PR #49, opened 2016-11-14
[core] Update to TypeScript 2.0

hippee-lee:

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "systemjs-builder": "^0.15.13",
     "through2": "^2.0.1",
     "tslint": "3.7.0-dev.5",
-    "typescript": "1.9.0-dev.20160505",
+    "typescript": "^2.0.9",
     "typings": "^1.4.0"
   },
   "scripts": {


package.json
```diff
@@ -59,7 +59,7 @@
     "systemjs-builder": "^0.15.13",
     "through2": "^2.0.1",
     "tslint": "3.7.0-dev.5",
-    "typescript": "1.9.0-dev.20160505",
+    "typescript": "^2.0.9",
```
adityarb88: do you think we should keep this version fixed? i remember last time there was an issue when we upgraded a typescript version. not sure what it was? cc: @youdz, @jeeyun

youdz: The issue was when we upgraded to a nightly build with breaking changes. I believe Typescript respects semantic versioning, so not being strict should be fine.

We do need a shrinkwrap pretty soon, though, now that we're open source and most tools we're using are stabilized. Things shouldn't be moving as fast as before, at least dependency-wise.

PR #58, opened 2016-11-15
[CORE] Copyright

youdz:
Adding the copyright to every single source file.


src/clarity/forms/demo/form-fields.demo.scss
```diff
@@ -1 +1,18 @@
+/*!
```
adityarb88: @youdz: since we are adding the copyright in a multi line comment, will it be included in the compiled file multiple times? just want to double check. should we use single line comments instead?

PR #66, opened 2016-11-16
[web] updating component page design

mathisscott:
Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ Safari
✔︎ IE10
✔︎ IE11
✔︎ Edge

• Implemented new header design for docs pages
• Added tablinks to every component page
• Changed workstream images to HTML5, CSS, NG shields
• Implemented in-progress versions of shields to denote work not completed
• Added javascript to handle scrolling inside docs content from header
• Added throttle to prevent event handlers from bogging down the browser
• Added responsive styling to component page header, some styling still needed on individual component pages
• changed bold + div items in components to h6 + para
• Added guideline anchors to every component page
• Implemented whitespace changes for headers and sections on the Get Started, What’s New, and Documentation pages
• changed all component-summary lines to h5s
• changed h4s to h5s as needed on component pages
• reviewed changes to what’s new and get started
• Fixed issue with sidenav highlighting on Get Started page
• Implement code that will allow #guidelines to be saved in browser history and directly linked to
• Implemented responsive changes for headers and sections on the Get Started, What’s New, and Documentation pages

screen shot 2016-11-15 at 7 20 49 pm


_data/componentlist.json
```diff
@@ -164,10 +164,10 @@
             "url": "/documentation/header",
             "text": "Header & Subnav",
             "type": "component",
-            "ux": 10,
-            "ui": 10,
+            "ux": 20,
+            "ui": 20,
```
adityarb88: I think this is still under implementation. Search and the Header Actions section are still pending, no?

mathisscott: My thinking was that we have a functional header. With the change away from the workstream bugs, it looks weird to have all this code on the page and the HTML, CSS, and Angular logo greyed out. It's confusing.

So my thinking is that if we have HTML/CSS/NG code that people can use, those should be set to `20`. The HTML/CSS/NG logos don't really represent the workstreams anymore. They just show if we have code there, what type it is, and if we don't are we planning to add it.

PR #72, opened 2016-11-16
chore(dropdown): remove dead code

wesleycho:
- Remove unused code

Signed-off-by: Wesley Cho <wesley.cho@gmail.com>

### DIFF
--- a/src/clarity/dropdown/dropdown.ts
+++ b/src/clarity/dropdown/dropdown.ts
@@ -81,18 +81,8 @@ export class Dropdown {
     //called on mouse clicks anywhere in the DOM.
     //Checks to see if the mouseclick happened on the host or outside
     @HostListener("document:click", ["$event.target"])
-    onMouseClick(target: any): void {
+    onMouseClick(): void {
         if (this._open) {
-            let current: any = target; //Get the element in the DOM on which the mouse was clicked
-            let dropdownHost: any = this.elementRef.nativeElement; //Get the current dropdown native HTML element
-
-            //Start checking if current and dropdownHost are equal. If not traverse to the parentNode and check again.
-            while (current) {
-                if (current === dropdownHost) {
-                    return;
-                }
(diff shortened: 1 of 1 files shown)


adityarb88:
Hi @wesleycho. Thanks for checking out Clarity. We really appreciate your suggestions/feedback. 

I tried removing the highlighted code in your PR locally but when I click on the `.dropdown-toggle`, the menu opens but its immediately closed because this logic, which was removed, detects if the click was inside or outside of the dropdown component. We want to close the menu when the click is outside of the dropdown component but also ignore click events which were inside of the dropdown component.

What this helped me realize is that there is a unit test in dropdown which is a false positive. I ll try to get that fixed ASAP. 

Thanks again!

wesleycho:
Ah shoot, I missed that there is a `return` - how come it isn't just checking for
```ts
if (dropdownHost.contains(target) {
```
?

adityarb88:
We tried that too but for I believe that `.contains` (please correct me if I am wrong here) only detects clicks on immediate children. If for example a `.dropdown-item` is clicked, the `.contains` method won't be able to detect it. Thats why we went with the while loop.

wesleycho:
I just selected an element in Chrome and did `$0.contains($0)`, it returned `true`.

adityarb88:
@wesleycho: I tested it again. It doesn't work. Try doing this:

```
if (dropdownHost.contains(current)) {
      console.log("Inside Host");
}
```

in the `onMouseClick` event handler and you wont get the console log when you click on the `.dropdown-item` but you will see the log when you click on the toggle or the menu

wesleycho:
Weird - going to close this PR since it looks like I erroneously deleted code and it seems like there's nothing worthwhile changing here atm.

I'll probably file some PRs cleaning up some other stuff I noticed at some point later this week when I fix my own ng2 issue at work :) .

adityarb88:
Thanks!

PR #78, opened 2016-11-16
[NG] Fix Dropdown Test

adityarb88:
Fixed a false positive test to detect clicks outside of the dropdown component.

Unit Tests:
Dropdown
    ✔ projects content
    ✔ adds the .dropdown class on clr-dropdown
    ✔ adds the .dropdown-toggle class on clrDropdownToggle
    ✔ adds the .dropdown-item class on clrDropdownItem
    ✔ supports clrMenuDirection option
    ✔ toggles the menu when clicked on the host
    ✔ closes the menu when clicked outside of the host
    ✔ supports clrMenuClosable option. Closes the dropdown menu when clrMenuClosable is set to true
    ✔ does not close the menu when a disabled item is clicked

Browser Tests:
Tested on: Chrome

Signed-off-by: Aditya Bhandari <arb492@nyu.edu>

### DIFF
--- a/src/clarity/dropdown/dropdown.spec.ts
+++ b/src/clarity/dropdown/dropdown.spec.ts
@@ -4,6 +4,9 @@ import {ClarityModule} from "../clarity.module";
 
 @Component({
     template: `
+        <button class="outside-click-test" (click)="outsideButtonClickHandler()">
+            Button to test clicks outside of the dropdown component
+        </button>
         <clr-dropdown [clrMenuPosition]="menuPosition" [clrCloseMenuOnItemClick]="menuClosable">
             <button class="btn btn-primary" type="button" clrDropdownToggle>
                 Dropdown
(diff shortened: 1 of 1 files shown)


src/clarity/dropdown/dropdown.spec.ts
```diff
@@ -112,10 +120,16 @@ describe("Dropdown", () => {
...
+        //check if the dropdown is closed
         expect(compiled.querySelector(".open")).toBeNull();
+
         //click outside the dropdown
-        fixture.nativeElement.click();
+        outsideButton.click();
+        //check if the click handler is triggered
+        expect(fixture.componentInstance.testCnt).toEqual(1);
+
         fixture.detectChanges();
```
youdz: Just to be safe, detect changes after a user action and before the assertion. So `outsideButton.click()`, then `fixture.detectChanges()`, and after that your two `expect`.

Issue #59, opened 2016-11-15
Provide a .forRoot() method on ClarityModule to prevent reimport of ClrResponsiveNavigationService in lazy-loaded feature modules


JohannesRudolph:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
`ClarityModule` needs a `.forRoot()` method so that it cooperates nicely with lazy-loaded feature modules. 

### Actual behavior
The current implementation results in separate instances of `ClrResponsiveNavigationService` per injector scope of lazy loaded feature modules

### Reproduction of behavior
For reference, see https://angular.io/docs/ts/latest/cookbook/ngmodule-faq.html#!#q-for-root

### Environment details

* **Angular version:** 2.2

* **Clarity version:** 0.7

* **OS and version:** OS X El Capitan

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

jaffoneh:
Thanks for filing this issue @JohannesRudolph. @youdz will try to get this fixed ASAP. Assigning this to him.

jaffoneh:
Actually, @hippee-lee will take care of this one!

PR #88, opened 2016-11-16
[ng] Fixes issue #59

hippee-lee:
- Adds forRoot() to ClarityModule
- Adds forRoot() call to specs

### DIFF
--- a/src/clarity/clarity.module.ts
+++ b/src/clarity/clarity.module.ts
@@ -54,10 +54,13 @@ import {ClrResponsiveNavigationService} from "./nav/clrResponsiveNavigationServi
         TABS_DIRECTIVES,
         WIZARD_DIRECTIVES,
         ICON_DIRECTIVES
-    ],
-    providers: [
-        ClrResponsiveNavigationService
     ]
 })
 export class ClarityModule {
+    static forRoot() {
+        return {
+            ngModule: ClarityModule,
+            providers: [ ClrResponsiveNavigationService ]
+        };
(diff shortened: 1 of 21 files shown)


youdz:
@vmware/clarity-maintainers: Please do not merge this in yet, it is a pretty big breaking change that needs to be properly document.

src/clarity/clarity.module.ts
```diff
@@ -54,10 +54,13 @@ import {ClrResponsiveNavigationService} from "./nav/clrResponsiveNavigationServi
...
-    ],
-    providers: [
-        ClrResponsiveNavigationService
     ]
 })
 export class ClarityModule {
+    static forRoot() {
+        return {
+            ngModule: ClarityModule,
+            providers: [ ClrResponsiveNavigationService ]
```
adityarb88: @hippee-lee, @youdz: Do we also need to include `ScrollingService` over here?

adityarb88: more details about `ScrollingService`: The service is used to support Modals in IE10. It adds the `.no-scrolling` class on `body` which sets `overflow: hidden` to avoid the content behind the backdrop from scrolling in IE10.

youdz: As far as I remember, ScrollingService doesn't need to be a singleton. Just like providers for other components (see Datagrid for instance), we can re-create it for each modal.

youdz:
Merged with #266.

PR #101, opened 2016-11-18
[NG] Select row by clicking the row in the datagrid

sivaah:
Currently the row can be selected only by clicking on the checkbox, this allows the user to click anywhere in the row to select the row.

Signed-off-by: Sivakumar Annamalai <kpa.siva@hotmail.com>

screen shot 2016-11-18 at 3 46 22 pm

### DIFF
--- a/src/clarity/datagrid/datagrid-row.ts
+++ b/src/clarity/datagrid/datagrid-row.ts
@@ -16,7 +16,8 @@ import {Selection} from "./providers/selection";
     `,
     host: {
         "[class.datagrid-row]": "true",
-        "[class.datagrid-selected]": "selected"
+        "[class.datagrid-selected]": "selected",
+        "(click)": "toggleSelection($event.target)"
     }
 })
 export class DatagridRow {
@@ -35,5 +36,15 @@ export class DatagridRow {
     }
     public set selected(value: boolean) {
         this.selection.setSelected(this.item, value);
+    }
+
+    /**
+     * Select / Unselect while click on row
+     */
+    public toggleSelection(element: HTMLElement) {
+        if (this.selection.selectable && element.nodeName !== "INPUT") {
+            this.selected = !this.selected;
+            return false;
+        }
     }
 }
--- a/src/clarity/datagrid/datagrid-row.spec.ts
+++ b/src/clarity/datagrid/datagrid-row.spec.ts
@@ -58,6 +58,20 @@ export default function(): void {
             expect(selectionProvider.current).toEqual([]);
         });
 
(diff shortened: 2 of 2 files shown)


youdz:
Hi @sivaah, thanks for the contribution. This is a request we received several times, but it turns out we cannot provide this functionality if we want to support any kind of content in Datagrid cells.
Many use cases involve having links inside of the datagrid, or action items to perform an action on a specific row. In these cases, we would end up with "conflicts" between click handlers that would result in a very frustrating experience for the end user. Kind of like a button within another button.

That said, I am going to review your code right now to show the way we work, even though this will not be merged in. Please do not lose time fixing what I point out.

I will leave this review open if you want to give us more feedback or share your thoughts on that, and you can close the PR when you want.

youdz:
On thing I forgot to mention is that this feature is quite easy to implement whenever you need it. It's just one binding in the template and you're done:
```html
<clr-dg-row #row *clrDgItems="let item of items" [clrDgItem]="item"
    (click)="row.selected = !row.selected">
```

src/clarity/datagrid/datagrid-row.ts
```diff
@@ -16,7 +16,8 @@ import {Selection} from "./providers/selection";
     `,
     host: {
         "[class.datagrid-row]": "true",
-        "[class.datagrid-selected]": "selected"
+        "[class.datagrid-selected]": "selected",
+        "(click)": "toggleSelection($event.target)"
```
youdz: You are adding a click handler on an inert element, which screenreaders will ignore.
For this to be accessible, you will need to add all the aria classes needed to describe the behavior you are implementing.

src/clarity/datagrid/datagrid-row.ts
```diff
@@ -36,4 +37,14 @@ export class DatagridRow {
     public set selected(value: boolean) {
         this.selection.setSelected(this.item, value);
     }
+
+    /**
+     * Select / Unselect while click on row
+     */
+    public toggleSelection(element: HTMLElement) {
+        if (this.selection.selectable && element.nodeName !== "INPUT") {
```
youdz: Working on HTMLElements means you cannot be rendered on the server side, or in webworkers, ... You need to work on Angular nodes and with the Renderer to be completely platform independent.

Admittedly, we still have a few of these in our existing code, but we're removing them as we go. Let's just not add more.

sivaah:
@youdz, I like the workaround you posted which solves my problem. I would have created an issue instead of stepping in and submitting a pull request :)

PR #120, opened 2016-11-19
update git clone instructions

gnomeontherun:
HTTPS is the preferred way to clone a repo by GitHub, as SSH only works if you have the key setup in your account.


gnomeontherun:
Appears when you edit in GitHub it assigns a different email to the commit, fixing...

PR #151, opened 2016-11-27
Null checking Subscription before unsubscribe

kucharzyk:
Subscribtion object should be checked for nullity before unsubscribe in ngOnDestroy methods

Without this fix unit test throws errors like this:

```
Chrome 54.0.2840 (Linux 0.0.0) Directive: shardisLarge should not make all elements yellow FAILED
	Error: Error in ./AppComponent class AppComponent - inline template:24:6 caused by: Cannot read property 'unsubscribe' of undefined
	TypeError: Cannot read property 'unsubscribe' of undefined
	    at Header.ngOnDestroy (webpack:///home/tomasz/dev/projects/shardis/~/clarity-angular/nav/header.js:73:0 <- src/main/typescript/test.ts:51016:27)
	    at Wrapper_Header.ngOnDestroy (/ClarityModule/Header/wrapper.ngfactory.js:14:16)
	    at CompiledTemplate.proxyViewClass.View_AppComponent0.destroyInternal (/DynamicTestModule/AppComponent/component.ngfactory.js:516:21)
	    at CompiledTemplate.proxyViewClass.AppView.destroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/view.js:143:0 <- src/main/typescript/test.ts:64598:14)
	    at CompiledTemplate.proxyViewClass.DebugAppView.destroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/view.js:371:0 <- src/main/typescript/test.ts:64826:38)
	    at CompiledTemplate.proxyViewClass.View_AppComponent_Host0.destroyInternal (/DynamicTestModule/AppComponent/host.ngfactory.js:33:19)
	    at CompiledTemplate.proxyViewClass.AppView.destroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/view.js:143:0 <- src/main/typescript/test.ts:64598:14)
	    at CompiledTemplate.proxyViewClass.DebugAppView.destroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/view.js:371:0 <- src/main/typescript/test.ts:64826:38)
	    at CompiledTemplate.proxyViewClass.AppView.detachAndDestroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/view.js:130:0 <- src/main/typescript/test.ts:64585:14)
	    at ComponentRef_.destroy (webpack:///home/tomasz/dev/projects/shardis/~/@angular/core/src/linker/component_factory.js:121:51 <- src/main/typescript/test.ts:30533:70)

```

### DIFF
--- a/src/clarity/layout/main-container.ts
+++ b/src/clarity/layout/main-container.ts
@@ -63,6 +63,6 @@ export class MainContainer implements OnDestroy, OnInit {
     }
 
     ngOnDestroy() {
-        this._subscription.unsubscribe();
+        if (this._subscription) { this._subscription.unsubscribe(); }
     }
 }
--- a/src/clarity/nav/header.ts
+++ b/src/clarity/nav/header.ts
@@ -93,6 +93,6 @@ export class Header implements OnDestroy, OnInit {
     }
 
     ngOnDestroy() {
-        this._subscription.unsubscribe();
+        if (this._subscription) { this._subscription.unsubscribe(); }
     }
 }


youdz:
This change seems fine by me, but I really believe the case where you have no subscription should absolutely never happen.

Even if it's in your unit tests, getting this error means you call `ngOnDestroy()` on your components but never `ngOnInit()`, which makes no sense. Either you don't call any of them because you're testing something that doesn't require the setup, of you call both of them because you're trying to test a compiled component.

I'm holding off on the :shipit: until someone else shares their opinion on this, because on one hand the change is innocuous, but on the other it encourages bad practice in unit tests by adding a safeguard where you shouldn't need one.

kucharzyk:
I've thougth that ngOnDestroy can't be called without ngOnInit but this error shows that I was wrong.
It makes no sense but it could happen.

After cleaning my tests and allways destroying fixture after tests they all pass now.  

My fix can avoid false errors like this for others.

jaffoneh:
@kucharzyk I would still be worried about introducing the change considering the statement @youdz made:

> It encourages bad practice in unit tests by adding a safeguard where you shouldn't need one.

What do you think?

youdz:
@kucharzyk: The fact that the absence of safeguard forced you to destroy fixtures after tests is my whole point. You should've been doing that before even using Clarity, or you end up with side effects and memory leaks, which can result in bugs that are horrible to identify. 

If we add your safeguard, people in the same situation as you will not see the error they already have, and they will only detect it much further down the line when it is harder to identify and fix.

Your answer pushes me towards not merging this in.

kucharzyk:
@youdz  You have right. Absence of safeguard forced me to properly destroy fixtuers.

On the other side some users may complain when stacktrace is pointing at error in Clarity class while real error is caused by improperly created test.

After all I think that You should't merge this. Creating safeguards for such situations makes little sense

PR #158, opened 2016-11-28
[NG] Modal keybinding cleanup

youdz:
We still close the modal when the user presses escape, but we can
remove our workaround now that Angular fixed the key filtering issue.

So much easier to read.


jeeyun:
👍

Issue #76, opened 2016-11-16
Document how to use dropdown in header-nav


JohannesRudolph:
**Select one ...**  (check one with "x")
```
[x] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
Documentation should include an example for how to correctly use a dropdown with a text label in the navbar.

### Actual behavior
Consider the following code: 
```
<clr-main-container>
  <clr-header>
    <div class="branding">
      <a routerLink="/home" class="nav-link">
        <span class="clr-icon clr-clarity-logo"></span>
        <span class="title">Logo</span>
      </a>
    </div>

    <div class="header-nav">
      <clr-dropdown clrMenuPosition="bottom-right">
        <a clrDropdownToggle class="nav-link" id="changeProject">
          <span class="nav-text">{{currentProject.name}}</span>
          <clr-icon shape="caret" class="icon-orient-down"></clr-icon>
        </a>
        <div class="dropdown-menu">
          <a clrDropdownItem *ngFor="let p of projects" [routerLink]="['p', p.tag]">{{p.name}}</a>
          <div class="dropdown-divider"></div>
          <a class="dropdown-item disabled" href="#">All</a>
        </div>
      </clr-dropdown>
...
```

This positions the caret in the center of the nav-link like so: 
![image](https://cloud.githubusercontent.com/assets/130103/20353818/4d0aa93a-ac1c-11e6-9e29-976fcb2bbe2e.png)

I can fix this using the following hack/workaround: 
```
.dropdown .dropdown-toggle:not(.btn) clr-icon[shape="caret"] {
  right: 0;
  left: initial;
}
```
Which then looks like this
![image](https://cloud.githubusercontent.com/assets/130103/20353724/11e71708-ac1c-11e6-876a-d5e3885a5f99.png)

I'm either doing something wrong OR this is a bug in clarity. 

### Reproduction of behavior
See above

### Environment details

* **Angular version:** 2.2.X

* **Clarity version:** 0.7

* **OS and version:** OS X el Capitan

* **Browser:** [Chrome 54, others likely affected as well]

adityarb88:
Hi @JohannesRudolph. Really appreciate your feedback. We will get this fixed this asap.

adityarb88:
[...]

PR #159, opened 2016-11-28
[UI] Header Actions Dropdown Text Example

adityarb88:
Added an example for Dropdown Toggle with Text in the Header

image

Fixes: #76 

### DIFF
--- a/src/clarity/nav/demo/header-types.demo.html
+++ b/src/clarity/nav/demo/header-types.demo.html
@@ -100,7 +100,7 @@
             </div>
         </header>
 
-        <header class="header-6">
+        <header class="header-6 space-below">
             <div class="branding">
                 <a href="javascript://" class="nav-link">
                     <clr-icon shape="vm-bug"></clr-icon>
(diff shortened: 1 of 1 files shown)


src/clarity/nav/demo/header-types.demo.html
```diff
@@ -113,6 +113,27 @@
...
+                <a href="javascript://" class="nav-link">
+                    <clr-icon shape="vm-bug"></clr-icon>
+                    <span class="title">Project Clarity</span>
+                </a>
+            </div>
+            <div class="header-actions">
+                <clr-dropdown class="dropdown bottom-right">
+                    <button class="nav-text" clrDropdownToggle>
+                        john.doe@vmware.com
+                        <clr-icon shape="caret" class="icon-orient-down"></clr-icon>
```
Shijir: This class `icon-orient-down` will be deprecated. Can you use `shape="caret down"`?

adityarb88: @shijir: oh okay. so this means that all dropdown examples will have to be updated.

Shijir: If they are still using `icon-orient-*` classess, we are going to replace them. I am not sure if you need to replace them in this PR though.

PR #163, opened 2016-11-29
[UI] Dropdown Caret Examples Update

adityarb88:
Updated carets in dropdowns to use this markup:
`<clr-icon shape=“caret” class=“icon-orient-down”></clr-icon>`

to:
`<clr-icon shape=“caret down”></clr-icon>`

Tested on: Chrome, IE10, IE11, Edge, Firefox, Safari

### DIFF
--- a/src/clarity/dropdown/_dropdown.clarity.scss
+++ b/src/clarity/dropdown/_dropdown.clarity.scss
@@ -35,7 +35,7 @@ $dropdown-white: clr-getColor(lightest);
                 margin: 0;
             }
 
-            clr-icon[shape="caret"] {
+            clr-icon[shape^="caret"] {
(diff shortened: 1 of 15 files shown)


src/clarity/dropdown/_dropdown.clarity.scss
```diff
@@ -35,7 +35,7 @@ $dropdown-white: clr-getColor(lightest);
                 margin: 0;
             }
 
-            clr-icon[shape="caret"] {
+            clr-icon[shape~="caret"] {
```
youdz: Nitpick: I think `[shape^="caret"]` is a tiny bit better here.
Same for the couple other ones further in that file.

PR #166, opened 2016-11-29
[web] Adds link to the Clarity Plunker Templage`

hippee-lee:


hippee-lee:
@scott - I didn't realize I was also committing changes to the node_module package.json files. Do I need to resubmit without them?

Issue #35, opened 2016-11-11
[WEB] Internal component pages in documentation can scroll left/right even when they shouldn't


jaffoneh:
**Select one ...**  (check one with "x")
```
[X ] bug
[ ] feature request
```

### Expected behavior
Scrolling should be smooth and work correctly on mobile devices.

### Actual behavior
Currently, pages like the card documentation page can scroll to show empty space to the right/left on the Clarity website. Some components have less space but can continue to scroll left/right as you try to scroll through the page.

### Reproduction of behavior
Visit the card documentation page or (almost) any other page on a mobile browser.

### Environment details

* **Angular version:** 2.0.X
2.1.0

* **Clarity version:** 
0.7.0

* **Browser:**
iOS 10 (Chrome or Safari)

PR #169, opened 2016-11-30
[web] fixing responsive horizontal scroll

mathisscott:
• had to go in page-by-page to address micro-issues that were causing horizontal scrolling
• also addressed what additional responsive issues I could
• some issues were in the demos. i will have a separate PR for that.
• tested some in iOS simulator. most testing was using Chrome DevTools

Tested in:
✔︎ Chrome
✔︎ IE10
✔︎ IE11
✔︎ Edge

Closes: #35


_includes/news/release_summary.html
```diff
@@ -32,6 +32,7 @@
         <a href="https://www.npmjs.com/package/clarity-angular"
            target="_blank">clarity-angular</a>
         {% if include.hasIcons %}
+           <br>
```
adityarb88: This might not work on Firefox. I remember using `br` for separating headers in the headers demo and it didn't work in Firefox. Might have to use margin

mathisscott: I will double check to see if it works on Firefox

PR #178, opened 2016-12-01
[core] Adds clarity version to generated css files

### LINKED ISSUE #110: on minified version of clarity, please have first characters depict the clarity.css version
```
[ ] bug
[ ] feature request
[X] enhancement
```

### Expected behavior
Being able to view clarity.css version when inspecting the clarity minified version.
This helps quite a lot to create consistency between different web-services using clarity-css.

### Actual behavior
upon inspecting clarity-ui.min.css, you will see the normalize version it uses:
"/*! normalize.css v4.2.0 | MIT License | github.com/necolas/normalize.css */html{font-family....."

### Reproduction of behavior
go to any web page using clarity-ui.min.css, inspect the file.
image

I can't tell which clarity version the clarity page uses :) I can tell which normalize version though! (although i don't care :( )

### Environment details
---

* **Angular version:** 2.0.X

* **Clarity version:** 
Any 

* **OS and version:** 
Any 

* **Browser:**
All

hippee-lee:
- Changes sass tasks to preprocess clarity-ui & clarity-icons css output with the clarity version number
- closes #110

### DIFF
--- a/build/tasks/sass.js
+++ b/build/tasks/sass.js
@@ -12,6 +12,8 @@ var autoprefixer = require('gulp-autoprefixer');
 var sass = require("gulp-sass");
 var stylestats = require('gulp-stylestats');
 var runSequence = require('run-sequence');
+var preprocess = require('gulp-preprocess');
+var util = require('gulp-util');
 
 var compressed = { outputStyle: "compressed" };
 var uncompressed = { sourceComments: 'map', errLogToConsole: true, sourceMap: 'sass' };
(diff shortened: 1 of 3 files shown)


src/clarity/main.scss
```diff
@@ -1,9 +1,13 @@
+
 /*
```
youdz: I just noticed this isn't loud, which means the copyright probably isn't in the minified CSS. Do you mind fixing this as part the your commit?

build/tasks/sass.js
```diff
@@ -36,6 +37,16 @@ var iconFiles = "src/icons/**/*.scss";
 
 var assets = ["src/clarity/img/*.png","src/clarity/img/*.svg"];
 
+/*
+ * Adds the current clarity version to generated css files
+ * Depends on this being in the *.scss files: \/* @echo VERSION *\/
+ * NOTE: for this to work we must generate a loud comment -> /*!
+ */
+var env = require('gulp-env');
+var pjson = require('../../package.json'); // Grab the projects package.json w/ version info
```
youdz: No need to parse the `package.json`, I believe `util.env.version;` is already set when running any npm script.

See [here](https://github.com/vmware/clarity/blob/master/build/tasks/publish.js#L15) for an example of how we do it when publishing to NPM.

hippee-lee:
@eudes changes made but my (force) push had an effect on your comments.

One thing to note and I'm not sure if its important - when using **util.env.version** substitution only happens when we publish **npm run publish:all**, when I run **npm run build** standalone, **util.env.version** is undefined.

src/clarity/main.scss
hippee-lee: Fixed.

build/tasks/sass.js
hippee-lee: One thing to note and I'm not sure if its important - when using **util.env.version** substitution only happens when we publish **npm run publish:all**, when I run **npm run build** standalone, **util.env.version** is undefined.

Issue #152, opened 2016-11-28
Wizard : it misses an event for cancel action


Rebolon:
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
When we use the Wizard component, we expect to bind an event to the CANCEL button

### Actual behavior
We cannot bind event to the cancel button whereas we can with the FINISH button

### Evolution
WizardPage class could have an onCancel method that could be bindable to the clr-wizard-page with a clrWizardPageOnCancel

Rebolon:
For those who wants to know why we may need this : 
 * on CANCEL/CLOSE we might want to do some action (route the user to another page, display alert box...) and for instance it hard to do that kind of features

Rebolon:
I did this PR that implement the wished feature https://github.com/vmware/clarity/pull/179
but the functional tests failed (see comment of the PR)

jaffoneh:
@Rebolon wanted to double-check and confirm you'd be okay with including your name on the [community page](https://vmware.github.io/clarity/news/) as a contributor once this change goes in. We would include it as: `Benjamin RICHARD`

Rebolon:
@jaffoneh it's ok for me it would be a pleasure. Thanks

PR #179, opened 2016-12-02
Feature/wizard oncancel

Rebolon:
for issue #152

need help on testing the feature: 
 * i want to test that the property of the component has been modified by the onCancel
 * i want to test that the _close method has been called

but my test doesn't work coz i have certainly missed something around the component


Rebolon:
The problem of this implementation is that the output is on the WizardPage whereas it should be on the Wizard. Will work on it and give a new implementation soon

PR #183, opened 2016-12-05
Implement an onCancel event fo Wizard component

Rebolon:
For issue #152

This is a new implementation of an onCancel event in wizard component (first try was on PR 179).
The first implementation was wrong because it was binded on wizardPage instead of main component.

I still need help on testing the feature:
 * i want to test that the property of the component has been modified by the onCancel method
 * i want to test that the _close method has been called after clicking on cancel

but my test doesn't work coz i have certainly missed something around the component

Signed-off-by: Benjamin RICHARD <benjamin.richard@ext.euronews.com>

### DIFF
--- a/src/clarity/wizard/wizard.spec.ts
+++ b/src/clarity/wizard/wizard.spec.ts
@@ -33,7 +33,10 @@ class BasicWizard {
 
 @Component({
     template: `
-    <clr-wizard [(clrWizardOpen)]="open">
+    <clr-wizard 
+        [(clrWizardOpen)]="open"
+         (clrWizardOnCancel)="myOnCancel($event)">
+         
          <div class="wizard-title">
             New Virtual Machine
          </div>
@@ -89,6 +92,7 @@ class AdvancedWizard {
     nextDisabled: boolean = false;
     content1: String = "Content1";
     content3: String = "Content3";
+    hasBeenCanceled = false;
 
     myOnLoad(): void {
         this.content1 = "This Works Better";
@@ -100,6 +104,10 @@ class AdvancedWizard {
 
     myOnCommit(event: any): void {
         event.preventDefault();
+    }
+
+    myOnCancel(event: any): void {
+        this.hasBeenCanceled = true;
     }
 };
 
@@ -117,6 +125,12 @@ describe("Wizard", () => {
     let moveToPrevious: Function = function (el: any): void {
         let back: HTMLElement = el.querySelector(".btn-outline");
         back.click();
+        fixture.detectChanges();
+    };
+
+    let doCancel: Function = function(el: any): void {
+        let cancel: HTMLElement = el.querySelector(".close");
+        cancel.click();
         fixture.detectChanges();
     };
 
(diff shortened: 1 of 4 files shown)


src/clarity/wizard/wizard.ts
```diff
@@ -124,6 +128,16 @@ export class Wizard extends Tabs {
         this._openChanged.emit(false);
     }
 
+    // _close --
+    //
+    // This is a private function that is called on the click of the close / cancel
+    // button and emits the onCancel event of the active tab.
+    _close(): void {
```
youdz: This isn't called when the user presses the Escape key, even though it do so closes the wizard. Can you add a listener for it? You can check out how [the modal](https://github.com/vmware/clarity/blob/master/src/clarity/modal/modal.ts#L78)  does it for instance.

src/clarity/wizard/wizard.spec.ts
```diff
@@ -466,6 +480,15 @@ describe("Wizard", () => {
             expect(tab3).toBeDefined();
             expect(tab4).toBeDefined();
         });
+
+        it("calls the user-defined onCancel handler when the Cancel button is clicked", () => {
+            fixture.detectChanges();
+            doCancel(compiled);
+
+            expect(fixture.componentInstance.hasBeenCanceled).toBe(true);
+            spyOn(instance, "_close");
```
youdz: Please avoid testing the internal implementation of the component. I think making sure the output is emitted is enough for this test.

src/clarity/wizard/wizard.ts
Rebolon: Ok it's now updated

src/clarity/wizard/wizard.spec.ts
Rebolon: i changed the test to listen to the wizard instance .onCancel method but it doesn't detect it.
If i check the component instance the test succeed : 

`

     const myComponentInstance = fixture.componentInstance;
     spyOn(myComponentInstance, "myOnCancel");
     doCancel(compiled);
     expect(myComponentInstance.myOnCancel).toHaveBeenCalled();
`

How could i modify the current test to make it works ?

src/clarity/wizard/wizard.spec.ts
```diff
@@ -466,6 +480,14 @@ describe("Wizard", () => {
             expect(tab3).toBeDefined();
             expect(tab4).toBeDefined();
         });
+
+        fit("calls the user-defined onCancel handler when the Cancel button is clicked", () => {
+            fixture.detectChanges();
+
+            spyOn(instance, "onCancel");
+            doCancel(compiled);
+            expect(instance.onCancel).toHaveBeenCalled();
```
youdz: I think there was a misunderstanding. What I meant here is that 
```typescript
doCancel(compiled);
expect(fixture.componentInstance.hasBeenCanceled).toBe(true);
```
was absolutely fine. That's all you need to test your feature.

Here, you're testing if an _object_ (in this case an EventEmitter) has been called, which doesn't make sense.
Your initial version was good, you just needed to remove the extra test on internal implementation. 😄

Rebolon: ok, sorry i'll fix itnow

Rebolon:
I have done the requested changes and rebased the history.

PR #196, opened 2016-12-07
[ICONS] created a class for icon error color

Shijir:
- made .icon-color-danger the same color as .icon-color-warning
- added .icon-color-error that is the same color as .icon-color-warning
- updated the icon colors demo to reflect the changes

### DIFF
--- a/src/icons/clarity-icons.scss
+++ b/src/icons/clarity-icons.scss
@@ -6,8 +6,9 @@
 
 $clr-icon-size: 16px;
 $clr-icon-color-success: #318700;
-$clr-icon-color-danger: #a32100;
+$clr-icon-color-danger: #e62700;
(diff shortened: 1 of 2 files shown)


src/icons/clarity-icons.scss
```diff
@@ -50,6 +51,9 @@ clr-icon {
     &.icon-color-success {
         @include setIconColor($clr-icon-color-success);
     }
+    &.icon-color-error {
+        @include setIconColor($clr-icon-color-error);
```
mathisscott: It seems like there is an opportunity to DRY this code out here. I will approve for now. But we should try to consolidate, instead of replicate.

Shijir: Yes. I agree. I will look into my options to DRY the code. One them could be SASS `for` loop, I think.

Issue #155, opened 2016-11-28
Wizard: should be able to hide close ("x") button


yaooluu:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

There should be a way to hide the close ("x") button on top-right of a wizard. This is especially an issue when there's an alert showed on top.

### Actual behavior

There is no way to hide the close ("x") button on top-right of a wizard

### Reproduction of behavior

Any Clarity wizard can reproduce this.

### Environment details

* **Angular version:** 2.1.1

* **Clarity version:** 0.6.0

* **OS and version:** OS X El Captain 10.11.6

* **Browser:** [ Electron ]

![image](https://cloud.githubusercontent.com/assets/8813134/20681726/ae4928e8-b559-11e6-91d4-8546db0c2175.png)

jaffoneh:
@yaooluu do you want to hide/show only when the error is there or are you looking for a way to not include the "X" at all?

On another note, could you provide more details as to why you're showing the error on top of the title in the Wizard. What is the use case where you need a page-wide error vs. inline errors to where the issues are with the possibility of having an error icon on the name of the page in the left-side TOC?

yaooluu:
Hi @jaffoneh, the use case here is that the error banner shows errors that are returned from backend after use submitted the form by clicking "Next" on wizard page. This is a convention we've followed in a product series.

I'm looking for a way not to include the "X" at all. Like the option we have for Modal `clrModalClosable`.

jaffoneh:
cc @youdz to provide an opinion here.

youdz:
@jaffoneh, @yaooluu: No, `[clrModalClosable]` will not work right now because the wizard currently _wraps_ the modal and doesn't offer a delegate for it. 

Yes, this is a valid request that we should address. It's as simple as offering a delegate `[clrWizardClosable]` on the wizard component that just passes it to the modal through `[clrModalClosable]`.

jaffoneh:
Thanks, @youdz. @adityarb88 could you take a look and see what we can do there?

adityarb88:
@jaffoneh: Sure. I will add this to my list

PR #202, opened 2016-12-07
[ng] adding clrWizardClosable attribute to wizards

mathisscott:
• added clrWizardClosable flag to wizard component
• this delegates to clrModalClosable
• effectively removes X from top-right corner of modal
• defaults to true, which is the current default
• added unit tests
• moved emission of onCancel event on close out of the private function
• when we moved the close functionality to be a delegate of the modal closing functionality, the onCancel event was not being fired

Tested in:
✔︎ Chrome

Closes: #155

### DIFF
--- a/src/clarity/wizard/wizard.spec.ts
+++ b/src/clarity/wizard/wizard.spec.ts
@@ -11,7 +11,7 @@ import {Wizard} from "./wizard";
 
 @Component({
     template: `
-    <clr-wizard [(clrWizardOpen)]="open">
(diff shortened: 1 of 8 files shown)


src/clarity/wizard/wizard.spec.ts
```diff
@@ -107,6 +106,8 @@ class AdvancedWizard {
     }
 
     myOnCancel(event: any): void {
+        console.log("here");
```
youdz: Nope. 😛

PR #203, opened 2016-12-07
[ui] Adds indeterminate state on checkboxes

### LINKED ISSUE #167: Need to add an indeterminate checkbox state. Design spec attached.
indeterminate checkbox

```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
Needed in a tree view to denote a group/folder that has some items checked, but not all.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

hippee-lee:
- closes #167


src/clarity/checkboxes/_checkbox.clarity.scss
```diff
@@ -0,0 +1,19 @@
...
+      width: 75px;
+      height: 75px;
+      border: 1px solid darkblue;
+    }
+
+    &:after {
+      width: 50px;
+      height: 50px;
+      border: 1px solid darkorange;
+    }
```
dragosrusu: would it be a good idea to have standard sizes in $clr? or % type of sizes instead maybe?

mathisscott: Re: BEM. Great suggestion, @dragosrusu. For a number of reasons worthy of a blogpost, the team used Bootstrap as a base. So our classnames tend to reflect their direction. If/when we move away from Bootstrap, BEM, SMACSS, and OOCSS are all directions we should investigate.

re: standard sizes -- we usually try to do that (use standard sizes based on our vertical baseline). Kind of odd that there's a hardcoded 'darkorange' border there.

Thanks for the comments and suggestions!

hippee-lee: Thanks for the review @dragosrusu and nice catch! This code wan't removed when I cleaned up after dev.

src/clarity/forms/_forms.clarity.scss
```diff
@@ -434,6 +433,23 @@ $clr-custom-checkbox-radio-top: ($clr_baselineRem_1 - $clr-checkbox-radio-height
...
+        input[type="checkbox"]:indeterminate + label::after {
+            position: absolute;
+            content: '';
+            display: inline-block;
+            height: ($clr-checkbox-radio-height / 4 + 1);
+            width: $clr-checkbox-radio-height / 2;
+            border-left: none;
+            border-bottom: 2px solid white;
+            top: ($clr_baselineRem_1 - $clr-checkbox-radio-height) / 2;
+            left: $clr-checkbox-radio-height / 4;
```
mathisscott: These repetitive operations should be in a scoped var. 

`$clr-checkbox-radio-height / 4`

`$clr-checkbox-radio-height / 2`

src/clarity/checkboxes/checkbox.ts
```diff
@@ -34,29 +35,54 @@ let latestId = 0;
...
+
+    public get indeterminate () {
+        return this._indeterminate;
+    }
+
+    @Input("clrIndeterminate")
+    public set indeterminate(value: boolean) {
+        this.checked = false;
+        this._indeterminate = value;
+        this.indeterminateChange.emit(this._indeterminate);
```
mathisscott: So the expectation is that whenever a checkbox is set to indeterminate state that its checked value will be false?

We need a unit test to validate this.

Same for the other way, if a checkbox's checked state is set to false or true we need to make sure its indeterminate value is false.

src/clarity/forms/_forms.clarity.scss
```diff
@@ -434,6 +433,23 @@ $clr-custom-checkbox-radio-top: ($clr_baselineRem_1 - $clr-checkbox-radio-height
             display: inline-block;
         }
 
+        input[type="checkbox"]:indeterminate + label::before {
```
mathisscott: My assumption is that these styles work outside of the angular code. If that is the case, can we make sure to put a static example on the website so that people who aren't using Clarity-NG will know how to use these?

And if it isn't the case, can we make sure that it is?

Please validate that this selector will work in IE10-11. I have a bad feeling about it...

hippee-lee: I double checked that this works in IE 10 / 11: 
ie_10_checkbox
ie_11_checkbox

src/clarity/checkboxes/checkbox.spec.ts
```diff
@@ -134,6 +143,12 @@ describe("Checkbox", () => {
         expect(fixture.nativeElement.querySelector(".checkbox-inline")).not.toBeNull();
     });
 
+    it("supports indeterminate state", () => {
+        createTestComponent(IndetermiateCheckbox);
+        expect(fixture.nativeElement.querySelector(".checkbox-indeterminate")).not.toBeNull();
+        expect(fixture.componentInstance.checkboxInstance.indeterminate).toBe(true);
```
hippee-lee: @mathisscott - here is the test I wrote for indeterminate state. Good call on checking that its false - I'll add one for that.

src/clarity/forms/_forms.clarity.scss
hippee-lee: Yes, this code will work outside of clarity-angular **but,** the only way to set a checkbox to indeterminate state is with javascript. One cannot do 
`<input type="checkbox" indeterminate>`
like you can for checked. 
When it was brought to my attention this was referenced:
[Checkbox Visual States](https://css-tricks.com/indeterminate-checkboxes/)
[Pseudoclass Spec / Example](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate)
I'll add an example to the website. I need to wait until this is release though correct?

src/clarity/forms/_forms.clarity.scss
```diff
@@ -19,6 +18,8 @@ $clr-form-field-disabled-color: $clr-gray !default;
 $clr-checkbox-radio-height: $clr-icon-dimension-sm !default;
 $clr-form-icon-dim: $clr-icon-dimension-sm !default;
 $clr-custom-checkbox-radio-top: ($clr_baselineRem_1 - $clr-checkbox-radio-height) / 2 !default;
+$clr-checkbox-height: ($clr-checkbox-radio-height / 4 + 1);
```
adityarb88: Is this the height of the checkbox or is it just the height of the hypen/dash symbol in the checkbox in the indeterminate state? If it is the second one, we should change the variable name to reflect that. Same for the width on the next line.

src/clarity/forms/_forms.clarity.scss
adityarb88: @hippee-lee: Just one small thing. If you are planning to add this demo to the website, we should use Clarity buttons

hippee-lee: Updated the demo to use Clarity buttons.

youdz:
Sorry, I just realized I never put in there what I asked you internally. 

This does look much much better with standard CSS selectors rather than custom classes. Very good job.

Issue #212, opened 2016-12-09
[UI] Input Field validation does not occupy entire width on smaller screens


adityarb88:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Input field validation occupies the entire width on smaller screens.

### Actual behavior

![image](https://cloud.githubusercontent.com/assets/1426805/21053658/4c2a59be-bddf-11e6-8b64-e0ffca1e2b4f.png)

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** All

* **OS and version:** All

* **Browser:** [all]

PR #213, opened 2016-12-09
[UI] Input Validation Bug Fix

adityarb88:
Fixed bug where Validated Inputs didn’t occupy 100% of the width on smaller screens

Before:
image

After:
image

Tested on:
Chrome,
IE 10,
IE 11,
Edge,
Safari,
Firefox

Fixes: #212 

### DIFF
--- a/src/clarity/forms/_forms.clarity.scss
+++ b/src/clarity/forms/_forms.clarity.scss
@@ -224,9 +224,21 @@ $clr-custom-checkbox-radio-top: ($clr_baselineRem_1 - $clr-checkbox-radio-height
 
             //All children have a top and bottom margin regarless if they are direct descendants or if they are
             //in a grid
-            label, span, #{$styledInputs}, .tooltip-validation, textarea,
-            .select, .checkbox-inline, .radio-inline, .checkbox, .radio,
-            .toggle-switch, button, a, #{$styledInputs1}, .btn {
+            label,
+            span,
+            #{$styledInputs},
+            .tooltip-validation,
+            textarea,
+            .select,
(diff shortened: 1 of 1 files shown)


src/clarity/forms/_forms.clarity.scss
```diff
@@ -258,34 +270,39 @@ $clr-custom-checkbox-radio-top: ($clr_baselineRem_1 - $clr-checkbox-radio-height
                 padding-left: 0;
                 margin-bottom: $clr_baselineRem_1;
 
-                & > label:first-of-type {
-                    position: relative;
+                & > label:first-child,
+                & > label:not(:first-child),
```
Shijir: Do both first-child and non-first child labels have the same style, ` flex: 1 1 100%;`? If then, I think we could just use `& > label`? I might be wrong here.

mathisscott: +1

adityarb88: @Shijir, @mathisscott: Nope. Since we have different flex properties for them on larger screens we have to override them both on small screens. Using only `& > label` has less specificity than mentioning the `first-child` rule and the previous styles override the behavior.

mathisscott: hmm. not loving it. i will be happy(er) when we can refactor to a CSS arch that is easier to maintain.

adityarb88: @mathisscott: yes and this is why i mentioned the other day to you that we might need to refactor forms because I actually didn't like this too.

PR #216, opened 2016-12-10
[ICONS] created component for an icon in icon-selection

Shijir:
- using a component, we don't need to manually repeat the shape names after each `clr-icon` in icon-selection-demo

- also, it will make it much easier to have interactivity on them later on. (i.e: click and show modal that shows their code usages, etc.) 


Shijir:
@mathisscott I didn't include the component in the the `exports` section of the module. So this component is private and should be only available in its iconography demo module.

mathisscott:
I was more talking about explicitly calling out that this is for the demo in the selector/directive name. That way anyone working with the code wouldn't get confused and think this is something they could use outside of the demos.

src/clarity/iconography/demo/icon-single-select.html
```diff
@@ -0,0 +1 @@
+<clr-icon [attr.shape]="shape"></clr-icon><span class="icon-shape-name">{{shape}}</span>
```
youdz: Just a remark, not a requested change: when your template is literally one line, you can just inline it in the Typescript file. It's easier to read for everyone. 😛

src/clarity/iconography/demo/icon-selection.demo.html
```diff
@@ -1,220 +1,220 @@
 <div class="clr-icon-selection">
     <div class="row">
         <div class="clr-icon-item col-xs-12 col-sm-6 col-md-4 col-lg-3">
-            <clr-icon shape="home"></clr-icon><span class="icon-shape-name">home</span>
+            <clr-icon-select [clrIconShape]="'home'"></clr-icon-select>
         </div>
```
youdz:
```html
<div *ngFor="let shape of shapes" class="clr-icon-item col-xs-12 col-sm-6 col-md-4 col-lg-3">
    <clr-icon-select [clrIconShape]="shape"></clr-icon-select>
</div>
```

Seriously.

src/clarity/iconography/demo/icon-single-select.html
Shijir: This component was created to remove the effort manually copy pasting each icon names next to the icons. Also, this template might grow big as we want to show icon usage of each icon in a modal.

src/clarity/iconography/demo/icon-selection.demo.html
Shijir: This is awesome! I will implement this now!

Shijir:
@mathisscott we no longer needed that new component now so I removed it. Please review the current updates.

PR #221, opened 2016-12-12
[UI] Datagrid column headers wrap fix

### LINKED ISSUE #207: [ng] Datagrid column headers wrap
```
[ ] bug
[ ] feature request
[X] enhancement
```

### Expected behavior

The data grid should either not wrap column headers or should wrap them intelligently. When a column header is wrapped, the filter icons across all headers should be positioned relative to the extra line height from the wrapped header text.

### Actual behavior
Data grid column headers wrap haphazardly and the filter icons are position as if the column headers are just a single line.

### Reproduction of behavior

https://embed.plnkr.co/6PaGJK/

screen shot 2016-12-08 at 10 58 02 am

### Environment details

* **Angular version:** 2.2.0

* **Clarity version:** 0.7.3

* **OS and version:** 

* **Browser:** [all]

Robertmw:
- Solve scaling column separator
- Solve placement of filter icon and title

- Closes #207

Signed-off-by: Robert Pamfile <robert.pamfile@gmail.com>

BY BYTEX ROMANIA

### DIFF
--- a/src/clarity/datagrid/_datagrid.clarity.scss
+++ b/src/clarity/datagrid/_datagrid.clarity.scss
@@ -89,11 +89,12 @@
                 position: absolute;
                 // Actual formula for top is: cellHeight - borderHeight / 2
                 // which is: baselineRem(1.5 - 1 / 2)
-                top: $clr_baselineRem_0_25;
+                top: 0;
                 right: 0;
-                height: $clr_baselineRem_1;
-                width: 0;
-                border-left: 1px solid $clr-lighter-midtone-gray;
+                bottom: 0;
+                width: 1px;
+                margin: $clr_baselineRem_0_25 0;
+                background-color: $clr-lighter-midtone-gray;
             }
 
             .datagrid-column-title {
@@ -104,12 +105,24 @@
                 }
             }
 
+            .datagrid-column-flex {
+                display: flex;
+            }
+
+            clr-dg-filter,
+            clr-dg-string-filter {
+                display: flex;
+                order: 99;
+                margin-left: auto;
+            }
+
             &.asc, &.desc {
                 font-weight: 600;
 
-                &::after {
+                .datagrid-column-flex::after {
                     content: "";
-                    display: inline-block;
+                    display: flex;
+                    flex: 0 0 auto;
                     vertical-align: middle;
                     width: $clr-datagrid-icon-size;
                     height: $clr-datagrid-icon-size;
@@ -119,11 +132,15 @@
                 }
             }
 
-            &.asc::after {
-                background-image: generateUpArrowIcon($action-blues-dark-midtone);
+            &.asc {
+                .datagrid-column-flex::after {
+                    background-image: generateUpArrowIcon($action-blues-dark-midtone);
+                }
             }
-            &.desc::after {
-                background-image: generateDownArrowIcon($action-blues-dark-midtone);
+            &.desc {
+                .datagrid-column-flex::after {
+                    background-image: generateDownArrowIcon($action-blues-dark-midtone);
+                }
             }
 
             .datagrid-filter-toggle {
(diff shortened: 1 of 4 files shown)


jaffoneh:
Hi @Robertmw: thanks for making a contribution, we appreciate it!

Seems like the commit belongs to robertp@plumgrid.com but it is being signed off by robert.pamfile@gmail.com. That's why the @vmwclabot is not happy. Could you sign it off with the correct email address?

Also, **could you provide a before/after images of the change**, this will help reviewers as they look at this. Thanks!

dragosrusu:
@Robertmw : please also attached "BYTEX ROMANIA" in your commit message. Thank you.

src/clarity/datagrid/datagrid-column.ts
```diff
@@ -16,14 +16,18 @@ import {Sort} from "./providers/sort";
...
         <!-- I'm really not happy with that select since it's not very scalable -->
-        <ng-content select="clr-dg-filter, clr-dg-string-filter"></ng-content>
-        
-        <clr-dg-string-filter *ngIf="field && !customFilter" 
-            [clrDgStringFilter]="defaultFieldFilter"></clr-dg-string-filter>
-            
-        <button class="datagrid-column-title" [disabled]="!sortable" (click)="sort()">
-            <ng-content></ng-content>
-        </button>
+        <div class="datagrid-column-flex">
```
youdz: Why add an extra wrapper here?

From what I see in the CSS and here, you should just add `display: flex;` for `.datagrid-column` instead of adding an extra wrapper and an extra class.

Robertmw: Adding `display:flex;` to `.datagrid-column` will break the flow. 

Indeed, flex can be added to `.datagrid-column` but that will require a total refactor for the whole datagrid component CSS (moving from `display: table-row;`  to `display: flex;`). If you guys are ok with such changes, I can start working on that.

@adityarb88 @mathisscott

mathisscott: I would rather not refactor the entire datagrid component CSS... What are your thoughts, @youdz?

youdz: My bad, you're right. I forgot about the `display: table-head` on the columns, that was obvious.

I agree with Scott, refactor here is overkill, what you have now is fine.

Robertmw:
Ok, I updated the email address and below you can find the before/after printscreens

Before fix:
image

After fix:
image

youdz:
This looks almost good to me, thanks for handling that. 👍 

There is still one last thing that needs to be fixed, it's the position of the "sort" arrow when a column is currently sorted. This is what's happening right now, with your change:
image

I believe it's as simple as changing the current pseudo-element selector from 
```css
.datagrid .datagrid-column.asc::after
```
to 
```css
.datagrid .datagrid-column.asc .datagrid-column-flex::after
```
and probably add some flex property to make sure the arrow always keeps the same size.

jaffoneh:
@Robertmw once this PR goes in, mind adding your name `Robert Pamfile` to the contributors page here: https://vmware.github.io/clarity/community/ ?

src/clarity/datagrid/_datagrid.clarity.scss
```diff
@@ -104,10 +105,21 @@
...
+                display: flex;
+                order: 99;
+                margin-left: auto;
+            }
+
             &.asc, &.desc {
                 font-weight: 600;
 
-                &::after {
+                .datagrid-column-flex::after {
```
youdz: This is missing `flex: 0 0 auto` The arrow icon shouldn't flex.

Checked your branch right now, this is what it looks like at the moment:
image
And this is with `flex: 0 0 auto`:
image

Robertmw: Done, @youdz please review again. Thank you!

Robertmw:
Another before/after screenshots with ascendant icon included.

Before:
image

After:
image

src/clarity/datagrid/_datagrid.clarity.scss
youdz: Adding `display: flex` to this element is not needed, it doesn't need any display property set since the container being flew is all we need. But that's a nitpick, it doesn't change anything so this is good to go.

Issue #220, opened 2016-12-12
ClarityModule is not an NgModule


BugsyFTW:
**Select one ...**  (check one with "x")
```
[ x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When using ng serve it should compile the app.

### Actual behavior
When i type ng serve it says that the ClarityModule is not an NgModule, i've followed correctly the tutorial on installation.

```
** NG Live Development Server is running on http://localhost:4200. **

 10% building modules 3/3 modules 0 active(node:844) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ClarityModule is not an NgModule
 
21% building modules 94/94 modules 0 active
```

adityarb88:
Hi @BugsyFTW. Does this happen when you use Angular-CLI version of the clarity-seed?

cc: @Shijir. It looks like an issue being raised over here: https://github.com/angular/angular-cli/issues/3426

hansl:
Hi @adityarb88. I'm the lead on the Angular-CLI project.

Basically this error likely happens because you did not publish `metadata.json` files to npm. These files contain static information for your module and components.

To generate these files, follow these steps: https://angular.io/docs/ts/latest/cookbook/aot-compiler.html (basically running `ngc` instead of `tsc` to build your project). You don't have to release the `.ngfactory.ts` files (you can use `skipTemplateCodegen: true`), just the metadata json files.

Cheers!

jeeyun:
@hansl is right. We are actively working on supporting AoT compilation (#62) right now and plan to publish the generated `metadata.json` files as soon as we have it working. @BugsyFTW feel free to subscribe to the issue on AoT compilation so you'd know as soon as we have it out.

PR #255, opened 2016-12-16
[core] Refactor the build for AoT compilation

### LINKED ISSUE #62: [NG] Support AoT compilation
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

Applications using Clarity components should be able to use AoT compilation.

### Actual behavior

Some of the Clarity components cannot be compiled ahead of time, mostly because they modify HTML elements directly instead of going through the renderer.

### Reproduction of behavior

Pretty hard to reproduce in a plunker...

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 0.7.0

* **OS and version:** all

* **Browser:** all

jeeyun:

Fixes #62, #220. 

Here's a refactoring of our project to support AoT compilation. Some notable changes are:
- Code changes to make our components AoT compatible:
   - make all inputs and outputs public
   - removed usage of our animation functions in components; the code is still there but not being used; for now the animation is hard-coded into the components (the ngc compiler couldn't statically analyze the functions passed into the component decorator; there were some suggestions to export them as named function but I couldn't get it working; let me know if anyone is able to get this working)
   - other minor edits to get ngc compiler to compile
- Moved demo components to under app (we previously had it nested under each component which required a lot of renaming and shifting things around to produce the npm packages)
- Reduced some of the "automagic" in our build flow:
   - renamed `src/clarity` to `src/clarity-angular`
   - no more auto re-exporting in `src/clarity-angular/index.ts` in favor of explicitly adding export statements for new components
- Now in dev mode, we no longer inline the templates (had to re-add module.id in most components to support this). this allows for the typescript compilation to not depend on sass compilation on watch tasks. prod mode will inline the template as before.
- Added build step to compile our components with ngc compiler. Our clarity-angular npm package will change to include es2015 compiled code (for AoT compilation) along with `metadata.json` files, original source code in `src` folder (as before), and `clarity-angular.min.js` in commonjs format (as before).
- No more demo npm package.

I left some of the unused gulp tasks commented out. Once we are sure we won't need them, I can delete those files.


jeeyun:
Made some additional changes, which I've squashed and merged into one commit:
-  removed the added module.id from clarity components (I couldn't get the components to be happy with webpack in dev mode. I think this issue is related: https://github.com/angular/angular-cli/issues/3576). Demo app and components will still use module.id. Clarity components will always be inlined, even in dev mode.
- now producing additional `clarity-angular.umd.js` file. I left the `clarity-angular.min.js` for now, so that is included in the npm package as before. apps using systemjs will still need to be updated to include the umd file in the config.

build/tasks/publish.js
```diff
@@ -106,28 +104,30 @@ gulp.task("npm:ui:readme", function () {
...
 
 /**
  * Preparing the clarity-angular package
  */
 
 /**
  * The deliverables for clarity-angular are:
- *   - the compiled js files in commonJS format
- *   - the minified JS bundle
+ *   - the compiled js files in es2015 format
```
youdz: I'm a bit scared of this. I think the standard is ES5 with ES2015 module format only, not the full features. Otherwise you either force Clarity's user to use yet another transpiler to ES5, or they forget about it and their app will only run on a handful of browsers.

For instance, when I look at the AoT documentation, they do this in their `tsconfig-aot.json`:
```javascript
{
  "compilerOptions": {
    "target": "es5",
    "module": "es2015",
    ...
  }
}
```

I'm guessing there is a good chance that's what you're doing, but the comments and commit message should probably be less ambiguous when you say "es2015 format". I know it's a mouthful, but "es5 with es2015 module format" will avoid confusion.

build/tasks/publish.js
```diff
@@ -106,28 +104,30 @@ gulp.task("npm:ui:readme", function () {
...
+ *   - the minified JS bundle (commonJS)
  *   - the Typescript declaration files for our components
  */
-gulp.task("npm:angular:deliverables", function () {
+gulp.task("npm:angular:bundles", function () {
     // We have to use 2 streams here because we want the angular 1 bundle to preserve its path.
     return es.merge(
         gulp.src([
             "dist/bundles/clarity-angular.min.js",
             "dist/bundles/angular1/clarity-angular1.min.js"
```
youdz: I think you can remove that. 😄

build/tasks/typescript.js
```diff
@@ -27,118 +27,109 @@ var runSequence = require('run-sequence');
...
 /**
  * Watches for changes in ts files (or files that will be inlined in typescript)
  * to retrigger typescript compilation.
  */
+var clarityHtmlFiles = ["src/clarity-angular/**/*.html"];
 gulp.task('typescript:clarity:watch', function () {
-    gulp.watch(claritySources, function () {
-        return runSequence('tslint:watch', 'typescript:clarity');
+    gulp.watch(claritySources.concat(clarityHtmlFiles), function () {
+        return runSequence('tslint:clarity:watch', 'typescript:clarity');
```
youdz: We could have a separate watcher on HTML that doesn't re-run tslint, if it becomes too slow at some point.

PS: `"tslint:clarity:watch"` is misleading, it made me pause and go look again because I thought our watcher was starting a new watcher every time. Maybe `"tslint:clarity:no-errors"` or something similar would be clearer.

gulpfile_old.js
```diff
@@ -0,0 +1,88 @@
+/*
```
youdz: Don't forget to remove this when you're done. 😄

src/app/alert/alert.demo.module.ts
```diff
@@ -0,0 +1,69 @@
+/*
```
youdz: This comment is _not_ something I want you to change right now, but I'm writing it down before I forget.

I was thinking about it this morning, and I think we should make it clearer what these are. I would write them in a separate `src/sandbox/` folder, that would be used by `src/app`. Of course, we'd remove a lot of the demos to stick to a single fully-featured example most of the time. So we'd do this as part of the demo cleanup rather than this PR.

src/app/app.module.ts
```diff
@@ -24,7 +24,8 @@ import {ClarityModule} from "../clarity/clarity.module";
         AppComponent,
         LandingComponent
     ],
-    bootstrap: [ AppComponent ]
+    bootstrap: [ AppComponent ],
+    schemas: [ CUSTOM_ELEMENTS_SCHEMA ]
```
youdz: This is not needed since the module already imports ClarityModule. I'd rather you didn't add the schema, and if you have errors we import ClarityModule from the various modules that need it. That way we can make sure it works without schema for most people. If you need help handling these errors, I'd be happy to.

src/clarity-angular/modal/modal.ts
```diff
@@ -0,0 +1,117 @@
...
+
+@Component({
+    selector: "clr-modal",
+    viewProviders: [ScrollingService],
+    templateUrl: "./modal.html",
+    styles: [`
+        :host { display: inline-block; }
+    `],
+    animations: [
+        trigger("fadeDown", [
```
youdz: Regarding this, I'm fine pushing with duplicate hard-coded animations for now, as long as we create an issue to fix it before we forget.

src/clarity-angular/wizard/wizard-page.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Emitter for Next button state changes
+    @Output("clrWizardPageNextDisabledChanged") nextDisabledChanged: EventEmitter<any> =
+        new EventEmitter(false);
+
+    constructor() {
+        super();
+    }
+
+    ngOnInit() {
+        let projectedTitleHTML: string = this.titleContainer.nativeElement.innerHTML.trim();
```
youdz: How does this work with AoT? We're accessing a property of the native element. I thought that was one of the things we couldn't do.

src/clarity-angular/wizard/wizard-step.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (!this.active && this.isCompleted) {
+            super.onClick();
+        }
+
+        return false;
+    }
+
+    ngOnInit(): void {
+        // The nav text will be the title of the main section.
+        this.title = this.elementRef.nativeElement.innerText.trim();
```
youdz: Same question, how does AoT accept that?

build/karma.conf.js
```diff
@@ -61,7 +61,7 @@ module.exports = function (config) {
             { pattern: 'node_modules/prismjs/components/prism-typescript.min.js', included: true, watched: false}
         ],
 
-        exclude: [],
+        exclude: [ 'node_modules/**/*spec.js' ],
```
jeeyun: Since we include other js files from `node_modules` in karma, making it `node_modules/**` overrides those include patterns declared earlier, and the tests won't be able to run. I had to introduce this because now we include `@angular/compiler-cli` in our project for AoT, and one of their spec files, will throw an error if we don't exclude (see: https://github.com/angular/angular/issues/13115). I think this is generic and safe enough. Let me know if you still disagree.

build/tasks/aot.js
```diff
@@ -0,0 +1,67 @@
...
+gulp.task("aot:copy", function() {
+    var claritySources = [
+        'src/clarity-angular/**/*.ts',
+        'src/clarity-angular/**/*.html',
+        '!src/clarity-angular/**/*.spec.ts',
+        '!src/clarity-angular/**/*.mock.ts'
+    ];
+
+    gulp.src(claritySources)
+        .pipe(inlineNg2Template({
```
jeeyun: If we opt to not inline the templates, we need to include our template files (which we haven't been) in our npm packages. Do we want to do that?

build/tasks/clean.js
```diff
@@ -12,8 +12,9 @@ var del = require("del");
  */
 gulp.task("clean", function () {
 	return del([
+		"aot-compiled/*",
 		"dist/*",
-		"tmp/"
+		"tmp/*"
```
jeeyun: There isn't. Actually let me clean up the clean task to delete all the generated folders themselves.

build/tasks/publish.js
jeeyun: 👍

build/tasks/typescript.js
jeeyun: I did separate out the tslint tasks to only run on the files that it's watching (earlier we had one monolithic tslint that ran for each typescript subgroups that the project was watching) so hopefully it's a bit better. The clarity components one is a special case because we inline the html. Of course this won't have to be the case if we opt to publish our html files (see my earlier comment on this).

I agree with the naming; I'll rename the tslint tasks.

src/app/app.module.ts
jeeyun: Not sure why I introduced this in the first place. Reverting.

build/tasks/publish.js
jeeyun: Absolutely. Yes, we are doing ES5 with ES2015 module format only as you guessed. Will update the comments.

src/clarity-angular/wizard/wizard-page.ts
jeeyun: Hmm I wasn't aware of limitations wrt accessing a native element's property in AoT. That being said, I tested the components in AoT, and they are working as expected. Let me know if you have any specific documentation or issue that you saw this in and I can take a look.

jeeyun:
@youdz: I think I addressed all of your comments. Let me know if I missed any. I will amend the commit message with the details when it's ready for merge.

Regarding a gulp task to test AoT compiled component in our demo app, I think it can be done but would it be ok if I file a separate issue for it? I don't know if I can fit it in for this week's release.

build/tasks/publish.js
```diff
@@ -112,16 +112,17 @@ gulp.task("npm:ui", ["npm:ui:bundles", "npm:ui:sources", "npm:ui:package", "npm:
...
- *   - the compiled js files in es2015 format
- *   - the minified JS bundle (commonJS)
+ *   - the compiled js files in es5 with es2015 module format
+ *   - the umd JS bundle
+ *   - the minified JS bundle (in register format)
  *   - the Typescript declaration files for our components
  */
 gulp.task("npm:angular:bundles", function () {
     // We have to use 2 streams here because we want the angular 1 bundle to preserve its path.
     return es.merge(
```
youdz: Nitpick: We don't need to merge two streams here anymore. We can just `gulp.src()` all at once.

Don't bother updating the review for this, it doesn't even matter if it stays that way, this is really a negligible nitpick.

Issue #186, opened 2016-12-05
Search component (new)


reddolan:
### Description
- a field or icon that can be interacted with to return a specific result or results
- search from header so that it is in a consistent place site-wide

### Use cases under consideration
- autocomplete
- placeholder text
- recent searches
- hot keys
- advanced search
- responsive design

### Future use cases under consideration
- operators, modifiers, wildcards
- saved searches
- formatted autosuggestions (different designs based on context)
- search from grid

#### Header Options
Search will be available to all Header design patterns

![headers](https://cloud.githubusercontent.com/assets/1858226/21439231/6486ae86-c852-11e6-8c95-28baae2b7e91.png)

Search Interaction
- Clicking on the Search icon or field will activate the Search interaction
- icon area will turn white (box around icon)
- icon will turn Active blue 
- white search box will expand to fill the Header space up to 1000px (see Responsive)
- expansion will start from the search icon/field origination point
- expansion will cover all icons and links
- expansion will stop to the left of the product name area
- search field will retain the placeholder text if there was, or display Search if it was just an icon

Starting point
![start](https://cloud.githubusercontent.com/assets/1858226/21439674/ff5cded8-c854-11e6-8850-950d2c13b4b4.png)

Beginning of interaction
![state 2](https://cloud.githubusercontent.com/assets/1858226/21439679/05828416-c855-11e6-8bab-21b378c01588.png)

Search field fully open
![state 3](https://cloud.githubusercontent.com/assets/1858226/21439687/0ec8b69e-c855-11e6-863f-14c7dfe412f9.png)

#### Recent Searches
- display recently searched  items when available
- shows immediately after expansion animation is finished
- display pages and/or search queries
- pages are displayed blue like a link
- icons associated with different object types (optional)
- Search experience can be closed by: clicking “cancel”, clicking outside of the search field or dropdown, and also by hitting the Esc key

![recent searches](https://cloud.githubusercontent.com/assets/1858226/21439705/1d264774-c855-11e6-8bc2-9568f68a0607.png)

#### Suggestions Interaction 
- Recent Searches stays visible while typing, up to 2 characters
- once the 3 character is typed, display the Searching dropdown
- if the suggestions come back quickly (~1 sec or less), immediately show the Suggestions dropdown and bypass showing the Searching dropdown
- show up to 5 of the most recent search queries

While typing
[...]

PR #259, opened 2016-12-16
[feature request] Add a Search component

varticvasile:
#186

initial commit to be reviewed

Signed-off-by: Vasile Vartic <vasile.vartic@bytex.ro>

BY BYTEX ROMANIA

### DIFF
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -32,6 +32,7 @@ export const APP_ROUTES: Routes  = [
     { path: "progress-bars", loadChildren: "clarity-demos/progress-bars/demo/progress-bars.demo.module" },
     { path: "radios", loadChildren: "clarity-demos/radios/demo/radios.demo.module" },
     { path: "selects", loadChildren: "clarity-demos/selects/demo/selects.demo.module" },
+    { path: "search", loadChildren: "clarity-demos/search/demo/search.demo.module" },
     { path: "spinners", loadChildren: "clarity-demos/spinners/demo/spinners.demo.module" },
     { path: "stack-view", loadChildren: "clarity-demos/stack-view/demo/stack-view.demo.module" },
     { path: "tables", loadChildren: "clarity-demos/tables/demo/tables.demo.module" },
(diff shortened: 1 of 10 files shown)


jaffoneh:
@varticvasile could you provide some screenshots of how this works. @reddolan is currently working on the design specs for search, maybe we can work together on getting that moving? Thanks for your contribution and looking forward to collaborating with you on this!

dragosrusu:
I think @varticvasile implemented a demo page. Maybe test there.
At this stage we should worry about functionality. Look and feel can be easily changed.

youdz:
Thank your for trying to implement a first draft of this component, @varticvasile.

First of all, I recommend you read our [coding guidelines](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md). Will cannot accept contributions that do not respect these.

In light of this, here are a few examples of things that would need to change. This list is not exhaustive, there might be other details in the code that might need to be fixed, but I just want to show examples from the coding guidelines rules.
- You need to add ARIA attributes to make sure this component is accessible. See https://www.w3.org/TR/wai-aria/roles#combobox and
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_listbox_role. Also implement the keyboard shortcuts required in these specs.
- This use case is too narrow, we want to offer full customizability from the beginning. In this case, at least support a custom filtering function, custom display of results, and non-string data (typically objects).
- Use projection for displayed results (after previous point).
- Support ngModel for even better customizability an integration. I see you started that based on commented out imports, you were right to try it, it's the way to go.
- Add the "clr" prefix to all inputs/outputs. See our other components.

Finally, I don't understand the point of the demos with no data. It's literally an `<input type="text">`.  All you are exposing is a `[placeholder]` input, a `[(value)]` to-way binding and a `(change)` event, all of which are already natively offered by inputs. So the only reason to use this component is for the results dropdown.

I like the idea of supporting regexes for power users, you wrote plenty of tests, and you even took the time to write a demo. Good initiative on these.

jaffoneh:
Based on the final comment by @youdz and no additional updates. I am closing this pull request, please let me know in a comment if you think it should be reopened and provide more information.

For big components like this with a big code contribution, we should discuss first a set of APIs that we agree on and a technical spec in order to match what we're looking to implement within Clarity.

Thanks again for making the contributions and please do take a look at open issues, open new components, and feel free to bring up any places where you can contribute. We're grateful for our community and contributors.

PR #271, opened 2016-12-20
[ICONS] enabled extending clarity icons

Shijir:
This is in progress! You can review it, but no approval is needed atm :)


src/clarity-icons/extend-clarity-icons.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    addAliases(aliases: IconAlias): void {
+
+        for (let shapeName in aliases) {
+
+            if (aliases.hasOwnProperty(shapeName)) {
+
+                if (SVG_ICON_TEMPLATES.hasOwnProperty(shapeName)) {
```
mathisscott: the issue here is that `SVG_ICON_TEMPLATES` AND `EXTENDED_ICON_TEMPLATES` are now mutable. yet they are cast as constants. We need to fix that and probably add some sort of `get()` in the API to abstract the retrieval of these icons.

src/clarity-icons/extend-clarity-icons.ts
```diff
@@ -0,0 +1,98 @@
...
+                    //set an alias to the icon if it exists in SVG_ICON_TEMPLATES.
+
+                    this.setIconAliases(SVG_ICON_TEMPLATES, shapeName, aliases[shapeName]);
+
+                } else if (SVG_IMG_ICON_TEMPLATES.hasOwnProperty(shapeName)) {
+                    //set an alias to the icon if it exists in SVG_IMG_ICON_TEMPLATES.
+
+                    this.setIconAliases(SVG_IMG_ICON_TEMPLATES, shapeName, aliases[shapeName]);
+
+                } else if (EXTENDED_ICON_TEMPLATES.hasOwnProperty(shapeName)) {
```
mathisscott: Why are we having `SVG_ICON_TEMPLATES`, `SVG_IMG_ICON_TEMPLATES` and `EXTENDED_ICON_TEMPLATES`?

This seems unnecessarily complex. we should have one source where all icons reside. 

These constant values are better used for when we bring in icon sets.

src/clarity-icons/extend-clarity-icons.ts
```diff
@@ -0,0 +1,98 @@
...
+    private setIconTemplate(shapeName: string, shapeTemplate: string): void {
+
+        if (this.validateName(shapeName) && this.validateTemplate(shapeTemplate)) {
+
+            EXTENDED_ICON_TEMPLATES[ shapeName ] = shapeTemplate;
+        }
+
+    }
+
+    addIcons(icons: IconTemplate): void {
```
Shijir: This is how it works currently. You can either add a single icon or add multiple icons in an object. 

For example:
```
ClarityIcons.addIcons({facebook: `<svg>...</svg>`});
ClarityIcons.addIcons({linkedin: `<svg>...</svg>`});
```
or 
```
ClarityIcons.addIcons({
facebook: `<svg>...</svg>`, 
linkedin: `<svg>...</svg>`,
...
...
...
});
```

src/typings.d.ts
```diff
@@ -18,4 +18,6 @@ declare var customElements: CustomElements;
 
 declare interface CustomElements {
     define: any;
-}
\ No newline at end of file
+}
+
+interface Window { ClarityIcons: any; }
```
Shijir: Yes. As long as, that's how they would use `ClarityIcons.add()` after loading `clarity-icons.min.js`.

src/clarity-icons/extend-clarity-icons.ts
Shijir: We can remove `SVG_IMG_ICON_TEMPLATES` and include the content in `SVG_ICON_TEMPLATES` as it was previously. So all default ClarityIcons will exist in `SVG_ICON_TEMPLATES`. But I think it's better to have user defined shapes separate from our default clarity icons shapes. That means will have only two source objects: `SVG_ICON_TEMPLATES` and `EXTENDED_ICON_TEMPLATES`.

src/clarity-icons/extend-clarity-icons.ts
```diff
@@ -0,0 +1,98 @@
...
+                get: function () {
+                    return templates[ shapeName ];
+                },
+                enumerable: true,
+                configurable: true
+            });
+        }
+    }
+
+    addAliases(aliases: IconAlias): void {
```
Shijir: Sure. Btw, this is how `ClarityIcons.addAliases` works:

```
ClarityIcons.addAliases({caret: ["arrow", "chevron", "my-caret"]});
```

So `caret` will have aliases such as `arrow`, `chevron`, and `my-caret`.

src/clarity-icons/extend-clarity-icons.ts
Shijir: Both `SVG_ICON_TEMPLATES` AND `EXTENDED_ICON_TEMPLATES` are not accessible to users. They can be mutable only through `ClarityIcons` API methods.

PR #313, opened 2017-01-04
Adding entry point for clarity-icons

katallaxie:
This adds an entry point for `clarity-icons`. It allows to better use `webpack` or other module loaders.


katallaxie:
@youdz yep, it just for convenience. For tree-shaking etc. we would have to add some other things. Would have to look into that.

mathisscott:
Before we can merge this change in @katallaxie, we'll need you to edit your commit message to follow the contribution guidelines...

```
[icons] <title goes here>

Signed-off-by: Sebastian Döll <your.email@example.com>
```

Thanks.

PR #327, opened 2017-01-06
[NG] Checkbox Checked & Indeterminate State Fix

adityarb88:
Fixed the Indeterminate and Checked state in the custom checkbox component.

Issues that were present before:
1. The checkbox did not emit the indeterminate state when the checkbox was toggled.
2. Checkbox and Indeterminate values were being set multiple times due to their inter dependency. For eg: If a checked value is set the indeterminate value is set and since the indeterminate value is set the checked value is again set.

Other changes:
Added additional tests.
Fixed the demo. 
Fixed existing tests.

**NOTE:** We need to port this to the website branch.

New Demo
image

Tested on Chrome. No other browser tests needed

**Unit Tests:**
Checkbox
    ✔ displays a native checkbox and label
    ✔ projects into the label
    ✔ toggles the checked state based on [clrChecked] input
    ✔ toggles the checked state based on user actions
    ✔ applies the given name attribute to the native input
    ✔ supports inline checkboxes
    ✔ supports indeterminate state
    ✔ sets indeterminate state to false when the checked state is toggled by user actions
    ngModel support
      ✔ toggles the checked state based on [ngModel] input
      ✔ emits changes to (NgModelChange) based on user actions

Should not be merged without @youdz's review.


src/clarity-angular/checkboxes/checkbox.ts
```diff
@@ -82,14 +86,17 @@ export class Checkbox implements ControlValueAccessor {
     public change = new EventEmitter<boolean>(false);
 
     public toggle() {
-        this._indeterminate = false;
```
youdz: If we set the private property here, how do we emit the `indeterminateChange` output?

adityarb88: @youdz: I removed that line. So we are setting only the checked field from here and the checked setter takes care of the indeterminate field too through the indeterminate setter using the pattern that we discussed last week.

youdz: Oh, yeah, I misread that.

I was thinking of a case where checked stays the same so we wouldn't set indeterminate to false, but that a `toggle()` method, so obviously it can't happen.

👍

PR #329, opened 2017-01-07
[docs] Update readme / issue files with links to multiple plunkers

hippee-lee:
I added multiple plunker links for latest and the last three Clarity Versions. This may not be the best way to surface multiple Clarity Plunker Templates in the project files so I am open to suggestions.


hippee-lee:
Painful to maintain -> yes.

In this case of a major version upgrade when the user has not upgraded to the latest yet they may find/report issues that have been fixed in latest and not back ported.

But perhaps only the version before?

PR #336, opened 2017-01-09
[ICONS] made clarity-icons importable in typescript

Shijir:
- Made Clarity-Icons importable and loadable in typescript files in the following way:
`import { ClarityIcons } from "clarity-icons"`

### DIFF
--- a/build/npm/clarity-icons.json
+++ b/build/npm/clarity-icons.json
@@ -3,7 +3,8 @@
     "version": "/* @echo VERSION */",
     "description": "Custom Element Icons for Clarity",
     "homepage": "https://vmware.github.io/clarity/",
-    "main": "clarity-icons.min.js",
+    "main": "index.js",
+    "types": "index.d.ts",
     "repository" : {
         "type" : "git",
         "url" : "ssh://git@git.eng.vmware.com/clarity.git"
--- a/build/tasks/publish.js
+++ b/build/tasks/publish.js
@@ -23,7 +23,9 @@ gulp.task("npm:icons:bundles", function () {
     return gulp.src([
         "dist/bundles/clarity-icons.min.js",
         "dist/bundles/clarity-icons.min.css",
-        "dist/clarity-icons/svg-icon-templates.js"
+        "dist/clarity-icons/**/*.ts",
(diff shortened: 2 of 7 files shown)


build/npm/clarity-icons.json
```diff
@@ -3,7 +3,8 @@
     "version": "/* @echo VERSION */",
     "description": "Custom Element Icons for Clarity",
     "homepage": "https://vmware.github.io/clarity/",
-    "main": "clarity-icons.min.js",
+    "main": "index.js",
+    "types": "index.d.ts",
```
jeeyun: How is the types setting used?

build/tasks/publish.js
```diff
@@ -23,11 +23,30 @@ gulp.task("npm:icons:bundles", function () {
...
         "dist/bundles/clarity-icons.min.css",
-        "dist/clarity-icons/svg-icon-templates.js"
+        "dist/clarity-icons/svg-icon-templates.js",
+        "dist/clarity-icons/svg-icon-templates.d.ts",
+        "dist/clarity-icons/clarity-icons-api.d.ts",
+        "dist/clarity-icons/clarity-icons-api.js",
+        "dist/clarity-icons/clarity-icon.d.ts",
+        "dist/clarity-icons/clarity-icon.js",
+        "dist/clarity-icons/index.d.ts",
+        "dist/clarity-icons/index.js"
```
jeeyun: why not use wildcard here? For example:
```
     "dist/clarity-icons/*.js",
     "dist/clarity-icons/*.d.ts",
```

build/npm/clarity-icons.json
Shijir: It's basically pointing to the proper declaration file, which is index.d.ts in this case. We don't have to specify this `types` field value as the declaration is already named `index.d.ts`. However, typescript documentation says it's it is advisable to do so.

build/tasks/publish.js
Shijir: Thanks for this recommendation. Please take a look at the update.

Shijir:
@jeeyun Yes. I have tested it in angular-cli's AoT compiled version.

Issue #156, opened 2016-11-28
[NG] Wizard refactor


youdz:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
The Wizard component's API should match our other components. At the very least, it should follow our own [coding guidelines](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md).

### Actual behavior
Currently, the Wizard component's API is drastically different from anything else we offer, which means users have to learn new patterns just for it. In addition, these different patterns are often not used in the main Clarity components for good reasons, typically because they hinder customizability. We end up with many feature requests because people have no workaround available, there only hope is that we get to these features as soon as possible.

Most importantly, the current Wizard contains hard-coded English text, which completely breaks any attempt at i18n.

A couple examples of issues to fix (this list is _not_ exhaustive):
- [Wizard template](https://github.com/vmware/clarity/blob/master/src/clarity/wizard/wizard.html#L36): a full footer of buttons with hard-coded text
- [Wizard page template](https://github.com/vmware/clarity/blob/master/src/clarity/wizard/wizard-page.html#L9): Interpolating the title instead of projecting it.
- [Wizard page title detection](https://github.com/vmware/clarity/blob/master/src/clarity/wizard/wizard-page.ts#L66): Using the native element **and** testing its content on ngInit, which cannot handle anything dynamic.

### Environment details

* **Angular version:** all

* **Clarity version:** 0.7.x

* **OS and version:** all

* **Browser:** all

yaooluu:
@youdz Hi Eudes, I noticed that when you have: 

`<clr-wizard-step>This is a title</clr-wizard-step>`

Both wizard TOC and page will have this same title. However, if I have: 

`<clr-wizard-step>{{ titleVar }}</clr-wizard-step>`

TOC shows the title correctly, but wizard page shows no title at all. 

I looked into the code of wizard and was trying to figure out the relationship between interpolation, projection, 1-way binding and 2 way binding. Can you confirm that my above code is the second issue you mentioned?

youdz:
@yaooluu: Yes, your example is one of the many that have to be handled with projection. Right now the code for the wizard is playing with native HTML elements on component initialization, which cannot handle anything dynamic.

[...]

PR #346, opened 2017-01-10
[ui] updating wizard styles to match specs

mathisscott:
• Part of the Wizard Refactor (#156) that can be done without breaking change
• Updated title text color and font size
• added styles to remove top margin from first child in the content area (having a paragraph or `h?` element would push content down)
• updated font weight and text color of active TOC element
• cleaned up text color lookups to use clr-getTextColor function

wizardstyleexample

Tested in:
✔︎ Chrome


src/clarity-angular/wizard/_wizard.clarity.scss
```diff
@@ -93,11 +95,30 @@ $clr-wizard-shadow-size: 4px;
         form {
             padding-top: 0;
         }
+
+        .content-heading {
+            margin: 0;
+            font-size: rem(16/$clr-rem-denominator);
```
hippee-lee: Does this variable come from variables.clarity.scss ?

mathisscott: yep. i believe it does.

hippee-lee: Thanks. Im just knowledge xfering here. :-)

PR #381, opened 2017-01-19
[ng] Make the checkbox component public

### LINKED ISSUE #319: [NG] Make the checkbox component public
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
The `<clr-checkbox>` component should be documented with proper demos, so that our users can test this "shortcut" and give us feedback on its usefulness. The point being to release this type of component for all inputs, so that our users won't have to constantly check the correct markup and classes needed by our static form controls.

### Actual behavior
Currently, the [component exists](https://github.com/vmware/clarity/blob/master/src/clarity-angular/checkboxes/checkbox.ts) and [is extensively tested](https://github.com/vmware/clarity/blob/master/src/clarity-angular/checkboxes/checkbox.spec.ts), but is just not documented. It is used internally by some of our other components.
It is also missing a few of the usual checkbox attributes, like `disabled` or `required` (although that last one probably works with `ngModel` on the `<clr-checkbox>` element itself. These should be added as part of the publication of the component if time allows, or be requested as a separate issue if not.

hippee-lee:
This is a first pass PR for the doc and ng updates to checkboxes. 
I'm looking for specific feedback on:

1. Is the native inderterminate example enough or should we show the angular code to toggle and set indeterminate state?
2. Any issues with how I implemented ngForm & ngModel in the second example?

Once this is accepted I'll use the demo code to update the website branch. Also, this demo code will depend on the new 'disabled' input for clr-checkbox. 

- Adds a demo for using a native checkbox with indeterminate state
- Adds a demo using clr-checkbox with ngForm and ngModel
- Exposes a 'disabled' property for the clr-checkbox component
- Adds a provider with fake server data for the checkbox demo
- Adds a demo with a list of clr-checkbox(es) using fake server data and disabled status
- closes #319
- Update checkbox reference image for css tests


src/app/checkboxes/checkboxes.demo.html
```diff
@@ -61,9 +61,51 @@
...
+            </div>
+        </div>
+    </section>
+</form>
+
+<form #form="ngForm" novalidate>
+    <section class="form-block">
+        <div class="form-group">
+            <label>With ngForm & ngModel</label>
+            <div class="checkbox-inline">
```
adityarb88: I am a bit confused by this. I thought that you dont have to wrap the `.checkbox-inline` class on the `clr-checkbox`. `clr-checkbox` takes an input `clrInline` and adds the class on the host itself (`clr-checkbox`) based on the input, no?

src/app/checkboxes/checkboxes.demo.html
```diff
@@ -61,9 +61,51 @@
...
+                <clr-checkbox
+                        name="termsCheckbox"
+                        id="checkrads_11"
+                        [clrChecked]="termsAgreement"
+                        [ngModel]="termsAgreement">
+                    I agree to the terms and conditions.
+                </clr-checkbox>
+            </div>
+            <div class="checkbox-inline">
```
adityarb88: This `.checkbox-inline` wrapper has not `input` of `type=checkbox` inside of it?

src/app/checkboxes/checkboxes.demo.html
```diff
@@ -61,9 +61,51 @@
...
+            </div>
+        </div>
     </section>
 </form>
 
+<form>
+    <section class="form-block">
+        <div class="form-group">
+            <label>With a list of objects</label>
+            <div class="checkbox" *ngFor="let item of list">
```
adityarb88: Same here. I thought the `.checkbox` class is added to the host automatically. I do see it on the demo app:
image

I am not sure why we have to wrap it again in `.checkbox`

src/clarity-angular/checkboxes/checkbox.ts
```diff
@@ -46,6 +47,10 @@ export class Checkbox implements ControlValueAccessor {
     @Input("name")
     public name: string = null;
 
+    // If the host is disabled we apply it to the checkbox
+    @Input("clrDisabled")
+    public disabled: boolean = false;
```
adityarb88: I think along with setting the `:disabled` property, we also set a class `.disabled` on the checkbox because that affects the label styles for the text too. I don't remember if we need to do both but we should double check.

src/app/checkboxes/checkboxes.demo.html
hippee-lee: Doh! - still learning angular and how to read typescript.

src/app/checkboxes/checkboxes.demo.html
```diff
@@ -49,21 +49,55 @@
...
     </section>
 </form>
 
+<form #form="ngForm" novalidate>
+    <section class="form-block">
+        <div class="form-group">
+            <label>With ngForm & ngModel</label>
+            <clr-checkbox
+                    name="termsCheckbox"
+                    class="checkbox-inline"
```
youdz: You don't need to manually add the class since you already have `[clrInline]="true"`.

src/app/checkboxes/checkboxes.demo.html
```diff
@@ -49,21 +49,55 @@
...
+</form>
+
+<form>
+    <section class="form-block">
+        <div class="form-group">
+            <label>With a list of objects</label>
+            <clr-checkbox *ngFor="let item of list"
+                          [(clrChecked)]="item.running"
+                          [clrDisabled]="item.disabled">
+                {{ item.name }} at {{ item.ip }} {{ item.disabled ? ' Is disabled.' : 'Is ok.' }}
```
youdz: To make the example more relevant,  you should change this to:
```
{{ item.disabled ? ' is disabled.' : (item.running ? 'is running.': 'is off.') }}
```
Or something similar that shows how checking a box actually modifies the model through `[(clrChecked)]`.

Issue #239, opened 2016-12-13
Design and implement a Tree view


lil-kim:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
Design and implement a tree view with options for icons or checkboxes in the nodes. See attached files for and visual spec and behavior info.

## Types of trees and Behavior
**Terminology**
Highlight = Click the text node to navigate to that content.
Select = Choose items to apply an action to, ie: via checkboxes.

**View Tree**
1 click target. 
Clicking the caret expands or collapses the list. 
This tree is view only, nodes can’t be highlighted or selected.

**Basic Tree with or without icons**
2 click targets: 1) caret, 2) text (plus icons, when present). 
Clicking the caret expands or collapses the list. 
User can highlight just the text to navigate to that content. 
Multi-selection is not supported.

**Checkbox Tree**
3 click targets: 1) caret, 2) checkbox, 3) text
Multi-selection supported via checkboxes
Text nodes can be highlighted to navigate to content

## Loading Behavior
Cursor:
Hover/click over caret: Cursor doesn’t change (standard arrow). Color of caret changes to #000000 on hover/click.

Hover over text: Cursor changes to hand, row BG color: #EEEEEE (hover); #D9E4EA (select)

Loading spinner position:
Upon click, caret is replaced by the spinner.

Animation
Caret should animate smoothly in a 90° arc.
When expanded, the child nodes should cascade in from top to bottom

![tree types1](https://cloud.githubusercontent.com/assets/23196444/21165286/49646374-c153-11e6-9e45-81b079ba7dfc.png)
![tree node specs](https://cloud.githubusercontent.com/assets/23196444/21165281/410559ae-c153-11e6-8a71-4738d33d7682.png)

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@youdz and I discussed the following use cases/examples that we plan to cover during implementation:

**1. Basic Example**
```
<clr-tree-node>
    A1
    <clr-tree-node>
        B1
        <clr-tree-node>C1</clr-tree-node>
        <clr-tree-node>C2</clr-tree-node>
[...]

PR #382, opened 2017-01-19
[NG] Tree View

adityarb88:
Fixes: #239 

@youdz and I implemented the TreeView component with the following features and the API mentioned in #239 (with an exception to the last example in the API - selection is of the type `TreeSelection[]` instead of `TreeSelection`):

**Note that these are not the final examples**

1. Basic Tree Nodes
image

2. Tree Node Expanded Input/Output
image

image

3. Building Trees Dynamically
image

4. Lazy Loading Tree Nodes
image

image

5. Selectable Tree Nodes
image

6. Toggling Individual Tree Nodes
image

7. Routing using Tree Nodes
image

8. Prepopulating Trees
image

Added an experimental feature compact trees in addition to the API mentioned in #239. Have to test this on smaller screens before releasing

**Please note** that this PR is not final. I am creating this PR so that reviewers can get a head start to check out the code and suggest any major changes in the logic. I still need to:

1. Still need to make styling changes to the component. The classnames are not final and might change after I discuss them with @mathisscott 
2. The tree node examples have to be updated and cleaned up.
3. Add more tests
4. Browser test styling
5. CSS Regression Tests

**Tested on:** Chrome only. Will browser test this week.

Unit Tests:
Tree View Directives
    Tree Node
      Tree Node Basics
        ✔ projects content
        ✔ projects child tree nodes under .clr-tree-node-children
        ✔ projects content under .clr-tree-node-content
        ✔ shows caret under .clr-tree-node-caret only for tree nodes having child tree nodes
        ✔ has hasChildren set to true for the parent tree node
        ✔ has hasChildren set to false for the child tree node
        ✔ has expandable set to false all tree nodes by default
        ✔ has selectable set to false all tree nodes by default
        ✔ has expanded set to false on tree nodes by default
        ✔ toggles 'expanded' on the tree node when its expanded & collapsed
        ✔ has caretDirection set to 'right' on tree nodes by default
        ✔ toggles caretDirection on the tree node when its expanded & collapsed
      Tree Node Default Expand Option
        ✔ sets the tree node 'expanded' property to the input provided by the user
        ✔ emits the 'expanded' event back to the user when the caret button is toggled
      Tree Node Expandable Option for Lazy Loading
        ✔ does not show the caret when expandable is set to false
        ✔ shows the caret when expandable is set to true
        ✔ shows the spinner when expandable & loading is set to true
        ✔ hides the caret when spinner is visible and vice-versa
      Building Trees Dynamically
        ✔ projects content
      Basic Tree Node Selection
        ✔ enables the selection based on the clrTreeSelected input
        ✔ accepts user input for tree node selection
        ✔ emits a change event after tree node selection
      Parent Child Tree Node Selection
        ✔ initializes all tree nodes to selected false
        ✔ initializes using the user input for model
        ✔ checks all children when the parent is checked
        ✔ checks all grand children when the parent is checked
        ✔ sets the parent as indeterminate when the one child is selected while others are not
        ✔ selects the parent when all the children are selected
        ✔ emits event when node is selected
    Tree View
      Pre-populating Trees
        ✔ pre populates tree based on the tree selection input
      Tree Selection
        ✔ does not emit anything when first initialized
        ✔ emits an array with all the root nodes when selection is updated
        ✔ emits an array with right selection


src/app/tree-view/basic-tree-view/tree-view-basic.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    nodes: string[] = [];
+
+    onSelectionChange(value: any[]) {
+        this.selection = value;
+        this.nodes = this.populateSelectedNodes(value);
+    }
+
+    //BTW, this could be our flatten logic
+    populateSelectedNodes(selectionArr: any[]): string[] {
```
youdz: Yes, I think this might be one of the first requests we receive.

src/app/tree-view/tree-node-selection/tree-node-selection.html
```diff
@@ -0,0 +1,29 @@
...
+            Documents
+            <clr-tree-node [clrTreeModel]="'doc1'">Document1.txt</clr-tree-node>
+            <clr-tree-node [clrTreeModel]="'doc2'">Document2.txt</clr-tree-node>
+            <clr-tree-node [clrTreeModel]="'doc3'">Document3.txt</clr-tree-node>
+        </clr-tree-node>
+
+        <clr-tree-node
+                [clrTreeModel]="'Pictures'"
+                [clrTreeNodeSelected]="picturesNodeSelection"
+                (clrTreeNodeSelectedChange)="onPicturesNodeSelectionChange($event)">
```
youdz: Same here, please use simple two-way binding.

src/app/tree-view/tree-node-selection/tree-node-selection.html
```diff
@@ -0,0 +1,29 @@
+<!--
+  ~ Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+<clr-tree [(clrTreeSelected)]="selection">
```
youdz: Isn't this the exact same example as the tree-view-basic one?

src/clarity-angular/tree-view/abstract-tree-selection.ts
```diff
@@ -0,0 +1,125 @@
...
+import {TreeSelectionService} from "./providers/treeSelection.service";
+
+export abstract class AbstractTreeSelection {
+    constructor(public parent: AbstractTreeSelection,
+                public treeSelectionService: TreeSelectionService) {
+    }
+
+    public model: any;
+
+    abstract get children(): AbstractTreeSelection[];
```
youdz: I think that by registering TreeNodes as AbstractTreeSelection providers, we can use @ContentChildren directly here and add more logic to this class so we can reuse it more easily in other components (in our case, at least Stack View and Datagrid).

Please don't fix this right now, we'll worry about it when we reuse the class.

src/clarity-angular/tree-view/abstract-tree-selection.ts
```diff
@@ -0,0 +1,125 @@
...
+                    break;
+                }
+            }
+        }
+        // FIXME: Only update the parent if this node really changed
+        // But when doing that, we need to make sure the root node is still warned that the overall selection
+        // has changed.
+        if (this.parent) {
+            this.parent.childChanged();
+        }
```
youdz: You can handle that FIXME right now: by having the TreeView observe the singleton service instead of waiting for its children to bubble up a change, you can stop the propagation from child to parent as soon as one of the parents is unchanged.

src/clarity-angular/tree-view/all.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import TreeNodeSpecs from "./tree-node.spec";
+import TreeViewSpecs from "./tree-view.spec";
+
+describe("Tree View Directives", () => {
+    TreeNodeSpecs();
+    TreeViewSpecs();
+});
```
youdz: We seriously need a spec on AbstractTreeSelection. That's where 90% of the logic is.

It would also be good to have specs on the TreeSelectionService, since it apparently has a validation method.

src/clarity-angular/tree-view/test.txt
```diff
@@ -0,0 +1,120 @@
+/*
```
youdz: Forgot to delete this file?

src/clarity-angular/tree-view/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import {TreeView} from "./tree-view";
+import {TreeNode} from "./tree-node";
+
+export * from "./tree-view";
+export * from "./tree-node";
+
+export const TREE_VIEW_DIRECTIVES: Type<any>[] = [
+    TreeNode,
+    TreeView
+];
```
youdz: You're not exporting the TreeSelection interface! The whole point of it is for consumers to import it and type our outputs.

src/app/tree-view/tree-node-selection/tree-node-selection.html
adityarb88: Nope. In this example we toggle a node using the `clrTreeNodeSelected` input. This example is going to be updated though.

src/app/tree-view/tree-node-selection/tree-node-selection.html
adityarb88: will do. ty

src/clarity-angular/tree-view/test.txt
adityarb88: Yes :D

src/clarity-angular/tree-view/index.ts
adityarb88: I totally forgot. Thanks for catching this!

hippee-lee:
Not being close to this, everything I read makes sense.

hippee-lee:
What I read looks good. I'll try it out if I have time.

src/clarity-angular/main.scss
```diff
@@ -46,11 +46,12 @@
 @import "./spinners/spinner.clarity"; // depends on variables.clarity, mixins.clarity, color.clarity, helpers.clarity, icons.clarity
 @import "./stack-view/stack-view.clarity"; // depends on variables.clarity, mixins.clarity, color.clarity, helpers.clarity, forms.clarity
 @import "./tables/tables.clarity"; // depends on variables.clarity, mixins.clarity, helpers.clarity, typography.clarity
+@import "./tree-view/tree-view.clarity";
```
adityarb88: thanks for catching this. will add the dependencies

src/clarity-angular/tree-view/abstract-tree-selection.ts
```diff
@@ -70,14 +78,15 @@ export abstract class AbstractTreeSelection {
...
         }
-        // FIXME: Only update the parent if this node really changed
-        // But when doing that, we need to make sure the root node is still warned that the overall selection
-        // has changed.
-        if (this.parent) {
+
+        if (this.parent
+                && (this._selected !== previousSelectedValue
+                    || this._indeterminate !== previousIndeterminateValue)) {
+            //console.log("Calling ", this.parent.model, " child changed method from ", this.model);
```
youdz: You forgot to clean this up. 😛

PR #392, opened 2017-01-20
[ng,web] Make the checkbox component public

### LINKED ISSUE #319: [NG] Make the checkbox component public

hippee-lee:
**NOTE: Do not merge this to the website branch until after v0.8.4 is release to npm.**

- Adds code and docs for the website for clr-checkbox, native checkbox usage by porting over the demo app from clarity-angular
- closes #319

## Screen shot of the new demo:
screen shot 2017-01-20 at 12 18 28 pm

### DIFF
--- a/src/demos/checkboxes/checkboxes.demo.html
+++ b/src/demos/checkboxes/checkboxes.demo.html
@@ -48,59 +48,148 @@
         <div class="form-group">
             <label>Indeterminate Checkbox</label>
             <div class="checkbox-inline">
-                <clr-checkbox type="checkbox" id="checkrads_9" [clrIndeterminate]="indeterminateState"></clr-checkbox>
-                <label for="checkrads_9">
-                    <button class="btn btn-primary" (click)="onToggleIndeterminateState($event)">
+                <clr-checkbox type="checkbox" id="checkbox_9" [(clrIndeterminate)]="indeterminateState">
+                    Clarity indeterminate checkbox
+                </clr-checkbox>
+                <div>
+                    <button class="btn btn-sm" (click)="onToggleIndeterminateState($event)" type="button">
                         Toggle Indeterminate State
                     </button>
-                </label>
+                </div>
+                <div class="checkbox-inline">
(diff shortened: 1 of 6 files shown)


src/demos/checkboxes/checkboxes.demo.html
```diff
@@ -48,59 +48,148 @@
...
+    <section class="form-block">
+        <div class="form-group">
+            <label>With a list of objects</label>
+            <clr-checkbox *ngFor="let item of list"
+                          [(clrChecked)]="item.running"
+                          [clrDisabled]="item.disabled">
+                {{ item.disabled ? ' is disabled.' : (item.running ? ' is running.' : 'is off.') }}
+            </clr-checkbox>
+        </div>
+    </section>
```
mathisscott: this is a gigantic code-block. do we need all of this in the pre > code section below?

hippee-lee: Not sure if we need the code in the pre section. But, I was operating on the assumption that a user would want to copy paste the code for the example they saw that interested them. If thats not the case let me know and I can truncate it.

PR #405, opened 2017-01-24
[ui] Remove active style for focused anchor tag

### LINKED ISSUE #374: Two class.active on nav-link when browser's back button clicked.
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Only one class.active should be displayed.

### Actual behavior

I can see two class.active are displayed until clicking other area on the page than nav-link.
image

### Reproduction of behavior
https://plnkr.co/edit/pEcTShQmRykVaIhbmmcn?p=preview

Click About --> Home --> About again, then click browser's back button. You will see two class.active. Then click any area on the Home page, you will see class.active on About disappeared.

### Environment details

* **Angular version:** 2.4.3

* **Clarity version:** 0.8.2

* **OS and version:** Windows 7

* **Browser:** tested on Chrome 55

jeeyun:
Currently, we apply the "active" style of an anchor tag on hover, active, and focus. This introduces a bug in which when the user navigates to another page through back/forward button on browser, the previous active anchor tag in nav keeps the "active" style until the focus is taken out by user clicking on another area. Let me know if this fix of removing the style for focus sounds reasonable, or if there's a better alternative fix.

Tested the newly generated css on seed.

Fixes #374.

### DIFF
--- a/src/clarity-angular/nav/_nav.clarity.scss
+++ b/src/clarity-angular/nav/_nav.clarity.scss
@@ -49,8 +49,7 @@ $clr-active-nav-link-shadow: 0 -3px 0 $action-blues-dark-midtone inset;
             }
 
             &:hover,
-            &.active,
-            &:focus {
+            &.active {
                 box-shadow: $clr-active-nav-link-shadow;
                 transition: box-shadow 0.2s ease-in;
                 outline: 0; //box-shadow above will ensure that the links are accessible.


src/clarity-angular/nav/_nav.clarity.scss
```diff
@@ -49,8 +49,7 @@ $clr-active-nav-link-shadow: 0 -3px 0 $action-blues-dark-midtone inset;
             }
 
             &:hover,
-            &.active,
-            &:focus {
```
adityarb88: @jeeyun: are we removing the focus states altogether? Will this mean that the tabs revert back to the browser default focus states and wont the browser default states still show when the user hits the back button? 

What if we programmatically remove the focus when the user switches a tab? (btw eudes didn't want us to use blur to remove focus states because it resets the focus to the top. so we should check with him once on this if we decide to go ahead with this approach) cc: @youdz

jeeyun: @adityarb88: yes, with the change above the browser default focus state will show but only if the user tabbed to it and hit the back button (after fix screenshot below). If the user clicks on it and hits the back button then it won't highlight it at all.

Let me know if my explanation is confusing.

Before fix:
screen shot 2017-01-25 at 12 01 24 pm

After fix:
screen shot 2017-01-25 at 12 00 40 pm

adityarb88: cool. since the focus state is changing we need to make sure that @yenmade knows about this and approves it.

adityarb88: also we can add a `outline-offset: -3px` or some other -ve offset to make sure that the focus state is within the tab component

PR #417, opened 2017-01-27
Adding Composer support for Clarity.

scrothers:
I have added a standard Composer package metadata file to the project. Composer is a package management system for PHP projects, similar to NPM or or Bower and such.

I have rooted the namespace of the project to "vmw" as currently, in Packagist (the repository where Composer packages are stored) the "vmware" namespace is currently in use located [here](https://packagist.org/packages/vmware/) by this [organization](https://github.com/purple-dbu).

In the future if VMware discusses the transfer of this namespace, it can be accomplished easily by editing the composer.json file.

Some reference information about Composer and Packagist is located here:
* https://getcomposer.org/doc/00-intro.md#introduction
* https://packagist.org/about

Once this merge request is accepted, I will repoint the namespace I already registered to this project. If a VMware team member would like to contact me, I'll gladly relinquish the ownership of the "vmw" namespace I have already registered on Packagist. Maintaining security over your namespace is important, as it prevents malicious actors from claiming to be VMware falsely to users and forcing them to download code from unauthorized sources or locations.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ _site/
 js/
 .vscode
 .stylelintrc
+/vendor/
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,14 @@
+{
+    "name": "vmw/clarity",
+    "description": "Clarity Design System: UX guidelines, HTML/CSS framework, and Angular 2 components working together to craft exceptional experiences.",
+    "type": "library",
+    "license": "MIT",
+    "keywords": ["angular", "angular2", "javascript", "html", "css", "design"],
+    "homepage": "https://vmware.github.io/clarity/",
+    "support": {
+        "issues": "https://github.com/vmware/clarity/issues",
+        "wiki": "https://github.com/vmware/clarity/wiki",
+        "source": "https://github.com/vmware/clarity",
+        "docs": "https://vmware.github.io/clarity/documentation"
+    }
+}


youdz:
We talked with the team about this, and tried to understand it. It's all very intriguing for us.
Here are some of the questions that popped:
- Why would a PHP project depend on an Angular library? I can understand trying to pull the pure CSS or the icons, but the whole thing?
- How does Composer work without us deploying something? Because if we don't do that, all it can do is pull our source code, which is completely unusable in a PHP project. You"d still have to compile Sass, Typescript, bundle, ... Not to mention rebuilding yourself all of Clarity from the source is something we never want to support.
- If we have to build and deploy for it to be available on Packagist, I didn't see any information on it after quickly reading their documentation. As a small team, anything that makes the release process longer is something that we really need to evaluate and decide if it's worth it.
- Why do we need to ignore the `/vendor/` folder in our own project?
- Our deliverables are currently split into 3 packages: `clarity-icons`, `clarity-ui` and `clarity-angular`. Any form of delivery (here for Composer) would need to respect that, which is not the case here.

Could you maybe shed some light on these, so we can make an informed decision? Right now, it makes very little sense for us. Thank you!

jaffoneh:
Closing this one based on @youdz's last comment. Please provide more info in a comment before reopening this to make sure it makes sense for us to reopen and discuss. Thanks!

PR #434, opened 2017-01-31
[core] Update published scss sources to use proper bootstrap import string

hippee-lee:
This fixes an issue that occurred after upgrading the website to use the latest angular-cli (**v1.0.0-beta.26**). The scss imports for bootstrap started failing when building the website for production / deployment. 

This proposed change will leave clarity source files intact and update them when they are published on npm for consumption by the user in an angular-cli project. 

### Needed for us to build and publish clarity-ui:
`@import "node_modules/bootstrap/scss/variables";`

### Needed for an angular-cli  (**v1.0.0-beta.26**) project (e.g. website):
`@import "~bootstrap/scss/variables";`

### DIFF
--- a/build/tasks/publish.js
+++ b/build/tasks/publish.js
@@ -9,6 +9,7 @@ var rename = require('gulp-rename');
 var preprocess = require('gulp-preprocess');
 var util = require('gulp-util');
 var es = require('event-stream');
+var replace = require('gulp-replace');
 
 // All packages share the same version number.
 var VERSION = util.env.version;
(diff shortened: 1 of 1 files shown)


build/tasks/publish.js
```diff
@@ -94,6 +95,7 @@ gulp.task("npm:ui:sources", function () {
         "src/clarity-angular/main.scss",
         "src/clarity-angular/**/*.clarity.scss"
     ])
+        .pipe(replace('node_modules/bootstrap', '~bootstrap'))
```
Shijir: I think this could be only a temporary solution. 

Ideally, we would want to publish our source code as it is without making changes to it. If users try to build `clarity-ui` by themselves from our published `src`, they would run into the similar problems we've faced.  The problem could be solved automatically once we get off Bootstrap as a dependency.

PR #443, opened 2017-02-02
[AFTER-PUBLISH][WEB] Update icon documentation

Shijir:
Also, updated the favicon because the old one was a bit pixelated on retina screens.
screenshot 2017-02-01 20 59 48


src/index.html
```diff
@@ -15,6 +15,8 @@
 
     <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
     <link rel="icon" href="favicon.ico" type="image/x-icon">
```
mathisscott: Why do we want to keep the old favicon around?

PR #455, opened 2017-02-04
[ICONS] publish icon svg files

Shijir:
- Fixes #386

Signed-off-by: Shijir Tsogoo <stsogoo@vmware.com>

PR #458, opened 2017-02-06
[ui/ng] Datagrid single row selection / action

### LINKED ISSUE #17: Need a single row selection/action pattern on datagrid
```
[ ] bug
[X ] feature request
```

### Expected behavior
We need a pattern for selecting a single row in datagrid and taking action on it especially when patch operations are not available. Think of a datagrid containing user accounts and a single action (Edit). How do you select one user and edit their account.

### Actual behavior
Currently, we don't have a pattern for this.

### Reproduction of behavior
N/A

### Environment details
N/A

* **Angular version:** 2.0.X
N/A

* **Clarity version:** 
Future

* **Browser:**
all

jeeyun:
This is an implementation of the datagrid single row selection / action. 

on click action overflow menu:
screen shot 2017-02-07 at 6 07 23 pm

context sensitive action bar (one selection vs multiple):
screen shot 2017-02-07 at 6 06 47 pm
screen shot 2017-02-07 at 6 06 53 pm

I have also updated the css regression tests to include data grid but re-commented it out because our datagrid data selection is dynamic, which means the width of the columns differs based on the data text. I still want to check in the commented out code, because it has good examples of how to scroll the component into view, and test cases that should be relevant if we decide to always render datagrid in a predictable/repeatable way for css regression.

Resolves #17.


soates:
Fantastic change, it something we have implements but no where near as clean as this..

chunghha:
How about providing a new option, called "clrDgSelectionType"? So We can specify the option to begin with into template of Datagrid something like [clrDgSelectionType]="single" (default to multiple if not specified) then Datagrid won't display the checkbox on Header and clears checkbox on previously selected row when new selection made.

youdz:
@chunghha: That's a pretty good idea. There's been quite a bit of discussion on #17 about single-row selection, mostly involving no checkboxes and the full row being clickable. That has issues that were explained in detail in the comments, but your idea sounds like a neat compromise.

We'll look into it, it would probably just be a different two-way binding than the current one, rather than a separate "option" input.
Current is: 
```<clr-datagrid [(clrDgSelected)]="listOfSelectedItems">```
And we could provide:
```<clr-datagrid [(clrDgSingleSelected)]="selectedItem">```
The first one binds to a list, the second one to a single item. We'd handle the "radio" behavior instead of checkboxes for you.

We will probably wait for some feedback on the pattern @jeeyun just implemented, but I could definitely see your request added to our backlog soon.

**Comment by chunghha** [comment]
@youdz yes what I'd like to have is "radio" behavior for single row selection in addition to @jeeyun's implementation here. And, I concur with your idea which is better than a separate option input I suggested.

File: src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+export default function(): void {
+    describe("DatagridActionBar component", function() {
+        let context: TestContext<DatagridActionOverflow, SimpleTest>;
+
+        beforeEach(function() {
+            context = this.create(DatagridActionOverflow, SimpleTest);
+        });
+
+        it("projects content", function() {
+            expect(context.clarityElement.textContent.trim()).toMatch("Hello world");
```
youdz: Again, this should not pass. The action overflow is not open, so the test element should only contain the ellipsis icon.

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+            expect(context.clarityDirective.open).toBe(false);
+            toggle.click();
+            context.detectChanges();
+            expect(context.clarityDirective.open).toBe(true);
+            toggle.click();
+            context.detectChanges();
+            expect(context.clarityDirective.open).toBe(false);
+        });
+    });
+}
```
youdz: Please add the following tests (list not exhaustive):
- Register/unregister with the service
- Two-way binding on `[(clrDgActionOverflowOpen)]`
- Close on click outside

src/clarity-angular/datagrid/datagrid-row.ts
```diff
@@ -3,15 +3,19 @@
...
 import {Selection} from "./providers/selection";
+import {DatagridActionOverflow} from "./datagrid-action-overflow";
 
 @Component({
     selector: "clr-dg-row",
     template: `
         <clr-dg-cell *ngIf="selection.selectable" class="datagrid-select">
             <clr-checkbox [ngModel]="selected" (ngModelChange)="toggle($event)"></clr-checkbox>
         </clr-dg-cell>
+        <clr-dg-cell *ngIf="actionableRows.length > 0" class="datagrid-single-select">
```
youdz: Nitpick: do you mind renaming this `datagrid-single-select` class to `datagrid-row-actions`?
We might have actual single selection patterns coming in later, and I want to avoid potential conflicts.

src/clarity-angular/datagrid/datagrid-row.ts
```diff
@@ -20,6 +24,11 @@ import {Selection} from "./providers/selection";
     }
 })
 export class DatagridRow {
+
+    @ContentChildren(DatagridActionOverflow) private _actionableRows: QueryList<DatagridActionOverflow>;
+    public get actionableRows(): QueryList<DatagridActionOverflow> {
+        return this._actionableRows;
+    }
```
youdz: The logic seems off to me here, I'm having a hard time making sense of it.

You should have 2 `*ngIf`s here:
- one on the `clr-dg-cell`, which should be true if `RowActionService.actionableCount` is greater than 0. That means somewhere, one of the rows has an extra cell for actions, so they should all have the extra cell to preserve the table uniformity.
- one on the action overflow `<ng-content select="clr-dg-action-overflow">` itself, which should be true if this row _does_ have actions projected. If it doesn't, there should still be an empty cell but no ellipsis icon.

src/clarity-angular/datagrid/datagrid.ts
```diff
@@ -21,16 +21,17 @@ import {Items} from "./providers/items";
...
     selector: "clr-datagrid",
     templateUrl: "./datagrid.html",
-    providers: [Selection, Sort, Filters, Page, Items]
+    providers: [Selection, Sort, Filters, Page, RowActionService, Items]
 })
 export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
 
     constructor(private selection: Selection, private sort: Sort, private filters: Filters,
-                private page: Page, public items: Items) {}
+                private page: Page, private rowActionService: RowActionService, public items: Items) {}
```
youdz: Used in the template, needs to be public.

src/app/datagrid/selection/selection.html
```diff
@@ -52,11 +72,9 @@
...
 
     <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
-        <clr-dg-action-overflow>
-            <ul class="list-unstyled">
-                <li><button type="submit" class="btn btn-sm btn-link">Edit</button></li>
-                <li><button type="submit" class="btn btn-sm btn-link">Delete</button></li>
-            </ul>
+        <clr-dg-action-overflow [clrDgActionItem]="user">
+            <a class="action-item" href="#" (click)="onEdit()" clrDgActionOverflowToggle>Edit</a>
+            <a class="action-item" href="#" (click)="onDelete()" clrDgActionOverflowToggle>Delete</a>
```
youdz: If I understand correctly, `clrDgActionOverflowToggle` is here to close the action overflow menu when one of the actions is clicked. While I do approve this behavior, the implementation right now bothers me, it requires a bit too much from the consumer for my taste.

How about just closing on any click on the document except for clicks on the action menu _itself_? Right now it closes only on clicks outside, but by not going up the tree and only checking if the target is the menu itself, any clicks on links inside the menu would still close it.

src/app/datagrid/selection/selection.html
```diff
@@ -35,15 +35,35 @@
...
-        <button type="submit" class="btn btn-link"><clr-icon shape="close" size="16"></clr-icon> Delete</button>
-        <button type="submit" class="btn btn-link" *ngIf="selected.length == 1">
-            <clr-icon shape="pencil" size="16"></clr-icon> Edit
-        </button>
+        <div class="btn-group" role="group" aria-label="Available Actions">
+            <button type="button" class="btn btn-sm btn-secondary" (click)="onAdd()"><clr-icon shape="plus"></clr-icon> Add
+                to
+                group</button>
+            <button type="button" class="btn btn-sm btn-secondary"><clr-icon shape="close"></clr-icon> Delete</button>
+            <button type="button" class="btn btn-sm btn-secondary" *ngIf="selected.length == 1"><clr-icon shape="pencil"></clr-icon> Edit</button>
```
youdz: You're missing `onEdit()` and `onDelete()` on these buttons.

src/app/datagrid/selection/selection.ts
```diff
@@ -31,10 +31,27 @@ export class DatagridSelectionDemo {
...
+    toEdit: User;
 
     constructor(private inventory: Inventory) {
         inventory.size = 10;
         inventory.reset();
         this.users = inventory.all;
     }
+
+    onDelete() {
+        this.toDelete = this.selected[0];
```
youdz: In your example, the Delete button in the action bar allows batch deletes. This doesn't account for it.

src/clarity-angular/datagrid/datagrid-action-overflow.ts
```diff
@@ -61,6 +70,11 @@ export class DatagridActionOverflow {
             //Start checking if current and actionMenuHost are equal. If not traverse to the parentNode and check again.
             while (current) {
                 if (current === actionMenuHost) {
+                    // if this datagrid has checkboxes, uncheck all but this current one
+                    if (this.selection.selectable) {
+                        this.selection.current = [];
+                        this.selection.setSelected(this.item, true);
+                    }
```
youdz: Ok, we need to talk about this. This is really not the behavior I'm expecting from the UX specs.

From what I see, actions in the overflow menu should be completely decoupled from selection, they just allow instant actions on the model of a row. Which also means you shouldn't need the model, since you already have it from the row. 

But even if we have ways to grab the model, I think this behavior is too quirky and too strict. Basically, you're going for this:
```html
<clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
    <clr-dg-action-overflow [clrDgActionItem]="user">
        <a class="action-item" (click)="onEdit()">Edit</a>
    </clr-dg-action-overflow>
    …
</clr-dg-row>
```
but I believe this is more natural, requires the user to learn less inputs/outputs, and is more flexible at the same time:
```html
<clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
    <clr-dg-action-overflow>
        <a class="action-item" (click)="onEdit(user)">Edit</a>
    </clr-dg-action-overflow>
    …
</clr-dg-row>
```

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -258,6 +317,17 @@
         }
     }
 
+    .datagrid-action-bar {
+        display: flex;
+        flex-flow: row nowrap;
+        align-items: stretch;
+        width: 100%;
```
jeeyun: Yes I'm thinking we can revisit this because we are recommending the use of button groups in the action bar, and the button groups themselves are supposed to be responsive as far as I know. But the button groups haven't been implemented yet. Of course the action bars aren't constrained to contain just button groups even if we recommend it, so we should definitely raise this for more UX input. cc @yenmade @reddolan

src/clarity-angular/datagrid/datagrid-action-bar.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    describe("DatagridActionBar component", function() {
+        let context: TestContext<DatagridActionBar, SimpleTest>;
+
+        beforeEach(function() {
+            context = this.create(DatagridActionBar, SimpleTest);
+        });
+
+        it("projects content", function() {
+            expect(context.clarityElement.textContent.trim()).toMatch("Hello world");
+        });
```
jeeyun: fixed

src/clarity-angular/datagrid/datagrid-row.ts
jeeyun: done

src/clarity-angular/datagrid/datagrid-action-overflow.ts
jeeyun: I had a discussion with @yenmade about this. The wanted behavior from UX point of view is to de-select other selected rows if you click on the ellipses icon (yes, that part is missing in the original spec/doc). I will modify the behavior per your suggestion.

youdz: If you need the model or the row, just inject the `DatagridRow` parent in the constructor, its `item` property is public. No need to ask the user to re-specify it.

src/clarity-angular/datagrid/datagrid-action-bar.ts
```diff
@@ -0,0 +1,21 @@
...
+    selector: "clr-dg-action-bar",
+    template: `
+        <div class="datagrid-action-bar">
+            <ng-content *ngIf="selection.current.length > 0"></ng-content>
+        </div>
+    `
+})
+export class DatagridActionBar {
+
+    constructor(private selection: Selection) {}
```
jeeyun: fixed

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component} from "@angular/core";
+import {TestContext} from "./helpers.spec";
+import {DatagridActionOverflow} from "./datagrid-action-overflow";
+
+export default function(): void {
+    describe("DatagridActionBar component", function() {
```
jeeyun: fixed

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
jeeyun: fixed

src/clarity-angular/datagrid/datagrid.ts
jeeyun: fixed

src/app/datagrid/selection/selection.html
jeeyun: I treated these demo components more like playground with incomplete details (as I am going to break up to several demo components for the website). But I updated it just for the sake of completeness.

src/app/datagrid/selection/selection.ts
jeeyun: I treated these demo components more like playground with incomplete details (as I am going to break up to several demo components for the website). But I updated it just for the sake of completeness.

src/clarity-angular/datagrid/datagrid-row.ts
jeeyun: Thanks for catching that. I actually don't think the second `*ngIf` is needed. I was checking for the wrong value there. It should be:
```
<clr-dg-cell *ngIf="rowActionService.actionableCount > 0" class="datagrid-single-select">
   <ng-content select="clr-dg-action-overflow"></ng-content>
</clr-dg-cell>
```
 This revised markup will always render the extra cell if any one of the rows have the extra cell for action. The ellipsis icon will only render if `<clr-dg-action-overflow>` exists in that row.

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
jeeyun: tests added

src/clarity-angular/datagrid/datagrid-action-overflow.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component, EventEmitter, HostListener, Input, Output, ElementRef} from "@angular/core";
+import "../../clarity-icons/shapes/essential-shapes";
```
jeeyun: Just discussed with @Shijir. We're actually supposed to package all icons clarity components depend on into core (as opposed to essential). He's going to move ellipses into core in #455. Will remove this and revert the change in the bundle task as well.

src/clarity-angular/datagrid/datagrid-action-overflow.ts
jeeyun: Thanks, will make that change. Are you comfortable with the logic that you originally raised questions about:
```
if (this.selection.selectable) {
   this.selection.current = [];
   this.selection.setSelected(this.item, true);
}
```
to achieve:
> de-select other selected rows if you click on the ellipses icon

src/app/datagrid/selection/selection.html
jeeyun: Thanks for the good suggestion. Since the action menu is a part of the actionMenuHost, I achieved this by looking for the class name of the overflow menu container in the target (and its parents), and hiding it if found. Let me know if this is a clean enough a solution.

build/tasks/bundle.js
```diff
@@ -101,7 +101,8 @@ gulp.task("bundle:clarity:js", ["typescript:clarity"], function () {
             "rxjs*": {build: false}
         },
         packages: {
-            'clarity-angular': {main: 'index.js', defaultExtension: 'js'}
+            'clarity-angular': {main: 'index.js', defaultExtension: 'js'},
+            'clarity-icons': {defaultExtension: 'js'}
```
jeeyun: reverted as per my other comment about putting all icons that clarity components depend on into core (vs essential).

src/app/datagrid/selection/selection.html
```diff
@@ -38,12 +38,27 @@
 </div>
 
 <clr-datagrid [(clrDgSelected)]="selected">
+    <clr-dg-action-bar>
+        <button type="submit" class="btn btn-link"><clr-icon shape="plus" size="16"></clr-icon> Add to group</button>
```
jeeyun: fixed

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
```diff
@@ -29,11 +52,62 @@ export default function(): void {
...
+
+        it("registers to the RowActionServiceProvider provider", function() {
+            expect(rowActionServiceProvider.actionableCount).toEqual(1);
+        });
+
+        it("unregisters when destroyed", function() {
+            context.clarityDirective.ngOnDestroy();
+            expect(rowActionServiceProvider.actionableCount).toEqual(0);
+
+        });
```
youdz: Thanks for writing such detailed unit tests here, awesome job. 👍

src/sample-app/app.component.html
```diff
@@ -62,6 +62,56 @@
...
+                    <clr-dg-action-overflow>
+                        <a class="action-item" href="#" (click)="onEdit()">Edit</a>
+                        <a class="action-item" href="#" (click)="onDelete()">Delete</a>
+                    </clr-dg-action-overflow>
+
+                    <clr-dg-cell>{{user.name}}</clr-dg-cell>
+                </clr-dg-row>
+
+                <clr-dg-footer>{{users.length}} users</clr-dg-footer>
+            </clr-datagrid>
```
youdz: Are we supposed to commit to that file? I'm not sure what our process on this is.

src/clarity-angular/datagrid/datagrid-row.ts
youdz: Oh, yes, of course the second one is not needed, I went too far there. 😝

src/sample-app/app.component.html
jeeyun: Yeah I think it's loose at this point, as this file doesn't impact anything in our build pipeline. I wanted to keep in the datagrid example there, since it's one of the more complex ones that we'd probably need to keep testing as we add more features to it.

src/app/datagrid/selection/selection.html
```diff
@@ -35,15 +35,48 @@
...
+    </clr-dg-action-bar>
+
     <clr-dg-column>User ID</clr-dg-column>
     <clr-dg-column>Name</clr-dg-column>
     <clr-dg-column>Creation date</clr-dg-column>
     <clr-dg-column>Favorite color</clr-dg-column>
 
     <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-action-overflow>
+            <a class="action-item" href="#" (click)="onEdit(user)">Edit</a>
```
adityarb88: Shouldn't these be buttons?

src/app/datagrid/selection/selection.ts
```diff
@@ -31,10 +31,31 @@ export class DatagridSelectionDemo {
...
         this.users = inventory.all;
     }
+
+    onDelete(user: User) {
+        if (user) {
+            this.toDelete = [ user ];
+        } else {
+            this.toDelete = this.selected.slice();
+        }
+        return false;
```
adityarb88: do we return false to stop the page from redirecting? Can we avoid this if we use a button instead of a link in the overflow menu?

src/app/datagrid/selection/selection.ts
```diff
@@ -31,10 +31,31 @@ export class DatagridSelectionDemo {
...
+    }
+
+    onEdit(user: User) {
+        this.toEdit = user;
+        return false;
```
adityarb88: same thing over here? do we return false to stop the page from redirecting?

src/app/datagrid/selection/selection.ts
jeeyun: Yes. I think what you suggested makes sense. I'll change them to buttons.

src/clarity-angular/datagrid/datagrid-action-overflow.spec.ts
```diff
@@ -80,11 +73,18 @@ export default function(): void {
...
+            context.detectChanges();
+            expect(context.clarityDirective.open).toBe(true);
+
+            // should close when a menu item is clicked
+            let actionItem: HTMLElement = context.clarityElement.querySelector(".action-item");
             actionItem.click();
             context.detectChanges();
             expect(context.clarityDirective.open).toBe(false);
+
         });
```
youdz: This test is starting to do a bit too much for my taste. The first part (closed initially, opens on ellipsis clicked) is already tested separately, so it can be removed. Then a test that has "XXX but YYY" clearly tests 2 things, so it could easily be split in 2 atomic tests, "XXX" and "YYY".

Still, I'm not going to block this PR for this, just a remark if you have time.

jeeyun: I agree that unit tests should be as simple / atomic as possible. Will update this before merging.

PR #509, opened 2017-02-18
[NG] Datagrid: two way binding for datagrid-column and datagrid-string-filter filters

hippee-lee:
Adds two way binding to filters on columns and string filters.

### DIFF
--- a/src/clarity-angular/datagrid/built-in/filters/datagrid-property-string-filter.ts
+++ b/src/clarity-angular/datagrid/built-in/filters/datagrid-property-string-filter.ts
@@ -5,13 +5,18 @@
  */
 import {StringFilter} from "../../interfaces/string-filter";
 import {NestedProperty} from "../nested-property";
+import {OnInit} from "@angular/core";
 
-export class DatagridPropertyStringFilter implements StringFilter<any> {
+export class DatagridPropertyStringFilter implements StringFilter<any>, OnInit {
 
     private nestedProp: NestedProperty;
 
     constructor(public prop: string, public exact = false) {
         this.nestedProp = new NestedProperty(prop);
+    }
+
+    ngOnInit() {
+        this.nestedProp = new NestedProperty(this.prop);
     }
 
     accepts(item: any, search: string): boolean {
@@ -24,5 +29,4 @@ export class DatagridPropertyStringFilter implements StringFilter<any> {
             return ("" + propValue).toLowerCase().indexOf(search) >= 0;
         }
     }
-
 }
--- a/src/clarity-angular/datagrid/built-in/filters/datagrid-property-string-filter.ts
+++ b/src/clarity-angular/datagrid/built-in/filters/datagrid-property-string-filter.ts
@@ -5,18 +5,13 @@
  */
 import {StringFilter} from "../../interfaces/string-filter";
 import {NestedProperty} from "../nested-property";
-import {OnInit} from "@angular/core";
 
-export class DatagridPropertyStringFilter implements StringFilter<any>, OnInit {
+export class DatagridPropertyStringFilter implements StringFilter<any> {
 
     private nestedProp: NestedProperty;
 
     constructor(public prop: string, public exact = false) {
         this.nestedProp = new NestedProperty(prop);
-    }
-
-    ngOnInit() {
-        this.nestedProp = new NestedProperty(this.prop);
     }
 
     accepts(item: any, search: string): boolean {
--- a/src/clarity-angular/datagrid/built-in/filters/datagrid-string-filter.spec.ts
+++ b/src/clarity-angular/datagrid/built-in/filters/datagrid-string-filter.spec.ts
@@ -3,20 +3,22 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+
 import {Component, ViewChild} from "@angular/core";
 import {TestBed, tick, fakeAsync} from "@angular/core/testing";
 import {TestContext} from "../../helpers.spec";
 import {DatagridStringFilter} from "./datagrid-string-filter";
-import {Filters} from "../../providers/filters";
+import {FiltersProvider} from "../../providers/filters";
 import {CustomFilter} from "../../providers/custom-filter";
 import {StringFilter} from "../../interfaces/string-filter";
+import {DatagridStringFilterImpl} from "./datagrid-string-filter-impl";
 
 export default function (): void {
     describe("DatagridStringFilter component", function () {
         // Until we can properly type "this"
         let context: TestContext<DatagridStringFilter, FullTest>;
         let filter: TestFilter;
-        let filtersInstance: Filters;
+        let filtersInstance: FiltersProvider;
 
         function openFilter() {
             context.clarityElement.querySelector(".datagrid-filter-toggle").click();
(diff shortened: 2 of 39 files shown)


src/clarity-angular/datagrid/built-in/filters/datagrid-property-string-filter.ts
```diff
@@ -5,15 +5,20 @@
...
+export class DatagridPropertyStringFilter implements StringFilter<any>, OnInit {
 
     private nestedProp: NestedProperty;
 
     constructor(public prop: string, public exact = false) {
         this.nestedProp = new NestedProperty(prop);
     }
 
+    ngOnInit() {
+        this.nestedProp = new NestedProperty(this.prop);
```
hippee-lee: I'm not sure why this is in here. I'll remove it.

src/clarity-angular/datagrid/datagrid-column.spec.ts
```diff
@@ -79,46 +81,77 @@ export default function(): void {
...
-                // We make the filter active to see if the Filters provider knows about it
+            it("offers two way binding on the filtered state", function () {
+                this.context = this.create(DatagridColumn, PreFilterTest, [Sort, FiltersProvider]);
+                this.context.testComponent.field = "test";
+                this.context.testComponent.filterValue = "M";
+                this.context.clarityDirective.filterValueChange.subscribe((m: string) => {
+                    expect(m).toBe("M");
+                });
+                this.context.detectChanges();
+                expect(this.context.clarityDirective.filterValue).toBe("M");
```
youdz: Same here, that's not how you want to test 2-way binding. Your test component already uses two-way binding, that's what you want to test:

```typescript
this.context.testComponent.filterValue = "M";
this.context.detectChanges();
expect(this.context.clarityDirective.filterValue).toBe("M");
this.context.clarityDirective.filterValue = "t";
this.context.detectChanges();
expect(this.context.testComponent.filterValue).toBe("t");
```

src/clarity-angular/datagrid/built-in/filters/datagrid-string-filter.spec.ts
```diff
@@ -95,24 +75,33 @@ export default function(): void {
...
-        xit("closes when the user presses Escape in the input", function() {
+        xit("closes when the user presses Escape in the input", function () {
             // TODO
             openFilter();
         });
 
-        xit("exposes an Observable to follow filter changes", fakeAsync(function() {
-            // TODO
+        xit("exposes an Observable to follow filter changes", fakeAsync(function () {
+            // TODO -  Can I prune this (Moved to filters provider line #58?)
```
youdz: Yes, I think you can remove this. We'll add more tests in the future if we have time.

src/clarity-angular/datagrid/datagrid-filter.spec.ts
```diff
@@ -34,22 +34,22 @@ export default function(): void {
                 expect(component.open).toBe(true);
             });
 
-            it("registers to the Filters provider", function() {
+            it("registers to the FiltersProvider provider", function() {
                 expect(filterService.getActiveFilters()).toEqual([]);
-                component.filter = filter;
+                component.setFilter( filter );
```
youdz: Please use `component.customFilter = filter` everywhere in this spec, since that's what the `@Input` does. You jumped a step there, not a big deal right now but the setter with the `@Input` might do more in the future.

src/clarity-angular/datagrid/providers/filters.spec.ts
```diff
@@ -74,21 +74,21 @@ export default function(): void {
...
             this.filtersInstance.change.subscribe(() => nbChanges++);
             filter.toggle();
             expect(this.filtersInstance.getActiveFilters()).toEqual([filter]);
             expect(nbChanges).toBe(1);
-            unregister();
+            registeredFilter.unregister();
             expect(this.filtersInstance.getActiveFilters()).toEqual([]);
             filter.toggle();
-            expect(nbChanges).toBe(1);
+            expect(nbChanges).toBe(2); // first is toggle(). second is unregister()
```
youdz: It took me a little while to understand what was happening there. The unregister triggers a change _because the filter was active_. So here, to make this clear, you just need to add another expect:
```typescript
registeredFilter.unregister();
expect(this.filtersInstance.getActiveFilters()).toEqual([]);
expect(nbChanges).toBe(2);
filter.toggle();
expect(nbChanges).toBe(2);
```
It makes it clear that the toggle isn't the cause of the extra change.

That said, I would really like to have two unit tests for this "unregister triggers a change" behavior:
1) Unregister an inactive filter, no change expected.
2) Unregister an active filter, change expected.

src/clarity-angular/datagrid/providers/filters.ts
```diff
@@ -57,14 +56,21 @@ export class Filters {
...
+        let hasUnregistered = false;
+        let registered = new RegisteredFilter(filter, () => {
+            if (hasUnregistered) { return; }
             subscription.unsubscribe();
             this._all.splice(index, 1);
-        };
+            if (filter.isActive()) {
+                this._change.next([]);
+            }
+            hasUnregistered = true;
```
youdz: Could you add a unit test for this `hasUnregistered` check, just to be sure?
Manually register 3 filters, then unregister the middle one twice in a row. Change should only be fired 1 time if the filter was active, and the other two filters should still be present.

src/clarity-angular/datagrid/datagrid-column.spec.ts
```diff
@@ -196,32 +229,46 @@ export default function(): void {
...
+                this.context.testComponent.show = true;
+                this.context.clarityDirective.filters.add(new TestFilter());
+                this.context.clarityDirective.filters.add(new TestFilter());
+                this.context.detectChanges();
+                this.context.testComponent.show = false;
+                this.context.detectChanges();
+                let filters = this.context.clarityDirective.filters.getActiveFilters();
+                filters.map((filter: TestFilter) => {
+                    expect(filter instanceof TestFilter).toEqual(true);
+                });
```
youdz: You're not checking that they're still here. For all you know, `filters` might be an empty array and the test would still pass. 😛

src/clarity-angular/datagrid/built-in/filters/datagrid-string-filter.ts
```diff
@@ -77,8 +77,12 @@ export class DatagridStringFilter extends DatagridFilterRegistrar<DatagridString
     /**
      * Common setter for the input value
      */
+    public get value() {
+        return this.filter.value;
+    }
     @Input("clrFilterValue")
     public set value(value: string) {
```
youdz: You HTML `<input>` still has `[(ngModel)]="filter.value"`. So if I type something in it, it will update `this.filter.value`, not `this.value`.  Which means you "bypass" this setter, so the `filterValueChange` output is never emitted.
All you need to do to fix that is switch to `[(ngModel)]="value"`.

I'm still figuring out what the best pattern for our unit tests is, because this could have slipped through while the tests do pass. And manually grabbing a native input element and typing in it in Jasmine is really something I would not want to do in every test we have. Until we figure out a solution, the change I mentioned in this comment is good enough for me.

src/clarity-angular/datagrid/datagrid-filter.spec.ts
```diff
@@ -36,7 +36,8 @@ export default function(): void {
 
             it("registers to the FiltersProvider provider", function() {
                 expect(filterService.getActiveFilters()).toEqual([]);
-                component.setFilter( filter );
+                // component.setFilter( filter );
+                component.customFilter = filter;
```
youdz: You have a few more of these in this `datagrid-filter.spec.ts` file, that's what I meant by replacing all occurrences in this spec. I can't comment on them because they haven't change, but I see them on lines 45 and 53.

src/clarity-angular/datagrid/built-in/filters/datagrid-string-filter.ts
hippee-lee: Gotcha. Now I understand.

PR #511, opened 2017-02-21
[NG] Datagrid: switch to flexbox layout

youdz:
## This is ready to go 👍 

We moved from `table-layout: auto`, which was very useful for smartly sizing columns but far too strict to allow all the features we are planning, to a flexbox layout.
    
Because of this, we have to manually handle column sizes, which bleeds quite a lot into our Typescript code. To make sure this is done as cleanly as possible, I wrote all the sizing logic in separate providers and classes.
    
One important pattern to note is the use of the renderer to directly style elements and add classes to them, rather than use more "natural" style and host binding with Angular. The reason is that said bindings introduce watchers and are factored in change detection, which can severely impact the performance of our datagrid. This is why I went with custom event-based size recomputation, to make sure we always did the strict minimum of work, and Angular could simply ignore it during change detection.

This change has been written in a way that does not break existing integration, it only modified our own components' internal template and styles.
    
This refactor allows 2 nice features out of the box:
1. Datagrids can now scroll easily if you restrict them to their container. The header and footer stay fixed, only the body containing the rows scrolls. It's as simple as setting the height of the datagrid to a specific value, or making it 100% to fill its container.
2. Specifying a header width through CSS or `[style.width]` binding locks the corresponding column size and makes it non-flexible. No need to duplicate that information on the cells, just the header is enough.

Tested on Chrome, Firefox, Safari, IE10, IE11 and Edge.

Here is an example of "smart" column sizing, where absolutely no width is specified by the customer:
smart-columns-size

And two examples of scrolling datagrids, vertically and horizontally:
vertical-scrolling
horizontal-scrolling

### DIFF
--- a/src/clarity-angular/popover/popover.ts
+++ b/src/clarity-angular/popover/popover.ts
@@ -1,0 +1,132 @@
+/*
+ * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+/*
+ * Do NOT Angular this up. It assumes we're in the DOM, plays with native elements, ...
+ * It could potentially be used as part of clarity-ui as a vanilla Javascript helper.
+ *
+ * WARNING: This is a quick prototype version, use at your own risks.
+ */
+
+export enum Direction {
+    RIGHT
+}
+
+export interface PopoverOptions {
+    offsetX?: number;
+    offsetY?: number;
+    userAnchorParent?: boolean;
+}
+
+const POSITION_RELATIVE = "relative";
+const POSITION_ABSOLUTE = "absolute";
+const POSITION_FIXED = "fixed";
+
+const OVERFLOW_SCROLL = "scroll";
+const OVERFLOW_AUTO = "auto";
+
+interface InlineOverflow {
+    both: string;
+    x: string;
+    y: string;
+}
+
+const OVERFLOW_HIDDEN: InlineOverflow = {
+    both: "hidden",
+    x: "hidden",
+    y: "hidden"
+};
+
+export class Popover {
+    constructor(private element: any) {
+        // Browsers don't agree with what to do if some of these are not specified, so we set them all to be safe.
+        element.style.position = POSITION_ABSOLUTE;
+        element.style.top = 0;
+        element.style.left = 0;
+    }
+
+    public anchor(anchor: any, direction: Direction,
+                  {offsetX = 0, offsetY = 0, userAnchorParent = false}: PopoverOptions = {}) {
+        // TODO: we are assuming here that the popover is inside or next to the anchor.
+        // We'd need to go up the popover tree too otherwise
+        this.preventScrolling(anchor);
+        if (userAnchorParent) {
+            anchor = anchor.parentNode;
+        }
+        let anchorRect = anchor.getBoundingClientRect();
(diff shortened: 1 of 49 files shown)


src/clarity-angular/popover/popover.ts
```diff
@@ -0,0 +1,132 @@
...
+        return position === POSITION_RELATIVE || position === POSITION_ABSOLUTE || position === POSITION_FIXED;
+    }
+
+    /*
+     * We prevent the containers up to the first positioned one from scrolling
+     */
+
+    private originalOverflows: {e: any, overflow: InlineOverflow}[] = [];
+
+    private preventScrolling(e: any) {
```
jeeyun: We have an existing `ScrollingService` class with two methods (`stopScrolling` and `resumeScrolling`) that I believe the modal is using. Would this be a good candidate to be moved out to there as we generalize `ScrollingService` to be used by multiple components? Doesn't need to be addressed in this PR per se, but just making a note. Maybe I need to look into that task as we come up with an overall solution for popovers.

src/clarity-angular/popover/popover.ts
```diff
@@ -0,0 +1,132 @@
...
+            anchor = anchor.parentNode;
+        }
+        let anchorRect = anchor.getBoundingClientRect();
+        let popoverRect = this.element.getBoundingClientRect();
+        let leftDiff: number;
+        let topDiff: number;
+        // TODO: obviously, handle direction
+        switch (direction) {
+            case Direction.RIGHT:
+                leftDiff = anchorRect.left + anchorRect.width - popoverRect.left + offsetX;
```
jeeyun: I think the `- popoverRect.left` isn't needed here..? Shouldn't `anchorRect.left + anchorRect.width + offsetX` suffice?

youdz: If you remove the `- popoverRect.left`, that means you're only positioning with respect to the anchor. In other words, you assume the popover's containing block (the closest parent with position absolute or relative) is the body. So it would break as soon as you have a positioned element anywhere up the tree.

Here is how I would explain what's happening here: by preventing parents* to scroll, we kind of "lock" the position of both the anchor and the popover on the screen. We have no idea what these positions are, because they depend on where the relative and absolute parents are, their own CSS styles, ... So the safest bet is simply to look at the **difference** of their positions, and adjust this with `transform: translate` to put the popover right next to the anchor. By using this difference, we do not rely in any way on parents or the rest of the DOM, the structure could be super complex and it would still position correctly.


* It's a little more involved than just all parents, see the walk-up-the-tree from `preventScrolling()` for more info.

src/clarity-angular/popover/popover.ts
youdz: `ScrollingService` only prevents the body from scrolling. This walks up the tree and prevents any number of components from scrolling.
On top of that, we stop going up the tree when we find a positioned element (`absolute`, `relative` or `fixed`) which is very popover-specific and doesn't make sense for the Modal case.
I don't think there is anything reusable between these two.

src/clarity-angular/popover/popover.ts
jeeyun: Thanks for the detailed explanation. I think I have something that'll work for different direction cases now. Coming up in a future PR.

youdz:
Updated with a cleaner architecture. Now it's time to fix the broken unit tests because of the new providers, and write new ones for all the added features.

Issue #508, opened 2017-02-17
Icon bug fixes for 8.8 release


lil-kim:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
- Ellipsis icon (horizontal) should be vertically centered on canvas
- There should be a solid version of the Filter icon on the Iconography page

### Updated files attached:
[icon fixes.zip](https://github.com/vmware/clarity/files/784419/icon.fixes.zip)

hippee-lee:
PR #514 is merged. Closing this.

Issue #498, opened 2017-02-16
Release icons for 0.8.8


Shijir:
Release the following icons for 0.8.8: 
[icons.2017-02-01.zip](https://github.com/vmware/clarity/files/781102/icons.2017-02-01.zip)

jaffoneh:
@Shijir can we merge this with #500?

hippee-lee:
PR #514 is merged. Closing this.

PR #514, opened 2017-02-21
[ICONS] Adding new icons

### LINKED ISSUE #500: New set of Icons (vol 4) to be Published
```
[ ] bug
[ ] feature request
[X] enhancement
```

### Description
The purpose of this ticket is to track icon deliveries and implementation (post to the site; add to codebase). A new set of icons will be attached to this ticket once a week by Lilia. Every 2weeks, the 2 sets are to be released to the Clarity site.

Shijir:

Resolves #500, #498, and #508.


Shijir:
clarity

Jinnie:
Great work!
I was looking for both ellipsis and angle-doubles yesterday and had to consider substitutes.
Btw, can't ellipsis be manipulated with direction keywords/attribs? Or the base is different and vertical is centered?

Shijir:
Hi @Jinnie, You can manipulate ellipsis-horizontal and ellipsis-vertical shapes through `dir` attribute just like any other shapes. We could have chosen the route of offering only of them and let users rotate the icon to their choice of direction. The problem with that approach is that the icon has a "badged" state. So when the icon is rotated, users will face cases that badges don't appear in the appropriate position, which is the upper-right side.

As you can see, I added `dir="left"` attribute to the `ellipsis-vertical` shape and its badge's position looks off compared to other icons in `badged` state.

screenshot 2017-02-22 10 39 08

Jinnie:
Thanks for the explanation. This probably applies to all icons. I'll keep it in mind

PR #519, opened 2017-02-22
[ui] hide inactive tab content if media type is screen

### LINKED ISSUE #457: All sections inside tab is always visible even it contains aria-hidden attribute as true
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
The tab section which contains the attribute aria-hidden as "true" then it should not be visible when its respective list is in-active.

### Actual behavior
The tab-content(all sections) inside the tab template is always visible irrespective of contains attribute aria-hidden=true .

### Reproduction of behavior
Tried the example mentioned in the documentation, this issue occurs with that code itself.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 0.8.4

* **OS and version:** Windows 8.1

* **Browser:** all

jeeyun:
I think at some point we changed from using `aria-hidden` to `data-hidden` for hiding inactive content panels. Should we just keep this new style:
```
  @media screen {
    section[aria-hidden="true"] {
      display: none;
    }
  }
```
which won't block the screen readers from reading the content of the inactive tabs, and get rid of using `data-hidden` all together?

resolves #457.


src/clarity-angular/tabs/_tabs.clarity.scss
```diff
@@ -5,6 +5,12 @@
 @import "../utils/mixins.clarity";
 
 @include exports('tabs.clarity') {
+  @media screen {
+    section[aria-hidden="true"] {
+      display: none;
+    }
+  }
+
   [data-hidden="true"] {
```
mathisscott: I don't know if we need data-hidden. I removed it from the new wizard. Maybe we should just deprecate it and remove it in 0.9?

jeeyun: Sounds good. I will leave it in for this PR. I did check and only the wizard was using it, so I think it'll get removed cleanly in 0.9.

PR #522, opened 2017-02-22
[CORE] Fix Import Order [Do Not Review or Merge]

adityarb88:
The order of imports in `main.scss` was definitely incorrect. There were some files which were importing `variables.clarity.scss` but `variables.clarity.scss` itself was being imported after those files. I have rearranged the imports to get these issues fixed.

CSS Regression Tests Pass https://travis-ci.org/adityarb88/clarity/builds/204293171

Tested on: Chrome

### DIFF
--- a/src/clarity-angular/main.scss
+++ b/src/clarity-angular/main.scss
@@ -7,50 +7,79 @@
 
 /* @echo VERSION */
 
-@import "./utils/mixins.clarity";
-@import "./utils/normalize.clarity";
-@import "./utils/reboot.clarity";
-@import "./utils/helpers.clarity";
-@import "./utils/layers.clarity";
+/** Order of imports is important **/
+@import "./utils/layers.clarity"; // no dependencies on other clarity scss
+@import "./utils/helpers.clarity"; // no dependencies on other clarity scss
+@import "./color/color.clarity"; // depends on colors.clarity, helpers.clarity, contrast-cache.clarity
+@import "./utils/mixins.clarity"; // depends on color.clarity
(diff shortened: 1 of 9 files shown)


src/clarity-angular/main.scss
```diff
@@ -7,52 +7,80 @@
...
+/** Order of imports is important **/
 @import "./utils/overwrites.clarity";
-@import "./utils/mixins.clarity";
-@import "./utils/normalize.clarity";
-@import "./utils/reboot.clarity";
-@import "./utils/helpers.clarity";
-@import "./utils/layers.clarity";
+@import "./utils/layers.clarity"; // no dependencies on other clarity scss
+@import "./utils/helpers.clarity"; // no dependencies on other clarity scss
+@import "./color/color.clarity"; // depends on colors.clarity, helpers.clarity, contrast-cache.clarity
```
jeeyun: If `mixins.clarity` always imports `color.clarity` inside it, is it necessary to import it again in this `main.scss`? Or is this so that the dependency is more explicit and clear? I don't think there's harm in importing twice, since we export style so that they are not included twice in the compiled css, but just noting it here.

I guess in general maintaining this 'master list' of imports is kind of a pain and error-prone. I don't have an alternative solution at the moment, but maybe something we can evaluate again in the future.

adityarb88: I think we import it again because there are few files like `variables.clarity.scss` which import `color` but not `mixins`, no?

adityarb88: But you are right. We need to think of a better solution here.

PR #526, opened 2017-02-23
[NG] Fix datagrid pagination page values when datagrid is empty

### LINKED ISSUE #518: [UI] Datagrid pagination page values incorrect
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When the datagrid is empty, and pagination is activated, as on the documentation page, a string like "0 - 0 of 0 parameters" is expected in the footer

### Actual behavior
The actual string is "1 - 8 of 0 parameters"

screen shot 2017-02-22 at 5 01 51 pm

### Reproduction of behavior
I tried creating a plunker with this behaviour, but the result is even worse. No pagination at all, plus incorrect values.
https://plnkr.co/mEjMJMtRSUH7AqdEEc4c

### Environment details

* **Angular version: 2.4.8**

* **Clarity version: 0.8.7** 
 
* **OS and version: OS X El Capitan** 

* **Browser: Tested with Chrome and Firefox**

Jinnie:
When there are 0 items in the datagrid, and paging is active, firstItem is reported as 1, and lastItem -as the page size. This could lead to footers like "1 - 10 of 0 items". It is correct to return -1 in this case, as these values are indices, and 0 would be a valid index of a non-existing 1st element (thus footers like "1 - 1 of 0" will be produced).

Fixes #518

### DIFF
--- a/src/clarity-angular/datagrid/providers/page.spec.ts
+++ b/src/clarity-angular/datagrid/providers/page.spec.ts
@@ -25,6 +25,13 @@ export default function(): void {
             expect(this.pageInstance.last).toBe(5);
         });
 
+        it("has firstItem and lastItem -1 when totalItems are 0", function() {
+            this.pageInstance.size = 10;
+            this.pageInstance.totalItems = 0;
+            expect(this.pageInstance.firstItem).toBe(-1);
+            expect(this.pageInstance.firstItem).toBe(-1);
+        });
+
         it("computes the indexes of the first and last displayed items", function() {
             this.pageInstance.size = 10;
             this.pageInstance.totalItems = 42;
--- a/src/clarity-angular/datagrid/providers/page.ts
+++ b/src/clarity-angular/datagrid/providers/page.ts
@@ -108,6 +108,9 @@ export class Page {
      * Index of the first item displayed on the current page, starting at 0
      */
     public get firstItem(): number {
+        if (this.totalItems === 0) {
+            return -1;
+        }
         if (this.size === 0) {
             return 0;
         }
@@ -118,6 +121,9 @@ export class Page {
      * Index of the last item displayed on the current page, starting at 0
      */
     public get lastItem(): number {
+        if (this.totalItems === 0) {
+            return -1;
+        }
         if (this.size === 0) {
             return this.totalItems - 1;
         }


Jinnie:
Also a manual testing screenshot:
screen shot 2017-02-23 at 4 24 58 pm

Jinnie:
Hi @youdz,
Thanks for the detailed explanation. I understand the problem is complex. I even rebuilt the app in clarity source to reproduce it. If [clrDgTotalItems] is bound to a previously known data size, all works well. Even with my commit code :)
But If I remove it (size unknown scenario) - then your prediction gets live and an empty datagrid is shown.
Then I rolled back my change and tried again, still without setting clrDgTotalItems. First 10 items were immediately loaded, proving your point.
But what I can't understand is how the user will load the next page of data then, as we are only showing him the first 10 items with no option to navigate further, to get new refresh event and load more items.
See this screenshot (it's from "/src/app/datagrid/server-driven"):
screen shot 2017-02-25 at 12 03 03 am

My own use case is not so important, I have ways to workaround it, but I am curious about the solution, as I've been working on an "indefinite datagrid" web component in the past (in vanilla JS) and the topic is still interesting to me, so I'd appreciate some more information on it when you have some time :)

Jinnie:
Another glitch that we hit with the unknown size scenario is when the server returns less items than the page size is:
screen shot 2017-02-25 at 8 25 15 am

It is clear to me how to fix the code and tests for the case where expected size is defined in the page.totalItems, as in the expamle
`        <clr-dg-pagination #pagination [clrDgTotalItems]="total"></clr-dg-pagination>
` 
But the scenario with the unknown size still remains quite vague to me. The easy answer is just not to try to display item data in the footer (~~"1 - x of y"~~). And although the page service works correctly in terms of internal use for pagination mechanics, I don't think that the pagination API itself, which is public (datagrid-pagination#firsItem, ~#lastItem, ~#totalItems), should be presenting incorrect data, based on assumptions.
I need more clarity 😃 on this, and on my question from the previous comment - how the user is expected to load new dynamic data and get to next page when the total size is unknown, and there are no navigation handlers for more pages.
Once I have this clear, the fix should be pretty straightforward + bonus tests for both cases 😈 

Best regards!

Jinnie:
I would prefer to close this pull request now, as there are at least two viable workarounds, with either conditional footers, or using pluralization, and I don't like stuck requests myself 😃 
I will leave the [related issue](https://github.com/vmware/clarity/issues/518) for you to decide if it gets closed too, or remains open for future tracking.

Best regards!

PR #530, opened 2017-02-25
Fix the Event name for Wizard Open state change

kars7e:
Angular 2 adds 'Change', not 'Changed' when performing a two-way binding.

### DIFF
--- a/src/app/wizard/wizard-options.demo.html
+++ b/src/app/wizard/wizard-options.demo.html
@@ -24,7 +24,7 @@
         </td>
     </tr>
     <tr>
-        <td class="left">(clrWizardOpenChanged)</td>
+        <td class="left">(clrWizardOpenChange)</td>
         <td>true, false</td>
         <td>N/A</td>
         <td class="left">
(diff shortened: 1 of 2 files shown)


kars7e:
I'm not sure how to find out which test failed, and why

Jinnie:
Hi @kars7e,

This is the test run:
https://travis-ci.org/vmware/clarity/jobs/205167207
So, this is the failing test:
https://github.com/vmware/clarity/blob/master/gemini/tests/wizard.js

kars7e:
Thanks @Jinnie, it seems that I cannot access the screenshots.zip artifact, I don't see any URL on Travis build page. Is there anyway I can access them, or do I have to run gemini locally?

hippee-lee:
Hi @kars7e 
I'm looking into this for you. Forks have to be configured to upload the images to s3.
I'm running it on a local setup to see what is going on and will update here when I know more.

hippee-lee:
Hey @kars7e,

It looks like there is a timing issue with the css regression test on this component that is causing that test to fail. 

When I talked to the team about this it was highlighted that renaming `clrWizardOpenChanged` would be a breaking change for Clarity users. 

FYI - the Wizard is currently being refactored. You can keep track of it here: #156

kars7e:
Thanks @hippee-lee. I've used:
`[clrWizardOpen]="open" (clrWizardOpenChanged)="open=$event"`
as a workaround for now in my project. Should we add a comment in documentation that two-way binding doesn't work atm and the above workaround until the refactoring is finished? I spent couple hours debugging it before I found that bug, maybe it would save others some time.

mathisscott:
Closing this pull request. Added #542 to note the workaround in our documentation.

PR #543, opened 2017-03-01
[CORE] AoT sanity test

### LINKED ISSUE #525: [NG] Templates accessing private properties of components
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Since templates are compiled down to a different object than the component itself, they should not access private properties of the component. In particular, doing so raises Typescript errors when using AoT compilation.

### Actual behavior

Many of our templates still access private properties of components. For instance, the `page` property of the DatagridPagination component: https://github.com/vmware/clarity/blob/master/src/clarity-angular/datagrid/datagrid-pagination.ts#L40

### Reproduction of behavior

Try compiling [DatagridPagination](https://github.com/vmware/clarity/blob/master/src/clarity-angular/datagrid/datagrid-pagination.ts#L40) in AoT mode.

Shijir:
Fixes #525.

### DIFF
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "declaration": false,
+    "module": "umd",
+    "moduleResolution": "node",
+    "sourceMap": false,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "lib": [ "es2015", "dom" ],
+    "removeComments": false,
+    "noImplicitAny": true,
+    "typeRoots": [
+      "node_modules/@types/"
+    ]
+  },
+  "include": [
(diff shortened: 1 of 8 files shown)


tsconfig.test.json
```diff
@@ -0,0 +1,25 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "declaration": true,
```
youdz: You don't need to produce declaration files or anything extra, you're just throwing away the result of the compilation anyway.

tsconfig.test.json
```diff
@@ -0,0 +1,25 @@
...
+    "module": "umd",
+    "moduleResolution": "node",
+    "sourceMap": true,
```
youdz: Same here. No point in emitting sourcemaps, you're just losing time.

PR #549, opened 2017-03-02
[WEB] Updates datagrid documentation for presets and two way binding

hippee-lee:


src/demos/datagrid/filtering/examples.ts
```diff
@@ -1,4 +1,4 @@
-/*
+``/*
```
Shijir: what is this :p

PR #551, opened 2017-03-02
[UI] Cleanup sidenav css

jeeyun:
- removing sidenav header styling (part of original spec, but no longer used/documented/supported)
- collapsible sidenav list now does not rely on css position property (necessary prerequisite for fixing clipping popover)

I don't know why but there are minor pixel differences in our icons so I'm replacing the images with new reference shots. For example (the purple part is the diff):

screen shot 2017-03-02 at 2 18 16 pm

The differences are very subtle, and very unlikely that they come from my changes in `_sidenav.clarity.scss`. Nevertheless, Travis build will not pass with the diff so I'm checking in the new images.


jeeyun:
@adityarb88: yes, manually/visually inspected in all browsers that we support.

Issue #529, opened 2017-02-24
[ng] datagrid action overflow throws an error on loading for AoT compiled app


jeeyun:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Should not throw any exception when loading a datagrid with action overflow menu

### Actual behavior

Upon loading an application that is AoT compiled, I get an exception in console:

```EXCEPTION: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'.```

However, the datagrid functions fine after that, and renders and works normally. This does not happen in JiT compiled application.

### Reproduction of behavior

https://plnkr.co/edit/bAYeDO 
This loads fine in plunkr.. but same code doesn't work for AoT.

### Environment details

* **Angular version:** 2.4.1

* **Clarity version:**  0.8.8

* **OS and version:** all

* **Browser:** all

**Comment by ryanpflem** [comment]
I'm also getting this exception, except I'm using JIT and the datagrid does not function after.
I'm on Clarity v. 0.8.8 and Angular v. 2.4.8

My datagrid:
```
<clr-datagrid *ngIf="claims">
  <clr-dg-column [clrDgField]="'claimNumber'">Claim No.</clr-dg-column>
  <clr-dg-column [clrDgField]="'claimantFirstName'">First Name</clr-dg-column>
  <clr-dg-column [clrDgField]="'claimantLastName'">Last Name</clr-dg-column>
  <clr-dg-column [clrDgField]="'_changed'">Last Adjusted</clr-dg-column>

  <clr-dg-row *clrDgItems="let claim of claims" [clrDgItem]="claim">
      <clr-dg-action-overflow>
        <button class="action-item" (click)="onSelect(claim)">View</button>
      </clr-dg-action-overflow>
    <clr-dg-cell>{{claim?.claimNumber}}</clr-dg-cell>
    <clr-dg-cell>{{claim?.claimantFirstName}}</clr-dg-cell>
    <clr-dg-cell>{{claim?.claimantLastName}}</clr-dg-cell>
    <clr-dg-cell>{{claim?._changed | date:'shortDate'}}</clr-dg-cell>
  </clr-dg-row>

  <clr-dg-footer>
    {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}} of {{pagination.totalItems}} Claims
    <clr-dg-pagination #pagination [clrDgPageSize]="10"></clr-dg-pagination>
[...]

PR #563, opened 2017-03-06
[NG] fixed expression has changed error

Shijir:
Resolves https://github.com/vmware/clarity/issues/529


src/clarity-angular/datagrid/datagrid.ts
```diff
@@ -35,7 +39,10 @@ import {DatagridRenderOrganizer} from "./render/render-organizer";
 export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
 
     constructor(public selection: Selection, private sort: Sort, private filters: FiltersProvider,
-                private page: Page, public rowActionService: RowActionService, public items: Items) {}
+                private page: Page, public items: Items, public rowActionService: RowActionService) {
+    }
+
+    public hasActionableRow = false;
```
jeeyun: do you need this? looks like the template is using rowActionService.hasActionableRow directly?

src/clarity-angular/datagrid/datagrid.spec.ts
```diff
@@ -157,13 +158,53 @@ export default function(): void {
...
         });
 
+        describe("Actionable rows", function () {
+            let context: TestContext<Datagrid, ActionableRowTest>;
+            let rowActionService: RowActionService;
+            let headActionOverflowCell: HTMLElement;
+            let actionOverflowCell: HTMLElement[];
+            let actionOverflow: HTMLElement[];
+
+            it("have cells for action overflows if there is at least one of them.", function () {
```
youdz: Nitpick: "it **has** cells..." 😛 

Same for the second test added.

src/clarity-angular/datagrid/datagrid.spec.ts
```diff
@@ -157,13 +158,53 @@ export default function(): void {
...
+            it("have no cells for action overflows if there is none of them.", function () {
+                context = this.create(Datagrid, ActionableRowTest);
+                rowActionService = context.getClarityProvider(RowActionService);
+                context.testComponent.showIfGreaterThan = 10;
+                context.detectChanges();
+                actionOverflow = context.clarityElement.querySelectorAll("clr-dg-action-overflow");
+                expect(actionOverflow.length).toEqual(0);
+                context.detectChanges();
+                expect(rowActionService.hasActionableRow).toBe(false);
+                context.detectChanges();
```
youdz: Why do you need these 2 extra change detections? That's surprising. You're not performing any new action, just doing`expect()`.

src/clarity-angular/datagrid/datagrid.ts
Shijir: That's a good catch. I will remove it.

youdz: Not used anymore, you should remove it.

src/clarity-angular/datagrid/datagrid.ts
```diff
@@ -176,18 +186,36 @@ export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
...
+     * We get deep nested DatagridActionOverflow children components, listen to the changes in them,
+     * and figure out if the datagrid has at least one actionable row.
+     */
+    @ContentChildren(DatagridActionOverflow, { descendants: true }) public actionableRows: QueryList<DatagridRow>;
+
     ngAfterContentInit() {
-        // TODO: Move all this to ngOnInit() once https://github.com/angular/angular/issues/12818 goes in.
-        // And when we do that, remove the manual step for each one.
+
+        this.actionableRows.changes.subscribe(() => {
```
youdz: Why did you remove this comment, it should stay there.

Also, it's incredibly relevant here since you forgot to write the explicitly mentioned "manual step". You should do `this.rowActionService.hasActionableRow = this.actionableRows.length > 0;` once on initialization outside of the subscription too, because at that point the QueryList is already populated with the initial value. (See the same behavior on `this.items` right underneath.)

Finally, please add this subscription to `this._subscriptions`, to make sure it's properly unsubscribed on destroy. Once again, see the existing code right after this.

src/clarity-angular/datagrid/datagrid.spec.ts
Shijir: Addressed this issue by having two consecutive change detectors and added in TODO explanation.

src/clarity-angular/datagrid/datagrid.spec.ts
Shijir: Addressed this issue.

src/clarity-angular/datagrid/datagrid.ts
Shijir: Addressed this issue.

src/clarity-angular/datagrid/datagrid.ts
Shijir: Sorry. The comment was lost during fixing the rebase conflicts. Also, I've added the subscribtion to this._subscriptions.

Issue #247, opened 2016-12-14
Button Group component (new)


reddolan:
### Description
A Button Group is a collection of Buttons aligned side by side with no space between them. This can be used to create a collection of similar actions, to save space, or to respond to different size containers.

### Use cases under consideration
- mixing colored buttons for action differentiation
- adding dropdown menus for additional options
- a responsive experience to handle smaller containers
- icon buttons
- checkbox and radio group selections

#### Basic
- Normal Button design options are available to Button Groups as well
- Available for each type of button
- Group together similar types of actions that a user can do
- Can be used to maximize UI space

![basic](https://cloud.githubusercontent.com/assets/1858226/24614548/2143289e-1851-11e7-9258-0e1a1acd04b0.png)

#### Mixed Classes
- action types can be mixed in a Button Group
- we recommend that primary actions be the first in a group, followed by secondary actions
- we recommend that warning actions be the last in a group

![mixed-classes](https://cloud.githubusercontent.com/assets/1858226/24615320/8fac6c08-1853-11e7-805c-29ee2f8dcb76.png)

#### Icons
- icons used in Button Group will have minimum button width of 72px
- flat Buttons have no minimum width. There is only 24px space between icons
- flat Button Groups can Action Blue or Grey. Grey is used with Grid and Table actions

![icons](https://cloud.githubusercontent.com/assets/1858226/24615795/e748d342-1854-11e7-9975-943b2a1e0951.png)

#### Dropdowns
- used when a button has a set of actions related to it
- caret is to the right of the text or icon
- the dropdown will align to the left-most edge of the button or caret button in Split

![dropdowns](https://cloud.githubusercontent.com/assets/1858226/24615811/f02aa6d4-1854-11e7-826b-cfe4757939d6.png)

#### Overflow
- used for preserving space when many buttons are present
- can be mannually configured or set to automatically show based on space
- can be configured “x” number of buttons and “y” number of actions in the dropdown

![overflow](https://cloud.githubusercontent.com/assets/1858226/24615828/faa8bf06-1854-11e7-8a05-97d1d3f1e741.png)

#### Multiple Groups
- Button Groups next to one another
- use when you want similar actions to be grouped together and separated from others
- uses normal spacing between Buttons (12px)
- we recommend starting with primary actions followed by secondary actions

![multiple-groups](https://cloud.githubusercontent.com/assets/1858226/24615836/07905738-1855-11e7-8035-28e3125f2267.png)

#### Checkbox
[...]

PR #567, opened 2017-03-07
Button Groups [DO NOT MERGE]

adityarb88:
Fixes: #247 

Version 1 contains:
- `clr-button` and `clr-button-group` component
- Input (`clrInMenu`) on `clr-button` to determine whether it belongs inline or in the overflow menu (this version will not listen to the parent container size to figure out responsiveness)
- Styles for button groups - Solid, Outline, Flat, Mixed
- Icon Button Groups
- Checkbox and Radio Button Groups

Markup:

```
    <clr-button-group>
        <clr-button>1</clr-button>
        <clr-button>2</clr-button>
        <clr-button>3</clr-button>
        <clr-button>4</clr-button>
        <clr-button [clrInMenu]="true">5</clr-button>
        <clr-button [clrInMenu]="true">6</clr-button>
    </clr-button-group>
```

Gemini Screenshots attached

Browser Tests:
Chrome, IE 10, 11, Edge, Firefox, Safari


src/clarity-angular/button-group/_button-group.clarity.scss
```diff
@@ -7,24 +7,86 @@
...
+            border-bottom-right-radius: 0;
+        }
 
-    .btn:not(:first-child) {
-      border-top-left-radius: 0;
-      border-bottom-left-radius: 0;
-    }
+        &.btn-primary .btn:not(:last-child),
+        &.btn-success .btn:not(:last-child),
+        &.btn-danger .btn:not(:last-child),
```
mathisscott: While I generally approve of all this, I just cleaned up the modal.scss in the Wizard refactor and removed 2+ layers of specificity to the CSS selectors due to code like this.

Please make sure that all this specificity is needed. Specifically the child selectors and nesting.

src/clarity-angular/button-group/_button-group.clarity.scss
```diff
@@ -7,24 +7,86 @@
...
 
-    display: flex;
+        .btn {
+            margin: 0;
+        }
 
-    .btn {
-      margin-right: 0;
-    }
+        .btn:not(:first-child) {
```
mathisscott: Can this be cleaned up by moving the `:not`s into the `.btn` declaration above?

```
.btn {
  &:not(...) {
  }
}
```

src/clarity-angular/buttons/_buttons.clarity.scss
```diff
@@ -215,6 +241,64 @@ $clr-button-appearance-map: (
         @include generateButton();
         @include clr-getStylesFromMap($clr-button-appearance-map, form);
         @include populateButtonProperties(default);
+    }
+
+    .btn-group {
+        //.btn-outline-<state> is the new BS4 naming convention
+        //TODO: we support both the previous one and the new one but we will deprecate the old ones from the example in 0.5.0
+        &.btn-secondary > .btn,
```
mathisscott: Since we aren't getting any real efficiency out of duplicating the `btn-*` classes here, I would prefer to change it so that they follow a pattern similar to SuitCSS. So...

`.btn-group--success`, `.btn-group--primary`, etc.

That way we can remove all the `&.btn-*` selectors and go straight to `.btn-group--*` selectors.

That seems cleaner to me.

src/clarity-angular/buttons/_buttons.clarity.scss
```diff
@@ -283,4 +360,171 @@ $clr-button-appearance-map: (
             transform: translate3d(0px, -2px, 0);
         }
     }
+
+    //Current Margins - We might be getting rid of this in 0.9.0
+    .btn {
+        margin: $clr-button-vertical-margin $clr-button-horizontal-margin $clr-button-vertical-margin 0;
+
+        &.btn-link {
```
mathisscott: Why does this need to be nested?

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -67,162 +67,187 @@ $dropdown-white: clr-getColor(lightest);
...
+                opacity: 0.4;
+                user-select: none;
+            }
 
-                &.disabled:active,
-                &.disabled:focus {
-                    background: none;
-                    box-shadow: none;
-                }
+            &.disabled:hover {
```
mathisscott: Seems like these pseudo-selectors could be better organized in the SCSS.

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -67,162 +67,187 @@ $dropdown-white: clr-getColor(lightest);
...
 
-        &.open .dropdown-menu {
+        .dropdown-divider {
+            border-bottom: 1px solid clr-getColor(light);
+            margin: $clr_baselineRem_0_75 0;
+        }
+    }
+
+    //Directions
+    .btn-group-overflow,
```
mathisscott: I understand why we are doing it this way. But we may want to rethink this approach because it inherently doubles the weight of this block in the generated CSS.

What used to be one block of CSS is now two blocks.

src/clarity-angular/buttons/_buttons.clarity.scss
adityarb88: I discussed this with @youdz during a code pair and he mentioned that we should change our class names at one go when we decide to adopt the new naming conventions. So thats why I used to old convention that we have for buttons too. @youdz is that right?

mathisscott: I don't think that should apply to net-new components.

It doesn't apply to the wizard refactor, for example.

Nor does it apply to vertical tabs.

I'm fine leaving the _existing_ classes as they are. But what we are doing here is introducing NEW classes that are using an old, bad way of doing that.

I'm not okay with that.

adityarb88: I actually am still not sure about it. Button groups use the same class names as buttons. If we change button groups to use the new naming convention (`.btn-group--primary` instead of `.btn-group-primary`) won't that create more confusion? Users will have to remember 2 different classnames which do the same thing.

mathisscott: My main issue is that we are applying button classes to button groups which is requiring us to introduce unnecessary specificity and bloat our CSS as shown below.

`.btn-group--primary` is the direction we want to go, right? I see little difference between choosing `.btn-group-primary` or `.btn-group--primary` beyond another breaking change when we finalize our CSS naming conventions in code.

What I am against is `.btn-group.btn-whatever > .btn`. I think that's an unnecessary layer of specificity as compared to `.btn-group-whatever > .btn`.


Given how eager we have been to rip off the band aid with other components, this level of reluctance seems out of character.

adityarb88:
Closing this PR and opening a new one because there have been a lot of changes to the code.

PR #571, opened 2017-03-08
[UI] CSS Regression Test for Dropdown in a Header

adityarb88:
Added a CSS Regression test to test dropdowns in headers


gemini/tests/dropdown.js
```diff
@@ -63,4 +63,19 @@ gemini.suite('dropdown', (child) => {
             .capture('default');
     });
 
-});
\ No newline at end of file
+    gemini.suite('dropdown-header', (child) => {
+        child.setUrl('/dropdown/dropdown-header')
+            .before((actions, find) => {
```
hippee-lee: Do we need the second arg, `find` here?

adityarb88: @hippee-lee: Nope. we don't :-) removed it. thanks!

PR #608, opened 2017-03-16
[UI,NG] Adds Vertical Tab Component

### LINKED ISSUE #452: Vertical Tab component (new)
### Description
Vertical tabs offer a clean, easy way to navigate sections of content in an app while maximizing vertical space for content.

### Usage 

- When a longer list of tab links can’t be accommodated in a horizontal tab 
- When vertical space needs to be maximized for content
- When group headings are needed
- As the sole navigation in a single-tab UI or as the primary/secondary navigation in a multi-tab UI. In the latter case, layer the nested navigation by alternating vertical and horizontal orientation. For example, vertical tabs should be nested under horizontal tabs and vice versa. 

### Features
- One level with or without a group heading  
- Flat list, not collapsible/expandable
- Scrollable
- May be used on the left or right side of the content
- Individual tab links have hover and click/selected states. Group headings do not.

### Vertical tab 
example without headings

### Vertical tab with headings
vertical tab with headings

### Vertical Tabs vs other forms of vertical navigation
**Vertical tabs:**
- The simplest form of vertical navigation with only one level of navigation that is not collapsible.
- Scrolls with content

**Sidenav:** Features a disclosure arrow(optional) in the group heading to collapse and expand lists of links, making it a better option for longer lists with deep hierarchies.

**Tree:** The most complex form of vertical navigation with multiple levels of nested hierarchy that is collapsible and expandable.

### Typography
- Headings and tab links are left aligned. No indents.
- Section names should be short and concise. Wrapping text should be avoided if possible

Sketch file attached:
[Vertical Tabs Design_452.zip](https://github.com/vmware/clarity/files/761728/Vertical.Tabs.Design_452.zip)

hippee-lee:
- Can be used with reverse content/tab-navigation: `[clrReversed]="true"`
- Has `(clrVTabsNavChange)` event on active link change
- Has `(clrVTabsContentChange)` event on content change
- Can set a specific active nav link / content with `<clr-vtabs-nav-link [clrActivatedNavLink]="true">`
- Supports link navigation groups
- 18 unit tests
- closes #452


src/clarity-angular/clarity.module.ts
```diff
@@ -56,7 +56,7 @@ import {ClrResponsiveNavigationService} from "./nav/clrResponsiveNavigationServi
         TABS_DIRECTIVES,
         TREE_VIEW_DIRECTIVES,
         WIZARD_DIRECTIVES,
-        ICON_DIRECTIVES
+        ICON_DIRECTIVES,
```
mathisscott: Why a trailing comma here?

src/clarity-angular/nav/_nav.clarity.scss
```diff
@@ -6,7 +6,7 @@ $clr-nav-shadow: 0 -1px 0 $gray-light-midtone inset;
 $clr-active-nav-link-shadow: 0 -3px 0 $action-blues-dark-midtone inset;
 
 @include exports('nav.clarity') {
-    .nav {
+    .nav {  // is a ul this may not work the way I expected it to.
```
mathisscott: Comment is not clear. Please elaborate.

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
```diff
@@ -0,0 +1,84 @@
...
+        margin-top: 24px;
+
+        .nav-group-header {
+            @include clr-getTypePropertiesForDomElement(nav_group_header, (font-size, font-weight));
+            padding-left: 9px;
+        }
+
+        .nav-list {
+            list-style-type: none;
+            .nav-link {
```
mathisscott: Is it necessary to nest .nav-link and all its children inside .nav-list?

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
```diff
@@ -0,0 +1,84 @@
...
+                }
+
+                &:hover,
+                &:focus,
+                &:active,
+                &.active {
+                    text-decoration: none;
+                }
+
+                .btn.btn-link { // Overrides for vertical tabs specific style
```
mathisscott: Is it possible to reduce specificity here by...

1) Using `.btn-link` instead of `.btn.btn-link`?

2) Nesting it under just `.nav-link` instead of `.nav-list .nav-link`?

src/clarity-angular/tabs/vtabs-content.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            context.detectChanges();
+            let textHeader = context.clarityElement.querySelector(".vtabs-content");
+            expect(textHeader.textContent.trim()).toEqual("VTabs Test Content");
+        });
+    });
+}
+
+@Component({
+    template: `
+        <clr-vtabs-content>VTabs Test Content</clr-vtabs-content>
```
mathisscott: For the wizard tests, I have had to do a lot more work to verify projection. Either this isn't a complete test or @youdz is trolling me...

src/clarity-angular/tabs/vtabs.ts
```diff
@@ -0,0 +1,212 @@
...
+     * @param linkIndex - the index of the link calling this method
+     * @param id - the id string to set for the matching VTabContent item
+     */
+    mapContentId(linkIndex: number, id: string) {
+        this.vtabContents.forEach((content, contentIndex) => {
+            if (linkIndex === contentIndex) {
+                content.id = id;
+            }
+        });
+    }
```
mathisscott: I don't see the skip/unskip functionality that we have in horizontal tabs. Did I miss something?

src/clarity-angular/tabs/vtabs.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+}
+
+@Component({
+    template: `
+        <clr-vtabs (clrVTabsNavChange)="handleVTabsNavChange($event)"
+                   (clrVTabsNavChange)="handleVTabsContentChange($event)">
+            <clr-vtabs-nav-link>The Gunslinger</clr-vtabs-nav-link>
+            <clr-vtabs-nav-link>The Wastelands</clr-vtabs-nav-link>
+            <clr-vtabs-content>
```
mathisscott: Do we plan to write a test that validates what happens when a user generates content/nav-links with ng-fors?

What about if items are hidden with ng-if?

These are all things I'm having to test in the wizard, so I'm perhaps feeling a little picked on here.

src/clarity-angular/tabs/vtabs.ts
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * @description
+     * Method to update VTabContent items as follows:
+     *
+     * 1. Deactivate all VTabContent items not equal to the linkIndex arg
+     * 2. Activate the VTabContent iotem that has the same index as the linkIndex arg
+     *
+     * @param linkIndex - the index to activate in this.vtabsContent QueryList
+     */
+    activateTabContentAtIndex(linkIndex: number) {
```
mathisscott: Did we write any tests for this? I see very high-level tests. But not many tests for the API in isolation.

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2016 - 2017 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@import "../utils/variables.clarity";
```
adityarb88: We don't need these imports here now. Everything has been moved into `main.scss`

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
```diff
@@ -0,0 +1,84 @@
...
+
+                &:hover,
+                &:focus,
+                &:active,
+                &.active {
+                    text-decoration: none;
+                }
+
+                .btn.btn-link { // Overrides for vertical tabs specific style
+                    height: 24px;
```
adityarb88: We can use $clr-baselineRem-1 here, no?

src/clarity-angular/nav/_nav.clarity.scss
hippee-lee: Was a note to myself early on when I was figuring out the DOM structure. WIll remove.

src/clarity-angular/tabs/vtabs-content.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @description
+ * A child component of VTabs that:
+ * 1. Provides an id property set and used by the VTabs class to map VTabsNavLink items to the corresponding
+ *    VTabsContent items
+ * 2. Provides an activated property that triggers Angular to add/remove the content to the DOM via *ngIf
+ */
+@Component({
+    selector: "clr-vtabs-content",
+    template: `
+        <section class="vtabs-content" *ngIf="activated" role="tabpanel">
```
adityarb88: Can we not add the `vtabs-content` class and the `role` on the host element and replace this with `ng-container` with just the `*ngIf`? The angular component otherwise will be even more nested no? We now have 
clr-vtabs-content > .vtabs-content > ng-content whereas the static version has just the last 2.

src/clarity-angular/clarity.module.ts
hippee-lee: will remove.

src/clarity-angular/tabs/vtabs.ts
mathisscott: I have wizard on the brain too much. The _old_ wizard had added skip/unskip tab functionality. It's not part of tabs.

Feel free to ignore this comment. My bad.

src/clarity-angular/tabs/vtabs-nav-group.ts
```diff
@@ -0,0 +1,25 @@
...
+ * An optional label can be passed in to give the group a title.
+ * clr-vtabs-nav-links are projected to the VTabsNavLink class.
+ *
+ * Style is set in the template, not dynamically.
+ */
+@Component({
+    selector: "clr-vtabs-nav-group",
+    template: `
+        <div class="nav-group">
+            <ng-content select="label"></ng-content>
```
adityarb88: I think @youdz mentioned to me earlier that we are not querying by class or by tag name. Can you please check with him once? Also, what exactly is `label`? Is it the nav-group header? If yes, we should style it using a class no?

src/clarity-angular/tabs/vtabs-nav-link.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @description
+ * A child component of VTabs that:
+ *  1. Creates the navigation links
+ *  2. Exposes an input property that allows binding to preselect a specific tab
+ *  3. Has functionality to allow an instance to become active and emit the change up to VTabs
+ */
+@Component({
+    selector: "clr-vtabs-nav-link",
+    template: `
+        <li (click)="activateLink()" >
```
adityarb88: `click` on the `li` might not work like we discussed with @youdz

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
hippee-lee: Correct, it does not need to be nested. Will correct.

src/clarity-angular/tabs/vtabs.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * @property linkChange
+     *
+     * @description
+     * EventEmitter for the tab link change event.
+     *
+     * @type {EventEmitter<VTabsNavLink>}
+     */
+    @Output("clrVTabsNavChange") linkChange: EventEmitter<VTabsNavLink> = new EventEmitter<VTabsNavLink>(false);
```
adityarb88: @youdz are we still passing `false` when we initializing the `EventEmitter`?

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
hippee-lee: Correct, this does not need to be nested and can be kept to `.btn-link` only. Will correct.

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
hippee-lee: Yes. Will update it.

src/clarity-angular/tabs/vtabs-content.ts
mathisscott: So this may be something I take issue with....

In the DOM example above, we are using `aria-hidden="true"` to hide `.vtabs-content`. But here we are using `*ngIf`.

First, I have an issue with presenting two different ways to do the same thing.

Second, I have questions about how using `*ngIf` in this manner affects accessibility.

Is there a specific reason we are using `*ngIf="activated"` instead of `[attr.aria-hidden]="activated"`?

hippee-lee: For `First`: I used `aria-hidden: true` b/c it was already built into the css and easy to re-use for the static example. My thinking was not to have a special style for vtabs-content in the static [UI] case. And, I planned on using *ngIf to control which content container was visible for the child component.

For `Second`: `when aria-hidden is used on an element, it removes all relevance to that element.` but you still have to hide it from the user. *ngIf removes the element from the DOM. So, I'm thinking that those are equivalent wrt accessibility. 

My preference would be to keep the *ngIf in the component and remove aria-hidden from the static example since a clarity-ui user will be implementing the hide/show functionality for the active content container.

hippee-lee: You are correct. I will change the template structure for the component.

mathisscott: As regards `aria-hidden` and accessibility, you should discuss this with @youdz and @adityarb88 because this is a pattern we have used literally _everywhere_. So if the plan is to change it here, you may as well change it everywhere else.

Additionally, we don't use `aria-hidden` to remove content from screenreaders. But ngIf clearly does.

Lastly, if we are only offering hide/show on the angular component, don't you think that's really gimping the HTML/CSS implementation.

You are asking users to come up with their own methods of hiding/showing the active tab content. That's a little much, IMO.

I'm not okay with that.

src/clarity-angular/tabs/vtabs-nav-group.ts
mathisscott: The new wizard queries some elements by class, tag-name, or even attribute. And that was @youdz idea... so there is clearly a disconnect.

The one thing I would say is that the classnames we query by are generally internal and not used by the CSS for styling. That's what I would try to follow (such that if we decide to change the CSS naming convention we won't break the component).

The only thing I would ping here is that label is a bit too general to the implementation. I actually don't have a huge problem with this though.

src/clarity-angular/tabs/vtabs.ts
hippee-lee: My approach to testing was to look at the functionality of each class and write a tests that made sure the output of a function behaved as expected.

For this `activateTabContextAtIndex` I use `it("activates content when link is clicked", function () { ...}`

The thinking behind this is: a specific link is clicked and a specific element of content should be activated. What else should I consider for this?

src/clarity-angular/tabs/vtabs-nav-group.ts
adityarb88: @hippee-lee, @mathisscott: Oh I didn't know that the wizard was doing the same thing. I might be wrong here then. But I remember that dropdowns had an issue because we used to query by class `.dropdown-item` and `.dropdown-toggle`. We introduced a breaking change to replace them by directives because of the number of issues it was causing in the demo app. But that definitely might a case just with dropdowns. 
We should check with @youdz on whats the right approach to handle such cases. But irrespective of that querying by `label` i think can be improved to query by `.nav-group-header` class just to be more specific and I say that because if someone use a `span` instead of a `label` this might break.

adityarb88: @hippee-lee: Actually you can ignore my comment about the dropdowns. I looked and the issue was dropdown requiring an attribute selector was different. You can definitely use css selectors like @mathisscott mentioned.

src/clarity-angular/tabs/vtabs.ts
mathisscott: I generally make very small unit tests that test a specific path. I would have individual tests that validated the `contentIndex === linkIndex` as well as `content.activated = false` fallthrough.

@youdz also directed me to divide my tests into TypescriptAPI, Template API, and View/Behavior groupings.

What you've written is a high-level test that goes through both states that would fit in the Template API grouping. But you would also want to verify any CSS classes are added to the component as expected in a test in the View/Behavior grouping.

An assumption that I see here is that vtabContents is always populated. I would hope to see more resiliency in handling an edge case if vtabContents was an empty array as well as if it came through as undefined or null.

These are questions I've been pressed to answer in the wizard. Do we throw an error if we find out vtabContents has been cleared out? Or something in the architecture messes up?

Your tests are good. And bonus points for using @youdz text context helpers. My issue is more around the rigor being expected of the tests I'm having to write is not reflected here.

So I'm wondering (for my own self) why there is a delta?

Your one test (were it in the wizard) would be broken up into 3-4 tests across 2-3 groupings. Yet I don't see any groupings here and the tests are much broader/birds-eye than I've been told to write.

So I don't know what to tell you other than this is not how I have been told to write tests for components. I don't know why I've been told to do them differently.

src/clarity-angular/tabs/_vertical-tabs.clarity.scss
```diff
@@ -0,0 +1,84 @@
...
+        flex: 0 180px;
+        padding-right: 12px;
+    }
+
+    &-content {
+        flex: 1;
+    }
+
+    .nav-group {
+        $navlink-padding: 4px;
```
Shijir: Users may want to customize these values so we might need to set them as default values.

src/clarity-angular/tabs/vtabs.ts
```diff
@@ -0,0 +1,212 @@
...
+     * 4. Sets the first link and first content as active, if needed.
+     *
+     * No params, called as part of the angular lifecycle
+     */
+    ngAfterContentInit() {
+        // Initialize tabs and content.
+
+        this.vtabLinks.forEach((link, index) => {
+            link.id = "clr-vtab-link-" + nbVTabLinks.toString();
+            nbVTabLinks++;
```
Shijir: nit: the following could be more concise:
```
 "clr-vtab-link-" + nbVTabLinks++;
```

src/clarity-angular/tabs/vtabs-nav-link.ts
youdz: Oh, yes, that's a no-go. Click listener goes on the button, definitely.
And once that's done, you don't want a list item anymore, just the host with the correct classes, the button, and `role="listitem"` on the host for accessibility (with the parent having `role="list"`). I don't believe we do this in horizontal tabs, by the way, we should go back to it at some point.

src/clarity-angular/tabs/vtabs.ts
youdz: It's safer, we explicitly state we want them synchronous. We've been burnt once when the Angular team made them asynchronous by default without announcing it.

src/clarity-angular/tabs/vtabs-nav-link.ts
```diff
@@ -0,0 +1,79 @@
...
+     * 1. Check if instance is already active
+     * 2. Set the instance of VTabsNavLink to active
+     * 2. Emit the change to subscribers (VTabs) of activatedChange
+     */
+    activateLink() {
+        if (!this.activated) {
+            this.activated = true;
+            this.activatedChange.emit(this);
+        }
+    }
```
youdz: You never emit `activatedChange` when a tab link de-activates. So your 2-way binding simply doesn't work.

This goes with Scott's remarks complaining that the tests are not split and exhaustive enough. If you had a simple two-way binding test on this, you would notice it's not working immediately.

jaffoneh:
More work is required here. Closing for now and will reopen as we start working on this.

PR #632, opened 2017-03-22
[NG] Datagrid Selection Clear - DO NOT MERGE

adityarb88:
Before:
User selection was cleared whenever any change to items, filtering or page changes occurred.

After:
1. The selection is cleared when a new filter is applied on the datagrid
2. The selection is preserved when a page is changed
3. When the user uses `clrDgItems` or the smart datagrid, we make use of the trackBy function if provided to preserve selection from the updated items.

Issues:
We still have a `Expression has changed` issue on the demo that @youdz and I have to resolve.

UX Feedback:
@yen and @red have to discuss complex cases when the selection might have to be preserved after a filter has been applied. This PR is blocked until we get more info on which pattern to use.

Tested on Chrome

### DIFF
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -269,6 +269,28 @@
             }
         }
 
+        .datagrid-footer-select.checkbox {
+            line-height: inherit;
+            flex: 1 1 100%;
+
+            label {
+                color: unset;
+                cursor: default;
+                opacity: 1;
+            }
+
+            input[type="checkbox"] + label {
+                &::before, &::after {
+                    top: 8px;
+                }
+
+                &::after {
+                    border-left-color: $clr-white;
+                    border-bottom-color: $clr-white;
+                }
+            }
+        }
+
         .datagrid-row-actions {
             flex: 0 0 $clr-datagrid-select-column-size;
             // IE10 needs this
--- a/src/clarity-angular/datagrid/datagrid-items.ts
+++ b/src/clarity-angular/datagrid/datagrid-items.ts
@@ -25,15 +25,15 @@ export class DatagridItems implements OnChanges, DoCheck {
         if ("rawItems" in changes) {
             const currentItems = changes["rawItems"].currentValue;
             if (!this._differ && currentItems) {
-                this._differ = this._differs.find(currentItems).create(this._changeDetector, this.trackBy);
+                this._differ = this._differs.find(currentItems).create(this._changeDetector, this._items.trackBy);
             }
         }
     }
 
-    /**
-     * Tracking function to identify objects. Default is reference equality.
-     */
-    @Input("clrDgItemsTrackBy") public trackBy: TrackByFn = (index: number, item: any) => item;
+    @Input("clrDgItemsTrackBy")
+    set trackBy(value: TrackByFn) {
+        this._items.trackBy = value;
+    }
 
     ngDoCheck() {
         if (this._differ) {
--- a/src/clarity-angular/datagrid/providers/items.ts
+++ b/src/clarity-angular/datagrid/providers/items.ts
@@ -3,7 +3,7 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Injectable} from "@angular/core";
+import {Injectable, TrackByFn} from "@angular/core";
 import {Subject} from "rxjs/Subject";
 import {Subscription} from "rxjs/Subscription";
 import {Observable} from "rxjs/Observable";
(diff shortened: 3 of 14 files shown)


src/clarity-angular/datagrid/providers/selection.ts
```diff
@@ -22,16 +23,65 @@ export enum SelectionType {
 export class Selection {
     public id: string;
 
-    constructor(private _items: Items) {
+    constructor(private _items: Items, private _filters: FiltersProvider) {
         this.id = "clr-dg-selection" + (nbSelection++);
-        this._itemsSub = _items.change.subscribe(() => {
-            if (!this._selectable || this.debounce) {
+        /* TODO */
```
youdz: You can probably remove the TODO now. 😛

src/clarity-angular/datagrid/providers/selection.ts
```diff
@@ -22,16 +23,65 @@ export enum SelectionType {
...
+         */
+
+        /*
+            //TODO: Remove this
+            this._itemsSub = _items.change.subscribe(() => {
+                if (!this._selectable || this.debounce) {
+                    return;
+                }
+            });
+        */
```
youdz: Clean up. 😉

src/clarity-angular/datagrid/providers/selection.ts
```diff
@@ -53,17 +103,21 @@ export class Selection {
     /**
      * Ignore items changes in the same change detection cycle.
      */
-    private debounce = false;
+    private debounce: boolean = false;
```
youdz: We're not using this anymore, are we? If that's the case, we want to remove it altogether.

src/clarity-angular/datagrid/providers/selection.spec.ts
```diff
@@ -135,5 +141,95 @@ export default function(): void {
...
+
+        it("clears selection when a filter is added", function() {
+            this.selectionInstance.selectionType = SelectionType.Multi;
+            this.selectionInstance.current = [4, 2];
+
+            let evenFilter: EvenFilter = new EvenFilter();
+
+            this.filterInstance.change.subscribe(() => {
+                expect(this.selectionInstance.current.length).toBe(0);
+            });
```
youdz: I don't like this pattern, because if someone removed the filter listener in the Selection provider, this test would _always_ pass, even if it didn't clear anymore.

I think you just want the `expect()` to go right after the `this.filterInstance.add(evenFilter);`, you don't need to subscribe to `.change` in the test.

src/clarity-angular/datagrid/providers/selection.spec.ts
```diff
@@ -135,5 +141,95 @@ export default function(): void {
...
+
+        it("keeps only the remaining selection when the items are updated", function() {
+            this.selectionInstance.selectionType = SelectionType.Multi;
+            this.selectionInstance.current = [4, 2];
+
+            let tempSub = this.itemsInstance.allChanges.subscribe(() => {
+                expect(this.selectionInstance.current.length).toBe(1);
+                expect(this.selectionInstance.current).toEqual([2]);
+                tempSub.unsubscribe();
+            });
```
youdz: Same here, and all the tests under it.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -266,6 +266,28 @@
...
+
+            label {
+                color: unset;
+                cursor: default;
+                opacity: 1;
+            }
+
+            input[type="checkbox"] + label {
+                &::before, &::after {
+                    top: 8px;
```
adityarb88: I need to refactor this in checkbox.clarity.scss. I should have probably added a TODO but for now the hardcoded pixels is fine.

src/clarity-angular/datagrid/datagrid-items.ts
```diff
@@ -12,7 +12,16 @@ import {Items} from "./providers/items";
...
+    private _rawItems: any[];
+
+    public get rawItems(): any[] {
+        return this._rawItems;
+    }
+
+    @Input("clrDgItemsOf")
+    public set rawItems(value: any[]) {
+        this._rawItems = value;
+    }
```
adityarb88: Yep. I had some logic in there but that was not needed later on. Thanks for catching this.

src/clarity-angular/datagrid/providers/selection.ts
adityarb88: No! :trollface:

src/clarity-angular/datagrid/providers/selection.ts
adityarb88: You do use it for single selection. I just added the `boolean` type to it.

src/clarity-angular/datagrid/providers/selection.spec.ts
adityarb88: Fixed

src/clarity-angular/datagrid/providers/items.ts
```diff
@@ -22,6 +22,11 @@ export class Items {
     public loading = false;
 
     /**
+     * Tracking function to identify objects. Default is reference equality.
+     */
+    public trackBy: TrackByFn = (index: number, item: any) => item;
```
adityarb88: I am adding a TODO for this for now. Will look at it later. Is that okay?

luis-cortina:
Is there any way to preserve the selection even when filtering happens?

PR #636, opened 2017-03-22
[NG] Allow reverse order in datagrid 2-way binding sort attribute

### LINKED ISSUE #586: [NG] Datagrid: allow reversing the sort order of a column dynamically
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

I should be able to set the initial sort order of the datagrid, even if it is a "reverse" order (the one you obtain by clicking a second time on the same column).
Something like a `[(clrDgSortedReverse)]` boolean two-way binding on the column would work. 
_Be careful of potential race conditions between `[(clrDgSorted)]` and `[(clrDgSortedReverse)]`._

### Actual behavior

#509 introduced a way to set the initial sort order of the datagrid, but it doesn't offer any way to make it reverse (or descending).

clementcur:
Changing current datagrid-column implementation to take into account the reverse sorting order. 

Instead of introducing a new 2-way binding as advised in the original issue #586, I replaced the existing  boolean `[(clrDgSorted)]` by a number `[(clrDgSortOrder)]` one. I chose not to go with a `[(clrDgSortedReverse)]` 2-way binding for two main reasons:
 - it would have complexify the usage of the datagrid component, by having to maintain 2 boolean flags for the same concept
 - race conditions would have occurred when the value of each of those 2 flags changes: they are dependent on each other and there's no way to predict the order in which they are updated

Details about the proposed implementation:
- Introduced an exported const Enum `SortOrder` that defines the 3 possible
sorting states for a datagrid column: Unsorted, Asc and Desc.
- Removed `[(clrDgSorted)]`:boolean 2-way binding
- Introduced `[(clrDgSortOrder)]`: SortOrder 2-way binding:
  - replaced the existing logic around the sorted state (2 states: ON or OFF)
  with a tri-state sorting logic (following the 3 `SortOrder` enum values)
  - made sure all existing behaviour is preserved:
    - the `datagrid-column` reacts to the `Sort` provider change events, i.e. when
    sort is cleared or toggled
    - modified the public `sort()` method to allow forcing the reverse sorting
- Modified the `Sort` provider toggle method to allow forcing the reverse sorting. It
is an optional parameter, so it does not break backward compatibility within
clarity-angular implementation
- modified and added some new test cases for the `datagrid-column` component and the
`Sort` provider

I left the original boolean `clrDgSorted` 2-way binding in clr-dg-column to allow
backward compatibility for the component consumption. I added some deprecation
annotations around all code that needs to be removed in the next major release.

/!\ Using both 2-way bindings at the same time will have unpredicted results, so
the using code needs to either use `clrDgSorted` or `clrDgSortOrder`. This needs
be explicitly mentioned in the doc.

This closes #586

Signed-off-by: Clement Cureau <clement.cureau@gmail.com>

### DIFF
--- a/src/clarity-angular/datagrid/datagrid-column.spec.ts
+++ b/src/clarity-angular/datagrid/datagrid-column.spec.ts
@@ -3,20 +3,21 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component, ViewChild} from "@angular/core";
-import {Subject} from "rxjs/Subject";
-import {TestContext} from "./helpers.spec";
-import {DatagridColumn} from "./datagrid-column";
-import {Sort} from "./providers/sort";
-import {Comparator} from "./interfaces/comparator";
-import {FiltersProvider} from "./providers/filters";
-import {Filter} from "./interfaces/filter";
-import {DatagridPropertyComparator} from "./built-in/comparators/datagrid-property-comparator";
-import {StringFilter} from "./interfaces/string-filter";
-import {TestBed} from "@angular/core/testing";
-import {DatagridStringFilter} from "./built-in/filters/datagrid-string-filter";
-import {DatagridRenderOrganizer} from "./render/render-organizer";
-import {DomAdapter} from "./render/dom-adapter";
+import { Component, ViewChild } from "@angular/core";
+import { Subject } from "rxjs/Subject";
+import { TestContext } from "./helpers.spec";
+import { DatagridColumn } from "./datagrid-column";
+import { Sort } from "./providers/sort";
+import { Comparator } from "./interfaces/comparator";
+import { FiltersProvider } from "./providers/filters";
+import { Filter } from "./interfaces/filter";
+import { DatagridPropertyComparator } from "./built-in/comparators/datagrid-property-comparator";
+import { StringFilter } from "./interfaces/string-filter";
+import { TestBed } from "@angular/core/testing";
+import { DatagridStringFilter } from "./built-in/filters/datagrid-string-filter";
+import { DatagridRenderOrganizer } from "./render/render-organizer";
+import { DomAdapter } from "./render/dom-adapter";
+import { SortOrder } from "./interfaces/sort-order";
 
 const PROVIDERS_NEEDED = [Sort, FiltersProvider, DatagridRenderOrganizer, DomAdapter];
 
@@ -67,6 +68,26 @@ export default function (): void {
                 expect(component.sorted).toBe(true);
             });
 
+            it("sorts according to the optional input parameter", function () {
+                component.sortBy = comparator;
+                expect(component.sortOrder).toBe(SortOrder.Unsorted);
+                component.sort(true);
+                expect(component.sortOrder).toBe(SortOrder.Desc);
+                component.sort(true);
+                expect(component.sortOrder).toBe(SortOrder.Desc);
+                component.sort(false);
+                expect(component.sortOrder).toBe(SortOrder.Asc);
+            });
+
+            it("knows the column current sorting order", function () {
+                component.sortBy = comparator;
+                expect(component.sortOrder).toBe(SortOrder.Unsorted);
+                component.sort();
+                expect(component.sortOrder).toBe(SortOrder.Asc);
+                component.sort();
+                expect(component.sortOrder).toBe(SortOrder.Desc);
+            });
+
             it("knows if the column is currently sorted in ascending order", function () {
                 component.sortBy = comparator;
                 expect(component.asc).toBe(false);
@@ -93,8 +114,8 @@ export default function (): void {
             });
         });
 
-        describe("Template API", function() {
-            it("receives an input for the comparator", function() {
+        describe("Template API", function () {
+            it("receives an input for the comparator", function () {
                 this.context = this.create(DatagridColumn, SimpleTest, PROVIDERS_NEEDED);
                 this.comparator = new TestComparator();
                 this.context.testComponent.comparator = this.comparator;
@@ -102,7 +123,7 @@ export default function (): void {
                 expect(this.context.clarityDirective.sortBy).toBe(this.comparator);
             });
 
-            it("receives an input for the property name", function() {
+            it("receives an input for the property name", function () {
                 this.context = this.create(DatagridColumn, SimpleTest, PROVIDERS_NEEDED);
                 this.context.testComponent.field = "test";
                 this.context.detectChanges();
@@ -118,7 +139,7 @@ export default function (): void {
             });
 
             it("offers two-way binding on the sorted state", function () {
-                this.context = this.create(DatagridColumn, SimpleTest, PROVIDERS_NEEDED);
+                this.context = this.create(DatagridColumn, SimpleDeprecatedTest, PROVIDERS_NEEDED);
                 this.comparator = new TestComparator();
                 this.context.testComponent.comparator = this.comparator;
                 this.context.testComponent.sorted = true;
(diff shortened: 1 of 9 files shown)


clementcur:
Sorry about all the formating changes, I forgot to disable the autosave option of my workspace beautify plugin... Let me know if you prefer me to amend my commit without those.

src/clarity-angular/datagrid/providers/sort.ts
```diff
@@ -48,15 +48,24 @@ export class Sort {
...
+     * @param {boolean} [forceReverse] `true` to force sorting descendingly
+     *
+     * @memberof Sort
      */
-    public toggle(sortBy: Comparator<any>) {
+    public toggle(sortBy: Comparator<any>, forceReverse?: boolean) {
         // We modify private properties directly, to batch the change event
         if (this.comparator === sortBy) {
-            this._reverse = !this._reverse;
+            this._reverse = forceReverse !== undefined
```
youdz: Please use `typeof forceReverse !== "undefined"`, it's safer on older browsers that let you reassign `undefined`. Not sure if we support any of these, but it doesn't cost much.

src/clarity-angular/datagrid/providers/sort.ts
```diff
@@ -48,15 +48,24 @@ export class Sort {
...
+                ? forceReverse || !this._reverse
+                : !this._reverse;
         } else {
             this._comparator = sortBy;
-            this._reverse = false;
+            this._reverse = forceReverse !== undefined ? forceReverse : false;
```
youdz: Same here.

src/clarity-angular/datagrid/datagrid-column.spec.ts
```diff
@@ -117,16 +129,19 @@ export default function (): void {
...
                 this.context.detectChanges();
-                expect(this.context.clarityDirective.sorted).toBe(true); //dg col instance
+                expect(this.context.clarityDirective.sortOrder).toBe(SortOrder.Desc); //dg col instance
                 this.context.getClarityProvider(Sort).clear();
                 this.context.detectChanges();
-                expect(this.context.testComponent.sorted).toBe(false);
+                expect(this.context.testComponent.sortOrder).toBe(SortOrder.Unsorted);
+                this.context.testComponent.sortOrder = SortOrder.Asc;
+                this.context.detectChanges();
+                expect(this.context.clarityDirective.sortOrder).toBe(SortOrder.Asc); //dg col instance
```
clementcur: sure, fixed.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -74,49 +75,62 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
...
+    @Input("clrDgSortOrder")
+    public set sortOrder(value: SortOrder) {
+        // only if the incoming order is different from the current one
+        if (this._sortOrder === value) {
+            return;
+        }
+
+        switch (value) {
+            default: case SortOrder.Unsorted:
+                this._sort.clear();
```
clementcur: done

src/clarity-angular/datagrid/interfaces/sort-order.ts
```diff
@@ -0,0 +1,18 @@
...
+ * i.e. each value needs to be treated as a `number`, starting at index 0.
+ *
+ * @export
+ * @enum {number}
+ */
+export const enum SortOrder {
+    Unsorted,
+    Asc,
+    Desc
+}
```
clementcur: makes sense, done.

src/clarity-angular/datagrid/providers/sort.ts
clementcur: ok, done.

src/clarity-angular/datagrid/providers/sort.ts
clementcur: done too.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -43,10 +44,16 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
         super(filters);
         this._sortSubscription = _sort.change.subscribe(sort => {
             // We're only listening to make sure we emit an event when the column goes from sorted to unsorted
```
mathisscott: Is this comment still accurate? It doesn't seem like that is or should be the case. Shouldn't `sortOrderChange` emit when I reverse the sort order too?

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -87,36 +100,92 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
...
+    public sort(reverse?: boolean) {
         if (!this.sortable) {
             return;
         }
+
+        this._sort.toggle(this.sortBy, reverse);
+
+        // setting the private variable to not retrigger the setter logic
+        this._sortOrder = this._sort.reverse ? SortOrder.Desc : SortOrder.Asc;
+        this.sortOrderChange.emit(this._sortOrder);
```
mathisscott: This is more of a stylistic comment than anything and nothing I would go out of my way to correct if you don't want to.

But it appears we are emitting `sortOrderChange` in a few places in this code. My preference is to focus event emission in one location if possible.

I mean, given the subscription on `_sort.change` wouldn't this component fire multiple emissions for one sort change?

src/clarity-angular/datagrid/interfaces/sort-order.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+/**
+ * Enumeration representing the sorting order of a datagrid column. It is a constant Enum,
+ * i.e. each value needs to be treated as a `number`, starting at index 0.
+ *
+ * @export
+ * @enum {number}
+ */
+export const enum SortOrder {
```
mathisscott: Bonus points for the enum. Well done.

src/clarity-angular/datagrid/providers/sort.spec.ts
```diff
@@ -3,37 +3,39 @@
...
-        beforeEach(function() {
+export default function (): void {
+    describe("Sort provider", function () {
+        beforeEach(function () {
             this.sortInstance = new Sort();
             this.comparator = new TestComparator();
         });
 
-        it("compares according to the current comparator", function() {
+        it("compares according to the current comparator", function () {
```
mathisscott: Do we have any tests for the default setting or if a comparator comes through as null or undefined?

What happens in those cases?

src/clarity-angular/datagrid/datagrid-column.ts
youdz: No, we're emitting it in the sort method itself. This is only here for the case where another column becomes sorted, which automatically de-activates this one.

mathisscott: So do we need to update this comment? It doesn't seem to make any sense....

src/clarity-angular/datagrid/datagrid-column.ts
youdz: As I mentioned in response to your previous comment, we have 2 emission which correspond to 2 sources of change: this one handles the case where the column itself is the source of change (becomes sorted, or toggles asc/desc), whereas the one in the listener handles the case where another column becomes sorted, in which case this one becomes unsorted.

The fact that the change comes from such different sources means trying to handle them both in the same place leads to lifecycle errors. I'm fine having these 2 separate emissions, as long as it's just 2 of them.

src/clarity-angular/datagrid/datagrid-column.ts
youdz: It makes complete sense. It still describes exactly what the `if` on the very next line does:
```typescript
if (this.sortOrder !== SortOrder.Unsorted && sort.comparator !== this.sortBy) {
```
We only emit when the column is currently sorted and the provider just switched to a comparator that's not ours. That means another column became sorted, so we're unsorted now.

clementcur: Exactly, the sort action can also be performed outside this class, i.e. in the Sort provider via the public method.

jeeyun:
Hi @clementcur could you rebase and squash your commits on this PR so we can schedule it for a release?

clementcur:
@jeeyun I just rebased, I'll keep an eye on the CI job!

jeeyun:
Thank you for your contribution @clementcur! Your PR is merged to master, and is scheduled to go out with this week's release!

Issue #631, opened 2017-03-22
Paging to last page of large dataset --> grid disappears !


kodemasta:
**Select one ...**  (check one with "x")
```
[X] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
paging should work for large number of rows independent of total row count and pageSize.

### Actual behavior
For some paging settings --> selecting the last page causes the grid to disappear including header and footer ! Expected last page of data is not accessible !

### Reproduction of behavior
Create a grid with 1200 Users (rows)
Set pageSize to 473. Then select 3rd page and grid disappears.
We were able to repro with various page and data sizes as well.

(some sort of floating point paging calculation error ?)

plunkr: https://plnkr.co/edit/zZZOW0utKuNZS572CaGI?p=preview

### Environment details

* **Angular version:** 2.4.1

* **Clarity version:** 0.8.11 

* **OS and version:** MacOs 10.11.2

* **Browser:** Chrome, Firefox,...

youdz:
Could you please post a plunker reproducing the error? Thank you.

bobsheehan1:
OK Added plunker to bug description: https://plnkr.co/edit/zZZOW0utKuNZS572CaGI?p=preview

Jinnie:
@bobsheehan1 
The datagrid on the last page gets "squeezed" up to the top by unpredicted sub-component behavior, which probably conflicts with the recent datagrid content scroll feature.

Here's a plunker that simplifies and illustrates the case:
https://plnkr.co/edit/CdJH5W7PUE4jKPkdKLYp?p=preview
The bottom half is growing in size (min-height) if you play with the page size.

@youdz, it seems someone had a premonition, writing this comment :rofl: 
> I hate doing this, with these 36px...

youdz:
Thank you for the plunker. 👍 

As Jinnie mentioned, this is a general issue with the placeholder since the switch to flexbox. We have 2 or 3 issues related to this, hopefully we can fix them all at once.

Issue #583, opened 2017-03-10
Datagrid style looks broken during filtering 


Shijir:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Actual behavior

Initial state:

<img width="1147" alt="screenshot 2017-03-10 02 26 07" src="https://cloud.githubusercontent.com/assets/3958008/23791899/a9fe047c-0539-11e7-8c26-8f8c9c128326.png">

it looks like some parts of datagrid get cut off once a user starts typing in something inside the filtering input:

<img width="1153" alt="screenshot 2017-03-10 02 26 23" src="https://cloud.githubusercontent.com/assets/3958008/23791895/a5516d1a-0539-11e7-95d8-ac8c4ec16dd8.png">

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

mathisscott:
@adityarb88 : Could you verify whether this is still an issue or not after your PR goes in?

adityarb88:
@mathisscott: It is. I linked it. But I am doubtful if my PR will go in tomorrow :-(

mathisscott:
No worries. If you feel it can make it into 0.9.2, would you mind moving it to that milestone?

PR #637, opened 2017-03-22
[NG] Datagrid Filter + Placeholder Fix

adityarb88:
Fixes: #631, #583 

1. Added a minimum height (equal to the height of the header + the height of the filter popover) to `.datagrid-table-wrapper` to make space for the filter popover to appear without clipping.

**Before:**
1. When no rows are displayed
image

2. When only one row is displayed
image

3. When two rows are displayed
image

**After:**
1. When no rows are displayed
image

2. When only 1 row is displayed
image

3. When 2 rows are displayed
image

Custom placeholders work as expected
image

Custom placeholders when filtering is present:
image

Tested on Chrome

### DIFF
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -9,6 +9,8 @@
     $clr-datagrid-icon-size: rem(1);
     $clr-datagrid-action-arrow-size: $clr_baselineRem_0_25;
     $clr-datagrid-select-column-size: 2*$clr-table-cellpadding + $clr-table-lineheight;
+    $clr-datagrid-filter-padding: $clr_baselineRem_0_75;
+    $clr-datagrid-filter-top-margin: $clr_baselineRem_0_2;
 
     .datagrid-container {
         display: flex;
@@ -39,6 +41,11 @@
         min-height: 0;
         // IE10 (maybe later versions too) needs this
         overflow: hidden;
+
+        &.has-filter {
+            // space for (filter padding + 2 * 24 (close button + input field) + margin top + Header height + 1px for vertical rhythm)
+            min-height: $clr-datagrid-filter-padding * 2 + $clr_baselineRem_2 + $clr-datagrid-filter-top-margin + $clr_baselineRem_1_5 + 1;
+        }
     }
     .datagrid-body {
         flex: 1 1 auto;
@@ -223,9 +230,9 @@
                 position: absolute;
                 top: 100%;
                 right: 0;
-                margin-top: $clr_baselineRem_0_2;
+                margin-top: $clr-datagrid-filter-top-margin;
                 background: $clr-white;
-                padding: $clr_baselineRem_0_75;
+                padding: $clr-datagrid-filter-padding;
                 border: 1px solid $gray-light-midtone;
                 box-shadow: 0 1px 3px rgba($gray-dark, 0.25);
                 border-radius: $clr-default-borderradius;
@@ -234,6 +241,7 @@
 
                 .datagrid-filter-close-wrapper {
                     text-align: right;
+                    height: $clr_baselineRem_1;
 
                     .close {
                         float: none;
--- a/src/clarity-angular/datagrid/datagrid.html
+++ b/src/clarity-angular/datagrid/datagrid.html
@@ -6,7 +6,7 @@
 
 <ng-content select="clr-dg-action-bar"></ng-content>
 <div class="datagrid">
-    <div clrDgTableWrapper class="datagrid-table-wrapper">
+    <div clrDgTableWrapper class="datagrid-table-wrapper" [class.has-filter]="hasFilters()">
         <div clrDgHead class="datagrid-head">
             <div class="datagrid-row">
                 <!-- header for datagrid where you can select multiple rows -->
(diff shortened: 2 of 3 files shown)


src/clarity-angular/datagrid/datagrid.html
```diff
@@ -36,7 +36,7 @@
 
     <!-- Custom placeholder overrides the default empty one -->
     <ng-content select="clr-dg-placeholder"></ng-content>
-    <clr-dg-placeholder *ngIf="!placeholder"></clr-dg-placeholder>
+    <clr-dg-placeholder *ngIf="!placeholder && items.displayed.length == 0"></clr-dg-placeholder>
```
Jinnie: I have a feeling that this change may beneficially effect this issue: https://github.com/vmware/clarity/issues/631
Best regards 😃

src/clarity-angular/datagrid/datagrid.ts
```diff
@@ -116,6 +116,10 @@ export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
         this.refresh.emit(state);
     }
 
+    public hasFilters(): boolean {
+        return this.filters.getActiveFilters().length > 0 ? true : false;
```
mathisscott: Is there any possibility that `this.filters` could come through as undefined or null?

adityarb88: I dont think so @mathisscott. I will confirm with @youdz once. I think the filters might be empty but not null.

youdz: Nope, it's a provider declared by the Datagrid component itself.

src/clarity-angular/datagrid/datagrid.html
adityarb88: @Jinnie: Yes. I tested that case too while working on this fix. I forgot to tag all the issues which submitting the PR

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -39,6 +41,11 @@
         min-height: 0;
         // IE10 (maybe later versions too) needs this
         overflow: hidden;
+
+        &.has-filter {
+            // space for (filter padding + 2 * 24 (close button + input field) + margin top + Header height + 1px for vertical rhythm)
+            min-height: $clr-datagrid-filter-padding * 2 + $clr_baselineRem_2 + $clr-datagrid-filter-top-margin + $clr_baselineRem_1_5 + 1;
```
youdz: This only works for our string filters. It doesn't work for any kind of custom filter, see our colorpicker filter for instance.

src/clarity-angular/datagrid/datagrid.ts
youdz: Active filters are just the ones which are currently filtering something. For instance, a string filter with an empty input (the user hasn't typed anything yet) isn't active, so this will return `false` and you'll get a jacked up display even though the user can open the filter.

src/clarity-angular/datagrid/datagrid.html
youdz: This is not the behavior we want. It makes the last page jump when there are fewer items on it, which is the opposite of what we want.

Jinnie: What do you mean by "jump"?
Sorry, I've been playing with this problem for a while, want to see if I can help, but I don't understand your point.

youdz: There is actually a bigger issue behind this, that I didn't point out explicitly: the purpose of this `*ngIf` is to always have a default placeholder when the user doesn't project one. So any sizing logic of the placeholder should never go here, but in the placeholder component itself. With this fix, the default placeholder and a custom one will behave completely different.

Now, regarding the "jump": one of the placeholder's purpose is to ensure the datagrid keeps the same height when moving to the last page. For instance, if you have a page size of 10 items, but the last page contains only 2, the placeholder will take the height of the missing 8 items to preserve the size of the datagrid.
If we remove this, the datagrid's height will change when moving to the last page, so whatever's under it will "jump" up. Then if you move back to the previous page, the content under it "jumps" back down.

Jinnie: But, with the flex layout and the scrolling implementation, won't the grid always be as high, as defined by its container?
For me, setting the datagrid height in number of rows was a problem, because row height can not be predicted due to word wrapping, etc.
Before you introduced the internal datagrid scrolling, I had to adopt the rather questionable approach of having page size that's higher than the viewport, and scrolling outside of the datagrid. That's how gmail is implemented, as a reference.
So, my point is, isn't the correct use case now to have a fixed height on the datagrid, but in px, %, vh, whatever, and scrolling inside the datagrid itself?

youdz: We still have to support datagrids without a fixed height, though. They should expand as much as needed without needing to scroll. That's where the placeholder still had to do some work. Admittedly it didn't work well with multi-line rows, and we had that on our backlog.

A solution now would be to force the datagrid height on initial load with the first page, and keep it fixed from there on. That's a bit debatable, because a datagrid that didn't scroll on page 1 might start scrolling on page 2, even though the consumer never wanted it to scroll.

PR #649, opened 2017-03-24
[ui/ng] popover

jeeyun:
This is v1 of the popover implementation which will prevent the clipping of elements that popover (tooltips, dropdowns, etc). Note that this solution uses javascript (TypeScript) to absolutely position the element using translate properties. This means that our css-only (non-angular) versions of those components will continue to be clipped off. Therefore, please note that this PR does not fix the clipping of the validation tooltip, which is not yet an angular component.

This PR introduces a new angular-component (tooltip) which behaves exactly as the existing static tooltip but will not clip the element when contained inside a scrolling container.

In a later iteration of this, we might want to consider adding smart functionalities, such as repositioning the popover if the position specified is outside of the browser's viewport.

fixes ##37.


src/clarity-angular/popover/popover.ts
```diff
@@ -11,14 +11,25 @@
  * WARNING: This is a quick prototype version, use at your own risks.
```
mathisscott: Is this still considered a prototype?

src/clarity-angular/popover/popover.ts
```diff
@@ -45,33 +56,203 @@ export class Popover {
...
             anchor = anchor.parentNode;
         }
+        // explicitly override anchor's style to static
+        anchor.style.position = "static";
+
+        let marginLeft = parseInt(getComputedStyle(this.element).marginLeft, 10);
+        let marginRight = parseInt(getComputedStyle(this.element).marginRight, 10);
+        let marginTop = parseInt(getComputedStyle(this.element).marginTop, 10);
+        let marginBottom = parseInt(getComputedStyle(this.element).marginBottom, 10);
+
```
mathisscott: There are a lot of lookups for `this.element`, `anchorRect.*`, and `popoverRect.*` in `anchor()`.

Should we consider using a variable inside of `anchor` for some of these values? I think it would make any future refactor easier to handle and fewer lookups generally helps performance.

src/clarity-angular/popover/popover.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+        expect(container.style.overflow).toEqual("scroll");
+    });
+
+    it("positions the popover according to align points specified", function () {
+        // popovers above the anchor
+        popoverInstance.anchor(anchor, Point.TOP_LEFT, Point.BOTTOM_RIGHT);
+        expect(popover.style.transform).toEqual("translateX(0px) translateY(0px)");
+
+        popoverInstance.anchor(anchor, Point.TOP_LEFT, Point.BOTTOM_LEFT);
+        expect(popover.style.transform).toEqual("translateX(60px) translateY(0px)");
```
mathisscott: These tests would be more resilient if the translate values were computed instead of hard-coded. If we know that we want `popover.style.height` to be 50px then can't we use computed values (adding in whatever value we use for additional margin/positioning from the component added in?

If the component changes the additional 10px to an additional anything-other-than-10px, all these tests will break. But if the values were computed, it wouldn't.

This isn't something I would worry about for the PR here. But maybe we can add a `// TODO:`?

src/app/nav/header-types.demo.html
```diff
@@ -218,16 +217,16 @@
             &lt;/label&gt;
         &lt;/form&gt;
         &lt;div class=&quot;header-actions&quot;&gt;
-            &lt;clr-dropdown class=&quot;dropdown bottom-right&quot;&gt;
+            &lt;clr-dropdown c[clrMenuPosition]=&quot;'bottom-right'&quot;&gt;
```
mathisscott: Extra "c" in here...

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -84,6 +84,18 @@ $dropdown-white: clr-getColor(lightest);
             visibility: hidden;
             z-index: map-get($clr-layers, dropdown-menu);
 
+            .dropdown-menu {
```
Shijir: This `.dropdown-menu` is inside another `.dropdown-menu`? I am a bit confused about how this is used in the template.

Shijir: oh sorry, I think misunderstood the purpose of it. I guess this is a temporary hack that overrides the style that should take effects in 0.9.0. If then, we could make use of `& {` instead of using the class name again?

src/clarity-angular/popover/popover.ts
jeeyun: Nope, will remove the comment.

src/clarity-angular/popover/popover.ts
jeeyun: Hmm `element` is already a private variable of the class, so it seems weird to assign that to yet another local variable. As per your suggestion, I created a variable so that `getComputedStyle(this.element)` is called only once. 

I will leave the `anchorRect.*`, and `popoverRect.*` as is. Creating additional variables like:
```
let anchorRectLeft = anchorRec.left;
```
doesn't change the refactorability too much. And because the `anchor()` function goes through a series of switch statements, the look up for `anchorRect.*` / `popoverRect.*` is only happening once each per `anchor()` function call.

src/clarity-angular/popover/popover.spec.ts
jeeyun: I totally agree. Adding a TODO here.

src/app/nav/header-types.demo.html
jeeyun: Thanks for catching my sloppiness 👍

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -127,7 +131,6 @@
         }
 
         .datagrid-column {
-            position: relative;
```
jeeyun: We have had the 
```
container (with position: relative) + child (with position:absolute)
``` 
as a pattern for popovers. However, this results in popovers clipping. This is removed as we are now using our popover implementation for the filter popover rather than relying on the `position: relative`.

src/clarity-angular/dropdown/_dropdown.clarity.scss
jeeyun: As per discussion offline, I'm leaving this in as is. This helps support backward compatibility until the markup for dropdown < 0.8.13 is deprecated.

src/clarity-angular/popover/popover.directive.ts
```diff
@@ -0,0 +1,61 @@
...
+                createPopoverFn();
+            }
+        }
+    }
+
+    createPopover() {
+        let embeddedViewRef = this.viewContainer.createEmbeddedView(this.templateRef);
+        this._popoverInstance = new Popover(embeddedViewRef.rootNodes[0]);
+        this._popoverInstance.anchor(
+            this.anchorElem, this.anchorPoint, this.popoverPoint, this.popoverOptions);
```
youdz: In the future, we might need a way to account for popover parameters that change dynamically. Especially positioning. Just throwing this out there, I'm not asking you to handle it now, of course. 😄

src/clarity-angular/popover/popover.ts
youdz: You're right, this comment should not be here anymore.

src/clarity-angular/popover/popover.ts
```diff
@@ -45,33 +56,203 @@ export class Popover {
...
+        if (useAnchorParent) {
             anchor = anchor.parentNode;
         }
+        // explicitly override anchor's style to static
+        anchor.style.position = "static";
+
+        let marginLeft = parseInt(getComputedStyle(this.element).marginLeft, 10);
+        let marginRight = parseInt(getComputedStyle(this.element).marginRight, 10);
+        let marginTop = parseInt(getComputedStyle(this.element).marginTop, 10);
+        let marginBottom = parseInt(getComputedStyle(this.element).marginBottom, 10);
```
youdz: You want to get the computed styles only once, on some browsers this might actually be a heavy-ish operation.

src/clarity-angular/datagrid/datagrid-action-overflow.ts
```diff
@@ -4,49 +4,41 @@
...
     `
 })
 
-export class DatagridActionOverflow implements OnDestroy, AfterViewInit {
+export class DatagridActionOverflow {
 
-    constructor(private elementRef: ElementRef, private rowActionService: RowActionService) {}
+    public anchorPoint: Point = Point.RIGHT_CENTER;
+    public popoverPoint: Point = Point.LEFT_CENTER;
+    public popoverOptions: PopoverOptions = { useAnchorParent: true, offsetX: 5 };
```
jeeyun: I got away without using popoverOptions all together here by using the margin as per your suggestion.

src/app/nav/header-types.demo.html
```diff
@@ -51,16 +51,16 @@
                 </a>
             </div>
             <div class="header-actions">
-                <clr-dropdown class="dropdown bottom-right">
+                <clr-dropdown [clrMenuPosition]="'bottom-right'">
```
jeeyun: updated for website

src/clarity-angular/clarity.module.ts
```diff
@@ -52,8 +56,10 @@ import {ClrResponsiveNavigationService} from "./nav/clrResponsiveNavigationServi
         LAYOUT_DIRECTIVES,
         MODAL_DIRECTIVES,
         NAVIGATION_DIRECTIVES,
+        POPOVER_DIRECTIVES,
```
jeeyun: done.

src/clarity-angular/popover/popover.directive.ts
jeeyun: made a note with a todo.

jbgarr:
@jeeyun So are there any updates related to the validation tooltip being clipped when a form is used in a modal? Any proposed workarounds if that is still a work in progress? I'm facing the issue now and I'm not sure how to best handle it. Thanks for any help you can provide.

PR #657, opened 2017-03-29
[NG] alpha wizard for clarity angular

### LINKED ISSUE #156: [NG] Wizard refactor

mathisscott:
This PR completes the alpha effort for the new clarity wizard. A quick overview of the new features/capabilities follows:

• i18n buttons and text throughout
• better support for two-way binding, more public convenience methods, and a great many more inputs and outputs at all levels
• removed dependency on tabs component
• improved DOM structure, removing dependency that nav-items and pages be presented in the same order
• allowing for pages to be generated and skipped with angular directives
• implementation of ghost pages from original design spec
• improved support for long title names
• allowing for buttons to be overridden at the page level or for a default wizard level set of buttons to be used
• header actions
• in-line wizard
• allowing for the navigation inside the wizard to be manipulated programmatically
• implementing a means whereby users could circumvent default actions like cancel to override them with their own code

This effort constitutes a 100% rewrite of the wizard component while attempting to maintain as much backwards compatibility as possible. As this is an alpha, there are known issues that we will try to address before final release.

Note also that some PR change requests may be pushed to post-alpha as we work to make the new wizard available as soon as possible.

More information on backwards compatibility and migration will be coming in the 0.9.0-alpha release notes.

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ Safari
✔︎ IE10
✔︎ IE11
✔︎ Edge

Known Issues:

- There is an issue with using anchor tags as header actions in some browsers. This may not be addressed before 0.9.0.
- There is an issue with ghost pages animation in IE 10, Safari, and the latest Firefox. This may also be pushed until after 0.9.0.
- IE 10 has some alignment issues with the title bar in the content area. These are waiting to be addressed until after we decide if we continue support for IE 10 in 0.9.0.
- The button bar in the footer is not usable in IE 10. This is a nontrivial flexbox issue. This is also on hold until we decide if we continue support for IE 10 in 0.9.0.
- There is a known issue with page collection syncing up after pages are added or removed from the wizard with ngIf. The plan is to address this before the 0.9.0 release.
- There is a known issue with bizarre behavior around alt-cancel routines, especially when there is an alt cancel routine at the wizard _and_ page level. The plan is to address this before the 0.9.0 release.
- There are additional tests that need to be written and some cleanup that generally needs to be done around existing unit tests. Some of this will be done before the 0.9.0 release, but maybe not all.
- Unit tests need to be written for the wizard as a whole. A test plan has already been put into place. The intent is to complete the test plan before the 0.9.0 release, although some tests may need to wait.

Any issues that are pushed beyond 0.9.0 will be added to our backlog as individual github issues and added to our 1.0 backlog if appropriate.

Closes #156


youdz:
Why, oh why? That's going to take a while. 😭

src/clarity-angular/wizard/wizard-header-action.ts
```diff
@@ -0,0 +1,61 @@
...
+    click(): void {
+        if (this.disabled) {
+            return;
+        }
+
+        // passing the header action id allows users to have one method that
+        // routes to many different actions based on the type of header action
+        // clicked. this is further aided by users being able to specify ids
+        // for their header actions.
+        this.headerActionClicked.emit(this.id);
```
youdz: I think you want to emit `this._id`, which is the id provided by the user. Otherwise, you return to them an id that you prefixed with a pretty long string, which seems strange.

src/clarity-angular/wizard/wizard-page.spec.ts
```diff
@@ -39,31 +82,1201 @@ class TestComponent {
...
+                            <span class="alert-text">
+                                i believe the answer is {{innerProjector/2}}
+                            </span>
+                        </div>
+                    </clr-alert>
+                </clr-wizard-page>
+            </clr-wizard>
+            `
+        })
+        class ViewTestComponent {
```
youdz: Not only does creating this monster of a component for each and every test seem like it would slow down everything and potentially create issue, you also _declare the whole class_ inside a Jasmine function. I have no idea how many times this is declared, how many times Angular compiles it on the fly, how many instances you get every time you compile the component after this, ... This should really, _really_ be at the module level.

src/clarity-angular/wizard/wizard-step.spec.ts
```diff
@@ -1,49 +1,49 @@
...
+// class TestComponent {
+//     @ViewChildren(WizardStep) wizardStepChildren: QueryList<WizardStep>;
+//     open: boolean = true;
+// }
 
-describe("WizardStep", () => {
-    let fixture: ComponentFixture<any>;
-    let compiled: any;
-    let wizardSteps: WizardStep[];
+// describe("WizardStep", () => {
```
youdz: I'm guessing this file should just be deleted, it was part of the old wizard?

src/clarity-angular/wizard/wizard-stepnav-item.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+    page: MockPage;
+    projector: string = "foo";
+
+    @ViewChild(WizardStepnavItem) stepNavItem: WizardStepnavItem;
+    @ViewChild(WizardPageNavTitleDirective) navTitleRef: WizardPageNavTitleDirective;
+
+// TODO: AfterContentInit TO AFTERVIEWINIT?
+    public ngAfterContentInit(): void {
+        this.page.navTitle = this.navTitleRef.pageNavTitleTemplateRef;
+    }
```
youdz: Yes, AfterViewInit. This is a ViewChild, not a ContentChild.

src/clarity-angular/wizard/providers/button-hub.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    describe("Button Hub Service", function() {
+
+        let context: TestContext<Wizard, ButtonHubTest>;
+        let buttonHubService: ButtonHubService;
+        let wizardNavigationService: WizardNavigationService;
+        let pageCollectionService: PageCollectionService;
+
+        beforeEach(function() {
+            context = this.create(Wizard, ButtonHubTest);
+            buttonHubService = context.getClarityProvider(ButtonHubService);
```
mathisscott: Yes, I agree. It was a lot of code at the backend of this component. Much of it, I need to go back over. I have an action item to do that before or after the wizard is finalized.

src/clarity-angular/wizard/wizard-button.ts
```diff
@@ -0,0 +1,174 @@
...
+    danger: "danger"
+};
+
+export const CUSTOM_BUTTON_TYPES: any = {
+    cancel: "custom-cancel",
+    previous: "custom-previous",
+    next: "custom-next",
+    finish: "custom-finish",
+    danger: "custom-danger"
+};
```
mathisscott: you can designate a button as "custom-finish" and it will inherit the look and feel (as well as the functionality making it available, unavailable, enabled, and disabled) as a regular finish button.

it's a convenience for people needing custom buttons.

src/clarity-angular/wizard/providers/page-collection.ts
```diff
@@ -0,0 +1,237 @@
...
+// TODO: create Observables for when list of pages changes?
+
+    public pages: QueryList<WizardPage>;
+
+    public get pagesAsArray(): WizardPage[] {
+        return this.pages ? this.pages.toArray() : [];
+    }
+
+    public get pagesCount(): number {
+        return this.pagesAsArray.length;
```
mathisscott: i'll change this. thanks.

src/clarity-angular/wizard/wizard-page.spec.ts
mathisscott: got it. will look into that when i revisit the unit tests.

src/clarity-angular/wizard/wizard-step.spec.ts
mathisscott: yes, i'll do that. thanks

src/clarity-angular/wizard/wizard-stepnav-item.spec.ts
mathisscott: thanks. will change.

src/clarity-angular/wizard/wizard-header-action.ts
mathisscott: Makes sense. I'll change that.

src/clarity-angular/wizard/providers/page-collection.ts
mathisscott: done. coming in next commit.

src/clarity-angular/wizard/wizard-header-action.ts
mathisscott: done. will be in next commit.

PR #664, opened 2017-03-29
[ng] removing const in SortOrder enum so it's not removed during compilation

jeeyun:
"Const enums can only use constant enum expressions and unlike regular enums they are completely removed during compilation."

reference: https://www.typescriptlang.org/docs/handbook/enums.html

### DIFF
--- a/src/app/datagrid/sorting/sorting.ts
+++ b/src/app/datagrid/sorting/sorting.ts
@@ -5,8 +5,7 @@
  */
 import { Component } from "@angular/core";
 
-import { SortOrder } from "../../../clarity-angular/datagrid";
-
+import { SortOrder } from "../../../clarity-angular";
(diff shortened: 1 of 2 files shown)


clementcur:
@jeeyun have you tried to run the app with that change? I'm seeing and Angular error, see screenshot.
screen shot 2017-03-29 at 3 28 26 pm

jeeyun:
Funny, I only tested consuming of the built package, but not in the demo app. I'm able to see replicate the error that you are seeeing. Will update after investigation.

jeeyun:
@clementcur i believe the second commit should fix the problem.

clementcur:
great, just tried it, yep it fixes the issue! sorry for the trouble!

jeeyun:
No, thank you for catching that! 👍

Issue #691, opened 2017-04-08
[Angular] Button group click event doesn't tell what child button is clicked. Child buttons also not responding to their own click event


whizkidwwe1217:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Can't determine what button is clicked.

### Actual behavior
Should be able to determine what button is clicked in a button group.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

hippee-lee:
Hi @whizkidwwe1217 

Can you submit a [https://plnkr.co/edit/8TwwdL?p=info](plunker) with an example of your usage so we can investigate?

whizkidwwe1217:
Hi,
There's nothing special about the code. Just wire a simple click event to a
button group or to a child button in a button group. Here's my sample code:

app.component.html

**
<clr-button-group class="btn-primary" *(click)="groupClicked($event)"*>
  <clr-button id="btnCreate">Create</clr-button>
  <clr-button id="btnFavorite">Favorite</clr-button>
</clr-button-group>
**
<clr-button-group class="btn-primary">
  <clr-button id="btnCreate2"
*(click)="childClicked($event)"*>Create</clr-button>
  <clr-button id="btnFavorite2">Favorite</clr-button>
</clr-button-group>

app.component.ts

groupClicked(e) {
  // How can I get the button that was clicked?
  console.log(e);
}

childClicked(e) {
  console.log(e);
}

whizkidwwe1217:
Hi,
There's nothing special about the code. Just wire a simple click event to a button group or to a child button in a button group. Here's my sample code:

**app.component.html**
```
<!-- Sample button group with click event. THIS WORKS BUT CANT TELL WHAT BUTTON WAS CLICKED -->
<clr-button-group class="btn-primary" (click)="groupClicked($event)">
  <clr-button id="btnCreate">Create</clr-button>
  <clr-button id="btnFavorite">Favorite</clr-button>
</clr-button-group>

<!-- Sample button group with child button click event. THIS DOES NOT WORK -->
<clr-button-group class="btn-primary">
[...]

PR #698, opened 2017-04-11
[NG] Button Group Click Event Fix

adityarb88:
Fixes: https://github.com/vmware/clarity/issues/691

Added a `click` `EventEmitter` to the Angular `clr-button` component.

Tested on Chrome.

----EDIT----

Note: This PR only adds a `EventEmitter` for the `clr-button` component and not the `clr-button-group` component.

### DIFF
--- a/src/clarity-angular/button-group/button.ts
+++ b/src/clarity-angular/button-group/button.ts
@@ -5,14 +5,14 @@
  */
 
 
-import {Component, ViewChild, TemplateRef, Input, SkipSelf, Optional} from "@angular/core";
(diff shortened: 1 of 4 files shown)


src/clarity-angular/button-group/button.ts
```diff
@@ -58,6 +58,12 @@ export class Button {
         }
     }
 
+    @Output("click") _click: EventEmitter<boolean> = new EventEmitter<boolean>(true);
```
youdz: You want false. This should **not** be async.

Issue #669, opened 2017-04-02
clr-dg-column separators improperly positioned


jarinwhite:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Borders separating datagrid columns should be positioned after their parent element

### Actual behavior
Borders are drawn relative to the document (or nearest relatively-positioned parent) rather than the parent column

### Reproduction of behavior
https://plnkr.co/edit/absCsKRnnNxW9PaQ6rak?p=preview

Seems to be fixed with `position: relative` on clr-dg-column (`.datagrid .datagrid-column`): https://plnkr.co/edit/RdlY4u2IHGqOuRDVl4tx?p=preview

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 0.8.13

* **OS and version:** Windows 10

* **Browser:** Chrome 56 | Firefox 52

Issue #621, opened 2017-03-20
Datagrid: Resizable columns


Shijir:
**Select one ...**  (check one with "x")
```
[ ] bug
[ x] feature request
[ ] enhancement
```

### Expected behavior

We need to make Datagrid columns resizable so users can adjust the width of the Datagrid columns by dragging the end of the column header. To implement this feature, we need to investigate various potential solutions. For example, right now off top of my head, we could make use of generic `mouse` and `touch` events or native HTML5 Drag and Drop API (Not fully supported in all browsers, so might require polyfills).

This feature is first requested in https://github.com/vmware/clarity/issues/145.
This could be also related to https://github.com/vmware/clarity/issues/620.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

dabkowitz:
This is what Olga and I came up with...  
Resizing Columns:
- The cursor changes when the user hovers on or near the separator line dividing column headers.
- Only adjacent columns resize at one time.
- The user resize actions should be persistent between log in sessions.

Reordering Columns:
- Optimally this would be a drag and drop action.   User click and drag the column name as opposed to click for sort.
- Optionally reordering could be accomplished in the dialog where the user can select which columns are visible.   This is sub optimal because it is removed and vertical movement does not map well to the horizontal movement of the columns.

Shijir:
Hi @dabkowitz, thanks a lot for your suggestions. I am currently working on column resizing and close to wrapping up. 

>1. The cursor changes when the user hovers on or near the separator line dividing column headers.
>2. Only adjacent columns resize at one time.
>3. The user resize actions should be persistent between log in sessions.

As your first suggestion, the cursor changes when a user tries to resize the column in the current implementation. Also in the current implantation, when a user resizes a column, only that column will be affected. Finally, users can listen to column resizing event through `(clrDgColumnResize)` so they can save the resized width of the columns, apply the width on the next session. 

You can look at the current implementation of column resizing (work in progress) demo by downloading this [video](https://www.dropbox.com/s/71uk115jd54ei8g/datagrid-column-resizing.mov?dl=0).

Jinnie:
Hi @Shijir,
[...]

PR #699, opened 2017-04-11
[NG] Fix datagrid column seperator

Shijir:
Resolves https://github.com/vmware/clarity/issues/669
This fix will be also needed for https://github.com/vmware/clarity/issues/621

### DIFF
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -52,9 +52,16 @@
         display: flex;
         flex-flow: row nowrap;
     }
-    .datagrid-column, .datagrid-cell {
+    .datagrid-column {
+        display: flex;
         flex: 1 1 auto;
+        &.datagrid-fixed-width {
+            flex: 0 0 auto;
+        }
+    }
 
+    .datagrid-cell {
+        flex: 1 1 auto;
         &.datagrid-fixed-width {
             flex: 0 0 auto;
         }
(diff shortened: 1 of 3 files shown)


src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -36,7 +36,7 @@
         display: flex;
         flex-flow: column nowrap;
         // Firefox needs this
-        min-height: 0;
+        min-height: $clr_baselineRem_6;
```
youdz: Wait, what? Why are we making the table part of the datagrid at least 6 baselines high? That's far more than the visual spec.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -110,7 +110,12 @@
 
         .datagrid-body {
             .datagrid-row:last-child {
+
                 .datagrid-cell {
+
+                    border-bottom: 1px solid $clr-lighter-midtone-gray;
+                    margin-bottom: -1px;
```
youdz: Once again, what does this have to do with the column separator? 

Note that the negative margin-bottom was actually tried initially, and it created unwanted scrollbars with various browsers (especially IE). So this is probably breaking a lot of things, unless you tested all the possible cases (rows filling the datagrid, not filling it, scrollable datagrid, flexible height datagrid, no pagination, ...) on every browser we support, and with scrollbars displayed at all times on Mac.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -131,6 +136,7 @@
         }
 
         .datagrid-column {
+            position: relative;
```
youdz: @jeeyun removed this on purpose, adding it will break the filters again. Please check up with her.

src/clarity-angular/datagrid/datagrid.html
```diff
@@ -32,11 +32,11 @@
...
         </div>
-    </div>
 
-    <!-- Custom placeholder overrides the default empty one -->
-    <ng-content select="clr-dg-placeholder"></ng-content>
-    <clr-dg-placeholder *ngIf="!placeholder"></clr-dg-placeholder>
+        <!-- Custom placeholder overrides the default empty one -->
+        <ng-content select="clr-dg-placeholder"></ng-content>
+        <clr-dg-placeholder *ngIf="!placeholder"></clr-dg-placeholder>
+    </div>
```
youdz: Once again, what's this doing here? This is a major change that completely impacts our column size calculator. Right now, I'm pretty sure it's actually breaking it.

Shijir:
@youdz having `position:relative` was solving the separator positioning issue. Also, it was going to be needed for the column resizer. However, with this solution, the filter clipping issue would come back if there are no rows. That's why I set the minimum height which prevents the filter box from getting clipped and go under the`.datagrid-table-wrapper`. Having a minimum height on  `.datagrid-table-wrapper` was also pushing the `clr-dg-placeholder` down when there are no rows. So I placed that inside `.datagrid-table-wrapper`. 

But I will replace this solution with another as this one forces these unwanted collateral changes.

PR #700, opened 2017-04-11
[NG] Datagrid expandable rows

youdz:
Expandable rows as a solution to the master/detail pattern.

This PR also introduces a new pattern for "loading listeners". If it works well, we will use it everywhere we have a spinner for lazy loading: treeview, stackview, tabs, ...

Several future improvements already noted:
1. We will be able to move to full Angular animations once 4.1 is out. Until then, we have to use this manual web-animations hack.
2. Many animation tests are not passing in PhantomJS, I had to comment them out for now. Some of these are important, we need to fix them ASAP. I'm expecting Angular 4.1 to help a lot.
3. The detail container should not be cleared immediately when collapsing, it should wait for the animation to be over.
4. The headers re-align immediately if a scrollbar is supposed to appear during the animation, rather than when the scrollbar actually appears.

### DIFF
--- a/src/app/datagrid/expandable-rows/expandable-rows.html
+++ b/src/app/datagrid/expandable-rows/expandable-rows.html
@@ -1,0 +1,87 @@
+<!--
+  ~ Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Expandable rows</h2>
+
+<form>
+    <div class="form-group">
+        <label>Type of detail</label>
+        <div class="radio-inline">
+            <input type="radio" name="detail" id="detail-default" value="default" [(ngModel)]="detail">
+            <label for="detail-default">Default</label>
+        </div>
+        <div class="radio-inline">
+            <input type="radio" name="detail" id="detail-columns" value="columns" [(ngModel)]="detail">
+            <label for="detail-columns">Columns</label>
+        </div>
+    </div>
+    <div class="form-group">
+        <div class="checkbox-inline">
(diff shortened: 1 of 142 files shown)


src/app/datagrid/expandable-rows/expandable-rows.html
```diff
@@ -0,0 +1,87 @@
...
+                    quam. Proin rhoncus lacus et tempor dignissim. Vivamus sem quam, pellentesque aliquet suscipit eget,
+                    pellentesque sed arcu. Vivamus in dui lectus. Suspendisse cursus est ac nisl imperdiet viverra. Aenean
+                    sagittis nibh lacus, in eleifend urna ultrices et. Mauris porttitor nisi nec velit pharetra porttitor.
+                    Vestibulum vulputate sollicitudin dolor ut tincidunt. Phasellus vitae blandit felis. Nullam posuere
+                    ipsum tincidunt velit pellentesque rhoncus. Morbi faucibus ut ipsum at malesuada. Nam vestibulum felis
+                    sit amet metus finibus hendrerit. Fusce faucibus odio eget ex vulputate rhoncus. Fusce nec aliquam leo,
+                    at suscipit diam.
+                </template>
+
+                <template [ngIf]="detail === 'columns'">
```
mathisscott: If you are merging to master, you're going to want to change these to `ng-template`

src/clarity-angular/loading/loading-listener.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
```
mathisscott: Not a big deal but I thought we were updating copyrights to 2016-2017 when we came across them.

PR #708, opened 2017-04-12
[NG] Fixing alt cancel issue in new wizard

mathisscott:
These changes Address the following issues with the 0.9.0 wizard:

• an issue where and alt cancel routine was fired twice at the wizard level
• an issue where a page level alt cancel routine fired and was followed by a wizard level alt cancel
• an issue with enter and leave animations in the modal

To resolve this issue, I gave precedence to page level alt cancel routines. If the current page has defined either and preventDefault or stopCancel, it will prevent comparable events at the wizard level from firing. This was already the case for finish, previous, next, and danger buttons. The logic was not implemented for cancel buttons, however.

Tested in:
✔︎ Chrome

### DIFF
--- a/src/clarity-angular/modal/modal.ts
+++ b/src/clarity-angular/modal/modal.ts
@@ -118,11 +118,13 @@ export class Modal implements OnChanges, OnDestroy {
     @Input("clrModalClosable") closable: boolean = true;
     @Input("clrModalSize") size: string;
     @Input("clrModalStaticBackdrop") staticBackdrop: boolean = false;
-    @Input("clrModalSkipAnimation") skipAnimation: boolean = false;
+    @Input("clrModalSkipAnimation") skipAnimation: string = "false";
 
     // presently this is only used by wizards
     @Input("clrModalGhostPageState") ghostPageState: string = "hidden";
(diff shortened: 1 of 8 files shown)


src/clarity-angular/modal/modal.ts
```diff
@@ -118,11 +118,13 @@ export class Modal implements OnChanges, OnDestroy {
     @Input("clrModalClosable") closable: boolean = true;
     @Input("clrModalSize") size: string;
     @Input("clrModalStaticBackdrop") staticBackdrop: boolean = false;
-    @Input("clrModalSkipAnimation") skipAnimation: boolean = false;
+    @Input("clrModalSkipAnimation") skipAnimation: string = "false";
```
jeeyun: why was this changed to string?

mathisscott: Because the animations are looking for it as a string. When I had it as a boolean, the animations were messed up because it was being passed to the animation/transition as boolean false.

This was the only way I could fix the animation in the modal. It did not affect the wizard, however, because that passed the value down through an input which the animation apparently converted from boolean to string at some point.

Issue #706, opened 2017-04-12
Datagrid filter "jumps" down when it changed in Server-driven mode


savelievser:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Filter stays on it's place.

### Actual behavior
Filter jumps to the bottom of the grid while data is loading and returns back when loading is over.

### Reproduction of behavior
Clarity's own documentation page exhibits this behavior.
https://vmware.github.io/clarity/documentation/datagrid
1. Scroll to Server-driven grid
2. Open any filter and start typing

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** latest Chrome

jeeyun:
This should be fixed by #710.

PR #710, opened 2017-04-12
[ng] allow multiple filters to be open

### LINKED ISSUE #668: Datagrid filters broken when multiple filters are open
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When opening multiple filters on a datagrid, each filter should display as "open" and operate independently.

### Actual behavior

After opening one filter and leaving it open, opening another filter takes two clicks. Upon the second click, a box appears with only an "x" and no filter contents. This small box cannot be closed. Continuing to click the filter icon causes erratic behavior, such as opening new filter windows on top of the existing ones.

image

image

### Reproduction of behavior
Clarity's own documentation page exhibits this behavior. 
https://vmware.github.io/clarity/documentation/datagrid

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** Verified in latest Chrome and Firefox

jeeyun:
Addressed in this commit are:
- change in popover to allow multiple to be open (thru options)
- ui/ng change to prevent filter popovers from shifting when the loading spinner is on (i.e. refactor loading spinner to not rely on datagrid class having `position:relative`)
- changing popover implementation to close on scroll (as per discussion with @reddolan)

Fixes #668 and #706.

### DIFF
--- a/src/clarity-angular/datagrid/datagrid.html
+++ b/src/clarity-angular/datagrid/datagrid.html
@@ -5,7 +5,14 @@
   -->
 
 <ng-content select="clr-dg-action-bar"></ng-content>
-<div class="datagrid" [class.loading]="loading">
+<div class="datagrid" #datagrid>
+    <div class="datagrid-spinner-wrapper" *ngIf="loading">
+        <!--we are using the offsetHeight here to set the height of spinner which is brittle;
+            re-evaluate later to see if there is a better fix for this-->
+        <div class="datagrid-spinner" [style.height.px]="datagrid.offsetHeight">
+            <div class="spinner">Loading...</div>
+        </div>
+    </div>
     <div clrDgTableWrapper class="datagrid-table-wrapper">
         <div clrDgHead class="datagrid-head">
             <div class="datagrid-row datagrid-row-flex">
@@ -44,7 +51,8 @@
 
             <!-- Custom placeholder overrides the default empty one -->
             <ng-content select="clr-dg-placeholder"></ng-content>
-            <clr-dg-placeholder *ngIf="!placeholder"></clr-dg-placeholder></div>
+            <clr-dg-placeholder *ngIf="!placeholder"></clr-dg-placeholder>
+        </div>
     </div>
 
     <!--
@@ -52,8 +60,4 @@
         everything when using custom elements with display:table-cell.
     -->
     <ng-content select="clr-dg-footer"></ng-content>
-
-    <div class="datagrid-spinner" *ngIf="loading">
-        <div class="spinner">Loading...</div>
-    </div>
 </div>
--- a/src/clarity-angular/popover/popover.directive.ts
+++ b/src/clarity-angular/popover/popover.directive.ts
@@ -1,5 +1,6 @@
-import {Directive, EmbeddedViewRef, Input, TemplateRef, ViewContainerRef} from "@angular/core";
+import {Directive, EmbeddedViewRef, EventEmitter, Input, Output, TemplateRef, ViewContainerRef} from "@angular/core";
 import {PopoverOptions, Point, Popover} from "./popover";
+import {Subscription} from "rxjs/Subscription";
 
 let openCount: number = 0;
 let waiting: Array<() => void> = []; // pending create functions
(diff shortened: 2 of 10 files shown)


src/clarity-angular/datagrid/datagrid.ts
```diff
@@ -223,9 +227,18 @@ export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
...
-        this._subscriptions.push(this.sort.change.subscribe(() => this.triggerRefresh()));
-        this._subscriptions.push(this.filters.change.subscribe(() => this.triggerRefresh()));
-        this._subscriptions.push(this.page.change.subscribe(() => this.triggerRefresh()));
+
+        // these changes (sort, filters, page) all should call triggerRefresh. The triggerRefresh
+        // function takes a snapshot of the most recent state and emits the refresh event. Here, we combine
+        // and debounce them so that the rapid successive changes don't throw the infamous
+        // "expression has changed after it was checked" error.
+        let merged = Observable.merge(this.sort.change, this.filters.change, this.page.change)
+            .debounce(() => Observable.interval(300));
```
youdz: 300ms debounce seems gigantic. As I mentioned on Slack, I'd much rather not re-emit duplicate states than do a debounce. And I don't believe it would be complicated to implement: you still combine the observables, turn them into an observable of `State` objects, and remove duplicates. Or something like this. You can even manually just test if the state has changed in the `triggerRefresh()` method, and abort if it hasn't. Many possible solutions.

But if really _really_ we have to debounce, then at least put something that the user can't actually see, like 30ms.

src/clarity-angular/datagrid/datagrid.spec.ts
```diff
@@ -77,29 +78,32 @@ export default function (): void {
                     expect(context.testComponent.nbRefreshed).toBe(1);
                 });
 
-                it("emits once when the sort order changes", function () {
+                it("emits once when the sort order changes",  fakeAsync(() => {
                     context.testComponent.nbRefreshed = 0;
                     let sort: Sort = context.getClarityProvider(Sort);
                     sort.toggle(new TestComparator());
+                    tick(1000);
```
youdz: Ugh. That's why debounce is so horrible in a plain component library. Debouncing should happen on the app level.

src/clarity-angular/popover/popover.directive.ts
```diff
@@ -16,24 +16,28 @@ export class PopoverDirective {
...
-                });
+                } else {
+                    waiting.push(() => {
+                        this.createPopover();
+                    });
+                }
             }
         } else {
             this.viewContainer.clear();
             this.destroyPopover();
```
youdz: So if I open 2 popovers, when I close the first one, everything can scroll again?
In that case, my second popover will just stay there in the wrong spot as I scroll.

src/clarity-angular/datagrid/datagrid.ts
jeeyun: The two states that produce the EHCAIWC are different (but I think the second state change fires mid-cycle of the first refresh?), so not re-emitting duplicate states won't fix this problem. I'm willing to dig this a bit more to see if there are alternative solutions rather than debounce.

src/clarity-angular/popover/popover.directive.ts
jeeyun: Ugh you are right. Let me fix this.

jeeyun: As per discussion with @youdz and @reddolan we are moving to an implementation where we close all popovers on scroll.

src/clarity-angular/popover/popover.directive.ts
```diff
@@ -47,14 +55,23 @@ export class PopoverDirective {
         //inside of the popover. For Eg: Button Groups
         embeddedViewRef.detectChanges();
 
-        this._popoverInstance = new Popover(embeddedViewRef.rootNodes[0]);
-        this._popoverInstance.anchor(
-            this.anchorElem, this.anchorPoint, this.popoverPoint, this.popoverOptions);
+        // filter out other nodes in the view ref so we are only left with element nodes
+        let elementNodes: HTMLElement[] = embeddedViewRef.rootNodes.filter((node: any) => {
+           return node.nodeType === 1;
+        });
```
youdz: So `<ng-container *clrPopover="true">Hello World</ng-container>` doesn't work? I'm okay with that, I just want to make sure I understand the constraints.

src/clarity-angular/popover/popover.ts
```diff
@@ -249,11 +242,11 @@ export class Popover {
         }
 
         this.element.style.transform = `translateX(${leftDiff}px) translateY(${topDiff}px)`;
+        return this._scroll.asObservable();
     }
 
     public destroy() {
-        this.element.style.transform = "none";
-        this.resumeScrolling();
```
youdz: Is there a reason why you removed the `this.element.style.transform` cleanup? The destroy here is not guaranteed to be called only when the element is destroyed, so we might leave an element out there with unwanted styles.

src/clarity-angular/popover/popover.ts
```diff
@@ -285,25 +277,11 @@ export class Popover {
         }
     }
 
-    private resumeScrolling() {
-        for (let container of this.originalOverflows) {
-            this.setInlineOverflow(container.e, container.overflow);
+    private removeScrollEventListeners() {
+        for (let elem of this.scrollableElements) {
+            elem.removeEventListener("scroll");
```
youdz: You should have a single listener function which you specifically remove by calling `elem.removeEventListener("scroll", listener);`. Because right now, you might remove an scroll listener that's not ours. See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Memory_issues (which means that on top of safely removing only our listener, we save some memory).

src/clarity-angular/datagrid/datagrid.html
```diff
@@ -5,7 +5,12 @@
   -->
 
 <ng-content select="clr-dg-action-bar"></ng-content>
-<div class="datagrid" [class.loading]="loading">
+<div class="datagrid" #datagrid>
+    <div class="datagrid-spinner-wrapper" *ngIf="loading">
+        <div class="datagrid-spinner" [style.height.px]="datagrid.offsetHeight">
```
jeeyun: Yes, I'll add a comment to this.

src/clarity-angular/popover/popover.directive.ts
jeeyun: Yes.. depending on the syntax we use, the rootNodes could contain just the element we want to use, or contain textnodes before and after.. so it became unreliable to try to pass in a specific node with something like `rootNodes[i]` (where `i` would be either 0 or 1 in the cases that I've seen so far).

src/clarity-angular/popover/popover.ts
jeeyun: Actually that's a mistake on my part. Re-adding the cleanup.

src/clarity-angular/popover/popover.ts
jeeyun: Yup will change this.

PR #735, opened 2017-04-17
[UI] Datagrid Footer CSS Bug Fixes

### LINKED ISSUE #730: Datagrid footer unit count description cutoff with items selected.
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Datagrid footer unit count description should be displayed whether items are selected or not.

### Actual behavior

When you have selection enabled on a datagrid and you select one or more items, the unit count in the datagrid footer gets cut off from "10 users" to "10 " with missing unit description.

### Reproduction of behavior

You can see it in the Clarity documentation website for both static counts and pagination.

Datagrid with selection enabled but not selected (proper footer displaying units):
image

Datagrid with selection enabled and one or more items selected (missing footer unit description):
image

### Environment details

* **Angular version:**
4.0.1

* **Clarity version:** 
0.9.0

* **OS and version:** 
Windows 10 and MacOS 10.12

* **Browser:** [all | Chrome 57 | Firefox 52.0.2 ]

adityarb88:
Fixes #730

Also fixes an existing bug with datagrid footer in IE10.

Before:
image

After:
image

Tested on:
IE10, 11, Edge, Chrome, Firefox, Safari

### DIFF
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -296,7 +296,8 @@
             }
         }
 
-        .datagrid-footer-select.checkbox {
+        .datagrid-foot-select.checkbox {
+            display: block;
             line-height: inherit;
             flex: 1 1 100%;
 
@@ -540,6 +541,11 @@
             margin-left: $clr_baselineRem_1_5;
             height: $clr_baselineRem_1_5 - 2px;
         }
+    }
+
+    .datagrid-foot-description {
+        display: block;
+        white-space: nowrap;
     }
 
     // Yes, this is not .datagrid-pagination on purpose.
--- a/src/clarity-angular/datagrid/datagrid-footer.spec.ts
+++ b/src/clarity-angular/datagrid/datagrid-footer.spec.ts
@@ -36,7 +36,7 @@ export default function(): void {
 
             context.detectChanges();
 
-            expect(context.clarityElement.querySelector(".datagrid-footer-select")).toBeNull();
+            expect(context.clarityElement.querySelector(".datagrid-foot-select")).toBeNull();
         });
 
         it("does not show the selection details when selection type is single", function() {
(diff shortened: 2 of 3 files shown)


src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -542,6 +543,11 @@
         }
     }
 
+    .datagrid-foot-description {
```
mathisscott: Why are we going from `footer` to `foot`? This seems inconsistent with other components where we use `footer`...

adityarb88: The other components in datagrid use `.foot` instead of `.footer` and I wanted to be consistent with that. @youdz, can i change all of them to `.footer` instead?

mathisscott: I don't think we need to do it for this PR. But at some point we need to settle on a convention and stick with it.

Whatever is decided, would you mind opening an issue for it?

adityarb88: I agree. I will check with @youdz and create an issue for it.

youdz: Actually, the reasoning was was that our classes would match standard HTML elements: 
- `.datagrid-head` <-> `thead`
- `.datagrid-body` <-> `tbody`
- `.datagrid-foot` <-> `tfoot`

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -542,6 +543,11 @@
...
+        display: block;
+        white-space: nowrap;
```
youdz: If we go with `white-space: nowrap`, then we should probably handle overflow, right? Otherwise long text will be really messed up here.

adityarb88: Yes we should. Before the text was being hidden because I think the height of the footer is fixed? Should I check with @reddolan on what he wants to do here? May be `text-overflow: ellipsis`?

reddolan:
We don't want to impose a predetermined width on the container. That begins to introduce many more complexities than are necessary. Having one in place means that teams might have to regularly override it for their case. Having the container expand to the content needed makes more sense and keeps teams from having to override.

In the cases where there is too much text in the container and it gets trimmed due to word-wrap, teams will need to create their own breakpoints or logic to handle the long content.

PR #738, opened 2017-04-17
[NG] Column Resizing

### LINKED ISSUE #621: Datagrid: Resizable columns

Shijir:
Tested on Chrome, Firefox, Safari, Edge, IE 11, & IE 10

resolves #621

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -77,7 +77,7 @@ module.exports = function (config) {
             pageTitle: 'Unit Tests',
             subPageTitle: __dirname
         },
-
+        browserNoActivityTimeout: 100000,
         port: 9876,
         colors: true,
         logLevel: config.LOG_INFO,
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -86,6 +86,9 @@
         .datagrid-column, .datagrid-cell {
             display: table-cell;
         }
+        .datagrid-column-separator {
+            display: none;
+        }
     }
 
     $clr-datagrid-icon-size: rem(1);
@@ -137,8 +140,14 @@
             border-bottom: 2*$clr-table-borderwidth solid $clr-light-midtone-gray;
         }
 
+
+
         .datagrid-column, .datagrid-cell {
             text-align: left;
+        }
+
+        clr-dg-column {
+            min-width: $clr_baselineRem_4;
         }
 
         .datagrid-column {
@@ -149,16 +158,18 @@
             .datagrid-column-title {
                 @include clr-no-styles-button();
                 color: $clr-dark-gray;
-                &:not([disabled]):hover {
-                    text-decoration: underline;
-                }
                 text-align: left;
                 flex: 1 1 auto;
                 align-self: flex-start;
             }
 
+            button.datagrid-column-title:hover {
+                text-decoration: underline;
+            }
+
             .datagrid-column-separator {
                 position: relative;
+                left: $clr-baselineRem_0_5;
                 flex: 0 0 auto;
                 width: 1px;
                 order: 100;
@@ -170,8 +181,32 @@
                     height: calc(100% + #{$clr-baselineRem_0_5 - 1});
                     width: 1px;
                     top: -$clr-baselineRem_0_25 + 1;
-                    left: $clr-baselineRem_0_5;
+                    left: 0;
                     background-color: $clr-lighter-midtone-gray;
+                }
+
+                .datagrid-column-handle {
+                    @include clr-no-styles-button();
+                    display: block;
+                    position: absolute;
+                    width: $clr-baselineRem_0_5+1;
+                    right: -$clr-baselineRem_0_25;
+                    top: -$clr-baselineRem_0_25;
+                    cursor: col-resize;
+                    height: calc(100% + #{$clr-baselineRem_0_5 - 1});
+                    z-index: map-get($clr-layers, dropdown-menu);
+                }
+                .datagrid-column-handle-tracker {
+                    position: absolute;
+                    right: 0;
+                    top: -$clr-baselineRem_0_5;
+                    display: none;
+                    width: 0;
+                    height: 100vh;
+                    border-right: 1px dotted $clr-action-blue-lighter;
+                }
+                .exceeded-max {
+                    border-right: 1px dotted rgba($clr-red-light-midtone, 0.3);
                 }
             }
 
@@ -274,6 +309,7 @@
                 }
             }
         }
+
 
         .datagrid-fixed-column {
             flex: 0 0 $clr-datagrid-fixed-column-size;
@@ -605,5 +641,6 @@
         padding: 0 !important;
         width: 0;
         flex: 0 0 auto !important;
+        min-width: 0 !important;
     }
 }
--- a/src/clarity-angular/datagrid/datagrid-column.ts
+++ b/src/clarity-angular/datagrid/datagrid-column.ts
@@ -4,19 +4,20 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import {
-    Component, ContentChild, HostBinding, Input, Output, EventEmitter
+    Component, ContentChild, HostBinding, Input, Output, EventEmitter, ElementRef, ViewChild
 } from "@angular/core";
-import { Subscription } from "rxjs";
+import {Subscription} from "rxjs";
 
-import { DatagridPropertyComparator } from "./built-in/comparators/datagrid-property-comparator";
-import { DatagridPropertyStringFilter } from "./built-in/filters/datagrid-property-string-filter";
-import { Comparator } from "./interfaces/comparator";
-import { CustomFilter } from "./providers/custom-filter";
-import { Sort } from "./providers/sort";
-import { DatagridFilterRegistrar } from "./utils/datagrid-filter-registrar";
-import { FiltersProvider } from "./providers/filters";
-import { DatagridStringFilterImpl } from "./built-in/filters/datagrid-string-filter-impl";
-import { SortOrder } from "./interfaces/sort-order";
+import {DatagridPropertyComparator} from "./built-in/comparators/datagrid-property-comparator";
+import {DatagridPropertyStringFilter} from "./built-in/filters/datagrid-property-string-filter";
+import {Comparator} from "./interfaces/comparator";
+import {CustomFilter} from "./providers/custom-filter";
+import {Sort} from "./providers/sort";
+import {DatagridFilterRegistrar} from "./utils/datagrid-filter-registrar";
+import {FiltersProvider} from "./providers/filters";
+import {DatagridStringFilterImpl} from "./built-in/filters/datagrid-string-filter-impl";
+import {SortOrder} from "./interfaces/sort-order";
+import {DragDispatcher} from "./providers/drag-dispatcher";
 
 @Component({
     selector: "clr-dg-column",
@@ -30,10 +31,20 @@ import { SortOrder } from "./interfaces/sort-order";
                 [clrDgStringFilter]="registered"
                 [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
 
-            <button class="datagrid-column-title" [disabled]="!sortable" (click)="sort()">
-                <ng-content></ng-content>
+            <ng-template #columnTitle><ng-content></ng-content></ng-template>
+            
+            <button class="datagrid-column-title" *ngIf="sortable" (click)="sort()">
+               <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
             </button>
-            <div class="datagrid-column-separator"></div>        
+            
+            <span class="datagrid-column-title" *ngIf="!sortable">
+               <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
+            </span>
+            
+            <div class="datagrid-column-separator">
+                <button #columnHandle class="datagrid-column-handle" tabindex="-1"></button>
+                <div #columnHandleTracker class="datagrid-column-handle-tracker"></div>
+            </div>        
         </div>
     `,
     host: {
@@ -41,7 +52,7 @@ import { SortOrder } from "./interfaces/sort-order";
     }
 })
 export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilterImpl> {
-    constructor(private _sort: Sort, filters: FiltersProvider) {
+    constructor(private _sort: Sort, filters: FiltersProvider, private _dragDispatcher: DragDispatcher) {
         super(filters);
         this._sortSubscription = _sort.change.subscribe(sort => {
             // We're only listening to make sure we emit an event when the column goes from sorted to unsorted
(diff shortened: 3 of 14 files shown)


src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -604,5 +637,6 @@
         padding: 0 !important;
         width: 0;
         flex: 0 0 auto !important;
+        min-width: 0 !important;
```
youdz: Yes, we really need to get rid of anything that's not actual `width` to properly compute the width of a column. That includes paddings that have been defined by the user, for instance, so this is the only way we have to override them.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -149,16 +152,18 @@
...
                 align-self: flex-start;
             }
 
+            button.datagrid-column-title:hover {
+                text-decoration: underline;
+            }
+
             .datagrid-column-separator {
                 position: relative;
+                left: 12px;
```
Shijir: Thanks! I've addressed it now.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -170,9 +175,37 @@
                     height: calc(100% + #{$clr-baselineRem_0_5 - 1});
                     width: 1px;
                     top: -$clr-baselineRem_0_25 + 1;
-                    left: $clr-baselineRem_0_5;
+                    left: 0;
                     background-color: $clr-lighter-midtone-gray;
                 }
+
+                .separator-dragger {
```
Shijir: I can make it `.datagrid-column-dragger`?

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -170,9 +175,37 @@
...
+                    background-color: transparent;
+                    border: none;
+                    cursor: col-resize;
+                    margin: 0;
+                    outline: none;
+                    padding: 0;
+                    height: calc(100% + #{$clr-baselineRem_0_5 - 1});
+                    z-index: map-get($clr-layers, dropdown-menu);
+                }
+                .separator-tracker {
```
Shijir: I might resort to `.datagrid-column-dragger` and `.datagrid-column-tracker`.

src/clarity-angular/datagrid/_datagrid.clarity.scss
hippee-lee: incorporating `handle` might help describe what it is.  e.g - `.datagrid-column-handle`

src/clarity-angular/datagrid/_datagrid.clarity.scss
hippee-lee: Some thought as above, `handle` might describe it better than _dragger_ or _tracker_

build/karma.conf.js
```diff
@@ -77,10 +77,14 @@ module.exports = function (config) {
             pageTitle: 'Unit Tests',
             subPageTitle: __dirname
         },
-
+        browserNoActivityTimeout: 100000,
```
mathisscott: Hahahahahaha.

Band aid...

src/clarity-angular/datagrid/_datagrid.clarity.scss
Shijir: @hippee-lee oh that's much better. I will use that instead of the current class name. Thanks!

src/clarity-angular/datagrid/_datagrid.clarity.scss
Shijir: I am thinking about making it `datagrid-column-handle-tracker` instead of `datagrid-column-tracker` ... 🤔

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    constructor(private el: ElementRef,
+                private renderer: Renderer,
+                private organizer: DatagridRenderOrganizer,
+                private domAdapter: DomAdapter) {
+
+        this.columnEl = el.nativeElement;
+    }
+
+    columnEl: Element;
```
youdz: `Element` seems weird here. I could see either `HTMLElement` or just `Node`. But either way, you can't actually type the native element because it would be different for each platform (browser, server-side, native mobile apps, ...). So I think `any` is safer as long as we use this in an Angular directive.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    ngOnDestroy() {
+        this.columnResizeEvents.map(event => event());
+    }
+
+
+    addDragListener(): Function[] {
+
+        this.draggerEl = this.columnEl.querySelector("button.separator-dragger");
+        this.draggerTracker = this.columnEl.querySelector(".separator-tracker");
```
youdz: Hmmm. This is normally be a no-go for me. It's very anti-Angular, ties our actual logic to style-related CSS classes, completely breaks outside the browser (which is the main reason we use the DomAdapter, to group these breaking native APIs in the same place), ...

I have to think about it for a moment, I would really like a cleaner solution that preserves the current architecture.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+                this.onDragEnd();
+
+                //Unsubscribing from itself
+                touchEndListener();
+            });
+        });
+
+        //returning mouseDownListener and touchStartListener so they can be unsubscribed.
+        return [mouseDownListener, touchStartListener];
+    }
```
youdz: I remember we said we would make this a vanilla Typescript class to handle drag&drop, which would be wrapped by an Angular directive like this one. Just like @jeeyun did for Popover.

I don't think I'm going to block the PR for this, but at the very least I'll open an issue to extract this soon.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+    columnResizeBy: number = 0;
+
+    draggerEl: Element;
+    draggerElWidth: number;
+    draggerTracker: Element;
+
+    pageStartPositionX: number;
+    dragDistancePositionX: number; //relative to pageStartPosition
+
+    @Output("clrDgColumnResize") resizeEmitter: EventEmitter<any> = new EventEmitter();
```
youdz: This should be an `EventEmitter<number>`, emitting the new width of the column in pixels.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+            touchEndListener = this.renderer.listenGlobal("document", "touchend", () => {
+
+                //Unsubscribing from mouseMoveListener
+                touchMoveListener();
+                this.onDragEnd();
+
+                //Unsubscribing from itself
+                touchEndListener();
+            });
+        });
```
youdz: Please make a method abstracting the "start, move, stop" pattern so that we don't have the exact same code twice, one for mouse and one for touch.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+import {DatagridRenderOrganizer} from "./render-organizer";
+
+@Directive({
+    selector: "clr-dg-column"
+})
+export class DatagridColumnResizer implements AfterViewInit, OnDestroy {
+
+
+    constructor(private el: ElementRef,
+                private renderer: Renderer,
```
youdz: With Angular 4, you should now use `Renderer2`.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    }
+
+    onDragMove(moveEvent: any): void {
+        let pageMovePosition = moveEvent.pageX || moveEvent.changedTouches[0].pageX;
+
+        this.dragDistancePositionX = this.getPositionWithinMax(pageMovePosition - this.pageStartPositionX);
+
+        this.renderer.setElementStyle(this.draggerTracker, "right", -1 * this.dragDistancePositionX + "px");
+        this.renderer.setElementStyle(document.body, "cursor", "col-resize");
```
youdz: Shouldn't you set the cursor on the whole body in `onDragStart()`? It seems counterproductive to do many times per second when you only need to do it once.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+        let mouseUpListener: any;
+
+        let touchStartListener: any;
+        let touchMoveListener: any;
+        let touchEndListener: any;
+
+        mouseDownListener = this.renderer.listen(this.draggerEl, "mousedown", () => {
+            this.onDragStart();
+
+            mouseMoveListener = this.renderer.listenGlobal("document", "mousemove", (mouseMoveEvent: any) => {
```
youdz: I think we might want to run this listener outside of the NgZone. At least we should try.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+    dragWithinMinWidth: boolean = false;
+
+    private _columnMinWidth: number;
+
+    get columnMinWidth(): number {
+        return this._columnMinWidth;
+    }
+
+    set columnMinWidth(value: number) {
+        this._columnMinWidth = value || 96;
```
youdz: Why are we handling the default min-width manually in the Javascript? I get that the resizer needs to know about the column's min-width to prevent it from shrinking too much, but we really shouldn't have that `96` hard coded here.

You just need to put `min-width: 96px` on `..datagrid-column` in the CSS and you're good to go.

src/clarity-angular/datagrid/render/header-renderer.ts
```diff
@@ -47,16 +63,32 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
+        }
+
+        if (!width) {
+
+            width = this.domAdapter.scrollWidth(this.el.nativeElement);
+
+            // a column shouldn't shrink beyond its minWidth when other columns get larger by resizing
+            if (width < this.minWidth) {
+                width = this.minWidth;
+            }
```
youdz: Why do we have to handle this manually? The way we do it right now, if you set a min-width on the header, the `scrollWidth` computed takes it into account already.

I actually have an issue with this header renderer having to worry about the resizer's min width. See my previous comment about putting the default min-width of 96px in the CSS, it would mean this renderer would never have to know anything about the resizer.

src/clarity-angular/datagrid/render/column-resizer.ts
```diff
@@ -0,0 +1,190 @@
...
+
+            this.columnResizeBy = this.dragDistancePositionX;
+
+            this.resizeEmitter.emit({
+                change: this.columnResizeBy,
+                newWidth: this.columnRectWidth + this.columnResizeBy,
+                minWidth: this.columnMinWidth
+            });
+
+            this.organizer.resize();
```
youdz: I don't know what happened, but this architecture changed since the last time I saw it. Or I misread last time, but either way I'm absolutely sure I explained how I wanted the resizer to update the organizer's strict widths directly, as opposed to having the various directives being so tightly coupled.

The problem is that this would be a pretty large change, and I don't want to postpone this anymore if it's working fine right now. But I'm definitely not happy with that component communication, it makes the various directives depend far too much on one another, which completely defeats the purpose of having multiple directives on the same selector.

src/clarity-angular/datagrid/render/main-renderer.ts
```diff
@@ -47,7 +49,22 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, On
      * Makes each header compute its width.
      */
     private computeHeadersWidth() {
+
+        let nbColumns: number = this.headers.length;
+        let nbStrictWidthColumns: number = this.headers.toArray().filter(column => column.strictWidth).length;
```
youdz: You're using the old width array, since you're about to compute it. So I'm expecting bugs here. You should just be using a boolean that becomes true if you see at least one non-strict column as you iterate.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -30,10 +30,20 @@ import { SortOrder } from "./interfaces/sort-order";
                 [clrDgStringFilter]="registered"
                 [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
 
-            <button class="datagrid-column-title" [disabled]="!sortable" (click)="sort()">
-                <ng-content></ng-content>
+            <ng-template #columnTitle><ng-content></ng-content></ng-template>
+            
+            <button class="datagrid-column-title" *ngIf="sortable" (click)="sort()">
+               <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
```
Shijir: 😢

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -170,9 +175,37 @@
...
+                    position: absolute;
+                    width: $clr-baselineRem_0_5+1;
+                    right: -$clr-baselineRem_0_25;
+                    top: -$clr-baselineRem_0_25;
+                    background-color: transparent;
+                    border: none;
+                    cursor: col-resize;
+                    margin: 0;
+                    outline: none;
+                    padding: 0;
```
Shijir: Thanks, I've incorporated it now.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: Using Renderer2 now.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: Now, it's emitting the new column width.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: This was really bad. 😞   Having `min-wdith: 96px` makes the code much simpler. 😃  I specified the min-width on `clr-dg-column` as `.datagrid-column` is also applied the selection column.

src/clarity-angular/datagrid/render/main-renderer.ts
Shijir: This has been updated as we've discussed on this.

src/clarity-angular/datagrid/render/header-renderer.ts
Shijir: I have removed it now :)

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: Addressed this issue.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: You are right! It's been moved to onDragStart() now.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: Addressed this issue.

src/clarity-angular/datagrid/render/column-resizer.ts
Shijir: Addressed this issue

PR #746, opened 2017-04-18
[NG] Datagrid Placeholder Fix

adityarb88:
Fixes: #631, #583

Changes:
1. Moves the placeholder into the `.datagrid-body`
2. `.datagrid-body` is now `display: flex`. This helps to position to placeholder in the center of the body.
3. `main-renderer` now calculates and sets the height on paginated datagrids. This ensures that the datagrid height is consistent when pages are changed and when the items displayed are less/more that what the page height can support.
4. Removes the datagrid placeholder min height logic and the corresponding tests.

Tested on:
IE11, Edge, Chrome, Safari, Firefox
My VM kept freezing when I ran IE10 so couldn't test on that.

### DIFF
--- a/src/app/datagrid/test-cases/test-cases.html
+++ b/src/app/datagrid/test-cases/test-cases.html
@@ -1,0 +1,239 @@
+<!--
+  ~ Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+<div class="clr-example">
+    <h4>Zero Rows</h4>
+    <p>Datagrid should be high enough to accomodate the placeholder.</p>
+    <clr-datagrid>
+        <clr-dg-column>User ID</clr-dg-column>
+        <clr-dg-column [clrDgField]="'name'">Name</clr-dg-column>
+        <clr-dg-column>Creation date</clr-dg-column>
+        <clr-dg-column>Favorite color</clr-dg-column>
+
+        <clr-dg-row *clrDgItems="let user of zeroUsers">
+            <clr-dg-cell>{{user.id}}</clr-dg-cell>
+            <clr-dg-cell>{{user.name}}</clr-dg-cell>
+            <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+            <clr-dg-cell>
+                <span class="color-square" [style.backgroundColor]="user.color"></span>
+            </clr-dg-cell>
+        </clr-dg-row>
+
+        <clr-dg-footer>
+            {{zeroUsers.length}} users
+        </clr-dg-footer>
+    </clr-datagrid>
+</div>
+
+<div class="clr-example">
+    <h4>Zero Rows + Fixed Height</h4>
+    <p>Datagrid height should be fixed irrespective on the number of rows displayed.</p>
+    <div style="height: 300px">
+        <clr-datagrid style="height: 100%">
+            <clr-dg-column>User ID</clr-dg-column>
+            <clr-dg-column [clrDgField]="'name'">Name</clr-dg-column>
+            <clr-dg-column>Creation date</clr-dg-column>
+            <clr-dg-column>Favorite color</clr-dg-column>
+
+            <clr-dg-row *clrDgItems="let user of zeroUsers">
+                <clr-dg-cell>{{user.id}}</clr-dg-cell>
+                <clr-dg-cell>{{user.name}}</clr-dg-cell>
+                <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+                <clr-dg-cell>
+                    <span class="color-square" [style.backgroundColor]="user.color"></span>
+                </clr-dg-cell>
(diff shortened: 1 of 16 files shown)


src/app/datagrid/test-cases/test-cases.html
```diff
@@ -0,0 +1,210 @@
+<!--
```
youdz: This file is going to be super helpful, thanks a lot. 👍

src/clarity-angular/datagrid/datagrid-placeholder.spec.ts
```diff
@@ -76,20 +57,11 @@ export default function(): void {
...
-
-            // Yuck. But hey, if we're hard-coding a height, we might as well make sure it we don't mess it up.
-            it("adds 36px to its height for each row needed", function() {
-                pageProvider.size = 10;
-                context.detectChanges();
-                expect(context.clarityElement.scrollHeight).toBe(360);
-                itemsProvider.all = new Array(9);
-                context.detectChanges();
-                expect(context.clarityElement.scrollHeight).toBe(36);
+                expect(context.clarityElement.scrollHeight).toBe(1);
```
youdz: This is a question, not a change request: how does this work when you have enough rows to fill the datagrid? Does it still add that 1px under the last one? Isn't that a double border with the footer one?

src/clarity-angular/datagrid/render/main-renderer.ts
```diff
@@ -3,21 +3,37 @@
...
+        private domAdapter: DomAdapter,
+        private el: ElementRef,
+        private renderer: Renderer) {
+            this._subscriptions.push(organizer.computeWidths.subscribe(() => this.computeHeadersWidth()));
+            this._subscriptions.push(organizer.done.subscribe(() => {
+                if (!this._heightSet && this.page.size > 0) {
+                    this.computeDatagridHeight();
+                }
+                this._heightSet = true;
+            }));
```
youdz: I anticipate one issue here: a paginated datagrid where the data arrives asynchronously. You should test this with a `setTimeout()` in a demo. I believe you will end up setting a height that is the one of an empty datagrid.

I think what you want is to compute the height when the page size is set _and_ the items are here.

src/clarity-angular/datagrid/render/main-renderer.ts
```diff
@@ -53,6 +69,23 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, On
...
+     * NOTE: We had to choose to set the height instead of the min-height because
+     * IE 11 requires the height on the parent for the children flex grow/shrink properties to work.
+     * When we used min-height, 1 1 auto doesn't used to work in IE11 :-(
+     * But this doesn't affect the fix. It works in both fixed & variable height datagrids.
+     *
+     * Refer: http://stackoverflow.com/questions/24396205/flex-grow-not-working-in-internet-explorer-11-0
+     */
+    private computeDatagridHeight() {
+        let value: number = this.domAdapter.computedHeight(this.el.nativeElement);
+        this.renderer.setElementStyle(this.el.nativeElement, "height", value + "px");
```
youdz: If you don't mind, I'd rather you set `min-height` for now, instead of `height`.

adityarb88: Just FYI, @youdz and I discussed to keep this as `height` because of the IE11 issues.

src/clarity-angular/datagrid/datagrid-placeholder.spec.ts
adityarb88: Good catch but I don't really have a good solution for this right now. I will add a TODO and keep looking for one.

src/app/datagrid/test-cases/test-cases.html
adityarb88: Added the test cases that we discussed after this review was posted.

src/clarity-angular/datagrid/render/main-renderer.ts
adityarb88: Added the required test cases + also added a test case for when the page size is updated by the user.

src/clarity-angular/datagrid/render/main-renderer.ts
```diff
@@ -84,6 +83,27 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, On
     private computeDatagridHeight() {
         let value: number = this.domAdapter.computedHeight(this.el.nativeElement);
         this.renderer.setElementStyle(this.el.nativeElement, "height", value + "px");
+        this._heightSet = true;
+    }
+
+    private resetDatagridHeight() {
+        this.renderer.setElementStyle(this.el.nativeElement, "height", "auto");
```
youdz: I don't think you want `auto`, because that is not actually neutral. It would override a height given through CSS, for instance. The correct reset is: `null` or the empty string `""`. Both should work fine and actually _remove_ the height property from the style inlined on the element.

adityarb88: fixed!

Issue #743, opened 2017-04-18
Checkbox alignment breaks when used in table TD


brankoleone:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

The checkboxes and their labels should appear aligned vertically on the middle as on the examples when used in the table TDs.

### Actual behavior

When integrating the checkbox component into a table cell the alignment is broken.
The checkbox input and the label text are not aligned vertically.

### Reproduction of behavior

http://plnkr.co/E7b8JQ

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

http://plnkr.co/E7b8JQ

### Environment details

* **Angular version:** 4.0.X

* **Clarity version:** 0.9.0

* **OS and version:** Ubuntu 16.4

* **Browser:** all

reddolan:
Alignment would be preferable. I would also want to make sure that if a cell wraps to multiple lines that we still have alignment with the first line. Text wraps under checkboxes which will be fine

PR #750, opened 2017-04-19
[UI] Add line height to Checkbox and Radio labels

adityarb88:
Fixes: #743

Before:
image

After:
image

Tested on Chrome

EDIT:
Had to update a screenshot because I updated the demo markup too.


src/clarity-angular/forms/utils/_mixins.clarity.scss
```diff
@@ -22,7 +22,7 @@
     width: $height;
 }
 
-@mixin checkbox-radio-label-styles($min-height, $padding-left) {
+@mixin checkbox-radio-label-styles($min-height, $padding-left, $line-height: $clr_baselineRem_1) {
```
hippee-lee: How is `$clr_baselineRem_1 used in the mixin?

adityarb88: Thats the default value in case the user doesn't supply a line-height :-)

hippee-lee: Right, I misread the `:`. Need more ☕️

PR #752, opened 2017-04-19
[NG] Add a resize method on Datagrid

adityarb88:
Adds a `resize` method on the datagrid component for users to be able to manually trigger the resizing the datagrid.

Note that I am using the `done` Observable. It wasn't just added for the test. Its coming with the datagrid placeholder fix anyway :-)

Tested on Chrome.


mattmutt:
thanks for the turnaround. is resize going to be a synchronous activity or broken up into several tasks. Sometimes calculating a resize can't be done in one tick especially when scrollbars and lots of  components have to be interrogated  and themselves resized to optimize the layout.

youdz:
At the moment, it's synchronous. Our resize calculation is mostly the browser natively handling things, so we haven't had any performance issues so far with it.

Note that we can't control how long the browser takes to re-render the resized element, so that one is pretty much out of our control. Still, I'd be surprised if a datagrid has such complex components inside that it takes time to render. Of course, we are not touching or impacting anything outside of the datagrid, so the whole page doesn't re-render or re-compute anything.

PR #760, opened 2017-04-20
[NG] Buttons: allow spinner inside of button

### LINKED ISSUE #463: Feature Request: Spinner Button
```
[ ] bug
[X] feature request
[ ] enhancement
```

### Expected behavior
I would like to have Spinner Button component.

### Actual behavior
Currently Clarity doesn't have the component.

### Use Case
Let's assume that users click on a submit button to get response from backend. And, we want to disable the button for users to prevent resubmission while running and show a proper indication message with Spinner to users like this picture,
image

We can do that with Clarity's Button/Spinner currently provided like this code:
```
<span class="input-group-btn">
     <button class="btn btn-outline" type="submit" [disabled]="groupNumber || isValidating">Validate</button>
</span>
<div *ngIf="isValidating">
     <span class="spinner spinner-inline">
     </span>
     <span>
            Validating...
     </span>
</div>
```

Then, by looking at the implementation of "Loading Data" in recently introduced Tree View, I am now thinking that Spinner inside Button might be more giving consistent Look&Feel to my users instead of disabled Button with spinner-inline message.

image

So would you take a look at this request to provide Spinner Button ng component?

https://hashnode.com/post/ux-where-to-display-a-spinner-on-a-button-ciw12i39115c3ob53cfqkzmwb

victormejia:
Leverages the `[clrLoading]` directive.

Wanted to get some input here. The injected `LoadingListener` in the `Loading` directive seems to be coming in as null: https://github.com/vmware/clarity/blob/master/src/clarity-angular/loading/loading.ts#L16

I'm unsure why, the new `LoadingButton` component here is implementing the `LoadingListener` abstract class. While trying to get it to work, I changed the type from `LoadingListener` to the `LoadingButton` type, and the instance seems to be injected fine. Any thoughts?

Resolves #463

Signed-off-by: Victor Mejia <victor.a.mejia@gmail.com>

### DIFF
--- a/src/app/buttons/button-loading.html
+++ b/src/app/buttons/button-loading.html
@@ -1,0 +1,17 @@
+
+<!--
+  ~ Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h4>Loading Buttons</h4>
+<button [clrLoading]="validateLoading" class="btn btn-info-outline" (click)="validateDemo()">Validate</button>
+<button [clrLoading]="submitLoading" type="submit" class="btn btn-success-outline" (click)="submitDemo()">Submit</button>
+
+<pre>
+    <code clr-code-highlight="language-html">
+    &lt;button [clrLoading]=&quot;validateLoading&quot; class=&quot;btn btn-info-outline&quot;&gt;Validate&lt;/button&gt;
+    &lt;button [clrLoading]=&quot;submitLoading&quot; type=&quot;submit&quot; class=&quot;btn btn-success-outline&quot;&gt;Success&lt;/button&gt;
+    </code>
+</pre>
--- a/src/app/buttons/button-loading.ts
+++ b/src/app/buttons/button-loading.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component} from "@angular/core";
+
+@Component({
+    moduleId: module.id,
+    selector: "clr-buttons-demo-button-loading",
+    templateUrl: "./button-loading.html",
+    styleUrls: ["./buttons.demo.css"]
+})
+export class ButtonLoadingDemo {
+  private validateLoading: boolean = false;
+  private submitLoading: boolean = false;
+
+  validateDemo() {
+    this.validateLoading = true;
+    setTimeout(() => this.validateLoading = false, 1500);
+  }
+
+  submitDemo() {
+    this.submitLoading = true;
+    setTimeout(() => this.submitLoading = false, 1500);
+  }
+}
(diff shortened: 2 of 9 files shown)


src/clarity-angular/button-group/button-loading.ts
```diff
@@ -0,0 +1,30 @@
...
+import {Component} from "@angular/core";
+import {LoadingListener} from "../loading";
+
+@Component({
+    selector: "button[clrLoading]",
+    template: `
+        <span class="spinner spinner-inline" *ngIf="loading"></span>
+        <ng-content></ng-content>
+    `
+})
```
youdz: What you're missing here is `providers: [{provide: LoadingListener, useExisting: LoadingButton}]`.

Angular can't make every component or provider available when requesting either their own class or any interface/abstract class they implement/extend. So you just have to manually register your `LoadingButton` component as a provider for the token `LoadingListener`.

victormejia: Thanks! Learned something new today 😄

victormejia:
Working now, screenshot below:

screen recording 2017-04-21 at 12 04 pm

victormejia:
Working now, screenshot below:

screen recording 2017-04-21 at 12 04 pm

adityarb88:
Thanks for the another contribution @victormejia!

src/clarity-angular/button-group/button-loading.ts
```diff
@@ -0,0 +1,32 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+
+import {Component} from "@angular/core";
+import {LoadingListener} from "../loading/loading-listener";
+
+@Component({
+    selector: "button[clrLoading]",
```
Shijir: Do this `clrLoading` need to be a bit more specific to buttons, like `clrBtnLoading`?

victormejia: We are leveraging the `Loading` directive for this case, which is `[clrLoading]`

src/clarity-angular/button-group/button-loading.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    `
+})
+class TestLoadingButtonComponent {
+    @ViewChild(LoadingButton) loadingButtonInstance: LoadingButton;
+
+    flag: boolean = false;
+}
+
+@Injectable()
+class DummyListener implements LoadingListener {
```
youdz: Why do you still have this? I don't believe you're using it anywhere, you should remove it and its declaration as a provider in the testing module.

src/clarity-angular/button-group/index.ts
```diff
@@ -6,8 +6,10 @@
 import { Type } from "@angular/core";
 import {Button} from "./button";
 import {ButtonGroup} from "./button-group";
+import {LoadingButton} from "./button-loading";
 
 export const BUTTON_GROUP_DIRECTIVES: Type<any>[] = [
     ButtonGroup,
-    Button
+    Button,
+    LoadingButton
```
youdz: This is not really part of the button group, so I'd rather you exported a separate array to keep things flexible.

src/clarity-angular/button-group/all.spec.ts
```diff
@@ -6,10 +6,12 @@
 
 import ButtonGroupSpecs from "./button-group.spec";
 import ButtonSpecs from "./button.spec";
+import LoadingButtonSpecs from "./button-loading.spec";
 import ButtonInGroupServiceSpecs from "./providers/buttonInGroup.service.spec";
 
 describe("Button Group Directives", () => {
     ButtonSpecs();
+    LoadingButtonSpecs();
```
youdz: Same as the array of directives, this new spec should probably go in its own describe, since it's not part of button groups.

victormejia:
Thanks for all the feedback @youdz. Made the changes.

PR #779, opened 2017-04-24
[NG] fixing pageOnLoad call in new wizard

### LINKED ISSUE #765: 0.9 Wizard not firing pageOnLoad when a page is loaded into wizard as the current page
```
[X] bug
[ ] feature request
[ ] enhancement
```

Here is the current behavior:

https://plnkr.co/edit/9OD2nCK80LnZN1lFXAQI?p=preview

Note that pageOnLoad only fires once (during wizard ngOnInit) and never fires again. It should be firing whenever a page is made current.

Here is where the only call is being made...

https://github.com/vmware/clarity/blob/master/src/clarity-angular/wizard/wizard-page.ts#L252

Here is a workaround in the interim:

https://plnkr.co/edit/eiMQddsWHAzrV9MK81Nj?p=preview

mathisscott:
• pageOnLoad was only being emitted on page.makeCurrent() which means we had two different outcomes for making a page current depending upon which method was used.

• this fix unites the two paths to make a page the current page and now pageOnLoad fires no matter how a page is made the current page.

Closes #765

NOTE: Depending on my meeting today with H5 devs, this may need to be released in a 0.9.2-alpha release...

### DIFF
--- a/src/clarity-angular/wizard/providers/wizard-navigation.ts
+++ b/src/clarity-angular/wizard/providers/wizard-navigation.ts
@@ -143,7 +143,7 @@ export class WizardNavigationService implements OnDestroy {
 
     public setCurrentPage(page: WizardPage): void {
         this.currentPage = page;
-// TODO: emit page onload here
+        page.onLoad.emit(page.id);
         this._currentChanged.next(page);
     }
 
--- a/src/clarity-angular/wizard/wizard-page.spec.ts
+++ b/src/clarity-angular/wizard/wizard-page.spec.ts
@@ -55,6 +55,185 @@ class TypescriptTestComponent {
 
 @Component({
     template: `
+        <clr-wizard-page #nav
+            [(clrWizardPageNextDisabled)]="navTestNextDisabled"
+            [(clrWizardPagePreviousDisabled)]="navTestPreviousDisabled"
+            [(clrWizardPagePreventDefaultCancel)]="navStopCancel"
+        >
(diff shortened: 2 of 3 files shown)


src/clarity-angular/wizard/providers/wizard-navigation.ts
```diff
@@ -143,7 +143,7 @@ export class WizardNavigationService implements OnDestroy {
 
     public setCurrentPage(page: WizardPage): void {
         this.currentPage = page;
-// TODO: emit page onload here
+        page.onLoad.emit(page.id);
```
hippee-lee: One thing here is the naming `onLoad`. How hard would it be to rename it to something related to what its doing - `makeCurrent` ? 

- I'm assuming `setCurrentPage` is called many times and to me I would think that `onLoad` is only called once.

mathisscott: `makeCurrent` is already a method on `WizardPage`. `onLoad` is an event emitter. The naming of the emitter has been carried over from 0.8 to minimize breaking changes/learning curve for users migrating from one to the other.

`setCurrentPage()` here is the final step in making a page the current page. All roads to making a page the current page (via the `Wizard`, `WizardPage`, or `WizardNavigationService`) wind up here. Per the legacy emitter, `onLoad` is emitted when a page becomes the current page.

Hopefully, that clarifies the architecture.

PR #796, opened 2017-04-26
[NG] adding alt-next feature to wizard

mathisscott:
• installer team needed this
• needs some unit testing but works and doesn’t break any existing tests
• next step is adding dox to website

### DIFF
--- a/src/clarity-angular/wizard/providers/wizard-navigation.ts
+++ b/src/clarity-angular/wizard/providers/wizard-navigation.ts
@@ -37,55 +37,15 @@ export class WizardNavigationService implements OnDestroy {
         });
 
         this.nextButtonSubscription = this.buttonService.nextBtnClicked.subscribe(() => {
-            let currentPage = this.currentPage;
-            if (!this.currentPageIsLast && currentPage.readyToComplete) {
-                currentPage.primaryButtonClicked.emit("next");
-                currentPage.nextButtonClicked.emit(currentPage);
-                if (currentPage.preventDefault) {
-                    this.pageCollection.commitPage(currentPage);
-                } else {
-                    this.next();
-                }
-            }
+            this.checkAndCommitCurrentPage("next");
         });
 
         this.dangerButtonSubscription = this.buttonService.dangerBtnClicked.subscribe(() => {
-            let currentPage: WizardPage = this.currentPage;
-
-            if (!currentPage.readyToComplete) {
-                return;
-            }
-
-            // because we made it past the guard up there ^ we know the current page is ready
-            if (this.currentPageIsLast) {
-                currentPage.primaryButtonClicked.emit("danger");
-                currentPage.dangerButtonClicked.emit(currentPage);
-                if (currentPage.preventDefault) {
-                    this.pageCollection.commitPage(currentPage);
-                } else {
-                    this.finish();
-                }
-            } else {
-                currentPage.primaryButtonClicked.emit("danger");
-                currentPage.dangerButtonClicked.emit(currentPage);
-                if (currentPage.preventDefault) {
-                    this.pageCollection.commitPage(currentPage);
-                } else {
-                    this.next();
-                }
-            }
+            this.checkAndCommitCurrentPage("danger");
         });
 
         this.finishButtonSubscription = this.buttonService.finishBtnClicked.subscribe(() => {
-            let currentPage = this.currentPage;
-
-            if (currentPage.readyToComplete && this.currentPageIsLast) {
-                if (currentPage.preventDefault) {
-                    this.pageCollection.commitPage(currentPage);
-                } else {
-                    this.finish();
-                }
-            }
+            this.checkAndCommitCurrentPage("finish");
         });
 
         this.customButtonSubscription = this.buttonService.customBtnClicked.subscribe((type: string) => {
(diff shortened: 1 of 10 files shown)


src/clarity-angular/wizard/providers/wizard-navigation.ts
```diff
@@ -164,50 +125,100 @@ export class WizardNavigationService implements OnDestroy {
...
+            // this is a state that alt next flows can get themselves in...
+            this.pageCollection.commitPage(currentPage);
+        }
+        this.setCurrentPage(nextPage);
+        // SPECME
     }
 
-    public finish(): void {
-        let currentPage = this.currentPage;
+    public checkAndCommitCurrentPage(buttonType: string): void {
```
Shijir: It's a nit, but it would be easier to read if we keep these early exits at the top? 
```
        if (!this.currentPage.readyToComplete) {
            return;
        }
        if (isFinish && ! this.currentPageIsLast) {
            return;
        }
        let currentPage: WizardPage = this.currentPage;
        let iAmTheLastPage: boolean = this.currentPageIsLast;
        let isNext: boolean =  buttonType === "next";
        let isDanger: boolean = buttonType === "danger";
        let isDangerNext: boolean = isDanger && !iAmTheLastPage;
        let isDangerFinish: boolean =  isDanger && iAmTheLastPage;
        let isFinish: boolean =  buttonType === "finish" || isDangerFinish;

```

src/clarity-angular/wizard/wizard-page.ts
```diff
@@ -119,6 +119,25 @@ export class WizardPage implements OnInit {
...
 
+    // overrides next from the page level, so you can use an alternate function for
+    // validation or data-munging with clrWizardPageOnNext
+    private _stopNext = false;
+    public get stopNext(): boolean {
+        return this._stopNext;
+    }
+    @Input("clrWizardPagePreventDefaultNext")
+    public set stopNext(val: boolean) {
+        let valBool = !!val;
```
Shijir: Do we need `let valBool = !!val;` if `val` is a already boolean parameter as `!!val` will have the same value as the `val` parameter, right?

src/clarity-angular/wizard/wizard.ts
```diff
@@ -124,6 +126,8 @@ export class Wizard implements OnInit, OnDestroy, AfterContentInit, DoCheck {
     @Output("clrWizardOnPrevious") onMovePrevious: EventEmitter<any> =
         new EventEmitter<any>(false);
 
+    @Input("clrWizardPreventDefaultNext") stopNext: boolean = false;
+
     @Input("clrWizardPreventDefaultCancel") stopCancel: boolean = false;
 
     @Input("clrWizardPreventModalAnimation") _stopModalAnimations: boolean = false;
```
Shijir: Don't we use underscore only on private properties?

src/clarity-angular/wizard/providers/wizard-navigation.ts
mathisscott: more of a stylistic question. i usually declare variables at the top and then do early returns.

src/clarity-angular/wizard/wizard-page.ts
mathisscott: This code is carried over from similar implementations in the datagrid. I presume the `!!val` is a guard against a non-boolean being sent by accident.

src/clarity-angular/wizard/wizard.ts
mathisscott: stopModalAnimations is given an underscore to denote that it should only be used internally (privately). The wizard needs it to make the inline/static wizard. It's how it tells the modal not to run animations.

src/clarity-angular/wizard/providers/wizard-navigation.ts
```diff
@@ -164,50 +125,100 @@ export class WizardNavigationService implements OnDestroy {
...
-        currentPage.primaryButtonClicked.emit("finish");
-        currentPage.finishButtonClicked.emit(currentPage);
-        if (!currentPage.preventDefault) {
-            // when preventDefault is enacted, this is called manually
-            // after an onCommit. put check in here to avoid duplicate
-            // onCommit calls
-            this.pageCollection.commitPage(currentPage);
+        iAmTheLastPage = this.currentPageIsLast;
+
+        if (isFinish && !iAmTheLastPage) {
```
youdz: I don't understand this `if`. `isFinish` was declared 5 lines above, left undefined, and nothing in these 5 lines modifies it. So this test can't ever true, right?

src/clarity-angular/wizard/wizard-page.ts
youdz: We could receive `null` or `undefined` at runtime.

src/clarity-angular/wizard/wizard-page.ts
```diff
@@ -119,6 +119,25 @@ export class WizardPage implements OnInit {
...
+        if (valBool !== this._stopNext) {
+            this._stopNext = valBool;
+            this.stopNextChange.emit(valBool);
+        }
+        // SPECME
+    }
+
+    @Output("clrWizardPagePreventDefaultNextChange") stopNextChange: EventEmitter <boolean> =
+        new EventEmitter();
```
youdz: This output doesn't make sense to me. What's a use case that would need it? From what I can see, the only way this is emitted is if the user gave us the corresponding input, which is a case where we _don't_ want to emit anything to avoid loops.

mathisscott: Yeah. This is a carry over from legacy defaultCancel code. It doesn't make sense to me either.

src/clarity-angular/wizard/providers/wizard-navigation.ts
mathisscott: You are correct. I need to fix this.

PR #798, opened 2017-04-26
[NG] Adds Hide/Unhide capability for columns in Datagrid

### LINKED ISSUE #401: Hide Unhide capability for Columns in Clarity DataGrid
```
[ ] bug
[x ] feature request
[ ] enhancement
```

### Expected behavior
As part of Clarity DataGrid currently we do not have a capability to hide few columns by default and if user wishes to unhide them user could do it as per their preference.

Hide/Unhide feature is extensively used in our existing UI projects.

image
image

hippee-lee:
- implements show/hide behavior for both columns and cells
- reuses column templates in the toggle component
- (conditionally) adds the toggle widget to the footer when hideable columns are present
- add the `*clrDgHideableColumn` directive
- closes #401 

Tested in: Chrome, IE11, Edge, Safari & Firefox

## Toggle UI
hide-show-ui

## Select All is disabled when they are all selected
hide-show-select-all-disabled

## Last showing column is disabled
hide-show-last-col-disabled

## Hideable Columns (and Cells) can be used with `*ngIf`
hide-show-no-id-date


src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -0,0 +1,87 @@
+<!--
+  ~ Copyright (c) 2016 VMware, Inc. All Rights Reserved.
```
mathisscott: Not a big deal for this PR but in the future we are doing copyright 2016-2017 on files...

src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -0,0 +1,87 @@
...
+
+    <clr-dg-footer>
+        {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}}
+        of {{pagination.totalItems}} users
+        <clr-dg-pagination #pagination [clrDgPageSize]="currentPageSize"></clr-dg-pagination>
+    </clr-dg-footer>
+</clr-datagrid>
+
+<h2>Datagrid: simple demo</h2>
+<!--<clr-datagrid>-->
```
mathisscott: A lot is commented out here. On purpose?

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -584,6 +597,69 @@
...
+
+                display: flex;
+                flex-direction: column;
+
+                .switch-header {
+                    label {
+                        font-weight: clr-getTypePropertyValueForDomElement(datagrid_column_toggle_title, font-weight);
+                        font-size: clr-getTypePropertyValueForDomElement(datagrid_column_toggle_title, font-size);
+                    }
+                    padding-bottom: 12px;
```
mathisscott: Maybe a note here that this should be `$clr_baselineRem_0_5` instead of 12px?

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -584,6 +597,69 @@
...
+                }
+
+                .switch-content {
+                    max-height: 300px;
+                    overflow-y: auto;
+                    min-height: 25px; // prevents scrollbars in IE.
+
+                    li {
+                        line-height: 24px;
```
mathisscott: `$clr_baselineRem_1`

src/clarity-angular/datagrid/datagrid-cell.ts
```diff
@@ -3,16 +3,56 @@
...
+     *
+     * @type {string}
+     */
+    private _id: string;
+
+    /**
+     * @property hidden
+     *
+     * @description
+     * Property used to apply a css class to this cell that hides it when hidden = true.
```
mathisscott: Love the use of JSDocs here. I wish we could start doing this throughout our code.

src/clarity-icons/shapes/core-shapes.ts
```diff
@@ -409,8 +409,16 @@ const coreShapes: any = {
...
-        `
+        `,
+
+    "view-columns": `
+            <svg version="1.1" viewBox="0 0 36 36" preserveAspectRatio="xMidYMid meet"
+                xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+                <title>view-columns</title>
+                <path class="clr-i-outline clr-i-outline-path-1" d="M31,5H5A2,2,0,0,0,3,7V29a2,2,0,0,0,2,2H31a2,2,0,0,0,2-2V7A2,2,0,0,0,31,5ZM13,29H5V7h8Zm10,0H15V7h8Z"/>
+            </svg>
+        `,
```
mathisscott: Trailing comma here?

src/app/datagrid/hide-show-columns/hide-show.html
hippee-lee: Good catch - that one slipped through. Will correct.

src/app/datagrid/hide-show-columns/hide-show.html
hippee-lee: Totally on purpose aren't comments removed when we compile? har har - will remove.

src/clarity-angular/datagrid/_datagrid.clarity.scss
hippee-lee: Thanks for catching that - will update.

src/clarity-angular/datagrid/_datagrid.clarity.scss
hippee-lee: Will change it.

src/clarity-icons/shapes/core-shapes.ts
hippee-lee: Will remove.

Jinnie:
Hi Clarity 😄 
Won't the selected-items icon be placed in the same bottom-left corner, once the selection-preservation feature is complete? Won't it get a bit crowded there?
Most datagrids I've seen that offer column hide/unhide capability have the columns settings icon on the top-right corner of the grid, in the end of the header line. That's where my expectation to find such a menu is, as it defines columns by name, which names are found in the header.
Just curious why did you prefer the bottom-left...

mathisscott:
@Jinnie : I believe @hippee-lee and @reddolan have already hashed this out.

@reddolan or @hippee-lee : Would either of you be willing to share the ultimate verdict on that?

hippee-lee:
Hi @Jinnie,

We integrated them in the bottom left of the datagrid footer. They work together or separately since one may or may not have hideable columns. As far as locating the toggle component in the top left - i'm not sure we have space on top. The top right is part of the last column which might have filters on it. 

screen shot 2017-04-27 at 1 01 28 pm
screen shot 2017-04-27 at 1 03 01 pm

Jinnie:
@hippee-lee, thanks for the explanation. It's your choice, of course, and I respect that. Just sharing my feedback, with the idea to provoke a discussion, or restart an old one if you already had it 😈. Because my perception for the footer of a component is more of a _status bar_, and less of a place where you can make settings. 
I don't think you've not already evaluated the different possibilities, though, so if you have a complete agreement on the footer choice - 👍

hippee-lee:
No worries @Jinnie - I appreciate the feedback and I think its a valid point. The question for me is, given the current datagrid, where else could the button for the column toggle go?

src/clarity-angular/datagrid/providers/hideable-column.service.ts
```diff
@@ -0,0 +1,192 @@
...
+     */
+    public get checkForAllColumnsVisible(): boolean {
+        let allColumns = this._columnList.filter(col => col);
+        let hiddenColumns = this._columnList.filter(col => {
+            if ( col && !col.hidden ) {
+                return col;
+            }
+        });
+        return ( allColumns.length === hiddenColumns.length );
+    }
```
youdz: Fun tip:
```typescript
public get checkForAllColumnsVisible(): boolean {
    return !this._columnList.some(col => col && col.hidden);
}
```

src/clarity-angular/datagrid/providers/hideable-column.service.ts
```diff
@@ -0,0 +1,192 @@
...
+     * @description
+     * Creates an array of DatagridHideableColumn's || null based column array passed as param.
+     * Is dependent on the order in @ContentChildren in Datagrid.
+     *
+     * @param columns
+     *
+     */
+    public updateColumnList( columns: DatagridHideableColumn[] ) {
+        this._columnList.length = 0; // clear the list
+        columns.forEach(col => this._columnList.push(col)); // Update it with the new
```
youdz: Why not just `this._columnList = columns`? Do you really need to manually clone the array?

src/clarity-angular/datagrid/providers/hideable-column.service.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+export default function (): void {
+    describe("DatagridHideableColumn provider", function () {
+
+        let columnService: HideableColumnService;
+        let filtersService: FiltersProvider;
+        let sortService: Sort;
+
+        beforeEach(function () {
+            sortService = new Sort();
+            filtersService = new FiltersProvider();
```
youdz: You're not using Sort or Filter here, you can remove them.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -584,6 +597,79 @@
             margin-left: $clr_baselineRem_1_5;
             height: $clr_baselineRem_1_5 - 2px;
         }
+
+        .column-switch-wrapper {
+            // Has a possible sibling in the .datagrid-foot-select.checkbox container
+            flex: 1 1 100%;
```
hippee-lee: Correct. Will change to $clr_baselineRem_1.

src/clarity-angular/datagrid/datagrid-cell.spec.ts
```diff
@@ -3,31 +3,33 @@
...
-        it("projects content", function() {
-            expect(context.clarityElement.textContent.trim()).toMatch("Hello world");
+        it ("projects content", function () {
+            expect (context.clarityElement.textContent.trim ()).toMatch ("Hello world");
         });
 
-        it("adds the .datagrid-cell class to the host", function() {
-            expect(context.clarityElement.classList.contains("datagrid-cell")).toBeTruthy();
+        it ("adds the .datagrid-cell class to the host", function () {
+            expect (context.clarityElement.classList.contains ("datagrid-cell")).toBeTruthy ();
```
hippee-lee: Sorry, IntelliJ was mis-configured for `Reformat Code` and I didn't catch this file when I cleaned up. Will fix.

src/clarity-angular/datagrid/datagrid-cell.ts
```diff
@@ -3,16 +3,56 @@
...
 
-@Component({
+@Component ( {
     selector: "clr-dg-cell",
     template: `
         <ng-content></ng-content>
     `,
     host: {
         "[class.datagrid-cell]": "true",
+        "[class.datagrid-cell--hidden]": "hidden"
```
hippee-lee: Will add the test.

src/clarity-angular/datagrid/datagrid-cell.ts
```diff
@@ -3,16 +3,56 @@
...
+     *
+     * @param val
+     */
+    public set id ( val: string ) {
+        this._id = val;
+    }
+
+    public get id (): string {
+        return this._id;
+    }
```
hippee-lee: Gotcha, will remove.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -4,32 +4,44 @@
...
+import { DatagridHideableColumn } from "./datagrid-hideable-column";
+
+let nbCount: number = 0;
+
 
 @Component({
     selector: "clr-dg-column",
     template: `
-        <div class="datagrid-column-flex">
+        <div class="datagrid-column-flex" #hideableColumn>
```
hippee-lee: Yep, leftover for initial prototype / exploration of the solution.
Will remove.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -67,8 +82,52 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
...
+     *
+     * @description
+     * A property that allows the column to be hidden / shown with css
+     * Note the default allows the DatagridColumn to have an *ngIf on it. (EHCAIWC - will occur if its not initialized)
+     *
+     * @default false
+     *
+     * @type boolean
+     */
+    public hidden: boolean = false;
```
hippee-lee: yup. Will change.

src/clarity-angular/datagrid/datagrid-hideable-column.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(changeValue).toBeUndefined();
+            expect(nbChanges).toBe(0);
+        });
+
+    });
+}
+
+@Component({
+    template: `
+        <clr-dg-column>Test Column</clr-dg-column>
```
hippee-lee: Will add a comment.

src/clarity-angular/datagrid/render/row-renderer.ts
```diff
@@ -3,44 +3,132 @@
...
 
+    /*****
+     * property dgCells
+     *
+     * @description
+     * A Query List of the Datagrid cells in this row.
+     *
+     * @type QueryList<DatagridCell>
+     */
+    @ContentChildren(DatagridCell) dgCells: QueryList<DatagridCell>;
```
hippee-lee: Moved it to the Row. For some reason I started in RowRenderer early on and things just stayed there.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -0,0 +1,123 @@
...
+                #anchor 
+                (click)="toggleUI()"
+                class="btn btn-sm btn-link column-toggle--action">
+            <clr-icon shape="view-columns"></clr-icon>
+        </button>
+        <div class="column-switch"
+             *clrPopover="open; anchor: anchor; anchorPoint: anchorPoint; popoverPoint: popoverPoint">
+            <div class="switch-header">
+                    Show Columns
+                    <button style="min-width: 0; margin: 0; padding: 0"
```
hippee-lee: Yep - this is actually cleanup I forgot about. Styles are already in the scss file.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -0,0 +1,123 @@
...
+                            class="btn btn-sm btn-link"
+                            (click)="toggleUI()">
+                        <clr-icon
+                                shape="close"></clr-icon>
+                    </button>
+            </div>
+            <ul *ngIf="open" class="switch-content list-unstyled">
```
hippee-lee: Appendix for the initial prototype before popover made it into my branch. 
Will remove.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -0,0 +1,123 @@
...
+            </ul>
+            <div class="switch-footer">
+                <div>
+                    <button
+                            class="btn btn-sm btn-link p6 text-uppercase"
+                            [disabled]="allColumnsVisible"
+                            (click)="selectAll()">Select All
+                    </button>
+                </div>
+                <div style="display: flex; justify-content: flex-end">
```
hippee-lee: Removed.

src/clarity-angular/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+ */
+import { Component, QueryList, TemplateRef, ViewChildren } from "@angular/core";
+import { TestContext } from "./helpers.spec";
+import { HideableColumnService } from "./providers/hideable-column.service";
+import { DatagridColumnToggle } from "./datagrid-column-toggle";
+import { DatagridHideableColumn } from "./datagrid-hideable-column";
+
+export default function (): void {
+    describe("Datagrid Column Toggle component", function () {
+        describe("Typescript API", function () {
```
Shijir: Do we test whether the same column keeps its previous state such as style, classes etc when it gets hidden and redisplayed again? Because in the app demo, when I use toggle buttons to hide and open the same column, their strict width classes are lost.

Jinnie:
Something like this, I guess:
column-toggle

Yes, it eats part of the last column's header (header only, the last cells in the rows go to the end). When you first said that it may interfere with the filter icon, I totally agreed, but after seeing the edited screenshot, I don't think it's really a problem - the filtering icon has a generous margin on the right.

I'm not saying this is the best solution, maybe I am biased because some Angular 1 datagrids I've used had this icon there and my "expectations" are modelled after this experience. 😃 

The strongest point in defence of this solution is that you have the columns settings next to the column names, not on the opposite side of the datagrid. If the grid is not bound to the viewport size, then this toggle may even go offscreen.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    ngOnInit() {
+        this.hideableColumnService.columnListChange.subscribe(( columnList ) => {
+            // Reset the list of columns
+            this.columns.length = 0;
+
+            // Add only the hidden columns to the toggler.
+            columnList.forEach(( col ) => {
+                if ( col ) {
+                    col.lastVisibleColumn = this.hideableColumnService.canHideNextColumn;
```
hippee-lee: Correct, we don't need it here, or below.

src/clarity-angular/datagrid/providers/hideable-column.service.ts
hippee-lee: Nope, don't need to clone here. 
Will change.

src/clarity-angular/datagrid/providers/hideable-column.service.ts
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @return void
+     *
+     */
+    public updateForLastVisibleColumn(): void {
+        let visibleCount: number = this._columnList.length - this._columnList.filter(col => {
+                if ( col && col.hidden ) {
+                    return col;
+                }
+            }).length;
```
hippee-lee: Will remove.

src/clarity-angular/datagrid/providers/hideable-column.service.spec.ts
hippee-lee: Will Remove.

src/clarity-angular/datagrid/datagrid-column-toggle.spec.ts
hippee-lee: I'm not testing for the draggable classes yet. Is that something that should go into this release?

src/clarity-angular/datagrid/datagrid-hideable-column.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * @property hiddenChanges
+     *
+     * @description
+     * A stream of state changes an instance of DatagridHideableColumn will broadcast to subscribers.
+     *
+     * @type {Subject<boolean>}
+     */
+    private hiddenChanges: Subject<boolean> = new Subject<boolean>();
```
hippee-lee: I'll change the Subject to `hiddenChangesState` and the Observable to `hiddenChangeState`.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -338,7 +346,12 @@
         .datagrid-foot-select.checkbox {
             display: block;
             line-height: inherit;
-            flex: 1 1 100%;
+            flex: 1 1 34px;
```
hippee-lee: Will update to use the variable.

src/clarity-angular/datagrid/datagrid-column-toggle.spec.ts
hippee-lee: fyi - it is preserving the widths when I hide/show after dragging tho ... 
out

src/clarity-angular/datagrid/datagrid-hidable-column.directive.ts
```diff
@@ -0,0 +1,121 @@
...
+     * *clrDgHideableColumn
+     * *clrDgHideableColumn={hidden: false}
+     * *clrDgHideableColumn={hidden: true}
+     *
+     * @param value
+     *
+     */
+    @Input("clrDgHideableColumn")
+    set clrDgHideableColumn( value: any ) {
+        this._hidden = (value) ? value.hidden : false;
```
hippee-lee: Good catch. Will update.

hippee-lee: Wondering about throwing an error though. Do we want to throw an error or have a simple default (`false`) ?

Shijir: I am not sure how a default (`false`) could tackle the error that could arise when a user passes an object with a property that has a different name than `hidden`. But do we have to ask an object as a parameter here instead of asking just a boolean value so we could avoid all these checkings?  If the value parameter has to be an object, we could also use a simple interface on the value parameter?

src/clarity-angular/datagrid/datagrid-hideable-column.ts
Shijir: `hiddenChanges` suggests to me that changes were hidden, but I believe it's a property that notifies the state of hideable columns. But It shouldn't block this PR at all as this is so minor nit.

src/clarity-angular/datagrid/datagrid-column-toggle.spec.ts
Shijir: This was happening when I hide and unhide the `User ID` column by clicking on that `TOGGLE ID` button.

src/clarity-angular/datagrid/datagrid-hidable-column.directive.ts
hippee-lee: I discussed this with @youdz during development. We opted for the more explicit form: `*clrDgHideableColumn="{hidden: value}"` in order to make it clear what we being set. 
The reason being questions that arise with just a value as the directive param: 
`*clrDgHideableColumn="true" - is that now a hideable column, what is it's state hidden or showing? 

@youdz - Should we revisit this and make it simpler / less explicit before it gets released into the wild?

src/clarity-angular/datagrid/datagrid-column-toggle.spec.ts
hippee-lee: Ah, yes, thats using `*ngIf` so the cells are destroyed and recreated. This is still there so I can make sure it works but isn;t part of the feature set to be documented. I'm not sure if it will be a recommended pattern yet either.

Shijir: Ah make sense.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -74,46 +75,48 @@ export class DatagridColumnToggle implements OnInit, OnDestroy {
...
+            this.hideableColumnService.updateForLastVisibleColumn();
+            this.allColumnsVisible = this.hideableColumnService.checkForAllColumnsVisible;
 
             // Add only the hidden columns to the toggler.
             columnList.forEach(( col ) => {
                 if ( col ) {
-                    col.lastVisibleColumn = this.hideableColumnService.canHideNextColumn;
-                    this.allColumnsVisible = this.hideableColumnService.checkForAllColumnsVisible;
-
                     this.columns.push(col);
```
youdz: So this could now become `this.columns = columnList.filter(col => !!col);`. 😉 
It doesn't matter either way, just pointing out a little extra cleanup step we can take.

src/clarity-angular/datagrid/datagrid-column-toggle.ts
```diff
@@ -74,46 +75,48 @@ export class DatagridColumnToggle implements OnInit, OnDestroy {
...
-                    );
                 }
             });
         });
     }
 
     ngOnDestroy() {
-        this._hideableColumnChangeSubscriptions.forEach(sub => sub.unsubscribe());
+        // this._hideableColumnChangeSubscription.unsubscribe();
     }
```
youdz: Any reason this is commented? I'm guessing it was to test, but we should put it back.

src/clarity-angular/datagrid/datagrid-hidable-column.directive.ts
youdz: I think we'll go with this for now. Structural directives inputs are not great right now, so we need to make some compromises.

Maybe if https://github.com/angular/angular/issues/15067 we can look for something simpler. And I can see ways for us to progressively deprecate the current explicit form as to avoid breaking changes.

PR #803, opened 2017-04-27
[core] fixing karma timeout error in travis (#666)

### LINKED ISSUE #666: [core] tests are throwing error in Travis
```
[ ] bug
[ ] feature request
[x] enhancement
```

Unit tests are throwing error sometimes (though no test failures), but only in Travis. Need to investigate.

### Related posts 
http://stackoverflow.com/questions/24119506/karma-jasmine-times-out-without-running-tests
http://stackoverflow.com/questions/27310184/warn-phantomjs-1-9-8-mac-os-x-disconnected-1-times-because-no-message-in
https://github.com/karma-runner/karma/issues/598
https://github.com/karma-runner/karma-phantomjs-launcher/issues/126

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

jeeyun:
We are moving away from PhantomJS as the memory leak crashes Travis, and the [maintainer stepped down from the project](https://groups.google.com/forum/#!msg/phantomjs/9aI5d-LDuNE/5Z3SMZrqAQAJ). We are now using headless Chrome.  This change assumes that the development machine has Chrome Canary installed. I'd prefer to keep the browser for dev/build server the same (that is, both Chrome). This may not be flexible enough though depending on what we assume about the development machine. Open to feedback on that.

I also had to change up the popover test, as the subsequent calls to anchor the popover only needed transform values relative to position set by the previous call. That was not the case with phantomJS. This issue is fixed by setting the transform to none before each call. I also cleaned up the expected value to be calculated based on variables. I'm not 10% confident all the math is right. It tests for the same values as before and passes the tests.

Fixes #666.


jeeyun:
Thanks @mathisscott! Yes, I also think it should go in 0.9.3.

Issue #360, opened 2017-01-12
Pagination links state is confusing


ssanghavi:
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Once user clicks page# 2 on data-grid pagination footer, only page# 2 link should appear as selected.

### Actual behavior
Page# 2 link has a border at bottom to make it look like selected. But at the same time Page# 3 link is showing outline which appears as currently active page.
Please see attached screenshot to understand it better.

### Reproduction of behavior
Did not create separate plunkr because screenshot is taken from clarity documentaion/samples.

* **Browser:** [Chrome 55]
On Firefox it seem to work fine.

ssanghavi:
![screen shot 2017-01-12 at 2 53 25 pm](https://cloud.githubusercontent.com/assets/5438020/21884206/5de0c94a-d8d8-11e6-8a86-51a9143b9a36.png)

mathisscott:
The blue outline is the browser highlighting the "3" because it has focus. Did you hit tab when page 2 loaded or does it just load in like that?

youdz:
I think this happens without hitting tab. Because of the structure of our pagination template, I believe the "clicked" link becomes the next page naturally, without being destroyed, so the browser keeps focus on it even though angular modifies its content.

Pretty interesting, not sure if the best fix is to move to `*ngFor` to better track pagination links, or to start hacking with native elements focus. Probably the former.

ssanghavi:
As @youdz already mentioned this happens without hitting tab.

youdz:
I'm marking this as `contributions-welcome`.

If anyone wants to tackle this, I would like them to use `*ngFor` on the 3 pages centered on the current one: `current - 1`, `current` and `current + 1`. The links for the first and last page are fine the way they are. This should properly destroy and recreate page links, making them lose focus.

PR #806, opened 2017-04-27
Fix for #360

SrinivasPadidapelli:
Hi,
I have removed outline using css and tested. It’s working fine. Can you
please review?

Signed-off-by: srinivas padidapelli srinuwd@gmail.com

### DIFF
--- a/src/clarity-angular/datagrid/_datagrid.clarity.scss
+++ b/src/clarity-angular/datagrid/_datagrid.clarity.scss
@@ -638,6 +638,7 @@
             color: $gray-dark;
             // FIXME: this should be in the general reboot
             cursor: pointer;
+            outline: none;
         }
     }
 


adityarb88:
@SrinivasPadidapelli Please amend your commit to correct the sign off message. You need to make sure that the format is like this: `Signed-off-by: John Doe <john.doe@email.org>`. Note that your email must be wrapped by `<>`.

src/clarity-angular/datagrid/_datagrid.clarity.scss
```diff
@@ -638,6 +638,7 @@
             color: $gray-dark;
             // FIXME: this should be in the general reboot
             cursor: pointer;
+            outline: none;
```
adityarb88: I don't think we can use `outline: none` because then the pagination links become inaccessible via keyboard.

SrinivasPadidapelli: HI Aditya,
I am tested in chrome it is accessible via keyboard also. can you please test and let me know. If its not working. i will close the PR.
Thanks,
Srinivas P

Issue #781, opened 2017-04-24
Datagrid: Undefined [clrDgSortOrder] overrides value on previous columns


jarinwhite:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Datagrid should presumably ignore `[clrDgSortOrder]` when undefined so that sorting can be conditionally applied.

### Actual behavior
If `[clrDgSortOrder]` is defined on a column and the value of `[clrDgSortOrder]` on a sequential column is falsy, the value of `[clrDgSortOrder]` on prior columns is ignored.

### Reproduction of behavior
https://plnkr.co/edit/A77ndTiWhJAfk8JWpzX0?p=preview (Sort on 'ID' is overridden by later columns even though `[clrDgSortOrder]` is undefined)

https://plnkr.co/edit/GLAXWc8kOHGQzgsogiSv?p=preview (Sorts properly on 'Age' column because it is last in line)

### Environment details

* **Angular version:** 4.0.1

* **Clarity version:** 0.9.1

* **OS and version:** Windows 10

* **Browser:** Chrome 58 | Firefox 56

mathisscott:
@Shijir : are you working on this? if so, could you assign it to yourself?

Issue #771, opened 2017-04-21
Datagrid: Default to DatagridPropertyComparator when [clrDgSortBy] is undefined


jarinwhite:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior
Unable to use [clrDgSortOrder] to specify default sort on columns unless [clrDgSortBy] is also defined. In many cases it's useful to just fall back to the default comparator. Currently I have to import DatagridPropertyComparator and pass it to [clrDgSortBy] manually. If not a fallback, could we have an easier way to specify that the default comparator should be used?

### Actual behavior
Does not sort unless [clrDgSortBy] is specified

### Reproduction of behavior
https://plnkr.co/edit/wUpQqjhf1Jcpgaa5DZI5?p=preview (No clrDgSortBy, does not apply sort)

https://plnkr.co/edit/V7dytjr0PN1weaRs4SLc?p=preview (Passed DatagridPropertyComparator instance, applies sort correctly)

### Environment details

* **Angular version:** 4.0.1

* **Clarity version:** 0.9.1

* **OS and version:** 

* **Browser:** Chrome 58

youdz:
`[clrDgField]` already sets the default comparator, that's the whole point of it. You should not have to specify `[clrDgSortBy]` in addition to it.
 So I think this is actually a bug where the sort order input is processed before the field input. Thanks for raising this.

PR #809, opened 2017-04-27
[NG] Fixes for #771 and #781

Shijir:
- Needed to define `[clrDgField]` input property before `[clrDgSortOrder]`
- Guarded against using undefined for Sort comparator
- Used "pointer" for the sortable column header hover cursor

### DIFF
--- a/src/clarity-angular/datagrid/datagrid-column.ts
+++ b/src/clarity-angular/datagrid/datagrid-column.ts
@@ -86,6 +86,28 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
         this._sortSubscription.unsubscribe();
     }
 
+    /*
+     * Simple object property shortcut, activates both sorting and filtering
+     * based on native comparison of the specified property on the items.
+     */
+    private _field: string;
+    public get field() {
+        return this._field;
+    }
+
+    @Input("clrDgField")
+    public set field(field: string) {
+        if (typeof field === "string") {
+            this._field = field;
+            if (!this.customFilter) {
+                this.setFilter(new DatagridStringFilterImpl(new DatagridPropertyStringFilter(field)));
+            }
+            if (!this.sortBy) {
+                this.sortBy = new DatagridPropertyComparator(field);
+            }
+        }
+    }
+
     /**
      * Comparator to use when sorting the column
      */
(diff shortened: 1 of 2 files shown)


Shijir:
@adityarb88 they were closely related Datagrid sorting issues.

src/clarity-angular/datagrid/datagrid-column.ts
```diff
@@ -136,7 +158,6 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
     public get sortOrder() {
         return this._sortOrder;
     }
-
     @Input("clrDgSortOrder")
     public set sortOrder(value: SortOrder) {
         // only if the incoming order is different from the current one
```
youdz: The correct fix to #781  is here:
```javascript
if (typeof value === undefined) { return; }
```

src/clarity-angular/datagrid/providers/sort.ts
```diff
@@ -18,9 +18,12 @@ export class Sort {
...
         return this._comparator;
     }
+
     public set comparator(value: Comparator<any>) {
-        this._comparator = value;
-        this.emitChange();
+        if (typeof value !== "undefined") {
+            this._comparator = value;
+            this.emitChange();
+        }
```
youdz: ... not here.

If you're handling undefined user input, do it directly at the input level, not at the deep internal provider level. This change has too many hidden effects (that might not matter right now) for what it solves. The surface solution is safer.

src/clarity-angular/datagrid/datagrid-column.ts
mathisscott: Only catch is `typeof` gives you a string. So `typeof value === "undefined"`...

youdz: I don't know what you mean. 😁

mathisscott: ROFLMAO

Well played. :1st_place_medal:

src/clarity-angular/datagrid/providers/sort.ts
Shijir: @youdz thanks for the review here. This one was the one I was also apprehensive about so I was waiting until I get your review here 😛 Will fix it now.

PR #829, opened 2017-05-02
[NG] allowing numeric ids for wizard pages

mathisscott:
• this is partly a bug and partly an enhancement.
• some people were using array indices as their ids but WizardPage was only set up to handle strings being passed to it. this fixes that.
• unit tests too! because TDD, yo!


imcotton:
Maybe have a look over #830 first?

PR #830, opened 2017-05-02
[NG] default wizard page id

### LINKED ISSUE #783: Wizard: page ids are off
```
[X] bug
[ ] feature request
[ ] enhancement
```
This is not a critical issue because the wizard doesn't really use the ids for anything but it could affect users who are trying to use ids.

### Reproduction of behavior
https://plnkr.co/edit/QSY2ChJ9p9Mwk2EZ35eD?p=preview

The logic in assigning a page an id is failing. I think this is due to a falsy return when the id counter (or the id passed in) is zero.

Here's a screenshot of devtools inspecting the problem. Note the disconnect between the first page's `_id` value and the value assigned to the id. Zero versus 3... Those should not be different.

screen shot 2017-04-24 at 5 32 04 pm

imcotton:
Default page `id` should only taken place by absence (~~not~~ `undefined`) of `@Input("id")`.

Fixes #783

### DIFF
--- a/src/clarity-angular/wizard/wizard-page.ts
+++ b/src/clarity-angular/wizard/wizard-page.ts
@@ -170,15 +170,13 @@ export class WizardPage implements OnInit {
         new EventEmitter();
 
     // If our host has an ID attribute, we use this instead of our index.
-    // sometimes people are using array indices here. that's why it's "any".
-    // most often a string or a number.
     @Input("id")
-    _id: any = (wizardPageIndex++).toString();
+    _id: string = `${wizardPageIndex++}`;
 
     public get id() {
-        if (!this._id && this._id !== 0) {
+        if (this._id == null) {
             // guard here in the event that input becomes undefined or null by accident
-            this._id = (wizardPageIndex++).toString();
+            this._id = `${wizardPageIndex}`;
         }
         return `clr-wizard-page-${this._id}`;
     }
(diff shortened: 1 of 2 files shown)


imcotton:
@mathisscott I made cherry-pick of your PR in #829 for the TDD specs

src/clarity-angular/wizard/wizard-page.ts
```diff
@@ -171,12 +171,12 @@ export class WizardPage implements OnInit {
 
     // If our host has an ID attribute, we use this instead of our index.
     @Input("id")
-    _id: string = (wizardPageIndex++).toString();
+    _id: string = `${wizardPageIndex++}`;
 
     public get id() {
-        if (!this._id) {
+        if (this._id == null) {
```
mathisscott: Hi @imcotton 
First, our linting calls for triple equals. So I'm not sure how this passes.
Second, this check would allow values of empty string and false to pass through. But that does give me an idea for more unit tests on the other PR to guard against those.

mathisscott:
I'm closing this PR because it's a duplicate of one we already have approved.

src/clarity-angular/wizard/wizard-page.ts
imcotton: - It allows compare to `null` with double equals by `allow-null-check` flag.
- I thought AoT would perform type guard through @Input() decorator, turns out still not there yet, therefore it's going to require runtime type checking against every non-primitive type as well.

mathisscott: Thanks for the feedback @imcotton. We will take it into consideration.

Issue #734, opened 2017-04-17
Should not show the hint of the Clarity icon


steven-zou:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
No DOM native hint shown for the Clarity icon.

### Actual behavior
Hint is displayed when hover on the icon.
<img width="152" alt="2017-04-17 10 56 14" src="https://cloud.githubusercontent.com/assets/5753287/25077450/875c354c-235c-11e7-8b62-dd57a552c913.png">

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:
[Legacy - 0.8.15]

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:**  
<=0.8.15

* **OS and version:**  OS 10.11.4

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Firefox 52 | Safari 9.0

Jinnie:
Second time this gets reported:
https://github.com/vmware/clarity/issues/678

mathisscott:
Thanks for the reminder, @Jinnie. #678 was a broad "this happens in places" issue. This one is specific to tooltips and gives us something to move forward with.

steven-zou:
Guys,

  Any progress?

mathisscott:
@steven-zou 
This has been added to our backlog and is currently behind other priorities. If you give us your use case, we may be able to move it in priority. 

Since the fix is most likely to use CSS to turn pointer events off for clr-icon elements in some cases, this is also something you could probably fix as well if you need it urgently.

victormejia:
@mathisscott let me know if you'd like me to tackle this one.

mathisscott:
Absolutely, @victormejia . It would be greatly appreciated.

I suspect it could be as simple as using `pointer-events: none` in the CSS when a `clr-icon > svg` is inside a tooltip triggering element.

Jinnie:
I had a plunker with this some time ago, but it seems I've lost it in a recent plunker space cleanup. 😞 
It was exactly as @mathisscott proposes, with pointer-events:none, but this also prevents any mouse interaction with the svg. So you need to have some html element as container, to be able to intercept clicks, hovers, change the pointer.

mathisscott:
[...]

PR #833, opened 2017-05-03
Fix for "Should not show the hint of the Clarity icon #734"

chandranive:
Changed the css tooltips.clarity.scss, to add "pointer-events: none;".

### DIFF
--- a/src/clarity-angular/tooltips/_tooltips.clarity.scss
+++ b/src/clarity-angular/tooltips/_tooltips.clarity.scss
@@ -186,10 +186,7 @@ $clr-tooltip-adjusted-margin: $margin-value + $arrow-width * 2 !default;
     .tooltip {
         & > .clr-icon{
             margin-right: 0;
+            pointer-events: none;
         }
-        clr-icon > svg {
-        pointer-events: none;
-  }
-
     }
 }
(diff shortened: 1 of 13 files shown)


chandranive:
Signed-off-by: Nivedita Chandra <chandran@vmware.com>

src/clarity-angular/tooltips/_tooltips.clarity.scss
```diff
@@ -186,6 +186,7 @@ $clr-tooltip-adjusted-margin: $margin-value + $arrow-width * 2 !default;
     .tooltip {
         & > .clr-icon{
```
mathisscott: `.clr-icon` isn't a class that is used with the Clarity Icon library. It's legacy from before our custom element based Icon framework.

Also, you probably want the pointer-events to be on the `svg` element inside the Clarity Icon.

So this line should be more like...

```
.tooltip {
  & > .clr-icon {
    margin-right: 0;
  }

  clr-icon > svg {
    pointer-events: none;
  }
}
```

mathisscott:
`Signed-off-by: Nivedita Chandra <chandran@vmware.com>` needs to be the last line of your commit message. Rebase off master and add that to your commit message and then force push up to your forked branch.

src/clarity-angular/tooltips/_tooltips.clarity.scss
chandranive: Thanks @mathisscott for your suggestion. I have made the required changes and done a commit. Do I need to have a new Pull request too?

Issue #610, opened 2017-03-17
Clicking next step in wizard should move to the next page


ScottLangham:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x ] enhancement
```

### Expected behavior
After completing a page, clicking on the next step (at the left) should move the user to the page for that step (assuming the current page is valid). It should do the same as the NEXT button.

After watching users trial our product, it's been noted they try to hit on the step (on the left of the wizard) to go to the next page, as well as to go backwards.

### Actual behavior
Clicking on the next step does not take the user to the page for that step. They have to use the NEXT button. (However, users can click on steps to go backwards, so it seems sensible that they should be able to click steps to go forward too).

### Reproduction of behavior
The examples on this page show the behavior: https://vmware.github.io/clarity/documentation/wizards

### Environment details

* **Angular version:** 2.4.9

* **Clarity version:** 0.8.9

* **OS and version:** Windows 8

* **Browser:** all

reddolan:
I agree, if we are able to make this interaction available

mathisscott:
I'll check to see if it is possible in the new wizard. If not, I'll assign this as an enhancement after the new wizard is out.

mathisscott:
I've verified that the new wizard does not allow this. It follows the interaction/behavior of the pre-0.9 wizard.

I will add this to the backlog. I can see where it is confusing. But it will have to be an enhancement we add after the 0.9 wizard is final.

Issue #810, opened 2017-04-27
[Wizard] Stepnav is jumping to the wrong page when pages are hidden/shown (bullet pt #2 below)


yaooluu:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
* Use *ngIf to control page skipping and unskipping for dynamically created pages.

### Actual behavior
* Open wizard, hit "Toggle page 2 visibility" button twice. Click Next to page 2. The dynamically generated page content is missing.
* Relevant issue: Open wizard, navigate to page 3, then navigate back to page 1. Hit button twice. It jumps to page 3 directly.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates: https://plnkr.co/edit/8bkNAtzXiadhq4THKQS7?p=preview

### Environment details

* **Angular version:** 4.0.1

* **Clarity version:** 0.9.1

mathisscott:
@yaooluu : This appears to be two separate issues.

The first issue is the generated page content going away after toggling the second step. This is because the generated page content is only being inserted on `ngAfterViewInit`. You will likely want to change this to insert the page content either with `ngOnChanges` or in both `ngAfterViewInit` _and_ `ngOnChanges`. I'll ping @youdz to set me straight on this if I'm confused.

The second issue is the navigation jumping to the third step. I believe that to be a bug.

If my assessment of the first issue is accurate, it's not a bug but a way that the dynamically generated wizard pages will need to interact/integrate with Angular.

If that is correct, we'll convert this to a stepnav bug and triage accordingly.

youdz:
Hi @yaooluu, I think adding the dynamic pages this from the parent component (in this case, `AppComponent`) creates a lot of issues if you want to stay dynamic, and you end up reinventing the wheel by listening to every lifecycle event that happens. Right now, you'd need to listen to the `Querylist` changes, identify the ones that appeared, and re-insert the dynamic component. Not only is this painful to write and maintain, it also has no reason to fall under the responsibility of the app component. 

I believe the correct solution is simply to use a "placeholder" page component that appends the real page to itself on init. I modified your plunker to show how it would work:
https://plnkr.co/edit/RjLGlkFkUmhpohNUCSvy?p=preview
[...]

Issue #784, opened 2017-04-24
Wizard: allow for resetting of previous pages when jumping back using stepnav


mathisscott:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[X] enhancement
```

### Expected behavior
Offer an input that when `true` will reset the completed state between the current page and a page behind it in the stepnav when a stepnav link is clicked.

### Actual behavior
The stepnav allows users to jump back to any completed page. This causes an issue if a wizard is setup to do validation when the "Next" button is clicked and a user changes something on a previous page that causes an invalid state on later pages.

mathisscott:
The scope of this issue requires a number of fixes and refactors to the wizard navigation. These could constitute minor breaking changes to the API but also fix issues that are currently wrong with the wizard that haven't been discovered yet.

Due to this, I'm increasing the estimate and flagging this item as potentially getting pushed to the 0.9.4 release.

mathisscott:
Given the environment issues that have cost me this morning. There is no way this is making it into 0.9.3.

Issue #769, opened 2017-04-21
Clean up new wizard code and work on tests


mathisscott:
Some cleanup items remain on the new wizard code.

Also need to try to get the failing PhantomJS issues resolved, if possible.

PR #843, opened 2017-05-04
[NG] wizard force forward feature and nav fixes

mathisscott:
• adds a forceForward input to the wizard that marks pages incomplete when navigating backwards. useful for when you need to do validation on the primary button click of every page.

• fixes an issue where step-nav items that could not be navigated to but could still get focus

• fixes an issue where reset was happening before the last page was committed, meaning the final page would show up completed even after reset

• fixes an issue where users could not click not the step nav item to navigate to the next step

Closes: #610, #784

[NG] wizard bifurcating next/finish from wizard calls

• adding ability to pass a boolean to wizard.next() and wizard.finish() to avoid running event emissions and checks. this can cause an event loop if fired from an onNext or onFinish event handler

• can also explicitly call wizard.forceNext() or wizard.forceFinish(). These likewise bypass page checking and event emission

• by default, wizard.next() and wizard.finish() bypass state checks and event emission because 80% of the time that is what people seem to want to do

• the only time you might want to do wizard.next(false) or wizard.finish(false) to force state checks and event emission is with custom buttons that you want to duplicate existing functionality with. pretty sure those aren’t being used to a large extent.

[NG] unit tests

[NG] unit tests

[NG] unit tests

[NG] unit tests

### DIFF
--- a/src/clarity-angular/wizard/providers/wizard-navigation.ts
+++ b/src/clarity-angular/wizard/providers/wizard-navigation.ts
@@ -58,11 +58,10 @@ export class WizardNavigationService implements OnDestroy {
             } else {
                 this.cancel();
             }
-            // SPECME
         });
 
         this.pagesResetSubscription = this.pageCollection.pagesReset.subscribe(() => {
-            this.setLastEnabledPageCurrent();
+            this.setFirstPageCurrent();
         });
     }
 
@@ -85,6 +84,7 @@ export class WizardNavigationService implements OnDestroy {
 
     public currentPage: WizardPage;
     public navServiceLoaded = false;
+    public forceForwardNavigation = false;
 
     public get currentPageTitle(): TemplateRef<any> {
         // when the querylist of pages is empty. this is the first place it fails...
@@ -139,7 +139,6 @@ export class WizardNavigationService implements OnDestroy {
         if (!this.wizardHasAltNext) {
             this._movedToNextPage.next(true);
         }
-        // SPECME
     }
 
     public forceNext(): void {
@@ -156,7 +155,6 @@ export class WizardNavigationService implements OnDestroy {
             this.pageCollection.commitPage(currentPage);
         }
         this.setCurrentPage(nextPage);
-        // SPECME
     }
 
     public checkAndCommitCurrentPage(buttonType: string): void {
@@ -200,11 +198,16 @@ export class WizardNavigationService implements OnDestroy {
             return;
         }
 
+        // order is very important with these emitters!
         if (isFinish || isDangerFinish) {
+            // mark page as complete
+            this.pageCollection.commitPage(currentPage);
             this._wizardFinished.next();
         }
 
         if (this.wizardHasAltNext) {
+            this.pageCollection.commitPage(currentPage);
+
             if (isNext || isDangerNext) {
                 this._movedToNextPage.next(true);
             }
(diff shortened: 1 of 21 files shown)


mathisscott:
This is a feature the installer team requested that I though would be easy but really, really wasn't. 

There are some test plans in here that I didn't get to finish out. I have #769 for going back through and working those.

src/clarity-angular/wizard/providers/wizard-navigation.ts
```diff
@@ -58,11 +58,10 @@ export class WizardNavigationService implements OnDestroy {
             } else {
                 this.cancel();
             }
-            // SPECME
         });
 
         this.pagesResetSubscription = this.pageCollection.pagesReset.subscribe(() => {
-            this.setLastEnabledPageCurrent();
+            this.setFirstPageCurrent();
```
adityarb88: Sorry I am new to the wizard so don't know much but this looks like a behavior change. Does this mean that instead of the last enabled page, the first page of the wizard will be the current page when the reset is called?

mathisscott: Ah. No. This is just being more explicit about what we want the wizard to do here. Reset should reset the navigation to its pristine state. 

This change here might actually fix issue #810

src/clarity-angular/wizard/wizard.ts
```diff
@@ -162,12 +163,17 @@ export class Wizard implements OnInit, OnDestroy, AfterContentInit, DoCheck {
...
+        let navService = this.navService;
+
         this.pageCollection.pages = this.pages;
-        this.navService.wizardHasAltCancel = this.stopCancel;
-        this.navService.wizardHasAltNext = this.stopNext;
         this.headerActionService.wizardHeaderActions = this.headerActions;
+
+        navService.wizardHasAltCancel = this.stopCancel;
+        navService.wizardHasAltNext = this.stopNext;
+        navService.forceForwardNavigation = this.forceForward;
```
youdz: I just realized, what if these inputs change dynamically at some point? They're not propagated to the navService, so the service will be out of sync with the inputs given by the user.

Why not just use:
```typescript
@Input("clrWizardPreventDefaultCancel") 
set stopCancel(value: boolean) {
    this.navService.wizardHasAltCancel = value;
}
```

src/clarity-angular/wizard/wizard.ts
```diff
@@ -252,12 +271,20 @@ export class Wizard implements OnInit, OnDestroy, AfterContentInit, DoCheck {
     }
 
     // LEGACY: Naming convention matches old wizard
-    public next(): void {
-        this.navService.next();
+    public next(skipChecksAndEmits: boolean = true): void {
+        if (skipChecksAndEmits) {
```
youdz: Isn't setting the default to `true` a breaking change?

src/clarity-angular/wizard/wizard.ts
mathisscott: Good suggestion. I'll add that to the wizard cleanup I'm currently working on.

src/clarity-angular/wizard/wizard.ts
mathisscott: No. Not for how `Wizard.next()` is used 80% of the time. The most common use case is programmatically navigating the wizard when you've prevented default navigation. In that case the event emitters and checks have already run. You don't want to run them again or you'll get an infinite loop of event emissions.

PR #849, opened 2017-05-05
[ui/ng] nested menu

### LINKED ISSUE #565: Nested Menu Component
```
[ ] bug
[X] feature request
[ ] enhancement
```

### Description
Nested menus save space and organize items/actions into categories. The user can hover to reveal deeper levels as needed.
- Hierarchy: darker border and slight overlap (4px) of nested menus help to show stack order.
- 168 px /336 px  Min/max width 
- Default ine height is 30 px. 36 px for x-small breakpoint (<544px)
 -- Current dropdown menu line height (24px) to be updated to 30px
- Group name stays in selected state to serve as a breadcrumb
- By default nested menus should pop up on the right side of the previous menu. If space isn’t available, the menu should appear on the left side. A third menu would appear on the right side of the second menu in a zig-zag pattern.

### Interaction
- With mouse: Hover-over group name to reveal nested menus. Add a delay to mitigate user errors and disappearing menus. On touch screens: nested menus are revealed on touch.
- Clicking outside of menu dismisses all menus
nm_example

### Menu stack order and positioning
By default nested menus pop up to the right side of the previous menu 
default positioning

If there isn’t enough space, the second menu should pop up on the left side of the first menu. The third menu would appear on the right side of the second menu in a zig-zag pattern.

responsive

nm_redline

jeeyun:
Implementation of nested menu. Resolves #565.

Static/UI:
screen shot 2017-05-05 at 1 32 27 pm

Angular:
dropdown

A few notes:

- Opting for a markup (both in UI and NG) that allows us to have dropdown items that are not nested. Otherwise, we could end up with a button inside button, which is not good.
- Using desugared syntax for our popover and moving the `dropdown-menu` class to `<clr-dropdown-menu>` allowed us to get rid of a hack in our css. Yay!
- open to feedback on name changes for the new css classes and the angular component's input `clrDropdownSubmenuOpenDirection`
- Updated the css tests to include dropdown with nested menu items.


src/app/dropdown/dropdown-angular-close-item-false.demo.html
```diff
@@ -10,14 +10,14 @@
             <clr-icon shape="error" class="is-error" size="24"></clr-icon>
             <clr-icon shape="caret down"></clr-icon>
         </button>
-        <div class="dropdown-menu">
+        <clr-dropdown-menu>
```
mathisscott: I know it may seem redundant but with the wizard I was adding the component tag as a classname (like `class="clr-dropdown-menu"`) so we (and end users) could have a hook to styling that was not reliant on the component's tagname.

src/clarity-angular/dropdown/dropdown-menu.ts
```diff
@@ -4,6 +4,11 @@ import {Component} from "@angular/core";
     selector: "clr-dropdown-menu",
     template: `
         <ng-content></ng-content>
-    `
+    `,
+    host: {
+        "[class.dropdown-menu]" : "true",
```
mathisscott: This class should probably be `clr-dropdown-menu` to bring it into line with our current path with CSS naming conventions.

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,41 @@
+import {Component, ContentChildren, Input, QueryList} from "@angular/core";
+import {DropdownItem} from "./dropdown-item";
+
+@Component({
+    selector: "clr-dropdown-submenu",
+    template: `
+        <ng-content></ng-content>
+    `,
+    host: {
+        "[class.dropdown-submenu]" : "true",
```
mathisscott: This should likewise be `clr-dropdown-submenu`.

src/clarity-angular/dropdown/dropdown-menu.ts
```diff
@@ -4,6 +4,11 @@ import {Component} from "@angular/core";
...
+        "[class.open]" : "true" // always true in angular; it's the responsibility of containing component to show/hide
```
adityarb88: why is the open class on the `dropdown-menu`? I thought it has to be on the `clr-dropdown`. if this relates to the sub menu then shouldn't we add this class only when the sub menu is open?

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,41 @@
...
+        "[class.open]" : "open",
+        "[class.right]" : "openDirection === 'right'",
+        "[class.left]" : "openDirection === 'left'",
+        "(click)": "open = true",
```
adityarb88: this is not an issue but just a question. angular recommends that we use `HostListener`s instead of this but I am not sure what convention we are following. I know we do regularly for classes but very rarely for events, no?

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,41 @@
...
+        "(mouseenter)": "open = true",
+        "(mouseleave)": "open = false"
+    }
+})
+export class DropdownSubmenu {
+    private _open: boolean = false;
+    @Input ("clrDropdownSubmenuOpenDirection") openDirection: string = "right";
+
+    @ContentChildren(DropdownItem, {descendants: false}) menuItems: QueryList<DropdownItem>;
```
adityarb88: oh nice. learned a new thing today. didn't know about the `descendants` option :D. so this basically gets all submenus no matter how deeply nested they are right?

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    ngAfterContentInit() {
+        if (this.menuItems.length === 0) {
+            throw new Error("A <clr-dropdown-submenu> must contain a child element with a clrDropdownItem directive. ");
```
adityarb88: this is nice. but i didn't find it in our main `clr-dropdown-menu` component. you think we should do it there too? or do we not use `ContentChildren` there?

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+        this.menuItems.first.expandable = true;
+    }
+
+    get open() {
+        return this._open;
+    }
+
+    set open(open: boolean) {
+        console.log("open", open);
```
adityarb88: `console.log("do we need this?")` :D

src/clarity-angular/dropdown/dropdown.ts
```diff
@@ -15,17 +15,15 @@ import {
...
 @Component({
     selector: "clr-dropdown",
     template: `
         <ng-content select="[clrDropdownToggle]"></ng-content>
         <ng-template [(clrPopover)]="open" [clrPopoverAnchor]="anchor" [clrPopoverAnchorPoint]="anchorPoint"
                      [clrPopoverPopoverPoint]="popoverPoint">
-            <div class="dropdown-menu">
-                <ng-content select="[clr-dropdown-menu, .dropdown-menu]"></ng-content>
-            </div>
+            <ng-content select="[clr-dropdown-menu, .dropdown-menu]"></ng-content>
```
adityarb88: again a question. I think we have both selectors here for backward compatibility but eventually we will get rid of the css selector. is that correct?

src/clarity-angular/dropdown/dropdown-menu.ts
jeeyun: Will leave this as is for now for backward compatibility reasons. We should tackle this when we address css naming convention across the board.

src/app/dropdown/dropdown-angular-nested.demo.html
```diff
@@ -0,0 +1,66 @@
+<!--
+  ~ Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+
+<div class="clr-example squeeze demo-menu-height dropdown-demo">
```
jeeyun: Thank you. Will double check this when I work on the website documentation.

src/app/dropdown/dropdown-angular-nested.demo.html
```diff
@@ -0,0 +1,66 @@
...
+            <label class="dropdown-header">Dropdown header</label>
+            <a href="javascript://" clrDropdownItem>Action 1</a>
+            <a href="javascript://" class="disabled" clrDropdownItem>Disabled Action</a>
+            <div class="dropdown-divider"></div>
+            <clr-dropdown-submenu clrDropdownSubmenuOpenDirection="right">
+                <a href="javascript://" clrDropdownItem>Link 1</a>
+                <clr-dropdown-menu>
+                    <a href="javascript://" clrDropdownItem>Foo</a>
+                    <clr-dropdown-submenu clrDropdownSubmenuOpenDirection="left">
+                        <a href="javascript://" clrDropdownItem>Bar</a>
```
jeeyun: You are right. I will change them into buttons.

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -144,6 +132,24 @@ $dropdown-white: clr-getColor(lightest);
                 text-decoration: none;
             }
 
+            &.expandable {
+                $clr-dropdown-caret-xOffset: baselineRem(8/24);
```
jeeyun: This one is not in the spec per se.. but it's used for placing the caret for the expandable items (items with submenu). I'll still change it to use the `$clr_baselineRem_1` here.

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -203,6 +209,41 @@ $dropdown-white: clr-getColor(lightest);
             visibility: visible;
         }
 
+        .dropdown-submenu {
+            $clr-dropdown-submenu-overlap-margin: baselineRem(4/24);
```
jeeyun: Done.

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -203,6 +209,41 @@ $dropdown-white: clr-getColor(lightest);
...
+            position: relative;
+
+          .dropdown-menu {
+            border-color: #9a9a9a;
```
jeeyun: Yes you are right. Will use `$clr-dark-midtone-gray`.

src/clarity-angular/dropdown/dropdown-menu.ts
jeeyun: Oops that should not be needed. Removing.

src/clarity-angular/dropdown/dropdown-submenu.ts
jeeyun: I have refactored them to use `HostListener`s.

src/clarity-angular/dropdown/dropdown-submenu.ts
jeeyun: Actually, the opposite. By passing in `{descendants: false}` (it's `true` by default if you don't specify), we are only checking for the direct children. That way at each level of submenu, we validate the existence of a child element with a `clrDropdownItem` directive.

src/clarity-angular/dropdown/dropdown-submenu.ts
jeeyun: Yes this is something new I've added while working on this. Should we add a similar feature to `dropdown-menu`?

src/clarity-angular/dropdown/dropdown.ts
jeeyun: Yes, that's correct. There is a comment in the same file above noting that. :)

src/clarity-angular/dropdown/dropdown-submenu.ts
jeeyun: Nope. Thanks for catching that. Cleaning up.

src/clarity-angular/dropdown/dropdown-submenu.ts
adityarb88: Had the same question. May be we can ask @youdz about that. I don't think its absolutely needed but it might be a nice to have.

src/app/dropdown/dropdown-angular-close-item-false.demo.html
youdz: What we tend to do for this is have the _component_ itself add the CSS class to the host. Less verbosity for the user.

src/clarity-angular/dropdown/dropdown-item.ts
```diff
@@ -14,17 +14,23 @@ import { Dropdown } from "./dropdown";
...
+    active: boolean = false;
+
     constructor(private _dropdown: Dropdown, private el: ElementRef) {
     }
 
     @HostListener("click")
     onDropdownItemClick(): void {
-        if (this._dropdown.isMenuClosable && !this.el.nativeElement.classList.contains("disabled")) {
+        if (this._dropdown.isMenuClosable && !this.el.nativeElement.classList.contains("disabled")
+                && !this.el.nativeElement.classList.contains("expandable")) {
```
youdz: Note for later: we really need to clean this up. Checking for `.disabled` on the native classlist is really bad news. 😕 

That said, in your case, you have absolutely no excuse to use this pattern. You literally have an `expandable` property already on your component. So please use that instead.

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,50 @@
+import {Component, ContentChildren, HostListener, Input, QueryList} from "@angular/core";
+import {DropdownItem} from "./dropdown-item";
+
+@Component({
+    selector: "clr-dropdown-submenu",
+    template: `
+        <ng-content></ng-content>
+    `,
```
youdz: This bothers me. Yet another component to learn about. I forgot to post a precise spec for nested menus before you started working on it, that's on me. We'll talk tomorrow morning and see our options.

src/clarity-angular/dropdown/dropdown-submenu.ts
youdz: Wow, descendants is now true by default? It was false not long ago, that's actually a breaking change on their part. When did it happen, 4.0?

src/clarity-angular/dropdown/dropdown-submenu.ts
```diff
@@ -0,0 +1,50 @@
...
+    host: {
+        "[class.dropdown-submenu]" : "true",
+        "[class.open]" : "open",
+        "[class.right]" : "openDirection === 'right'",
+        "[class.left]" : "openDirection === 'left'"
```
youdz: I forgot this comment on this: right/left is probably not enough, we need the full positioning power. If your submenu is too long and you're close to the bottom of the screen, you need to be able to have it go up, with the bottom of the link and the sub-menu aligned.

src/clarity-angular/dropdown/_dropdown.clarity.scss
```diff
@@ -85,18 +85,6 @@ $dropdown-white: clr-getColor(lightest);
         visibility: hidden;
         z-index: map-get($clr-layers, dropdown-menu);
 
-        .dropdown-menu {
-            // TODO: this is a hack to prevent breaking changes as we implement
-            // the popover directive. Remove this after the < 0.9.0 markup for
-            // dropdown (angular component) is deprecated.
```
jeeyun: Hmm I think you misunderstand; the code changes in this PR made it possible to remove the hack and the deprecated feature still works as expected. But, I'll leave this comment as is, as we are still undergoing API design changes.

youdz: Oh, my bad then. If the deprecated feature still works, then we're good. 👍

jeeyun:
@youdz would appreciate your review on the code changes. Had a discussion with @lil-kim also and decided to go forward with reveal on click (as opposed to hover) to make the experience consistent. I'll hold off on rebasing until your review.

PR #859, opened 2017-05-08
[NG] Input tab trap directive

### LINKED ISSUE #767: Tabbing within modals should keep the focus on the modal
```
[X] bug
[] feature request
[ ] enhancement
```

### Expected behavior
When tabbed in the modal, the tab focus should stay within the modal. This helps in better user accessibility.

### Actual behavior
When tabbed in the modal, the tab focus leaves the modal.

### Reproduction of behavior
This behavior can be reproduced in the clarity design website
https://vmware.github.io/clarity/documentation/modals

### Environment details
* **Angular version:** 2.0.X
N/A
* **Clarity version:** 
latest clarity version
* **OS and version:** 
MAC os
* **Browser:** 
Chrome

victormejia:
New `clrInputTrap` directive to trap the user tabbing to the inputs within the container.

Would love some input here, as not entirely sure how exhaustive the `focusableInputs` list should be. Thanks!

screen recording 2017-05-08 at 03 15 pm

Resolves #767

Signed-off-by: Victor Mejia <victor.a.mejia@gmail.com>

### DIFF
--- a/src/clarity-angular/input-trap/input-trap.directive.ts
+++ b/src/clarity-angular/input-trap/input-trap.directive.ts
@@ -1,0 +1,47 @@
+import { Directive, HostListener, ElementRef, OnInit } from "@angular/core";
+
+@Directive({
+    selector: "[clrInputTrap]"
+})
+export class InputTrapDirective implements OnInit {
+
+    static TAB_KEY = 9;
+
+    // tslint:disable-next-line:max-line-length
+    // reference: https://github.com/jkup/focusable
+    static FOCUSABLE_INPUTS = "a[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled])";
+
+    firstInput: HTMLElement;
+    lastInput: HTMLElement;
+
+    constructor(public elementRef: ElementRef) { }
+
+    @HostListener("document:keydown", ["$event"])
+    onkeydown(event: any) {
+        const isShiftKey = !!event.shiftKey;
+
+        if (event.keyCode === InputTrapDirective.TAB_KEY) {
+            if (isShiftKey) {
+                if (document.activeElement === this.firstInput) {
+                    event.preventDefault();
+                    this.lastInput.focus();
+                }
+            }  else if (document.activeElement === this.lastInput) {
+                event.preventDefault();
+                this.firstInput.focus();
+          }
+        }
+    }
+
+    ngOnInit() {
+        const element = this.elementRef.nativeElement;
+
+        const focusableInputs = Array.from(
+            element.querySelectorAll(InputTrapDirective.FOCUSABLE_INPUTS)
+        ) as HTMLElement[];
+
+        this.firstInput = focusableInputs[0];
+        this.lastInput = focusableInputs[focusableInputs.length - 1];
+    }
+
+}
--- a/src/app/modal/modal-trap.demo.html
+++ b/src/app/modal/modal-trap.demo.html
@@ -1,0 +1,63 @@
+<!--
+  ~ Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<p>For accessiblity purposes, you may want to trap user input tabbing to within the modal.
+You can use the <code>clrInputTrap</code> directive. Both <code>SHIFT</code> and <code>SHIFT+TAB</code>
+will keep the user tabbing from within the modal.
+</p>
+
+<div class="clr-example">
+    <div class="backdrop-example-container">
+        <div class="modal static">
+            <div class="modal-dialog" role="dialog" aria-hidden="true">
+                <div clrInputTrap class="modal-content">
+                    <div class="modal-header">
+                        <button aria-label="Close" class="close" type="button">
+                            <clr-icon aria-hidden="true" shape="close"></clr-icon>
+                        </button>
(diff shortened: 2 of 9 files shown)


src/clarity-angular/input-trap/input-trap.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+          }
+        }
+    }
+
+    ngOnInit() {
+        const element = this.elementRef.nativeElement;
+
+        const focusableInputs = Array.from(
+            element.querySelectorAll(InputTrapDirective.FOCUSABLE_INPUTS)
+        ) as HTMLElement[];
```
mathisscott: What happens if an element is removed from the DOM or added to the DOM with a structural directive?

Would it be better to have this be a dynamic check at the time of focus instead of static at init?

victormejia: ah very good point, didn't take this into account. I'll make the changes.

src/clarity-angular/input-trap/input-trap.directive.ts
```diff
@@ -0,0 +1,47 @@
+import { Directive, HostListener, ElementRef, OnInit } from "@angular/core";
+
+@Directive({
+    selector: "[clrInputTrap]"
+})
+export class InputTrapDirective implements OnInit {
```
youdz: This should be named "Focus trap". 😉 
You could very well have a modal with no inputs, just buttons, that wants to prevent focus outside.

src/clarity-angular/input-trap/input-trap.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+    // reference: https://github.com/jkup/focusable
+    static FOCUSABLE_INPUTS = "a[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled])";
+
+    firstInput: HTMLElement;
+    lastInput: HTMLElement;
+
+    constructor(public elementRef: ElementRef) { }
+
+    @HostListener("document:keydown", ["$event"])
+    onkeydown(event: any) {
```
youdz: We talked about this a couple weeks ago, and the solution we wanted to go with was to listen to the `focusin` event on the whole document, rather than `keydown`. See https://developer.mozilla.org/en-US/docs/Web/Events/focusin.

The keydown solution you have implemented doesn't support `tabindex` (it even breaks if the last element in the DOM of the trap is _not_ the last one in the tab order), it tries to tab to elements even if they are not displayed, it doesn't work when the user interacts manually outside of the trap like the URL bar, ... The focusin solution is far less hackier and much more robust.

src/clarity-angular/input-trap/input-trap.directive.ts
youdz: The problem of a dynamic check is that it can be very bad for performances when the user starts typing anywhere on the page. See my comment about `focusin` to avoid all this focusable element tracking.

youdz: The problem of a dynamic check is that it can be very bad for performance when done on a complex element, because the keydown event on the document can be triggered a lot when typing text.

reddolan:
@victormejia do you happen to have a plnkr of this or some interactive demo?

The GIF you provided looks good and the tab ordering seems accurate as well. I am wondering where the focus starts when the modal first opens and want to test it on top of content

src/clarity-angular/input-trap/input-trap.directive.ts
victormejia: yup, will go ahead and rename

src/clarity-angular/input-trap/input-trap.directive.ts
victormejia: @youdz Will go ahead and change to use the `focusin` event. However, don't we need to still determine what is considered the first and last focusable elements? Would I expand the list to include those with `tabindex`? Then that would include more logic, right? Since we then have to take into account those indices.

Just want to make sure I understand everything here 😄

victormejia:
@reddolan will try to get a plunkr going as soon as I make some changes...

src/clarity-angular/input-trap/input-trap.directive.ts
youdz: @victormejia: Actually, the `focusin` pattern wouldn't need you to track focusable elements at all anymore. What you do is check the target of the `focusin` event:
- If it's in the modal, you do nothing.
- If it's outside of the modal, then you re-focus *on the modal itself*. You don't need to focus necessarily on a focusable element, focusing on the modal element is good enough.

This would make it support naturally tabindex, ignore hidden elements, handle dynamic elements without killing performances, ...

src/clarity-angular/input-trap/input-trap.directive.ts
victormejia: very interesting, learning something new every day 😄 ....

src/clarity-angular/input-trap/input-trap.directive.ts
victormejia: very interesting, learning something new every day 😄 ....

jaffoneh:
@victormejia just wanted to quickly ping you and see if you have an update for this PR?

victormejia:
hi @jaffoneh ! Yup, should have an update to this by tomorrow morning.

victormejia:
Closing this PR as it's been a while and things have changed quite a bit :)

New PR: https://github.com/vmware/clarity/pull/1056

PR #866, opened 2017-05-09
[CORE] upgrade tslint version

chunghha:
please review rules updated especially no-trailing-whitespace/semicolon which are just temporarily set to avoid lots of lint errors.

Signed-off-by: Chris Ha  <chunghha@users.noreply.github.com>

### DIFF
--- a/build/tslint.json
+++ b/build/tslint.json
@@ -1,4 +1,7 @@
 {
+  "linterOptions": {
+      "typeCheck": true
+  },
   "rules": {
     "align": [
       true,
@@ -7,7 +10,9 @@
     ],
     "ban": false,
     "class-name": true,
-    "comment-format": false,
+    "comment-format": [
+      false
+    ],
     "curly": true,
     "eofline": false,
     "forin": true,
@@ -15,7 +20,9 @@
       true,
       "spaces"
     ],
-    "interface-name": false,
+    "interface-name": [
+      false
+    ],
     "jsdoc-format": true,
     "label-position": true,
     "max-line-length": [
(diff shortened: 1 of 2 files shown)


build/tslint.json
```diff
@@ -39,19 +50,20 @@
...
+      false
+    ],
     "no-internal-module": true,
     "no-null-keyword": false,
     "no-require-imports": true,
     "no-shadowed-variable": true,
     "no-string-literal": false,
     "no-switch-case-fall-through": true,
-    "no-trailing-whitespace": true,
+    "no-trailing-whitespace": false,
```
adityarb88: why disable this? we have this rule set in `.editorconfig` too.

build/tslint.json
```diff
@@ -72,14 +84,18 @@
       "avoid-escape"
     ],
     "radix": true,
-    "semicolon": [true, "always"],
+    "semicolon": [false, "always"],
```
Shijir: Are these necessary changes in TSlint rules as we upgrade its version?

build/tslint.json
chunghha: If not set it to false, I see errors from this kind of code.

```
throw new Error(
                `The argument must be an object literal passed in the following pattern: 
                { "shape-name": "shape-template" }`
            );
```

I also like to set this one true if team will be willing to change code.

build/tslint.json
chunghha: Same as above, I see errors if not set it to false. Please try 'gulp tslint' and compare with true/false. I would like to set this one true but I am not sure it's about time to change code to fix the errors.

chunghha:
This PR may be incomplete due to consequently code change will be needed as well to keep better rules pointed above. I just put this to the table for discussion and hope Clarity team will have enough time to review per team's plan/direction at this point.

youdz:
I'm going to close this until we figure out our formatting + linting strategy. I would very much like our linter to never worry about format, and just check code patterns, typings, and other healthy-code concerns.

It's becoming more and more of a priority, I don't think we'll postpone it much longer. 👍

PR #870, opened 2017-05-09
[core] CSS regression tests for datagrid (#845)

### LINKED ISSUE #845: CSS regression tests for datagrid
We need them...

jeeyun:
Created a kitchen sink page with datagrid demos for css regression. With this PR I am also breaking down the css regression tests into 2 sets to help speed up the build a bit. It doesn't quite cut the build time in half but shortens it a bit (17 min vs 23 min). 

Sample build for reference: https://travis-ci.org/jeeyun/clarity/builds/230163713

Resolves #845.


.travis.yml
```diff
@@ -5,9 +5,11 @@ env:
   global:
     - CXX=g++-4.8
   matrix:
-    - TEST_SUITE=test
-    - TEST_SUITE=css:test
-    - TEST_SUITE=css:reference
+    - TEST_SUITE='test'
```
jeeyun: good question. because they are run in parallel, that would not be possible. but because they are run in parallel, it shouldn't impact how long our build takes.

PR #890, opened 2017-05-11
[WEB] deleting unused wizard demo

mathisscott:


mathisscott:
This is part of the wizard cleanup ( #769 ). Not critical because it is just removing an unused demo. So zero effect on anything. Can be done after 0.9.4 release.

Issue #847, opened 2017-05-05
Datagrid: Error when loading active


siteslave:
**Select one ...**  (check one with "x")
```
[ x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Datagrid error when loading active.

![2017-05-05_145448](https://cloud.githubusercontent.com/assets/526890/25737634/19d287ec-31a3-11e7-961e-af49b7a61e76.png)

### Actual behavior

https://plnkr.co/edit/QuDQM0suEiiy9h8I7BvF?p=preview

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

### Environment details

* **Angular version:** 4.0.1

* **Clarity version:** 0.9.3

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Jinnie:
Hello @siteslave,

This seems to be an initialization timing issue, which you can workaround if you change the "loading" var assignment in your constructor to this:
    setTimeout(() => {
      this.loading = true;
    });
As it is without a timeout period parameter, the only effect it will have will be to yield the thread, so other code that waits on the line will have chance to finish its initialization.

Best regards

siteslave:
@Jinnie ,
In version 0.9.2, no errors were found.

youdz:
This is a bug introduced by the latest spinner refactor (needed to avoid clipping popovers). Assigning to @jeeyun for preliminary investigation since it's hopefully still fresh for her. We can then assign the actual fix for someone else.

PR #905, opened 2017-05-11
[NG] Fixes vertical scrolling for fixed height datagrids

hippee-lee:
I missed something or changed something in my code and didn't re-check the vertical scrolling with the fix for #847. 
This fix makes fixed height datagrids have vertical scrolling.  


adityarb88:
Does this work in IE11?

hippee-lee:
Yes, tested in all the browsers.

PR #912, opened 2017-05-12
[NG] Move away from deprecated API

chunghha:
`Renderer` to `Renderer2`

### DIFF
--- a/src/clarity-angular/data/datagrid/built-in/filters/datagrid-string-filter.ts
+++ b/src/clarity-angular/data/datagrid/built-in/filters/datagrid-string-filter.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import {
-    Component, Input, ViewChild, ElementRef, Renderer, AfterViewInit, EventEmitter, Output
+    Component, Input, ViewChild, ElementRef, Renderer2, AfterViewInit, EventEmitter, Output
 } from "@angular/core";
 import {StringFilter} from "../../interfaces/string-filter";
 import {CustomFilter} from "../../providers/custom-filter";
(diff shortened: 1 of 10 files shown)


src/clarity-angular/data/datagrid/built-in/filters/datagrid-string-filter.ts
```diff
@@ -68,7 +68,7 @@ export class DatagridStringFilter extends DatagridFilterRegistrar<DatagridString
                 // We need the timeout because at the time this executes, the input isn't
                 // displayed yet.
                 setTimeout(() => {
-                    this.renderer.invokeElementMethod(this.input.nativeElement, "focus");
+                    this.input.nativeElement.focus();
```
youdz: Ok, I'm going to ask you to do a small change even though what you wrote here is the exact same behavior as we had before. The thing is that at the time we wrote the original code, `renderer.invokeElementMethod()` was supposed to work in webworkers. I'm aware that the Angular Renderer has changed since then and now fails when calling this method in a webworker.

We want to try a open the possibility, in the future, to have the datagrid work in webworker, with server-side rendering, or in any other headless mode. To do this, we made sure that every access to native elements API for the datagrid is in our own [DomAdapter](https://github.com/vmware/clarity/blob/master/src/clarity-angular/data/datagrid/render/dom-adapter.ts). So I would like you to write a one-line method in it:
```typescript
focus(element: any) {
   element.focus();
}
```
And just call this method here. Yes, it seems extreme, but that way we group all our not-headless-compatible code in a single class, in which we could add guards, tests or even a separate implementation for server-side rendering.

chunghha: Thank you, @youdz for this kind and detail explanation. I just checked the angular issue here, https://github.com/angular/angular/issues/13818

chunghha:
By looking at the summary of 1st test, I am not sure which one failed because that says "0 test failed". Please check and let me know if there is any further change required for my PR.

chunghha:
I was misreading '8' as '0' last night. Anyway, updated the spec too to pass checks.

chunghha:
@adityarb88 please see if the commit 177130d works for you.

adityarb88:
@chunghha: I don't think the commits were squashed. I now see 5 commits instead of 1:

image

chunghha:
@adityarb88 Hope the commit 66159c8 works this time.

adityarb88:
@chunghha: This one looks good. Will wait for our tests to pass and merge it in. Thanks!

PR #921, opened 2017-05-12
[CORE] gemini tests for new wizard

mathisscott:
• another piece of #769

[CORE] whoops

[CORE] updating gemini screens

[CORE] changing gemini test suites for animations

[CORE] update gemini screens

[CORE] gemini test fix

[CORE] update gemini screen


jeeyun:
@mathisscott you might want to make the whole-page screenshots more specific (to only include the wizard part). The reason for this is that when the background changes (for example, we add a new component so the left sidenav content changes) then the gemini tests will fail. But if you think it's tolerable to update screenshots for those cases, I'm okay with leaving them as is.

mathisscott:
I wanted one or two whole page shots so that we could tell if issues like the Firefox issue popped up. If it becomes a PITA, it's easy enough to change and I will be happy to do so.

gemini/tests/set2/wizard-deprecated.js
```diff
@@ -0,0 +1,27 @@
+var WAIT_TIME = 5000;
+var WAIT_LOAD_TIME = 2000;
```
hippee-lee: I looks like this is only using WAIT_TIME but not WAIT_LOAD_TIME.

mathisscott: yup. this was a pre-existing issue. i will fix it. thanks for the heads up

mathisscott: Done

PR #922, opened 2017-05-12
Deleted spaces in the imports

marcelocarmona:


mathisscott:
@Shijir , @jeeyun : Is this a code convention we are following?

marcelocarmona:
Yes I just thought that was the code convention you guys follow 👍

youdz:
Actually, the problem is that we haven't agreed on a convention for this yet. We'll do a large reformatting commit once we decide on a formatter, but I'd rather not have intermediate small reformats that might very well be rolled back by anyone with a different convention in the next few weeks (or months). Especially since you only updated a single file, when this format issue is present in many more in the project.

Sorry for discarding your PR. Please bear with us a little while longer until we all sync up and integrate a formatter directly in the build! 😉

PR #945, opened 2017-05-19
[ICONS] Using Object.defineProperty on aliases instead of getter acce…

Shijir:
…ssor

UglifyJS was treating kebab cases in a correct fashion. But when it's used after getter accessor, UglifyJS strips the wrapper quotes.  For example, UglifyJS turned the following getter accessor
```
get() "ellipses-vertical"
```
to 

```
get() ellipses-vertical
```
resulting in an invalid JS syntax.

To avoid this, we could resort to `Object.defineProperty` instead of `get` accessor.


src/clarity-icons/shapes/essential-shapes.ts
```diff
@@ -512,14 +503,10 @@ export const essentialShapes: any = {
         </svg>`,
 
     /*TODO: ellispes-vertical is deprecated and will be removed in 0.9.0*/
-    get "ellipses-vertical"() {
-        return this[ "ellipsis-vertical" ];
-    },
+
 
     /*TODO: ellispes-horizontal is deprecated and will be removed in 0.9.0*/
```
mathisscott: We need to change these deprecation TODOs to 0.10.0 and make sure they carry over with the new code.

src/clarity-icons/shapes/essential-shapes.ts
```diff
@@ -1703,13 +1655,29 @@ export const essentialShapes: any = {
         </svg>`
 
 
-
 };
 
+Object.defineProperty(essentialShapes, "edit", descriptorConfig(essentialShapes["pencil"]));
+Object.defineProperty(essentialShapes, "note-edit", descriptorConfig(essentialShapes["note"]));
+Object.defineProperty(essentialShapes, "group", descriptorConfig(essentialShapes["users"]));
+Object.defineProperty(essentialShapes, "ellipses-vertical", descriptorConfig(essentialShapes["ellipsis-vertical"]));
```
mathisscott: TODOs should be moved here so we can get rid of `ellipses-vertical` and `ellipses-horizontal`.

src/clarity-icons/shapes/essential-shapes.ts
Shijir: Saw them but forgot to update them 😛. Now they are updated to 0.10.0 and moved down to `Object.defineProperty` area.

PR #986, opened 2017-05-26
[ICONS] Added ability to assign non-SVG templates to ClarityIcons namespace

mathisscott:
• allows for the assigning of any white-listed html element to the ClarityIcons namespace
• sanitization added for increased security
• existing unit tests updated and new unit tests added

[ICONS] wip

[ICONS] wip

[ICONS] wip -- build working

[ICONS] wip - working global

[ICONS] wip - working with imports

[ICONS] wip

[ICONS] wip

[ICONS] wip

### DIFF
--- a/src/app/index.html
+++ b/src/app/index.html
@@ -32,6 +32,7 @@
     <link rel="stylesheet" href="node_modules/font-awesome/css/font-awesome.min.css">
     <link rel="stylesheet" href="clarity-icons.css">
     <link rel="stylesheet" href="clarity-ui.css">
+    <script src="node_modules/dompurify/dist/purify.min.js"></script>
     <!-- @endif -->
 
     <!-- @if NODE_ENV='prod' -->
@@ -57,6 +58,7 @@
             var map = {
                 'app':                                  'app',
                 'rxjs':                                 'node_modules/rxjs',
+                'dompurify':                            'node_modules/dompurify/dist/purify.min.js',
                 '@angular/animations':                  'node_modules/@angular/animations/bundles/animations.umd.js',
                 '@angular/animations/browser':          'node_modules/@angular/animations/bundles/animations-browser.umd.js',
                 '@angular/core':                        'node_modules/@angular/core/bundles/core.umd.js',
--- a/src/clarity-icons/clarity-icons-api.ts
+++ b/src/clarity-icons/clarity-icons-api.ts
@@ -6,6 +6,7 @@
 
 import { IconTemplate } from "./interfaces/icon-template";
 import { IconAlias } from "./interfaces/icon-alias";
+import * as DomPurify from "dompurify";
 
 let iconShapeSources: IconTemplate = {};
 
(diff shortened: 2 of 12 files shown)


src/clarity-icons/clarity-icons-element.ts
```diff
@@ -36,52 +33,49 @@ ClarityIconElement.prototype.constructor = ClarityIconElement;
 
 let generateIcon =
     function (element: any, shape: string) {
+        let cIcons = ClarityIconsApi.instance;
```
Shijir: I think it's better to set `cIcons` variable outside of `generateIcon` function unless there is a good reason to have it inside here. Otherwise, currently, we are accessing the common `ClarityIconsApi.instance` and setting `cIcons ` to it on every single icon.

Can we also rename this `cIcons` variable? As it's specifically referring to the `ClarityIconsApi.instance`, could we perhaps use `clrIconsApi` or something like that 😛 ?

src/clarity-icons/clarity-icons-api.ts
```diff
@@ -17,64 +18,60 @@ export class ClarityIconsApi {
...
+            "feFuncB", "feFuncG", "feFuncR", "feGaussianBlur", "feImage", "feMerge", "feMergeNode",
+            "feMorphology", "feOffset", "fePointLight", "feSpecularLighting", "feSpotLight",
+            "feTile", "feTurbulence", "filter", "g", "line", "linearGradient", "marker",
+            "mask", "mpath", "path", "pattern", "polygon", "polyline", "radialGradient", "rect",
+            "stop", "symbol", "text", "textPath", "title", "use", "view"
+        ];
+
+        const sanitizeOptions = {
+            SAFE_FOR_TEMPLATES: true,
+            FORBID_ATTR: ["style"],
```
Shijir: Do we need to forbid `style` attributes? I think allowing users to have `style` attribute could enable them to have greater customizability and control over their icons' look and feel.

src/clarity-icons/clarity-icons-api.ts
```diff
@@ -17,64 +18,60 @@ export class ClarityIconsApi {
...
-
             throw new Error("Shape name or alias must not contain any whitespace characters!");
-
         }
 
         return true;
-
     }
 
-    private validateTemplate(template: string): boolean {
```
Shijir: Forgot to say I ❤️  this refactoring.

src/app/index.html
```diff
@@ -90,6 +92,7 @@
       window.addEventListener("load", function() {
 
         System.import('app/boot').then(null, console.error.bind(console));
+        System.import('dompurify').then(null, console.error.bind(console));
```
youdz: I don't understand this line. We already
```typescript
import * as DomPurify from "dompurify";
```
in `clarity-icons-api.ts`, why do we need to manually import DomPurify from the page itself? If we actually _need_ it, then this is an issue because our users would need it too.

src/clarity-icons/clarity-icons-element.ts
mathisscott: Yup. I'll do that. Thanks!

src/clarity-icons/clarity-icons-api.ts
mathisscott: Yes, sadly. There's way too many harmful things that can be done with style attrs. The recommendation is to load your SVG in an img tag if it requires styles. Or move those styles up into your own CSS files.

src/app/index.html
mathisscott: I don't know. I can try removing that line and see if it works as expected.

src/clarity-icons/clarity-icons-element.ts
mathisscott: I decided to revert that variable. It was only a convenience for three lookups below. So there was no point in pushing it out of the generateIcon function. So I removed cIcons and replaced it with the more verbose ClarityIconsApi.instance.

src/app/index.html
mathisscott: @youdz : I think this was left in from an early attempt to get this working. I've removed that line and everything still works fine. App runs. Tests run...

PR #992, opened 2017-05-30
[NG] Adding input to Wizard that prevents all navigation

mathisscott:
• The use case here is to disable navigation across the whole wizard for a long-running process
• Added clrWizardPreventNavigation input that disables all buttons, stepnav links, and even prevents closing the wizard
• Prevents firing of events as well, so cannot be used to implement alternative flows
• Fixed a naming collision issue in the deprecated wizard demos
• Also added a demo of the stop navigation functionality
• When working on the demo, I noticed we were needing an additional path to just disable the stepnav by itself. So I did that and wrote tests for it!

[NG] unit tests

[NG] Aditya’s fix for unit tests

[NG] Demo

• This added a demo of the stop navigation functionality
• When working on the demo, I noticed we were needing an additional path to just disable the stepnav by itself. So I did that and wrote tests for it!


adityarb88:
@mathisscott: not sure why but some of the CSS regression images are failing on this PR.

mathisscott:
@adityarb88 
I know. I'm looking into it.

src/clarity-angular/wizard/wizard-stepnav-item.ts
```diff
@@ -49,7 +49,7 @@ export class WizardStepnavItem {
 
     public get isDisabled(): boolean {
         this.pageGuard();
-        return this.page.disabled;
+        return this.page.disabled || this.navService.wizardStopNavigation || this.navService.wizardDisableStepnav;
```
hippee-lee: These will always have a value right?
IIR correctly this evaluates each expression and returns the first true value or the last false value. But, if they are all null it will return that.

mathisscott: The navService values init with `false`. So at the very worst, you get `false`.

PR #993, opened 2017-05-31
[NG] Add change detector ref to footer

### LINKED ISSUE #950: Datagrid when using *clrDgHideableColumn in hidden div throws console error.
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Expected not to throw errors in browser console.

### Actual behavior
I'm using Datagrid with *clrDgHideableColumn inside a div which is not visible (*ngIf="false").
When showing the containing div, an error is thrown in the browser console.

```
ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'.
```

### Reproduction of behavior

https://plnkr.co/pWs4dd

### Environment details

* **Angular version:** 4.0.X

* **Clarity version:** 0.9.4

* **OS and version:** Ubuntu 16.4

* **Browser:** [ all ]

hippee-lee:
- closes #950
- Fixes issue with  component when using  on the datagrid itself that casued  error

I had to make a change to the footer spec related to using ChangeDetectorRef to manage updates to the child component. I'm not 100% sure its the right way to use it in testing. 

Tested in Chrome, Firefox, Safari, Edge & IE11

### DIFF
--- a/src/clarity-angular/data/datagrid/datagrid-column-toggle.ts
+++ b/src/clarity-angular/data/datagrid/datagrid-column-toggle.ts
@@ -75,7 +75,6 @@ export class DatagridColumnToggle implements OnInit, OnDestroy {
      * @type {Array}
      */
     public columns: DatagridHideableColumn[] = [];
-    // public lastColumnshowing: boolean = false;
 
     public get allColumnsVisible(): boolean {
         return this._allColumnsVisible;
@@ -102,11 +101,11 @@ export class DatagridColumnToggle implements OnInit, OnDestroy {
                     this.columns.push(col);
                 }
             });
-        });
+            });
     }
 
     ngOnDestroy() {
-        // this._hideableColumnChangeSubscription.unsubscribe();
+        this._hideableColumnChangeSubscription.unsubscribe();
     }
 
     selectAll() {
--- a/src/app/datagrid/selection/selection.html
+++ b/src/app/datagrid/selection/selection.html
@@ -70,9 +70,7 @@
     <clr-dg-column>Name</clr-dg-column>
     <clr-dg-column>Creation date</clr-dg-column>
     <clr-dg-column>
-        <ng-container *clrDgHideableColumn>
-            Favorite color
-        </ng-container>
+        Favorite color
     </clr-dg-column>
 
     <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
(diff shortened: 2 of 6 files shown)


src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -12,49 +12,49 @@
...
-        <ng-container *clrDgHideableColumn>
-            Pokemon
-        </ng-container>
-    </clr-dg-column>
-    <clr-dg-column>
-        <ng-container *clrDgHideableColumn>
-            Favorite color
-        </ng-container>
-    </clr-dg-column>
+    <clr-datagrid>
```
adityarb88: Not sure if this diff is because of a formatting change but normally formatting changes should be done in a separate PR because it creates a longer diff not related to the actual change and just in case we revert this commit, the formatting changes are lost too. same for the other formatting changes in this file.

src/clarity-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -36,15 +36,15 @@ import { Point } from "../../popover/common/popover";
...
+                        class="btn btn-sm btn-link p6 text-uppercase"
+                        [disabled]="allColumnsVisible"
+                        (click)="selectAll()">Select All
                     </button>
                 </div>
                 <div class="action-right">
                     <button
-                            (click)="toggleUI()"
-                            class="btn btn-primary">
+                        (click)="toggleUI()"
```
adityarb88: same for the formatting changes in this file.

src/app/datagrid/hide-show-columns/hide-show.html
hippee-lee: 🤔  - I didn't catch that when I created the PR. Let me see if I can get rid of them.

hippee-lee:
@adityarb88 - I think I got them all back to what was in master.
I do want to format things but want to wait until we decide on and use the standard convention. That way I can just implement / use EditorConfig to see it my way but save it the Clarity way.

hippee-lee:
I can add one in the datagrid.specs where we are doing more comprehensive testing.
This is hard to test in isolation because we don't have a full datagrid, hideable column provider or hideable columns and the error was triggered because of an `*ngIf` on `<clr-datagrid>`.

I will file an issue and self assign.

hippee-lee:
@youdz - My thinking for `markForCheck()` was to leave the selections alone / almost as they were and let Angular take care of them with normal change detection. 

If we find its not a great pattern to implement I can update it.

PR #1017, opened 2017-06-06
[NG] Adds Signpost Component

### LINKED ISSUE #486: Signpost Help Component (New)
```
[ ] bug
[X] feature request
[ ] enhancement
```

### Description
Signposts provide contextual info/help in a popover that can be initiated or dismissed by the user

### Features
- User initiated viewing or hiding of the popover
- Headings, images, text/image links (optional)
- A vertical scroll bar (default browser) may be used if the content exceeds the max height (504 px)
- A trigger element: When clicked, the “i” icon triggers the popover into view
- Anchor arrows may be positioned on any corner or centered on any side of the popover. See diagram for placement. (May be subject to change in implementation)
- Max width is 360 px, obviating the need for responsiveness on smaller screens

### Usage
- Use to show contextual help or information 
- May be used when a tool tip is too small for the amount of content
- When the information presented needs to stay in view at length. Unlike tooltips, the user is able to mouse away without losing the view.

### Interaction
- By default, popovers should appear on the right side of the trigger element. Alternate positions may be used if the default is not possible.
- The popover may be dismissed by either clicking outside of it or by clicking the “X” icon.
- To keep the interface clean, show one popover at a time.  
- When a popover is visible, clicking an icon to open another will automatically dismiss the first one.

### Recommendations
- The popover should not be so large that it dominates the screen
- Make sure that the popover doesn’t obstruct other important information on the screen.
- With the exception of text links, don’t insert any actionable components such as buttons in the popover. 

### Basic Signpost Design
basic ex1

states

### Signpost help with header / text link / image
w_header_image

### Min/Max Width and Height (including anchor point)
Min-width  216 px;  Min-height  84 px
Max-width  360 px;  Max-height  504 px
sph-scroll

### Basic Spec
- On hover/click, the grey “i” icon (trigger element) becomes blue #007CBB. It remains blue until the popover is dismissed.
- When the “i” icon is clicked, the popover becomes visible.
spec 1

### Heading and Image Spec
H4 (Metropolis Light)
heading spec

w_image

### Icon Positioning
Align the icon 6 px to the right of the item. If a table or stack view has multiple signposts, the icon may be placed in a separate column or aligned to the right side of the grid.
icon alignment

### Anchor Point Positioning

anchor point positioning

1. Anchor placed along center line of popover
2, 3. Anchor placed on corner of popover

### Popover positioning
popover - icon position

* **Clarity version:** 

* **OS and version:** 

hippee-lee:
Gemini tests and updated demos will follow in a separate PR. 

This Includes the following:
- `clr-signpost` with default trigger 'i' icon
- `[clrSignpostPosition]="POSITION_VALUE"` 
    - signpost_positions
- `clr-signpost-content` component (that uses `*clrIfOpen` for details
- 2-way binding for de-sugared *clrIfOpen template syntax.
- closes #486

Tested on: Chrome, Safari, Firefox, Edge & IE11.


src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -0,0 +1,389 @@
...
+
+    &:hover {
+        cursor: pointer;
+    }
+
+    button {
+        min-width: $clr_baselineRem_1;
+        margin: 0 $clr_baselineRem_0_3;
+        padding: 0;
+        color: $clr-dark-midtone-gray!important;
```
adityarb88: why do we need `!important`? We usually avoid using this because it makes customizability from the user side difficult.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -0,0 +1,389 @@
...
+        outline: none;
+
+        clr-icon {
+            width: $clr_baselineRem_1;
+            height: $clr_baselineRem_1;
+        }
+
+        &:hover, &.active {
+            color: $clr-action-blue!important;
```
adityarb88: same here. why do we need `!important`?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -0,0 +1,389 @@
...
+    /***
+        The signpost is above the icon top-{HORIZONTAL_POSITION}
+    */
+    &.top-left {
+        .signpost-content {
+            .popover-pointer {
+                width: 0;
+                height: 0;
+                border-left: 12px solid transparent;
+                border-top: 12px solid #9a9a9a;
```
adityarb88: We have a color variable in `variables.clarity.scss` that we can reuse here instead of `#9a9a9a` and same for the ones below. I think you didn't use the baseline rem variables for 12px here because this is for the pointer arrows right?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -0,0 +1,389 @@
...
+            color: $clr-action-blue!important;
+        }
+    }
+
+    /***
+        The signpost is above the icon top-{HORIZONTAL_POSITION}
+    */
+    &.top-left {
+        .signpost-content {
+            .popover-pointer {
```
adityarb88: We should make a note that we will create a mixin for this later on and reuse that in tooltips too.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -0,0 +1,389 @@
...
+                width: 0;
```
adityarb88: can extract the `width` and `height` which remain `0` throughout & also the `position: absolute` right?

src/clarity-angular/popover/signpost/if-open.directive.ts
```diff
@@ -0,0 +1,112 @@
...
+     * @description
+     * A setter for the directive property _open.
+     *
+     * @param value
+     */
+    @Input("clrIfOpen")
+    public set open( value: boolean ) {
+        value = !!value;
+        this._open = value;
+        this.openChange.emit(value);
```
adityarb88: should we emit events on user inputs? I remember @youdz mentioned that we should avoid doing this but you should check with him.

src/clarity-angular/popover/signpost/if-open.directive.ts
```diff
@@ -0,0 +1,112 @@
...
+    /**********
+     * @function ngOnInit
+     *
+     * @description
+     * Angular lifecycle function that subscribes to the Observable changes to the open propoerty of the
+     * IfOpenService. When changes are heard it calls updateView with the change value.
+     */
+    ngOnInit() {
+        this.ifOpenService.openedChange.subscribe((change) => {
+            this.updateView(change);
```
adityarb88: I think that you should emit the event here but again we should check with @youdz.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I changed this to use a class so that I can override it with the natural cascade.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I changed this to use a class so that I can override it with the natural cascade.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: Added a note.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: Yes. I created `%popover-pointer` as an extendable directive.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: Yes, this is for the pointer arrows.
I'll update the color to the correct variable.

adityarb88:
Oh forgot to ask. This was tested on all browsers right?

hippee-lee:
Yeah. I forgot to put that in the notes. I'll update them.

src/clarity-angular/popover/signpost/if-open.service.ts
```diff
@@ -0,0 +1,84 @@
...
+     * @description
+     * A setter function that updates the current state of _open for this instance of IfOpen structural directive. And,
+     * broadcasts the new value to all subscribers.
+     *
+     * @param value
+     */
+    public set open(value: boolean) {
+        value = !!value;
+        this._open = value;
+        this._opened.next(value);
```
youdz: You should wrap these last two lines in `if (this._open !== value)`, to make sure you can't fire the observable without an actual change.

src/clarity-angular/popover/signpost/if-open.directive.ts
```diff
@@ -0,0 +1,112 @@
...
+    /*********
+     * @property _open
+     *
+     * @description
+     * A property that holds the open or closed state for the template implementing this directive.
+     *
+     * @type {boolean}
+     * @private
+     */
+    private _open: boolean = false;
```
youdz: The service tracks the state already, you don't want 2 sources of truth. This directive should just use `this.ifOpenService.open` when needed.

src/clarity-angular/popover/signpost/if-open.directive.ts
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @description
+     * A setter for the directive property _open.
+     *
+     * @param value
+     */
+    @Input("clrIfOpen")
+    public set open( value: boolean ) {
+        value = !!value;
+        this._open = value;
```
youdz: This input doesn't do `this.ifOpenService.open = value`, so the service doesn't have the correct state.

src/clarity-angular/popover/signpost/if-open.directive.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+                let testValue: boolean = true;
+                this.testComponent.openState = testValue;
+                this.fixture.detectChanges();
+                expect(this.clarityDirective.open).toEqual(testValue);
+            });
+
+            it("provides the current value of the open state", function() {
+                let testValue = this.clarityDirective.open;
+                expect(this.testComponent.openState).toBe(testValue);
+            });
```
youdz: I don't understand what this test does. Your component doesn't test the output at all, and this test just checks the initial state.

src/clarity-angular/popover/signpost/signpost-content.ts
```diff
@@ -0,0 +1,83 @@
...
+    /**********
+     * @function close
+     *
+     * @description
+     * Close function that uses the signpost instance to toggle the state of the content popover.
+     *
+     */
+    close() {
+        // This can only be called when a signpost content is open
+        this.signpost.toggleSignpost();
```
youdz: Why is the click on the close icon using a custom method rather than just `this.ifOpenService.open = false;`?

The point of the service is that it can be injected everywhere in the various sub-components, and that it handles both the state and all internal communications.

src/clarity-angular/popover/signpost/signpost-content.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    constructor( private signpost: Signpost, private elementRef: ElementRef ) { }
```
youdz: You probably want to inject directly the `IfOpenService` rather than use the one from the Signpost parent component. I'm pretty sure you don't even need the parent component anymore.

src/clarity-angular/popover/signpost/signpost-toggle.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    /**********
+     * @function onSignpostContentClick
+     *
+     * @description
+     * click handler for the Signpost trigger button used to hide/show SignpostContent.
+     */
+    @HostListener("click")
+    onSignpostContentClick(): void {
+        this._signpost.toggleSignpost();
```
youdz: You want to use the service itself here, not the parent component.

src/clarity-angular/popover/signpost/signpost-toggle.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+    constructor(private _signpost: Signpost) { }
+
+    /**********
+     * @function onSignpostContentClick
+     *
+     * @description
+     * click handler for the Signpost trigger button used to hide/show SignpostContent.
+     */
+    @HostListener("click")
+    onSignpostContentClick(): void {
```
youdz: Misleading method name. 😛

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -0,0 +1,347 @@
...
+     *
+     * @function open
+     *
+     * @description
+     * A getter function for the current value of the open state for the SignpostContent popover.
+     * @returns {boolean}
+     */
+    public get open(): boolean {
+        return this._open;
+    }
```
youdz: Do you still need this? I should simply use `this.ifOpenService.open` directly, no need to mirror the property.

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -0,0 +1,347 @@
...
+    selector: "clr-signpost",
+    template: `
+        <button
+            #anchor
+            type="button"
+            class="signpost-action btn btn-small btn-link"
+            [ngClass]="{active: open}"
+            clrSignpostToggle>
+            <clr-icon shape="info"></clr-icon>
+        </button> <!-- <==turn into ng-content to handle default signpost icon and user input. -->
```
youdz: I don't remember if this `<ng-content>` feature was part of the initial MVC or not. I'm fine pushing it for later, we just need to make sure we open an issue for it.

src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -0,0 +1,485 @@
...
+            expect(content).toBeNull();
+        });
+
+        it("closes the signpost when clicked outside host", function() {
+            let signpostToggle: HTMLElement = clarityElement.querySelector(".btn");
+            signpostToggle.click();
+            fixture.detectChanges();
+            expect(clarityElement.querySelector(".signpost-content")).not.toBeNull();
+
+            let outsideButton: HTMLElement = clarityElement.querySelector(".outside-click-test");
```
youdz: You probably want to test that the signpost doesn't close when its own content is clicked.

src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -0,0 +1,485 @@
...
+            expect(signpost.popoverPoint).toBe(Point.RIGHT_TOP);
+            expect(signpost.signpostOptions.offsetY).toBe(18);
+            expect(signpost.signpostOptions.offsetX).toBe(-14);
+        });
+    });
+
+    describe("SignpostContent", function() {
+        let fixture: ComponentFixture<any>;
+        let clarityElement: any;
+        let signpostContent: SignpostContent;
```
youdz: This part of the spec should be in `signpost-content.spec.ts`, right?

src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -0,0 +1,485 @@
...
+            fixture.detectChanges();
+
+            let testClass = clarityElement.querySelector(".top-left");
+            expect(testClass).toBeDefined();
+
+            expect(signpost.anchorPoint).toBe(Point.TOP_LEFT);
+            expect(signpost.popoverPoint).toBe(Point.BOTTOM_RIGHT);
+            expect(signpost.signpostOptions.offsetY).toBe(-8);
+            expect(signpost.signpostOptions.offsetX).toBe(10);
+        });
```
youdz: Nitpick: you can create helper functions for unit tests, this is still just Javascript (well... Typescript).

This position test is copy-pasted a bunch of time, you could just write a simple function that takes the position string, the expected points and offsets, and runs this. Only do this if you have enough time before the release, it's really not a blocker.

hippee-lee: Yeah, I got on the 3rd or 4th one and had the same thought but it conflicted with just getting the PR out for review. I'll revisit it if/when I can. 

Side note - I think I addressed all of the changes requested but I lost my comments in a page reload so let me know if anything is incorrect.

src/clarity-angular/popover/signpost/if-open.directive.specs.ts
```diff
@@ -35,9 +35,12 @@ export default function(): void {
...
 
-            it("provides the current value of the open state", function() {
-                let testValue = this.clarityDirective.open;
-                expect(this.testComponent.openState).toBe(testValue);
+            it("gets the current value of the open state", function() {
+                let testValue: boolean = this.ifOpenService.open;
+
+                expect(this.ifOpenService.open).toEqual(this.clarityDirective.open);
+                this.ifOpenService.open = !testValue;
+                expect(this.ifOpenService.open).toEqual(this.clarityDirective.open);
```
youdz: What you want here is to test the two-way binding, right?

So you should have the de-sugarized syntax `<ng-template [(clrIfOpen)]="openState">` in your test component, and then check the `testComponent` itself:
```typescript
this.ifOpenService.open = true;
expect(this.testComponent. openState).toEqual(true);
this.ifOpenService.open = false;
expect(this.testComponent. openState).toEqual(false);
```

src/clarity-angular/popover/signpost/signpost-content.ts
```diff
@@ -44,10 +44,10 @@ export class SignpostContent/* implements OnDestroy, AfterViewInit*/ {
      */
     close() {
         // This can only be called when a signpost content is open
-        this.signpost.toggleSignpost();
+        this.ifOpenService.open = !this.ifOpenService.open;
```
youdz: Then you might as well just `this.ifOpenService.open = false`, right? 😛  Easier to read. (Yes, this is a nitpick.)

src/clarity-angular/popover/signpost/signpost-trigger.directive.specs.ts
```diff
@@ -0,0 +1,32 @@
...
+                // ifOpenService = new IfOpenService();
+                // signpost = new Signpost(ifOpenService);
+                // toggler = new SignpostTriggerDirective(signpost);
+            });
+
+            xit("should call the signpost parent toggle function", function() {
+                // signpostSpy = spyOn(signpost, "toggleSignpost");
+                // toggler.onSignpostContentClick();
+                // expect(signpostSpy).toHaveBeenCalled();
+            });
```
youdz: What's with these tests all commented out?

hippee-lee: Merge gone wrong. I'll find the tests and update.

src/clarity-angular/popover/signpost/if-open.directive.specs.ts
hippee-lee: I will update this.

src/clarity-angular/popover/signpost/if-open.directive.ts
hippee-lee: Yes. This was where he recommended.

PR #1027, opened 2017-06-08
[CORE] Remove the html reporter from gemini test command.

hippee-lee:
It looks like Gemini removed the html reporter recently: 
https://github.com/gemini-testing/gemini/commit/e2613e4c80f013902f3f8df3a92e0bd37fa1d2c4

I'm not sure if we need/use the html reporter but this PR will allow builds to run gemini tests again until we change the gemini setup.


hippee-lee:
@jeeyun - I would like to discuss this when you get back to make sure its inline with what you want for gemini.

PR #1035, opened 2017-06-12
[NG] Fixes HideableColumnService error when looking up by undefined id

### LINKED ISSUE #837: [ng] HideableColumnService is throwing an error
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Actual behavior

HideableColumnService is throwing an error when you do a refresh of the page in the Datagrid selection demo page. Error log from Chrome Dev Tool  console:
```
errors.ts:42 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'id' of undefined
TypeError: Cannot read property 'id' of undefined
    at eval (http://localhost:3000/clarity-angular/datagrid/providers/hideable-column.service.js:199:75)
```
### Reproduction of behavior

Go to datagrid's selection page and refresh the browser.

### Environment details

* **Angular version:** 4.0.X

* **Clarity version:** 0.9.3

* **OS and version:** 

* **Browser:** all

hippee-lee:
- Adds a guard to getColumnById lookup
- Adds a test for getColumnById fn (both defined and undefined hideable columns
- closes #837

### DIFF
--- a/src/clarity-angular/data/datagrid/providers/hideable-column.service.ts
+++ b/src/clarity-angular/data/datagrid/providers/hideable-column.service.ts
@@ -192,9 +192,9 @@ export class HideableColumnService {
      * @returns HideableColumn
      *
      */
-    public getColumnById( id: string ): DatagridHideableColumn {
+    public getColumnById( id: string ): undefined | DatagridHideableColumn {
         if ( id ) {
-            return this._columnList.find(column => column.id === id);
+            return this._columnList.find(column => column && column.id === id);
         }
         return;
     }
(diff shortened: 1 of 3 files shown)


src/clarity-angular/data/datagrid/providers/hideable-column.service.ts
```diff
@@ -192,9 +192,9 @@ export class HideableColumnService {
      * @returns HideableColumn
      *
      */
-    public getColumnById( id: string ): DatagridHideableColumn {
+    public getColumnById( id: string ): any { // Can't be DatagridHideableColumn b/c we don't always return something.
```
mathisscott: Instead of declaring `any` here, would it make more sense to declare a union type?

```
public getColumnById( id: string ): void | DatagridHideableColumn {
    ...
}
```

hippee-lee: A wat? (j/k) I've seen them a couple of times but forgot all about them here. 
It absolutely makes sense. I'll push an update.

PR #1045, opened 2017-06-13
[CORE] Use gemini version 4.19.3

hippee-lee:
- This uses an older version of gemini until we can work out why the html-reporter plugin is not working with the v5x alpha


hippee-lee:
I have a discussion going with the folks who provided the plugin and will re-visit at a later date.
https://github.com/gemini-testing/html-reporter/issues/3

PR #1051, opened 2017-06-13
[NG] Prevented undefined sortBy to replace default comparator value

Shijir:

### DIFF
--- a/src/clarity-angular/data/datagrid/datagrid-column.ts
+++ b/src/clarity-angular/data/datagrid/datagrid-column.ts
@@ -13,7 +13,7 @@ import {
     ElementRef,
     ViewChild
 } from "@angular/core";
-import {Subscription} from "rxjs";
+import { Subscription } from "rxjs";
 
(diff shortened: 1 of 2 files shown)


src/clarity-angular/data/datagrid/datagrid-column.ts
```diff
@@ -155,13 +155,29 @@ export class DatagridColumn extends DatagridFilterRegistrar<DatagridStringFilter
...
+    }
+
+    @Input("clrDgSortBy")
+    public set sortBy(comparator: Comparator<any>) {
+        if (comparator) {
+            this._sortBy = comparator;
+        } else {
+            if (!this._field) {
+                delete this._sortBy;
+            }
```
youdz: I think you're missing a case:
```typescript
column.sortBy = myComparator;
column.field = "hello";
column.sortBy = undefined;
```
This sequence should make the column use the default `DatagridPropertyComparator` from the field, but right now at the end of this sequence, the column is still using `myComparator`, which was **not** deleted. You can write a unit test for it, you'll see. 😉

PR #1059, opened 2017-06-14
[NG] Checkbox IE Indeterminate Fix | DO NOT MERGE

adityarb88:
Fixes an issue with checkboxes in IE & Edge

Indeterminate Checkboxes when clicked behave differently on IE than other browsers.

Indeterminate Checkbox + Click = Unchecked State + No change event raised (in IE & Edge)
Indeterminate Checkbox + Click = Checked State + Change event raised (other browsers)

This leads to weird states in Trees, for eg:

Before A is clicked:
image

After A is clicked:
image

This PR makes checkboxes in IE behave like in other browsers.

Tested on: Chrome, Edge, IE, Safari, Firefox 


src/clarity-angular/forms/checkbox/checkbox.ts
```diff
@@ -55,39 +68,51 @@ export class Checkbox implements ControlValueAccessor {
...
+    @Input("clrChecked")
     public set checked(value: boolean) {
         if (value !== this._checked) {
-            this.indeterminate = false;
-            this._checked = value;
-            this.change.emit(this.checked);
+            this.setChecked(value);
+            if (this._indeterminate) {
+                this.setIndeterminate(false);
+            }
```
youdz: You're changing the order of the event emitters here, I remember vaguely the order was finely tuned because it created issues otherwise. If all the unit tests are working it should be fine, but I'm just pointing out there is a risk here.

adityarb88: Thanks for catching this. This slipped through because I was trying a different approach before. Will revert the order back to what it was before.

PR #1062, opened 2017-06-14
[CORE] Adds html-reporter plugin for gemini@latest

hippee-lee:
I figured out what the issue was, the html-reporter plug-in needs to also be installed globally (duh). 
This PR will get us back on the latest gemini with html-reporter.


hippee-lee:
This worked on my Ubuntu VM but not here. Going to close this for now.

PR #1080, opened 2017-06-16
[NG] refactor of alerts to make them use Clarity Icons

mathisscott:
• this removes the generated icons that were previously used in alerts
• it also converts the alerts to flexbox so they no longer rely on position relative/absolute
• some generate*Icon functions have been removed from SASS
• some .clr-icon-* utility classes were also removed (they were deprecated and only used internally)
• tooltips demos in demo app now use clr-icons and no longer use the generated icons

screen shot 2017-06-16 at 5 59 29 pm

[NG] wip

• got things working for basic alerts, moving on to small alerts

[NG] wip

• got regular alerts, small alerts, and responsive alerts working

[NG] wip

• app level alerts working

[NG] wip

• updated static demos
• updated code demos in angular and static demos
• updated copyright notices in all alert files

[NG] wip

• removed .clr-icon-info, .icon-info, .clr-icon-info-white utility classes
• removed generateInfoIcon function from SASS
• removed .clr-icon-success, .icon-success, .clr-icon-success-white utility classes
• removed generateSuccessIcon function from SASS
• removed .clr-icon-update utility class and generateUpdateIcon function from SASS
• removed .clr-icon-danger, .icon-danger, .clr-icon-danger-white, .icon-danger-white utility classes
• removed .clr-icon-caret utility class

• had to leave in the generateErrorIcon function in the SASS because it is still being used in the form page and login page validation

• had to leave in the generateCaretIcon function in the SASS because it is still being used in the stack view, select box, sidenav and dropdowns

• had to leave in the .clr-icon-warning, .icon-warning, .clr-icon-warning-white utility classes because they are still being used in tooltips demos

[NG] wip

• removed multiple icon generators from the CSS that were only being used in the tooltips demo
• updated tooltips demos to use Clarity Icons

[wip] ready for browser testing

[wip] browser testing and unit testing done

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ Safari
✔︎ IE11
✔︎ Edge


src/clarity-angular/emphasis/alert/alert.module.ts
```diff
@@ -21,6 +21,9 @@ import { ALERT_DIRECTIVES } from "./index";
     ],
     exports: [
         ALERT_DIRECTIVES
+    ],
+    schemas: [
```
adityarb88: not exactly sure why we need this? I don't see this being used anywhere in the project. is it needed because users can add their own icons in alerts?

src/clarity-angular/emphasis/alert/alert.ts
```diff
@@ -10,20 +10,38 @@ import {
...
 export class Alert {
+    constructor(public iconService: AlertIconAndTypesService) {
+    }
+
     @Input("clrAlertSizeSmall") isSmall: boolean = false;
-    @Input("clrAlertType") type: string = "alert-info";
     @Input("clrAlertClosable") closable: boolean = true;
     @Input("clrAlertAppLevel") isAppLevel: boolean = false;
 
+    // TODO: THIS SHOULD BE AN ANGULAR TWO-WAY BINDING...
```
adityarb88: this TODO still needed? I see that you have both Input and Output below.

src/clarity-angular/emphasis/alert/alert.module.ts
mathisscott: Without it, the module complains that it doesn't know what `clr-icon` is.

src/clarity-angular/emphasis/alert/alert.ts
mathisscott: Yeah, I can get rid of it. Thanks.

src/clarity-angular/emphasis/alert/alert.module.ts
adityarb88: Weird. It actually shouldn't because you are already importing clrIconsModule too. Thats what datagrid and other components do too.

src/clarity-angular/emphasis/alert/_alert.clarity.scss
```diff
@@ -81,30 +77,70 @@ $clr-alert-colors: (
...
     }
+
+    .alert-icon {
+        color: map-get($propMap, icon-color)
+    }
 }
 
 @include exports('alert.clarity') {
-    %icon-styles {
+    $alertIconWidth: $clr_baselineRem_1 + 1;
```
hippee-lee: Question for knowledge xfer: What does adding 1px to the alert icon width do for us?

src/clarity-angular/emphasis/alert/alert.module.ts
youdz: @mathisscott: Importing `ClrIconModule` should be enough. Unless you're expecting consumers to write `<clr-icon>` in their own template, in which case you might want to also export `ClrIconModule`. But this schema is actually a blocker, it hides too many errors to ship with it.

src/clarity-angular/emphasis/alert/alert.html
```diff
@@ -1,17 +1,19 @@
...
-    [ngClass]="alertType"
+    [ngClass]="alertClass"
     [class.alert-sm]="isSmall"
     [class.alert-app-level]="isAppLevel">
+    <div class="alert-items">
+        <ng-content></ng-content>
+    </div>
     <button type="button" class="close" aria-label="Close" *ngIf="closable" (click)="close()">
         <clr-icon aria-hidden="true" shape="close"></clr-icon>
     </button>
```
youdz: This might be a stupid question, but does the order really matter? We could simply use `flex-order: 100` on the `.close` button and it would always be the last element in the flexbox, no matter the order of the actual DOM.

src/clarity-angular/emphasis/alert/index.ts
```diff
@@ -1,12 +1,15 @@
...
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+
 import { Type } from "@angular/core";
 import { Alert } from "./alert";
+import { AlertItem } from "./alert-item";
 
 export * from "./alert";
```
youdz: The standard is to also `export * from "./alert-item"` here. Otherwise consumers have no way to get the actual `AlertItem` class if they need to.

src/clarity-angular/emphasis/alert/alert.ts
```diff
@@ -10,20 +10,41 @@ import {
...
+        this.iconService.alertType = val;
+    }
+    get alertType(): string {
+        return this.iconService.alertType;
+    }
+
+    @Input("clrAlertIcon")
+    set alertIconShape(value: string) {
+        this.iconService.alertIconShape = value;
+    }
```
youdz: Ok, so I want to prefix what I'm about to write by saying I'm ok with this PR. Barring a few changes I posted in this review, I think it's good to go.

But I don't think forcing Clarity icons here is the best way to go. We've always tried to stay open and allow people to project anything most of the time, so that would be more consistent with other components. I'm a bit scared of the precedent this sets.

src/clarity-angular/emphasis/alert/providers/icon-and-types-service.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (this.alertTypes.indexOf(val) > -1) {
+            this._alertType = val;
+        }
+    }
+
+    // alert-* types are deprecated!
+    private alertTypes: string[] = [
+        "alert-info", "alert-warning", "alert-danger", "alert-success",
+        "info", "warning", "danger", "success"
+    ];
```
youdz: Please make this a constant at the module level, right now you're instantiating a different array for every single alert in the app.

src/clarity-angular/emphasis/alert/providers/icon-and-types-service.ts
```diff
@@ -0,0 +1,72 @@
...
+                returnShape = "check-circle";
+                returnClass = "alert-success";
+                break;
+            default:
+                returnShape = this.defaultIconShape;
+                returnClass = "alert-info";
+                break;
+        }
+
+        return "shape" === classOrShape ? returnShape : returnClass;
```
youdz: I'd feel more comfortable if you simply returned
```typescript
{
  shape: returnShape,
  cssClass: returnClass
}
```
Relying on a 2-value string as an argument is a hack, in particular it can't be typechecked. Just having callers do:
```typescript
service.iconInfoFromType("warning").shape
```
is strictly less characters (one space and 2 quotes less, to be precise :trollface: ) and throws errors at compile time as opposed to runtime.

src/clarity-angular/emphasis/alert/alert.html
mathisscott: Good point. I can make that change.

Also, with the schema comment, I'll find a way to take that out.

src/clarity-angular/emphasis/alert/index.ts
mathisscott: Will do

src/clarity-angular/emphasis/alert/alert.ts
mathisscott: I will have to think about this one. Mainly, I did this so that this would not be a breaking change for people using `clr-alert` today. If we want people to project anything into the icon wrapper, that's a little different from what clarity alerts have always done. At least now, they can use any custom icon they want.

I would say that this suggestion would require a templateRef in the DOM and potentially a fallback to clarity icons. I don't know if that's something we can manage in this changeset.

src/clarity-angular/emphasis/alert/providers/icon-and-types-service.ts
mathisscott: I agree. Good suggestion. I was trying not to duplicate the logic in two different functions.

src/clarity-angular/emphasis/alert/_alert.clarity.scss
mathisscott: Because the icons are a very explicit size in the specs and due to the nature of needing to have them centered in the min-height of the alert, I had to do some pixel pushing. You'll notice that the icon itself is larger than its wrapper by a decent amount. The sizing, padding, and negative margins help keep everything lined up.

mathisscott: I'm counting the days until CSS grid is widely adopted. This whole task would've been much easier with CSS grid.

src/clarity-angular/emphasis/alert/alert.ts
mathisscott: This one I'm not going to change. We already have components where people have to use Clarity Icons. I will open a github issue to allow for full customization of putting whatever anyone wants in that container, but it _would_ involve a templateRef to stamp in the content ala Wizard page buttons or whatnot.

Right now, users can create a Clarity Icon with any html they want using `ClarityIcons.add()` and inject content that way.

youdz: As I said in this comment, I did not expect you to change this for this PR. I was raising a concern, and giving you a heads up for a potential future issue opened on this.

PR #1083, opened 2017-06-19
[ICONS] Chore: remove deprecated items for 0.10.0 release

hippee-lee:
This removes deprecated items from `ClarityIcons`. 
This has breaking changes in it and is part of the v0.10.0 release. 
## CSS Classes Removed
1. **.logo and .clr-icon** from .header > .branding in _header.clarity
2. **.logo and .clr-icon** from `.main-container > header > .header-hamburger-trigger + .branding` in _responsive-nav.clarity
3. **.logo and .clr-icon** from `.main-container > header > .header-nav + .header-overflow-trigger` in _responsive-nav.clarity


jeeyun:
Doesn't our documentation follow the old bootstrap class names for the buttons? Is this considered a breaking change then?

src/clarity-angular/layout/nav/_header.clarity.scss
```diff
@@ -375,26 +362,12 @@ $clr-header-nav-text-horizontal-padding: $clr_baselineRem_1;
                 //for the height of the header.
                 .dropdown-menu {
                     margin-top: -1 * $clr_baselineRem_0_375;
-
-                    .dropdown-menu {
```
hippee-lee: @jeeyun - just double checking that this is the correct itme to remove this. Can you confirm that we no longer need it?

src/clarity-angular/layout/nav/_responsive-nav.clarity.scss
```diff
@@ -275,13 +275,6 @@ $clr-transition-style: ease;
 
                 .header-hamburger-trigger + .branding {
                     padding-left: 0;
-
-                    //deprecate .clr-icon, .logo
-                    .clr-icon,
-                    .logo,
-                    clr-icon {
```
mathisscott: Do we want to get rid of `clr-icon` here?

hippee-lee: I think we need to keep it for now. Its going to be a breaking change for anyone using that markup structure. We had a slack discussion about not removing those things until we do the css standardization work.

mathisscott: i agree. but are we putting it back? right now, it looks like it was deleted.

hippee-lee: Duh, thats red, not green. I'll put it back.

PR #1087, opened 2017-06-20
[ng] Reusable *clrIfOpen structural directive (#891)

### LINKED ISSUE #891: [NG] Create a reusable *clrIfOpen structural directive

jeeyun:
Reusable `clrIfOpen` structural directive, eventually to be used by all of our components with popovers. This PR only includes:

- factoring out `clrIfOpen` to a new utils package
- converting Dropdown to use the new popover (attribute directive)
- renaming old popover structural directive
- refactoring other components to use the renamed old directive

Fixes #891.


src/clarity-angular/popover/dropdown/dropdown-toggle.ts
```diff
@@ -30,7 +30,7 @@ export class DropdownToggle {
     }
 
     get active(): boolean {
-        return this._dropdown.open;
+        return this._dropdown.ifOpenService.open;
```
youdz: I'd rather you inject the ifOpenService directly, even if we still need the dropdown itself for other features. Very small change, and probably subjective, but it "feels" cleaner.

src/clarity-angular/popover/dropdown/dropdown-toggle.ts
```diff
@@ -30,7 +30,7 @@ export class DropdownToggle {
...
     }
 
     @HostListener("click")
```
youdz: This HostListener should just `ifOpenService.open = !ifOpenService.open`. It shouldn't delegate to the dropdown.

src/clarity-angular/popover/dropdown/dropdown.module.ts
```diff
@@ -9,12 +9,14 @@ import { CommonModule } from "@angular/common";
...
 import { ClrIconModule } from "../../icon/icon.module";
 import { DROPDOWN_DIRECTIVES } from "./index";
+import { ClrConditonalModule } from "../../utils/conditional/conditional.module";
 
 @NgModule({
     imports: [
         CommonModule,
         ClrCommonPopoverModule,
-        ClrIconModule
+        ClrIconModule,
```
youdz: Why do you need to import the IconModule? You're never using a `<clr-icon>` in one of the templates, right?
But it does make sense to **export** it from this module (without importing it) if consumers need to manually write an icon every time.

src/clarity-angular/popover/dropdown/dropdown.ts
```diff
@@ -16,26 +16,28 @@ import {
...
 
 // TODO: the ng-content inside the ng-template should ideally just be
 // <ng-content select="clr-dropdown-menu"></ng-content>. Remove .dropdown-menu in 1.0?
 @Component({
     selector: "clr-dropdown",
     template: `
         <ng-content select="[clrDropdownToggle]"></ng-content>
-        <ng-template [(clrPopover)]="open" [clrPopoverAnchor]="anchor" [clrPopoverAnchorPoint]="anchorPoint"
-                     [clrPopoverPopoverPoint]="popoverPoint" [clrPopoverOptions]="popoverOptions">
+        <div class="dropdown-menu-wrapper" [clrPopoverAnchor]="anchor" [clrPopoverAnchorPoint]="anchorPoint"
```
youdz: This is one of the things I'd like to revisit in a little while once Angular has more features. Right now an extra wrapper is the only way for us to use the Popover directive, but in the future we might be able to dynamically add it to the host of the DropdownMenu component.

Just a note for later.

src/clarity-angular/popover/dropdown/dropdown.module.ts
```diff
@@ -9,12 +9,14 @@ import { CommonModule } from "@angular/common";
...
+        ClrConditonalModule
```
youdz: Same thing, but for this one I'm 100% sure. We don't want to import `ClrConditonalModule`, we want to export it. That way consumers importing `ClrDropdownModule` don't need to import anything else, they get all the dependencies automatically. That's the point of dependencies management. 😛

src/clarity-angular/popover/dropdown/dropdown.spec.ts
```diff
@@ -19,8 +21,9 @@ export default function(): void {
 
         beforeEach(() => {
             TestBed.configureTestingModule({
-                imports: [ClrDropdownModule, ClrIconModule],
-                declarations: [TestComponent]
+                imports: [ClrDropdownModule, ClrIconModule, ClrConditonalModule],
```
youdz: So you shouldn't need to import anything other than `ClrDropdownModule` here. That'd be more representative of what we want our consumers to do.

src/clarity-angular/popover/signpost/signpost.module.ts
```diff
@@ -9,12 +9,14 @@ import {CommonModule} from "@angular/common";
...
 import {SIGNPOST_DIRECTIVES} from "./index";
+import {ClrConditonalModule} from "../../utils/conditional/conditional.module";
 
 @NgModule({
     imports: [
         CommonModule,
         ClrCommonPopoverModule,
-        ClrIconModule
+        ClrIconModule,
+        ClrConditonalModule
```
youdz: Same thing here, except in this case we do want the `ClrIconModule` imported. But the `ClrConditonalModule` should be exported, not imported.

src/clarity-angular/utils/conditional/all.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+import IfOpenDirectiveSpecs from "./if-open.directive.specs";
+import IfOpenServiceSpecs from "./if-open.service.specs";
+
+describe("Conditional Utils", function() {
+    describe("Services", function() {
+        IfOpenServiceSpecs();
+    });
+    describe("Directives", function() {
+        IfOpenDirectiveSpecs();
+    });
```
youdz: I think it makes more sense to group by feature rather than by service/directive. 😛 

Doesn't matter anyway, it's just the order in the reporter, so it's subjective and doesn't break anything.

src/clarity-angular/popover/common/popover.directive.spec.ts
```diff
@@ -15,111 +16,45 @@ describe("Popover directive", () => {
...
 
-    it("queues up opening of subsequent popovers if one is already open", function () {
-        fixture = TestBed.createComponent(TestComponent);
-        compiled = fixture.nativeElement;
-
-        fixture.componentInstance.open1 = true;
-        fixture.componentInstance.open2 = true;
-        fixture.componentInstance.open3 = true;
-        fixture.detectChanges();
+    it("positions popover only if the open property of ifOpenService is true", function () {
```
jeeyun: All the actual positioning is done by `popover.ts` and is covered by `popover.spec.ts`.

src/clarity-angular/utils/conditional/conditional.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import {NgModule} from "@angular/core";
+import {CommonModule} from "@angular/common";
+import {CONDITIONAL_DIRECTIVES} from "./index";
+
+@NgModule({
+    imports: [CommonModule],
+    declarations: [CONDITIONAL_DIRECTIVES],
+    exports: [CONDITIONAL_DIRECTIVES]
+})
+export class ClrConditonalModule {
```
jeeyun: Oops. Corrected.

src/clarity-angular/popover/dropdown/dropdown-toggle.ts
jeeyun: Agreed.

src/clarity-angular/popover/dropdown/dropdown-toggle.ts
jeeyun: Actually this getter is only being used by its own template to set the active class on the host element.. I've made changes so that it gets the value directly from the injected service.

src/clarity-angular/popover/dropdown/dropdown.module.ts
jeeyun: You are right, we aren't using it in the template. But since our example uses it to render the caret in the consuming side, I will export it.

src/clarity-angular/popover/common/popover.directive.spec.ts
youdz: So we're good. 👍  Thanks.

src/clarity-angular/popover/dropdown/dropdown-toggle.ts
youdz: @jeeyun: I walk talking about the click listener. I couldn't comment on the exact line because you didn't modify it. 😉

jeeyun: I was wrong; it's using the getter so I've just edited it to use the injected service.

src/clarity-angular/popover/dropdown/dropdown.spec.ts
jeeyun: Thanks for catching that. Updated the spec.

src/clarity-angular/utils/conditional/all.spec.ts
jeeyun: Regrouped.

src/clarity-angular/popover/dropdown/dropdown-toggle.ts
jeeyun: Updated the HostListener and got rid of the toggle method in dropdown!

PR #1097, opened 2017-06-22
[NG] Adds custom trigger to signposts

### LINKED ISSUE #1032: Add ability to use trigger for a Signpost trigger
```
[ ] bug
[ ] feature request
[x ] enhancement
```

Add `clr-signpost-trigger` as an optional component for users to specify which ClarityIcon they want to use for their signpost

## Default Icon (Shows the `info` icon when no declared in the template)
```
<clr-signpost>
    <clr-signpost-content *clrIfOpen="openState">
        <h3 style="margin-top: 0">Default Signpost</h3>
        <p>Position: <code class="clr-code">right-middle</code></p>
    </clr-signpost-content>
</clr-signpost>
```

## Specified Icon(shows the Clarity Icon specified as the trigger)
```
<clr-signpost>
    <clr-signpost-trigger shape="help-info"></clr-icon-trigger>
    <clr-signpost-content *clrIfOpen="openState">
        <h3 style="margin-top: 0">Default Signpost</h3>
        <p>Position: <code class="clr-code">right-middle</code></p>
    </clr-signpost-content>
</clr-signpost>
```

hippee-lee:
This adds the following for Signposts:
- A Gemini testing for signpost
- Update to signpost tests for new position values
- Functionality to declare a custom signpost trigger
- Changes Signpost template to use new Popover attribute directive instead of old structural directive
- closes #1032


src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -11,385 +11,503 @@ import { Signpost } from "./signpost";
...
+            });
 
-        it("takes an input for position", function() {
-            fixture.componentInstance.position = "top-middle";
-            fixture.detectChanges();
-            expect(signpost.signpostPosition).toBe("top-middle");
-        });
+            it("does not allow multiple open poovers", function() {
+                expect(signpost.signpostOptions.allowMultipleOpen).toBe(false);
+            });
```
mathisscott: As part of this spec, do we want to verify that if I click on one popover and then another, only one actually is open? Or do we feel this is covered adequately in another spec?

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -211,33 +230,32 @@ export class Signpost {
...
         }
         // // set the popover values based on menu position
         switch ( this._signpostPostion ) {
             case ("top-left"):
-                this.anchorPoint = Point.TOP_LEFT;
+                this.anchorPoint = Point.TOP_CENTER;
                 this.popoverPoint = Point.BOTTOM_RIGHT;
-                this.signpostOptions.offsetY = -8;
-                this.signpostOptions.offsetX = 10;
+                this.signpostOptions.offsetY = -10;
```
mathisscott: It seems that storing these offset values in an enum might be easier to maintain over time. Would you agree?

This isn't something I would block the PR on. But could you add a `TODO` for it? It would make a great `contributions-welcome` task/chore...

src/clarity-angular/popover/signpost/signpost.specs.ts
hippee-lee: I don't have a test for that. I can add one here.

src/clarity-angular/popover/signpost/signpost.ts
hippee-lee: For these three I could. The concern I have is that only with the custom trigger changes did these three values und up being the same. 

I'll see if I can abstract the commonality.

src/clarity-angular/popover/signpost/signpost.specs.ts
hippee-lee: Just re-read the spec and I have a test that makes sure we configure the popover to `            expect(signpost.signpostOptions.allowMultipleOpen).toBe(false);` [here](https://github.com/vmware/clarity/blob/master/src/clarity-angular/popover/signpost/signpost.specs.ts#L76)

Since this is tested in the popover spec I think it covers it. Are you ok with that?

mathisscott: Absolutely. Thanks for tracking that down.

src/clarity-angular/popover/signpost/signpost.ts
mathisscott: Yeah. I don't consider it a blocker. But generally when I'm doing things like this, @youdz has me put the values in an enum.

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -152,10 +145,27 @@ export class Signpost {
...
+
+    /*********
+     * @function ngAfterContentInit
+     *
+     * @description
+     * Angular lifecycle hook to set the signpost trigger
+     */
+    ngAfterContentInit() {
+        this.setTrigger(this.signpostTrigger);
+    }
```
youdz: So this whole `setTrigger()` in `ngAfterContentInit()` thing is a bit obfuscated, and as such it has a bug: if I start by providing a custom trigger, then remove it (for instance with `*ngIf`), the `useCustomTrigger` property will never revert back to false.

There is a much more straightforward solution:
```typescript
public useCustomTrigger: boolean = false;

@ContentChild(SignpostTriggerDirective) 
set customTrigger(trigger: SignpostTriggerDirective) {
  this.useCustomTrigger = !!trigger;
}
```

src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -11,385 +11,503 @@ import { Signpost } from "./signpost";
...
+                fixture.destroy();
+            });
 
-        });
+            function testPosition( sp: Signpost, test: TestPosition): void {
+                expect(sp.anchorPoint).toBe(test.anchorPoint);
+                expect(sp.popoverPoint).toBe(test.popoverPoint);
+                expect(sp.signpostOptions.offsetY).toBe(test.offsetY);
+                expect(sp.signpostOptions.offsetX).toBe(test.offsetX);
+            }
```
youdz: Good first step to reduce code duplication, now take it even further! 😛

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -26,14 +29,19 @@ const signpostPositions: string[] = [
...
+                    type="button"
+                    class="signpost-action btn btn-small btn-link"
+                    [ngClass]="{active: open}"
+                    clrSignpostTrigger>
+                    <clr-icon shape="info"></clr-icon>
+                </button>
+            </ng-container>
+        </div>
+
         <ng-template [(clrPopover)]="ifOpenService.open"
```
youdz: Oh, I just realized this is pre-Jeeyun's PR. You're going to have a few conflicts.

src/clarity-angular/popover/signpost/signpost.specs.ts
```diff
@@ -11,385 +11,503 @@ import { Signpost } from "./signpost";
...
+
+            it("projects a custom trigger element to hide/show content", function() {
+                let signpostTrigger: HTMLElement =
+                    clarityElement.querySelector(".signpost-trigger > .signpost-action");
+                let signpostContent: HTMLElement;
+
+                expect(signpostTrigger.textContent.trim()).toBe("Custom trigger");
+
+                // Test we have a trigger
+                expect(signpostTrigger).not.toBeNull();
```
youdz: Any way you could check that the default trigger is not also displayed? I could see this happening after a refactor or Angular breaking change and us not catching it.

src/clarity-angular/popover/signpost/signpost-positions.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+
+export class SignpostPositions {
+    public default: Position = {
+        anchorPoint: Point.RIGHT_CENTER,
+        popoverPoint: Point.LEFT_CENTER,
+        offsetY: 6,
+        offsetX: 14
+    };
```
youdz: So the problem here is that you're declaring a big constant, but you have to instantiate it every time so each popover creates a dozen different objects in memory. You want a shared constant.
On top of that, if you used the class names as keys, you could drastically reduce the bloating of the code in the component itself. So I recommend doing something like:

```typescript
export const SIGNPOST_POSITIONS: { [input: string]: Position } = {
    "top-left": {
        anchorPoint: Point.TOP_CENTER,
        popoverPoint: Point.BOTTOM_RIGHT,
        offsetY: -10,
        offsetX: 0
    },
    ...
}
```

src/clarity-angular/popover/signpost/signpost.ts
```diff
@@ -235,82 +245,82 @@ export class Signpost implements AfterContentInit {
...
             case ("top-left"):
-                this.anchorPoint = Point.TOP_CENTER;
-                this.popoverPoint = Point.BOTTOM_RIGHT;
-                this.signpostOptions.offsetY = -10;
-                this.signpostOptions.offsetX = 0;
+                this.anchorPoint = this.positions.topLeft.anchorPoint;
+                this.popoverPoint = this.positions.topLeft.popoverPoint;
+                this.signpostOptions.offsetY = this.positions.topLeft.offsetY;
+                this.signpostOptions.offsetX = this.positions.topLeft.offsetX;
                 break;
```
youdz: And this 80-lines switch becomes:
```typescript
let position = SIGNPOST_POSITIONS[this._signpostPostion];
this.anchorPoint = position.anchorPoint;
this.popoverPoint = position.popoverPoint;
this.signpostOptions.offsetY = position.offsetY;
this.signpostOptions.offsetX = position.offsetX;
```

PR #1110, opened 2017-06-26
[WEB] bugfix: Code Highlight example not work

dna2github:
- '{' and '}' make code highlight example break (angular reporst template syntax error)
- use '&#' style to escape

Signed-off-by: Seven Lju <dna2gle@gmail.com>

### DIFF
--- a/src/app/documentation/demos/code/code-highlight-snippet.ts
+++ b/src/app/documentation/demos/code/code-highlight-snippet.ts
@@ -6,15 +6,16 @@
 import {Component} from "@angular/core";
 
 const HTML_EXAMPLE = `
+<!-- "&#123;" left curly brace / "&#125;" right curly brace -->
 <pre>
     <code clr-code-highlight="language-css">
-        .some-component {
+        .some-component &#123;
             display: flex;
             flex-direction: column;
             justify-content: space-between;
             align-items: center;
             flex-wrap: wrap;
-        }
+        &#125;
     </code>
 </pre>
 `;


src/app/documentation/demos/code/code-highlight-snippet.ts
```diff
@@ -6,15 +6,16 @@
 import {Component} from "@angular/core";
 
 const HTML_EXAMPLE = `
+<!-- "&#123;" left curly brace / "&#125;" right curly brace -->
 <pre>
     <code clr-code-highlight="language-css">
-        .some-component {
+        .some-component &#123;
```
adityarb88: @dna2github: The code highlight example does work. Not exactly sure what the issue is here. Could you please provide more details. If you are saying that the CSS is not being highlighted its because the language we are highlight is html. The pre code blocks are getting highlighted. Also, you dont have to escape/encode the `{` `}` in a string. This change will make the code snippet look like this:

image

dna2github:
The error:

```
Uncaught Error: Template parse errors:
Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ '{' }}") to escape it.) ("
        }
    </code>
</pre>
[ERROR ->]"): ng:///AppModule/xxx.html@42:0
Invalid ICU message. Missing '}'. ("
        }
    </code>
</pre>
[ERROR ->]"): ng:///AppModule/xxx.html@42:0
```

src/app/documentation/demos/code/code-highlight-snippet.ts
dna2github: The error:
```
Uncaught Error: Template parse errors:
Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ '{' }}") to escape it.) ("
        }
    </code>
</pre>
[ERROR ->]"): ng:///AppModule/xxx.html@42:0
Invalid ICU message. Missing '}'. ("
        }
    </code>
</pre>
[ERROR ->]"): ng:///AppModule/xxx.html@42:0
```

adityarb88: I am not exactly sure how you are getting that error. We don't get that error. Also the `new-website` branch is for internal use only. Its used for publishing our website. Why are you using that branch?

dna2github: how:
1. clone `clarity-seed`, npm install
2. edit `about.component.html`, insert the code highlight example html
3. run `ng serve`
4. open browser, visit http://localhost:4200/about; watch console
=> see the error.

I am a VMware employer and just read the documentation and try all of them to have a check. Just thinking about how to do G11N work on the site and to be more familiar with Clarity, I tried the documentation.

adityarb88: sorry, i understand the issue now. there are multiple ways to escape `{` `}` but we dont in particular recommend one. for eg: you could also use `ngNonBindable` in some cases. let me check with the team to see if we need to update the code highlight documentation to reflect that. thanks a lot!

dna2github: thx for consideration. and seems ngNonBindable not work in angular 4 (tested). current i know two ways: `{{ '{' }}` or `&#123;` for escape. i tried `{{ '{' }}` and not work for this case.

adityarb88: @dna2github: thanks for pointing that out. I will check with the team and get back to you soon.

youdz:
Yes, escaping brackets in Angular is a nightmare. We decided to use interpolation for this purpose in our website. Our example is poorly chosen here, and we should use something else. But we cannot recommend one specific way of escaping brackets, because all of them have issues and compromises.

We should probably change our example to something without curly brackets, to avoid the free trolling.

dna2github:
@youdz yes, <code `[innerHTML]=variable`></code> and define `variable` in ts file is a choice i think.

mathisscott:
@adityarb88

jeeyun:
@mathisscott since we have two approvals, I can merge this in with today's website update.

youdz:
Sure, you can merge. 

But since when has code highlight been a component we fully support for consumers? This was an internal tool that was never meant to be released, but now we apparently have to maintain our docs to showcase code escaping, we have 2 issues open for dynamic example support, ... Please make sure this does not set a precedent, and do NOT consider the current code highlight directive as a stable and supported API. Any attempt at making it dynamic will be a breaking change anyway.

Issue #1126, opened 2017-06-28
[NG] Popover positioning applies to the child element instead of the Host element


adityarb88:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

1. `position: absolute` styling should be applied to the Host element on which the `clrPopoverAnchor` directive is placed.
2. Positioning should work when `*clrIfOpen` directive is used with `clrPopoverAnchor`

### Actual behavior

Popover positioning applies to the child element instead of the host element

1. Popover positioning applies to the host element instead of the child element.
2. Popover positioning breaks when `*clrIfOpen` is used along with the `clrPopoverAnchor` directive.

Fix discussed with @youdz and @jeeyun: https://github.com/adityarb88/clarity/tree/topic/aditya/DatePicker

This doesn't work with dropdowns. Need to investigate that.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 0.10.0-alpha.2

* **OS and version:** 

* **Browser:** [all]

youdz:
Closed by #1238.

PR #1131, opened 2017-06-29
[NG] Popover Anchor Fix. DO NOT MERGE

adityarb88:
Fixes: #1126

Had to use a nasty hack. Do not merge without @youdz review.

Why use the hack?
Hack is needed to make sure that the `.dropdown-menu` does not have `position: absolute` because `.dropdown-menu-wrapper` which is only present in the Angular component already has it because of `clrPopoverAnchor`. If we don't do this, a `position: absolute` on both wrapper and menu causes the wrapper width to become 0 and all *-right positions are rendered incorrectly.

Components Tested: Dropdowns & Signposts

### DIFF
--- a/src/app/signpost/signpost.demo.html
+++ b/src/app/signpost/signpost.demo.html
@@ -4,14 +4,10 @@
     <div class="content-area">
         <div class="signpost-demo">
             <div class="signpost-item"
-                *ngFor="let position of positions">
+                 *ngFor="let position of positions">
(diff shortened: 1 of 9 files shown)


mathisscott:
@youdz
Is this something we want to merge in or should we close this PR?

youdz:
This won't go in, since we're planning a larger popovers refactor. So I'm closing for now.

Issue #1007, opened 2017-06-02
Craft Library is not for newest Library version


ehausen:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
Expected Craft Library to import without notices

### Actual behavior
Craft Library gives a notice that the imported library is for old version

### Reproduction of behavior
Install newest Skecth
Install newest Craft plugins
Download latest Clarity Craft Library
Import Clarity Craft Library

### Environment details
CRAFT LIBRARY (0.9.0)

See support article from Invision: [How to update a library created with Craft Library 1.0.5 and earlier to work with version 1.0.28 and later](https://support.invisionapp.com/hc/en-us/articles/115002209843)

reddolan:
@ehausen looking into this right now. Thanks for the heads up

reddolan:
Icebox for now while we get resolution on a resizing bug within Craft

reddolan:
Will be pushing an update out for this. There have been many small updates to the UI Template that belong in the Craft Library even though the resizing attributes have a bug.

reddolan:
Updates from the 0.9.9 UI Template were merged into Craft Library. This update fixes the versioning and adds some updates from #1077 
[Clarity-0.9.9.library.zip](https://github.com/vmware/clarity/files/1085278/Clarity-0.9.9.library.zip)

There is still the Craft Library bug with resizing which will be addressed in a future release.

reddolan:
@ehausen can you test the attached zip and see if you are still getting the error?

reddolan:
@ehausen There still seems to be an issue with the versioning on the Library. InVision is currently looking into this

reddolan:
Here is a clean Library that isn't throwing the error anymore. Please replace the zip file on the site with this one.
[Clarity-0.9.9.library.zip](https://github.com/vmware/clarity/files/1088788/Clarity-0.9.9.library.zip)

mathisscott:
We need to add the new link to the website.

reddolan:
I am not sure what you mean by "add the new link". We need to replace the zip with the one above.

mathisscott:
That is what I mean. Thanks for clarifying, @reddolan

PR #1143, opened 2017-06-30
[WEB] replacing craft library file with new download

mathisscott:
Closes: #1007


adityarb88:
@mathisscott: We have started to update the version number in the button in the get started page.

image

Not sure if we need that here to update to 0.9.12?

mathisscott:
@reddolan just said to switch out the files. I can update the craft library version number if he says I need to.

reddolan:
We can keep the version number the same v0.9.9

PR #1151, opened 2017-07-05
[ng] Reusable *clrIfActive structural directive (#892)

### LINKED ISSUE #892: [NG] Create a reusable *clrIfActive structural directive
The difference between `*clrIfActive` and the other structural directives that we are introducing is that several instances of this directive will need to communicate with a shared service and behave differently (one will create its view, the others will destroy theirs). Because of this, to be able to identify themselves while still allowing an easy integration in our components, we'll need 2 providers instead of one.

## Integration in components

I am going to takes `Tabs`, `Tab` and `TabLink` as an example because it's the simplest, but the integration code should be the same for the other components that will use this.

First, `Tabs` will want to declare the overall provider that will track the general state, specifically which tab is currently active:
```typescript
@Component({
    ...
    providers: [IfActiveService]
})
export class Tabs {
}
```
Then `Tab` will want to declare a provider that lets its subcomponents and directives (in particular, the `IfActiveDirective`) share a common identifier when communicating with the overall service. Since all we need is an identifier, a simple numeric id should be fine. This is how they'll know what tab they belong too without having to grab other components instances. For simple integration, we should provide a ready-to-use provider:
```typescript
@Component({
    ...
    providers: [IF_ACTIVE_ID_PROVIDER]
})
export class Tab {
}
```
Finally, any component that needs to interact with these two providers can inject them and use the identifier to check its current status or activate itself. For instance, a quick implementation of `TabLink` could be:
```typescript
class TabLink {
    constructor(private ifActiveService: IfActiveService, 
                @Inject(IF_ACTIVE_ID) private id: number) {}

    @HostListener("click")
    activate() {
        this.ifActiveService.active = this.id;
    }

    @HostBinding("class.active")
    get isActive() {
        return this.ifActiveService.active === this.id;
    }
}
```
Note that just like `*clrIfOpen`, the two-way binding on the current active state should not happen at the component level, but directly in the `*clrIfActive` directive. That way it'll be implemented once for all components using this pattern.

## Implementation

Ok, so how do we get the API described above?

First, the signature of shared provider that will track the current active element:
```typescript
class IfActiveService {
  active: number,
  activeChange: Observable<number>;
}
```
Obviously, the implementation is going to have some private fields and use getter and setters to simulate these public properties, but the goal is this.

Then the provider for an identifier for each "activatable" item could be implemented like this:
```typescript
let counter = 0;

export const IF_ACTIVE_ID = new OpaqueToken("IF_ACTIVE_ID");

export const IF_ACTIVE_ID_PROVIDER = {
    provide: IF_ACTIVE_ID,
    useFactory: () => ++counter
};
```
We now get to the `IfActiveDirective` itself. It will look a lot like `IfOpenDirective`, but it will need both of the providers mentioned above to determine if it should render the template or not:
```typescript
class IfActiveDirective implement OnInit {
    constructor(private ifActiveService: IfActiveService, 
                @Inject(IF_ACTIVE_ID) private id: number) {}

    ngOnInit() {
        this.ifActiveService.activeChange.subscribe(newActive => {
            this.updateView(newActive === this.id);
        });
    }
}
```
The input and output (both boolean) will be implemented like the ones in `IfOpenDirective`, directly using `this.ifActiveService.active` but now comparing with `this.id` to get the boolean.

jeeyun:
Also some minor clean up in `clrIfOpen`, such as moving its change subscription to constructor.

Fixes #892.


src/clarity-angular/layout/tabs/tab-content.spec.ts
```diff
@@ -25,18 +26,18 @@ describe("TabContent", () => {
...
 
-    it("initializes the correct property values", () => {
-        let tabContents: TabContent[] = fixture.componentInstance.tabContentChildren.toArray();
-
-        expect(tabContents[0].active).toEqual(true);
-        expect(tabContents[1].active).toEqual(false);
+    it("creates ng-templates that can be rendered through templateOutlet", () => {
+        let tabContents: HTMLElement[] = compiled.querySelectorAll("clr-tab-content");
+        expect(tabContents.length).toEqual(2);
+        expect(tabContents[0].childNodes[0].nodeType).toEqual(Node.TEXT_NODE);
```
youdz: This is a strange way to test it, I'm not sure what's happening. Why not just test `tabContent. templateRef`?

src/clarity-angular/layout/tabs/tab-content.spec.ts
```diff
@@ -5,17 +5,18 @@
...
-import {TabContent} from "./tab-content";
 import { ClrTabsModule } from "./tabs.module";
+import {IfActiveService} from "../../utils/conditional/if-active.service";
+import {IF_ACTIVE_ID_PROVIDER, Tab} from "./tab";
 
 @Component({
     template: `
-        <clr-tab-content [clrTabContentActive]="true">Content1</clr-tab-content>
+        <clr-tab-content>Content1</clr-tab-content>
         <clr-tab-content>Content2</clr-tab-content>
```
youdz: Why do you need to test 2 of them? Does it change anything?

src/clarity-angular/utils/conditional/if-open.directive.ts
```diff
@@ -30,7 +32,9 @@ import { IfOpenService } from "./if-open.service";
  * using it in the component template.
  *
  */
-export class IfOpenDirective implements OnInit {
+export class IfOpenDirective implements OnDestroy {
+
+    private ngUnsubscribe: Subject<void> = new Subject<void>();
```
jeeyun: Hmm I have no idea what happened here. I thought I had it correctly (as you describe below). Anyways, thanks for catching it!

src/app/tabs/tabs-angular.ts
```diff
@@ -12,4 +13,13 @@ import { Component } from "@angular/core";
     templateUrl: "./tabs-angular.demo.html"
 })
 export class TabsAngularDemo {
+    @ViewChild(Tabs) myTab: Tabs;
+
+    public test: boolean = true;
+
+    ngAfterViewInit() {
+        this.myTab.ifActiveService.currentChange.subscribe((active: any) => {
+            console.log(this.myTab.activeTab);
```
jeeyun: This is a log in the demo, but I'll remove it.

src/clarity-angular/utils/conditional/if-active.directive.ts
```diff
@@ -0,0 +1,108 @@
...
+    Output,
+    EventEmitter,
+    Inject,
+    OnDestroy
+} from "@angular/core";
+
+import { IfActiveService } from "./if-active.service";
+import {IF_ACTIVE_ID} from "../../layout/tabs/tab";
+import {Subject} from "rxjs/Subject";
+import "rxjs/add/operator/takeUntil";
```
jeeyun: I guess I responded too soon above :) Ok, I'm fine with reverting.

src/clarity-angular/utils/conditional/if-active.directive.ts
```diff
@@ -0,0 +1,108 @@
...
+ * A structural directive that controls whether or not the associated TemplateRef is instantiated or not.
+ * It makes use of a Component instance level service: IfActiveService to maintain state between itself and
+ * the component using it in the component template.
+ *
+ */
+export class IfActiveDirective implements OnDestroy {
+
+    private ngUnsubscribe: Subject<void> = new Subject<void>();
+
+    constructor( private ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) private self: any,
```
jeeyun: in `tab-content.ts` and `tab-link.ts`, `id` was already taken, so I was trying to use an alternative name (hence the `self`) and that got carried over to `if-active-directive.ts`. I'll rename the id's in `tab-content.ts` and `tab-link.ts` to `tabContentId` and `tabLinkId` and use `id` in every place for the `IF_ACTIVE_ID`.

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -4,158 +4,45 @@
...
 @Component({
     selector: "clr-tabs",
-    templateUrl: "./tabs.html"
+    template: `
+        <ul class="nav" role="tablist">
+            <ng-content select="clr-tab"></ng-content>
+        </ul>
+        <ng-container *ngFor="let tab of tabChildren">
+            <ng-template *ngIf="tab.tabContent" [ngTemplateOutlet]="tab.tabContent.templateRef"></ng-template>
+        </ng-container>
```
jeeyun: Yes. Just as a note we do the same thing for button groups. I also couldn't find a way to achieve this visually (with css) if I were to not separate out the tabContents from being inside each tab. I'm not sure what'd be a clean way to achieve this. I'll add a comment to the file so we are reminded to come back to it.

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -4,158 +4,45 @@
...
-        if (index < this.tabContents.length) {
-            selectedTabContent = this.tabContents[index];
-            this.currentTabContent = selectedTabContent;
-            selectedTabContent.active = true;
+    get activeTab() {
+        let activeTab: Tab;
+        if (this.tabChildren) {
+            activeTab = this.tabChildren.find((tab: Tab) => {
+                return tab.active;
+            });
```
jeeyun: Do you think this is a common functionality that the end user might want though? When they subscribe to currentChange, wouldn't they have to write the above logic to programmatically to map it back to which one is the active Tab? Thought it might be good to provide as a helper function. Or is this not a common scenario?

src/clarity-angular/layout/tabs/tabs.spec.ts
```diff
@@ -67,88 +68,24 @@ describe("Tabs", () => {
...
-    it("emits a clrTabSelected event when a tab is clicked", (done: any) => {
-        instance.currentTabLinkChanged.subscribe((selectedTab: TabLink) => {
-            expect(selectedTab).toEqual(instance.tabLinks[0]);
-            done();
-        });
-
-        instance.selectTab(instance.tabLinks[0]);
+    it("sets the designated tab as active", () => {
+        let tabs = instance.tabChildren.toArray();
+        expect(instance.activeTab).toEqual(tabs[0]);
```
jeeyun: I had some logic to designate the first tab as active, but I removed it to resolve the EHCAIWC error with dynamic tabs. Removing corresponding tests. Comment or ping me if you think we should preserve this functionality.

src/clarity-angular/layout/tabs/tab.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor(public ifActiveService: IfActiveService, @Inject(forwardRef(() => IF_ACTIVE_ID)) public self: any) {
+    }
+
+    ngAfterContentChecked() {
+        // the tabContent is only there if this tab is active
+        if (this.tabLink && this.tabContent) {
+            this.tabLink.ariaControls = this.tabContent.id;
+            this.tabContent.ariaLabelledBy = this.tabLink.id;
+        }
```
jeeyun: Getting rid of the parent's `ngAfterContentInit` got rid of the chocolate error. Let me know if you think that's insufficient and I can explore this more.

src/clarity-angular/layout/tabs/tab-content.spec.ts
jeeyun: Nope. Removing the second one in the test.

src/clarity-angular/layout/tabs/tab-content.spec.ts
```diff
@@ -5,17 +5,18 @@
...
    `
 })
 class TestComponent {
-    @ViewChildren(TabContent) tabContentChildren: QueryList<TabContent>;
+    @ViewChildren(Tab) tabChildren: QueryList<Tab>;
```
jeeyun: Removed.

src/clarity-angular/layout/tabs/tab-link.ts
```diff
@@ -5,34 +5,48 @@
...
-import { Tabs } from "./tabs";
+let nbTabLinkComponents: number = 0;
 
 @Component({
     selector: "clr-tab-link",
-    templateUrl: "./tab-link.html",
+    template: `
+        <button class="btn btn-link nav-link" type="button">
+            <ng-content></ng-content>
+        </button>
```
jeeyun: This is definitely doable. I just tested it real quick and the only problem is our css doesn't work as is since we assume a level of hierarchy that disappears with this approach. I can't guarantee how clean the changed css will be, especially since I assume we're gonna try to preserve backward compatibility for all navs using the same css. Do you still want me to try?

src/clarity-angular/layout/tabs/tab-content.spec.ts
jeeyun: Updated.

src/clarity-angular/utils/conditional/if-active.directive.ts
```diff
@@ -0,0 +1,108 @@
...
+                 private template: TemplateRef<any>, private container: ViewContainerRef ) {
+
+        this.ifActiveService.currentChange
+            .takeUntil(this.ngUnsubscribe)
+            .subscribe((id) => {
+                this.updateView(id === this.self);
+                this.activeChange.emit(id === this.self);
```
jeeyun: It doesn't seem possible for the directive to know what the previous value was (in order to determine if the boolean actually changed). Do you suggest I change the observable to send both old and new id's or should this state be stored by the directive? Maybe I'm missing something obvious here.

youdz: You can keep a private boolean property on the directive that's the last value you emitted. Just emit when `(id === this.self) != this.wasActive`.

src/clarity-angular/layout/tabs/tabs.ts
youdz: Well, we don't have currentChange anymore. I'm ok leaving this here for now if you think it really is useful, but people don't want to work with _our_ Tab objects, that's uncomfortable most of the time. So I'm curious to see a use case where the per-tab two-way binding isn't just simpler.

src/clarity-angular/layout/tabs/tabs.spec.ts
youdz: I think this functionality is necessary. Having the first tab displayed by default is Tabs 101, that's something our users don't want to have to worry about. We need to put it back and solve the EHCAIWC error.

src/clarity-angular/utils/conditional/if-active.directive.ts
```diff
@@ -34,15 +34,22 @@ import {Subscription} from "rxjs/Subscription";
...
 
         this.subscription = this.ifActiveService.currentChange.subscribe((newId) => {
-                this.updateView(newId === this.id);
-                this.activeChange.emit(newId === this.id);
+                let isNowActive = newId === this.id;
+                this.updateView(isNowActive);
+
+                // only emit if the new active state is changed since last time.
+                if (isNowActive !== this.wasActive) {
+                    this.activeChange.emit(isNowActive);
```
youdz: Shouldn't the `this.updateView(isNowActive);` be inside the `if` ? I'm afraid we might have an edge case in the future where we duplicate a tab content if this observable fires with the same id twice. I don't think it can happen right now, but just putting everything inside the `if` doesn't cost anything and is safer.

src/clarity-angular/layout/tabs/tab.ts
```diff
@@ -20,10 +21,19 @@ import {
...
+    constructor(public ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) public id: number,
+                private tabsService: TabsService) {
+
+        tabsService.register(this);
+
+        // if there's no active tab, set itself as active; it will be overridden if
+        // a tab created after this one sets it explicitly
+        if (!this.ifActiveService.current) {
+            this.ifActiveService.current = id;
+        }
```
youdz: Do you mind adding a comment to this saying it might be worth moving to the `IfActiveService` itself? If all of our "if active" components want the first item to be active by default, we'll end up with this small piece of code duplicated everywhere.

You don't need to move it right now, just adding a comment is good enough so that people won't just copy-paste blindly but instead refactor slightly if they are doing the same thing.

src/clarity-angular/utils/conditional/if-active.directive.ts
jeeyun: That makes sense. Moving it inside the if.

src/clarity-angular/layout/tabs/tabs.spec.ts
```diff
@@ -54,7 +55,7 @@ describe("Tabs", () => {
         TestBed.configureTestingModule({
             imports: [ClrTabsModule],
             declarations: [TestComponent],
-            providers: [TabLink, TabContent]
+            providers: [Tabs, IfActiveService]
```
adityarb88: Why do we need to register `Tabs` as a provider here?

src/clarity-angular/utils/conditional/if-open.directive.ts
```diff
@@ -68,7 +70,13 @@ export class IfOpenDirective implements OnInit {
     }
 
     constructor( private ifOpenService: IfOpenService, private template: TemplateRef<any>,
```
hippee-lee: Is the subscription to services in the constructor the preferred way to listen to observables?
It seems like we are moving the subscriptions here instead of OnInit. (I think I saw similar changes in the Chocolate PR too)
 /cc @youdz

src/clarity-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -0,0 +1,42 @@
...
+
+let nbTabLinkComponents: number = 0;
+
+@Directive({
+    selector: "[clrTabLink]",
+    host: {
+        "[id]": "tabLinkId",
+        "[attr.aria-selected]": "active",
+        "[attr.aria-controls]" : "ariaControls",
+        "role": "presentation",
```
jeeyun: yeah so the ones with the `[]` treat the right side as an input that needs to be evaluated. so the id for example, is set to the value of `tabLinkId` variable. The role is just for a11y and the hard-coded value of `presentation` is assigned.

src/clarity-angular/utils/conditional/if-open.directive.ts
jeeyun: For the conditional directives, subscribing to them inside `OnInit` is too late, as the value whose change we are subscribing to can be set earlier as input in the component and the directive needs to react to that value. It just depends on the scenario. Copy/pasting from our thread outside of GitHub:

OnInit is an excellent place for:
- setup that needs the inputs to be ready
- heavy initial action that you don't want to trigger when you just instantiate the component manually in tests (an HTTP request, for instance)

src/clarity-angular/layout/tabs/tabs.spec.ts
jeeyun: Oops, that shouldn't be there. I replaced it with `TabsService` which is the correct provider it needs.

PR #1156, opened 2017-07-06
[core] Upgrade Clarity to use @angular/* @ ^4.2.0

jeeyun:
Earlier we made our angular versions stricter since angular had a bug that prevented rollup: https://github.com/angular/angular/issues/17389

Now that the bug is fixed, we should develop against latest angular to ensure our packages can support users who are on the latest.


jeeyun:
I couldn't figure out a fix for the two gemini test failures. I can replicate the scenario/discrepancy manually so it's not a gemini issue but rather some changes occurring as a result of upgrade to angular@4.2.x. If you guys are comfortable with going forward with this + filing a separate bug for those two, lmk. Otherwise, this PR is on hold.

mathisscott:
@youdz
Could you let @jeeyun know if we can more forward with this PR and get it merged in for 0.9.14?

youdz:
I'm ok merging this in if it doesn't block every single future PR because of the 2 failing Gemini tests.

jeeyun:
- Upgraded to ^4.3.0
- Updated reference images for datagrid css changes (they just seem to shift the columns slightly with the upgrade but doesn't break anything)
- Commented out one failing Gemini test with Stackview inside modal. Filed a "future" bug for it that should become valid after the upgrade (https://github.com/vmware/clarity/issues/1235)

PR #1185, opened 2017-07-18
Check if the subscription exists before calling the unsubscribe method.

o1ena:
This brakes the VMC operator console.


o1ena:
Signed-off-by: Olena Iarova <olena.iarova@gmail.com>

Issue #1188, opened 2017-07-19
Placement of button groups for data grids


tishun:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
- the action buttons in a button group in the header of a data grid are placed close to the grid
- the entire surface of the the action buttons in a button group in the header of a data grid is clickable

### Actual behavior
- the action buttons in a button group in the header of a data grid are placed with a large amount of free space from the datagrid itself; this is probably for the purpose of being able to host larger buttons, but especially for btn-link types of buttons it looks as the action buttons are not associated with the data grid itself

- around 30% of the surface (located in the lower part of the button) is not clickable, which is especially visible when having btn-link buttons - basically the user could place the mouse over the lower part of the button text and not be able to click it

### Reproduction of behavior
Visible in the Clarity showcase https://vmware.github.io/clarity/documentation/datagrid/batch-action
Screenshot : http://i67.tinypic.com/xnhz6d.jpg

### Environment details

* **Angular version:** any

* **Clarity version:** 0.9.5

* **OS and version:** any

* **Browser:** all

tishun:
Awesome job, guys. What milestone release do we expect to contain this fix?

adityarb88:
@tishun: This will go in both 0.9.15 & 0.10.0

adityarb88:
@tishun: Actually we are just going to release this for 0.10.0 since the fix for this is a breaking change visually. I will add more details in the release notes. thanks!

PR #1197, opened 2017-07-24
[UI] Datagrid action bar click target fix

adityarb88:
Fixes: #1188 

Fixes an issue with the spacing and click target of the datagrid action bar.

Tested on: Firefox, Chrome, Safari, IE

Screenshots: 
image


adityarb88:
@youdz: yep. I don't believe we should keep a fixed height reserved for a button group just in case there is a use case that requires more or less space there. do you think this is a breaking change btw?

cc: @reddolan

reddolan:
I agree. I worry that the additional weight to manage the fixed height will bite us and our users in the butt. I recommend using disabled buttons vs hiding them and then showing them. So if a team uses this logic, there won't be any content moving up and down. If for some reason teams want to show/hide buttons, they can apply a fixed height to the datagrid scenario themselves.

adityarb88:
merging in master based on @youdz and @reddolan's feedback

PR #1200, opened 2017-07-24
[NG] Emit clrDropdownOpenChange state on changes:

### LINKED ISSUE #1195: Dropdown does not emit clrDropdownMenuOpenChange when click is made outside the host element
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When dropdown is open, and click is made outside the host element, the dropdown is closed. The clrDropdownMenuOpenChange emitter should emit this change of state accordingly.

### Actual behavior
Dropdown is closed, but clrDropdownMenuOpenChange is not emitted. 

### Reproduction of behavior
Create any dropdown and register a console.log(event) to be executed when clrDropdownMenuOpenChange is emitted. When dropdown is opened, the event is emitted (`true` is printed to console). When dropdown is closed by clicking outside of the host element, nothing is printed.
 
I believe this line: https://github.com/vmware/clarity/blob/v0.9.14/src/clarity-angular/popover/dropdown/dropdown.ts#L142 should read:
`this.open = true;` 
instead of 
`this._open = true;`
to trigger the setter and properly invoke the emitter.

### Environment details

* **Angular version:** 4.1.1

* **Clarity version:** 0.9.5

* **OS and version:** macOS

* **Browser:** Chrome

hippee-lee:
- Uses a subscription to IfOpenService.openChange
- Unsubscribes in ngOnDestroy
- closes #1195

## Notes

- We talked about implementing this unsubscribe pattern before and since the `IfOpenService.openChange` observable is of the type infinate value I used the takeUntil operator / pattern for unsubscribe. I can unsubscribe the old way if we do not want to do this pattern here. 
- There are a few formatting changes in `dropdown.spec.ts` as I try to get the files to the formatting standard we discussed (no spaces in the import braces).

### DIFF
--- a/src/clarity-angular/popover/dropdown/dropdown.ts
+++ b/src/clarity-angular/popover/dropdown/dropdown.ts
@@ -9,6 +9,7 @@ import {
     EventEmitter,
     HostListener,
     Input,
+    OnDestroy,
     Optional,
     Output,
     SkipSelf
@@ -17,6 +18,7 @@ import {
 import { Point, PopoverOptions } from "../common/popover";
 import { menuPositions } from "./menu-positions";
 import {IfOpenService} from "../../utils/conditional/if-open.service";
+import {Subscription} from "rxjs/Subscription";
 
 // TODO: the ng-content inside the ng-template should ideally just be
 // <ng-content select="clr-dropdown-menu"></ng-content>. Remove .dropdown-menu in 1.0?
@@ -39,7 +41,9 @@ import {IfOpenService} from "../../utils/conditional/if-open.service";
     },
     providers: [ IfOpenService ]
 })
-export class Dropdown {
+export class Dropdown implements OnDestroy {
+
+    private ifOpenServiceUnsubscribe: Subscription;
 
     @Input("clrDropdownMenuOpen") _open: boolean = false;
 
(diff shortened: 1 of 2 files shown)


src/clarity-angular/popover/dropdown/dropdown.ts
```diff
@@ -56,6 +61,11 @@ export class Dropdown {
     constructor(public elementRef: ElementRef, @SkipSelf() @Optional() public parent: Dropdown,
                 public ifOpenService: IfOpenService) {
         this.anchor = elementRef.nativeElement;
+
+        // Subscribe and broadcast openChanges from the service
+        ifOpenService.openChange.takeUntil(this.ngUnsubscribe).subscribe(change => {
```
adityarb88: I remember @youdz and @jeeyun deciding not to use the takeUntil pattern. @jeeyun please correct me if I am wrong.

hippee-lee: I wasn't 100% sure what we decided. This is easy to change to a Subscription unsubscribe as we do other places if it was decided not to go this route. 

One thing I like about it though is that the `takeUntil` just operates on the subscription provided by the service and we don't have to assign it to a new one that later gets cleaned up 
idk if its better but it seems cleaner to me. 

```
let mySubscription = ifOpenService.openChange.subscribe(...)
```

jeeyun: Yes, we decided to not go forward with the pattern because it'd require the additional import (dependency) and the gain seemed not significant. Not sure if this will change in the future though but for now, would it be ok if you reverted the pattern?

hippee-lee: Sure, no problem. I'll push an update to the pr in a bit.

hippee-lee: One thing though wrt to: 
```
 it'd require the additional import (dependency)
```
We gain the Subject import but lose the Subscription import. You are probably right though, it should be unsubscribed the same way as the others.

Issue #1161, opened 2017-07-07
Create a "media" category for Clarity Icons


mathisscott:
Currently, we have a number of "media/entertainment" icons packed into the Essentials, Social, and Technology categories. This bloats essentials with icons that non-music or image-based webapps may not use.

We'll need @lil-kim to answer three questions:

1. Do we need a separate category for media/entertainment icons? Why not if the answer is "no"?
2. What should we call the new category?
3. Which icons do we need to relocate there? (Please update the confluence doc to reflect this migration. Whoever implements this change should use the confluence doc as a guide.)

Note also that this is complicated by similarly themed icons spread across different categories. Like play, stop, and fast-forward being in Essentials but volume-up/down being in technology. Or camera being in social but video-camera being in technology -- while film-strip is in Essentials.

Then we'll need @Shijir or whomever is assigned to implement this to perform the migration from the old categories to the new category.

This constitutes a breaking change so it should go into a minor release. If possible, I would prefer it go into 0.10. If not, it will have to go into 0.11 and should be completed before Clarity Icons is finalized.

lil-kim:
1, 2. Please create a new category, it should be called: Media
3. Refer to the Confluence icon list. You'll find the icons to be moved in the "Media" column.
https://confluence.eng.vmware.com/display/CLAR/Icon+Inventory+and+Classification
Thank you!

Issue #1160, opened 2017-07-07
Create "travel" category for Clarity Icons


mathisscott:
Currently, we have a number of "travel/transportation" icons packed into the Social category. This bloats social with icons that are not applicable to that category.

We'll need @lil-kim to answer two questions:

1. Do we need a separate category for travel/transportation icons? Why not if the answer is "no"?
2. What should we call the category?
3. Which icons do we need to relocate there? (Please update the confluence doc to reflect this migration. Whoever implements this change should use the confluence doc as a guide.)

Then we'll need @Shijir or whomever is assigned to implement this to perform the migration from Social to the new category.

This constitutes a breaking change so it should go into a minor release. If possible, I would prefer it go into 0.10. If not, it will have to go into 0.11 and should be completed before Clarity Icons is finalized.

lil-kim:
1, 2. Please create a new category, it should be called: Travel
3. Refer to the Confluence icon list. You'll find the icons to be moved in the "Travel" column.
https://confluence.eng.vmware.com/display/CLAR/Icon+Inventory+and+Classification
Thank you!

PR #1207, opened 2017-07-25
[ICONS] created new categories for icons

### LINKED ISSUE #1159: Create "commerce" category for Clarity Icons
Currently, we have a number of "commerce/finance" icons packed into the Essentials category. This bloats essentials with icons that many non-ecommerce webapps may not use. Note the non-trademarked icons in #501 would go in this category as well.

We'll need @lil-kim to answer two questions:

1. Do we need a separate category for commerce/finance/marketplace icons? Why not if the answer is "no"?
2. What should we call the new category?
3. Which icons do we need to relocate there? (Please update the confluence doc to reflect this migration. Whoever implements this change should use the confluence doc as a guide.)

Then we'll need @Shijir or whomever is assigned to implement this to perform the migration from Essentials to the new category.

This constitutes a breaking change so it should go into a minor release. If possible, I would prefer it go into 0.10. If not, it will have to go into 0.11 and should be completed before Clarity Icons is finalized.

Shijir:
Closes #1159, #1160, #1161

- Are there any icons you would suggest to move to another set?
- Do we still want to keep duplicates of icons across different sets? As we are moving icons into smaller and more specialized sets, I think we shouldn't have the same icons in different sets. 

clarityiconsnewcategories


mathisscott:
At a glance, the `repeat` icon should probably be in Essentials, not Media. That type of icon gets used for all sorts of things in web apps.

CC @lil-kim

mathisscott:
I think duplicating among sets is something we should avoid.

Shijir:
Fyi, here is the published svg files. [all-shapes.zip](https://github.com/vmware/clarity/files/1174521/all-shapes.zip)

Shijir:
clarity

PR #1211, opened 2017-07-26
[core] Adding clang-format (#1210)

### LINKED ISSUE #1210: [core] add formatter and clean up linter in our build process
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior

Right now we have a strict linter but no formatter, and often time this is counter-productive to our development. We want to be smart about when they run and when they stop you from going forward vs just outputting a warning.

jeeyun:
Starting point for adding clang-format to our code base. My initial changes/thoughts:

- Demoted tslint to only output warning but never stop the build (but we'll always run tslint both for initial startup and on watch)
- Removed tslint rules that will be covered by clang-formatter
- Added a gulp task `format` which will format and replace source code.
- Bonus: Added `reloadDelay` and `reloadDebounce` to live reloading so we don't get a million reloads with small changes. 

Here, I'm relying on the developers/contributors to:
- pay attention to tslint logs and fix them on their own
- run the `format` task before checking in your code

I'm open to enforcing running of `format` in some points in the process. Just not sure yet where the sweet spot would be. Feedback welcome.

Resolves #1210.


build/tasks/clang-format.js
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+var gulp = require('gulp');
```
adityarb88: @jeeyun: Quick question. Do we not need a clang format config file for setting things that we want to check?

jeeyun: By default it'll use the "Google-style" format. Should I add a config file so we override it?

adityarb88: Oh i didn't know that. Do you think we should? Is there anything that we are doing differently formatting wise? I think may be we should just in case we want modify options later on.

jeeyun: good point. I will add one with just a couple of options there.

build/tasks/typescript.js
```diff
@@ -124,31 +124,31 @@ gulp.task('typescript:app', function () {
 var clarityHtmlFiles = ["src/clarity-angular/**/*.html"];
 gulp.task('typescript:clarity:watch', function () {
     gulp.watch(claritySources.concat(clarityHtmlFiles), function () {
-        return runSequence('tslint:clarity:no-error', 'typescript:clarity');
+        return runSequence('tslint:clarity', 'typescript:clarity');
```
youdz: No, linting should not be part of the build anymore, at all. It should be a separate task, like formatting, that only runs when required or as part as an overall check on Travis.

build/tasks/clang-format.js
```diff
@@ -0,0 +1,35 @@
...
+        .pipe(gulp.dest('src/app/'));
+});
+
+gulp.task('format:icons', function() {
+    return gulp.src(iconsSources)
+        .pipe(format.format('file'))
+        .pipe(gulp.dest('src/clarity-icons/'));
+});
+
+gulp.task("format", ["format:clarity", "format:app", "format:icons"], function(){});
```
youdz: You are missing a `format:check` task that we can run on Travis.

jeeyun:
@youdz I have tried formatting, then linting, then running both again to verify that they wouldn't conflict with each other or cause additional failures. FYI I have separated out the commits so that the first commit only shows what files were changed to make the new build process happen, and the subsequent one contains still the unfortunate manual work post clang formatter that one would have to do to make tslinter happy (this is what'd be left to fix after clang formatter +  tslinters autofix). As discussed we really need to coordinate when this will be merged. It's going to be "fun."

PR #1226, opened 2017-07-28
[NG] Tree View Animation Fix - 0.10.0 - DO NOT MERGE

adityarb88:
This change is not really needed because in the collapse mode in the 0.10.0 tree, the nodes are removed from the DOM but just adding it for safety.

Tested on: Safari, Firefox, Chrome


adityarb88:
@youdz: I should have added a comment for that. I am doing this because, Safari for some reason applies the overflow styles and then removes it. Adding it to CSS always makes sure that overflow: hidden is applied to the node. Do you think I should investigate this more rather than doing this?

PR #1236, opened 2017-07-31
[ng] Tabs overflow (#602)

### LINKED ISSUE #602: Support more tabs than can fit on the page
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

If I have more tabs to display than can fit on the page or in their container, they should either scroll, or wrap, or offer some way to access the last ones.

### Actual behavior

At the moment, they simply overflow the `<clr-tabs>` element. If it scrolls, the whole content of the tabs scroll, the delimiter doesn't go across all the links, ... If it doesn't scroll, the last tabs are simply not accessible.

### Reproduction of behavior

https://plnkr.co/edit/IMzGJy?p=preview

### Environment details

* **Angular version:** 2.4.X

* **Clarity version:** 0.8.X

* **Browser:** all

jeeyun:
tab-overflow

The goal of this PR is to provide a solution for tabs that do not fit into the horizontal space that's available (#602). Eventual goal is to be fully responsive, but for now we'll provide a manual solution for immediate work around for those that are needing a solution.

Some notes:
- This PR assumes that popovers refactor will merged in before this one, though I didn't actually code it like the refactored code was available. Therefore, I'll still need to rebase and remove some duplicate code once that happens (one popover test case fails now in the PR).
- I named the class that we are using to render tab links on demand `TemplateRefContainer`. Suggestions welcome as far as naming goes, as well as its final destination (right now, it's inside the tabs folder.

Fixes #602.


reddolan:
The interactions and design look spot-on!

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -3,35 +3,84 @@
...
+                        <ng-container #overflowTabLinks></ng-container>
+                    </clr-tab-overflow-content>
+                </div>
+            </ng-container>
         </ul>
+        <!--this is the tab content, rendered outside the ul element above-->
+        <!--TODO: find a better way to achieve this-->
         <ng-container *ngFor="let tab of tabsService.children">
             <ng-template *ngIf="tab.tabContent" [ngTemplateOutlet]="tab.tabContent.templateRef"></ng-template>
         </ng-container>
```
youdz: We can get rid of that now!

Since you are now "extracting" the links rather than the content, you can project the `<clr-tab>` components themselves directly with `<ng-content>`. You don't even need a selector on it.

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -3,35 +3,84 @@
...
 
-    get activeTab() {
-        let activeTab: Tab;
-        if (this.tabsService.children.length > 0) {
-            activeTab = this.tabsService.children.find((tab: Tab) => {
-                return tab.active;
-            });
+    ngAfterViewInit() {
+        this.renderTabLinks();
+    }
```
youdz: Doing the same thing after view init and after view checked is redundant. `ngAfterViewChecked` is called after `ngAfterViewInit` initialization.

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -3,35 +3,84 @@
...
+
+    ngAfterViewChecked() {
+        if (this.overflowTabsChanged()) {
+            this.renderTabLinks();
+            this._prevOverflowTabs = this.tabsService.overflowTabs;
```
youdz: I believe you now have a problem with dynamic inline tabs. They do not update properly.

Rather than address this manually, I would like you to use a behavior closer to the button groups one: only trigger a redraw when a tab link's `inOverflow` input actually changes.
The reason behind this change request is that I'm a bit scared of the potential performance impact in the future if we let a longer operation slip in the `if` check.

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -3,35 +3,84 @@
...
+        return (this.tabsService.overflowTabs.length !== this._prevOverflowTabs.length);
+    }
+
+    renderTabLinks() {
+        // Set the ContainerRefs and render the view. One of the refs might have been undefined
+        // earlier, so we always set it before calling renderView here rather than
+        // setting them inside ngAfterViewInit() function.
+        this.tabsService.inlineTabLinksContainerRef = this.inlineTabLinksContainerRef;
+        this.tabsService.overflowTabLinksContainerRef = this.overflowTabLinksContainerRef;
+        this.tabsService.renderView();
```
youdz: So now that I have read everything, I think you can get away with a much, much simpler solution. Just use:
```html
<ng-container *ngFor="let link of tabLinks">
    <ng-container *ngIf="!link.inOverflow"
        [ngTemplateOutlet]="link.templateRefContainer.template"></ng-container>
</ng-container>
```
and same thing for the inOverflow. No more `@ViewChild`, no more manual creation, ...
And you don't need to keep track of two lists, you're using the same in both and just trusting `*ngIf`.

src/clarity-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -39,4 +49,17 @@ export class TabLinkDirective {
         return this.ifActiveService.current === this.id;
     }
 
+    ngOnInit() {
```
jeeyun: done.

src/clarity-angular/layout/tabs/tabs-service.ts
```diff
@@ -19,10 +23,46 @@ export class TabsService {
...
         }
+
+    }
+
+    renderView() {
+        this._links.forEach((link: TabLinkDirective) => {
+           if (link.inOverflow) {
+               this.overflowTabLinksContainerRef.createEmbeddedView(link.templateRefContainer.template);
+           } else {
+               this.inlineTabLinksContainerRef.createEmbeddedView(link.templateRefContainer.template);
```
jeeyun: Hmm actually I think it's much cleaner and simpler to move this logic into `Tabs`. Take a look and lmk what you think!

src/clarity-angular/layout/tabs/tabs.ts
jeeyun: I have to always call it on `ngAfterViewInit` for the initial rendering, but only want to call it inside `ngAfterViewChecked` if the overflow tabs changes later on. In the initial render, the `if` condition inside `ngAfterViewChecked` isn't actually true, so the second call to render the view doesn't happen. Let me know if my explanation is confusing.

jeeyun: Ignore the earlier comment; we're now relying on `ng-container`s with `ngIf`s without having to manage this manually.

src/clarity-angular/layout/tabs/tabs.ts
jeeyun: I tried to project it directly and now I'm getting the chocolate error while trying to set the aria attributes in the `ngAfterContentChecked` of `clr-tab`. Is there a better place to manage/set the aria attributes? Will continue to investigate.

src/clarity-angular/layout/tabs/template-ref-container.ts
```diff
@@ -0,0 +1,12 @@
+import {Component, TemplateRef, ViewChild} from "@angular/core";
+
+@Component({
+    template: `
+      <ng-template>
+        <ng-content></ng-content>
+      </ng-template>
+    `,
+})
+export class TemplateRefContainer {
```
jeeyun: Yup. Moved it to `utils/` with its own module.

src/clarity-angular/layout/tabs/tabs.module.ts
```diff
@@ -8,15 +8,25 @@ import { NgModule } from "@angular/core";
...
+import {ClrOutsideClickModule} from "../../utils/outside-click/outside-click.module";
 
 @NgModule({
     imports: [
         CommonModule,
-        ClrConditionalModule
+        ClrCommonPopoverModule,
+        ClrConditionalModule,
+        ClrIconModule,
+        ClrOutsideClickModule
```
jeeyun: Removed.

src/clarity-angular/layout/tabs/tabs.module.ts
```diff
@@ -8,15 +8,25 @@ import { NgModule } from "@angular/core";
...
     ],
     declarations: [
         TABS_DIRECTIVES
     ],
+    entryComponents: [
+        TemplateRefContainer
+    ],
```
jeeyun: Done as part of moving `template-ref-container.ts` to `utils`.

src/clarity-angular/layout/tabs/tabs-service.ts
jeeyun: Now this logic doesn't exist anymore. Using the simpler markup relying on `ng-container`s with `*ngIf`s.

build/tasks/live.js
```diff
@@ -24,7 +24,9 @@ gulp.task("live", function() {
 				historyApiFallback({ index: "./app/index.html" })
 			]
 		},
-		files: ["dist/**/*"]
+		files: ["dist/**/*"],
+        reloadDebounce: 1000,
+        reloadDelay: 1000
```
hippee-lee: Thanks for this!

src/clarity-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -16,18 +20,49 @@ let nbTabLinkComponents: number = 0;
...
 export class TabLinkDirective {
-    @Input("clrTabLinkId") tabLinkId: string;
-    ariaControls: string;
 
-    constructor(public ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) private id: number) {
+    @Input("clrTabLinkInOverflow") inOverflow: boolean;
+    templateRefContainer: TemplateRefContainer;
+
+    constructor(public ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) private id: number,
+                private  ariaService: AriaService, private el: ElementRef,
```
hippee-lee: Is this private because its returned from `(public) ariaControls` but not used directly in the view? I noticed that it's public in `tab-content.ts`

src/clarity-angular/layout/tabs/aria-service.ts
```diff
@@ -0,0 +1,7 @@
+import {Injectable} from "@angular/core";
+
+@Injectable()
+export class AriaService {
```
hippee-lee: Do we have plans to / should we have plans to move this to utils so it can be re-used with other components?

src/clarity-angular/layout/tabs/tab-link.directive.ts
jeeyun: Oh good catch. I'll make the other one private as well. Yes, it's exactly as you said. Earlier I had them as public and used them directly in the view, but later decided to use the getter that uses the service instead.

src/clarity-angular/layout/tabs/aria-service.ts
jeeyun: Hmm that's a good point. I'm not sure of other components that will use them yet. `aria-controls` and `aria-labelledby` are pretty limited in use cases since they depend on each other and only certain components like tabs would have them. I'm open to moving them though, if we find another component that could use an `aria-` related service. I'll add a TODO with a note to consider moving this if we run into another case where this service is useful.

youdz: I considered this too, but different components need different attributes. I looked at it quickly, and it seemed to me like there was very little reusability. 😞

PR #1261, opened 2017-08-04
[CORE] Store/Use gemini ref images using S3

jeeyun:
Sorry this didn't come earlier. With this change we won't be keeping copies of ref images in our repo. Instead, Travis will sync with the "golden image set" from S3, whose location is defined by our Travis environment variable. The jobs that are taking ref images are going to continue to upload to S3 as before. If we want to update the reference images in the future, we simply need to update that one Travis environment variable.


youdz:
I don't really understand what's happening here, so I'll refrain from approving blindly. Feel free to merge if you have enough :shipit:.

I'd be interested in a quick explanation on a call next week, if you have the time.

mathisscott:
+1 to what Eudes said

jeeyun:
Sure, I can schedule a quick call early next week to go over this before merging.

hippee-lee:
Also, with the docker changes I am working on this change also means core devs will need to have the aws-cli tools installed.

ref-img.sh
```diff
@@ -0,0 +1,9 @@
+if [ "$TEST_SUITE" = "css:test --set set1" ]; then
+    wget $SET1;
```
hippee-lee: Where does $SET1 get set? I thought it would be in the Travis config but didn't see it.

build/tasks/gemini.js
```diff
@@ -20,7 +20,7 @@ var config = {
     }
 };
 
-gulp.task('css:reference', ["build"], function() {
+gulp.task('css:reference', ["build"], function(cb) {
```
hippee-lee: I'm assuming this is for callback, right? 
Isn't cb supposed to be called somewhere at the end  of this task? `cb()`?

adityarb88:
@jeeyun @hippee-lee: What are we doing with this PR? Can we close this?

hippee-lee:
We should close it for now.

Issue #875, opened 2017-05-09
Navigation doesn't scroll reliably on iOS


reddolan:
```
[X] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
when opening the collapsed navigation on a phone, the long list of navigation items should scroll up and down

### Actual behavior
- from the Documentation pages
- click vertical ellipses to show offcanvas navigation
- try to scroll the navigation to the lower menu options
- the main content scrolls, the nav does not
- not sure if this is just isolated to the website or is a component-level issue
- I can get the nav to scroll but only after collapsing and expanding one of the nav headers
- it seems like it is a focus issue

### Reproduction of behavior
![ios-navigation-issue](https://user-images.githubusercontent.com/1858226/29224436-3184a0cc-7e7f-11e7-9a2d-1b9b8253222b.gif)

### Environment details
- Clarity version: 0.9.3
- OS and version: iOS 10.3.1
- Browser: iOS Safari, iOS Chrome

bojan88:
This bug is affecting all touch devices. I noticed it on Chrome for Android.
The problem is with CSS - `overflow-y: auto` is set on hover on non-touch devices, but on touch devices it's set on click event.
I filed a pull request for this.

I created a gif if it's not clear from the comments.
![clarity-menu-bug](https://user-images.githubusercontent.com/1783133/29211302-8d2cafdc-7e98-11e7-818d-cf278cf25355.gif)

adityarb88:
Closing because the fix was merged in 0.10.1 in this PR #1286

PR #1286, opened 2017-08-11
[UX] Collapsible navigation menu scroll bug fix

bojan88:
Fixes issue #875

Signed-off-by: Bojan Djurdjevic <bojan88@gmail.com>

### DIFF
--- a/src/clarity-angular/layout/nav/_sidenav.clarity.scss
+++ b/src/clarity-angular/layout/nav/_sidenav.clarity.scss
@@ -29,12 +29,8 @@ $clr-sidenav-border-color: $clr-light-midtone-gray !default;
 
         .sidenav-content {
             flex: 1 1 auto;
-            overflow: hidden;
+            overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


youdz:
@adityarb88: Wasn't the point of the `overflow: hidden` + `:hover { overflow: scroll }` pattern (which is absolutely loathe) to not have the page jump around in some weird situation for non-Mac OS?

I just want to make sure we're not going in circles by fixing the same two bugs over and over, always negating the previous fix.

adityarb88:
@youdz Yes that's why we added those styles. But now I think we should get rid of them because they really don't solve the issue completely plus it introduces a bug in touch devices. What do you think? I think this change looks good to me. If the content moves when the scroll bar appears, we can't do much about it in Clarity UI. Vertical Nav doesn't use this pattern.

adityarb88:
@bojan88: Thanks for the contribution. Your name has been added to our community page: https://vmware.github.io/clarity/community

bojan88:
My pleasure. I will try to do it more often.

PR #1305, opened 2017-08-15
Improve the word wrapping of the stack-view

### LINKED ISSUE #1304: Content of the Stack View doesn't overflow correctly in Firefox
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
The expectation is the text to overflow on new line.

### Actual behavior
The text doesn't overflow and a scroller appears, because the text goes outside of the content.
This is reproducible only in Firefox.
image

### Reproduction of behavior
https://plnkr.co/edit/qdO0U7LCAOZciWm8eSHX?p=preview

### Environment details

* **Clarity version:** 
v0.10.0-patch 

* **OS and version:** 
Windows 10

* **Browser:**
Firefox 54.0.1 (64-bit) - Currently Latest

matonev:
When the content of the stack-view appears to be too long in some of the
browsers it doesn't fit in the container and scrollers appear.

This will:
* Allow the browser to brake unbreakable words, so the content fits in the container
* Avoid appearance of the scrollers around the stack-view

Fixes #1304

Before:
image

After:
image

### DIFF
--- a/src/clarity-angular/data/stack-view/_stack-view.clarity.scss
+++ b/src/clarity-angular/data/stack-view/_stack-view.clarity.scss
@@ -44,6 +44,7 @@ $clr-stack-font-weight: clr-getTypePropertyValueForDomElement(stackview_text, fo
         border-radius: $clr-default-borderradius;
         overflow-y: auto;
         background-color: $gray-lighter;
+        word-wrap: break-word;
 
         /* this fixes the overflow problem of children elements in a wrapper with a border-radius in Chrome */
         -webkit-mask-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAA5JREFUeNpiYGBgAAgwAAAEAAGbA+oJAAAAAElFTkSuQmCC);


src/clarity-angular/data/stack-view/_stack-view.clarity.scss
```diff
@@ -44,6 +44,7 @@ $clr-stack-font-weight: clr-getTypePropertyValueForDomElement(stackview_text, fo
         border-radius: $clr-default-borderradius;
         overflow-y: auto;
         background-color: $gray-lighter;
+        word-wrap: break-word;
```
adityarb88: Shouldn't this be `break-all`? https://developer.mozilla.org/en-US/docs/Web/CSS/word-break. I dont see `break-word` as a property value. Please correct me if I am wrong.

matonev: This is `word-wrap` not `word-break`:
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

The difference is that `break-all` will break between any two letters, while with `break-word` it will try to break the words and only if the browser cannot overflow but it should, then it will break the unbreakable words.

adityarb88: Oh right. I got confused between the two properties. Tested again. Works fine!

adityarb88:
@matonev: Thanks for the contribution. Your name has been added to our community page: https://vmware.github.io/clarity/community

PR #1323, opened 2017-08-17
[NG] refactor to use abstract popover

gnomeontherun:
Refactoring the component to also use the abstract popover so it conforms to the new API design.

Looking for a review but currently the bindings to check if the actions are open or not is not functional. Need to tidy that up and get the tests passing (yes there is an fdescribe in there).


gnomeontherun:
Declining for the time being, bugs were addressed separately.

Issue #1292, opened 2017-08-12
clr-dropdown doesnt close all dropdown menus when a nested menu item is clicked


envyN:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
If a dropdown has multiple levels and a user clicks on a menu item in any of the levels, the whole menu system should close.

### Actual behavior
Only the menu content of the clicked item and all after that are closed right now. does not close the whole dropdown

### Reproduction of behavior
Can be seen on the site itself Click on the angular dropdown button then link1 then bar then baz.
the menu system till foo and bar is not closed

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:
0.10.0

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@envyN: Thanks for reporting this. I do believe that all the menus should close when an item is clicked but we will get a confirmation from our UX team on the correct behavior and then decide what needs to be done based on their input. cc: @mathisscott, @reddolan

mathisscott:
@adityarb88 
The only instance where a dropdown menu should not close when an item is clicked is when an item is a form input. Think about clicking on a checkbox or radio button inside a dropdown.

Everytime a user clicks on a navigational element inside a dropdown menu, it should close all menus.

reddolan:
@adityarb88 I second @mathisscott's comment. All menus should close when a selection is made.

jeeyun:
Hmm this looks like a regression bug of some sort. It used to close all menus. I am able to reproduce on our doc site.

reddolan:
I was able to reproduce as well

PR #1342, opened 2017-08-21
[NG] Close All Menus on Item Click

adityarb88:
Fixes: #1292

Tested on Chrome


src/clarity-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -16,7 +16,11 @@ export class DropdownItem {
     @HostListener("click")
     onDropdownItemClick(): void {
         if (this.dropdown.isMenuClosable && !this.el.nativeElement.classList.contains("disabled")) {
-            this.ifOpenService.open = !this.ifOpenService.open;
```
hippee-lee: Each Dropdown instance gets its own `IfOpenService` right? Is `tempDropdown` needed here?

adityarb88: actually after discussing with @youdz I am changing the solution completely. Closing this PR

PR #1343, opened 2017-08-21
[NG] Close all menus on item click

### LINKED ISSUE #1292: clr-dropdown doesnt close all dropdown menus when a nested menu item is clicked

adityarb88:
Fixes #1292 

Tested on Chrome

### DIFF
--- a/src/clarity-angular/popover/dropdown/dropdown.ts
+++ b/src/clarity-angular/popover/dropdown/dropdown.ts
@@ -3,9 +3,11 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component, Input, Optional, SkipSelf} from "@angular/core";
+import {Component, Input, OnDestroy, Optional, SkipSelf} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
 
 import {IfOpenService} from "../../utils/conditional/if-open.service";
+import {ROOT_DROPDOWN_PROVIDER, RootDropdownService} from "./providers/dropdown.service";
 
 @Component({
     selector: "clr-dropdown",
@@ -15,10 +17,19 @@ import {IfOpenService} from "../../utils/conditional/if-open.service";
         // FIXME: remove this as soon as we stop supporting this old <div class="dropdown-menu"> syntax
         "[class.open]": "ifOpenService.open"
     },
-    providers: [IfOpenService]
+    providers: [IfOpenService, ROOT_DROPDOWN_PROVIDER]
 })
-export class Dropdown {
-    constructor(@SkipSelf() @Optional() public parent: Dropdown, public ifOpenService: IfOpenService) {}
+export class Dropdown implements OnDestroy {
+    private _subscription: Subscription;
+
+    constructor(@SkipSelf() @Optional() public parent: Dropdown, public ifOpenService: IfOpenService,
+                dropdownService: RootDropdownService) {
+        this._subscription = dropdownService.changes.subscribe(value => this.ifOpenService.open = value);
+    }
 
     @Input("clrCloseMenuOnItemClick") isMenuClosable: boolean = true;
+
+    ngOnDestroy() {
+        this._subscription.unsubscribe();
+    }
 }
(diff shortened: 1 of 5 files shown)


src/clarity-angular/popover/providers/dropdown.provider.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {DropdownService} from "./dropdown.service";
+
+export function clrDropdownProviderFactory(existing: DropdownService) {
```
youdz: You can put this inline in the providers declaration if you want, a separate file is a bit overkill.

src/clarity-angular/popover/dropdown/dropdown.ts
```diff
@@ -15,10 +18,25 @@ import {IfOpenService} from "../../utils/conditional/if-open.service";
         // FIXME: remove this as soon as we stop supporting this old <div class="dropdown-menu"> syntax
         "[class.open]": "ifOpenService.open"
     },
-    providers: [IfOpenService]
+    providers: [
+        IfOpenService, {
+            provide: DropdownService,
+            useFactory: clrDropdownProviderFactory,
+            deps: [[new Optional(), new SkipSelf(), DropdownService]]
+        }
```
youdz: If you do end up keeping a separate file, please move the actual Provider into it, rather than just the factory. The current halfway split doesn't make much sense. I know we did the same in TreeView, but I didn't pick it up during the review.

src/clarity-angular/popover/providers/dropdown.service.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Injectable} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+
+@Injectable()
+export class DropdownService {
```
youdz: Please rename this `RootDropdownService` or something similar. We want to make it clear that it's a singleton per dropdown + nested dropdowns, declared at the root, so that we can move some of the "if root" logic in there.

src/clarity-angular/popover/dropdown/dropdown.ts
```diff
@@ -15,10 +18,25 @@ import {IfOpenService} from "../../utils/conditional/if-open.service";
...
+    ]
 })
-export class Dropdown {
-    constructor(@SkipSelf() @Optional() public parent: Dropdown, public ifOpenService: IfOpenService) {}
+export class Dropdown implements OnDestroy {
+    private _subscription: Subscription;
+
+    constructor(@SkipSelf() @Optional() public parent: Dropdown, public ifOpenService: IfOpenService,
+                public dropdownService: DropdownService) {
```
youdz: This doesn't need to be a property on the object at the moment, you're only using it in the constructor. The linter will fail once we add typing. So you can just remove the `public` keyword here.

src/clarity-angular/popover/providers/dropdown.provider.ts
adityarb88: i am moving this to the same file as the service instead

src/clarity-angular/popover/dropdown/dropdown.ts
adityarb88: fixed

src/clarity-angular/popover/dropdown/dropdown.ts
adityarb88: fixed

src/clarity-angular/popover/providers/dropdown.service.ts
adityarb88: fixed

PR #1348, opened 2017-08-22
[NG] custom label for button group menu

### LINKED ISSUE #1347: [Button Group] Allow dropdowns
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

This shouldn't be hardcoded, I would propose this to be the default but overwritable. Like this:

- [clrMenuLabel]: String - This would replace the icon with a Label
- [clrMenuIcon]: String - This would just change the icon itself

### Actual behavior

https://github.com/vmware/clarity/blob/f0f79cd76d2480c5f0e78b45649e8139a1a8b75c/src/clarity-angular/button/button-group/button-group.html#L12
Hardcoded.

### Reproduction of behavior

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

marvinosswald:
`[clrMenuLabel]` on a `<clr-button-group>` sets a custom button label instead of the 'ellipsis-horizontal' icon

Partly Fixes #1347 

Signed-off-by: Marvin Osswald <mail@marvinosswald.de>

### DIFF
--- a/src/app/button-group/angular/button-group-angular.ts
+++ b/src/app/button-group/angular/button-group-angular.ts
@@ -19,6 +19,7 @@ import {Component} from "@angular/core";
             <li><a [routerLink]="['./hide-overflow']">Hide/Show Overflow Toggle</a></li>
             <li><a [routerLink]="['./mixed-buttons']">Mixed Buttons</a></li>
             <li><a [routerLink]="['./move-button-in-menu']">Move Button In Menu</a></li>
+            <li><a [routerLink]="['./move-button-in-menu-with-custom-label']">Move Button In Menu with Custom Label</a></li>
             <li><a [routerLink]="['./move-multiple-buttons-in-menu']">Move Multiple Buttons In Menu</a></li>
             <li><a [routerLink]="['./move-all-in-menu']">Move All Buttons In Menu</a></li>
             <li><a [routerLink]="['./projection-update-test-1']">Projection Update Test 1</a></li>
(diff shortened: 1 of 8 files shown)


mathisscott:
First off, I want to stress that you did good work here. This is a well-done piece of functionality. After discussing it at length with the team, however, our recommendation is not to include it in the Clarity Angular library.

The reasoning from our UX team is that it introduces an inconsistency in the interaction. Suddenly people may see some components where ellipses are used to present more options but then in other cases it will be text links… or further still, a different icon (if we were to continue this feature request to include a custom icon).

I feel really bad about this because you've done a good job. But it is not a feature that we can, at this time, push out to our users.

Please let me know if there's something I am misunderstanding about the use case here that might impact this recommendation.

marvinosswald:
@mathisscott while I understand your reasoning I want to lay down my case I've created this fix for.

Originally all I wanted was to have a dropdown included in a button group, but if I would just "put it there" in terms of code the toggle button wouldn't recognise that itself is inside a button group and would just render as a normal button. So I've found the "ellipses solution" and run with it until someone didn't understand the meaning of these three little dots... which made it a requirement for me to label it. 

But if it was just was a personal limitation of myself to implement a "normal" dropdown in a button group please let me know.

mathisscott:
@youdz and I had a discussion about this earlier today. I'll give him a chance to post his feedback. Ultimately, I think your use case is more an issue with allowing for dropdowns to be triggered inside of button groups and allowing button groups to handle that (even from the overflow).

youdz:
As @mathisscott said, I didn't realize your use case was using actual dropdowns inside a button group. 

This is a perfectly valid feature request that we have been planning for from the first version of button groups. At the time, we did not have "nested menus" yet, so we could not do it as part of the MVP. Now that we have these nested menus, we can start allowing dropdowns in button groups, and any dropdown that ends up in the "overflow" (the ellipsis dropdown) will become a nested dropdown itself. So in your case, you would just add a dropdown directly in the button group, with your own text or icon for the button, and you would just have to not put it in the overflow. 😉 

I will update your original issue to reflect this, and we will add it to our backlog. Thanks! 😄

PR #1356, opened 2017-08-23
[ICONS] remove dompurify from Clarity Icons

### LINKED ISSUE #1253: Very slow loading in Internet Explorer 11
```
[x] bug
```

### Expected behavior
Clarity application should be loaded without significant delays (as in Chrome/Firefox)

### Actual behavior
Loading of Clarity application in IE11 taking about 20 sec. 
Usually, IE11 is the only one allowed browser in a big companies. Was surprised to see what "enterprise-oriented" Clarity does not work with IE11:(
Investigation shows what issue is caused by the usage of html5 custom elements to render SVG icons, probably related to a known issue of IE with DOM fragments (https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/4561410/).

### Reproduction of behavior
Open https://vmware.github.io/clarity/documentation with Internet Explorer 11

### Environment details
* **Browser:** IE 11

Shijir:
Fixes #1253


Shijir:
@mathisscott I think so. As it fixes the load speed issue, so it could be considered as a bug/issue fix 🤔

mathisscott:
I agree. Let's try to get it into 0.9 this week, if we can.

PR #1358, opened 2017-08-23
Implementation of whole row selection 

### LINKED ISSUE #476: Implement selection of whole row when no clickable items in the row in datagrid
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

Need to evaluate if this is something we need to provide as a built-in option. There's no technical constraint even in the current version (see plunkr). However this pattern should never be used if your datagrid's row contains clickable cells. Need more input from UX side.

### Reproduction of behavior

https://plnkr.co/edit/Cwp2irAbmozAdivFelwg?p=preview

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** All

* **OS and version:** All

* **Browser:** All

matonev:
Implementation of whole row selection

* New `clDgRowSelection` input property for Datagrid indicating if the row selection should be enabled.
  Selection works the same way if you click on the input in the first column.
  For Multi selection on second click the row gets deselected.
* New `clDgRowSelection` input property for Datagrid indicating if the selection column should be available
  This automatically turns on `clDgRowSelection` in case of Single and Multi selection
* Added support for row actions in case of Row selection enabled
* Added "Selection (Row mode)" demo page on Datagrid demo page with examples of the newly introduced functunality

Fixes #476

Signed-off-by: Mincho Tonev <mtonev@vmware.com>

### DIFF
--- a/src/app/datagrid/selection-row-mode/selection-row-mode.html
+++ b/src/app/datagrid/selection-row-mode/selection-row-mode.html
@@ -1,0 +1,118 @@
+<!--
+  ~ Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Selection (Row mode)</h2>
+
+<p>
+    Depending on the use case, you might want to allow selection when clicking anywhere in the grid row.
+</p>
+<p>
+    Note that this is not recommeded when the rows are containing clickable cells. If your cells contain input, buttons, etc.
+    this option is not recommended
+</p>
+<ul>
+    <li>
+        For row selection mode add <code class="clr-code">[(clDgRowSelection)]</code> and set it to <code class="clr-code">true</code>.
+    </li>
+    <li>
+        For hiding the selection column add <code class="clr-code">[(clDgHideSelectionColumn)]</code> and set it to <code class="clr-code">true</code>.
+    </li>
+</ul>
+
+<p>
+    In the following examples, we simply display the names of the selected users
+</p>
+
+<div class="card card-block">
+    <p class="card-text username-list">
+        Selected user:
+        <em *ngIf="!singleSelected">No user selected.</em>
+        <span class="username" *ngIf="singleSelected">{{singleSelected.name}}</span>
+    </p>
+</div>
+
+<h3>Single selection</h3>
+
+<clr-datagrid [(clrDgSingleSelected)]="singleSelected" [clDgRowSelection]="true">
+
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>Name</clr-dg-column>
+    <clr-dg-column>Creation date</clr-dg-column>
+    <clr-dg-column>Favorite color</clr-dg-column>
+
+    <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>{{users.length}} users</clr-dg-footer>
+</clr-datagrid>
+
+<clr-example [clrCode]="singleSelectionExample" clrLanguage="html"></clr-example>
+
+<h3>Multi selection</h3>
+
+<div class="card card-block">
+    <p class="card-text username-list">
+        Selected users:
+        <em *ngIf="multiSelected.length == 0">No user selected.</em>
+        <span class="username" *ngFor="let user of multiSelected">{{user.name}}</span>
+    </p>
+</div>
+
+<clr-datagrid [(clrDgSelected)]="multiSelected" [clDgRowSelection]="true">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>Name</clr-dg-column>
+    <clr-dg-column>Creation date</clr-dg-column>
+    <clr-dg-column>Favorite color</clr-dg-column>
+
+    <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>{{users.length}} users</clr-dg-footer>
+</clr-datagrid>
+
+<clr-example [clrCode]="multiSelectionExample" clrLanguage="html"></clr-example>
+
+<h3>Multi selection with hidden selection column</h3>
+
+<div class="card card-block">
+    <p class="card-text username-list">
+        Selected users:
+        <em *ngIf="hiddenMultiSelected.length == 0">No user selected.</em>
+        <span class="username" *ngFor="let user of hiddenMultiSelected">{{user.name}}</span>
+    </p>
+</div>
+
+<clr-datagrid [(clrDgSelected)]="hiddenMultiSelected" clDgHideSelectionColumn="true">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>Name</clr-dg-column>
+    <clr-dg-column>Creation date</clr-dg-column>
+    <clr-dg-column>Favorite color</clr-dg-column>
+
+    <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>{{users.length}} users</clr-dg-footer>
+</clr-datagrid>
+
+<clr-example [clrCode]="hiddenSelectionExample" clrLanguage="html"></clr-example>
--- a/src/app/datagrid/selection-row-mode/selection-row-mode.html
+++ b/src/app/datagrid/selection-row-mode/selection-row-mode.html
@@ -17,9 +17,6 @@
     <li>
         For row selection mode add <code class="clr-code">[(clDgRowSelection)]</code> and set it to <code class="clr-code">true</code>.
     </li>
-    <li>
-        For hiding the selection column add <code class="clr-code">[(clDgHideSelectionColumn)]</code> and set it to <code class="clr-code">true</code>.
-    </li>
 </ul>
 
 <p>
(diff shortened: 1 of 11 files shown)


src/clarity-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -151,6 +151,82 @@ export default function(): void {
...
+                cell.click();
+                context.detectChanges();
+                expect(selectionProvider.currentSingle).toEqual(context.testComponent.item);
+            });
+
+            it("selects the model on click only when `rowSelectionMode` is enabled (Multi selection)", function() {
+                selectionProvider.selectionType = SelectionType.Multi;
+                context.testComponent.item = {id: 1};
+                context.detectChanges();
+                const cell = context.clarityElement.querySelector("clr-dg-cell");
```
youdz: I'm not a fan of adding `<clr-dg-cell>` elements to these unit tests. It breaks the isolation, we actually test 2 components at the same time. I don't see why you need it here.

src/clarity-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -151,6 +151,82 @@ export default function(): void {
...
+                expect(selectionProvider.current).toEqual([]);
+                cell.click();
```
youdz: Why not just click the row? It seems more straightforward.

src/clarity-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -151,6 +151,82 @@ export default function(): void {
...
+                expect(input).toBeDefined();
+
+                selectionProvider.hideSelectionColumn = true;
+                context.detectChanges();
+                input = context.clarityElement.querySelector("input");
+                expect(input).toBeNull();
+            });
+
+            it("doesn't have input for selection when hideSelectionColumn is enabled (Multi selection)", function() {
+                selectionProvider.selectionType = SelectionType.Single;
```
youdz: The description of the test says multi-selection, but you're still using single. Happens to all of us on copy-paste. 😄

youdz:
Oh, and I forgot to mention: if you have questions or need help regarding some of my requests, feel free to ask directly in the comments and I'll answer as soon as I can. Thanks again!

src/app/datagrid/selection-row-mode/selection-row-mode.html
```diff
@@ -0,0 +1,122 @@
...
+    <p class="card-text username-list">
+        Selected user:
+        <em *ngIf="!singleSelected">No user selected.</em>
+        <span class="username" *ngIf="singleSelected">{{singleSelected.name}}</span>
+    </p>
+</div>
+
+<h3>Single selection</h3>
+
+<clr-datagrid [(clrDgSingleSelected)]="singleSelected" clDgRowSelection="true">
```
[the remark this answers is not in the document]
matonev: Done.

src/app/datagrid/selection-row-mode/selection-row-mode.html
```diff
@@ -0,0 +1,122 @@
...
+<clr-datagrid [(clrDgSelected)]="multiSelected" clDgRowSelection="true">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>Name</clr-dg-column>
+    <clr-dg-column>Creation date</clr-dg-column>
+    <clr-dg-column>Favorite color</clr-dg-column>
+
+    <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-action-overflow>
+            <button class="action-item">Action</button>
+        </clr-dg-action-overflow>
```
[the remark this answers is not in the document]
matonev: I thought that you would said that :)
I just wanted to support that feature as well, but I agree that it isn't natural.
Removing it from the example.

src/app/datagrid/selection-row-mode/selection-row-mode.html
```diff
@@ -0,0 +1,122 @@
...
+
+<div class="card card-block">
+    <p class="card-text username-list">
+        Selected users:
+        <em *ngIf="multiSelected.length == 0">No user selected.</em>
+        <span class="username" *ngFor="let user of multiSelected">{{user.name}}</span>
+    </p>
+</div>
+
+<clr-datagrid [(clrDgSelected)]="multiSelected" clDgRowSelection="true">
```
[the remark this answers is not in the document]
matonev: That's true, but I have the following concerns with putting of that property on `<clr-dg-row>`:
* This way we give the power to the consumers to have inconsistent selection. - You might miss and put this on only of half of the rows.
* All the selection properties are put on `<clr-datagrid>` level, so it seems nature as a consumer to expect such feature to be there

src/app/datagrid/selection-row-mode/selection-row-mode.html
```diff
@@ -0,0 +1,122 @@
...
+
+<div class="card card-block">
+    <p class="card-text username-list">
+        Selected users:
+        <em *ngIf="hiddenMultiSelected.length == 0">No user selected.</em>
+        <span class="username" *ngFor="let user of hiddenMultiSelected">{{user.name}}</span>
+    </p>
+</div>
+
+<clr-datagrid [(clrDgSelected)]="hiddenMultiSelected" clDgHideSelectionColumn="true">
```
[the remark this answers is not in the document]
matonev: I disagree, because:
* Select all functionality (Checkbox in the first column's header) won't be supported, but also might need such a feature
* You might want to be explicit when showing the selection, so sometimes you might want to have row selection + checkboxes as a first column. This is widely use in our case when we show datagrids in dialogs.

src/clarity-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -79,4 +82,11 @@ export class DatagridActionOverflow implements OnDestroy {
         }
         this.open = false;
     }
+
+    /**
+     * Indicates that this is menu interaction
+     */
+    public menuInteraction(event: any) {
+        event.menuInteraction = true;
```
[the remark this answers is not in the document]
matonev: I agree. Since `<clr-dg-action-overflow>` won't be supported in grid having row selection enabled, I'm removing this.

src/clarity-angular/data/datagrid/datagrid-row.ts
```diff
@@ -22,12 +22,12 @@ let nbRow: number = 0;
 @Component({
     selector: "clr-dg-row",
     template: `
-        <clr-dg-row-master class="datagrid-row-flex">
-            <clr-dg-cell *ngIf="selection.selectionType === SELECTION_TYPE.Multi"
+        <clr-dg-row-master class="datagrid-row-flex" (click)="toggleSelection($event)">
```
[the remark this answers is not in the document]
matonev: Thanks! Done.

src/clarity-angular/data/datagrid/datagrid-row.spec.ts
matonev: Fixed.

src/clarity-angular/data/datagrid/datagrid-row.spec.ts
matonev: Agree. Fixed.

src/clarity-angular/data/datagrid/datagrid-row.spec.ts
matonev: Fixed. Thanks :)

matonev:
Thanks for the review, Eudes!
In general I applied most of your comments.
The only two which I didn't applied are:
* Remove the ability to display checkboxes and at the same time the grid to be in row selection mode
* `clDgRowSelection` to be property of `<clr-dg-row>`

Once we agree on the implementation and use-cases, I will update the pull request.

matonev:
Hey @youdz,

There are three tests failing as part of the gemini test suite.
I'm having trouble running those on my windows machine, I will really appreciate if you can give me a hand and look at them.
Thanks!

adityarb88:
Hi @matonev. Could you please rebase your PR on master? Some images from other components were failing. I fixed them and once you rebase, the tests should pass too. Please let me know if that works.

youdz:
Hi @matonev, I haven't had time to review your recent changes yet, but just know that Travis and Gemini are just playing hard these days, we're trying to solve the issues but we still have random failures happening. The tests failing on this PR have nothing to do with what you changed, so I believe you can just ignore them, not worry about it. We'll handle them however we can.

I will hopefully find time to review your latest changes before the end of the week. 😄

adityarb88:
@matonev: Actually could you please hold off on the rebase and squash? @youdz will first review the code based on your new commit. The CSS regression failures are not related to this PR. Once we get the approval from @youdz on this PR, we can rebase and fix the regression issues later on. Thanks a lot again for the contribution!

src/clarity-angular/data/datagrid/datagrid-row.ts
```diff
@@ -65,7 +74,11 @@ let nbRow: number = 0;
             <ng-content select="clr-dg-row-detail"></ng-content>
         </ng-template>
     `,
-    host: {"[class.datagrid-row]": "true", "[class.datagrid-selected]": "selected"},
+    host: {
+        "[class.datagrid-row]": "true",
+        "[class.datagrid-selected]": "selected",
+        "[attr.tabindex]": "selection.rowSelectionMode ? 0 : null"
```
youdz: You're still missing `role="button"`. 😉 
You made it tabbable but didn't indicate it was clickable.

To be honest, the best solution would be to have some ARIA attribute that lets you indicate the full row as a `label` for the checkbox or radio inside of it, but I could find one after a quick research. Maybe you might have heard of one.

youdz:
A lot of the code looks much better without the action overflow compatibility. 😄 Thanks for making these changes.

So I had 2 meetings today about the question of showing the checkboxes/radios when the full row is clickable or not. It turns out I was way off: the case we cannot support is actually hiding them. It would not be accessible anymore. The explanation is that when a user is in high-contrast mode, backgrounds are not displayed. So without the checkbox or radio, they would have no way of knowing which rows are currently selected and which ones aren't. On top of that blocker argument (which settles the decision by itself, yay accessibility), the concern from the UX side was also that without checkboxes or radios, the user would have absolutely no way of knowing if they can select one or more items. Or even if they can select items at all.

That means we have to drop the `clDgHideSelectionColumn` input. I'm sorry you spent time implementing it to drop it this late in the process, and we'll try to figure out a way to avoid this situation happening in the future.

Finally, there is one behavior I would like you to check. You have the checkbox in the row, which is itself clickable and toggles the selection. So you still have the "checkbox in a button" concern. For checkboxes, it's much more natural since this is actually valid: 
```html
<label>
  My label
  <input type="checkbox" />
</label>
```
The label behaves like a button in practice, since clicking on it toggles the checkbox. So we're fine here. My concern comes from the fact that naive implementation leads to a broken behavior, as you can see here: https://plnkr.co/edit/dBeC5uSMqPeM4gGUXK3c?p=preview
If you click anywhere on the button, the checkbox correctly toggles. But if you click on the checkbox itself, you have a conflict between the listeners and it starts behaving erratically.
Could you please check the behavior in your case when you click directly on the checkbox? Make sure it toggles on and off properly.

Thanks for all your work on this PR, from all the team! 👍

youdz:
@matonev: Just checking in to see if you were still planning on following this through.
We would absolutely understand if our last request was the last straw, in which case we would use your PR and update it ourselves. We would of course preserve your commit so you keep credit.

It's up to you! 😄

matonev:
@youdz Yes, I plan to address your comments. I just didn't have enough time to do this.
I hope that I will do it next week.

youdz:
No problem, take all the time you need, we'll leave this untouched so you can fully own it. 👍

src/clarity-angular/data/datagrid/datagrid-row.ts
matonev: Added the `role="button"` attribute and implemented the selection with space and enter.
This is the recommended approach.

matonev:
@youdz It is working correctly when clicking on the checkbox in the first column. It also puts the focus on the checkbox not the row, which is correct.
It is visible in the example of the Row Selection. Basically the checkbox prevents the propagation on the event.

matonev:
Thanks, guys! Can't wait to use it and remove all the workarounds in our code base.

jacobbutton:
Is there a way to use this to expand the details row instead of selecting the row?

adityarb88:
@jacobbutton: No we don't. But the implementation would be similar to this. Please feel free to submit a PR (if you do please make sure that a click performs only one action. either selection or expansion). All the logic should be the same. You would just have to switch the selection to expansion. If you are unable to submit a PR due to time constraints then I would recommend that you open a new issue for this. We can add a `contribution-welcome`/`help-wanted` label on that.

PR #1394, opened 2017-08-30
[NG] fix datagrid single selection to deselect when value is set to null

gnomeontherun:
This allows you to set the `clrDgSingleSelected` value to null, and clear the selection of that row. The previous check did not account for going from selected to null if you passed a non-truthy value.

It also adds plumber to the typescript gulp task so it doesn't die so easily on compile errors.

### DIFF
--- a/build/compile-ts.js
+++ b/build/compile-ts.js
@@ -5,6 +5,7 @@
  */
 
 var gulp = require("gulp");
+var plumber = require("gulp-plumber");
 var replace = require('gulp-replace');
 var sourcemaps = require("gulp-sourcemaps");
 var inlineNg2Template = require("gulp-inline-ng2-template");
@@ -31,6 +32,7 @@ module.exports = function(tsSources, options, destination) {
     var prod = process.env.NODE_ENV==="prod";
 
     var stream = gulp.src(allSources, {base: "src/"})
+        .pipe(plumber())
         // replace .scss references in styleUrls: [...] in the files
         .pipe(replace(/styleUrls\s*:\s*\[([^\]]*)\]/g, function(match, group) {
             // here, match = styleUrls: [...]  and group = string inside []
--- a/src/clarity-angular/data/datagrid/datagrid.spec.ts
+++ b/src/clarity-angular/data/datagrid/datagrid.spec.ts
@@ -248,6 +248,52 @@ export default function(): void {
             });
         });
 
+        describe("Single selection", function() {
+            let context: TestContext<Datagrid, SingleSelectionTest>;
+            let selection: Selection;
+
+            beforeEach(function() {
+                context = this.create(Datagrid, SingleSelectionTest, [Selection]);
+                selection = context.getClarityProvider(Selection);
+            });
+
+            describe("TypeScript API", function() {
+                // None for now, would duplicate tests of Selection provider
+            });
+
+            describe("Template API", function() {
+                it("sets the currentSingle binding", function() {
+                    expect(selection.currentSingle).toBeNull();
+                    context.testComponent.selected = 1;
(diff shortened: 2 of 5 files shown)


build/compile-ts.js
```diff
@@ -5,6 +5,7 @@
  */
 
 var gulp = require("gulp");
+var plumber = require("gulp-plumber");
```
mathisscott: What issue is `gulp-plumber` addressing?

mathisscott:
FYI, we wait for two approvals before merging. Please make sure @youdz takes a look at this before merging.

Thanks!

build/compile-ts.js
gnomeontherun: While running `gulp serve` the compile Typescript task can throw an error into the pipe when you have a syntax error, so plumber monkey patches NodeJS pipe to not kill the stream when an error is thrown.

src/clarity-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -356,6 +378,26 @@ class TrackByTest {
...
+            <clr-dg-cell>{{item.value * item.value}}</clr-dg-cell>
+        </clr-dg-row>
+    
+        <clr-dg-footer (click)="selected = null">{{selected?.value}}</clr-dg-footer>
+    </clr-datagrid>
+`
+})
+class SingleSelectionTest {
+    items = [{value: 1}, {value: 2}, {value: 3}];
+    selected: number;
```
youdz: I know we don't run our unit tests in AoT yet, but this would fail. Your `selected` property receives an object of type `{value: number}`, not an actual `number`.

Anyway, why switch to objects and not use simple numbers like the other tests? Especially since testing object equality in Jasmine opens up a traps that could easily be avoided with native numbers.

src/clarity-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -248,6 +248,28 @@ export default function(): void {
...
+                expect(context.selected).toBeUndefined();
+                const radio = context.clarityElement.querySelectorAll(".datagrid-select input[type='radio']")[0];
+                const row = context.clarityElement.querySelectorAll(".datagrid-row")[1];
+                const foot = context.clarityElement.querySelector(".datagrid-foot-description");
+                expect(row.classList.contains("datagrid-selected")).toBeFalsy();
+                // Select a row
+                context.testComponent.selected = {value: 1};
+                radio.click();
+                context.detectChanges();
+                expect(row.classList.contains("datagrid-selected")).toBeTruthy();
```
youdz: This unit test is really not "unit". The fact that you're grabbing shadow DOM elements, clicking on stuff, checking the result by looking up some class in the shadow DOM again, ... means your test is extremely fragile and is more akin to an end-to-end test.

You should use the two-way binding, check that the Selection provider of the Datagrid has been updated, and trust the unit test that check the click listeners, the view bindings and the dynamic CSS classes to do their job.

So your unit test would look like:
```typescript
it("allows to unselect in single selection mode", function() {
    const selection: Selection = context.getClarityProvider(Selection);
    selection.currentSingle = 2;
    context.detectChanges();
    context.testComponent.selected = null;
    context.detectChanges();
    expect(selection.currentSingle).toBeNull();
});
```

gnomeontherun: I wrote it this way to track the bug that was found, where it would not fully unselect onclick, because the styling remained like it was active but the model could be set to `null`. The only way to test this is to inspect the DOM for classes to be applied or not. Your version is fine to test the bindings, but doesn't verify the bug was fixed.

PR #1408, opened 2017-09-01
[ICONS] let users set custom title to icons

Shijir:
Tested on: Chrome, FF, Safari, Edge, IE 11

### DIFF
--- a/src/clarity-icons/clarity-icons-element.ts
+++ b/src/clarity-icons/clarity-icons-element.ts
@@ -7,6 +7,10 @@ import {ClarityIconsApi} from "./clarity-icons-api";
 
 /* CLR-ICON CUSTOM ELEMENT */
 
+let clrIconId = 0;
+const offScreenSpan = document.createElement("span");
+offScreenSpan.className = "is-off-screen";
+
 let parentConstructor = function() {
     return HTMLElement.apply(this, arguments);
 };
@@ -18,62 +22,143 @@ if (typeof Reflect === "object") {
 
 export function ClarityIconElement() {
     "use strict";
-    return (parentConstructor as any).apply(this, arguments);
+
+    const _this = (parentConstructor as any).apply(this, arguments);
+
+    _this.clrIconUniqId = "_clr_icon_" + clrIconId;
+    clrIconId++;
+
+    return _this;
 }
 
-(ClarityIconElement as any).observedAttributes = ["shape", "size"];
+(ClarityIconElement as any).observedAttributes = ["shape", "size", "title"];
 
 ClarityIconElement.prototype = Object.create(
     HTMLElement.prototype, {constructor: {configurable: true, writable: true, value: ClarityIconElement}});
 
 ClarityIconElement.prototype.constructor = ClarityIconElement;
 
-const generateIcon = function(element: any, shape: string) {
-    shape = shape.split(/\s/)[0];
-
-    if (shape !== element._shape) {
-        element._shape = shape;
-
-        // shape exists in set
-        if (ClarityIconsApi.instance.has(shape)) {
-            element.innerHTML = ClarityIconsApi.instance.get(shape);
-        } else {
-            console.error(`'${shape}' is not found in the Clarity Icons set.`);
-            element.innerHTML = ClarityIconsApi.instance.get("error");
-        }
-    }
+ClarityIconElement.prototype._appendCustomTitle = function() {
+    const cloneOffScreenSpan = <HTMLElement>offScreenSpan.cloneNode(false);
+    cloneOffScreenSpan.id = this.clrIconUniqId;
+    cloneOffScreenSpan.textContent = this.currentTitleAttrVal;
+    this.appendChild(cloneOffScreenSpan);
 };
 
-const setIconSize = function(element: any, size: string) {
+ClarityIconElement.prototype._setIconSize = function(size: string) {
     if (!Number(size) || Number(size) < 0) {
-        element.style.width = null;   // fallback to the original stylesheet value
-        element.style.height = null;  // fallback to the original stylesheet value
+        this.style.width = null;   // fallback to the original stylesheet value
+        this.style.height = null;  // fallback to the original stylesheet value
     } else {
-        element.style.width = size + "px";
-        element.style.height = size + "px";
+        this.style.width = size + "px";
+        this.style.height = size + "px";
     }
 };
 
 ClarityIconElement.prototype.connectedCallback = function() {
-    const host = this as HTMLElement;
+    // One thing to note here is that the attributeChangedCallback method is called for every attribute first
+    // before this connectedCallback method called only once when the custom element is inserted into the DOM.
+    // So we could check whether the attribute values really changed or not.
+    // If not, we don't need to execute the same codes again.
 
-    if (host.hasAttribute("shape")) {
-        generateIcon(host, host.getAttribute("shape"));
+    if (this.hasAttribute("size")) {
+        const sizeAttrValue = this.getAttribute("size");
+
+        if (this.currentSizeAttrVal !== sizeAttrValue) {
+            this.currentSizeAttrVal = sizeAttrValue;
+            this._setIconSize(sizeAttrValue);
+        }
     }
 
-    if (host.hasAttribute("size")) {
-        setIconSize(host, host.getAttribute("size"));
+    // Note: the size attribute is irrelevant from the shape template;
+    // That's why the size checking placed before detecting changes in shape and title attributes.
+    // This means even if the shape is not found, the injected shape will have the user-given size.
+
+    if (this.hasAttribute("shape")) {
+        const shapeAttrValue = this.getAttribute("shape").split(/\s/)[0];
+
+        if (shapeAttrValue === this.currentShapeAttrVal) {
+            return;
+        }
+
+        this.currentShapeAttrVal = shapeAttrValue;
+
+        if (ClarityIconsApi.instance.has(this.currentShapeAttrVal)) {
+            this.currentShapeTemplate = ClarityIconsApi.instance.get(this.currentShapeAttrVal);
+        } else {
+            this._injectErrorTemplate();
+            return;
+        }
     }
+
+
+    if (this.hasAttribute("title")) {
+        const titleAttrValue = this.getAttribute("title");
+
+        if (this.currentTitleAttrVal !== titleAttrValue) {
+            this.currentTitleAttrVal = titleAttrValue;
+        }
+
+        if (!this.currentShapeAttrVal) {
+            return;
+        }
+    }
+
+    this._injectTemplate();
 };
 
 ClarityIconElement.prototype.attributeChangedCallback = function(attributeName: string, oldValue: string,
                                                                  newValue: string) {
-    const host = this as HTMLElement;
+    if (attributeName === "size") {
+        this._setIconSize(newValue);
+    }
+
+    // Note: the size attribute is irrelavent from the shape template;
+    // That's why the size checking placced before detecting changes in shape and title attributes.
+    // This means even if the shape is not found, the injected shape will have the user-given size.
 
     if (attributeName === "shape") {
-        generateIcon(host, newValue);
+        this.currentShapeAttrVal = newValue.split(/\s/)[0];
+
+        if (ClarityIconsApi.instance.has(this.currentShapeAttrVal)) {
+            this.currentShapeTemplate = ClarityIconsApi.instance.get(this.currentShapeAttrVal);
+        } else {
+            this._injectErrorTemplate();
+            return;
+        }
     }
-    if (attributeName === "size") {
-        setIconSize(host, newValue);
+    if (attributeName === "title") {
+        this.currentTitleAttrVal = newValue;
+
+        if (!this.currentShapeAttrVal) {
+            return;
+        }
     }
+
+    this._injectTemplate();
+};
+
+ClarityIconElement.prototype._setAriaLabelledBy = function() {
+    const existingAriaLabelledBy: string = this.getAttribute("aria-labelledby");
+    if (!existingAriaLabelledBy) {
+        this.setAttribute("aria-labelledby", this.clrIconUniqId);
+    } else if (existingAriaLabelledBy && existingAriaLabelledBy.indexOf(this.clrIconUniqId) < 0) {
+        this.setAttribute("aria-labelledby", existingAriaLabelledBy + " " + this.clrIconUniqId);
+    }
+};
+
+ClarityIconElement.prototype._injectTemplate = function() {
+
+    this.innerHTML = this.currentShapeTemplate;
+
+    if (this.currentTitleAttrVal) {
+        this._setAriaLabelledBy();
+        this._appendCustomTitle();
+    }
+};
+
+ClarityIconElement.prototype._injectErrorTemplate = function() {
+    console.error(`'${this.currentShapeAttrVal}' is not found in the Clarity Icons set.`);
+    this.currentShapeTemplate = ClarityIconsApi.instance.get("error");
+    this._injectTemplate();
 };
--- a/src/clarity-icons/clarity-icons.scss
+++ b/src/clarity-icons/clarity-icons.scss
@@ -6,7 +6,6 @@
  * Clarity v@VERSION | MIT license | https://github.com/vmware/clarity
  *
  */
-
 $clr-icon-size: 16px;
 $clr-icon-color-success: #318700;
 $clr-icon-color-danger: #e62700;
@@ -99,9 +98,13 @@ clr-icon {
         height: 100%;
         width: 100%;
         display: block;
+        pointer-events: none;
     }
     & > svg {
         transition: inherit;
+        title {
+            display: none;
+        }
     }
     .clr-i-solid,
     .clr-i-solid--badged,
@@ -184,5 +187,19 @@ clr-icon {
         .clr-i-alert {
             @include setIconColor($clr-icon-color-warning);
         }
+    }
+    .is-off-screen {
+        position: fixed !important;
+        border: none !important;
+        height: 1px !important;
+        width: 1px !important;
+
+        left: 0px !important;
+        top: -1px !important;
+
+        overflow: hidden !important;
+
+        padding: 0 !important;
+        margin: 0 0 -1px 0 !important;
     }
 }
--- a/src/clarity-icons/clarity-icons.spec.ts
+++ b/src/clarity-icons/clarity-icons.spec.ts
@@ -3,8 +3,8 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-
-import {removeWhitespace, resetShapes, testAllShapes} from "./helpers.spec";
+import {ClarityIconElement} from "./clarity-icons-element";
+import {getErrorShape, giveAngleShapeTitle, removeWhitespace, resetShapes, testAllShapes} from "./helpers.spec";
 import {ClarityIcons} from "./index";
 import {AllShapes} from "./shapes/all-shapes";
 import {CommerceShapes} from "./shapes/commerce-shapes";
@@ -16,7 +16,6 @@ import {TechnologyShapes} from "./shapes/technology-shapes";
 import {TravelShapes} from "./shapes/travel-shapes";
 
 describe("ClarityIcons", () => {
-
     afterEach(() => {
         resetShapes();
     });
@@ -28,9 +27,7 @@ describe("ClarityIcons", () => {
     });
 
     describe("ClarityIconsApi.get()", () => {
-
         it("should return all icons when no argument is passed in", () => {
-
             ClarityIcons.add(CommerceShapes);
             ClarityIcons.add(EssentialShapes);
             ClarityIcons.add(SocialShapes);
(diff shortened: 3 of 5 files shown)


src/clarity-icons/clarity-icons-element.ts
```diff
@@ -20,28 +22,24 @@ export function ClarityIconElement() {
...
-        }
+const replaceTitleWithCustom = function(currentShapeTemplate: string, currentTitleAttrVal: string,
+                                        clrIconUniqId: string) {
+    // if the title tag doesn't exist right after that first svg tag, we will inject a
+    if (currentShapeTemplate.indexOf("<title") > -1 && currentShapeTemplate.indexOf("</title>") > -1) {
+        // if title tag exists, remove that fist
+        currentShapeTemplate = currentShapeTemplate.replace(/<title[^>]*>([^<]+)<\/title>/i, "");
     }
+
+    const openingSVGTag = currentShapeTemplate.match(/<svg\b[^>]*>/i)[0];  // will find the first occurrence of svg tag
```
mathisscott: Icons can be added that are not SVG. Does this code still support that?

Shijir: oh totally forgot about that. I will fix that.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -20,28 +22,30 @@ export function ClarityIconElement() {
...
-        if (ClarityIconsApi.instance.has(shape)) {
-            element.innerHTML = ClarityIconsApi.instance.get(shape);
-        } else {
-            console.error(`'${shape}' is not found in the Clarity Icons set.`);
-            element.innerHTML = ClarityIconsApi.instance.get("error");
-        }
+    // if the title tag doesn't exist right after that first svg tag, we will inject a
+    if (currentShapeTemplate.indexOf("<title") > -1 && currentShapeTemplate.indexOf("</title>") > -1) {
+        // if title tag exists, remove that fist
+        currentShapeTemplate = currentShapeTemplate.replace(/<title[^>]*>([^<]+)<\/title>/i, "");
```
youdz: Isn't it risky to nuke the title instead of replacing its content directly?
I could see it being an issue on custom SVGs the consumer adds to Clarity, they might have extra attributes on the title that we are not preserving here.

youdz: Another comment here: you literally just grabbed the indexes of the beginning and end of the title tag, do you really have to do another regex search on a potentially large string?
```typescript
currentShapeTemplate = currentShapeTemplate.slice(0, titleStart) + currentShapeTemplate.slice(titleEnd+8)
```
I believe that would be more performant, and it's pretty free.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -20,28 +22,30 @@ export function ClarityIconElement() {
...
 ClarityIconElement.prototype = Object.create(
     HTMLElement.prototype, {constructor: {configurable: true, writable: true, value: ClarityIconElement}});
 
 ClarityIconElement.prototype.constructor = ClarityIconElement;
 
-const generateIcon = function(element: any, shape: string) {
-    shape = shape.split(/\s/)[0];
+const replaceTitleWithCustom = function(currentShapeTemplate: string, currentTitleAttrVal: string,
+                                        clrIconUniqId: string) {
+    const hasSVGtag = currentShapeTemplate.match(/<svg\b[^>]*>/i);  // will find the first occurrence of svg tag
```
youdz: Do you need a regex here? `indexOf()` with plain strings is way faster than regex matching, and I think matching `"<svg"` is enough, isn't it?

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -20,28 +22,30 @@ export function ClarityIconElement() {
...
-            element.innerHTML = ClarityIconsApi.instance.get("error");
-        }
+    // if the title tag doesn't exist right after that first svg tag, we will inject a
+    if (currentShapeTemplate.indexOf("<title") > -1 && currentShapeTemplate.indexOf("</title>") > -1) {
+        // if title tag exists, remove that fist
+        currentShapeTemplate = currentShapeTemplate.replace(/<title[^>]*>([^<]+)<\/title>/i, "");
     }
+
+    const openingSVGTag = hasSVGtag[0];
+    const remainingSVG = currentShapeTemplate.split(/<svg\b[^>]*>/i)[1];
```
youdz: You literally did this exact regex match 15 lines above. You should just use that result.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
+            return;
+        }
+
+        this.currentShapeAttrVal = shapeAttrValue;
+
+        if (ClarityIconsApi.instance.has(this.currentShapeAttrVal)) {
+            this.currentShapeTemplate = ClarityIconsApi.instance.get(this.currentShapeAttrVal);
+        } else {
+            console.error(`'${this.currentShapeAttrVal}' is not found in the Clarity Icons set.`);
+            host.innerHTML = ClarityIconsApi.instance.get("error");
```
youdz: I believe you still want to put the user-given title even if you can't find the icon. That's super easy for us to do and helps mitigate the error when users can hover over the error icon to see what it was supposed to mean. Screen readers will even have the normal experience.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
+    // before this connectedCallback method called only once when the custom element is inserted into the DOM.
+    // So we could check whether the attribute values really changed or not.
+    // If not, we don't need to execute the same codes again.
     const host = this as HTMLElement;
 
     if (host.hasAttribute("shape")) {
-        generateIcon(host, host.getAttribute("shape"));
+        const shapeAttrValue = host.getAttribute("shape").split(/\s/)[0];
+
+        if (shapeAttrValue === this.currentShapeAttrVal) {
```
youdz: You inlined this from the function, but this test makes no sense for the `connectedCallback`, `this.currentShapeAttrVal` will always be undefined. It's only here for when the shape attribute changes.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
+    if (host.hasAttribute("title")) {
+        const titleAttrValue = host.getAttribute("title");
+
+        if (this.currentTitleAttrVal !== titleAttrValue) {
+            this.currentTitleAttrVal = titleAttrValue;
+        }
+
+        if (!this.currentShapeAttrVal) {
+            return;
+        }
```
youdz: Why is this here, in the middle of the title code? If the icon has a title but no shape, you're never setting its size, but if the title has no title and no shape you do set the size correctly?

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
+
+    this.injectTemplate(host);
 };
 
 ClarityIconElement.prototype.attributeChangedCallback = function(attributeName: string, oldValue: string,
                                                                  newValue: string) {
     const host = this as HTMLElement;
 
+    if (!this.initialized) {
+        this.clrIconUniqId = "_clr_icon_" + clrIconId;
```
youdz: Why is this set in the `attributeChangedCallback`? This is initialization, it clearly belongs to `connectedCallback`. You wouldn't even need the `this.initialized` property and test.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
+            host.innerHTML = ClarityIconsApi.instance.get("error");
+            return;
+        }
+    }
+    if (attributeName === "title") {
+        this.currentTitleAttrVal = newValue;
+
+        if (!this.currentShapeAttrVal) {
+            return;
+        }
```
youdz: That return I understand slightly more, but even then I don't see the point. And why is this only for the title and not the size? They should have exactly the same problems with no shape.

Just fallback "gracefully" to the error icon with the correct title and size, it's the most friendly behavior and we can have it for free.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
-        generateIcon(host, newValue);
+        this.currentShapeAttrVal = newValue.split(/\s/)[0];
+
+        if (ClarityIconsApi.instance.has(this.currentShapeAttrVal)) {
+            this.currentShapeTemplate = ClarityIconsApi.instance.get(this.currentShapeAttrVal);
+        } else {
+            console.error(`'${this.currentShapeAttrVal}' is not found in the Clarity Icons set.`);
+            host.innerHTML = ClarityIconsApi.instance.get("error");
+            return;
+        }
```
youdz: Why did you inline this but kept the external function for the icon size? You now have this code duplicated in two methods of `ClarityIconElement`, and I don't understand why you had to do it.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -55,25 +59,99 @@ const setIconSize = function(element: any, size: string) {
...
     if (attributeName === "size") {
         setIconSize(host, newValue);
     }
+
+    this.injectTemplate(host);
+};
+
+ClarityIconElement.prototype.injectTemplate = function(hostElement: any) {
+    if (this.currentTitleAttrVal) {
+        hostElement.setAttribute("aria-labelledby", this.clrIconUniqId);
```
youdz: When called from the `connectedCallback`, it sets `aria-labelledby` to undefined because the id is not ready yet. See my comment on the initialization happening in the wrong place.

src/clarity-icons/clarity-icons.spec.ts
```diff
@@ -373,15 +356,134 @@ describe("ClarityIcons", () => {
...
+
+            const clarityIcon = document.createElement("clr-icon");
+            clarityIcon.setAttribute("shape", "test-shape");
+            clarityIcon.setAttribute("title", "my-custom-title");
+
+            expect(removeWhitespace(clarityIcon.innerHTML)).toBe(removeWhitespace(testShape));
+        });
+
+        it("should not inject anything if the custom title is not given", () => {
+            const customTitle = "my-custom-title";
```
youdz: You're not using `customTitle` here. Happens to all of us with test copy-paste. 😛

src/clarity-icons/clarity-icons.spec.ts
```diff
@@ -373,15 +356,134 @@ describe("ClarityIcons", () => {
...
+
+            const clarityIcon = document.createElement("clr-icon");
+            clarityIcon.setAttribute("shape", "test-shape");
+
+            expect(clarityIcon.hasAttribute("aria-labelledby")).toBeFalsy();
+            expect(removeWhitespace(clarityIcon.innerHTML)).toBe(removeWhitespace(testShape));
+        });
+
+        it("should inject a custom title into the template if title attribute is present", () => {
+            const timeStart = performance.now();
```
youdz: Leftover performance helper.

src/clarity-icons/clarity-icons.spec.ts
```diff
@@ -373,15 +356,134 @@ describe("ClarityIcons", () => {
...
+
+            clarityIcon.setAttribute("shape", "angle");
+            const timeAngleShape = performance.now();
+            console.log("Icon load time: " + (timeAngleShape - timeStart));
+
+            clarityIcon.setAttribute("title", customTitle);
+            const timeCustomTitle = performance.now();
+            console.log("Icon load time: " + (timeCustomTitle - timeStart));
+
+            // tslint:disable-next-line:no-string-literal
```
youdz: If you want to avoid this, just declare a `ClarityIconElement` interface that extends `HTMLElement` with our few custom properties. It'll be 100 times cleaner and you can use it in the actual code too, allowing you to get proper type checking.

src/clarity-icons/clarity-icons.spec.ts
```diff
@@ -373,15 +356,134 @@ describe("ClarityIcons", () => {
...
+            const clrIconUniqId = (clarityIcon as any)["clrIconUniqId"];
+            const testShape = giveAngleShapeTitle(clrIconUniqId, customTitle);
+
+            expect(removeWhitespace(clarityIcon.innerHTML)).toBe(removeWhitespace(testShape));
+        });
+
+        it("should inject custom title if given and template doesn't contain title tag", () => {
```
youdz: I think you're missing a test to check that if there is a title in the SVG, it is properly *replaced* by the one given as attribute. You don't want 2 title tags in your SVG.

src/clarity-icons/clarity-icons-element.ts
Shijir: @youdz I've updated replaceTitleWithCustom function. This updated version doesn't rely on the regex; it doesn't remove elements such as title from user's template. It converts the template into a dom element and makes all the required changes into it before injecting it into the icon.

src/clarity-icons/clarity-icons-element.ts
Shijir: I have addressed it in the updated version of the method.

src/clarity-icons/clarity-icons-element.ts
Shijir: We set value to `this.currentShapeAttrVal` in `attributeChangedCallback()`, which gets invoked before `connectedCallback()`. So it won't always be undefined.

src/clarity-icons/clarity-icons-element.ts
Shijir: Good point! I will implement that!

src/clarity-icons/clarity-icons-element.ts
Shijir: What I figured out was that the `attributeChangedCallback()` is invoked before `connectedCallback()`. So I had to initialize the properties here with `this.initialized` guard. The `connectedCallback()` method gets called only once when it's actually inserted into the document tree. Of course, if any of the attributes are changed after the element is inserted into the dom, `attributeChangedCallback()` will be called after `connectedCallback()` as well.

src/clarity-icons/clarity-icons-element.ts
Shijir: I have now `injectErrorTemplate()` function takes care of this part of code.

src/clarity-icons/clarity-icons.spec.ts
Shijir: I intentionally left these logs to see how long it's taking to load an icon. 😛

src/clarity-icons/clarity-icons.spec.ts
Shijir: Thanks! I have added in a test with a following description: `it("should update existing title if custom title is specified")`

src/clarity-icons/clarity-icons.spec.ts
Shijir: fixed that.

src/clarity-icons/clarity-icons-element.ts
Shijir: Ideally, I needed to initialize them in the constructor. But `this` in the constructor and the ones in any of the prototype methods of `ClarityIconElement` were pointing to different object. But I figured out why that was so and made changes accordingly to fix that issue.

src/clarity-icons/clarity-icons-element.ts
Shijir: I have moved the size testing part to the area above both the shape attribute check. So we get hold of the size attribute before shape and title attributes.

As to falling back to the error icon, in cases when title attribute value happens to change right before shape attribute changes, why do we need to inject the error icon?

youdz: The order doesn't really matter, what I'm saying right now is that if I put a correct title but a wrong shape, I get an error icon _without_ my title. You should at least display the error icon with the correct title.
You noted this was a good idea and said you would do it, but I still don't see it anywhere.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -18,30 +20,45 @@ if (typeof Reflect === "object") {
...
-        }
+    if (!svgElement) {
+        return currentShapeTemplate;
     }
+
+    let titleElement = svgElement.querySelector("title");
+
+    if (titleElement) {
+        titleElement.setAttribute("id", clrIconUniqId);
+        titleElement.innerHTML = currentTitleAttrVal;
```
youdz: I think you want `textContent` here, not `innerHTML`. It's safer and you're not supporting HTML in SVG titles anyway.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -18,30 +20,45 @@ if (typeof Reflect === "object") {
...
+    } else {
+        titleElement = document.createElement("title");
+        titleElement.setAttribute("id", clrIconUniqId);
+        titleElement.innerHTML = currentTitleAttrVal;
```
youdz: Same here.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -31,21 +37,28 @@ ClarityIconElement.prototype.constructor = ClarityIconElement;
...
+        svgElement.insertAdjacentElement("afterbegin", titleElement);
     }
 
-    const openingSVGTag = hasSVGtag[0];
-    const remainingSVG = currentShapeTemplate.split(/<svg\b[^>]*>/i)[1];
-    return openingSVGTag + `<title id="${clrIconUniqId}">${currentTitleAttrVal}</title>` + remainingSVG;
+    return tempContainer.innerHTML;
```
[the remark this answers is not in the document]
mathisscott: We should not be going from `String -> Html -> String -> Html`

I think we can do what we want with basic (and _fast_!) string operations and then apply it to the DOM once.

src/clarity-icons/clarity-icons-element.ts
Shijir: I do think there are cases that users may want to manipulate `clr-icon` elements before or without injecting them into the DOM tree. In those cases, `clr-icon` element's attributes should update properly. Also, in environments with simulated or fake DOM tree, clr-icon attributes probably wouldn't work. For example, if I do:
```
let myIcon = document.createElement("clr-icon");
myIcon.setAttribute("shape", "user");
console.log(myIcon.innerHTML); // no SVG template as we have to now append to the DOM.
```

src/clarity-icons/clarity-icons-element.ts
Shijir: Wait 🤔  I've actually addressed that issue. I have even added a test for the case:
```
it("should inject error shape with title if icon is not found", ...);
```

src/clarity-icons/clarity-icons-element.ts
Shijir: This is here to escape calling `this.injectTemplate(host)` as a user hasn't yet provided the shape at this point.

src/clarity-icons/clarity-icons-element.ts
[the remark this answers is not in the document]
Shijir: @youdz @mathisscott I resorted to DOM operations as I thought this would be more robust and reliable than using regexes and string operations. When I do benchmark tests, this approach happened to show slightly better performance than the previous string manipulation approach I had. I think it doesn't perform badly because we are doing all the DOM operations **before** injecting it into the actual DOM tree. So there is no deep DOM traversal or no DOM reflow activity.  Also, in terms of readability, it looks much cleaner and easier to read IMO.

The only hesitation or question over this approach I have now is: can we accomplish better results if we find better string manipulating operations? I think better string manipulating operations means we should avoid regexes as much as possible because as soon as I use them, the benchmark tests are not showing better results, but actually worse. And manipulating string HTML template without regex to reach the desired result can be very difficult.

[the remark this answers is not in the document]
Shijir: Or we could also end up achieving better results if we use both dom operations and string manipulating operators. That's what I've been pondering, testing, and experimenting.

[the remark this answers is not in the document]
mathisscott: I would say:

• avoid regex for this

mathisscott:
@Shijir 
What is the status of this PR?

Shijir:
@mathisscott 
It's ready for reviews. I've just requested reviews from other team members. Also, just talked to @youdz about this issue. He said he would go over it as he is catching up on pull requests.

src/clarity-icons/clarity-icons.scss
```diff
@@ -95,11 +95,25 @@ clr-icon {
...
         width: 100%;
         display: block;
         transition: inherit;
+        title {
+            display: none;
+        }
+    }
+    & > span[id^=_clr_icon_] {
+        position: absolute;
+        left: -10000px;
```
adityarb88: This will cause scrollbars to appear if the display is changed from ltr to rtl. Could you reuse this class instead? https://github.com/vmware/clarity/blob/master/src/clarity-angular/popover/common/_popover.clarity.scss#L6

src/clarity-icons/clarity-icons.scss
```diff
@@ -198,7 +198,6 @@ clr-icon {
         top: -1px !important;
 
         overflow: hidden !important;
-        visibility: hidden !important;
```
adityarb88: Why are we removing `visibility: hidden`?

Shijir: The main reason was that `visibility: hidden;` makes screen readers ignore the content text. 
https://webaim.org/techniques/css/invisiblecontent/

adityarb88: Yep. Totally forgot about that. This is different than popovers.

PR #1420, opened 2017-09-05
[NG] Fixes datagrid height on page size changes

### LINKED ISSUE #1084: Datagrid dynamic pagination
```
[x] bug
[ ] feature request
[x] enhancement
```

### Expected behavior

If i set the pagination of the datagrid to show 25 records on each page and then change the number of each page to 10 the datagrid should resize itself to the appropriated height.

### Actual behavior

When i lower the number of records per page the datagrid creates a blank space of the previous size.

### Reproduction of behavior
https://plnkr.co/edit/NmkRkIdmvZaVbNbcqjMv?p=info

### Environment details

* **Angular version:** 
4.0

* **Clarity version:** 
0.9.9

hippee-lee:
- Closes #1084

### DIFF
--- a/src/clarity-angular/data/datagrid/render/main-renderer.spec.ts
+++ b/src/clarity-angular/data/datagrid/render/main-renderer.spec.ts
@@ -1,10 +1,14 @@
+
 /*
  * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component} from "@angular/core";
+import {Component, DebugElement} from "@angular/core";
+import {ComponentFixture, TestBed} from "@angular/core/testing";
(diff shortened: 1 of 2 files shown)


youdz:
Actually, I do have one question: why the update to the reference screenshots with no visible changes? I thought we were finally past that.

hippee-lee:
>  I thought we were finally past that.

You and me both.

There was a 1px shift in only those three images. It looks much like what I saw after we upgraded the angular version to 4.3.0 but before I had actually installed it. What I mean is I brought changes from master that updated package.json to angular 4.3.0 and new reference images but I hadn't installed it yet. Many tests failed much like these until I installed the angular 4.3.0 code. 

I'm also having clang issues atm but will update here once tests are happy.

hippee-lee:
In working through the clang issues I reverted the images and was able to get the regression tests to pass without updating the reference image.

Issue #1201, opened 2017-07-25
Add .btn-icon override class


jonathantrevor:
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior

Small icon-only buttons, and those with short text (e.g. 10m, 1d, 8d for time selection), especially in toolbars / trays / button groups seem far to wide. 

The current base .btn class has a min-width of 72px which is too large, especially when combined with 12px l/r padding.

I'd recommend going to a smaller min-width, say 24px, or perhaps making just .btn.sm small buttons have the smaller relative size and/or padding.

### Actual behavior

Unsuitable for trays or toolbars which have a lot of buttons, or as single dropdown "buttons" or action "icon" buttons.

### Environment details

* **Angular version:**
n/a

* **Clarity version:** 
0.9 and 0.10

* **OS and version:** 
n/a

* **Browser:** all

mathisscott:
When we have a button that only contains an icon, we should allow users to give it a `.btn-icon` class. This will adjust it so that the min-width is equal to the expected height. So that minimally, the sizing will be equilateral.

mathisscott:
Also needs a docs update on buttons page...

adityarb88:
@mathisscott, @reddolan: `This will adjust it so that the min-width is equal to the expected height. So that minimally, the sizing will be equilateral.` Are we also supporting small icon buttons? (24x24)

mathisscott:
@adityarb88 
To avoid confusion, we should support small icon buttons. Our recommendation should be a minimum click target of 36x36 but users should be able to create small icon buttons nonetheless.

PR #1422, opened 2017-09-05
Icon Buttons

adityarb88:
Fixes: #1201 

Adds Icon Buttons & Button Groups to Clarity. Screenshots attached.

Tested on:
Chrome, Firefox, Safari, IE, Edge


src/app/button-group/static/icon-buttons/icon-button-group.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
```
hippee-lee: This is a TODO for updating the dates in our file headers.

adityarb88: Yes. But I think we need to do this in one go for all files.

hippee-lee: Right, thats what I meant but it didn't come accross that way. I'll file an issue for doing it as a chore.

PR #1423, opened 2017-09-06
[CORE] Clang Format

adityarb88:

### DIFF
--- a/src/app/tree-view/recursive-selectable-tree/recursive-selectable-tree.ts
+++ b/src/app/tree-view/recursive-selectable-tree/recursive-selectable-tree.ts
@@ -92,14 +92,15 @@ export class RecursiveSelectableTreeDemo {
         name: "A1",
         selected: false,
         expanded: true,
-        children: [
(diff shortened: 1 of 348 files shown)


src/app/tree-view/recursive-selectable-tree/recursive-selectable-tree.ts
```diff
@@ -92,14 +92,15 @@ export class RecursiveSelectableTreeDemo {
         name: "A1",
         selected: false,
         expanded: true,
-        children: [
```
hippee-lee: @jeeyun - Are forms like these something we can specify with a rule? Or are they `BasedOnStyle: Google`?

jeeyun:
@adityarb88 @hippee-lee @youdz @mathisscott @gnomeontherun @Shijir can you also make sure you update your git hook to include the -style argument? something like:

`node_modules/.bin/clang-format -style=file -i "$file"`

This will ensure that the command will look for the `.clang-format` file and use it.

PR #1440, opened 2017-09-07
[ng] Reset to page 1 when filtering (#812)

### LINKED ISSUE #812: [NG] Page change when filtering too much in server-driven datagrid
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

If the user is on page 4, then changes a filter in the datagrid that reduces the available items down to 2 pages, the datagrid should end up moving from the page 4 to page 2 without issues.

### Actual behavior

This currently happens by emitting 2 state changes, one when the filter is modified, then another one when the data comes back from the server and we learn that there is no 4th page.
This second event is synchronous, and typically changes the `[clrLoading]` binding of the parent app, which results in the beloved `Expression has changed after it was checked` error.

Note that this is not such a big priority, because an easy workaround for the app is simply to wrap its `this.loading = true` (or whatever handles this logic) in a timeout: `setTimeout(() => this.loading = true)`.

### Reproduction of behavior

_I will add a plunker soon, I just do not want to forget this._

### Environment details

* **Angular version:** 4.x

* **Clarity version:** 0.9.x

* **Browser:** all

jeeyun:
Filtering may change the page number such that current page number doesn't exist in the filtered dataset. So we always set the current page to 1 when filtering, so that it'll fetch first page's data with the given filter.

Many trivial changes are introduced to tests simply because of the added provider (Page) to FiltersProvider.

Fixes #812.


src/clarity-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -107,6 +107,21 @@ export default function(): void {
                     expect(context.testComponent.nbRefreshed).toBe(1);
                 });
 
+                it("emits twice when the filters change when currentPage > 1", function() {
```
youdz: I'm surprised because I don't remember this being what we agreed on. Blocking the PR until we discuss it offline.

src/clarity-angular/data/datagrid/datagrid.ts
```diff
@@ -111,6 +111,7 @@ export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
                 state.filters.push(filter);
             }
         }
+        console.log(state);
```
youdz: You missed a `console.log()` during cleanup.

jeeyun: Thanks for catching it. Will remove.

PR #1441, opened 2017-09-08
[WIP] fix focus trap

Shijir:
This could be a potential solution to fix the issue of `clrFocusTrap` if it's used multiple times simultaneously on the same page. What this PR does is that it only enables the very last `clrFocusTrap`. If this looks like a reasonable fix to the problem, I will proceed to test this fix.


src/clarity-angular/utils/focus-trap/focus-trap-linker.service.ts
```diff
@@ -0,0 +1,13 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Injectable} from "@angular/core";
+import {FocusTrapDirective} from "./focus-trap.directive";
+
+@Injectable()
+export class FocusTrapLinker {
+    currentFocusTrap: FocusTrapDirective;
+    previousFocusTrap: FocusTrapDirective;
```
youdz: What if you have 3 modals? You support 2, but draw the line at 3? That seems arbitrary.
Please change this to a queue so that if we support nested modals, we support any number of them.

src/clarity-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -3,17 +3,32 @@
...
 @Directive({selector: "[clrFocusTrap]"})
-export class FocusTrapDirective implements AfterViewInit {
-    constructor(public elementRef: ElementRef) {}
+export class FocusTrapDirective implements AfterViewInit, OnDestroy {
+    prevFocusTrap: FocusTrapDirective;
+    focusTrapEnabled: boolean;
+
+    constructor(public elementRef: ElementRef, private focusTrapLinker: FocusTrapLinker) {
+        if (this.focusTrapLinker.currentFocusTrap && this.focusTrapLinker.currentFocusTrap.focusTrapEnabled) {
+            this.focusTrapLinker.currentFocusTrap.focusTrapEnabled = false;
```
youdz: I like the idea, but the execution here is messy. I mean, let's just phrase the current structure:
"Directives are directly setting public properties of other directives that they grab through a service to which said directives registered manually."
This is convoluted and will lead to bug, if you don't already have some. 

What you want is to move all of that logic to the service, and only do a basic registration here.
For instance, you can expose a method like this in the service:
```typescript
export class FocusTrapLinker {
    ...
    isEnabled(focusTrap: FocusTrapDirective): boolean {
        // Change this once you move the service to an array of focus traps, as mentioned in my other comment
        return this. currentFocusTrap === focusTrap;
    }
}
```

src/clarity-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -22,4 +37,11 @@ export class FocusTrapDirective implements AfterViewInit {
         const nativeElement: HTMLElement = this.elementRef.nativeElement;
         nativeElement.setAttribute("tabindex", "0");
     }
+
+    ngOnDestroy() {
+        if (this.focusTrapLinker.previousFocusTrap) {
+            this.focusTrapLinker.currentFocusTrap = this.focusTrapLinker.previousFocusTrap;
```
youdz: As I was guessing above, I see 2 bugs here:
- If I destroy a `FocusTrapDirective` that is not currently the active one, you switch from the current to the previous anyway, even though the current is still there. This will break everything if the case happens.
- If I destroy a `FocusTrapDirective` and I don't have a previous one, you never unset the `currentFocusTrap` property. That's a memory leak. Even worse, that means if I open a new focus trap later, when I destroy that new one the app will try to re-enable the old one that was destroyed long ago. Once again, potential crash or broken behavior.

src/clarity-angular/utils/focus-trap/focus-trap-linker.service.ts
Shijir: No, it can support any number of modals. FocusTrapLinker is merely playing the role of a linker; it's not responsible for the chain of references. The chain of references is saved in the `this.focusTrapLinker.previousFocusTrap`. So no limit on the number of modals.

youdz: Yes, there is a limit, you don't have a reference to the third modal anymore. Here is a step-by-step example, using your code (there are several issues in this scenario): let's say you have 3 modals A, B and C, and you open them one by one without closing the previous one.
1. Open modal A
```typescript
currentFocusTrap = A
previousFocusTrap = undefined
```
2. Open modal B
```typescript
currentFocusTrap = B
previousFocusTrap = A
```
3. Open modal C
```typescript
currentFocusTrap = C
previousFocusTrap = B
```
4. Close modal C
```typescript
currentFocusTrap = B
previousFocusTrap = B
```
(We already see something is wrong.)
5. Close modal B
```typescript
currentFocusTrap = B
previousFocusTrap = B
```
No focus trap on A, focus trap on B which has been destroyed, ... It's all broken.

Shijir: Oh lol, I just forgot to set `prevFocusTrap` in the `focus-trap.directive.ts`. If I set a value to it, I think it can support any number of modals. But, I will see and test it more than 3 nested modals.

src/clarity-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: I know this is super messy currently. I just wanted to get a feeling of your perspectives on this approach first. I will improve the overall code quality in this PR and include the tests in it.

src/clarity-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: Ah yes, I think you are taking about the case when `focusTrapEnabled` is not enabled, right? If so yeah, I know I have another thing to add there. The second one puzzles me a little bit. So I will discuss it with you on Slack.

PR #1442, opened 2017-09-08
[CORE] Lock docker image for css regression testing with a sha value

hippee-lee:
## Fix
This will hopefully fix the issues with images that have cropped up recently. By adding the sha for the same `selenium/standalone-chrome` docker image will always be used to run the browser tests. This change fixes my earlier mistake by not understanding exactly what docker was pulling for the test. 

## Enhancement
I broke the gemini tests up into four sets to speed up the overall build process. 

## Update
I updated the reference images for this specific docker image. 


src/app/vertical-nav/nested-menus/vertical-nav-nested-menus.demo.html
```diff
@@ -25,7 +25,7 @@
                 </a>
             </div>
         </header>
-        <div class="content-container">
+        <div class="content-container .nested-menus-text">
```
adityarb88: something looks incorrect here

hippee-lee: Yes, I'll fix it.

Issue #98, opened 2016-11-17
Code highlighting not applied when navigating through angular's router


amellnik:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Navigating to a component via a route should result in the same code highlighting as refreshing the page on that route.  

### Actual behavior
Code highlighting is only applied if the component is refreshed.  

### Reproduction of behavior
Load [this page](https://amellnik.github.io/clarity-demo/).  Code highlighting is applied. Next, navigate to a different route, then either hit the back button or navigate back.  Code highlighting is not correctly applied.   

### Environment details

* **Angular version:** 2.1.0

* **Clarity version:** 0.7.0

* **OS and version:** Win8.1 x64

* **Browser:** Seen in Chrome 54 and Firefox 48 & 50.  Code highlighting does not work at all in Edge, probably unrelated.

youdz:
Hi @amellnik, this is a good catch. A quick investigation reveals that when hitting back, Prism is run on the `<code>` element _before_ the example is interpolated in.

Our code highlight component was initially meant for our static website demos, and listening to changes to its inner text requires quite the refactor. I will raise this to the team and we will see how we can fix it and when.

In the meantime, we _do_ have a wrapper component that we have been using for more complex demos: [example.ts](https://github.com/vmware/clarity/blob/master/src/clarity/datagrid/demo/utils/example.ts)
It is slightly more dynamic, you don't have to worry about escaping brackets since you use interpolation (just like you've been doing in your demo app), and is slightly easier to use. That said, it still doesn't listen to example code changes, but that's as easy as using a setter on the input: 
```typescript
private _code: string;
public get code() {
    return this._code;
}
@Input("clrCode") public set code(raw: string) {
    this._code = raw;
    this.codeHighlight.redraw();
}
```

I hope this helps.

CrisLi:
@youdz I found there is a problem in the demo code (maybe a defect for CodeHighlight component). When the `@Input("clrCode") public code: string;` changed in the parent component, the code highlight block is still the old value. The component variable `code` is changed, but the rendered DOM is not changed.

youdz:
@CrisLi: Thanks for the bug report, we will look at it and probably fix it as part of a more general refactor of the code highlight component, because at the moment it really doesn't support anything dynamic.

amellnik:
[...]

PR #1447, opened 2017-09-08
[WEB] icon detail snippet fix

Shijir:
- added in more tags for searching as well

Preview: https://shijir.github.io/clarity/icons/icon-sets

### DIFF
--- a/src/app/icons/icons-sets/icon-detail-card/icon-detail-card.component.html
+++ b/src/app/icons/icons-sets/icon-detail-card/icon-detail-card.component.html
@@ -9,34 +9,25 @@
 
         <div [ngSwitch]="activeVariantClasses">
 
-            <clr-code-snippet *ngSwitchCase="'is-solid'"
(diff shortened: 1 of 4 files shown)


jeeyun:
With all the format changes, it's a bit hard to identify what's fixing the issue at hand. Can you briefly explain or just highlight the relevant code changes for me?

Shijir:
@jeeyun our `clr-code-snippet` component wasn't working as it should. When the value of its `[clrCode]` changes, it wasn't updating the view. I think the problem lies in how we used `prism.js` in the `clr-code-snippet` component. That's why I didn't use the component and resorted to using raw code snippets.

adityarb88:
@shijir refer to this issue: https://github.com/vmware/clarity/issues/98

adityarb88:
@shijir is this PR going to be merged in 0.10.5? If yes we should update the milestone so that @gnomeontherun knows to merge this after the publish.

PR #1451, opened 2017-09-11
set row actions click target to 30x30

### LINKED ISSUE #1402: [NG] Row actions for datagrid do not have the standard click target size.
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Any click/touch target in Clarity, barring UX-approved exceptions, should be at least 36x36 pixels. 

### Actual behavior

The ellipsis button to show row actions in datagrid is much smaller than the needed size. It was just added in https://github.com/vmware/clarity/pull/1395.
See the expandable row caret for an example of how to do increase the button size while not messing up the Datagrid layout.

gnomeontherun:
Set click target and padding for actions trigger to be 30x30

closes #1402

Signed-off-by: Jeremy Wilken <gnomation@gnomeontherun.com>


youdz:
Ugh, I just realized that the "good" click targets in datagrid are 30x30px. I don't remember this discussion to go with this instead of 36px.

@reddolan: Was there a "derogation" for datagrid click targets to go down to 30px because of row size constraints? I can't find the information.

mathisscott:
@youdz
Our decision with the UX of the datagrid what's that the rows would be 36px tall. At the time, this meant that the tallest icon button we could use was 30px.

The thinking behind the 30px x 30px icon buttons in the data grid was that are preferred click target of 36 x 36 was an affordance for smaller form factors like phones.

Our reasoning was that the data grid would not be used for a phone application. That didn't make sense to us. So we decided that 30 x 30 was okay.

reddolan:
Is there a reason why we can't take full advantage of the 36px row height? Could we use a 36px click target? The click target doesn't need to have a margin and being able to click anywhere in that 36x36 space would be easier than clicking just outside of the click target
image
image

gnomeontherun:
@reddolan At 36px it actually pushes the row size slightly https://raw.githubusercontent.com/gnomeontherun/clarity/a427dc50c7277475f6ed3dddb91be86c34f7dbd2/gemini/screens/datagrid/multl-selection/action-overflow/chrome.png

mathisscott:
@gnomeontherun 
We need it to maintain the existing size of the datagrid rows -- even if the icons' click targets are larger. Increasing the rows to 42px is not going to work.

reddolan:
@gnomeontherun any ideas as to why it pushes the row height?
@mathisscott agreed, I definitely don't want to increase the row height.

If the best we can do is 30x30, that is better than what exists for the current (16x17) single-row action touch target

gnomeontherun:
@reddolan The best we can do is 35px because the rows are 36px with a border (so inner is 35px). I've updated to 30x30, and this doesn't change any of the current gemini images.

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -147,6 +147,10 @@
         .datagrid-head {
             background-color: $clr-thead-bgcolor;
             border-bottom: 2*$clr-table-borderwidth solid $clr-light-midtone-gray;
+
+            .datagrid-row-actions {
+                padding: 10px 12px 11px;
```
adityarb88: Do you think we should create variables for these values with meaningful names so that other developers who work with this stylesheet change these values while maintaining the vertical rhythm?

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -518,7 +526,8 @@
                 color: $clr-dark-gray;
                 border: none;
                 background: none;
-                padding: 0;
+                padding: 8px;
```
adityarb88: Same for the ones here

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
gnomeontherun: I think that would be a larger refactor to ensure that changes to the vertical height remain consistent. Making this a variable doesn't have a lot of value IMO, because its a particular quirk to get the buttons exactly sized and positioned.

mathisscott: I think the key to creating meaningful variables in SASS is identifying what about them makes them "quirks". In this example...

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
mathisscott: If not a variable, this should probably be using our existing baselineRem variables.

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
mathisscott: Leaving comment so github will let me request changes.

src/clarity-angular/data/datagrid/_datagrid.clarity.scss
gnomeontherun: Settled on 
``` 
            .datagrid-action-toggle {
                padding: $clr_baselineRem_0_375 - $clr-table-borderwidth;
                margin: ($clr_baselineRem_0_125 - 1) $horizontalPadding;
            }```

File: src/clarity-angular/data/datagrid/_datagrid.clarity.scss (hunk 1 above)
Comment (gnomeontherun) [reply src/clarity-angular/data/datagrid/_datagrid.clarity.scss]: Replicated what I found elsewhere in cells.

```
padding: $clr-table-topcellpadding $clr-table-cellpadding $clr-table-bottomcellpadding;
```

PR #1475, opened 2017-09-15
[NG] Vertical Nav Fix

adityarb88:
Demo: http://clarity-vertical-nav.surge.sh/

1. Adds support for text on the nav group level which when clicked expands the nav group.
2. Fixes Nav Group Icon Highlight when the Nav Group is collapsed and one of the links under it is active.

Screenshots attached.

Tested on: IE11, Edge, Safari, Firefox, Chrome

**Should only be merged when the documentation is ready**


src/clarity-angular/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -11,9 +11,10 @@ $vertical-nav-top-padding: $clr_baselineRem_1;
 $vertical-nav-trigger-collapsed-margin: ($vertical-nav-min-width - $vertical-nav-item-height) / 4;
 
 //Icons
-$vertical-nav-icon-size: 16px;
+$vertical-nav-icon-size: $clr-icon-dimension-sm;
```
mathisscott: ❤️

src/clarity-angular/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -171,54 +121,95 @@ $vertical-nav-divider-margin: $clr_baselineRem_0_5;
...
+
+            //Others
+            transition: all 0.2s ease-in-out;
         }
 
-        .nav-icon + .nav-group-text {
-            padding-left: 0;
+        .nav-trigger + .nav-content {
+            //Dims
+            border-top: 1px solid rgba($gray-darker, 0.2);
```
mathisscott: Can we use `$clr-default-borderwidth` here?

src/clarity-angular/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -171,54 +121,95 @@ $vertical-nav-divider-margin: $clr_baselineRem_0_5;
...
+        .nav-group-text,
+        .nav-link {
             //Dims
+            height: $vertical-nav-item-height;
             padding: 0 $vertical-nav-link-padding-right 0 $vertical-nav-link-padding-left;
 
             //Others
-            @include clr-getTypePropertiesForDomElement(vertical_nav_header, (font-size, font-weight, letter-spacing));
             line-height: $vertical-nav-item-height;
+            outline-offset: -5px;
```
mathisscott: I've noticed this `-5px` for focus in a lot of places in our SCSS. We may want to consider making it a SCSS var at some point. Nothing for you to do here. Just want to know if you think it is a good idea.

If so, I will take care of it with the rem units refactor.

adityarb88: Yes this definitely is a good idea. We have it defined in all places which use a link in a component which has `overflow: hidden` & `overflow: auto` because in this case the component clips the outline styles. We should definitely do this later on.

src/clarity-angular/layout/vertical-nav/_vertical-nav.clarity.scss
adityarb88: @mathisscott: Sure. I can create a global variable for this. Thanks!

PR #1479, opened 2017-09-19
[NG] Internal structural directive for virtual scrolling

youdz:
See demos for code examples on how to use it.
It's still very crude and missing many customizations, but we can grow it organically as we go.

Do not merge, I still have to write unit tests. And they are going to be fun ones to write.


src/app/virtual-scroll/slots.ts
```diff
@@ -0,0 +1,39 @@
...
+export function slotGenerator(slot: number) {
+    /*
+     * Preferred solution, but doesn't work yet because Angular iterates over it.
+     */
+    // return function*() {
+    //     let index = yield images[imagesIndices[slot][0]];
+    //     while (true) {
+    //         index = yield images[imagesIndices[slot][positiveModulo(index, 7)]];
+    //     }
+    // };
```
mathisscott: Do we want to leave this comment here?

youdz: As it says 2 lines above this:
```
Preferred solution, but doesn't work yet because Angular iterates over it.
```
So if Angular offers a way to not iterate over it in dev mode, we'll want to go back to this solution. That's why I'd rather keep it in the code base, commented out.

src/clarity-angular/utils/virtual-scroll/virtual-for-of.ts
```diff
@@ -0,0 +1,139 @@
...
+enum Side {
+    START = -1,
+    END = 1
+}
+
+@Directive({selector: "[clrVirtualForOf]"})
+export class VirtualForOf<T> implements AfterViewInit {
+    constructor(private viewContainer: ViewContainerRef, private template: TemplateRef<{$implicit: T}>,
+                private renderer: Renderer2, private cdr: ChangeDetectorRef) {
+        this.containerEl = <HTMLElement>renderer.parentNode(viewContainer.element.nativeElement);
```
adityarb88: Why not just get `ElementRef` and then get the parent from there?
`this.elmentRef.nativeElement.parentElement`?

src/clarity-angular/utils/virtual-scroll/virtual-for-of.ts
```diff
@@ -0,0 +1,139 @@
...
+
+enum Side {
+    START = -1,
+    END = 1
+}
+
+@Directive({selector: "[clrVirtualForOf]"})
+export class VirtualForOf<T> implements AfterViewInit {
+    constructor(private viewContainer: ViewContainerRef, private template: TemplateRef<{$implicit: T}>,
+                private renderer: Renderer2, private cdr: ChangeDetectorRef) {
```
adityarb88: cdr is not used anywhere.

src/clarity-angular/utils/virtual-scroll/virtual-for-of.ts
youdz: Because your solution uses HTML native elements, whereas this i-one uses the Renderer. As you can see, I'm never going deeper than `elRef.nativeElement`.

It's the same difference as between `elRef.nativeElement. setAttribute("Hello", "World")` and `renderer.setAttribute(elRef.nativeElement, "Hello", "World")`. The second one will work for server-side rendering or other contexts.

src/clarity-angular/utils/virtual-scroll/virtual-for-of.ts
```diff
@@ -0,0 +1,139 @@
...
+    private containerEl: HTMLElement;
+
+    // Until we have a way to tell Angular not to iterate over the whole iterable as soon as it's passed as an input,
+    // we can't accept an NgIterable here so we have this weird Array | NonNgIterable type.
+    private _items: Array<T>|NonNgIterable<T>;
+    @Input("clrVirtualForOf")
+    set items(value: Array<T>|NonNgIterable<T>) {
+        this._items = value;
+        this.viewContainer.clear();
+        this.fill();
```
adityarb88: You asked me to remind you to add:

```
this.start = -1;
this.end = 0;
```

before

`this.fill()`.


Blocking PR to make sure that you read my comment :P

PR #1480, opened 2017-09-19
[WEB] disable focus trap in static modal and wizard demos

Shijir:
Should be merged in after PR https://github.com/vmware/clarity/pull/1441 gets incorporated into master.

### DIFF
--- a/src/app/documentation/demos/modal/modal-style-design.ts
+++ b/src/app/documentation/demos/modal/modal-style-design.ts
@@ -3,14 +3,18 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component} from "@angular/core";
+import { Component } from "@angular/core";
+import { FocusTrapTracker } from "clarity-angular/utils/focus-trap/focus-trap-tracker.service";
+import { DisableFocusTrap } from "../../utils/disable-focus-trap";
+
 
 @Component({
     selector: "clr-modal-style-design-demo",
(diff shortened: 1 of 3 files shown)


src/app/documentation/demos/modal/modal-style-design.ts
```diff
@@ -3,15 +3,26 @@
...
     host: {
         "[class.in-place-takeover]": "true"
-    }
+    },
+    providers: [{provide: FocusTrapTracker, useClass: FocusTrapServiceInterjectory}]
 })
 export class ModalStyleDesignDemo {
     public basic: boolean = true;
+
+    constructor() {}
```
mathisscott: Why do we need to call an empty constructor here?

src/app/documentation/demos/wizard/wizard-design.demo.ts
```diff
@@ -7,6 +7,15 @@
 import { Component, ViewChild } from "@angular/core";
 import { Wizard, WizardPage } from "clarity-angular";
 
+import { FocusTrapTracker } from "clarity-angular/utils/focus-trap/focus-trap-tracker.service";
+
+class FocusTrapServiceInterjectory {
```
mathisscott: Does it make sense to make `FocusTrapServiceInterjectory` a class we can import? Seems like we are already duplicating the code...

Shijir: Indeed, I will address these issues you've pointed out.

Issue #1434, opened 2017-09-07
[DOCS] Add UX recommendations for Icon Buttons


adityarb88:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior

Add UX recommendation for Icon Buttons

### Actual behavior

UX recommendation for Icon Buttons does not exist.

reddolan:
Also update the Button Group Icons section to reflect the recent changes

PR #1491, opened 2017-09-20
Icon Buttons UX Recommendations

adityarb88:
Demo: http://clarity-red-icon-buttons.surge.sh/

Fixes: #1434


src/app/documentation/demos/button-group/button-group.demo.html
```diff
@@ -227,46 +229,17 @@ <h6 class="clrweb-DoxMedia-do-dont">Don't</h6>
...
-                                </button>
-                            </div>
-                        </div>
-                    </div>
-                    <div class="clrweb-DoxMedia-text">
                         <h6 class="clrweb-DoxMedia-do-dont">Don't</h6>
                         <p>
-                            Make solid or outline buttons smaller than
-                            72px wide.
+                            Use small icon buttons in most cases. They are difficult to see and distinguish what
```
mathisscott: So should we be using the small icon buttons? The first sentence here doesn't align with the two after it.

PR #1495, opened 2017-09-21
[UI] Fix Textarea padding

adityarb88:
Before: 
image

After:
image

Tested on Chrome


rbirkgit:
This is happening again with 0.10.15. Not sure which version regressed. Maybe when the switch to rem happened.

PR #1500, opened 2017-09-22
[NG] fix bug with selection with pagination and server-driven datagrids

### LINKED ISSUE #1393: DataGrid - server side selection not working with pagination
[x ] bug
[ ] feature request
[ ] enhancement

### Expected behavior
datagrid should retain selection from previous page while paginating 

### Actual behavior
should retain selected rows of previous pages while navigating

### Reproduction of behavior
Step 1 - Select any row from page 1 of data grid
Step 2 - go to page 2
Step 3 - go to page 1 (step 1 selected row is deselected)

https://plnkr.co/edit/wnEUkRcVcFeNLdTyfrk6?p=preview

### Environment details

* **Angular version: 4.0.X

* **Clarity version:0.9.x

* **Browser: Chrome

gnomeontherun:
Since you cannot use clrDgItems with server-driven datagrids, we need to ensure the trackBy function is leveraged as part of the comparison process for determining if something is equal or not. This is due to server-driven datagrids getting new objects with each page (from an API for example), and those don't strictly equal one another. You must use trackBy for selection to display correctly, though selection does persist without it.

closes #1393

### DIFF
--- a/src/app/datagrid/datagrid.demo.html
+++ b/src/app/datagrid/datagrid.demo.html
@@ -16,11 +16,12 @@
     <li><a [routerLink]="['./filtering']">Custom filters</a></li>
     <li><a [routerLink]="['./string-filtering']">Built-in filters</a></li>
     <li><a [routerLink]="['./pagination']">Pagination</a></li>
-    <li><a [routerLink]="['./pagination-scrolling']">Pagination & Scrolling</a></li>
+    <li><a [routerLink]="['./pagination-scrolling']">Pagination &amp; Scrolling</a></li>
     <li><a [routerLink]="['./selection']">Selection</a></li>
     <li><a [routerLink]="['./selection-single']">Selection (Single)</a></li>
     <li><a [routerLink]="['./preserve-selection']">Preserve Selection</a></li>
     <li><a [routerLink]="['./server-driven']">Server-driven datagrid</a></li>
+    <li><a [routerLink]="['./selection-server']">Server-driven &amp; Selection</a></li>
     <li><a [routerLink]="['./placeholder']">Placeholder</a></li>
     <li><a [routerLink]="['./scrolling']">Scrolling</a></li>
     <li><a [routerLink]="['./column-sizing']">Column sizing</a></li>
--- a/src/app/datagrid/selection-server/selection-server.html
+++ b/src/app/datagrid/selection-server/selection-server.html
@@ -1,0 +1,51 @@
+<!--
+  ~ Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Server-driven &amp; Selection</h2>
+
+<h3>Single Selection</h3>
+
+<clr-datagrid (clrDgRefresh)="refreshSingle($event)" [clrDgLoading]="loadingSingle" [(clrDgSingleSelected)]="selected">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>First Name</clr-dg-column>
+    <clr-dg-column>Last Name</clr-dg-column>
+    <clr-dg-column>Avatar</clr-dg-column>
+
+    <clr-dg-row *ngFor="let user of usersSingle; trackBy: trackById" [clrDgItem]="user">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.first_name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.last_name}}</clr-dg-cell>
+        <clr-dg-cell><img [src]="user.avatar" width="14" /></clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>
+        {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}}
+        of {{totalSingle}} users
+        <clr-dg-pagination #pagination [clrDgTotalItems]="totalSingle" [clrDgPageSize]="perPage"></clr-dg-pagination>
+    </clr-dg-footer>
+</clr-datagrid>
+
+<h3>Multi-selection</h3>
+
+<clr-datagrid (clrDgRefresh)="refreshMulti($event)" [clrDgLoading]="loadingMulti" [(clrDgSelected)]="list">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>First Name</clr-dg-column>
+    <clr-dg-column>Last Name</clr-dg-column>
+    <clr-dg-column>Avatar</clr-dg-column>
+
+    <clr-dg-row *ngFor="let user of usersMulti; trackBy: trackById" [clrDgItem]="user">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.first_name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.last_name}}</clr-dg-cell>
+        <clr-dg-cell><img [src]="user.avatar" width="14" /></clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>
+        {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}}
+        of {{totalMulti}} users
+        <clr-dg-pagination #pagination [clrDgTotalItems]="totalMulti" [clrDgPageSize]="perPage"></clr-dg-pagination>
+    </clr-dg-footer>
+</clr-datagrid>
--- a/src/clarity-angular/data/datagrid/datagrid-items-trackby.ts
+++ b/src/clarity-angular/data/datagrid/datagrid-items-trackby.ts
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Directive, Input, Optional, TrackByFunction} from "@angular/core";
+
+import {Items} from "./providers/items";
+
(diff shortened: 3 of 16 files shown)


src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -156,9 +156,14 @@ export class Selection {
...
-            return this.currentSingle === item;
+            return (this.currentSingle) ? this._items.trackBy(0, this.currentSingle) === this._items.trackBy(0, item) :
+                                          false;
         } else if (this._selectionType === SelectionType.Multi) {
-            return this.current.indexOf(item) >= 0;
+            return this.current
+                       .filter((selected: any) => {
+                           return this._items.trackBy(0, selected) === this._items.trackBy(0, item);
+                       })
+                       .length > 0;
```
jeeyun: I have a slight preference for using `findIndex` instead of `filter`, so that the function will return upon finding a match as opposed to continue going through the rest of the array.
```
return this.current.findIndex((selected: any) => {
    return this._items.trackBy(0, selected) === this._items.trackBy(0, item);
}) >= 0;
```

src/app/datagrid/datagrid.demo.html
```diff
@@ -21,6 +21,7 @@
     <li><a [routerLink]="['./selection-single']">Selection (Single)</a></li>
     <li><a [routerLink]="['./preserve-selection']">Preserve Selection</a></li>
     <li><a [routerLink]="['./server-driven']">Server-driven datagrid</a></li>
+    <li><a [routerLink]="['./selection-server']">Server-driven & selection</a></li>
```
mathisscott: `&amp;` instead of `&`?

src/app/datagrid/selection-server/selection-server.html
```diff
@@ -0,0 +1,51 @@
+<!--
+  ~ Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Server-driven & selection</h2>
```
mathisscott: &amp; instead of &?

src/clarity-angular/data/datagrid/datagrid-items-trackby.ts
```diff
@@ -0,0 +1,25 @@
...
+} from "@angular/core";
+
+import {Items} from "./providers/items";
+
+@Directive({
+    selector: "[ngForTrackBy]",
+})
+export class DatagridItemsTrackBy {
+
+    constructor(private _items: Items) {}
```
youdz: This breaks for breaks for `ngForTrackBy` outside of a datagrid. The Items provider needs to be optional.

src/clarity-angular/data/datagrid/datagrid.ts
```diff
@@ -68,7 +68,7 @@ export class Datagrid implements AfterContentInit, AfterViewInit, OnDestroy {
     /**
      * Output emitted whenever the data needs to be refreshed, based on user action or external ones
      */
-    @Output("clrDgRefresh") public refresh = new EventEmitter<State>(false);
+    @Output("clrDgRefresh") public refresh = new EventEmitter<State>(true);
```
youdz: You can revert this now that the proper fix is in.

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -39,7 +39,7 @@ export class Selection {
             }
             let leftOver: any[];
             if (this._items.trackBy) {
-                const trackBy: TrackByFn = this._items.trackBy;
+                const trackBy: TrackByFunction<Function> = this._items.trackBy;
                 const updatedTracked: any[] = updatedItems.map((item, index) => trackBy(index, item));
                 leftOver = this.current.filter((selected, index) => {
                     return updatedTracked.indexOf(trackBy(index, selected)) > -1;
```
youdz: I didn't realize this before, but this is wrong. The index of the item in `current` has nothing to do with the index of the item in the array of items itself. So any trackBy function that actually uses the index would break.

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -156,9 +156,14 @@ export class Selection {
      */
     public isSelected(item: any): boolean {
         if (this._selectionType === SelectionType.Single) {
-            return this.currentSingle === item;
+            return (this.currentSingle)
+                ? this._items.trackBy(0, this.currentSingle) === this._items.trackBy(0, item)
```
youdz: My previous comment becomes immediately obvious here, because we don't have an index so we use `0` instead of a wrong index. But the problem is the same, neither works for index-dependent trackBy functions.

To test this, the easiest is to use a track by index: `(index: number, item: any) => index`.
I believe in your server-driven case if I use this trackBy function, select one item on the first page, then move to the second page, every item on the new page will be selected.

build/tasks/bundle.js
```diff
@@ -145,7 +145,8 @@ gulp.task("bundle:clarity:js:ng1", ["typescript:clarity"], function () {
             '@angular/platform-browser': 'node_modules/@angular/platform-browser/bundles/platform-browser.umd.js',
             '@angular/platform-browser-dynamic': 'node_modules/@angular/platform-browser-dynamic/bundles/platform-browser-dynamic.umd.js',
             '@angular/router': 'node_modules/@angular/router/bundles/router.umd.js',
-            '@angular/forms': 'node_modules/@angular/forms/bundles/forms.umd.js'
+            '@angular/forms': 'node_modules/@angular/forms/bundles/forms.umd.js',
+            '@angular/http': 'node_modules/@angular/http/bundles/http.umd.js'
```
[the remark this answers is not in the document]
gnomeontherun: I did a search for places it was bundled, and added it. Wasn't sure if there were any tasks that used it.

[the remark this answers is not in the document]
gnomeontherun: I added it here to be complete about the places it was declared, in case this was used for anything I wasn't aware of.

src/clarity-angular/data/datagrid/datagrid.ts
gnomeontherun: This has been done.

src/clarity-angular/data/datagrid/providers/selection.ts
gnomeontherun: However, logically using the index for trackby and server driven list wouldn't yield the results you want. So while it doesn't send the right index, the use of that index (regardless of the value) would not be reflective of the actual index (of a paginated server driven list). 

So what would be the 'right' index?

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -154,11 +167,33 @@ export class Selection {
...
+            }
+            return false;
+        }
+
+        // Have to lookup against the precalculated
+        const index = this.allReferences.findIndex((refItem: Reference) => {
+            return refItem.original === item;
+        });
+        if (this._selectionType === SelectionType.Single && this.allReferences[index]) {
+            return (this.currentSingle) ? this.currentSingle === this.allReferences[index].original : false;
```
youdz: So _if_ we did need to use trackBy here, wouldn't it be the opposite? Rows always have the latest reference to the item, but `this.currentSingle` is the one that might be stale.

src/clarity-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -114,6 +114,33 @@ export default function(): void {
             expect(selectionInstance.isSelected(4)).toBe(false);
         });
 
+        it("handles clrDgItems trackBy index expressions in multi selection type", function() {
+            // IMPORTANT, using clrDgItems with trackBy will use trackBy reference in multi
+            selectionInstance.selectionType = SelectionType.Multi;
+            itemsInstance.trackBy = (index, item) => index;
+            // Must set the trackBy before setting the items
+            itemsInstance.all = ["a", "b", "c", "d", "e"];
```
youdz: If you really want to test `trackBy`, you should test with several objects that are equal but the trackBy can differentiate them. Something like `["a", "a", "a", "b", "b"]`.

src/clarity-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -114,6 +114,33 @@ export default function(): void {
...
+            selectionInstance.current = [1, 3];
```
youdz: Yeah, that's what I feared. By changing the meaning of `current`, you break the behavior of our two-way binding on our current selection. People want to select the items, not their trackBys.

gnomeontherun: I thought this is what we discussed though. If you have a trackBy, the current becomes an array of the trackBy references (server side), otherwise its the exact items (client side). Looks like the rest of the comments hinge upon this point. If we need to use a different property to maintain the list references so we can track the selection and the raw items, but we'd have duplicates of the arrays for client side data grids.

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -154,11 +167,33 @@ export class Selection {
     /**
      * Checks if an item is currently selected
      */
-    public isSelected(item: any): boolean {
-        if (this._selectionType === SelectionType.Single) {
-            return this.currentSingle === item;
-        } else if (this._selectionType === SelectionType.Multi) {
-            return this.current.indexOf(item) >= 0;
+    public isSelected(item: any, rowIndex?: number): boolean {
```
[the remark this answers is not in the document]
gnomeontherun: We couldn't do simple equality checks on the selection array if we have trackBy values in the selection property. `isSelected` always sends the item, not the reference, so we have to compute its reference and compare. Server driven grids cannot compare equality otherwise.

gnomeontherun:
Going to reopen with a different PR to address pieces individually.

PR #1503, opened 2017-09-23
[NG] Properly debounce state changes for Datagrid

### LINKED ISSUE #1099: clr-datagrid should not call clr-dgRefresh when it gets Destroyed/Initialized.
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Hi, I am facing problem with clr-datagrid. It calls clrDgRefresh multiple times whenever gets destroyed. Is it expected behaviour ?

When i did some research, found that it calls clrDgRefresh for all Active Filters whenever gets destroyed.

Small Analysis: clr-datagrid calls unsubscribe when gets destroyed which in-turn calls unregisterFilters which finally calls clrDgRefresh for all active filters.

I believe clrDgRefresh should even not get called when initialize clr-datagrid. This is not directly related to this issue. But please consider this scenario as well.

### Actual behavior
clrDgRefresh gets called for all Active filters when clr-datagrid destroys.

### Reproduction of behavior
Steps to Reproduce:
1.) Create clr-datagrid(would be good if with custom filter).
2.) In UI, filter on multiple column(you just need to make multiple filters isActive() to return true)
3.) Do something that destroys clr-datagrid(may be with a button)
4.) clrDgRefresh gets called for every filter which has isActive() true.

Templates:

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

youdz:
Our previous solution actually emitted too late, which broke
server-driven datagrids that relied on the `[clrDgLoading]` input.

~Also fixes #1099 as a nice side-effect.~ My bad, it didn't.

I am not 100% happy with this pattern, having to remember to call `changeStart()` and `changeDone()` everywhere is tedious, but at the same time allows us to catch more of these cases if they come from somewhere else the the 3 main state providers. I tried using zones for this, only emitting when they were no more microtasks left, but Angular subscribes before us so change detection happens before we can emit (subscribers are called in order).

Also, it's pretty painful to add a single provider anywhere in the Datagrid right now, we have to go through dozens of spec files to update them. We need to centralize this somehow.

### DIFF
--- a/src/clarity-angular/data/datagrid/built-in/filters/datagrid-string-filter.spec.ts
+++ b/src/clarity-angular/data/datagrid/built-in/filters/datagrid-string-filter.spec.ts
@@ -12,12 +12,13 @@ import {StringFilter} from "../../interfaces/string-filter";
 import {CustomFilter} from "../../providers/custom-filter";
 import {FiltersProvider} from "../../providers/filters";
 import {Page} from "../../providers/page";
+import {StateDebouncer} from "../../providers/state-debouncer.provider";
 import {DomAdapter} from "../../render/dom-adapter";
 
 import {DatagridStringFilter} from "./datagrid-string-filter";
 import {DatagridStringFilterImpl} from "./datagrid-string-filter-impl";
 
-const PROVIDERS = [FiltersProvider, DomAdapter, Page];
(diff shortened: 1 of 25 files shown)


gnomeontherun:
It makes sense to me, though I agree its painful that changeStart/End have to be called.

PR #1509, opened 2017-09-25
[NG] webpack version of our build

jeeyun:
Hi team,
Please start reviewing my PR for moving from gulp to webpack.

What's working:
- `npm start`: running in dev mode
- `npm test`: running unit test
- `npm run build`: packaging npm candidates for `clarity-angular`, `clarity-ui`, and `clarity-icons`
- `ks-app`works now with the caveat that we now make it install the tgz file produced by our build. I couldn't get the non tzg file to install and work properly and still not sure why there's a difference in consuming the package.
- docker + gemini working
- tslint and clang format check for travis build


webpack.icons.config.js
```diff
@@ -0,0 +1,54 @@
...
+        "shapes/social-shapes.min": "./src/clarity-icons/shapes/social-shapes.ts",
+        "shapes/travel-shapes": "./src/clarity-icons/shapes/travel-shapes.ts",
+        "shapes/travel-shapes.min": "./src/clarity-icons/shapes/travel-shapes.ts",
+        "shapes/technology-shapes": "./src/clarity-icons/shapes/technology-shapes.ts",
+        "shapes/technology-shapes.min": "./src/clarity-icons/shapes/technology-shapes.ts"
+    },
+    output: {
+        path: path.resolve(__dirname, 'dist/clarity-icons'),
+        filename: '[name].js',
+        libraryTarget: 'umd'
```
Shijir: I think we need to also use `library` property to expose `ClarityIcons`.

webpack.icons.config.js
```diff
@@ -0,0 +1,54 @@
+const path = require("path");
+const {UglifyJsPlugin} = require('webpack').optimize;
+
+module.exports = {
+    entry: {
+        "clarity-icons-light.min": "./src/clarity-icons/index.ts",
+        "clarity-icons.min": "./src/clarity-icons/clarity-icons-sfx.ts",
+        "shapes/commerce-shapes": "./src/clarity-icons/shapes/commerce-shapes.ts",
```
Shijir: So for each shape set, are we publishing two `.js` files with the same content; one is just `.js` and `.min.js`?

jeeyun: The `.min.js` is the minified version of the `.js` files.

webpack.icons.config.js
jeeyun: Thanks, I'll add this.

jeeyun: Hi @Shijir did a quick test with this and seems like we actually shouldn't be adding the `library` property. Here's some relevant doc around it: https://webpack.js.org/configuration/output/#module-definition-systems

In particular:
```Note that omitting library will result in the assignment of all properties returned by the entry point be assigned directly to the root object, as documented under the object assignment section```

When we produce the package with the `library` defined, I'm seeing some errors. I suspect that this is because we expect `ClarityIcons` to be globally defined/available. The package produced without the `library` property is working as expected, in both importing through TypeScript or via inclusion of the script tags.

**Comment by chunghha** [comment]
@jeeyun, fyi https://github.com/angular/angular-cli/pull/7888

File: tests/tests.entry.ts [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+import "zone.js/dist/async-test";
+import "zone.js/dist/fake-async-test";
+import "zone.js/dist/sync-test";
+import "zone.js/dist/proxy";
+import "zone.js/dist/jasmine-patch";
+
+const browserTesting = require("@angular/platform-browser-dynamic/testing");
+const coreTesting = require("@angular/core/testing");
+// include every .spec.ts file inside src, except for those in ks-app folder
+const context = (require as any).context("../src/", true, /^((?![\\/]ks-app[\\/]).)*\.spec\.ts$/);
```
hippee-lee: Ugh. This makes me wonder if having the ks-app in here is going to be more trouble than its worth.

webpack.config.js
```diff
@@ -0,0 +1,400 @@
...
+                    }
+                    else {
+                        // Join together base-href, deploy-url and the original URL.
+                        // Also dedupe multiple slashes into single ones.
+                        return `/${baseHref}/${deployUrl}/${URL}`.replace(/\/\/+/g, '/');
+                    }
+                }
+            }),
+            autoprefixer({
+                browsers: ['last 2 versions','ie 11','> 5%','Firefox > 35','Chrome > 35'],
```
adityarb88: If we have last 2 versions, do we need Firefox > 35 and Chrome > 35?

package.json
```diff
@@ -1,97 +1,106 @@
...
+        "test": "karma start ./karma.conf.js",
+        "test:travis": "npm-run-all clang:check tslint:check test",
+        "clang:check": "./scripts/clang-check.sh",
+        "clang:format": "node_modules/.bin/clang-format -style=file -i --glob=src/**/*.ts",
+        "prebuild:angular": "cp ./npm/clarity-angular/package.json ./src/clarity-angular/; replace '@VERSION' $npm_package_version ./src/clarity-angular/package.json",
+        "build:angular": "ng-packagr -p ng-package.json",
+        "postbuild:angular": "cd dist/clarity-angular; npm pack;",
+        "build:icons": "webpack --config webpack.icons.config.js",
+        "postbuild:icons": "node ./scripts/clarity-icons-svg.js ",
+        "npm:publish": "npm publish dist/clarity-ui; npm publish dist/clarity-angular/clarity-angular-$npm_package_version.tgz; npm publish dist/clarity-icons",
```
adityarb88: we don't need prepublish now because npm run build does that. is that right?

package.json
```diff
@@ -1,97 +1,106 @@
...
+        "start": "webpack-dev-server --port=4200 --host 0.0.0.0",
+        "clean": "del ./dist",
+        "build": "npm-run-all clean build:angular webpack build:icons",
+        "pretest": "webpack",
+        "test": "karma start ./karma.conf.js",
+        "test:travis": "npm-run-all clang:check tslint:check test",
+        "clang:check": "./scripts/clang-check.sh",
+        "clang:format": "node_modules/.bin/clang-format -style=file -i --glob=src/**/*.ts",
+        "prebuild:angular": "cp ./npm/clarity-angular/package.json ./src/clarity-angular/; replace '@VERSION' $npm_package_version ./src/clarity-angular/package.json",
+        "build:angular": "ng-packagr -p ng-package.json",
```
hippee-lee: 👍🏽

package.json
```diff
@@ -1,97 +1,106 @@
...
+        "bootstrap": "4.0.0-alpha.5",
+        "core-js": "^2.4.1",
+        "font-awesome": "^4.5.0",
+        "prismjs": "^1.6.0",
+        "rxjs": "^5.4.2",
+        "web-animations-js": "^2.3.1",
+        "zone.js": "^0.8.14"
+    },
+    "devDependencies": {
+        "@angular/cli": "1.4.0",
```
hippee-lee: Do we have a good reason to lock this to version 1.4.0? If not, my preference is for `^1.4.0`

src/app/app.module.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2017 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
hippee-lee: Re-reading this it might be better to handle all the copyright dates at once. Maybe a coordinated PR in early January could take care of all of the files?

src/app/card/card-old.ts
```diff
@@ -5,10 +5,5 @@
...
 import {Component} from "@angular/core";
 
-@Component({
-    moduleId: module.id,
-    selector: "clr-card-old-demo",
-    styleUrls: ["./card.demo.css"],
-    templateUrl: "./card-old.html"
-})
+@Component({selector: "clr-card-old-demo", styleUrls: ["./card.demo.scss"], templateUrl: "./card-old.html"})
 export class CardOldDemo {}
```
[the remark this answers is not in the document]
adityarb88: @hippee-lee: Good point. This is present because we did a non breaking change for cards. There are other such demos too. We need to clean them up as the deprecation period is already done for most of the old demos. But I think we should do that in a separate PR so that we can review whats needed and whats not.

tests/tests.entry.ts
adityarb88: @jeeyun @hippee-lee: I don't clearly understand the benefits of having the ks-app and the website in the master branch.

I have always liked things not related to our main code base separate (website for eg). This way reviews are not needed for website for small changes and the person publishing the website is not blocked. Same for the ks-app.

src/app/app.module.ts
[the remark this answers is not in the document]
jeeyun: Yeah that's a good point. I think we have a mixed variety and my regex rule accepts both (2017 or 2016-2017). Can we create a separate task for this?

src/app/dropdown/dropdown-static-icon-toggle.ts
```diff
@@ -6,10 +6,8 @@
 import {Component} from "@angular/core";
 
 @Component({
-    moduleId: module.id,
     selector: "clr-dropdown-static-icon-toggle-demo",
-    // Note the .css extension here, not .scss. That's the best we can have at the moment.
-    styleUrls: ["./dropdown.demo.css"],
+    styleUrls: ["./dropdown.demo.scss"],
```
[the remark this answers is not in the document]
jeeyun: search & replace ;)

tests/tests.entry.ts
hippee-lee: @adityarb88 - I see your point. Imo - they don't belong in a separate branch either though.

adityarb88: @hippee-lee: Trying to understand why though. Both the Kitchen Sink app and the website consume the build and have nothing to do with the source code right? So if thats the case then I am not exactly sure why we need to have them in the `master` branch.

hippee-lee: Sorry my comment wasn't clear enough. When I meant is that I do agree with you, but, perhaps since they are separate projects/codebases they should have their own repo. 

Here are a couple of thoughts about how we might move forward with decoupling them

- Keep the demo app (or ks-app with proper sourcemapping back to Clarity added) - we will need something to develop / debug components against 
- Move the app for regression testing into its own repo. Regression tests should still live in Clarity but the app it tests doesn't have to. Its feasible that we could publish it to a server (or surge or etc ...) and change the tests to point the browser in selenium-standalone there
- Move the website docs into it's own repo so that it has its own history, tags

> I have always liked things not related to our main code base separate 
Me too, but I don't think they should be on a separate branch in the same repo, I think they should have their own repo.

marvinosswald:
Ideas:
- I'm not sure if its a standard thing of webpack but if not it would be nice to have a catch around the watch process like known by the angular-cli so if something happens not the whole watch process crashes.
- Running only changed testsuites or having the option to manually run a single testsuite

package.json
jeeyun: yes, that's correct.

webpack.config.js
jeeyun: i am not sure. i copied this over from our previous config. maybe let's review this after this PR?

src/app/app.module.ts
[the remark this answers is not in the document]
hippee-lee: Yes, thats what I was thinking. 
I'll open an issue for it.

package.json
jeeyun: I think the cli is sometimes a bit volatile with breaking changes so I wanted to lock it. I can change it to `^1.4.0` and we monitor closely as we go forward.

gnomeontherun: Can do `~1.4.0` to keep it in 1.4 minor release.

jeeyun:
@marvinosswald: thanks for your input. Can you elaborate on the first point? Are you saying that angular-cli has known issues that halts/messes up the watch process?

As for the unit tests, I think that's a great idea. For now, it can be done by editing the regex in the `tests.entry.ts` file but maybe that can be externalized/configurable.

src/app/progress-bars/progress-bar-inline.ts
```diff
@@ -43,7 +42,7 @@ export class ProgressBarInlineDemo implements OnInit {
 
     runProgressBar(): void {
         this.stopProgressBar();
-        this.inlineProgressTimerId = setInterval(() => {
+        this.inlineProgressTimerId = window.setInterval(() => {
```
gnomeontherun: Why add `window` here? Not that we'll ever compile this sucker on a server, but it would fail.

jeeyun: Ah, it wasn't passing the linter (or type checker?) without it. Let me see if that's still the case and find ways to make it work without the `window`.

tests/tests.entry.ts
jeeyun: my two cents:
- I prefer having the `ks-app` in master. This is nice as we don't have to check out another project or switch to any branch to test out our npm packages
- I agree with your points on that it adds additional headache for `ks-app` to live under the `src` folder though.. is it possible for the `ks-app` to live in master but outside `src` folder? Or does this violate (philosophically) the structure we want?

package.json
jeeyun: Updated to `~1.4.0`.

tests/tests.entry.ts
hippee-lee: I will investigate that after we make the switch to webpack 😀

marvinosswald:
@jeeyun no quite the contrary, a default angular-cli setup includes some comfortable features.
And what i meant is if something happens, where the current gulp watch process crashes instantly, angular-cli simply displays the error in your browser and terminal then waits till you change any sources and reloads. 

If thats working out of the box perfect, if not it would be awesome to have.

package.json
chunghha: My vote goes to pin exact version of the cli and ng-packagr as start. I've seen even patch version of the cli update sometimes gets into trouble. Updates of ng-packagr seem pretty often nowadays.

How about duplicated webpack plugins and loaders in the cli and your package.json? I am wondering which version will be picked at build if there are mismatched. Just for thoughts from here.

PR #1511, opened 2017-09-25
[NG] fix wizard issue where buttons with NgIf would throw expression has changed error

### LINKED ISSUE #1439: ExpressionChangedAfterItHasBeenCheckedError has been thrown in Wizard
```
[ x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

No ExpressionChangedAfterItHasBeenCheckedError.

### Actual behavior

It could be replicable in our demo app on `http://localhost:3000/wizard/stop-navigation`. Also, it's visible on our website, but only in the dev mode.

### Reproduction of behavior

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

gnomeontherun:
The problem was in the lifecycle of the buttons the host binding for isHidden on a button would flip between true/false, due to the change of state in the navigation service provider. This delays the evaluation of the button's host binding until later in the cycle.

closes #1439

### DIFF
--- a/src/clarity-angular/wizard/providers/button-hub.ts
+++ b/src/clarity-angular/wizard/providers/button-hub.ts
@@ -10,6 +10,8 @@ import {Subject} from "rxjs/Subject";
 
 @Injectable()
 export class ButtonHubService {
+    public buttonsReady: boolean = false;
+
     private _previousBtnClicked = new Subject<any>();
     public get previousBtnClicked(): Observable<any> {
         return this._previousBtnClicked.asObservable();
--- a/src/clarity-angular/wizard/wizard-button.spec.ts
+++ b/src/clarity-angular/wizard/wizard-button.spec.ts
@@ -23,10 +23,12 @@ import {ClrWizardModule} from "./wizard.module";
             [clrWizardButtonDisabled]="disableBtn"
             [clrWizardButtonHidden]="hideBtn"
             (clrWizardButtonClicked)="doClick($event)"
+            *ngIf="show"
         >hello {{ projector }}</clr-wizard-button>
     `
 })
 class ViewTestComponent {
+    public show: boolean = true;
     public btnType: string = "";
     public disableBtn: boolean = false;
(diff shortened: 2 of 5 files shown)


src/clarity-angular/wizard/providers/button-hub.ts
```diff
@@ -10,6 +10,14 @@ import {Subject} from "rxjs/Subject";
 
 @Injectable()
 export class ButtonHubService {
+    private _buttonsReady: boolean = false;
```
Shijir: I am curious why this property couldn't be just a public property?

gnomeontherun: It was originally an observable, but then I solved it without and left the getter setting. Will just make it public.

src/clarity-angular/wizard/wizard-button.spec.ts
```diff
@@ -909,6 +913,20 @@ export default function(): void {
...
+                    // Test with buttons ready.
+                    buttonHub.buttonsReady = true;
+                    fixture.detectChanges();
+                    expect(buttonComponent.isHidden).toBe(false, "should be the default value");
+                    // Test setting to hide
+                    myTestComponent.hideBtn = true;
+                    fixture.detectChanges();
+                    expect(buttonComponent.isHidden).toBe(true, "should now be hidden");
+                });
             });
```
jeeyun: Is there an equivalent test for `isDisabled` (i.e. testing for it in conjunction with `buttonsReady`'s value)?

mathisscott: That's a good point. It might be a good idea to test for that and would be trivial to add the test here now as opposed to later if a problem is found.

gnomeontherun: Added another test to be safe, though it doesn't suffer from being a host binding.

PR #1514, opened 2017-09-26
[NG] Fix tab issue with *ngIf directive

Shijir:

### DIFF
--- a/src/clarity-angular/layout/tabs/tab-content.ts
+++ b/src/clarity-angular/layout/tabs/tab-content.ts
@@ -3,8 +3,11 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component, Inject, Input, TemplateRef, ViewChild} from "@angular/core";
+import {Component, ElementRef, Inject, Input, OnDestroy, Renderer2, TemplateRef, ViewChild} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
+
 import {IF_ACTIVE_ID, IfActiveService} from "../../utils/conditional/if-active.service";
+
 import {AriaService} from "./aria-service";
 
 let nbTabContentComponents: number = 0;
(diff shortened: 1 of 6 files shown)


src/clarity-angular/layout/tabs/tab-content.ts
```diff
@@ -14,22 +17,27 @@ let nbTabContentComponents: number = 0;
...
 
     constructor(public ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) public id: number,
-                private ariaService: AriaService) {
+                private ariaService: AriaService, private renderer: Renderer2, private el: ElementRef) {
         if (!this.tabContentId) {
             this.tabContentId = "clr-tab-content-" + (nbTabContentComponents++);
         }
+
+        this.setAttributes();
+        this.ifActiveService.currentChange.subscribe(() => {
```
gnomeontherun: Looks like you're not adding this to the `subscriptions` array?

src/clarity-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -10,26 +10,26 @@ import {
...
 
 let nbTabLinkComponents: number = 0;
 
 @Directive({
     selector: "[clrTabLink]",
     host: {
-        "[id]": "tabLinkId",
-        "[attr.aria-selected]": "active",
-        "[attr.aria-controls]": "ariaControls",
         "role": "presentation",
```
gnomeontherun: If you've replaced some host bindings, would it be worthwhile to replace them all?

src/clarity-angular/layout/tabs/tab-content.ts
Shijir: Thanks. I have addressed the issue.

src/clarity-angular/layout/tabs/tab-link.directive.ts
Shijir: I don't see any compelling reasons to replace all of them for now. 🤔  The replaced host bindings were the ones that were holding dynamic values so I needed to use Renderer to set them in a certain lifecycle hook.

Issue #1523, opened 2017-09-28
Dropdown or any element with absolute position is getting clipped off in a tree node


whizkidwwe1217:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
[x] question
```

### Expected behavior

Hi, is there a way I can render an element above a tree node? I'm using the tree view to make a filter editor like in the DevExpress filter box and I want to display a dropdown (clr-dropdown is also getting clipped so I made a custom dropdown from scratch) but it always getting clipped in the treenode. I tried setting the z-index to higher value but to no avail.
![image](https://user-images.githubusercontent.com/1977391/30951456-aee867c8-a455-11e7-8a6e-ab64eabf2022.png)

### Actual behavior

![image](https://user-images.githubusercontent.com/1977391/30951495-e63e16e6-a455-11e7-9533-d8f1715bac08.png)

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

Here's a plunkr:
https://plnkr.co/edit/PnskAP?p=preview
### Environment details

* **Angular version:** 4.*.X

* **Clarity version:** 0.10.6

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

mathisscott:
@adityarb88 
Is this a `position: relative` issue in the treeview?

adityarb88:
Yes we have a `position: relative` in tree view to position the caret in the front. We can probably get rid of it. I will take a look at that and post an update here sometime next week.

PR #1524, opened 2017-09-28
[UI] Tree View Styles Rewrite

adityarb88:
**DO NOT MERGE**

Fixes: #1523

**Demo:** http://clarity-tree-view-test.surge.sh/treeviews

**Changes:**
1. Rewrote complete TreeView styles to get rid of `position: relative` to avoid popovers from clipping. 
2. Introduced new wrappers to ensure that the tree view looks the same as before (Had to update 1 screenshot due to display: flex)
3. Removed unnecessary nesting.

**Note:**
1. Had to add `display: flex` to the tree node content. This might break alignment for some of our users. Doesn't break anything in our demos but some issues might crop up later on. Not possible to all custom test cases. The default case that we recommend works fine.

@mathisscott, @hippee-lee: Sorry but your rebase might get a bit difficult :-(.

**Tested on:** Chrome, IE, Edge, Firefox, Safari


src/clarity-angular/data/tree-view/_tree-view.clarity.scss
```diff
@@ -14,232 +16,173 @@ $clr-tree-node-children-margin-big: $clr-tree-node-touch-target-big - $clr-tree-
...
 
-    //Tree Node Content
     .clr-treenode-content {
-
+        //Note to people reviewing: Adding display flex to the content area. You think this might break a few things
+        //on the consumer side?
         //Display
-        line-height: $touch-target;
-        padding-left: $touch-target;
+        flex: 1 1 auto;
```
hippee-lee: This is the same as `flex: auto`. It should be ok as flexbox should size the content item(s) based on their height/width.

src/clarity-angular/data/tree-view/tree-node.html
```diff
@@ -1,21 +1,27 @@
...
+    <clr-checkbox
+        class="clr-treenode-checkbox"
+        *ngIf="selectable"
+        [(ngModel)]="selected"
+        [(clrIndeterminate)]="indeterminate"></clr-checkbox>
+    <div class="clr-treenode-content">
+        <ng-content></ng-content>
+    </div>
 </div>
 <!-- FIXME: remove this string concatenation when boolean states are supported -->
```
hippee-lee: Is this referring to this? `[@childNodesState]="state"` (line number 29)

adityarb88: Yes. the comment was added to make sure we fix the animation when it allows us to handle boolean values :-)

src/clarity-angular/data/tree-view/_tree-view.clarity.scss
adityarb88: I have a habit of writing the whole thing because of IE10. IE 10 default was flex: 1 0 auto instead of flex: 1 1 auto. I know we don't officially support IE10 but I think its good to specify explicitly.

Issue #1318, opened 2017-08-17
*ngIf on clr-tab affects default tab


MRamos62786:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

I would expect the first tab that has an *ngIf that doesn't evaluate to false to be the default tab.
### Actual behavior

*ngIf affects the default tab. Though there is a work around.
### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

https://plnkr.co/edit/FpCzerJ6lqVpS9u1ObyM
Note that second tab is defaulted when run.
Work around is adding *ngIf="true" to all tabs
### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

youdz:
That's a tough one. When you use the pattern in your plunker, Angular actually instantiates the second tab first because `*ngIf` happens later in the lifecycle of your app component.

I would like whoever handles this to just try moving the "default tab picking" to `ngOnInit()` as opposed to the constructor. I think that should be enough.

mathisscott:
@youdz 
@adityarb88 tried your suggestion and it didn't work. Adjusting estimate accordingly.

Shijir:
Closed by #1514

PR #1533, opened 2017-09-29
Tabs chocolate

youdz:
#1514 introduced chocolate errors in tabs. So I reverted it and rewrote a new fix for #1318.

As now standard with chocolate errors, we let Willy Wonka and the Oompa Loompas handle it. The actual fix is 10 lines long, but I wanted to add unit tests for it and got sucked in when trying to improve the helpers slightly. So there is a bunch on non-related tests cleanups, all revolving around stronger, safer typing.

### DIFF
--- a/src/app/tabs/tabs-angular.demo.html
+++ b/src/app/tabs/tabs-angular.demo.html
@@ -7,7 +7,7 @@
 <div class="clr-example">
     <button class="btn btn-primary" (click)="inOverflow = !inOverflow">Toggle</button>
     <clr-tabs>
-        <clr-tab *ngIf="true">
+        <clr-tab>
             <button clrTabLink>Dashboard</button>
             <clr-tab-content clrTabContentId="content1" *clrIfActive>
                 <p>Content for Dashboard tab. Here is a <a href="javascript://">link</a> that can be accessed via clicking
--- a/src/clarity-angular/data/datagrid/render/header-renderer.spec.ts
+++ b/src/clarity-angular/data/datagrid/render/header-renderer.spec.ts
@@ -27,8 +27,8 @@ export default function(): void {
             context = this.create(
                 DatagridHeaderRenderer, SimpleTest,
                 [MOCK_ORGANIZER_PROVIDER, MOCK_DOM_ADAPTER_PROVIDER, Sort, FiltersProvider, Page, StateDebouncer]);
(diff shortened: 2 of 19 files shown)


src/app/tabs/tabs-angular.demo.html
```diff
@@ -7,7 +7,7 @@
 <div class="clr-example">
     <button class="btn btn-primary" (click)="inOverflow = !inOverflow">Toggle</button>
     <clr-tabs>
-        <clr-tab *ngIf="true">
+        <clr-tab>
```
Shijir: Why is `*ngIf="true"` removed though? It was added there for testing purposes.

src/clarity-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -27,8 +27,8 @@ export default function(): void {
             context = this.create(
                 DatagridHeaderRenderer, SimpleTest,
                 [MOCK_ORGANIZER_PROVIDER, MOCK_DOM_ADAPTER_PROVIDER, Sort, FiltersProvider, Page, StateDebouncer]);
-            domAdapter = context.getClarityProvider(DomAdapter);
-            organizer = context.getClarityProvider(DatagridRenderOrganizer);
+            domAdapter = <MockDomAdapter>context.getClarityProvider(DomAdapter);
```
Shijir: Are these type assertion changes in Datagrid related to Tabs?

src/app/tabs/tabs-angular.demo.html
youdz: We have several unit tests for it, so we're safe.

I can put it back if you want, but it just gives you a false sense of security which tricked me once already. I changed some code and suddenly it worked again, only for me to discover that instead of registering the tabs with `*ngIf` last, it just registered them first. So suddenly the broken behavior was with the first tab _without_ `*ngIf`, and the second tab _with_ it.

src/clarity-angular/data/datagrid/render/header-renderer.spec.ts
youdz: As I mentioned in the PR's description, a small change in the test helpers was needed for me to be pass typescript compilation on the tabs tests. And that change had a little impact on several Datagrid unit tests.

Issue #1268, opened 2017-08-08
ExpressionChangedAfterItHasBeenCheckedError in Header component


coldhamix:
**Select one ...**  (check one with "x")
```
[ x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
No errors should be produced.

### Actual behavior
I am trying to encapsulate the sidebar into its own component. When I'm adding `[clr-nav-level]` property to it, an `ExpressionChangedAfterItHasBeenCheckedError` happens in `clr-header` component. If we reverse the order of `clr-header` and `my-sidebar` so that `clr-header` appears after `my-sidebar` everything works as expected.

### Reproduction of behavior
https://plnkr.co/edit/2gDsu1AX4pU1os8rKeKQ?p=preview

### Environment details

* **Angular version:** 4.3.2

* **Clarity version: latest** 

* **OS and version: all** 

* **Browser: all**

adityarb88:
@coldhamix: Please use this as a workaround for now: https://plnkr.co/edit/KgJBKNzeg5Py7ffAqZZ7?p=preview but we will take a look and get back to you.

coldhamix:
@adityarb88 Did not work for me as I my sidebar is nested one more level deeper, but still thanks for the workaround and waiting for the issue to get resolved

mathisscott:
@adityarb88 ^

coldhamix:
@mathisscott @adityarb88 Will this at least be fixed when the Header component is ready?

coldhamix:
@mathisscott I'd like to contribute if possible

mathisscott:
Sure, @coldhamix !

With these EHCAIWC errors, much of the work is actually tracking down the value that is causing the error. Angular's error messages are usually not helpful with that.

After you find the value, it's then a question of whether we need to move the changing of the value to later in the lifecycle or if we need to use either our custom way of handling these errors or by using something like setTimeout.

Once you've identified where the error is happening, you can post back here and we can point you in the right direction if you need us to.

coldhamix:
@mathisscott I just found out where the issue happens. It's `clr-header` component in `layout/nav/header.ts` in method `initializeNavTriggers`.

```
navList.forEach(function (navLevel) {
    if (navLevel === ClrResponsiveNavCodes.NAV_LEVEL_1) {
        this.isNavLevel1OnPage = true;
    }
    else if (navLevel === ClrResponsiveNavCodes.NAV_LEVEL_2) {
[...]

PR #1536, opened 2017-10-02
[NG] Main Container + Header Oompa Loompa

adityarb88:
Fixes: #1268

clr-header - Oompa Loompa
clr-main-container - Willy Wonka

Tested on: Chrome

Also added the ks-app node-modules to the ignore sources array to avoid tslint from checking it.
cc: @hippee-lee 

### DIFF
--- a/src/clarity-angular/layout/nav/header.spec.ts
+++ b/src/clarity-angular/layout/nav/header.spec.ts
@@ -8,10 +8,11 @@ import {ComponentFixture, TestBed} from "@angular/core/testing";
 
 import {ClrIconModule} from "../../icon/icon.module";
 
+import {MainContainerWillyWonka} from "./main-container-willy-wonka";
 import {ClrNavigationModule} from "./navigation.module";
 
 @Component({
-    template: `
+    template: `        
         <clr-header class="header">
             <div class="branding">
                 <a href="#" class="nav-link">
@@ -35,7 +36,11 @@ describe("Header", () => {
     let compiled: any;
 
     beforeEach(() => {
-        TestBed.configureTestingModule({imports: [ClrNavigationModule, ClrIconModule], declarations: [TestComponent]});
+        TestBed.configureTestingModule({
+            imports: [ClrNavigationModule, ClrIconModule],
+            declarations: [TestComponent],
+            providers: [MainContainerWillyWonka]
+        });
 
         fixture = TestBed.createComponent(TestComponent);
         fixture.detectChanges();
--- a/src/clarity-angular/layout/nav/nav-detection-oompa-loompa.ts
+++ b/src/clarity-angular/layout/nav/nav-detection-oompa-loompa.ts
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {ChangeDetectorRef, Directive} from "@angular/core";
+import {OompaLoompa} from "../../utils/chocolate/oompa-loompa";
+import {ClrResponsiveNavigationService} from "./clrResponsiveNavigationService";
+import {MainContainerWillyWonka} from "./main-container-willy-wonka";
+
+@Directive({selector: "clr-header"})
+export class NavDetectionOompaLoompa extends OompaLoompa {
+    constructor(cdr: ChangeDetectorRef, willyWonka: MainContainerWillyWonka,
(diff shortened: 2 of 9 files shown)


src/clarity-angular/layout/nav/header.spec.ts
```diff
@@ -11,21 +11,23 @@ import {ClrIconModule} from "../../icon/icon.module";
...
-            </div>
-            <div class="header-nav" [clr-nav-level]="1">
-                <a class="active nav-link" href="javascript://"><span class="nav-text">Components</span></a>
-            </div>
-            <div class="header-nav" [clr-nav-level]="2">
-                <a class="active nav-link" href="javascript://"><span class="nav-text">About</span></a>
-            </div>
-        </clr-header>
+    template: `        
+        <clr-main-container>
```
adityarb88: @youdz: I had to wrap the `clr-header` with `clr-main-container` in the tests. Is there an alternative to this?

build/tasks/tslint.js
```diff
@@ -7,7 +7,7 @@
 var gulp = require('gulp');
 var tslint = require('gulp-tslint');
 
-var tsSources = ['src/**/*.ts', '!src/sample-app/aot-compiled/**/*'];
+var tsSources = ['src/**/*.ts', '!src/sample-app/aot-compiled/**/*', '!src/ks-app/node_modules/**/*'];
```
hippee-lee: Thanks for adding this.

src/clarity-angular/layout/nav/navLevelDirective.ts
```diff
@@ -3,18 +3,18 @@
...
 
 @Directive({selector: "[clr-nav-level]"})
-export class NavLevelDirective implements AfterContentInit {
+export class NavLevelDirective implements OnInit {
     @Input("clr-nav-level") _level: number;
 
     constructor(private responsiveNavService: ClrResponsiveNavigationService, private elementRef: ElementRef) {}
 
-    ngAfterContentInit() {
+    ngOnInit() {
```
Shijir: Since we are moving this block of code execution from `ngAfterContentInit` to `ngOnInit`, do we need to use oompa-loompa as well? I am assuming that the chocolate error's root cause was due to the fact that we changed certain things after the view was initialized.

src/clarity-angular/layout/nav/nav-detection-oompa-loompa.ts
```diff
@@ -0,0 +1,25 @@
...
+                private responsiveNavService: ClrResponsiveNavigationService) {
+        super(cdr, willyWonka);
+    }
+
+    // NavDetectionOompaLoompa is the addition of the nav levels
+    // Since we support 2 levels, the possibilities are 0, 1 or 3 (1 + 2)
+    get flavor() {
+        let sum: number = 0;
+        this.responsiveNavService.responsiveNavList.forEach(navLevel => sum = sum + navLevel);
+        return sum;
```
youdz: This is
```typescript
this.responsiveNavService.responsiveNavList.reduce((sum, navLevel) => sum+navLevel, 0);
```

😛

src/clarity-angular/layout/nav/navLevelDirective.ts
adityarb88: @Shijir: The error is thrown even if we move it earlier in the lifecycle. I just moved it earlier because @youdz mentioned that user inputs are available in `ngOnInit` itself so I wanted to use them as early as possible.

src/clarity-angular/layout/nav/nav-detection-oompa-loompa.ts
adityarb88: haha. thanks. will fix this

src/clarity-angular/layout/nav/navLevelDirective.ts
Shijir: I see.

adityarb88: @Shijir: Just realized that I didn't mention the cause of the issue before. The issue is happening because the header binding is being changed by the nav level directives after the header's view is already initialized.

Shijir:
@adityarb88 btw, is this using classList to add a class?
 https://github.com/vmware/clarity/pull/1536/files#diff-56a5778083a74a9c97a3c700f7590187R29

adityarb88:
@Shijir: Yes. It is using class list to add the class. The reason we do that is because the directive is the same but different levels use different class names. We could have done this with this host binding too but not sure why i did it this way :-/ :D

PR #1541, opened 2017-10-03
[UI] Fixes a11y for Signpost Triggers

### LINKED ISSUE #1274: Signposts have accessibility issues
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

- When tabbing to a signpost icon, the icon should get a focus outline

### Actual behavior
- The icon has the focus, but there is no visible indication of such.
- I am also getting a message about the icon in an internal accessibility testing tool - Minimum contrast of 2.81 with its background is not sufficient to meet WCAG AA requirements for text of size 12px and weight of 500.

### Reproduction of behavior
This can be reproduced on the Clarity web page https://vmware.github.io/clarity/documentation/signposts in the section titled "Signpost positions"

### Environment details

* **Angular version:** 4.1.3

* **Clarity version:** 0.9.13

* **OS and version:** Windows 7

* **Browser:** [all | Chrome 60 | IE 11

hippee-lee:
- Adds the focus outline to triggers when they recieve focus
- Updated reference images for these changes
- Added a gemini test for the hovered state
- Closes #1274 

## Before (Top left has focus)
screen shot 2017-10-03 at 9 21 15 am

## After (Top left has focus)
screen shot 2017-10-04 at 3 41 27 pm


gemini/tests/set3/signposts.js
```diff
@@ -9,6 +9,10 @@ gemini.suite('signposts', (child) => {
         })
         .setCaptureElements('.content-area')
```
adityarb88: I know this PR doesn't change this line but we should wrap all demos in a container like `.clr-example` and take that screenshot instead of the `.content-area`. Otherwise the gemini regression images add unnecessary height to the images because the `.content-area` is occupies the entire height thats left from the header. May be a TODO for later?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -24,8 +24,26 @@
                 height: $clr_baselineRem_1;
             }
 
-            &:hover, &.active {
+            &.active {
                 color: $clr-action-blue;
+                clr-icon {
```
adityarb88: :-( Targeting `clr-icon` directly is almost never a good idea because its possible that it might target other icons in that component but in this case I guess its needed because the icons are projected by the user. Is that right?

gemini/tests/set3/signposts.js
hippee-lee: I can make this change now. Thanks.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I showed usage of a clr-icon in my documentation on the website but we project anything they give us. 
What about targeting the first child of `.signpost` elements?

gemini/tests/set3/signposts.js
hippee-lee: I will add a TODO for later.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -24,8 +24,26 @@
...
+                    fill: $clr-action-blue;
+                }
+            }
+
+            &:hover {
```
Shijir: I don't understand why `&:hover, &.active` are separated into two different sections if they happen to have the same style.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
Shijir: We could probably attach a class to the trigger directive and target that class on `clr-icon` element, maybe?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: Right, I'll recombine them. I forget why I separated them.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I'll  take a look and see if that will work.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -12,20 +12,24 @@
...
+
+            &.active, &:hover {
+                color: $clr-action-blue;
+                background-color: transparent;
             }
 
-            &:hover, &.active {
+            &:focus {
                 color: $clr-action-blue;
+                outline: 0;
```
adityarb88: @hippee-lee: Sorry but I just remembered that UX wants our outline styles to look the same across all our components. I think you should confirm with UX on how the outline should look. If I remember correctly, outline styles should match that of regular buttons so you would have to remove `outline: 0` and the border styles below and just let the browser add the outline styles.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -12,20 +12,24 @@
...
+                border: 1px solid $clr-action-blue;
+                border-radius: $clr-default-borderradius;
```
Shijir: We have `tabindex=0` now on the trigger directive so we will have a default browser focus there. We don't need to now manipulate that default browser focus, right? So I think we don't need all these border and outline styles.

hippee-lee: Yeah, this is still in progress. I'm working through a slightly different issue. But, this code will be removed as you indicated. Thanks.

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -40,10 +43,18 @@ export class SignpostTriggerDirective implements OnDestroy {
      * @function onSignpostTriggerClick
      *
      * @description
-     * click handler for the Signpost trigger button used to hide/show SignpostContent.
+     * Enter and nnpclick handler for the Signpost trigger button used to hide/show SignpostContent.
+     *
+     * The enter HostListener handles the case when the directive is on a non-standard elemnent e.g clr-icon without
+     * a button wrapper
      */
+    @HostListener("keydown.enter", ["$event"])  // Works with the clr-icon trigger
```
Shijir: nice! Do we close a signpost on `esc` key? If no, we could tackle that later...

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -40,10 +43,18 @@ export class SignpostTriggerDirective implements OnDestroy {
...
+    @HostListener("keydown.space", ["$event"])
+    // Works with the clr-icon trigger needs preventDeault to keep page from scrolling.
     @HostListener("click", ["$event"])
-    onSignpostTriggerClick(event: Event): void {
+    onSignpostTriggerClick(event: KeyboardEvent|MouseEvent): void {
+        console.log(event);
```
Shijir: left over `console.log`?

hippee-lee: yep. 😳

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -12,19 +12,20 @@
...
         }
 
-        .signpost-action {
-            min-width: $clr_baselineRem_1;
-            margin: 0;
-            padding: 0;
+        .signpost-trigger {
             color: $clr-dark-midtone-gray;
-            outline: none;
+            border: none;
```
Shijir: hmm we need to remove this border styles too. If users want to use their own custom trigger element and if that element has a border, we are removing that here.

hippee-lee: ok. I'll remove it.

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -12,19 +12,19 @@
...
-            padding: 0;
+        .signpost-trigger {
             color: $clr-dark-midtone-gray;
-            outline: none;
 
-            clr-icon {
-                width: $clr_baselineRem_1;
-                height: $clr_baselineRem_1;
+            &.btn-icon {
+                border: none;
```
adityarb88: I am a bit confused by this. So if the trigger is a Clarity outline button, there is no border for it?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -12,19 +12,19 @@
...
+            }
+
+            &.active, &:hover {
+                color: $clr-action-blue;
+                background-color: transparent;
```
adityarb88: same for this. no background if the trigger is a primary button? I am assuming that UX mentioned that signposts are used over actionable icons only? which means a transparent button with just the icon.

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -8,7 +8,10 @@ import {Subscription} from "rxjs/Subscription";
 
 import {IfOpenService} from "../../utils/conditional/if-open.service";
 
-@Directive({selector: "[clrSignpostTrigger]"})
+@Directive({
+    selector: "[clrSignpostTrigger]",
+    host: {"[class.signpost-trigger]": "true", "[attr.tabindex]": "0", "role": "button"}
```
adityarb88: tab index is set to 0 always? Would that affect the focus order on the page?

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I'm using a btn.btn-icon which has a border. The specs do not have a border around any of the trigger examples.

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
Shijir: @adityarb88 `tabindex = 0`  obeys the sequential document tab order. So it should be fine. We have the same thing in our tooltips as well

src/clarity-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I forget what caused me to add that. I'll remove it. 
Our docs don't require SignpostTriggers to be actionable icons only. Triggers could be anything (span, label etc ...)

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
mathisscott: +1 to what @Shijir said

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -40,10 +43,16 @@ export class SignpostTriggerDirective implements OnDestroy {
      * @function onSignpostTriggerClick
      *
      * @description
-     * click handler for the Signpost trigger button used to hide/show SignpostContent.
+     * Enter key, Space key and click handlers for the Signpost trigger button are used to hide/show SignpostContent.
+     *
+     * The enter HostListener handles the case when the directive is on a non-standard element e.g clr-icon without
+     * a button wrapper and we preventDefault to keep the page from scrolling when the Space key is used.
```
adityarb88: I dont think we can do this. If users use it without the button wrapper then its a wrong way of using Signposts. Listening to the space key instead and changing the default behavior might cause unexpected behavior if the user is expecting to scroll the page using a spacebar

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -40,10 +43,16 @@ export class SignpostTriggerDirective implements OnDestroy {
...
      */
+    @HostListener("keydown.enter", ["$event"])
+    @HostListener("keydown.space", ["$event"])
     @HostListener("click", ["$event"])
-    onSignpostTriggerClick(event: Event): void {
+    onSignpostTriggerClick(event: KeyboardEvent|MouseEvent): void {
         this.ifOpenService.toggleWithEvent(event);
+        event.preventDefault();  // Keeps the page from scrolling down when the signpost is triggered with space
```
adityarb88: i remember @youdz mentioning that we cannot do this. Could you please check with him once?

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
adityarb88: 👍

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
```diff
@@ -40,10 +43,16 @@ export class SignpostTriggerDirective implements OnDestroy {
...
+     * Enter key, Space key and click handlers for the Signpost trigger button are used to hide/show SignpostContent.
+     *
+     * The enter HostListener handles the case when the directive is on a non-standard element e.g clr-icon without
+     * a button wrapper and we preventDefault to keep the page from scrolling when the Space key is used.
      */
+    @HostListener("keydown.enter", ["$event"])
+    @HostListener("keydown.space", ["$event"])
     @HostListener("click", ["$event"])
-    onSignpostTriggerClick(event: Event): void {
+    onSignpostTriggerClick(event: KeyboardEvent|MouseEvent): void {
```
adityarb88: More specific types! Nice :D

src/clarity-angular/popover/signpost/signpost-trigger.directive.ts
Shijir: Yeah, I was about to mention that `preventDefault` should be called only when the trigger is activated by clicking the space key. But now I think it shouldn't be called at all. Otherwise, if users use a checkbox or radio button as a signpost trigger, they wouldn't be checked or unchecked.

hippee-lee: Ok. I'll revisit adding support for Enter/Space keys. Maybe I should limit it to Enter for now?

hippee-lee:
Closing this for now based on different changes needed to address the issue.

PR #1549, opened 2017-10-05
0.10.8 Alpha Release Webpack Build Fixes

adityarb88:
Tested on Website Branch


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "clarity-all",
-    "version": "0.10.6",
```
Shijir: Why It's versioned from `0.10.6`?

adityarb88: Its not. Thats the deleted line :-)

Shijir: I know but why it's not 0.10.7? Was your branch not rebased?

adityarb88: Oh I checked with @jeeyun. It was actually an error in the webpack PR. The package.json file was rebased incorrectly.

PR #1556, opened 2017-10-06
[core] clean up clarity-icons package

jeeyun:
clean up clarity-icons package to produce: 
- all-shapes.zip

- correct file structure for zip files 

- add missing js files


scripts/clarity-icons-svg.js
```diff
@@ -12,7 +12,8 @@ const SHAPE_SETS = [
 ];
 
 writeSVGIcons(SHAPE_SETS, () => {
+    shell.exec("cd dist/clarity-icons/shapes; zip -r all-shapes.zip ./**/*");
     SHAPE_SETS.forEach((setName) => {
-        shell.exec(`zip -r dist/clarity-icons/shapes/svg-source/${setName}.zip dist/clarity-icons/shapes/svg-source/${setName}`);
```
adityarb88: So we getting rid of the `svg-source` folder altogether?

scripts/clarity-icons-svg.js
```diff
@@ -12,7 +12,8 @@ const SHAPE_SETS = [
...
+        shell.exec(`cd dist/clarity-icons/shapes; zip -r ${setName}.zip ./${setName}/*; rm -r ./${setName}`);
```
adityarb88: This looks like its not going in the `svg-source` folder.

scripts/clarity-icons-svg.js
jeeyun: correct. that wasn't there in the earlier version. meant to be a 'holding place' that gulp task used to use but it's not in the npm package.

PR #1601, opened 2017-10-17
[CORE] Place .jshintrc to root folder

chunghha:
This is simply not to display ES6 syntax warnings.


webpack.config.js
```diff
@@ -328,10 +328,11 @@ module.exports = {
         "vendor"
       ],
       "minChunks": (module) => {
-                return module.resource
```
hippee-lee: It looks like these are formatting changes not related to the file above. Can you please revert them?

chunghha: Without this change, we see these warnings when jshint runs. That's why I changed. Do you still want to revert?

webpack.config.js: line 332, col 21, Misleading line break before '&&'; readers may interpret this as an expression boundary.
webpack.config.js: line 333, col 25, Misleading line break before '||'; readers may interpret this as an expression boundary.
webpack.config.js: line 334, col 25, Misleading line break before '||'; readers may interpret this as an expression boundary.

hippee-lee: I see, yeah this looks good since jshint is also deprecating the `laxbreak` and `laxcomma` settings.

hippee-lee:
Thanks for submitting the PR. We discussed adding `jshint` to the project and decided not to include it at this time.

Issue #1512, opened 2017-09-26
DataGrid cells not aligned properly


koldaman:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Data in DataGrid should be aligned one below the other

### Actual behavior
When using clr-dg-pagination and you switch to other page, data in clr-datagrid moves to the right (more, if you have many colomns and the shift is more aparent in the latest columns). Header then does not align with the data in clr-dg-cell. Problem is observable in Google Chrome, IE behaves well. 

### Reproduction of behavior
It is observable even in your examples: (https://vmware.github.io/clarity/documentation/datagrid/pagination)

### Environment details

* **Angular version:** 4.3.0

* **Clarity version:** 0.10.1 (also tried latest 0.10.6)

* **OS and version:** 

* **Browser:** [ Chrome 60.0 ]

adityarb88:
@koldaman: Sorry but I am not able to understand the issue correctly. Could you please provide more details (Plnkr or a Gif)? On our docs pagination page, I cant see what the issue is: https://vmware.github.io/clarity/documentation/datagrid/pagination

Here is the GIF of the datagrid in Chrome and I dont see any issues with the alignment:

![nimbus-record-video-2017-09-26-09-19-42](https://user-images.githubusercontent.com/1426805/30862554-721ef128-a29c-11e7-816c-702cd4bc939c.gif)

koldaman:
Here is my behaviour:
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/22653353/30900459-64564136-a364-11e7-80d7-78ffb2409887.gif)

mathisscott:
This appears to be happening only on Chrome 60 and only when scrollbars are visible.

jayrungta:
Can confirm, this is happening to me too. Pagination messes up the datagrid cell widths. The headers and the normal cells have different widths. Is there any plan to address this issue?

jayrungta:
Just to add, this looks especially bad if you add borders to cells:

Before:
![1](https://user-images.githubusercontent.com/16365416/31197020-ba99761c-a91d-11e7-9c1a-d666ecd6fbb7.PNG)

After:
![2](https://user-images.githubusercontent.com/16365416/31197030-c4e6848e-a91d-11e7-8474-29ad4bb6bdb8.PNG)

[...]

PR #1602, opened 2017-10-17
[NG] Fix datagrid alignment bug

adityarb88:
Fixes: #1512

Tested on Chrome


src/clarity-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -137,7 +137,22 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewChecked,
...
+            // 1. The array which contains the items displayed is updated to contain the items on the new page.
+            // 2. An event is emitted which is subscribed to by the main renderer (this file) and this marks the
+            // shouldStabilizeColumns flag to true
+            // 3. While this is happening the datagrid is in the process of cleaning up the view. The view first
+            // renders the new displayed items and then cleans up the old items. But there is a point where the view
+            // contains the old items as well as the new items. So if the page size is 10 the view contains 20 items.
+            // This causes the datagrid body to overflow.
+            // Now since shouldStabilizeColumns was set to true, the scrollbar width is calculated
+            // and added to the datagrid header. Adding the setTimeout gives Angular time to clean up the view so that
+            // the scrollbar disappears.
```
youdz: Can you just mention the fact that this only needs a timeout because of https://github.com/angular/angular/issues/19094?

That way we can remove it as soon as the issue is fixed on the Angular side.

adityarb88: Done

PR #1608, opened 2017-10-18
[NG] update dependency ng-packagr to ^1.5.0

dherges:
New flagship feature of ng-packagr are secondary entrypoints
- https://github.com/dherges/ng-packagr#secondary-entry-points
- https://github.com/dherges/ng-packagr/pull/184

Do you plan to follow the [Angular Material style of splitting up the library into sub-packages](https://github.com/angular/material2/blob/master/CHANGELOG.md#breaking-changes-2)?

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
         "karma-webpack": "^2.0.4",
         "lite-server": "^2.3.0",
         "loader-utils": "^1.1.0",
-        "ng-packagr": "^1.2.0",
+        "ng-packagr": "^1.5.0",
         "npm-run-all": "^4.1.1",
         "optimize-css-assets-webpack-plugin": "^3.2.0",
         "postcss-loader": "^1.3.3",


gnomeontherun:
Hey @dherges we do not currently plan to split the Angular library into separate packages, though should that become an Angular recommendation then we'd certainly consider it. We also don't really expose any testing stubs, but if we ddi that might be an ideal use case. If your application can properly tree shake, there is less incentive IMO to make the consumption model of the library more complex through many packages. For the time being, our components are not truly isolated because they depend on clarity-ui to get much of their styling. We'll take a closer look at this version bump, and do you have any other expected changes (excluding bugfixes) before the next release?

We do need you to fix up your PR by adding a sign-off to your commit message for CLA compliance, if you don't mind!

dherges:
From 1.4.1 to 1.5.x:

 - bumped to @angular/tsc-wrapped ^4.4.5, solves dherges/ng-packagr#101 (external dependencies, doesn't affect clarity-angular)
 - minified UMD bundles

https://github.com/dherges/ng-packagr/blob/master/CHANGELOG.md#150-rc1-2017-10-23

Something I am thinking about and would appreciate your team's ideas about it, is whether the minimalistic configuration should be stored in `package.json` (and then maybe deprecate `ng-package.json`), see dherges/ng-packagr/issues#190.

gnomeontherun:
It looks fine to me, just asking for a quick sanity check from @hippee-lee or @jeeyun

adityarb88:
@jeeyun could you please review this?

dherges:
Hi all, I published ng-packagr@1.5.0. It seems to work with Angular 5, here is [a demo repo for ng-packagr and Angular 5](https://github.com/dherges/nx-packaged)

The changelog for the [collected changes from 1.4.1 to 1.5.0](https://github.com/dherges/ng-packagr/blob/master/CHANGELOG.md#150-2017-11-05)

Issue #1618, opened 2017-10-19
If clr-tooltip-content is nested inside of ng-container, there is an error


aaronfrost:
**Select one ...**  (check one with "x")
```
[X] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

when I wrap `ng-container` around `clr-tooltip-content` _(cause sometimes I optionally want to have the tooltip and othertimes not, so I use the `ng-container` to ngIf if I want to show it or not)_ it shouldn't throw an error. It should be fine, and not care if I wrapped it with a span or an ng-container. 

### Actual behavior

when I wrap `ng-container` around `clr-tooltip-content` _(cause sometimes I optionally want to have the tooltip and othertimes not, so I use the `ng-container` to ngIf if I want to show it or not)_ then the UI breaks when the tooltip is displayed. If I use any other wrapper besides ng-container, it works fine. It is just the ng-container that dies. This isn't a major bug, but cost me a bit of time trying to figure out. My current workaround is to use a div instead of ng-container. 

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:
https://plnkr.co/edit/oOKYBeZXoLB5ptWS3l6j?p=preview

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 0.10.8

* **OS and version:** OSX 10.12.6

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
Thanks again @aaronfrost. Will take a look and get back to you.

youdz:
Yes, we are just grabbing the parent host `ElementRef` and using it to position the tooltip, but adding an extra `ng-container` around the content means we get an undefined element. So we just need to get the `Tooltip` itself instead of the parent host, and use this to position.

Thanks for filing this. 👍

PR #1620, opened 2017-10-20
[NG] ng container bug fix for popovers

adityarb88:
Fixes: #1618

### DIFF
--- a/src/clarity-angular/popover/dropdown/dropdown-menu.ts
+++ b/src/clarity-angular/popover/dropdown/dropdown-menu.ts
@@ -3,9 +3,10 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component, ElementRef, Injector, Input, Optional, SkipSelf} from "@angular/core";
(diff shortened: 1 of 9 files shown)


src/clarity-angular/popover/dropdown/dropdown-menu.ts
```diff
@@ -17,7 +18,8 @@ import {Point} from "../common/popover";
     }
 })
 export class DropdownMenu extends AbstractPopover {
-    constructor(injector: Injector, @SkipSelf() parentHost: ElementRef, @Optional() @SkipSelf() nested: DropdownMenu) {
+    constructor(injector: Injector, @Inject(POPOVER_HOST_ANCHOR) parentHost: ElementRef,
```
youdz: Do we want to do like Willy Wonka, make this `@Optional` and throw our own error if it's not found?

adityarb88: We could. Do you want me to add that change with this PR itself or create a new one for it? I think I should create a new one because you already reviewed this one and you would have to review it again.

Issue #1551, opened 2017-10-06
"[clrWizardButtonDisabled]=true" is not working for clarity wizard buttons


vishwapratapsingh:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Clarity Wizard button should be disabled when "[clrWizardButtonDisabled]=true" is added to it for below provided code:

<clr-wizard-button [type]="'finish'" (click)="save()" [clrWizardButtonDisabled]="disableSubmitBtn">Submit</clr-wizard-button>

### Actual behavior

The disabled class is added to clr-wizard-button tag of the wizard button but still the button is clickable i.e. "disabled" attribute is not added to the relative wizard button

### Reproduction of behavior
None.

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 0.10.3

* **OS and version:** Windows 10 (64 bit)

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

mathisscott:
Thanks, @vishwapratapsingh 

I've verified that the `disabled` attribute is not being set. I'll mark this as a bug.

vishwapratapsingh:
Thanks, @mathisscott

dorthrithil:
`[clrWizardButtonDisabled]` stopped working for me after I upgraded from 0.10.4 to 0.10.8. So the bug was temporarily fixed in 0.10.4 and then reintroduced? Maybe this helps you tracking it down.

PR #1621, opened 2017-10-20
[NG] Set disabled attribute when clrWizardButtonDisabled is true

bdunkz:
Tested in Chrome, FF, IE11, and Edge.

Fixes: #1551

Signed-off-by: Brian Duncan <bduncan@vmware.com>


gnomeontherun:
@bdunkz looks like you need to check the code formatting, can you fix that up?

bdunkz:
@gnomeontherun - I have fixed the formatting.

PR #1626, opened 2017-10-20
[NG] should select/deselect items only on the current page in Datagri…

Shijir:
…d multiselect

### DIFF
--- a/src/clarity-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clarity-angular/data/datagrid/providers/selection.spec.ts
@@ -57,7 +57,7 @@ export default function(): void {
             expect(selectionInstance.current).toEqual([4, 2]);
         });
 
-        it("can select/deselect all items at once in multi selection type", function() {
+        it("can select/deselect all items at once in multi selection type if no pagination exist", function() {
             selectionInstance.selectionType = SelectionType.Multi;
             selectionInstance.toggleAll();
             expect(selectionInstance.current).toEqual(itemsInstance.displayed);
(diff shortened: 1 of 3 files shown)


src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -85,6 +86,7 @@ export class Selection {
     private get _selectable(): boolean {
         return (this._selectionType === SelectionType.Multi) || (this._selectionType === SelectionType.Single);
     }
+
```
youdz: Why do we have formatting changes even with clang-format? Is it still properly working?

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -215,11 +221,14 @@ export class Selection {
...
-         * If everything is already selected, we clear.
-         * If at least one row isn't selected, we select everything.
+         * If every currently displayed item is already selected, we clear them.
+         * If at least one item isn't selected, we select every currently displayed item.
          */
         if (this.isAllSelected()) {
-            this.current.length = 0;
+            const startRemovingAt: number = this.current.indexOf(this._items.displayed[0]);
+            if (startRemovingAt >= 0) {
+                this.current.splice(startRemovingAt, this._items.displayed.length);
```
youdz: This is wrong, `current` contains items in the order the user selected them, if I remember correctly. So if I go to page 2, select an item, go back to page 1, select another item, then go to page 2 again and select all then select none, you'll end up with the wrong selection.

See my comment on the unit test regarding how to catch that.

src/clarity-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -57,18 +57,31 @@ export default function(): void {
...
             expect(selectionInstance.current).toEqual([4]);
             selectionInstance.toggleAll();
             expect(selectionInstance.current.sort(numberSort)).toEqual(itemsInstance.displayed);
         });
 
+        it("can select/deselect items only on the current page", function() {
+            selectionInstance.selectionType = SelectionType.Multi;
+            pageInstance.size = 3;
+            pageInstance.current = 1;
+            selectionInstance.current = [1, 2];
```
youdz: Can you make the current selection `[4, 1, 2]`? That can happen if the user first selects an item on page 2, then goes back to page 1.

src/clarity-angular/data/datagrid/providers/selection.ts
hippee-lee: @Shijir - are you running the pre-commit hook or `npm run clang:format` before you commit changes? 
This is the pre-commit hook I'm currently using: https://gist.github.com/hippee-lee/5d3e83f02d90224abb5de9e50893deca

Shijir: Sorry about this, I usually add an empty line after methods to make it more readable. I do use that `clang:format` on every commit, but looks like it doens't remove them. I have manually removed these empty lines.

hippee-lee: Thanks for clarifying. The default is for clang:format to remove them but it seems something changed with the change to webpack. I'll investigate further.

Issue #1232, opened 2017-07-31
accessibility + visibility with clr-dropdowns and tooltips


envyN:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
If clr-dropdown is given top-left alignment, it should not take up the default space it currently takes

### Actual behavior
https://plnkr.co/edit/jArAZjF483ejoqnkpi1J?p=preview demonstrates the bug wherein an overflow scrollbar is visible even though there seems to be no element which needs the scrollbar to appear.
as soon as all action 2 dropdowns in the last row are clicked twice, the dropdown menu is repositioned correctly and the unnecessary page scrollbar disappears.
Even if the user doesnt click any of the action buttons in the last row, the dropdowns should correctly occupy their places.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:
[demo plunk](https://plnkr.co/edit/jArAZjF483ejoqnkpi1J?p=preview)
you may need to adjust the number of cards depending on your screen size.
have as many cards in the screen which can just be accomodated in the given space. you would still see a scrollbar.
then click on the action 2 buttons in the last row. the scroll bar would disappear.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@envyN: Thanks for raising this. We will take a look and get back to you.

hippee-lee:
This may be a bug related to the new popover implementation on `Dropdown` as `clr-dropdown-menu` element is always in the DOM even though we _seem_ to be using the new Popover with `*clrIfOpen`.

adityarb88:
@envyN: Which version of Clarity are you using?

envyN:
@adityarb88 its 0.10.0-alpha.6 but seen in alpha4 as well. upgraded today to 6 and still see the issue

adityarb88:
[...]

PR #1646, opened 2017-10-26
Popover Horizontal Scroll Fix

adityarb88:
Fixes: #1232 

Deleted deprecate popover directive (Separate commit just in case we want to revert the change)

Tested on: Chrome, Firefox, Safari, Edge, IE11

Dropdowns:
screen shot 2017-10-26 at 10 05 28 am
screen shot 2017-10-26 at 10 05 43 am

Tooltips:
screen shot 2017-10-26 at 10 06 01 am
screen shot 2017-10-26 at 10 06 08 am


src/clarity-angular/popover/common/popover.ts
```diff
@@ -45,7 +45,7 @@ export class Popover {
 
     constructor(private element: any) {
         // Browsers don't agree with what to do if some of these are not specified, so we set them all to be safe.
-        element.style.position = POSITION_ABSOLUTE;
+        element.style.position = POSITION_FIXED;
```
youdz: Wait, when the popover is open it needs `position: absolute`. I was expecting the `position: fixed` to be used only when the popover is closed (which you already do with the CSS class, so I don't see why you needed to change it here too).

src/clarity-angular/popover/common/_popover.clarity.scss
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('popover.clarity') {
+    .is-off-screen {
+        position: fixed !important;
+        height: 1px !important;
+        width: 1px !important;
+        top: -1px !important;
```
youdz: I think you'd be safer with: `left: 0px !important`
Otherwise browsers might put that 1px closed popover somewhere weird by default, and might still generate scrollbars.

src/clarity-angular/popover/common/_popover.clarity.scss
```diff
@@ -0,0 +1,15 @@
...
+        padding: 0 !important;
+        margin: 0 0 -1px 0 !important;
+        border: none !important;
```
youdz: `box-shadow`?

src/clarity-angular/popover/common/_popover.clarity.scss
```diff
@@ -0,0 +1,15 @@
...
+    }
```
youdz: Hum, you forgot `overflow: hidden`. Maybe it's applied somewhere else too, but it's so critical to this off-screen class that it absolutely needs to be here too.

src/clarity-angular/popover/common/abstract-popover.ts
```diff
@@ -88,6 +88,11 @@ export abstract class AbstractPopover implements AfterViewChecked, OnDestroy {
         return this.ifOpenService.open ? "visible" : "hidden";
```
youdz: You should just  put the `visibility: hidden` as part of the new CSS class, and drop this extra binding.

src/clarity-angular/popover/common/_popover.clarity.scss
adityarb88: 👍

src/clarity-angular/popover/common/_popover.clarity.scss
adityarb88: i only added properties which occupy space. thats why ignored it. do you think we need to add `box-shadow`?

src/clarity-angular/popover/common/_popover.clarity.scss
adityarb88: 👍

src/clarity-angular/popover/common/abstract-popover.ts
adityarb88: good idea. thanks

src/clarity-angular/popover/common/popover.ts
adityarb88: 👍

src/clarity-angular/popover/common/_popover.clarity.scss
youdz: My bad, I forgot that it was `visibility: hidden` so box-shadow wouldn't show. I was scared of it showing at the top of the page, not the scrollbars.

Issue #1622, opened 2017-10-20
A modal window takes place when it is in a hidden state


belyan:
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
A modal window does not create extra padding when it is in a hidden state.

### Actual behavior
A modal window takes place when it is in a hidden state.

### Reproduction of behavior
https://embed.plnkr.co/UC5LA68PKfVMjoUwtqss/

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** latest

* **OS and version:** Windows 10

* **Browser:** Chrome 61.0.3163.100

adityarb88:
@belyan: Not sure if I understand the question correctly.

![image](https://user-images.githubusercontent.com/1426805/31827455-cb9cfbf4-b585-11e7-98ba-ed710bb0d6f0.png)

Clarity buttons have margin around them. Are you talking about that? I don't believe modal takes any space because it is `position: fixed`

belyan:
@adityarb88 I cleared margins for paragraphs in CSS. Cause is in the custom element `clr-modal`. It has the style `display: inline-block` by default, therefore it creates extra padding between paragraphs.

adityarb88:
@belyan: I am still not sure what the issue is. Are you looking for this? If yes, you need to clear the margin for the button.

![image](https://user-images.githubusercontent.com/1426805/31830479-509aff4c-b58e-11e7-9db8-e51180bd3a07.png)


https://plnkr.co/edit/GxHX55zxg4TsUS7NTL5C?p=preview

belyan:
@adityarb88 Could you please add the CSS rule `clr-modal { display: block; }`, and you will see difference.

adityarb88:
@belyan: I understand your question now. I will take a look and get back to you.

belyan:
@adityarb88 Thanks.

mathisscott:
`clr-modal` needs a `display: none` on the host when it is not open. This should fix this extra space that is added.

govi2010:
Hello Team , I am working on this right now Will send PR soon @mathisscott .

adityarb88:
Thanks @govi2010

govi2010:
hello @mathisscott , I have made changes which are require. now i am doing `npm build`, then `npm start`. I noticed that my changes are not being applied. Sorry to ask but I am new to this project. am i doing something wrong?

govi2010:
@adityarb88  Sending PR.

[...]

PR #1648, opened 2017-10-27
Fix: A modal window takes place when it is in a hidden state #1622 

govi2010:
This file changes is to fixed white space taken by model (clr-modal) when it is hidden.

So when model is hidden, I have applied display property none and when model is open i have added display property inline.

### DIFF
--- a/src/clarity-angular/wizard/_wizard.clarity.scss
+++ b/src/clarity-angular/wizard/_wizard.clarity.scss
@@ -354,7 +354,7 @@ $clr-wizard-box-shadow: 0 $clr-rem-1px (2 * $clr-rem-1px) (2 * $clr-rem-1px) rgb
             }
         }
 
-        .spinner:not(.spinner-inline) {
+        .spinner:not(.spinner-inline):not(.clr-treenode-spinner) {
             left: calc(50% + 4.791667rem);
             position: absolute;
             top: 40%;


src/clarity-angular/modal/modal.ts
```diff
@@ -26,7 +27,7 @@ import {GHOST_PAGE_ANIMATION} from "./utils/ghost-page-animations";
     viewProviders: [ScrollingService],
     templateUrl: "./modal.html",
     styles: [`
-        :host { display: inline-block; }
+        :host { display: inline; }
```
adityarb88: Why was this changed to `inline` from `inline-block`? I thought we just had to add `display: none` when the modal isn't open. I think thats what @mathisscott mentioned here: https://github.com/vmware/clarity/issues/1622#issuecomment-339349623

src/clarity-angular/modal/modal.ts
```diff
@@ -106,8 +107,16 @@ export class Modal implements OnChanges, OnDestroy {
     ngOnChanges(changes: {[propName: string]: SimpleChange}): void {
         if (!this.bypassScrollService && changes && changes.hasOwnProperty("_open")) {
             if (changes._open.currentValue) {
+                const hostElement = this.elementRef.nativeElement;
```
adityarb88: Instead of getting the `ElementRef`, can we not just use a class which is applied to the host when the modal is not open? The class just has `display: none` on it and is removed when the modal is open.

src/clarity-angular/modal/modal.ts
govi2010: @adityarb88  because when Popup is open. Due to `inline-block` margin can bee seen in background.

Like When you open modal  in backdrop you can see margin again. So I changed `inline-block ` to `inline`

src/clarity-angular/modal/modal.ts
govi2010: I have removed `ElementRef ` and used  `@HostBinding("class.open")` so open class is added and added `:host { display: none; }  :host.open { display: inline; }` css.

src/clarity-angular/modal/modal.ts
```diff
@@ -92,7 +94,7 @@ export class Modal implements OnChanges, OnDestroy {
     @Input("clrModalPreventClose") stopClose: boolean = false;
     @Output("clrModalAlternateClose") altClose: EventEmitter<boolean> = new EventEmitter<boolean>(false);
 
-    constructor(private _scrollingService: ScrollingService) {}
+    constructor(private _scrollingService: ScrollingService, private elementRef: ElementRef) {}
```
adityarb88: Thanks for making the changes. I believe we can remove the ElementRef now since its not being used.

adityarb88:
@govi2010: Could you please squash your commits? I see 4 commits now but to merge this change in we need 1 commit only.

src/clarity-angular/modal/modal.ts
```diff
@@ -6,7 +6,9 @@
 import {animate, AnimationEvent, state, style, transition, trigger} from "@angular/animations";
 import {
     Component,
+    ElementRef,
```
adityarb88: We can get rid of this import too.

govi2010: Yes sure.

adityarb88:
@govi2010: Wanted to check if you could squash to commits so that we can merge this into master? Please let us know.

govi2010:
@adityarb88  bit new to github. So i have messed up with some commits.  What changes you need from me ? I can do that right a way.

adityarb88:
@govi2010: No worries. You can refer to this: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit for more information. Let me know if that helps.

adityarb88:
@govi2010 You could also submit a new PR if you are facing issues with this one. Its a small PR so I can review it again.

govi2010:
@adityarb88 I am finding bit difficulty. But I think I need to learn one day , and it better to start from today.  I am running command blindly don't know what it is affect.

adityarb88:
@govi2010: You can also refer to this video: https://www.youtube.com/watch?v=2E23I9PzplM. Let me know if that helps.

PR #1659, opened 2017-10-31
[NG] updates for Angular 5

gnomeontherun:
Fixes a deprecation removal and bumps TypeScript and RX versions to minimum expected versions for Angular 5.

Wanting to hold this open and continue testing with latest RC until Angular 5 goes stable.


chunghha:
Are you going to have another PR for Rxjs 5.5 lettable operators (ie. map operator used in several places)? And, I think minimum TS version for NG 5 is 2.4.2+.

gnomeontherun:
I didn't want to do the lettable operators yet, just to limit the changeset here to what was essential. That would take a longer time to bring through. 

Yes the minimum RxJS is now 5.5, and TypeScript is >=2.4.2. I don't see a blocking reason not to update to latest if we don't see any issues.

package.json
```diff
@@ -101,7 +101,7 @@
         "text-transform-loader": "^2.0.0",
         "ts-node": "~3.2.0",
         "tslint": "~5.7.0",
-        "typescript": "~2.3.3",
+        "typescript": "^2.5.3",
```
euskadi31: The version of Typescript supported by Angular 5 is 2.4.2, see [changelog](https://github.com/angular/angular/blob/master/CHANGELOG.md) (commit: [ca5aeba](https://github.com/angular/angular/commit/ca5aeba))

gnomeontherun: The compiler requires 2.4, and It is actually set to `"typescript": ">=2.4.2 <2.5",` for now, because there is still testing to be done on 2.5 or 2.6 (not that there is a known reason it fails to my knowledge).

euskadi31: ok

euskadi31: see: https://github.com/angular/angular/issues/19587#issuecomment-337713553

adityarb88: Reminder to update to ~2.4.2

WuglyakBolgoink:
> @angular/compiler-cli@5.0.0 requires typescript@'>=2.4.2 <2.5.0' but 2.6.1 was found instead.

gnomeontherun:
Updated also with rxjs changes since there was only one change.

package.json
gnomeontherun: Done

gnomeontherun:
Updated with package-lock.json file

PR #1670, opened 2017-11-03
[NG] access the minimum width only on the first drag attempt

### LINKED ISSUE #1537: Datagrid column cannot be shrunk
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
I have a wizard which has a datagrid in it. The datagrid column resizing works fine if the wizard is opened initially.

### Actual behavior
Initially the wizard is not opened initially, opening the wizard will cause datagrid column not able to shrink.

After debugging I found the following line NaN:
https://github.com/vmware/clarity/blob/55e90f4b474a25fa259d387829e8f7db66eb5b80/src/clarity-angular/data/datagrid/render/column-resizer.ts#L45

Which causes the following line NaN too:
https://github.com/vmware/clarity/blob/55e90f4b474a25fa259d387829e8f7db66eb5b80/src/clarity-angular/data/datagrid/render/column-resizer.ts#L95

I noticed that when wizard is not opened yet, the datagrid (Line 45) is already being initialized.

### Reproduction of behavior
https://plnkr.co/edit/lgSJrYiNBzKphdR9DBsH?p=preview

### Environment details

* **Angular version:** 4.0.1

* **Clarity version:** 0.9.2

* **OS and version:** MacOS El Captain

* **Browser:** Chrome 61

### LINKED ISSUE #1666: Can not shrink datagrid columns on Chrome
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
I am creating clarity datagrid columns by iterate column definitions with *ngFor.
I expect the created clarity datagrid columns to be shrinkable.
### Actual behavior
When create clarity datagrid columns (clr-dg-column elements)  with *ngFor, they can't be shrink on Chrome.
The shrink of the datagrid columns work on FF and IE. 

### Reproduction of behavior
https://plnkr.co/edit/UGgTYUNIQMOm19pxgbXp?p=preview

### Environment details

* **Angular version:4.x.x** 

* **Clarity version:10.06**

* **OS and version: Windows** 

* **Browser:** Chrome 62.0.3202.75

Shijir:
Fixes #1537
Fixes #1666 

This PR fixes the problem that Datagrid columns instantiate even when it's in a hidden state modal. As such, it tries to call methods in our DomAdapter, which returns `Nan` as the dom is not yet rendered. The solution I am providing here is that it will call the DomAdapter methods only on the first drag attempt by a user after the Datagrid is fully rendered. 

One of the unit tests check that if it sets the minumum width only after the first drag attempt (`dragStartHandler()`). I have added one more unit that makes sure that if one column keeps expanding, the other columns shouldn't shrink beyond its minimum width. Also, I've fixed some typos in the unit tests.

Tested on: Chrome, Firefox, Safari, Edge, IE.

### DIFF
--- a/src/clarity-angular/data/datagrid/render/column-resizer.ts
+++ b/src/clarity-angular/data/datagrid/render/column-resizer.ts
@@ -42,9 +42,7 @@ export class DatagridColumnResizer implements AfterViewInit, OnDestroy {
     }
 
     ngAfterViewInit() {
-        this.columnMinWidth = this.domAdapter.minWidth(this.columnEl);
         this.handleTrackerEl = this.dragDispatcher.handleTrackerRef.nativeElement;
-
         this.dragDispatcher.addDragListener();
         this.subscriptions.push(this.dragDispatcher.onDragStart.subscribe(() => this.dragStartHandler()));
         this.subscriptions.push(this.dragDispatcher.onDragMove.subscribe(($event) => this.dragMoveHandler($event)));
@@ -52,6 +50,10 @@ export class DatagridColumnResizer implements AfterViewInit, OnDestroy {
     }
 
     dragStartHandler(): void {
+        if (!this.columnMinWidth) {
+            // sets the min width only on the very first drag attempt
+            this.columnMinWidth = this.domAdapter.minWidth(this.columnEl);
+        }
         this.renderer.setStyle(this.handleTrackerEl, "display", "block");
         this.renderer.setStyle(document.body, "cursor", "col-resize");
         this.dragDistancePositionX = 0;
--- a/src/clarity-angular/data/datagrid/render/column-resizer.spec.ts
+++ b/src/clarity-angular/data/datagrid/render/column-resizer.spec.ts
@@ -70,12 +70,25 @@ export default function(): void {
             expect(column3DragDispatcher.handleTrackerRef.nativeElement).not.toBeUndefined();
             expect(column4DragDispatcher.handleTrackerRef.nativeElement).not.toBeUndefined();
         });
-        it("accesses column minumum width when the drag listener is added in", function() {
+        it("accesses column minimum width on the first drag attempt", function() {
+            column1ResizerDirective.dragStartHandler();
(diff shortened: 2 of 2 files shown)


adityarb88:
@Shijir: We should try to mention what issue your PR is trying to fix and provide more details on that in the PR description. Also, what testing was done to make sure that the issue is fixed.

Shijir:
@adityarb88 Sorry, my bad... you are absolutely right. I will update the PR and provide more contextual information as I get back.

src/clarity-angular/data/datagrid/render/column-resizer.ts
```diff
@@ -42,16 +42,18 @@ export class DatagridColumnResizer implements AfterViewInit, OnDestroy {
...
         this.handleTrackerEl = this.dragDispatcher.handleTrackerRef.nativeElement;
-
         this.dragDispatcher.addDragListener();
         this.subscriptions.push(this.dragDispatcher.onDragStart.subscribe(() => this.dragStartHandler()));
         this.subscriptions.push(this.dragDispatcher.onDragMove.subscribe(($event) => this.dragMoveHandler($event)));
         this.subscriptions.push(this.dragDispatcher.onDragEnd.subscribe(() => this.dragEndHandler()));
     }
 
     dragStartHandler(): void {
+        if (!this.columnMinWidth) {
```
adityarb88: Question for my understanding. How is this different than initializing the min-width on `ngAfterViewInit` since even that lifecycle method will be called only once, no? Is this change related to hideable columns?

adityarb88:
No worries :-) Thanks!

src/clarity-angular/data/datagrid/render/column-resizer.ts
Shijir: This `dragStartHandler()` is called only when a user drags the column. The problem was that `this.columnMinWidth` was getting `Nan` value because the datagrid was instantiating even when it's nested inside a hidden modal.

PR #1691, opened 2017-11-14
[NG] Import DOCUMENT from @angular/common

chunghha:
instead of @angular/platform-browser. NG5 support.


gnomeontherun:
Hi @chunghha thanks for this update, good catch on the deprecation.

Can you fix your PR by running `npm run clang:format` on your code change? Our formatter can be a bit opinionated and probably wants the imports to be sorted alphabetically by package name. That will allow the tests to run.

Issue #1695, opened 2017-11-15
Multi-app level alerts do not cycle correctly


reddolan:
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
- clicking on the right arrow while being on the last alert, should cycle it to the first alert
- clicking on the left arrow while being on the first alert, should cycle it to the last alert

### Actual behavior
- clicking the right arrow on the last alert does nothing
- clicking the left arrow on the first alert does nothing

### Reproduction of behavior
![mulit-app-level-alert](https://user-images.githubusercontent.com/1858226/32852025-97cadc72-c9eb-11e7-8aa0-ac9d2038bb84.gif)

### Environment details

* **Clarity version:** 0.10.13

* **OS and version:** 

* **Browser:** [all]

PR #1697, opened 2017-11-15
[NG] Enable cycling of alerts

adityarb88:
Fixes: #1695 

Tested on Chrome

nimbus-record-video-2017-11-15-13-33-36

### DIFF
--- a/src/clarity-angular/emphasis/alert/providers/multi-alert-service.ts
+++ b/src/clarity-angular/emphasis/alert/providers/multi-alert-service.ts
@@ -53,11 +53,11 @@ export class MultiAlertService {
     }
 
     next() {
-        this.current = (this.current === this.activeAlerts.length - 1) ? this.current : this.current + 1;
+        this.current = (this.current === this.activeAlerts.length - 1) ? 0 : this.current + 1;
     }
 
     previous() {
-        this.current = (this.current === 0) ? this.current : this.current - 1;
+        this.current = (this.current === 0) ? this.activeAlerts.length - 1 : this.current - 1;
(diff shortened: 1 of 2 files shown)


hippee-lee:
lgtm

src/clarity-angular/emphasis/alert/providers/multi-alert-service.ts
```diff
@@ -53,11 +53,11 @@ export class MultiAlertService {
     }
 
     next() {
-        this.current = (this.current === this.activeAlerts.length - 1) ? this.current : this.current + 1;
+        this.current = (this.current === this.activeAlerts.length - 1) ? 0 : this.current + 1;
```
Shijir: 🤔  Is cycling alerts a good behavior though? What If it has only one alert? When there is only alert, users can still keep clicking on both previous and next buttons?

reddolan: If there is only one alert, the cycle control should be showing.

reddolan: The cycle control should also disappear after the second to last alert is dismissed, then leaving only one alert

adityarb88: As @reddolan correctly pointed out, if there is only one alert, the pager disappears.

Shijir: Oh so the pager disappears. Thanks for the clarification.

PR #1698, opened 2017-11-15
[DO-NOT-REVIEW] Alerts Pager Bug Fix

adityarb88:
Fixes a bug with the alerts pager component where if a user is viewing the last alert and that last alert is removed from the DOM, the alert pager disappears. More info below:

Before:
before

After:
after

Tested on Chrome

### DIFF
--- a/src/clarity-angular/emphasis/alert/alerts.spec.ts
+++ b/src/clarity-angular/emphasis/alert/alerts.spec.ts
@@ -49,7 +49,6 @@ describe("Alerts component", function() {
                 TestBed.configureTestingModule({imports: [ClrEmphasisModule], declarations: [componentType]});
 
                 fixture = TestBed.createComponent(componentType);
-                // fixture.detectChanges();
             };
         });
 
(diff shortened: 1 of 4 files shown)


src/clarity-angular/emphasis/alert/alerts.spec.ts
```diff
@@ -139,6 +138,17 @@ describe("Alerts component", function() {
...
+
+        it("makes the previous alert active when the current alert is removed from the DOM", function() {
+            fixture.componentInstance.currentAlertIndex = 1;
+            fixture.detectChanges();
+            expect(alertElements[0].classList).toContain("alert-hidden");
+
+            fixture.componentInstance.alertInstances.reset([this.alert]);
+            fixture.detectChanges();
+
+            console.log(fixture.componentInstance.currentAlertIndex);
```
gnomeontherun: Make this an `expect` to confirm the result.

adityarb88: good catch. this means that my test didn't run itself. let me check and confirm if this change actually works.

PR #1702, opened 2017-11-20
Internal class name changes and organization updates

### LINKED ISSUE #1729: ng-packagr question
```
[ ] bug
[ ] feature request
[ ] enhancement
[x] Just a question
```

Hi,
I am learning about ng-packagr and found your project to be using it.
I would appreciate your clarification on my question.

Noticed that in this file:
https://github.com/vmware/clarity/blob/master/src/clarity-angular/index.ts

All the modules have been exported first. Wondering if there is a module per component why the module for each component is not exported on its own index.ts file along with components, directives, ... .

Thanks.

gnomeontherun:
This is renaming and organizing classes and objects based on our established conventions. It does not deal with input/output bindings.

* [x] Establish Patterns

* [x] Button
* [x] Code
* [x] Data
* [x] Emphasis
* [x] Forms
* [x] Icon
* [x] Layout
* [x] Modal
* [x] Popover
* [x] Wizard
* [x] Util

closes #1729


src/app/_utils/code-example.ts
```diff
@@ -20,7 +20,7 @@ import {CodeHighlight} from "../../clarity-angular/code/syntax-highlight/syntax-
     `]
 })
 export class CodeExample {
-    @ViewChild(CodeHighlight) codeHighlight: CodeHighlight;
+    @ViewChild(ClrCodeHighlight) codeHighlight: ClrCodeHighlight;
```
youdz: Unrelated to your change: I don't even remember why we have to query code highlights all over the place. But not only do we not need it for the demo app, this directive was also written in a few minutes and was never meant to make it in prod, let alone in our public API.

Had it been fixed before, we wouldn't have needed to rename it everywhere. 😢

src/clarity-angular/button/index.ts
```diff
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import {Type} from "@angular/core";
-import {Button} from "./button-group/button";
-import {ButtonGroup} from "./button-group/button-group";
+import {ClrButton as Button} from "./button-group/button";
+import {ClrButtonGroup as ButtonGroup} from "./button-group/button-group";
```
youdz: That's not what you want. I get that you're trying to avoid breaking changes, but the whole point is to deprecate _and_ offer the new standard, not just deprecate. Otherwise people have nothing to move to. As a side note, you apparently got unlucky again and took (one of?) the only module(s) that didn't export the directives by themselves, when it should.

All in all, this file should be:
```typescript
import {Type} from "@angular/core";
import {ClrButton} from "./button";
import {ClrButtonGroup} from "./button-group";

export {ClrButton} from "./button";
export {ClrButtonGroup} from "./button-group";

export const CLR_BUTTON_GROUP_DIRECTIVES: Type<any>[] = [ClrButton, ClrButtonGroup];

/* tslint:disable */
/**
 * @deprecated
 */
export const Button = ClrButton;
/**
 * @deprecated
 */
export const ButtonGroup = ClrButtonGroup;
/**
 * @deprecated
 */
export const BUTTON_GROUP_DIRECTIVES = CLR_BUTTON_GROUP_DIRECTIVES;
```

And this is what I get when I try to use them:
screen shot 2017-11-21 at 20 40 57

mathisscott: @youdz: We don't want to introduce these breaking changes in 0.11. We have a very specific scope of breaking changes for 0.11. So if we are not deprecating, we will need to hold off to merge this in until 0.12.

I'm not sure how this jumped up in prioritiy. I thought we were clear on what was going into 0.11.

youdz: That's the whole point of my comment, this is not a breaking change, just a deprecation.
It allows us to warn people that 0.12 will include the removal of these deprecated exports as a breaking change.

Whether this is 0.11.0, 0.11.1 or later, I don't think it matters. But the point is to have it early in 0.11.x, to give people a full minor's notice.

src/app/_utils/code-example.ts
gnomeontherun: If it was never intended, we can deprecate it?

src/clarity-angular/button/index.ts
gnomeontherun: I did this PR as a quick prototype so it could provide the pattern for the rest of the work. I was also thinking it needs to be as early as possible.

src/app/_utils/code-example.ts
youdz: Well, apparently people now completely depend on it, including the Prism dependency and all. We've had requests for bugfixes on this component.

So rather than deprecate it, I think we should own our mistake and clean it up to make it an up-to-standards Clarity component. But as usual, the priority on that is super low.

src/clarity-angular/button/index.ts
```diff
@@ -4,7 +4,18 @@
...
-export const BUTTON_GROUP_DIRECTIVES: Type<any>[] = [ButtonGroup, Button];
+export * from "./button-group/button";
+export * from "./button-group/button-group";
+export const CLR_BUTTON_GROUP_DIRECTIVES: Type<any>[] = [ClrButtonGroup, ClrButton];
+
+/* tslint:disable variable-name */
+/** @deprecated since 0.11 */
+export const Button = ClrButton;
+/** @deprecated since 0.11 */
+export const ButtonGroup = ClrButtonGroup;
```
youdz: You already do all this in `button-group/index.ts`, why are you doing it again here? Just the `export * from "./button-group/button-group";` is enough.

gnomeontherun: Turns out this file is never imported, so it could be removed.

youdz: Oh, main gets the submodules? We should make a tree of index files, should we?
Main imports from button which itself imports from button-group. That would make maintenance easier, otherwise we have two files to update every time.

src/clarity-angular/button/button-group/button-group.module.ts
```diff
@@ -5,16 +5,27 @@
...
 @NgModule({
     imports: [CommonModule, ClrIconModule, ClrCommonPopoverModule],
     declarations: [CLR_BUTTON_GROUP_DIRECTIVES],
     exports: [CLR_BUTTON_GROUP_DIRECTIVES]
 })
 export class ClrButtonGroupModule {}
+
+/* tslint:disable variable-name */
+/** @deprecated since 0.11 */
+export const Button = ClrButton;
```
youdz: This might seem like a nitpick, but these deprecated exports are pure API considerations, I believe they should have stayed in the index file, not the module one.

gnomeontherun: I wanted that originally too, but then I ran into circular dependency issues and moved any declarations from the index files. It makes the index cleaner since we don't have to import and then export, which was part of my goal as well.

PR #1728, opened 2017-11-28
[NG] tweaks for server side rendering

gnomeontherun:
There are a few places in our code that trip up server side rendering, and this puts guards around them so you don't have errors or hangups.

The primary way to do this is to use `isPlatformBrowser()` and pass in the `PLATFORM_ID`. This will check if the browser is rendering or if is the server.


src/clarity-angular/data/datagrid/render/dom-adapter.spec.ts
```diff
@@ -3,6 +3,8 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import {ɵPLATFORM_BROWSER_ID} from "@angular/common";  // Might be hacky but it is exported.
```
hippee-lee: Typo `ɵPLATFORM_BROWSER_ID` ?

gnomeontherun: Nope, that is what it is. Its similar to how we handle the focus trap export. Its public, but not overly public.

youdz: Did you look into maybe using `platformBrowserTesting()` or `platformBrowser()` to create the platform and get the ID from there? Yes, it's much heavier, but it would use 100% public API, which is safer.
Unless you feel confident this `ɵPLATFORM_BROWSER_ID` will never be changed in a minor version, which sounds a bit optimistic. 😛

src/clarity-angular/data/datagrid/render/dom-adapter.ts
```diff
@@ -10,13 +10,19 @@
...
 
 @Injectable()
 export class DomAdapter {
+    constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+
     userDefinedWidth(element: any): number {
         element.classList.add("datagrid-cell-width-zero");
-        const userDefinedWidth = parseInt(getComputedStyle(element).getPropertyValue("width"), 10);
+        let userDefinedWidth = 0;
+        if (isPlatformBrowser(this.platformId)) {
```
youdz: Guarding every single method of the class is a bit painful. I think a cleaner solution would be to have a separate `NoopDomAdapter` or other dummy name, that pretty much returns 0 for everything, and then you switch the provider declaration from `providers: [DomAdapter]` to:
```typescript
providers: [{
    provide: DomAdapter,
    useFactory: function(platformId: Object) {
        if (isPlatformBrowser(platformId)) {
            return new DomAdapter();
        } else {
            return new NoopDomAdapter();
        }
    },
    deps: [PLATFORM_ID]
}]
```

I think it's both more maintainable and flexible at the same time. If we ever were to have an actual implementation on the server-side, we would already have a clean separate class for it.
That was actually the point of isolating all DOM access for the Datagrid to a single class.

src/clarity-angular/data/datagrid/render/dom-adapter.spec.ts
gnomeontherun: It is a public API, so it should go through a deprecation before change. However it must be excluded from the docs for some reason. I guess there are no guarantees in life though

I'll look at the testing options, but no documentation on it anywhere...

src/clarity-angular/data/datagrid/render/dom-adapter.ts
gnomeontherun: Sure, will do.

src/clarity-angular/data/datagrid/render/dom-adapter.spec.ts
youdz: No, it's actually in the public API for internal reasons, and is not actually public. As such, they reserve the right to change this without even mentioning it in the changelog. They have several exports like that, all of them being here to allow their own packages to communicate.
I remember they did change the same ones for animations between 2 minors, with no warning and no note about it.

gnomeontherun: Well I've already refactored for the noop dom adapter and this won't be necessary then so 👍

PR #1732, opened 2017-11-29
[NG] Fix nested tabs link

Shijir:
- Fixes 1494

Tabs component will now only project its immediate TabLinkDirectives.  

Before:
screenshot 2017-11-29 13 07 12

After:
screenshot 2017-11-29 13 11 05

Preview: http://nested-tabs-demo.surge.sh/tabs

Tested on: Chrome

### DIFF
--- a/src/clarity-angular/layout/tabs/tab-link.directive.ts
+++ b/src/clarity-angular/layout/tabs/tab-link.directive.ts
@@ -15,6 +15,7 @@ import {
 import {IF_ACTIVE_ID, IfActiveService} from "../../utils/conditional/if-active.service";
 import {TemplateRefContainer} from "../../utils/template-ref/template-ref-container";
 import {AriaService} from "./aria-service";
+import {TABS_ID} from "./tabs-id.provider";
 
 let nbTabLinkComponents: number = 0;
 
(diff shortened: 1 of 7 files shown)


src/clarity-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -38,7 +39,7 @@ export class TabLinkDirective {
 
     constructor(public ifActiveService: IfActiveService, @Inject(IF_ACTIVE_ID) private id: number,
                 private ariaService: AriaService, private el: ElementRef, private cfr: ComponentFactoryResolver,
-                private viewContainerRef: ViewContainerRef) {
+                private viewContainerRef: ViewContainerRef, private tabs: Tabs) {
```
youdz: This may seem like a nitpick but I would like you to make the change: to avoid the whole circular dependency with `forwardRef()` and all, you should just declare a provider for the tabs id that's a simple value, and inject that instead of `Tabs`. It will also let you run the unit tests for tab links without having to wrap them in a tabs component, which is much better isolation.

Shijir: @youdz I completely overlooked an option of using providers. I wanted to actually use a service first, but I thought that would just make it complicated. I will opt for providers. Thanks for the suggestion

src/clarity-angular/layout/tabs/tabs.ts
```diff
@@ -3,26 +3,28 @@
...
-import {Component, ContentChildren, QueryList} from "@angular/core";
+import {AfterContentInit, Component, ContentChildren, forwardRef, QueryList} from "@angular/core";
 
 import {IfActiveService} from "../../utils/conditional/if-active.service";
 import {IfOpenService} from "../../utils/conditional/if-open.service";
 
 import {TabLinkDirective} from "./tab-link.directive";
 import {TabsService} from "./tabs-service";
 
+let ngTabsComponent = 0;
```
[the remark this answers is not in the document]
Shijir: Right 😂  it's not supposed to be `ngTabsComponent `.

PR #1742, opened 2017-12-04
[NG] Adjust step nav height for inline wizards.

### LINKED ISSUE #1711: Inline wizard navigation pane should have max-height: 100%
```
[ x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Using inline wizard, the navigation pane of the wizard should have maximum height of 100% instead of .75h.

The issue is filed by request of @adityarb88 from issue https://github.com/vmware/clarity/issues/1704

### Actual behavior

The inline wizard scales horizontally to consume all available width of the iframe but it does not scale vertically to consume all available height. It turns out clr-wizard--inline defines "position: static" and the max-height = .75h of navigation pane. Using the inline wizard css the content are max-height is changed to 100%. The same should apply for nav pane.
The following screenshot explains the problem with the Inline wizard:
inline-wizard-nav-pane-broken

### Reproduction of behavior

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

hippee-lee:
- closes #1711

## Browsers
Tested in:

- Chrome
- Safari
- Firefox
- IE 11
- Edge

### DIFF
--- a/src/clarity-angular/wizard/_wizard.clarity.scss
+++ b/src/clarity-angular/wizard/_wizard.clarity.scss
@@ -538,6 +538,10 @@
             min-height: 100%;
             height: auto;
             max-height: 100%;
+
+            .clr-wizard-stepnav {
+                height: 100%;
+            }
         }
     }
 


src/clarity-angular/wizard/_wizard.clarity.scss
```diff
@@ -538,6 +538,10 @@
             min-height: 100%;
             height: auto;
             max-height: 100%;
+
+            .clr-wizard-stepnav {
```
mathisscott: Do we want to do this everywhere for all wizards or does this just need to be for inline wizards?

Did we test this change to verify there's no unexpected issues with it?

hippee-lee: My understanding if the issue was that this was for inline wizards only. Added it for all wizards breaks our cssRegression tests for most wizardz ... they all get scroll bars as the `.clr-wizard-stepnav` pushes below the container. 

I'll update the description for browsers tested in.

mathisscott: Cool. That's what I thought. Thanks for the clarification.

Issue #1747, opened 2017-12-05
Two ways to write alert-types [clrAlertType]


digitlz:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[ x ] enhancement
```

### Expected behavior
When I'm creating a `<clr-alert>` I have to set `[clrAlertType]="'alert-info'"` . When I want to add pagination for multiple alerts I will have to use `<clr-alerts>` with `<clr-alert>` as children using `[clrAlertType]="'alert-info'"`. 

### Actual behavior
The styling will not allow a background-color in the multi-alerts-pagination if I use `[clrAlertType]="'alert-info'"` - it will only appear when I'm using `[clrAlertType]="'info'"` for each alert which doesn't make sense for me. Why is there a new declaration for the alert-type specifically if I want to use the pagination?
I had an `*ngFor` for AppLevelAlerts according to [the documentation](https://vmware.github.io/clarity/documentation/alerts) but after the update and after I implemented multi-page-alerts I had no background-color for the pagination. After some research I found out it was the `[clrAlertType]="'alert-info'"`, now I had to use `[clrAlertType]="'info'"`. Shouldn't it stay the same across all alerts (and ways to use alerts) to use "alert-..."?
[This file](https://github.com/vmware/clarity/blob/master/src/clarity-angular/emphasis/alert/utils/alert-types.ts) contains the types and I think It's quite redundant. Shouldn't there be one way instead of two?

### Reproduction of behavior
I created a simple Plunker which has one alert with `[clrAlertType]="'alert-info'"` and the other with `[clrAlertType]="'info'"`.
[Link to Plunker](https://embed.plnkr.co/eIRTIJykhoeC556IMieL/)

### Environment details

* **Angular version:** 4.4.6

* **Clarity version:** 0.10.16

* **OS and version:** macOS 10.13.1

* **Browser:** [Chrome 62 | Safari 11]

mathisscott:
@digitlz 
The comment [in the File your referenced](https://github.com/vmware/clarity/blob/master/src/clarity-angular/emphasis/alert/utils/alert-types.ts) points out that `alert-*` types have been deprecated and will be removed.

That said, the new Alerts pager should support both for the time being. I will mark this as a bug.

PR #1755, opened 2017-12-07
[NG] Update global alerts pager to recognize deprecated alert types

mathisscott:
• previously only recognized “info” and such
• now recognizes both “info” and deprecated “alert-info” alert types
• also added tests for this functionality

Tested in:
✔︎ Chrome

Closes: #1747


mathisscott:
This is ready to merge. Just needs one more review.

PR #1756, opened 2017-12-07
[NG] Update global alerts pager to recognize deprecated alert types - v0.10

mathisscott:
• previously only recognized “info” and such
• now recognizes both “info” and deprecated “alert-info” alert types
• also added tests for this functionality

Tested in:
✔︎ Chrome

Closes: #1747

### DIFF
--- a/src/clarity-angular/emphasis/alert/alerts.ts
+++ b/src/clarity-angular/emphasis/alert/alerts.ts
@@ -7,6 +7,7 @@ import {AfterContentInit, Component, ContentChildren, EventEmitter, Input, Outpu
 import {Alert} from "./alert";
 import {MultiAlertService} from "./providers/multi-alert-service";
 
+// alert types in the format of "alert-*" are deprecated and will be removed in 0.12 or later
 @Component({
     selector: "clr-alerts",
     templateUrl: "./alerts.html",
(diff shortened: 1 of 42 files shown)


src/clarity-angular/emphasis/alert/alerts.ts
```diff
@@ -13,10 +13,10 @@ import {MultiAlertService} from "./providers/multi-alert-service";
     providers: [MultiAlertService],
     host: {
         "[class.alerts]": "true",
-        "[class.alert-danger]": "this.currentAlertType == 'danger'",
-        "[class.alert-info]": "this.currentAlertType == 'info'",
-        "[class.alert-success]": "this.currentAlertType == 'success'",
-        "[class.alert-warning]": "this.currentAlertType == 'warning'"
+        "[class.alert-danger]": "this.currentAlertType == 'danger' || this.currentAlertType == 'alert-danger'",
```
gnomeontherun: Can we put a comment that these are deprecated so we can find it later? Same on the other PR now that I think of it.

`/** @deprecated */`

mathisscott: Yep. Will do. Thanks!

src/clarity-angular/emphasis/alert/alerts.ts
```diff
@@ -7,6 +7,7 @@ import {AfterContentInit, Component, ContentChildren, EventEmitter, Input, Outpu
 import {Alert} from "./alert";
 import {MultiAlertService} from "./providers/multi-alert-service";
 
+// alert types in the format of "alert-*" are deprecated and will be removed in 0.12 or later
```
mathisscott: @gnomeontherun 
Please verify that this is sufficient. Thanks!

mathisscott:
This is ready to merge. Just needs one more review and for @gnomeontherun on the run to verify my comment.

PR #1764, opened 2017-12-12
[NG] fix selection with pagination in datagrid

### LINKED ISSUE #1393: DataGrid - server side selection not working with pagination

gnomeontherun:
This will update the current value instance when new items are set, such as from pagination, by using trackby to ensure that current array or item maps correctly to the instances of the currently displayed items.

It does this by always updating the `current` or `currentSingle` value with the updated instance if it could be matched based on the trackBy. This way our current equality checks work and the only time that we have to do lookups is during setting the state of a selected item. Gets should remain as fast as before.

There are 8 scenarios for selection and pagination that we can properly support. They are the following four scenarios for both single and multi-selection (2x4=8, yay math). There are explicit unit tests for each of these 8 scenarios.

* client side, no trackBy
* client side, trackBy index
* client side, trackBy item
* server side, trackBy item

Server side trackBy index and server side no trackBy don't make any sense because we cannot properly track instances of objects over pages. We need trackBy to know how to identify the object key, and using the index is folly because we only know the index of the current page (which might be index `4` but on page 4 that would actually be the index `44` and the Datagrid doesn't know that). 

fixes #1393

### DIFF
--- a/src/app/datagrid/inventory/inventory.ts
+++ b/src/app/datagrid/inventory/inventory.ts
@@ -89,8 +89,9 @@ export class Inventory {
     }
 
     fetch(skip: number = 0, limit: number = this._currentQuery.length): Promise<FetchResult> {
-        const result:
-            FetchResult = {users: this._currentQuery.slice(skip, skip + limit), length: this._currentQuery.length};
+        // Stringify and parse to mimic new object creation like a real HTTP request
+        const items = JSON.stringify(this._currentQuery.slice(skip, skip + limit));
+        const result: FetchResult = {users: JSON.parse(items), length: this._currentQuery.length};
         this._currentQuery = null;
         return this._fakeHttp(result);
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "clr-all",
-    "version": "0.11.0-alpha.1",
+    "version": "0.11.0-beta.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/src/app/datagrid/datagrid.demo.html
+++ b/src/app/datagrid/datagrid.demo.html
@@ -6,7 +6,7 @@
 
 <h2>Datagrid</h2>
 
-<ul>
+<ul class="datagrid-nav">
     <li><a [routerLink]="['./kitchen-sink']">Kitchen Sink (for CSS regression)</a></li>
     <li><a [routerLink]="['./structure']">Basic structure</a></li>
     <li><a [routerLink]="['./custom-rendering']">Custom cell rendering</a></li>
--- a/src/app/datagrid/datagrid.demo.scss
+++ b/src/app/datagrid/datagrid.demo.scss
@@ -4,7 +4,7 @@
 
 @import "../_utils/code";
 
-ul {
+.datagrid-nav {
     column-count: 3;
 }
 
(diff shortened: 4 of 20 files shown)


src/app/datagrid/inventory/inventory.ts
```diff
@@ -89,8 +89,9 @@ export class Inventory {
     }
 
     fetch(skip: number = 0, limit: number = this._currentQuery.length): Promise<FetchResult> {
-        const result:
-            FetchResult = {users: this._currentQuery.slice(skip, skip + limit), length: this._currentQuery.length};
+        // Stringify and parse to mimic new object creation like a real HTTP request
+        const items = JSON.stringify(this._currentQuery.slice(skip, skip + limit));
```
gnomeontherun: Force our fetch method to create new instances of data for each request like a real HTTP request.

src/clarity-angular/data/datagrid/datagrid-row.ts
```diff
@@ -41,7 +41,7 @@ let nbRow: number = 0;
                          class="datagrid-select datagrid-fixed-column">
                 <div class="radio">
                     <input type="radio" [id]="id" [name]="selection.id + '-radio'" [value]="item"
-                           [(ngModel)]="selection.currentSingle">
+                           [(ngModel)]="selection.currentSingle" [checked]="selection.currentSingle === item">
```
gnomeontherun: Without this, the state of the input doesn't properly resolve between pages. I believe this is due to ngModel forcing a second change detection cycle. https://github.com/angular/angular/blob/5.1.0/packages/forms/src/directives/ng_model.ts#L30

src/clarity-angular/data/datagrid/providers/selection.ts
```diff
@@ -22,18 +22,53 @@ export enum SelectionType {
...
+                    if (typeof newSingle !== "undefined") {
+                        this.currentSingle = newSingle;
+                    }
+                    this.current = leftOver;
+                }, 0);
+            }
+        }));
+
+        this.subscriptions.push(this._items.allChanges.subscribe((updatedItems) => {
             if (!this._selectable) {
```
youdz: The two subscriptions look very similar, I'm wondering if we can avoid code duplication.

And saying this just made me realize there is a final bug left here: use a client-side datagrid with a "heartbeat", that refreshes the complete set of items every X seconds.
In this specific case, we're still using the wrong index in the trackBy of the `this._items.allChanges` subscription.

I believe both remarks can be solved by putting the new code you wrote inside the `this._items.allChanges` subscription, and update the Items provider to:
```typescript
private _all: any[];
public set all(items: any[]) {
    if (this.smart) {
        this._all = items;
        this.emitAllChanges();
        this._filterItems();
    } else {
        this._displayed = items;
        this.emitAllChanges();
        this.emitChange();
    }
}
```

PR #1766, opened 2017-12-12
[NG] fix chocolate error in columns hidden by default

Shijir:
- Fixes 1069

Datagrid cell was throwing a 🍫  error when a user sets columns hidden by default. The main reason was that mapping the cell with the corresponding column header was taking place after the cell's view was checked and as a result, the cell properties had to change. 

This PR solution is to give the hidden class to a cell only after the mapping process completes with the help of `Renderer2`.

I have added one unit test that makes sure there is no 🍫  error when the columns are hidden by default. Also, I have done a manual testing in our `ks-app`.
 

Tested on: Chrome

### DIFF
--- a/src/clarity-angular/data/datagrid/datagrid-cell.ts
+++ b/src/clarity-angular/data/datagrid/datagrid-cell.ts
@@ -3,11 +3,9 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import {Component, ContentChildren, QueryList} from "@angular/core";
-
+import {Component, ContentChildren, ElementRef, QueryList, Renderer2} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
 import {Signpost} from "../../popover/signpost/signpost";
-
-import {DatagridHideableColumn} from "./datagrid-hideable-column";
 import {HideableColumnService} from "./providers/hideable-column.service";
 
 @Component({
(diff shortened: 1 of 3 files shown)


src/clarity-angular/data/datagrid/datagrid-cell.ts
```diff
@@ -56,7 +37,43 @@ export class DatagridCell {
...
-    constructor(public hideableColumnService: HideableColumnService) {}
+        this.setHiddenClass(hideableColumn.hidden);
+        this._hiddenStateSubscription = hideableColumn.hiddenChangeState.subscribe(() => {
+            this.setHiddenClass(hideableColumn.hidden);
+        });
+    }
+
+    private setHiddenClass(hideableColumnValue: boolean) {
+        if (hideableColumnValue) {
+            this._renderer.addClass(this._el.nativeElement, "datagrid-cell--hidden");
```
hippee-lee: I like how this lets us remove `host` bindings.

src/clarity-angular/data/datagrid/datagrid-cell.ts
```diff
@@ -56,7 +37,43 @@ export class DatagridCell {
...
      */
-    public id: string;
+    private _id: string;
+
+    set id(value: string) {
+        this._id = value;
+        this.mapHideableColumn(this._id);
+    }
+
+    private _hiddenStateSubscription: Subscription;
```
gnomeontherun: I believe our conventions would want this to be `private hiddenStateSubscription: Subscription;` without the `_` since its not a get/set property like `id` above.

Shijir: Thanks @gnomeontherun. I've updated as requested.

PR #1777, opened 2017-12-13
[ICONS] Update already injected icons' templates if their templates are

Shijir:
updated in the API

- Fixes 1758

With this PR, it will be possible for users to update icons' templates when they are already in the DOM. This will also solve the case that if users load 3rd party JS libraries (e.g. `clarity-icons.min.js`) right before the closing body tag, not in the head.

I have added two unit tests that make sure the injected templates get updated whenever they are updated in the API and properly show the title if it's specified.  

Before:
https://plnkr.co/edit/3jcwnYDre4MUtUiYQssX?p=preview
After:
https://plnkr.co/edit/9hCdFy9xBA9InLEE23xE?p=preview

Fixes the errors reproduced in the original issue poster's pen:
https://codepen.io/anon/pen/opgMgP

Another example in Angular app:
https://plnkr.co/edit/BoqUNY?p=preview

Tested on: Chrome, Firefox, Safari, Edge, IE11

### DIFF
--- a/src/clr-icons/clr-icons-api.ts
+++ b/src/clr-icons/clr-icons-api.ts
@@ -4,10 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import {IconAlias} from "./interfaces/icon-alias";
-import {IconTemplate} from "./interfaces/icon-template";
+import {IconAlias, IconShapeSources} from "./interfaces/icon-interfaces";
+import {ShapeTemplateObserver} from "./utils/shape-template-observer";
 
-const iconShapeSources: IconTemplate = {};
+const iconShapeSources: IconShapeSources = {};
 
 export class ClarityIconsApi {
     private static singleInstance: ClarityIconsApi;
@@ -44,10 +44,12 @@ export class ClarityIconsApi {
             }
 
             iconShapeSources[shapeName] = trimmedShapeTemplate;
+
+            ShapeTemplateObserver.instance.emitChanges(shapeName, trimmedShapeTemplate);
         }
     }
 
-    private setIconAliases(templates: IconTemplate, shapeName: string, aliasNames: string[]): void {
+    private setIconAliases(templates: IconShapeSources, shapeName: string, aliasNames: string[]): void {
         for (const aliasName of aliasNames) {
             if (this.validateName(aliasName)) {
                 Object.defineProperty(templates, aliasName, {
@@ -61,7 +63,7 @@ export class ClarityIconsApi {
         }
     }
 
-    add(icons?: IconTemplate): void {
+    add(icons?: IconShapeSources): void {
         if (typeof icons !== "object") {
             throw new Error(`The argument must be an object literal passed in the following pattern: 
                 { "shape-name": "shape-template" }`);
@@ -88,11 +90,6 @@ export class ClarityIconsApi {
             throw new TypeError("Only string argument is allowed in this method.");
         }
 
-        // if shapeName doesn't exist in the icons templates, throw an error.
-        if (!this.has(shapeName)) {
-            throw new Error(`'${shapeName}' is not found in the Clarity Icons set.`);
-        }
-
         return iconShapeSources[shapeName];
     }
 
@@ -111,7 +108,8 @@ export class ClarityIconsApi {
                     // set an alias to the icon if it exists in iconShapeSources.
                     this.setIconAliases(iconShapeSources, shapeName, aliases[shapeName]);
                 } else {
-                    throw new Error("The icon '" + shapeName + "' you are trying to set an alias to doesn't exist!");
+                    throw new Error(`An icon "${
+                        shapeName}" you are trying to set aliases to doesn't exist in the Clarity Icons sets!`);
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


src/clarity-icons/clarity-icons.spec.ts
```diff
@@ -21,7 +21,7 @@ import {SocialShapes} from "./shapes/social-shapes";
 import {TechnologyShapes} from "./shapes/technology-shapes";
 import {TravelShapes} from "./shapes/travel-shapes";
 
-describe("ClarityIcons", () => {
+fdescribe("ClarityIcons", () => {
```
mathisscott: Can we remove this `fdescribe`?

Shijir: Oh yes, will do  🤦‍♂️ 😂

src/clarity-icons/clarity-icons-api.ts
```diff
@@ -88,12 +98,26 @@ export class ClarityIconsApi {
...
+    // This method will save which icon is used as well as the reference to its wrapper element, which is clr-icon.
+    // By saving reference to clr-icon, we could access it without DOM traversal and update the injected template
+    // if the icon's template is updated in the API.
+    _cacheAndRetrieve(iconElement: ClarityIconElement, shapeName: string, cachedIconName?: string) {
+        if (!cachedIconName) {
+            cachedIconName = shapeName;
+        }
+
+        if (!usedIconsTemplateCache[cachedIconName]) {
+            usedIconsTemplateCache[cachedIconName] = [iconElement];
```
gnomeontherun: Why are you caching the icon as an array like this? You're only storing it once, and if an override icon is set this cache wouldn't be updated right?

Shijir:
@youdz I think your first and second points are related. I am currently working on the part actually where If the element is disconnected, it should be removed from the array. 

I am also thinking that I should avoid the term "cache" as it's not really doing caching. It doesn't save any states, It's collecting only the references to icon instances.

src/clarity-icons/clarity-icons-api.ts
Shijir: I think the term "cache" is more misleading here. I am thinking about renaming it. It collects only the references to the icon instances. It doesn't save any data state. But I am working on a part where if the element is removed from the DOM, the instance gets removed from the cache.

src/clarity-icons/index.ts
```diff
@@ -12,16 +12,12 @@ const clarityIcons: ClarityIconsApi = ClarityIconsApi.instance;
...
-    }
-} else {
-    customElements.define("clr-icon", ClarityIconElement);
+if (typeof window !== "undefined" && !window.hasOwnProperty("ClarityIcons")) {
+    // Setting a global object called "ClarityIcons" to expose the ClarityIconsApi.
+    window.ClarityIcons = clarityIcons;
 }
 
+// Defining clr-icon custom element
+customElements.define("clr-icon", ClarityIconElement);
```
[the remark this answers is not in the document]
Shijir: Gotcha, I will address this. 👌

youdz:
These arrays and index lookup are still going to be a problem on pages that have 1000 times the same icon (yes, people do crazy things and expect us to make it work). Like a list of N items all with the same little icon in the front and on a button.

By keeping track of all icons, you've got way too much work for one class. You should do it the other way round: 
- make a singleton service in Typescript (not Angular, just a simple singleton instance of a class that you import and use) that emits some observable or something when the shapes' templates are updated ,
- have all icons instances listen to that service and re-inject their template if they need to.

It could be a single observable that fires an object, it could be one observable per shape, it could be anything you think is going to be more efficient than this.

src/clarity-icons/clarity-icons-element.ts
```diff
@@ -147,7 +173,13 @@ ClarityIconElement.prototype._setAriaLabelledBy = function() {
...
 
-ClarityIconElement.prototype._injectTemplate = function() {
+ClarityIconElement.prototype._injectTemplate = function(shapeTemplate?: string) {
+    // Accepting the argument, shapeTemplate, will help us to update the shape template
+    // right before the injection.
+    if (shapeTemplate && shapeTemplate !== this.currentShapeTemplate) {
+        this.currentShapeTemplate = shapeTemplate;
+    }
+
     this.innerHTML = this.currentShapeTemplate;
```
gnomeontherun: Should this and the following `if` be inside the `if` above so it only runs if a new template is provided?

Shijir: You mean like this?
```ts
 if (shapeTemplate && shapeTemplate !== this.currentShapeTemplate) {
       this.currentShapeTemplate = shapeTemplate;
       this.innerHTML = this.currentShapeTemplate;
}
```

gnomeontherun: Right, if I understand it we don't want to reapply the innerHTML unless it changes.

src/clarity-icons/utils/shape-template-observer.ts
```diff
@@ -0,0 +1,55 @@
...
+export const changeHandlerCallbacks: ShapeTemplateObservables = {};
+
+export class ShapeTemplateObserver {
+    private static singleInstance: ShapeTemplateObserver;
+
+    private callbacks: ShapeTemplateObservables = changeHandlerCallbacks;
+
+    private constructor() {}
+
+    public static get instance(): ShapeTemplateObserver {
```
youdz: I smiled at this for some reason, I got Java flashbacks.

PR #1791, opened 2017-12-19
[UI] update dark theme tab border to Lilia’s color - v0.10

mathisscott:


mathisscott:
This is the same code edit as the one before. Just backported. I will merge after tests pass and someone verifies.

PR #1792, opened 2017-12-19
Website server-side rendering!

gnomeontherun:
This incorporates the changes for building the documentation with SSR and deploying a static version of the site that is prerendered for SEO reasons.


src/app/icons/icons.component.ts
```diff
@@ -1,10 +1,18 @@
 import {
-    Component, AfterViewInit, OnDestroy, Renderer2, ElementRef, ChangeDetectorRef
+    Component,
```
hippee-lee: At some point do you think we should run `clang:format` or [https://prettier.io/](prettier.io) on the site code?
Maybe after it settles down from all the changes.

src/tsconfig.server.json
```diff
@@ -0,0 +1,21 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../out-tsc/app",
+    "baseUrl": "./",
+    // Set the module format to "commonjs":
```
hippee-lee: Comment in JSON?

src/tsconfig.server.json
```diff
@@ -0,0 +1,21 @@
...
+    "types": [
+      "node"
+    ]
+  },
+  "exclude": [
+    "test.ts",
+    "**/*.spec.ts"
+  ],
+  // Add "angularCompilerOptions" with the AppServerModule you wrote
+  // set as the "entryModule".
```
hippee-lee: Comment in JSON?

src/app/icons/icons.component.ts
gnomeontherun: I did not because that would cause major changes that distract from the real changes. When we merge the website into master we'll have to do that though.

src/app/documentation/demos/datagrid/datagrid.demo.main.scss
```diff
@@ -5,4 +5,9 @@
 
 .demo-pagination-next {
     margin-left: auto;
+}
+
+.demo-list {
+    column-count: 3;
```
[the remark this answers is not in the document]
gnomeontherun: Yeah, I have it dynamic so it grows based on the number of routes and lays out as 3 columns always.

src/tsconfig.server.json
gnomeontherun: I looked this up, and tsconfig files support comments. https://github.com/lacolaco/tsconfig-cli/pull/1

src/app/icons/icons.component.ts
hippee-lee: Yes, that's what I meant. Not for this change but something to consider in the future.

src/tsconfig.server.json
hippee-lee: Cool!

PR #1822, opened 2017-12-29
[NG] update deprecated exports to not be tree shaken

gnomeontherun:
It appears that something in the build chain tree-shakes unused variables. To address this, exporting a class that extends another does not tree-shake the objects.

### DIFF
--- a/src/ks-app/package.json
+++ b/src/ks-app/package.json
@@ -12,15 +12,15 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^4.0.0",
-    "@angular/common": "^4.0.0",
-    "@angular/compiler": "^4.0.0",
-    "@angular/core": "^4.0.0",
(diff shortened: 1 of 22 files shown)


src/ks-app/package.json
```diff
@@ -12,15 +12,15 @@
...
-    "@angular/animations": "^4.0.0",
-    "@angular/common": "^4.0.0",
-    "@angular/compiler": "^4.0.0",
-    "@angular/core": "^4.0.0",
-    "@angular/forms": "^4.0.0",
-    "@angular/http": "^4.0.0",
-    "@angular/platform-browser": "^4.0.0",
-    "@angular/platform-browser-dynamic": "^4.0.0",
-    "@angular/router": "^4.0.0",
+    "@angular/animations": "^5.0.0",
```
hippee-lee: Thanks for bumping these.

PR #1837, opened 2018-01-03
Issue/core copyright date updates

hippee-lee:
This updates the project files (`ts`, `scss` & `html`) for 2018. 

Because I did it automatically with IntelliJ's Copyright tool copyright comments should be in a standard format across all filetypes (i.e same comment format for `ts`, `scss` and `html` files). 

There are also some trailing spaces that were removed (Side effect from the Copyright tool).

### DIFF
--- a/src/app/alert/alert.demo.scss
+++ b/src/app/alert/alert.demo.scss
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
 
 @import "../_utils/code";
 
--- a/src/app/button-group/angular/hide-show-overflow-toggle/hide-show-overflow-toggle.html
+++ b/src/app/button-group/angular/hide-show-overflow-toggle/hide-show-overflow-toggle.html
@@ -1,3 +1,9 @@
+<!--
+  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
 <h4>Toggle Overflow Menu</h4>
 
 <div class="clr-example example-min-height">
--- a/src/app/color/color-palette.demo.scss
+++ b/src/app/color/color-palette.demo.scss
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
 
 @import "./color-shared.demo";
 
--- a/src/clr-angular/code/syntax-highlight/_code.clarity.scss
+++ b/src/clr-angular/code/syntax-highlight/_code.clarity.scss
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
 
 @include exports('code.clarity') {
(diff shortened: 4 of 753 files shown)


src/app/button-group/angular/hide-show-overflow-toggle/hide-show-overflow-toggle.html
```diff
@@ -1,3 +1,9 @@
+<!--
```
gnomeontherun: Do we really need to add these? When its bundled templates are moved into the JS anyways. This is also not shipped code, just our sample app.

src/app/color/color-palette.demo.scss
```diff
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
```
gnomeontherun: I got to thinking, is it even necessary to declare the year? Angular doesn't https://github.com/angular/angular/blob/master/packages/common/src/common.ts.

It links to a page on their website that contains the date. This is an annoying housekeeping task that would be nice to remove, but I suspect we'd have to talk to legal.

src/clr-angular/code/syntax-highlight/_code.clarity.scss
```diff
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
```
gnomeontherun: Why do all `scss` files have `/*!`?

src/clr-angular/forms/checkbox/_checkbox.clarity.scss
```diff
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
```
gnomeontherun: Noticed as well that Material doesn't put copyright notices on non typescript files (since they are bundled I guess), nor on spec files.

src/clr-angular/wizard/wizard.module.ts
```diff
@@ -1,5 +1,5 @@
-/**
- * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
```
gnomeontherun: Just wondering if we should put an explicit link?

src/app/button-group/angular/hide-show-overflow-toggle/hide-show-overflow-toggle.html
hippee-lee: IntelliJ added them when I processed the `/src` directory. I can see if I can change the way it adds multi-line comments.

src/app/color/color-palette.demo.scss
hippee-lee: Yes, I believe the copyright text for files was given to us. Perhaps it s worth asking if the year is needed?

src/clr-angular/code/syntax-highlight/_code.clarity.scss
hippee-lee: IntelliJ added it when I processed all of the `src`1 files. I can see if I can configure it to remove the !.

src/clr-angular/forms/checkbox/_checkbox.clarity.scss
hippee-lee: I've seen them on some of our scss files but not others. I thought this would be a good time to make them all the same.

src/clr-angular/wizard/wizard.module.ts
hippee-lee: Link to the LICENSE.txt file on github? Sure I can add that.

youdz:
To answer several of Jeremy's questions at once:
- The license wording has been given to us by the legal department at VMware, we have to respect it word for word.
- The wording included the year, so we're respecting that unless said legal department says it's fine to drop it.
- The license isn't here for deliverables, it's here for every line of code we've written. Anything here on Github is public, so they all need the license on them. Technically, I believe even our old gulp script files needed it. The only place we can't put it is configuration JSONs, because JSON doesn't allow comments.

src/app/alert/alert.demo.scss
```diff
@@ -1,6 +1,8 @@
-// Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
-// This software is released under MIT license.
-// The full license information can be found in LICENSE in the root directory of this project.
+/*!
```
youdz: These `/*!` "loud" comments are preserved in the compressed output by the Sass compiler. You **absolutely do not want** to have these on every source file, or the copyright will be written hundreds of times in our final CSS. You want non-loud comments all over the place, and a single loud comment at the top of `main.scss`.

Issue #1836, opened 2018-01-02
btn-sm button with loading indicator doesn't look good


rongery:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

A <button> with a class of .btn-sm and clrLoading="true" should lay out nicely.

### Actual behavior

See the plunker. The loading spinner seems too big and is not vertically centered.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

https://plnkr.co/edit/bw4dwl8tfXr8QUzMy5zV

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**
Latest (and earlier)

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
all

adityarb88:
@rongery: Thanks for raising this issue. I will mark this as a bug. Updated plnkr: https://plnkr.co/edit/I6wy9TUw75k1R5EcLp9S?p=preview because I think the default clarity plnkr was pasted above by mistake.

rongery:
Sorry about the bad paste job (I did create one and fork it but apparently
didn't get the right link), and thank you for correcting it. Other than the
text in the buttons, our plnkrs are pretty much identical.

Issue #1488, opened 2017-09-19
Spinner Button component (new)


gracesnoh:
### Description
![spinnerbtn](https://user-images.githubusercontent.com/8237984/30615409-bd346ee4-9d43-11e7-8c52-c2cb6808cd82.gif)
A spinner button component is a button component that generates a spinner inside of a button to indicate that a process is running. At the end of the process, a check icon is generated to indicate that the process is complete.

### Use cases under consideration
- page requests that require time to process
- buttons (all types & sizes)
- icon buttons
- forms: inline buttons

### Documentation
- Key features: add/remove spinners to all types of button components, apply screen reader text while process is running, and add/remove success state*
- We recommend using spinner buttons only for actions that require some time to process**
- We do not recommend using spinner buttons for links, navigating to different pages, or opening/closing a modal or wizard

*This feature may apply to situations where an action navigates to a new page but still requires time to process (i.e. logging in)

**For longer running processes, we recommend using a progress indicator or _circular progress indicator (not a currently existing component)_

#### States
![stateimg2](https://user-images.githubusercontent.com/8237984/30570399-f12d2294-9c95-11e7-9e44-b74f687bb2f6.png)

1. **Default state** acts just like a normal solid button would, darkening on hover.
2. **Loading state** transitions into a disabled state, hides the button label, and injects a rotating spinner.
3. **Success state** remains disabled, injects a check icon to indicate the success of an action, and returns back to default state.

##### Toggling Success State
- Success state can be toggled on/off according to the type of process that is taking place
- Ex) If a spinner button is applied to a “Log In” action, the success state should be toggled off because the process will result in navigating to a new page.

#### Usage
- For processes that require time to take place, such as creation or deletion actions
- For longer running processes, we recommend using a progress bar or a circular progress indicator (component in progress)

![do](https://user-images.githubusercontent.com/8237984/30570667-8a4147ca-9c97-11e7-85f9-8f7cdb49d6e1.png)
- Use spinner buttons for actions that update the content on the current page without navigating to a new one.

![dont](https://user-images.githubusercontent.com/8237984/30570679-a03234e0-9c97-11e7-9857-a04039e1da23.png)
- Use spinner buttons for actions that will navigate to a different page or open/close a modal or wizard.

#### Accessibility
- For accessibility, screen reader text is available for the Loading and Success button states
- Apply contextual text such as “Saving” and “Saved” to inform screen readers of what is happening

##### Solid
[...]

PR #1844, opened 2018-01-05
[UI] Fix spinners in small buttons

adityarb88:
Fixes: #1836 

Spinner in small buttons changed to 13px according to the specs mentioned here: https://github.com/vmware/clarity/issues/1488

smallbuttonspinner

Tested on: IE, Edge, Safari, Chrome, Firefox

Firefox alignment is a bit off but we can't do anything about it. Noticeable only if you stare at the screen very closely.


mathisscott:
Can we make sure this goes into 0.10?

adityarb88:
@mathisscott: Yes. I will make sure to port it to 0.10

Issue #1838, opened 2018-01-03
Conditional pagination issue


grreeenn:
**Select one ...**  (check one with "x")
```
[x ] bug 
[ ] feature request
[ ] enhancement
```
Not sure if it's a bug or an expected behavior..
I'm building a datagrid wrapper component, which serves many pages and gets the structure and data dynamically. As a part of a structure the component gets a boolean with a pagination switch - so the component may enable the pagination only in tables that should have it. So I've put the pagination-related code into ng-container with *ngIf, but...

### Expected behavior
Is to show the pagination just when the *ngIf condition is met.

### Actual behavior
The pages switch does not show up. Pagination variables, firstItem and lastItem shows up correctly, though.

### Reproduction of behavior
Put the pagination-related code into ng-container or any other element (even without a condition)
https://plnkr.co/edit/vXGb7eBFQlwL22LPSYBy?p=preview

### Workarounds tried
I've tried to put the condition on the clr-dg-pagination element itself - fails as firstItem and lastItem are undefined;
I've tried to put the condition on the clr-dg-footer itself - works fine until there're no hideable columns, when there are some - does the job but throws ExpressionChangedAfterItHasBeenCheckedError;
I've tried to put the whole footer into ng-container - it just fails silently

I have a feeling that I miss some rather simple workaround, but I still can't find it... Thank you for your help!

### Environment details

* **Angular version:** 5.1.2

* **Clarity version:** 0.10.19

* **OS and version:** Win 10

* **Browser:** Chrome  63.0.3239.84, FF 57.0.2

adityarb88:
@grreeenn: Not sure if I understand your question correctly. Are you looking for something like this? https://plnkr.co/edit/FzDlMGaRIN1GszssoKlF?p=preview

adityarb88:
@grreeenn: Please take a look at this plnkr: https://plnkr.co/edit/JLlJVp0YzRf48j7Hpcko?p=preview

This should work fine. I will check with @youdz if this is the solution that we were aiming for. If not, I will post a better one later on after further investigation.

adityarb88:
@grreeenn: We have fixed this bug in 0.11.1 & 0.10.23. Please refer to this file for more details: https://github.com/vmware/clarity/commit/df1ed6a99c72e292ede76f02f7edcfdeef3256e5#diff-c2816faca8eee9d8ca2c37224086529a

grreeenn:
[...]

PR #1854, opened 2018-01-10
[NG] Datagrid Pagination Bug Fix

adityarb88:
Fixes: #1838

1. Resets the page size when the pagination is destroyed.
2. Introduced a pagination description section.

pagination

Note: In case the datagrid height is restricted, it has to be restricted via a class. Adding a style attribute doesn't work because when you enable and then disable pagination, the height is reset when added through the style attribute. Works when you have a class though.

Tested on Chrome, IE, Edge, Safari, Firefox

### DIFF
--- a/src/app/datagrid/pagination-conditional/pagination-conditional.html
+++ b/src/app/datagrid/pagination-conditional/pagination-conditional.html
@@ -1,0 +1,40 @@
+<!--
+  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Conditional Pagination</h2>
+
+<p>
+    Test to conditional activate/deactivate pagination in a datagrid
+</p>
+
+<clr-datagrid [class.conditional-datagrid]="!paginationEnabled">
+    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column>Name</clr-dg-column>
+    <clr-dg-column>Creation date</clr-dg-column>
+    <clr-dg-column>Favorite color</clr-dg-column>
+
+    <clr-dg-row *clrDgItems="let user of users">
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>
+        <ng-container *ngIf="!paginationEnabled">
+            {{users.length}} users
+        </ng-container>
+        <clr-dg-pagination #pagination [clrDgPageSize]="5" *ngIf="paginationEnabled">
+            {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}} of {{pagination.totalItems}} users
+        </clr-dg-pagination>
+    </clr-dg-footer>
+</clr-datagrid>
+
+<button class="btn" type="button" (click)="toggle()">Toggle Pagination</button>
+
+<clr-example [clrCode]="example" clrLanguage="html"></clr-example>
--- a/src/app/datagrid/pagination-conditional/pagination-conditional.ts
+++ b/src/app/datagrid/pagination-conditional/pagination-conditional.ts
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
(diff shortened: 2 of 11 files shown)


src/app/datagrid/pagination-conditional/pagination-conditional.ts
```diff
@@ -0,0 +1,56 @@
...
+    providers: [Inventory],
+    templateUrl: "pagination-conditional.html",
+    styleUrls: ["../datagrid.demo.scss"]
+})
+export class DatagridConditionalPaginationDemo {
+    example = EXAMPLE;
+    users: User[];
+    paginationEnabled: boolean = false;
+
+    constructor(private inventory: Inventory, private cdr: ChangeDetectorRef) {
```
Shijir: I don't understand why `ChangeDetectorRef` is used in the demo app. Doesn't it mean users also will face the chocolate error and the need to use `ChangeDetectorRef`?

src/app/datagrid/pagination-conditional/pagination-conditional.html
```diff
@@ -0,0 +1,40 @@
...
+        <clr-dg-cell>{{user.id}}</clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>
+        <span *ngIf="pagination">
```
Shijir: Couldn't we include all the contents of `clr-dg-footer` in one `ng-container` and toggle it with single `paginationEnabled `?

adityarb88: But all the contents in `clr-dg-footer` are not hidden when the pagination is disabled. `users.length` is still visible.

src/app/datagrid/pagination-conditional/pagination-conditional.ts
adityarb88: I need @youdz to weigh in on this. I believe required because when the user toggles the pagination, the `pagination` variable is updated. Once that is updated, I get a chocolate error. Its happening because `ViewChildren`/`ViewChild` populates when the view changes and when the view changes the change detection has already begun.

src/app/datagrid/pagination-conditional/pagination-conditional.html
Shijir: I see. But we could still use only `paginationEnabled` in the place of `*ngIf="pagination"` and don't need to listen to `this.paginationViewChildren.changes` to set the `pagination` property to `null`.

adityarb88: I believe we do need to listen to the view children changes because thats exactly when i need to run cdr.detectChanges() to void the chocolate error. If the pagination is not enabled, i dont understand why we do not need to set the pagination property to null. it wont reflect whats on the view if we dont set it to null.

this will be more clear when we know if we need the change detector ref or not. i will wait for eudes to respond.

youdz: I might be wrong, but I believe doing something as simple as putting this span _after_ the pagination will fix your chocolate error. And that means you'd just use `paginationEnabled` like @Shijir suggested, and you can switch from the whole `@ViewChildren` + changes listener to a simple `@ViewChild` without any listeners.

adityarb88: @youdz I already tried that before submitting the PR :D Didn't work :-(

adityarb88: plnkr which replicates this: https://plnkr.co/edit/BMCrLP6AvFG6DWdhErB1?p=preview

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -69,6 +71,7 @@ export class ClrDatagridPagination implements OnDestroy, OnInit {
      */
     private _pageSubscription: Subscription;
     ngOnDestroy() {
+        this.pageSize = 0;
```
gnomeontherun: While this works, I personally would rather there be a method on the page service to reset so the intention here is more explicit.

adityarb88: yes. thats a good idea. i can make that change. will just add a method to the page service and set pageSize to 0

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -645,8 +645,18 @@
         border-radius: 0 0 $clr-global-borderradius $clr-global-borderradius;
 
         .pagination {
-            margin-left: 1.5rem;
-            height: calc(1.5rem - #{2 * $clr-default-borderwidth});
+            display: flex;
+            align-items: center;
```
Shijir: Have we tested it on different browsers?

adityarb88: image

Yep!

PR #1861, opened 2018-01-11
[NG] fix to remove datagrid column toggle hard coded strings

### LINKED ISSUE #1129: Datagrid localization for Hide/Show Columns
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior

It should be possible to set custom labels for the datagrid hide/show columns dialog box.

### Actual behavior

Currently, the datagrid hide/show dialog (https://vmware.github.io/clarity/documentation/datagrid/hide-show) uses fixed English labels (see for instance "Show Columns" in https://github.com/vmware/clarity/blob/master/src/clarity-angular/data/datagrid/datagrid-column-toggle.ts).

### Reproduction of behavior

There is no way to specify custom labels.

### Environment details

* **Angular version:** 4.0.X

* **Clarity version:** 0.9.10

* **OS and version:** all

* **Browser:** all

gnomeontherun:
Does not break compatibility because it will fall back to existing implementation. To use, just add the `clr-dg-column-toggle` component into the footer, and then project the `clr-dg-column-toggle-title` and `clr-dg-column-toggle-button` components into the footer.

This can be ported to 0.10 without a breaking change.

Fixes #1129

### DIFF
--- a/src/app/datagrid/hide-show-columns/hide-show.html
+++ b/src/app/datagrid/hide-show-columns/hide-show.html
@@ -82,3 +82,73 @@
         <clr-dg-pagination #pagination [clrDgPageSize]="currentPageSize"></clr-dg-pagination>
     </clr-dg-footer>
 </clr-datagrid>
+
+<h2>Hide-show columns demo with custom buttons</h2>
+
+<clr-datagrid>
+    <clr-dg-column>
+        <ng-container *clrDgHideableColumn>
+            User ID
+        </ng-container>
+    </clr-dg-column>
+    <clr-dg-column>
+        <!--Name-->
+        <ng-container *clrDgHideableColumn>
+            Name
+        </ng-container>
+    </clr-dg-column>
+    <clr-dg-column>
+        <ng-container *clrDgHideableColumn>
+            Creation date
+        </ng-container>
+    </clr-dg-column>
+    <clr-dg-column>
+        <ng-container *clrDgHideableColumn>
+            Pokemon
+        </ng-container>
+    </clr-dg-column>
+    <clr-dg-column>
+        <ng-container *clrDgHideableColumn>
+            Favorite color
+        </ng-container>
+    </clr-dg-column>
+
+    <clr-dg-placeholder>No users found</clr-dg-placeholder>
+
+    <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
+        <clr-dg-cell>
+            {{user.id}}
+        </clr-dg-cell>
+        <clr-dg-cell>{{user.name}}</clr-dg-cell>
+        <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+        <clr-dg-cell>
+            {{user.pokemon.name}}
+            <clr-signpost>
+                <button
+                    type="button"
+                    class="signpost-action btn btn-small btn-link"
+                    [class.active]="open"
+                    clrSignpostTrigger>
+                    <clr-icon shape="help-info"></clr-icon>
+                </button>
+                <clr-signpost-content *clrIfOpen [clrPosition]="'top-middle'">
+                    The pokemon is strong.
+                </clr-signpost-content>
+            </clr-signpost>
+        </clr-dg-cell>
+        <clr-dg-cell>
+            <span class="color-square" [style.backgroundColor]="user.color"></span>
+        </clr-dg-cell>
+    </clr-dg-row>
+
+    <clr-dg-footer>
+        <clr-dg-column-toggle>
+            <clr-dg-column-toggle-title>Kolumne Herauschauen!</clr-dg-column-toggle-title>
+            <clr-dg-column-toggle-button clrType="selectAll">Alle auswählen!</clr-dg-column-toggle-button>
+            <clr-dg-column-toggle-button clrType="ok"><clr-icon shape="check"></clr-icon></clr-dg-column-toggle-button>
+        </clr-dg-column-toggle>
+        {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}}
+        of {{pagination.totalItems}} users
+        <clr-dg-pagination #pagination [clrDgPageSize]="currentPageSize"></clr-dg-pagination>
+    </clr-dg-footer>
+</clr-datagrid>
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component, QueryList, TemplateRef, ViewChildren} from "@angular/core";
+
+import {ClrDatagridColumnToggle} from "./datagrid-column-toggle";
+import {ClrDatagridColumnToggleButton} from "./datagrid-column-toggle-button";
+import {DatagridHideableColumnModel} from "./datagrid-hideable-column.model";
+import {TestContext} from "./helpers.spec";
+import {ColumnToggleButtonsService} from "./providers/column-toggle-buttons.service";
+import {HideableColumnService} from "./providers/hideable-column.service";
+
+export default function(): void {
+    describe("Datagrid Column Toggle Button component", function() {
+        describe("Typescript API", function() {
+            let columnToggleButtons: ColumnToggleButtonsService;
+            let component: ClrDatagridColumnToggleButton;
+
+            beforeEach(function() {
+                columnToggleButtons = new ColumnToggleButtonsService();
+                component = new ClrDatagridColumnToggleButton(columnToggleButtons);
+            });
+
+            it("knows if the button isOk()", function() {
+                component.clrType = "ok";
+                expect(component.isOk()).toBe(true);
+                component.clrType = "selectAll";
+                expect(component.isOk()).toBe(false);
+            });
+
+            it("gets the correct classes", function() {
+                component.clrType = "ok";
+                expect(component.getClasses()).toEqual("btn btn-primary");
+                component.clrType = "selectAll";
+                expect(component.getClasses()).toEqual("btn btn-sm btn-link p6 text-uppercase");
+            });
+
+            it("calls the click method", function() {
+                component.clrType = "ok";
+                spyOn(columnToggleButtons, "buttonClicked");
+                component.click();
+                expect(columnToggleButtons.buttonClicked).toHaveBeenCalledWith(component.clrType);
+            });
+        });
+
+        describe("View", function() {
+            // Until we can properly type "this"
+            let context: TestContext<ClrDatagridColumnToggleButton, ButtonTest>;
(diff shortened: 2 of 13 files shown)


src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -38,8 +39,18 @@ import {DatagridRenderOrganizer} from "./render/render-organizer";
...
+        FiltersProvider,
+        Page,
+        Items,
+        DatagridRenderOrganizer,
+        RowActionService,
+        ExpandableRowsCount,
+        HideableColumnService,
+        StateDebouncer,
+        StateProvider,
+        ColumnToggleButtonsService,
```
mathisscott: trailing comma? do we care?

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    public buttonClicked(type: ColumnToggleButtons): void {
+        switch (type) {
+            case "ok":
+                this._okButtonClicked.next();
+                break;
+            case "selectAll":
+                this._selectAllButtonClicked.next();
+                break;
+            default:
```
mathisscott: I suspect the answer is "Yes, because JavaScript" but if we have a custom type that requires two strings (either "ok" or "selectAll") And we create a case statement that covers both of those types, do we still need the default fall through?

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
```diff
@@ -0,0 +1,46 @@
...
+            [ngClass]="getClasses()"
+            type="button">
+            <ng-content></ng-content>
+        </button>
+    `,
+    host: {"[class.action-right]": "isOk()", "[style.display]": "block"}
+})
+
+export class ClrDatagridColumnToggleButton {
+    @Input() type: ColumnToggleButtons;
```
adityarb88: Don't we want to name this Input like we do for other ones? `clrToggleButtonType` or something like that?

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
adityarb88: I believe we need it because tslint requires us to have it.

gnomeontherun: Yeah, tooling doesn't like no `default` case, and I _could_ have it use one of the two as default but then it makes sense to just use if/else.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
gnomeontherun: Its not meant to be part of our public API, we don't use the prefix.

src/clr-angular/data/datagrid/datagrid.ts
gnomeontherun: I've come to love them myself. Later I can add a new item to this list without having a diff on the line above just to add a `,`. Our linting tools do not seem to mind either way though.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
gnomeontherun: That said, it is a type, and we don't have a specific naming convention for types (we do for interfaces).

src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -77,6 +77,11 @@ <h2>Datagrid: Hide-show columns demo</h2>
     </clr-dg-row>
 
     <clr-dg-footer>
+        <clr-dg-column-toggle>
+            <clr-dg-column-toggle-title>Show Columns!!!</clr-dg-column-toggle-title>
+            <clr-dg-column-toggle-button [type]="'selectAll'">Select All!!!</clr-dg-column-toggle-button>
+            <clr-dg-column-toggle-button [type]="'ok'">OK!!!</clr-dg-column-toggle-button>
```
youdz: You had an opportunity to put German in our sample app! Too bad. I like 2-lines buttons. 😛 

(Everyone knows an "OK" button becomes "Ich stimme dieser Entscheidung zu" in German. :trollface: )

src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -77,6 +77,11 @@ <h2>Datagrid: Hide-show columns demo</h2>
     </clr-dg-row>
 
     <clr-dg-footer>
+        <clr-dg-column-toggle>
+            <clr-dg-column-toggle-title>Show Columns!!!</clr-dg-column-toggle-title>
+            <clr-dg-column-toggle-button [type]="'selectAll'">Select All!!!</clr-dg-column-toggle-button>
```
youdz: You probably know this, but when doing something as static as this with the button types, you can just do `type="selectAll"` instead of `[type]="'selectAll'"`. I'm pointing this because it would be more user-friendly in the documentation.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+                expect(component.isOk()).toBe(true);
+                component.type = "selectAll";
+                expect(component.isOk()).toBe(false);
+            });
+
+            it("gets the correct classes", function() {
+                component.type = "ok";
+                expect(component.getClasses()).toEqual("btn btn-primary");
+                component.type = "selectAll";
+                expect(component.getClasses()).toEqual("btn btn-sm btn-link p6 text-uppercase");
```
youdz: Nitpick, but this feels like a view test. You're pretty much testing that the correct CSS classes are added to the host.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            expect(service.okButtonClicked.subscribe).toBeDefined();
+            expect(service.selectAllButtonClicked.subscribe).toBeDefined();
+        });
+
+        it("should emit clicks", function(done) {
+            let calls = 0;
+            function handle() {
+                calls++;
+                if (calls === 2) {
+                    expect(calls).toEqual(2);
```
youdz: Wait, what? If `calls === 2`, expect `calls` to equal `2`? You're not testing anything here. I believe you just want the `expect` outside of the function, right after 
```typescript
service.buttonClicked("ok");
service.buttonClicked("selectAll");
```
Unless I'm missing something and not all pieces are synchronous.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type ColumnToggleButtons = "ok"|"selectAll";
+
+@Injectable()
+export class ColumnToggleButtonsService {
+    buttons: TemplateRef<any> = null;
+    selectAllDisabled: boolean = false;
+
+    private _okButtonClicked = new Subject<any>();
+    public get okButtonClicked(): Observable<any> {
```
youdz: I can't find where these two Observables are used. Do the buttons actually perform their respective actions? Based on the code in the PR, I don't see how that would be possible.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
gnomeontherun: Its on the fence, because I have a class method that returns these values. Its implicitly covered in the view "it has a button" test by checking the className there, but felt like this was useful to check that we didn't change the classes here accidentally.

src/app/datagrid/hide-show-columns/hide-show.html
gnomeontherun: Yup, at one point they were variables and I never flipped it to strings. Fixed up.

src/app/datagrid/hide-show-columns/hide-show.html
gnomeontherun: I added German, and an icon for fun.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.spec.ts
gnomeontherun: The `buttonClicked()` calls `next()` on an observable, so I'm testing that both clicks emitted correctly by listening and counting the times the handler was called. This will fail if one of the emits fails to be handled as a timeout. Its not particularly elegant, I could use a spy instead except the internals are private.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
gnomeontherun: Botched the original PR, pushing the right code which _uses_ the observables in the trigger component to call the corresponding methods.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
gnomeontherun: I just reread this and realize you meant the input should be `@Input('clrType') type: ColumnToggleButtons;`, fixing.

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -271,6 +275,30 @@ export default function(): void {
...
+                iconBtn.click();
+                context.detectChanges();
+                expect(context.clarityDirective.open).toBe(true);
+
+                const buttons = context.clarityElement.querySelectorAll(".switch-footer button");
+                expect(buttons[0].innerText).toEqual("Select All!".toUpperCase());
+                expect(buttons[1].innerText).toEqual("OK!".toUpperCase());
+                const title = context.clarityElement.querySelector(".switch-header");
+                expect(title.innerHTML).toContain("Custom Title");
+            });
```
youdz: We are missing a unit test for `buttonClicked` emit -> select all / toggle popover.

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -271,6 +275,30 @@ export default function(): void {
...
+            it("projects custom buttons and title", function() {
+                const iconBtn = context.clarityElement.querySelector(".column-switch-wrapper > button");
+                expect(context.clarityDirective.open).toBe(false);
+                iconBtn.click();
+                context.detectChanges();
+                expect(context.clarityDirective.open).toBe(true);
+
+                const buttons = context.clarityElement.querySelectorAll(".switch-footer button");
+                expect(buttons[0].innerText).toEqual("Select All!".toUpperCase());
+                expect(buttons[1].innerText).toEqual("OK!".toUpperCase());
```
mathisscott: Do we want a unit test to validate the defaults are coming through?

gnomeontherun: Added a test.

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
gnomeontherun: Added a test.

PR #1868, opened 2018-01-18
[ICONS] update the templates of icons in DOM if they are updated in the

Shijir:
API

- Backporting #1777 to v0.10


src/clarity-icons/utils/shape-template-observer.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
```
hippee-lee: Minor issue with the dates for 2017.

Issue #1866, opened 2018-01-17
Add sass variable that will bypass the base64 fonts in a generated theme


KANekT:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
How change Default Font from Metropolis?

### Reproduction of behavior
`// Pre variable and theme dependencies
@import "../../node_modules/@clr/ui/src/utils/dependencies.clarity";

// Theme (Variable overrides)
$styledInputs: inputs("text", "password", "number", "email", "url", "tel", "date", "time", "datetime-local", "search");
@import "fonts";
$clr-font: 'Exo 2', 'Helvetica Neue', Arial, sans-serif !default; //Default font stack for Clarity

// Clarity component code (consumes the variables and theme overrides)
@import "../../node_modules/@clr/ui/src/utils/components.clarity";
`

But is not change font-family in output css

### Environment details

* **Angular version:** 5.x.x

* **Clarity version:** 0.11.x

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

mathisscott:
You shouldn't use the `!default` SASS flag like that.

`!default` only sets the variable in question if it _doesn't already have a value_.

Try the following...

```
// Pre variable and theme dependencies
@import "../../node_modules//src/utils/dependencies.clarity";

// Theme (Variable overrides)
$styledInputs: inputs("text", "password", "number", "email", "url", "tel", "date", "time", "datetime-local", "search");
$clr-font: 'Exo 2', 'Helvetica Neue', Arial, sans-serif;
@import "fonts";
```

...and see if that works.

KANekT:
No it not working for me
Example in out css:
`    .sidenav .nav-group label {
      color: #313131;
      font-size: 0.58333rem;
      font-weight: 500;
      line-height: 1rem;
      font-family: Metropolis, "Avenir Next", "Helvetica Neue", Arial, sans-serif;
      letter-spacing: normal; }
`
i need 
`    .sidenav .nav-group label {
      color: #313131;
      font-size: 0.58333rem;
      font-weight: 500;
      line-height: 1rem;
      font-family: 'Exo 2', 'Helvetica Neue', Arial, sans-serif;
      letter-spacing: normal; }
`

[...]

PR #1893, opened 2018-01-30
[UI] introducing font override variable to SASS

mathisscott:
• added $clr-fontSkipBase64
• when set to true, $clr-fontSkipBase64 will keep the Metropolis base64 font out of the generated CSS
• $clr-fontSkipBase64 is false by default – meaning the default is for the base64 font to be included

Closes: #1866

Tested in:
✔︎ Chrome


paul-mcgrath:
Hi @mathisscott, I checked the Clarity Typography documentation for this change, but I found it difficult to follow and was wondering if you have a clearer example of how to accomplish the following:

1. Set $clr-fontSkipBase64 in an Angular app
2. Set a new font-family for Clarity in the same Angular app

Any advice would be incredibly helpful, thank you!

PR #1901, opened 2018-01-31
[NG] Host wrapping hack

youdz:
This commit provides an internal utility for a directive to wrap its host in a container element. For information on how to use it, check out the spec file.

### CAUTION: This feature is internal only and should never be exported or be made part of the public API!

Signed-off-by: Eudes Petonnet-Vincent <epetonnetvince@vmware.com>

### DIFF
--- a/src/clr-angular/utils/host-wrapping/dynamic-wrapper.ts
+++ b/src/clr-angular/utils/host-wrapping/dynamic-wrapper.ts
@@ -1,0 +1,7 @@
+/**
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export interface DynamicWrapper { _dynamic: boolean; }
--- a/src/clr-angular/utils/host-wrapping/empty-anchor.ts
+++ b/src/clr-angular/utils/host-wrapping/empty-anchor.ts
@@ -1,0 +1,12 @@
+/**
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Component} from "@angular/core";
+
+@Component({
+    template: "",
+})
+export class EmptyAnchor {}
--- a/src/clr-angular/utils/host-wrapping/host-wrapper.ts
+++ b/src/clr-angular/utils/host-wrapping/host-wrapper.ts
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {
+    ComponentFactory,
+    ComponentFactoryResolver,
+    ElementRef,
+    InjectionToken,
+    Injector,
+    Type,
+    ViewContainerRef
+} from "@angular/core";
+
+import {DynamicWrapper} from "./dynamic-wrapper";
+import {EmptyAnchor} from "./empty-anchor";
+
+export class HostWrapper<W extends DynamicWrapper> implements Injector {
+    constructor(containerType: Type<W>, vcr: ViewContainerRef) {
+        this.injector = vcr.injector;
+        // If the host is already wrapped, we don't do anything
+        if (!this.injector.get(containerType, null)) {
+            const cfr = this.injector.get(ComponentFactoryResolver);
+            const el = this.injector.get(ElementRef);
+
+            // We need a new anchor, since we're projecting the current one.
+            vcr.createComponent(cfr.resolveComponentFactory(EmptyAnchor));
+            const factory: ComponentFactory<W> = cfr.resolveComponentFactory(containerType);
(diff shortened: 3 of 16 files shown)


src/clr-angular/utils/host-wrapping/dynamic-wrapper.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
```
adityarb88: Copyright error :P

adityarb88: Almost all your files need to update the copyright.

src/clr-angular/utils/host-wrapping/host-wrapper.ts
```diff
@@ -0,0 +1,52 @@
...
+            const factory: ComponentFactory<W> = cfr.resolveComponentFactory(containerType);
+            // We're assuming only one projection slot, but in more complex cases we might want to provide
+            // a different array of projected elements.
+            const containerRef = vcr.createComponent(factory, undefined, undefined, [[el.nativeElement]]);
+            // We can now remove the useless anchor
+            vcr.remove(0);
+
+            // We note that the container was dynamically created
+            // (no idea why we have to force cast, typescript is weird since I updated)
+            containerRef.instance._dynamic = true;
```
adityarb88: I don't understand this comment. What do you mean that you had to force cast this? What would you have rather done if TS wasn't acting weird?

src/clr-angular/utils/host-wrapping/host-wrapping.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+
+interface TestContext<T extends WrappingTest|ExplicitTest|NgIfTest> {
+    fixture: ComponentFixture<T>;
+    wrapper: WrapperComponent;
+    wrapped: WrappedDirective;
+}
+
+fdescribe("Host wrapping", function() {
```
adityarb88: Wanted to remind to change this back to `describe`

src/clr-angular/utils/host-wrapping/host-wrapper.ts
adityarb88: What's the point of noting that the wrapper was dynamically created? We are not using that property anywhere. Am I missing something?

src/clr-angular/utils/host-wrapping/empty-anchor.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Component} from "@angular/core";
+
+@Component({
+    selector: "clr-empty-anchor",
```
gnomeontherun: Technically this is unnecessary, so would be good to remove it unless we actually want to use it.

src/clr-angular/utils/host-wrapping/host-wrapping.module.ts
```diff
@@ -0,0 +1,25 @@
...
+
+
+/**
+ * Internal module, please do not export!
+ */
+// We don't use animations, but the BrowserAnimationsModule introduces inconsistent behavior absolutely everywhere
+// depending on whether it's imported in some distant modue or not. So to be consistent, we include it everywhere.
+// I'm not happy about this hack.
+@NgModule({
+    imports: [BrowserAnimationsModule],
```
gnomeontherun: Did you try the `NoopAnimationsModule`?

src/clr-angular/utils/host-wrapping/host-wrapper.ts
gnomeontherun: The `_dynamic` property is meant to be consumed in the template of a container implementation so it can add a label if it is missing. Its a quirk specific to radio and checkboxes at least.

src/clr-angular/utils/host-wrapping/host-wrapping.spec.ts
youdz: Oops, my bad. Thanks. ❤️

src/clr-angular/utils/host-wrapping/host-wrapper.ts
youdz: Comment from a previous version, I was able to drop the force cast but forgot to remove the comment. Thanks for catching that.

And as @gnomeontherun pointed out, it provides the container with a way to perform specific work when generated dynamically. For radios and checkboxes, it can create an empty label. For email inputs, it could add a default error if the value isn't an email. There could be more cases, and it didn't cost much to put it there.

youdz:
@gnomeontherun: You're right, I should post the form control utility too. It's just that I didn't have unit tests for it yet. I'll update the PR tomorrow with it.

src/clr-angular/utils/host-wrapping/host-wrapping.module.ts
youdz: It would include the `NoopAnimationsModule` in actual production applications, which is kind of messed up when they already import the `BrowserAnimationsModule` too. So I went with the heavier, but real-life one. I have zero confidence in non-prod modules from the animation part, they simply don't care if they break everything.

PR #1913, opened 2018-02-02
[CORE] error when tests have `fdescribe` or `fit`

gnomeontherun:
This is a simple check to have our tests fail if we have focused tests, so that its obvious to fix before merging.


package.json
```diff
@@ -9,7 +9,7 @@
         "build": "npm-run-all clean build:angular webpack build:icons",
         "pretest": "webpack",
         "test": "karma start ./karma.conf.js",
```
hippee-lee: Does it make sense to add it here also and execute with npm-run-all?

gnomeontherun: You don't want this to run locally, so adding it here would fail the test command locally.

Issue #1958, opened 2018-02-06
Enable SauceLabs with TravisCI


gnomeontherun:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
We want to run our test suites against SauceLabs, using Chrome and Firefox.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

PR #1960, opened 2018-02-06
[NG] get unit tests working with Firefox

gnomeontherun:
This enables running the tests locally with Firefox as well as Chrome. There are a few tests that are disabled for Firefox only, due to differences in the way browsers handle some details like focusing.

This does not enable the Firefox to run by default, because it would fail on Travis.

Part of #1958


tests/tests.helpers.ts
```diff
@@ -0,0 +1,13 @@
+import * as Browser from "detect-browser";
+
+export const itIgnore = (browsers: string[], should: string, test: any, focus?: boolean) => {
+  if (browsers.length && browsers.indexOf(Browser.name) >= 0) {
+    return;
```
youdz: Would returning `xit(should, test)` make more sense? It would show the skipped tests in the report.

gnomeontherun: Actually yes, I have a second PR following that makes that change when I worked on Safari.

PR #1968, opened 2018-02-07
[NG] Preserve selection on items refresh for client-side datagrids

### LINKED ISSUE #1593: datagrid.selected is reassigned with an empty array
```
[x] bug
[ ] feature request
[ ] enhancement
```
https://plnkr.co/edit/WUNt57MsppVCqK18VtH1?p=preview
### Expected behavior
After datagrid.selected is assigned - the selection if populated on screen

### Actual behavior
The value is assigned and grid is redrawn, but after that this code from datagrid/selection.js seems to rewrite it
```
                // TODO: Discussed this with Eudes and this is fine for now.
                // But we need to figure out a different pattern for the
                // child triggering the parent change detection problem.
                // Using setTimeout for now to fix this.
                setTimeout(function () {
                    _this.current = leftOver;
                }, 0);
```

### Reproduction of behavior
https://plnkr.co/edit/WUNt57MsppVCqK18VtH1
Load plunker,
Populate some selection in both grids
Click refresh button
(the selection if grid1 is preserved, the multi-selection grid - loses selection)
### Environment details

* **Angular version:** 4.3.1

* **Clarity version:** 0.10.7 and 0.10.9

* **OS and version:** latest osx

* **Browser:** all

youdz:
Fixes several bugs at once where we used the wrong variables (or didn't use them at all) to try and preserve selection, which meant we cleared it every time.
Also fixes the corresponding spec, which were completely broken for two reasons: a nested `beforeEach()` that didn't do anything, and the fact that we tested synchronously that an asynchronous operation didn't change anything. In other words, we were testing that code that hasn't run yet did not have any effects...

Fixes #1593.

### DIFF
--- a/src/clr-angular/data/datagrid/providers/items.ts
+++ b/src/clr-angular/data/datagrid/providers/items.ts
@@ -79,14 +79,16 @@ export class Items {
      * List of all items in the datagrid
      */
     private _all: any[];
+    public get all() {
+        return this._all;
+    }
     public set all(items: any[]) {
+        this._all = items;
+        this.emitAllChanges(items);
         if (this.smart) {
-            this._all = items;
-            this.emitAllChanges(items);
             this._filterItems();
         } else {
             this._displayed = items;
-            this.emitAllChanges(items);
             this.emitChange();
         }
     }
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -3,6 +3,7 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import {TrackByFunction} from "@angular/core";
 import {fakeAsync, tick} from "@angular/core/testing";
 import {Subject} from "rxjs/Subject";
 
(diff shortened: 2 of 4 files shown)


src/clr-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -250,81 +251,79 @@ export default function(): void {
...
-                }
-
                 beforeEach(function() {
                     selectionInstance.selectionType = SelectionType.Multi;
                 });
 
-                it("should support no trackBy", function() {
-                    testMultiSelection();
+                it("should preserve selection on page change", function() {
+                    selectionInstance.setSelected(items[2], true);
```
adityarb88: Quick question about assuming defaults while testing: Should we have explicitly checked over here the initial current page value and that the item is actually not on the page after the page change? I know that the page size is 3 and you moved to page 3 so you are viewing items 6-9 but still would that additional `expect` be useful?

src/clr-angular/data/datagrid/providers/items.ts
```diff
@@ -79,12 +79,16 @@ export class Items {
      * List of all items in the datagrid
      */
     private _all: any[];
+    public get all() {
+        return this._all;
+    }
     public set all(items: any[]) {
         if (this.smart) {
             this._all = items;
```
gnomeontherun: Can reduce code here, though order of assignment might be important.

```
this._all = items;
this.emitAllChanges(items);
if (this.smart) {
  this._filterItems();
} else {
  this._displayed = items;
  this.emitChange();
}

src/clr-angular/data/datagrid/providers/selection.spec.ts
youdz: Isn't that what we already test in `page.spec.ts`?

adityarb88: cool. thanks!

KeshavRathi:
Is there a way to disable this default feature?

PR #1971, opened 2018-02-08
fix: [CORE] tslint config usage

hippee-lee:
### Highlights
- Updates npm script to use -p option and the tslint.json at root of project
- Fixes code and addresses tslint issues in codebase

### Notes
1. this ignores the ks-app (I was unable to get extends to work in the tslint.json for ks-app and there are a lot of no-unused-var errors b/c of the components testing aot/clarity api in the ks-app code)
2. I'm not sure if we still need  this npm script: `tslint:fix` but we don't use it in the project
3. constructors that have private removed: these are cases where we used it as an argument, not a property, it tslint fails b/c it expects anything declared private to be used somewhere with `this.thing`. [read more](https://github.com/palantir/tslint/issues/3150#issuecomment-324274921)
4. `no-use-before-declare` was removed: see the [tslint docs](https://palantir.github.io/tslint/rules/no-use-before-declare/)

### Warning
**If you use a precommit git hook to lint your changes you will need to update it.**
Here is mine (Note the `-p .` on the first line of the if block): 
```
# Run tslint on all staged ts files
echo "Running Linter"
for file in $(git diff --cached --name-only | grep -E '\.ts$')
do
    if [[ -f $file ]]; then
        git show ":$file" | node_modules/.bin/tslint -p . -c tslint.json "$file" # we only want to lint the staged changes, not any un-staged changes
        if [ $? -ne 0 ]; then
        echo "TSLint failed on staged file '$file'. Please check your code and try again. You can run TSLint manually via gulp tslint:check."
        exit 1 # exit with failure status
        fi
    fi
done
```


src/app/progress-bars/progbar-example.ts
```diff
@@ -8,14 +8,16 @@ let id: number = 0;
...
     value: number = 0;
     intervalTimeInMs: number = 100;
+    title: string;
+    isLabeled: boolean;
 
-    constructor(private label: string = "demo", private title: string = "Progress Bar",
-                private isLabeled: boolean = false) {}
+    constructor(private label: string = "demo", title: string = "Progress Bar", isLabeled: boolean = false) {
+        this.title = title;
+        this.isLabeled = isLabeled;
```
youdz: This one is surprising, but doesn't really matter. I would expect constructor's `private` to work here.

src/clr-angular/layout/vertical-nav/vertical-nav-link.ts
```diff
@@ -19,10 +19,12 @@ import {VerticalNavGroupService} from "./providers/vertical-nav-group.service";
 export class ClrVerticalNavLink {
     constructor(@Optional() private _navGroupService: VerticalNavGroupService) {}
 
+    /* tslint:disable */
     @HostListener("click")
     private expandParentNavGroup(): void {
```
youdz: Is TSLint complaining because this private method isn't used anywhere, and doesn't know about `@HostListener`? This method could very well be public, and on top of that I'm not 100% sure that Angular is happy about private `@HostListener`s (maybe the methods are called from another class).

hippee-lee: You are right, tslint did not like it being private. Will update.

PR #1979, opened 2018-02-09
[NG] added datagrid caption

jkleczkowski:
I've added a caption feature to datagrid. [crlCaption]="model or raw valu"

Signed-off-by: Jacek Kleczkowski <jkleczkowski@gmail.com>

### DIFF
--- a/src/app/datagrid/kitchen-sink/kitchen-sink.html
+++ b/src/app/datagrid/kitchen-sink/kitchen-sink.html
@@ -225,6 +225,32 @@
 
 </div>
 
+<h3>Table Caption</h3>
+<div id="table-caption">
+    <clr-datagrid [clrCaption]="'Table Caption'">
+        <clr-dg-column>User ID</clr-dg-column>
+        <clr-dg-column>Name</clr-dg-column>
+        <clr-dg-column>Creation date</clr-dg-column>
+        <clr-dg-column>Favorite color</clr-dg-column>
+
+        <clr-dg-placeholder>We couldn't find any users!</clr-dg-placeholder>
+        <clr-dg-row *ngFor="let user of []"></clr-dg-row>
+
+        <clr-dg-footer>{{[].length}} users</clr-dg-footer>
+    </clr-datagrid>
+
+    <clr-datagrid>
+        <clr-dg-column>User ID</clr-dg-column>
+        <clr-dg-column>Name</clr-dg-column>
+        <clr-dg-column>Creation date</clr-dg-column>
+        <clr-dg-column>Favorite color</clr-dg-column>
+
+        <clr-dg-row *ngFor="let user of []"></clr-dg-row>
+
+        <clr-dg-footer>{{[].length}} users</clr-dg-footer>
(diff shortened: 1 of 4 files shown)


youdz:
Sadly, we can't accept a PR where displayed text is given as a component input instead of being projected, because this limits reusability and customizability way too much, as mentioned in our [coding guidelines](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#html-templates).

On top of that, could you explain why you think this feature is needed for the datagrid? We can't accept everyone's features at least a little explanation of why the feature should be added. Whether it's describing your use case, or showing why a simple title before or after the datagrid doesn't work, ...
And finally, we'll need someone from our UX side to check out the design you're trying to submit, they will certainly have specific requirements on the positioning and style of the caption. If you could post a screenshot of what the caption looks like with your current code, this would help them a lot.

Thank you for your contribution.

jkleczkowski:
Hi.
Thank you  for response.
I'm just starting my adventure with Angular, so please forgive me for my inadequate knowledge of the subject.
I would be grateful if you explained to me or pointed to the documentation of the difference between 'component input' and 'being projected'

And now I will show you how I wanted to achieve:
(Intentionally I pointed lement 'class =' datagrid-overlay-wrapper " ')

1. Raw version

2. Your suggesstion:

3. My suggestion


When You add <caption></caption> to regular table it will looks like that:


<table class="table">
    <caption>Caption in regulart HTML table</caption>
        <thead>
            <tr>
                <th>Decimal</th>
                <th>Hexadecimal</th>
                <th>Binary</th>
                <th>Roman Numeral</th>
            </tr>
        </thead>
        <tbody>
            <tr>
                <td>1</td><td>1</td><td>1</td><td>I</td>
            </tr>
            <tr>
                <td>5</td><td>5</td><td>101</td><td>V</td>
            </tr>
            <tr>
                <td>10</td><td>A</td><td>1010</td><td>X</td>
            </tr>
            <tr>
                <td>15</td><td>F</td><td>1111</td><td>XV</td>
            </tr>
        </tbody>
    </table>    

Element before Datagrid could be unpredictible so I think that element inside main div of datagrid can be more manageable.

Please consider feature like this of course written in the proper way.

Thank you for your time.

Regards
Jacek Kleczkowski

youdz:
This is the difference between component input and projection: https://stackblitz.com/edit/content-projection-vs-input?file=app%2Fapp.component.html

As you can see, not only does projection allow better customization of the content by the user (interpolation, links, icons or anything else that would require HTML), it is also much more readable when dealing with longer texts.

mathisscott:
@reddolan : What is our UX/design recommendation for table captions?

adityarb88:
@reddolan: ^

reddolan:
I don't exactly understand what is trying to be accomplished here. The Stackblitz doesn't show what the PR seems to be describing. Is it possible to get something visual to look at here so that I can review?

reddolan:
We recommend that you use normal header tags `<h2>` or `<h3>`. We won't be including this in Clarity.

PR #1994, opened 2018-02-14
[WIP][NG] drag and drop directives - not tested yet

Shijir:


youdz:
Drag and drop is public, you have it in the ClarityModule but you're not exporting anything from the main Clarity `index.ts`.

Shijir:
@youdz yeah 😬 realized that when I was setting up the test. So the fix will be included in the commit that covers setting up unit tests.

src/app/drag-and-drop/drag-and-drop.demo.html
```diff
@@ -0,0 +1,179 @@
...
+
+<div class="clr-example">
+    <h4>Basic Draggable</h4>
+    <div class="row">
+        <div class="col-xs-12">
+            <p>
+                Use <code class="clr-code">clrDragAndDropGroup</code> directive to make particular draggables interact
+                with particular droppables.
+            </p>
+            <button class="btn btn-primary" clrDraggable clrDragAndDropGroup="dnd-1">Draggable Button</button>
```
adityarb88: Do you think we can just name this `clrDropGroup` at both places? with `clrDraggable` and `clrDroppable`?

src/app/drag-and-drop/drag-and-drop.demo.html
```diff
@@ -0,0 +1,179 @@
...
+                        ...
+                    </p>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+
+<div class="clr-example">
+    <h4>Custom Ghosts</h4>
```
adityarb88: Can we split this big file into different routes?

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
```diff
@@ -0,0 +1,105 @@
...
+import {Injectable, NgZone, Renderer2} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+
+import {DragPosition} from "../interfaces/client-position";
+import {Draggable} from "../interfaces/draggable";
+
+@Injectable()
+export class DragDispatcher {
+    private _listeners: Function[];
```
adityarb88: Replace `Function` by the actual type.

src/clr-angular/utils/drag-and-drop/interfaces/draggable.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+export interface Draggable {
```
adityarb88: Could you please comment what the interface does as well as what the properties are responsible for?

src/clr-angular/utils/drag-and-drop/interfaces/client-position.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+export interface DragPosition {
```
adityarb88: Why is the file name different?

src/clr-angular/utils/drag-and-drop/interfaces/client-position.ts
```diff
@@ -0,0 +1,9 @@
...
+    x: number;
```
adityarb88: Is this a coordinate or the distance dragged? A comment would be helpful here.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {DragPosition} from "./client-position";
+import {Draggable} from "./draggable";
+
+export interface DraggableEvent {
+    draggable: Draggable;
+    dragPosition: DragPosition;
+    dragAndDropGroup: string;
```
adityarb88: If the draggable belongs to more than one group, then this could be either a string or an array, right?

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {DragPosition} from "./client-position";
+import {Draggable} from "./draggable";
+
+export interface DraggableEvent {
```
adityarb88: Same thing that we mentioned above. Interfaces typically need comments because we dont have any implementation to understand what each property does.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
```diff
@@ -0,0 +1,105 @@
...
+        return this._onDragMove.asObservable();
+    }
+
+    get onDragEnd(): Observable<MouseEvent|TouchEvent> {
+        return this._onDragEnd.asObservable();
+    }
+
+    constructor(private _ngZone: NgZone, private _renderer: Renderer2) {}
+
+    initialize() {
```
adityarb88: Initialize should receive as arguments everything the class needs to setup. In particular, just give the `Draggable` object here and make the public `draggable` property private.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.draggable.handle = this.draggable.self;
+        }
+
+        this._listeners = [
+            this.customDragEvent(this.draggable.handle, "mousedown", "mousemove", "mouseup"),
+            this.customDragEvent(this.draggable.handle, "touchstart", "touchmove", "touchend")
+        ];
+    }
+
+    customDragEvent(element: Node, startOnEvent: string, moveOnEvent: string, endOnEvent: string): Function {
```
adityarb88: Should be private.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
```diff
@@ -0,0 +1,105 @@
...
+        let dragMoveListener: any;
+        let dragEndListener: any;
+
+        return this._renderer.listen(element, startOnEvent, (startEvent: any) => {
+            this._dragStartPosition = {x: startEvent.clientX, y: startEvent.clientY};
+            startEvent.stopPropagation();
```
adityarb88: Why was this added? It wasn't there before? I thought we decided that we weren't going to use this in our code.

adityarb88: No one else in the application will be able to listen to this event.

youdz:
The fact that this PR is 1200 lines *without* unit tests is a problem, we're still way bigger than what we want. You should have submitted early PRs for just the drag dispatchers and their unit tests, for instance.

src/ks-app/src/app/containers/drag-and-drop/drag-and-drop.component.html
```diff
@@ -0,0 +1,179 @@
...
+    <h4>Basic Draggable</h4>
+    <div class="row">
+        <div class="col-xs-12">
+            <p>
+                Use <code class="clr-code">clrDragAndDropGroup</code> directive to make particular draggables interact
+                with particular droppables.
+            </p>
+            <button class="btn btn-primary" clrDraggable clrDragAndDropGroup="dnd-1">Draggable Button</button>
+            <button class="btn btn-success" clrDraggable clrDragAndDropGroup="dnd-2">Draggable Button</button>
+            <button class="btn btn-warning" clrDraggable clrDragAndDropGroup="dnd-3">Draggable Button</button>
```
adityarb88: Can you try adding an array to the `clrDropGroup`?

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+import {IfDragged} from "./if-dragged";
+import {DragPosition} from "./interfaces/client-position";
+import {DraggableEvent} from "./interfaces/draggable-event";
+import {DragAndDropDispatcher} from "./providers/drag-and-drop-dispatcher";
+import {DragDispatcher} from "./providers/drag-dispatcher";
+import {CUSTOM_GHOST_STATE, CUSTOM_GHOST_STATE_PROVIDER, CustomGhostState} from "./providers/draggable-tracker";
+
+@Directive({selector: "[clrDraggable]", providers: [DragDispatcher, DomAdapter, CUSTOM_GHOST_STATE_PROVIDER]})
+export class Draggable implements AfterViewInit, OnDestroy {
+    constructor(private dragDispatcher: DragDispatcher, private dragAndDrop: DragAndDropDispatcher,
```
adityarb88: Your variables are named inconsistently. One is named `dragDispatcher` and the other is named `dragAndDrop`. Both are providers and both end with `Dispatcher`

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-dispatcher.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    get onDragEnd(): Observable<DraggableEvent> {
+        return this._onDragEnd.asObservable();
+    }
+
+    get onDrop(): Observable<void> {
+        return this._onDrop.asObservable();
+    }
+
+    dragStart(dragEvent: DraggableEvent) {
```
adityarb88: nit - @youdz do we have a naming convention for these types of methods where you are just calling the `next` method on the `Subject`. I think naming this method as `notifyDragStart` instead of just `dragStart` would help better understand what it is doing.

src/app/drag-and-drop/drag-and-drop.demo.html
Shijir: I agree `clrDropGroup` is better than `clrDragAndDropGoup`. Also, we could use `clrDroppableAccept`? Because I see that it's convention to use the term `accept` in other DnD libraries. What do you think? @youdz could you also, chime in on this?

src/app/drag-and-drop/drag-and-drop.demo.html
Shijir: Definitely, I will better organize and clean up the app demo part.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
Shijir: Are you recommending `() => void` instead of `Function`?

src/clr-angular/utils/drag-and-drop/interfaces/client-position.ts
Shijir: Thanks. Will address this!

src/app/drag-and-drop/drag-and-drop.demo.html
Shijir: Also, If we use `clrDroppableAccept` on `clrDroppable`, we need to use some other input name for clrDraggable.

Shijir: How about `clrDropLink` on `clrDraggable` and `clrDropLinkGroup` on `clrDroppable`?

src/clr-angular/utils/drag-and-drop/interfaces/client-position.ts
Shijir: Yes, it's coordinates on page. Will comment and make it clear.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: Ah this one is a bit different. This is why using the same input name "dragAndDropGroup"  on  both `clrDraggable` and `clrDroppable` makes it confusing. 

It's an event originated from `clrDraggable` so the `dragAndDropGroup` is going to be always a single string.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: Will address this issue.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
Shijir: Agree. This would be much better if pass the necessary items as parameters in the `initialize` method. Will address this issue.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
Shijir: Yes

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
Shijir: This was necessary in case users need to nest multiple draggables inside each other, which is common in DnD. Basically, events on children draggables shouldn't interfere with the ones emitted from the parent draggables.

src/clr-angular/utils/drag-and-drop/interfaces/draggable.ts
Shijir: Will do.

src/ks-app/src/app/containers/drag-and-drop/drag-and-drop.component.html
Shijir: Yes! I have updated the app to demonstrate this case.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
youdz: Exactly. :) It shows better what it does.

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: Will address this.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-dispatcher.ts
youdz: I don't think a convention brings much here, other than unnecessary verbosity. To be fair, I think a more natural method name would be `startDrag()`, but anything including `notify` feel overkill.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -101,12 +102,12 @@ export class ClrDatagridColumn extends DatagridFilterRegistrar<DatagridStringFil
...
     @ViewChild("columnHandle")
     set handleElRef(value: ElementRef) {
-        this._dragDispatcher.handleRef = value;
+        this._dragDispatcher.draggable.self = value.nativeElement;
     }
 
     @ViewChild("columnHandleTracker")
     set handleTrackerElRef(value: ElementRef) {
-        this._dragDispatcher.handleTrackerRef = value;
+        this._dragDispatcher.draggable.ghost = value.nativeElement;
```
youdz: As mentioned before, calling `dragDispatcher.initialize(...)` in `ngAfterViewInit()` with all the necessary parameters would be cleaner than atomically mutating a bunch of "deep" properties all over the place.

src/clr-angular/data/datagrid/render/dom-adapter.ts
```diff
@@ -37,10 +37,26 @@ export class DomAdapter {
         return parseInt(element.getBoundingClientRect().right, 10);
     }
 
+    clientRectLeft(element: any): number {
+        return parseInt(element.getBoundingClientRect().left, 10);
```
youdz: `getBoundingClientRect` can be expensive, you don't want to call it over and over. It was fine when you called it only once per element for the column resizer, but now you're calling it over and over. Please just expose a single method `clientRect(element: any): ClientRect`, store the result in your directive/component and then get the various properties you need from it.

src/clr-angular/utils/drag-and-drop/providers/drag-dispatcher.ts
youdz: I get the reason you did it, but we can't use `stopPropagation()`, simple as that. Use something else to ignore events from nested draggables, but let the event bubble all the way up. A library can never use `stopPropagation()`, it will break any apps or third-party libraries that expect the event to bubble up.

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
```diff
@@ -0,0 +1,43 @@
...
+    }
+    .in-drag {
+        cursor: grabbing;
+        user-select: none;
+        * {
+            pointer-events: none;
+        }
+    }
+    .drop-area {
+        min-height: 96px;
```
youdz: Hard coded pixel width. And it probably should be rems.

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
```diff
@@ -0,0 +1,43 @@
...
+        &.drop-enabled {
+            border: dashed 2px grey;
+        }
+        &.drop-allowed {
+            border: dashed 2px greenyellow;
```
youdz: What are these hardcoded colors? They're not in the palette.

src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {CommonModule} from "@angular/common";
+import {NgModule, Type} from "@angular/core";
+
+import {Draggable} from "./draggable";
```
youdz: Naming conventions! https://github.com/vmware/clarity/wiki/Naming-Conventions#typescript

src/clr-angular/utils/drag-and-drop/providers/draggable-tracker.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
youdz: This file name doesn't match what's in it.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Copyright (c) 2016-2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {animate, style, transition, trigger} from "@angular/animations";
+import {Component, ElementRef, HostBinding, Inject} from "@angular/core";
+
+import {DomAdapter} from "../../data/datagrid/render/dom-adapter";
```
youdz: Why are you using the Datagrid's DomAdapter here? If we end up having to use it everywhere, we should start extracting it to `utils`.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
youdz: I'll get to that part at some point in the review, but draggables can also belong to more than one group. I suspected you didn't take this into account as I went through the other files.

src/clr-angular/utils/drag-and-drop/providers/draggable-tracker.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    constructor() {
+        this._draggableId = "clrDraggable-" + (nbDraggableDirective++);
+    }
+}
+
+export const CUSTOM_GHOST_STATE_PROVIDER = {
+    provide: CUSTOM_GHOST_STATE,
+    useClass: CustomGhostState
```
youdz: Why the complex provider when you're using a simple class? I don't see the point.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,46 @@
...
+    constructor(private el: ElementRef, private dragDispatcher: DragDispatcher, private domAdapter: DomAdapter,
+                @Inject(CUSTOM_GHOST_STATE) private customGhostState: CustomGhostState) {
+        this.draggableGhostEl = el.nativeElement;
+
+        let originalLeft: string;
+        let originalTop: string;
+
+        if (customGhostState.isDirectChild) {
+            originalLeft = this.dragDispatcher.dragStartPosition.x + "px";
+            originalTop = this.dragDispatcher.dragStartPosition.y + "px";
```
youdz: Doing this in the constructor is going to create problems down the line, I believe. Especially when you have an `*ngFor` creating a bunch of draggables, or conditional ghosts.

It's very hard to review this PR because there are several unguarded accesses to deep properties of other classes like this. Most of these properties are undefined until something happens (like drag start), and they would throw undefined errors if called at the wrong time. In other words, stuff seems to be in one class or another randomly, with little thought for separation of concerns or privacy of internal state.

In  this specific case, because you seem to magically be able to access the `dragStartPosition` in the constructor of a component, I have no idea if you should move this to ngAfterViewInit, or to a subscription, or simply to guard it if the drag hasn't started yet.

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+
+        this.dropSubscription = this.dragAndDrop.onDrop.subscribe(() => {
+            // Only Draggable which dragStart event detected on should subscribe to the drop event.
+            // That's why this subscription is set inside this method instead of AfterViewInit.
+            // And it has to be unsubscribed not only inside this.ngOnDestroy() as well as this.dragEnd() method.
+            this.drop();
+        });
+        this.subscriptions.push(this.dropSubscription);
+
+        if (!this.customGhostState.isDirectChild) {
```
youdz: I've read every use of this `isDirectChild`, and I still have no idea at all why it's named like this. It seems to describe cases where the consumer provides a custom ghost, maybe?

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,46 @@
...
+        } else {
+            originalLeft = this.domAdapter.clientRectLeft(this.dragDispatcher.draggable.self) + "px";
+            originalTop = this.domAdapter.clientRectTop(this.dragDispatcher.draggable.self) + "px";
```
youdz: Why the different treatment here? Why do custom ghosts start on the dragstart position, while non-custom ones start on the original element's position? If I provide a custom ghost that's pretty much a copy of my draggable element, I don't want a completely different behavior, I want the same as the original with my custom ghost. 

I don't know which of the two behaviors is the right one, I'll let you see that with UX, but it should be the same one for both cases.

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    private isDragMoveStarted: boolean = false;
+
+    @Output("clrDragStart") private dragStartedEmitter: EventEmitter<DraggableEvent> = new EventEmitter();
+    @Output("clrDragMove") private dragMovedEmitter: EventEmitter<DraggableEvent> = new EventEmitter();
+    @Output("clrDragEnd") private dragEndedEmitter: EventEmitter<DraggableEvent> = new EventEmitter();
+
+    // This input is or transfering payload from draggable to dropppable
+    // We probably need to have a definite type on it with an interface instead of having any.
+    @Input("clrDraggable") data: any;
```
youdz: I think what you want is just:
```typescript
export class Draggable<T>
```
```typescript
@Input("clrDraggable") data: T;
```

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    @Input("clrDragAndDropGroup") groupKey: string;
```
youdz: Yup, there it is. Draggables can belong to more than one group, so this is `string | string[]` too.

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    private subscriptions: Subscription[] = [];
+
+    private dropSubscription: Subscription;
+
+    private _customHandle: DraggableHandle;
+    @ContentChild(DraggableHandle)
+    set customHandle(draggableHandle: DraggableHandle) {
```
youdz: Using `@ContentChild(DraggableHandle)` for draggable handles won't work on stuff like cards, because the handle will be a bit deeper down the card's DOM, not at the root.
Using `@ContentChild(DraggableHandle, {descendants: true})` will break nested draggables.
To top it all, `<my-component clrDraggable></my-component>` is a very valid use case, and it just won't work with `@ContentChild` at all.

 I think you're going to have to use registration/unregistration with dependency injection.

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -0,0 +1,246 @@
...
+        if (draggableHandle) {
+            this._customHandle = draggableHandle;
+            this.setDraggableStyles(this._customHandle.draggableHandleEl);
```
youdz: You have a specific directive for the drag handle, yet you do the styling in this big draggable class? The handle should handle its own styling, pun intended.

Issue #1878, opened 2018-01-19
Clarity Forms Redesign Implementation Plan


gnomeontherun:
This outlines the expected plan for refactoring and enhancing forms in Clarity. The specifics may change over time based on the actual versioning of Clarity, and any release dates or versions are subject to change.

## Initial changes for v0.12

First step is to release only what is new and non-breaking, and deprecate current checkbox.

* [x] Deprecate existing `ClrCheckbox`
* [x] Create `HostWrapper`
* [x] Reorganize SCSS
* [x] Update styles based on UX

##### `ClrInput`
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

##### `ClrPassword`
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

## After v0.12

This version is about adding more components, which don't currently have.

##### `ClrRadio`
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

##### `ClrCheckbox` (new version)
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

##### `ClrSelect`
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

##### `ClrTextarea`
  * [x] Default Styling
  * [x] Error Support
  * [x] Disabled & Read Only
  * [x] Horizontal & Compact Layouts
  * [x] Documentation

##### `ClrInputPassword`
  * [x] Show/hide

## Uncommitted 

* Additional input types [`number`, `date`, `email`, `range`, `search`, `tel`, `time`, `url`] if they need any custom support
* Select2 which may be provided from external contribution

MithilaTople:
Hi, @jeremydejiacomo and I were looking at some of the mockups for error handling in the new form update. The error icon is placed next to the field (except for radio buttons and checkboxes) and the error message appears below the field. We were wondering if the icon could be placed next to the error message, since the current placement seems a bit detached from the message and could be confused with a tooltip.

mathisscott:
@reddolan + @gnomeontherun ^

reddolan:
[...]

PR #1997, opened 2018-02-15
[NG] Radios and Checkboxes

youdz:
New radio and checkbox components, using the host wrapping trick to provide a simpler API. These two new components are part of #1878.

@gnomeontherun: What's the plan for forms deprecation here? We have a naming conflict, so I'm marking this PR as "Do not merge" until we know exactly what we want to do.

### DIFF
--- a/src/app/checkboxes/checkboxes.demo.html
+++ b/src/app/checkboxes/checkboxes.demo.html
@@ -116,118 +116,31 @@
     </section>
 </form>
 
-<pre>
-    <code clr-code-highlight="language-html">
-    &lt;form #testForm=&quot;ngForm&quot;&gt;
-        &lt;section class=&quot;form-block&quot;&gt;
-            &lt;label&gt;Checkboxes&lt;/label&gt;
-            &lt;div class=&quot;form-group&quot;&gt;
-                &lt;label&gt;Default/Stacked checkbox group&lt;/label&gt;
-                &lt;div class=&quot;checkbox&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_1&quot;&gt;
-                    &lt;label for=&quot;checkrads_1&quot;&gt;Checkbox 1&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_2&quot; checked&gt;
-                    &lt;label for=&quot;checkrads_2&quot;&gt;Checkbox 2&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox disabled&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_3&quot; disabled&gt;
-                    &lt;label for=&quot;checkrads_3&quot;&gt;A disabled and unchecked checkbox&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox disabled&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_4&quot; disabled checked&gt;
-                    &lt;label for=&quot;checkrads_4&quot;&gt;A disabled and checked checkbox&lt;/label&gt;
-                &lt;/div&gt;
-            &lt;/div&gt;
-            &lt;div class=&quot;form-group&quot;&gt;
-                &lt;label&gt;Inline checkbox group&lt;/label&gt;
-                &lt;div class=&quot;checkbox-inline&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_5&quot;&gt;
-                    &lt;label for=&quot;checkrads_5&quot;&gt;Checkbox 1&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox-inline&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_6&quot; checked&gt;
-                    &lt;label for=&quot;checkrads_6&quot;&gt;Checkbox 2&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox-inline disabled&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_7&quot; disabled&gt;
-                    &lt;label for=&quot;checkrads_7&quot;&gt;A disabled and unchecked checkbox&lt;/label&gt;
-                &lt;/div&gt;
-                &lt;div class=&quot;checkbox-inline disabled&quot;&gt;
-                    &lt;input type=&quot;checkbox&quot; id=&quot;checkrads_8&quot; disabled checked&gt;
-                    &lt;label for=&quot;checkrads_8&quot;&gt;A disabled and checked checkbox&lt;/label&gt;
-                &lt;/div&gt;
-            &lt;/div&gt;
-            &lt;div class=&quot;form-group&quot;&gt;
-                &lt;label&gt;Indeterminate Checkbox&lt;/label&gt;
-                &lt;div&gt;
-                    &lt;clr-checkbox type=&quot;checkbox&quot; id=&quot;checkbox_9&quot; [(clrIndeterminate)]=&quot;indeterminateState&quot;&gt;
-                        Clarity indeterminate checkbox
-                    &lt;/clr-checkbox&gt;
-                    &lt;div&gt;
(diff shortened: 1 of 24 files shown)


gnomeontherun:
My suggestion is we name the new checkbox module differently, until we remove the old one in 0.13. So essentially its not useable until then, but it (and radio) also needs to conform to the new layouts I'm working on. So I think we hold them out of the public API until 0.13, when we rename them and there is enough form support for people to wholesale move over.

youdz:
The problem is not just the checkbox module, the entire `ClrFormsModule` itself is going to be a conflict.

gnomeontherun:
I don't think we also want to make these available until they have the new forms layout, so the main conflict is the `ClrCheckboxModule` while the rest are just new.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -0,0 +1,26 @@
...
+        <ng-content select="[clrRadio]"></ng-content>
+        <ng-content></ng-content>
+        <label *ngIf="_dynamic"></label>
+    `,
+    host: {"[class.radio]": "true"},
+    providers: [FormControlService]
+})
+export class ClrRadioContainer implements DynamicWrapper {
+    // Indicates whether the container is dynamically created by the radio button itself
+    _dynamic = false;
```
hippee-lee: Same question as above.
Is this supposed to be private, per our [guidlines](https://github.com/vmware/clarity/wiki/Naming-Conventions#public-vs-private) ?

src/app/checkboxes/checkboxes.demo.html
```diff
@@ -116,118 +116,31 @@
...
-    &lt;/form&gt;
-    </code>
-</pre>
+<!-- User can't be bothered to give an ID -->
+<clr-checkbox-container>
+    <input clrCheckbox type="checkbox">
+    <label>No id</label>
+</clr-checkbox-container>
+
+<!-- Radio with no label -->
```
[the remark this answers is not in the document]
youdz: Yay copy-paste! My bad.

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -0,0 +1,26 @@
...
+        <ng-content select="[clrCheckbox]"></ng-content>
+        <ng-content></ng-content>
+        <label *ngIf="_dynamic"></label>
+    `,
+    host: {"[class.checkbox]": "true"},
+    providers: [FormControlService]
+})
+export class ClrCheckboxContainer implements DynamicWrapper {
+    // Indicates whether the container is dynamically created by the radio button itself
+    _dynamic = false;
```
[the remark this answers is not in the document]
youdz: This has to be public because another class will set the value, but it's internal wiring so we don't want consumers to mess with this property. This case is addressed in our guidelines:
>Finally, some properties or methods needs to be public for wiring purposes, but we don't want consumers to play with them. These will stay public but be prefixed with an underscore.

src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+import {Directive, ViewContainerRef} from "@angular/core";
+
+import {WrappedFormControl} from "../common/wrapped-form-control";
+
+import {ClrCheckboxContainer} from "./checkbox-container";
+
+@Directive({selector: "[clrCheckbox]"})
+export class ClrCheckboxNext extends WrappedFormControl<ClrCheckboxContainer> {
```
gnomeontherun: This means you agree with the `Next` annotation, which I'm happy with. Speak now or forever hold your peace.

src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -0,0 +1,20 @@
...
+    // Once again, several more elegant solutions were foiled by severity 3+ bugs on Angular that have been opened
+    // for 6 months to a year. So that's how we do it. Inheritance and ridiculous constructors. :-(
```
gnomeontherun: Please point to this ticket in the comment, otherwise we might lose track of it.

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: No, it is used in the template and therefore is `public`.

src/ks-app/src/app/app.module.ts
```diff
@@ -78,8 +79,10 @@ import {KSWizards} from "./containers/wizard/wizards.component";
         KSKillersDemo,  KSHotFussDemo,    KSProgressBars,  KSSpinners,
         KSTables,       KSTooltips
     ],
-    imports:
-        [BrowserModule, BrowserAnimationsModule, FormsModule, ReactiveFormsModule, ClarityModule, AppRoutingModule],
+    imports: [
+        BrowserModule, BrowserAnimationsModule, FormsModule, ReactiveFormsModule, ClarityModule, ClrRadioModule,
```
gnomeontherun: You should also import Checkboxes and add them to KS app

src/clr-angular/forms/radio/radio.ts
```diff
@@ -0,0 +1,20 @@
...
+import {Directive, ViewContainerRef} from "@angular/core";
+
+import {WrappedFormControl} from "../common/wrapped-form-control";
+
+import {ClrRadioContainer} from "./radio-container";
+
+@Directive({selector: "[clrRadio]"})
+export class ClrRadio extends WrappedFormControl<ClrRadioContainer> {
+    // Once again, several more elegant solutions were foiled by severity 3+ bugs on Angular that have been opened
+    // for 6 months to a year. So that's how we do it. Inheritance and ridiculous constructors. :-(
```
gnomeontherun: Same comment about ticket url

src/clr-angular/forms/radio/radio.module.ts
```diff
@@ -0,0 +1,25 @@
...
+import {NgModule} from "@angular/core";
+
+import {ClrHostWrappingModule} from "../../utils/host-wrapping/host-wrapping.module";
+import {ClrCommonFormsModule} from "../common/common.module";
+
+import {ClrRadio} from "./radio";
+import {ClrRadioContainer} from "./radio-container";
+
+/**
+ * Internal module, please do not export!
```
gnomeontherun: I forget if this compiles correct (TS seems to have an open PR that might indicate this wont work at the moment) but can we put `@private` on this comment for IDEs?

src/clr-angular/forms/checkbox/checkbox-container.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+            this.checkboxContainer = checkboxContainerDE.componentInstance;
+            this.checkboxContainerEl = checkboxContainerDE.nativeElement;
+        });
+
+        it("declares a FormControlService provider", function(this: TestContext) {
+            expect(this.formControlService).toBeTruthy();
+        });
+
+        it("implements DynamicWrapper", function(this: TestContext) {
+            // Typescript pretty much tests this for us, so this unit test is a bit nonsensical.
```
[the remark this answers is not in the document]
hippee-lee: Agreed.

src/clr-angular/forms-deprecated/forms.module.ts
```diff
@@ -7,9 +7,7 @@
 import {CommonModule} from "@angular/common";
 import {NgModule} from "@angular/core";
 
-import {ClrCommonFormsModule} from "../forms/common/common.module";
-
 import {ClrCheckboxModule} from "./checkbox/checkbox.module";
 
-@NgModule({imports: [CommonModule], exports: [ClrCommonFormsModule, ClrCheckboxModule]})
+@NgModule({imports: [CommonModule], exports: [ClrCheckboxModule]})
```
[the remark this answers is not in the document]
youdz: As discussed on Slack, I believe our deprecated forms don't need the `ClrCommonFormsModule` at all.

src/clr-angular/forms/checkbox/checkbox.ts
youdz: 🤐

src/clr-angular/forms/checkbox/checkbox.ts
youdz: I have subscriptions on the issues, but as mentioned before it's impossible to find tickets you're subscribed to... Anyway, once the newest compiler hits, maybe we'll want to do a pass on all of our abstract classes and inheritance patterns. But it's not going to be a priority.

src/ks-app/src/app/app.module.ts
youdz: I didn't because I was thinking of a checkbox conflict, but I guess I can do it now.

src/clr-angular/forms/radio/radio.module.ts
youdz: I'm not sure what the point of `@private` would be. All that matters here is that `clr-angular/index.ts` doesn't export this, right? If people start importing stuff from source files rather than our index, then they already know they're doing something not supported.

Issue #1957, opened 2018-02-06
Improper CSS targeting Clarity Icons in Dropdown


Shijir:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Clarity Icons in Dropdown should look pleasurable.

### Actual behavior

Clarity Icons are not targetted properly in dropdowns. As a result, all icons placed inside a dropdown component overlapped on top of each other. Please see the issue here: 

<img width="172" alt="screenshot 2018-02-05 23 21 26" src="https://user-images.githubusercontent.com/3958008/35846422-721cc898-0acb-11e8-8b28-98aa80f732f2.png">

https://stackblitz.com/edit/clarity-light-theme-v11-zfx7vw?file=app/app.component.html

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

gnomeontherun:
Resolved by #2012

PR #2012, opened 2018-02-23
[ui] fix overlapping clr-icons when inside header dropdown (#1957)

jeeyun:
The css that causes the overlap of icons exists for the case when we use clr-icon as the trigger for the dropdown on the header. I just increased the specificity of css selector so it doesn't apply to what goes in the dropdown content.

This is non-breaking change. I also added a css regression test for an example header with icons.

p.s. my travis builds kept failing one css regression tests in wizard (it was color difference in the code example that shows up behind the wizard). I was only able to replicate this when I upgraded my local node version and installed the same version of gemini and html-reporter as travis. So that one reference image is updated now. Not sure why this isn't an issue on vmware clarity repo and only on my forked repo.


src/app/nav/header-types.ts
```diff
@@ -3,6 +3,7 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import "clr-icons/shapes/essential-shapes";
```
gnomeontherun: This path should be relative to the actual directory

Shijir: Does it need to be relative? Because we have several component files that already use this type of non-relative module imports in our demo app. Also, Isn't it actually better to have non-relative module imports if we ever need to change the folder structures of our project in the future?

gnomeontherun: Even if we move folders, paths could still be wrong. However, I ran into an issue with these types of paths when working on https://github.com/vmware/clarity/pull/2015. I worked on this further, and it does not actually cause a problem now. I am wondering if I changed something with TypeScript (as I was tinkering with the solution above) that broke this. 

@jeeyun it appears I was wrong, and you could revert that change for relative path. Sorry :(

mathisscott:
@jeeyun 
Can we also port this fix back to 0.10?

jeeyun:
@mathisscott  Yup we definitely can.

src/app/nav/header-types.ts
jeeyun: Haha no worries. Will revert and also rebase since this is close to being merged.

PR #2015, opened 2018-02-23
[NG] fix issues for changes in angular cli

gnomeontherun:
The @ngtools/webpack module renamed the AotPlugin to AngularCompilerPlugin and we should use that name. There was a time they supported both names, but the AotPlugin name is no longer in master and will not work in Angular 6.

Signed-off-by: Jeremy Wilken <gnomation@gnomeontherun.com>


gnomeontherun:
@adityarb88 was there a reason these icons were being imported in the app? The paths were wrong but I'm a little uncertain as to why it didn't throw errors previously.

gnomeontherun:
Yes I put them back since I no longer need them for this changeset. It also passes now.

PR #2018, opened 2018-02-26
[NG] Fix stack view overflow-y issue with Angular universal

alyssabrunswick:
When using stack-view with angular universal, there is an error with the animation because overflow-y is not a supported CSS property for animations. This moves it from the animation to CSS.

Signed-off-by: Alyssa Brunswick <abrunswick@oriental.com>


gnomeontherun:
Thanks @alyssabrunswick for the contribution, we'll have to take a closer look at why the visual diff test is failing from this change.

adityarb88:
@alyssabrunswick: Could you please rebase on master? Rebasing on master should fix the visual diff issue.

alyssabrunswick:
I rebased on master. Thanks!

Issue #1681, opened 2017-11-09
button loading spinner not showing in button group


marvinosswald:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

A loading spinner should show on a button even if its in a button group

### Actual behavior

loading spinner doesn't show

### Reproduction of behavior

https://plnkr.co/edit/k0vC2vtiV4wvPP4RrRZv?p=preview

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 0.10.12

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@marvinosswald That plnkr contains a Modal. Could you please update to replicate the issue you are facing?

marvinosswald:
@adityarb88 my bad, updated the plnkr same url though

adityarb88:
@marvinosswald: Thanks for reporting this. This definitely is a bug and we will add this to our backlog.

govi2010:
@adityarb88  I think this is because of `selector: "button[clrLoading]"` in loading-button.ts. ...

adityarb88:
@govi2010: Yes right now the loading indicators works only for static buttons but we wont be able to extend the selector to `button[clrLoading], clr-button[clrLoading]` because `clr-button` already is a separate component. Angular will throw an error when two different templates are defined. I will think of a solution and discuss with the team on how we can resolve this.

govi2010:
@adityarb88  I fixed this issue by

govi2010:
<img width="600" alt="screen shot 2017-11-10 at 9 11 26 pm" src="https://user-images.githubusercontent.com/8776031/32665896-c1f1c7f6-c65b-11e7-87be-d6f6bd30be69.png">

govi2010:
@adityarb88  any update on my code review. Let me know it you want to fix this any other way

PR #1685, opened 2017-11-10
Fix for loader in clr-button   Issue #1681

govi2010:
Signed-off-by: Govinda Alwani <govi2010@gmail.com>

PR #2022, opened 2018-02-27
[NG] Support loading buttons in button groups

adityarb88:
Fixes: #1681

Adds tests to this PR: #1685
- Had to rewrite tests for the `clr-button` component as the previous tests were not testing everything.

Tested on Chrome
test 1

### DIFF
--- a/src/clr-angular/button/button-group/button.ts
+++ b/src/clr-angular/button/button-group/button.ts
@@ -7,6 +7,7 @@
 
 import {Component, EventEmitter, Input, Optional, Output, SkipSelf, TemplateRef, ViewChild} from "@angular/core";
 
+import {LoadingListener} from "../../utils/loading";
 import {ButtonInGroupService} from "../providers/button-in-group.service";
 
 @Component({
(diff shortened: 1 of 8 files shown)


src/clr-angular/button/button-group/button.ts
```diff
@@ -108,6 +111,16 @@ export class ClrButton {
         }
     }
 
+    public loading: boolean;
+
+    startLoading(): void {
```
mathisscott: Why wouldn't we want to move this into a reusable piece at some point? Seems unnecessary to have this boilerplate duplicated everytime someone wants to use `LoadingListener`.

youdz: Because annotations don't have access to the injector, and dynamically modifying a component's declaration breaks AoT.

Inheritance would be the main way to do it, but inheritance is a terrible pattern for such composable features, we can't use it.

mathisscott: I was thinking something along the lines of a mixin (ala Rails). But, yes, composition. If composition isn't viable in our framework of choice...

¯\_(ツ)_/¯

Issue #2020, opened 2018-02-27
Tooltip with long text is not shown properly


swapnilsvaidya1412:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

If tooltip is having long text, then tooltip height/width should gets adjusted automatically to accomodate all the text
### Actual behavior

If tooltip component is having long text, then tooltip text is not shown properly.
 Text overflows the tooltip element and if it is used in card then scroll bar appears on the card.
### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:
https://stackblitz.com/edit/clarity-dark-theme-v10-efelwd?file=app%2Fapp.component.html

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

gnanasekhar504:
Use tooltip-lg and break the long words.

adityarb88:
@swapnilsvaidya1412: 
For now you could add this style at your end to get this fixed immediately. We will add it at our end too:

```
.tooltip-content {
  word-break: break-word;
}
```
Info: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
Demo: https://stackblitz.com/edit/clarity-dark-theme-v10-fndbkb?file=app/app.component.css

BTW our UX recommendation is that you use tooltip for a line of text only. Adding a lot of content in a tooltip is not recommended.

adityarb88:
@swapnilsvaidya1412: I missed the discussion that our team did earlier regarding this. We will not be adding styles which break a word to wrap into multiple lines as this might cause unexpected behavior. For more info please read this: https://github.com/vmware/clarity/issues/1146#issuecomment-313105291

We recommend that you increase the size of the tooltip to let the content fit correctly or use the style we mentioned above in your CSS.

Closing this issue based on the above info. Please feel free to get in touch with us if you have more questions/feedback.

swAucxis:
Instead of defining a fixed width. We would prefer an auto width solution. And the xs, sm, md, lg determines the max width.

Issue #1146, opened 2017-07-03
Clarity datagrid does not word-wrap or show ellipsis on fields with underscore.


SandeshSomashekar:
Clarity datagrid does not word-wrap or show ellipsis on fields with underscore.

```
[ x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Ideally, columns with text longer than the available column width must either wrap or show ellipsis with mouse over for the whole text.

### Actual behavior
Text containing underscores overlaps with the text on the next column.

### Reproduction of behavior
Have a clarity column text with data containing several underscore characters and see how it is displayed. Attaching a screenshot.

Clarity Version: [Latest - 0.9.2]

### Environment details

* **Angular version:** 4.2.4

* **Clarity version: 0.9.2 

* **OS and version: OSX 10.12.5

* **Browser:** Chrome Version 58.0.3029.110 (64-bit)
<img width="955" alt="screen shot 2017-07-03 at 12 03 12 pm" src="https://user-images.githubusercontent.com/1157190/27805434-7b691b96-5fe9-11e7-8f8b-d9f956be0da6.png">

SandeshSomashekar:
For now I am overriding datagrid-cell style with the following to temporarily fix this -

.datagrid-cell {
   word-break: break-all;
}

youdz:
Two things:
1) Please post a plunker reproducing this, because Datagrid actually sizes columns smartly and will allocate enough space for them, even if it means adding a horizontal scroll.
2) Feel free to follow #1065 for a potential helper class if it is really needed. See #815 and #618 for discussion of why we will not truncate text by default. Note that the `word-break: break-all` solution is not something we'd implement by default either, since it leads to terrible word breaking on normal text. I'm closing this as duplicate, we can reopen if you do have a plunker that shows a different use case as the mentioned issues.

PR #2023, opened 2018-02-28
[UI] Tooltip Content Break Work

adityarb88:
Fixes: #2020

image

 More info: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

Tested in Chrome, Firefox, Safari

### DIFF
--- a/src/clr-angular/popover/tooltip/_tooltips.clarity.scss
+++ b/src/clr-angular/popover/tooltip/_tooltips.clarity.scss
@@ -86,6 +86,7 @@
             opacity: 0;
             transition: opacity 0.3s linear;
             white-space: normal;
+            word-wrap: break-word;
             z-index: map-get($clr-layers, tooltips);
         }
 
--- a/src/app/tooltips/tooltips-angular.html
+++ b/src/app/tooltips/tooltips-angular.html
@@ -113,3 +113,14 @@
     </clr-dropdown>
 </div>
 
+<h6>Tooltip Break Word</h6>
+<div class="clr-example" style="margin-bottom:480px">
+    <clr-tooltip>
+        <clr-icon clrTooltipTrigger shape="home"></clr-icon>
+        <clr-tooltip-content [clrSize]="'sm'">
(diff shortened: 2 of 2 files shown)


src/clr-angular/popover/tooltip/_tooltips.clarity.scss
```diff
@@ -86,6 +86,7 @@
             opacity: 0;
             transition: opacity 0.3s linear;
             white-space: normal;
+            word-wrap: break-word;
```
mathisscott: Is this something we want on by default? Or would it make more sense to offer a utility class that applies it to the tooltip?

gnomeontherun: I agree, best UX is to avoid putting more content and making this the default subverts that slightly. I would make an optional class to wrap, or just put some documentation for people to know how to override in the future. Not sure we really should proactively encourage this?

adityarb88: @mathisscott @gnomeontherun: 
Just read the description here: https://github.com/vmware/clarity/issues/1146
Will close this PR. It seems like we decided not to add `word-break` in our CSS at all and let the user add it at their end if they are facing such issues because it breaks the word.

I do feel that this style can be included because this won't affect how the tooltip content is rendered in general. If there are multiple words in one line which exceed the width of the content, then the entire last word will be wrapped. It won't be broken into two parts.

Eg: 
image

`word-break` just makes the styling more fool proof and avoids errors like the ones mentioned in #2020 

For now I will close this PR but may be we can talk about this later on.

mathisscott: I'm fine including it. I just wondered if it needed to be the default...

adityarb88: I am going to route this back to UX and see what they say about this. Will reopen this PR if they agree too. Thanks!

PR #2025, opened 2018-03-01
[ICONS] icon shapes are modularized

Shijir:


src/ks-app/src/app/app.component.ts
```diff
@@ -53,4 +58,15 @@ export class AppComponent {
         {path: "vertical-nav", title: "Vertical Nav"},
         {path: "wizards", title: "Wizards"}
     ];
+
+    constructor() {
+        ClarityIcons.add({
```
Shijir: Oops forgot to remove the `Shape` suffix.

src/clr-icons/shapes/core-shapes.ts
```diff
@@ -223,76 +139,84 @@ const coreShapes: any = {
...
+const searchShape = clrIconSVG(
+    `<path class="clr-i-outline clr-i-outline-path-1" d="M16.33,5.05A10.95,10.95,0,1,1,5.39,16,11,11,0,0,1,16.33,5.05m0-2.05a13,13,0,1,0,13,13,13,13,0,0,0-13-13Z"/>
+                <path class="clr-i-outline clr-i-outline-path-2" d="M35,33.29l-7.37-7.42-1.42,1.41,7.37,7.42A1,1,0,1,0,35,33.29Z"/>`);
+const viewColumnsShape = clrIconSVG(
```
gnomeontherun: Just do `export const ClrShapeViewColumns` and such for every one, so you don't have to reexport at the bottom and rename them.

Also, then should we include the shape like `ClrShapeCommerceViewColumns` in the name? That way we can avoid ambiguity and know the package.

Shijir: I tried to do it that way initially. But we have tslint rule that prevents us from declaring a capitalized variable. That's why I opted for this way. 

But we could make these shape files exempt from that tslint rule. What's your thought on that?

gnomeontherun: I'd rather exempt it personally, its easier to maintain long term by not having to export things separately

PR #2029, opened 2018-03-02
[NG] Adds aria-labelledby to clr-checkboxes

### LINKED ISSUE #1926: Checkboxes in tree view need aria-labelledby (see notes)
URL: http://clarity-ks.surge.sh/treeviews

Tools: Web Accessibility Toolbar (WAT2.0) / Inspect32

Reproduced Steps:
1. Login Clarity Kitchen Sink with web browser.
2. Go to Tree Views tab.
3. Check the name property of all check boxes in Tree Views tab.

Actual Results:
The name property of all check boxes is “”.  
It’s difficult to distinguish between them for Blind.
For more details, please refer to “TreeViews-NamePropertyIsEmpty.png”.

Expected Results:
The name property should have a descriptive and unique name property.

Note:
This issue also occurs on Vertical Nav tab.

Violation of Section 508 Standards: 
1194.21(d) Sufficient information about a user interface element including the identity, operation and state of the element shall be available to assistive technology. When an image represents a program element, the information conveyed by the image must also be available in text. 
User Impact: Blind.

Screen shot:
1.	TreeViews-NamePropertyIsEmpty.png

NOTE: this may also be an issue on the website

hippee-lee:
- Adds a generic id-generator service
- Fixes #1926

@youdz - I'm not happy about the one test in `id-generator.spec`. I only really test that the ids generated are unique. The main thing I ran into is that `let NB_INSTANCES = 0;` seems to retain its state across tests and it felt hacky to write a second test expecting the starting value to be 'clr-id-4'.  I wasn't able to reset t` NB_INSTANCES` between the first and a second test's. I tried using the `TestBed`'s

```
static resetTestEnvironment()
Reset the providers for the test injector.
```

([docs](https://angular.io/api/core/testing/TestBed#resetTestEnvironment))

But that was resetting `this` to `{}`. If you have something specific that should be tested please advise. 

### DIFF
--- a/src/clr-angular/data/tree-view/tree-node.ts
+++ b/src/clr-angular/data/tree-view/tree-node.ts
@@ -5,9 +5,10 @@
  */
 
 import {animate, state, style, transition, trigger} from "@angular/animations";
-import {Component, EventEmitter, Input, OnDestroy, Optional, Output, SkipSelf} from "@angular/core";
+import {Component, EventEmitter, Inject, Input, OnDestroy, Optional, Output, SkipSelf} from "@angular/core";
 
 import {Expand} from "../../utils/expand/providers/expand";
+import {UNIQUE_ID, UNIQUE_ID_PROVIDER} from "../../utils/id-generator/id-generator.service";
 import {LoadingListener} from "../../utils/loading/loading-listener";
 
 import {AbstractTreeSelection} from "./abstract-tree-selection";
@@ -22,7 +23,8 @@ import {TreeSelectionService} from "./providers/tree-selection.service";
             provide: TreeSelectionService,
             useFactory: clrTreeSelectionProviderFactory,
             deps: [[new Optional(), new SkipSelf(), TreeSelectionService]]
-        }
+        },
+        UNIQUE_ID_PROVIDER
     ],
     animations: [trigger("childNodesState",
                          [
@@ -30,11 +32,11 @@ import {TreeSelectionService} from "./providers/tree-selection.service";
                              state("collapsed", style({"height": 0, "overflow-y": "hidden"})),
                              transition("expanded <=> collapsed", animate("0.2s ease-in-out"))
                          ])],
-    host: {"class": ".clr-tree-node"}
+    host: {"class": "clr-tree-node"}
 })
 export class ClrTreeNode extends AbstractTreeSelection implements OnDestroy {
     constructor(public nodeExpand: Expand, @Optional() @SkipSelf() public parent: ClrTreeNode,
-                public treeSelectionService: TreeSelectionService) {
+                public treeSelectionService: TreeSelectionService, @Inject(UNIQUE_ID) public nodeId: string) {
         super(parent);
         if (this.parent) {
             this.parent.register(this);
--- a/src/clr-angular/utils/id-generator/id-generator.spec.ts
+++ b/src/clr-angular/utils/id-generator/id-generator.spec.ts
@@ -4,72 +4,51 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import {Component, Inject, NgModule, Type} from "@angular/core";
+import {Component, Inject, NgModule} from "@angular/core";
 import {ComponentFixture, TestBed} from "@angular/core/testing";
 import {By} from "@angular/platform-browser";
 
 import {UNIQUE_ID, UNIQUE_ID_PROVIDER} from "./id-generator.service";
 
 @Component({
-    selector: "test-zero",
+    selector: "id-test",
     template: `
         <div [id]="divId">My Div</div>`,
     providers: [UNIQUE_ID_PROVIDER]
 })
-class IdZero {
+class IdTest {
     constructor(@Inject(UNIQUE_ID) public divId: string) {}
 }
 
-@Component({
-    selector: "test-one",
-    template: `
-        <div [id]="divId">My Div</div>`,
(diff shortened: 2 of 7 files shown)


src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -22,19 +23,20 @@ import {TreeSelectionService} from "./providers/tree-selection.service";
...
+        },
+        UNIQUE_ID_PROVIDER
     ],
     animations: [trigger("childNodesState",
                          [
                              state("expanded", style({"height": "*", "overflow-y": "hidden"})),
                              state("collapsed", style({"height": 0, "overflow-y": "hidden"})),
                              transition("expanded <=> collapsed", animate("0.2s ease-in-out"))
                          ])],
-    host: {"class": ".clr-tree-node"}
```
youdz: Wow, can't believe we had that.

src/clr-angular/utils/id-generator/id-generator.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+@NgModule({declarations: [IdZero, IdOne, IdTwo], exports: [IdZero, IdOne, IdTwo]})
+class IdTestingModule {}
+
+@Component({
+    template: `
+        <test-zero></test-zero>
+        <test-one></test-one>
+        <test-two></test-two>`
```
youdz: Why do you need 3 different components when they're exact clones of one another?
Just use the same one 3 times, and when querying just iterate over the array of 3 elements to check the ids.

src/clr-angular/utils/id-generator/id-generator.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    idOne: IdOne;
+    idTwo: IdTwo;
+}
+
+describe("Unique ID Generator Service", function() {
+    function setupTest<T>(testContext: TestContext<T>, testComponent: Type<T>) {
+        TestBed.configureTestingModule(
+            {imports: [IdTestingModule], providers: [UNIQUE_ID_PROVIDER], declarations: [testComponent]});
+        testContext.fixture = TestBed.createComponent(testComponent);
+        testContext.idZero = testContext.fixture.debugElement.query(By.directive(IdZero)).componentInstance;
```
youdz: So this would be a `queryAll`.

src/clr-angular/utils/id-generator/id-generator.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        testContext.idOne = testContext.fixture.debugElement.query(By.directive(IdOne)).componentInstance;
+        testContext.idTwo = testContext.fixture.debugElement.query(By.directive(IdTwo)).componentInstance;
+    }
+
+    it("generates uniq id's in the correct order", function() {
+        setupTest(this, UniqueIdTest);
+        expect(this.idZero !== this.idOne !== this.idTwo).toBe(true);
```
youdz: This line doesn't work. This is interpreted in Javascript as 
```typescript
(this.idZero !== this.idOne) !== this.idTwo
```
So the first inequality returns a boolean, which is *always* `!==` from the last id.
You can easily try it out in your console, `1 !== 1 !== 1` is true.
What that means is this test always passes, even if the ids are the same. 😛 
On top of that, you're testing equality of the component instances, not their ids.

Anyway, based on my previous comments, what you want here is something like:
```typescript
const tests = [1, 2, 3];
for (const i=0; i<2; i++) {
    for (const j=i+1; j<3; j++) {
        expect(tests[i].divId).not.toEqual(tests[j].divId);
    }
}
```

src/clr-angular/utils/id-generator/id-generator.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        TestBed.configureTestingModule(
+            {imports: [IdTestingModule], providers: [UNIQUE_ID_PROVIDER], declarations: [testComponent]});
+        testContext.fixture = TestBed.createComponent(testComponent);
+        testContext.idZero = testContext.fixture.debugElement.query(By.directive(IdZero)).componentInstance;
+        testContext.idOne = testContext.fixture.debugElement.query(By.directive(IdOne)).componentInstance;
+        testContext.idTwo = testContext.fixture.debugElement.query(By.directive(IdTwo)).componentInstance;
+    }
+
+    it("generates uniq id's in the correct order", function() {
+        setupTest(this, UniqueIdTest);
```
youdz: Extracting a setup function when you have a single test is a bit overkill. 😛

src/clr-angular/utils/id-generator/id-generator.spec.ts
hippee-lee: Because more is better? j/k Duh, not sure what I was doing with all three. Removed.

src/clr-angular/utils/id-generator/id-generator.spec.ts
hippee-lee: yep

src/clr-angular/utils/id-generator/id-generator.spec.ts
hippee-lee: Updated.

src/clr-angular/utils/id-generator/id-generator.spec.ts
hippee-lee: Was planning to have more than one test but removed this for now.

Issue #1857, opened 2018-01-11
Binding alerts to array, new not shown after close all


limited-ed:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
I store alertt in array and binding whith ngFor to <clr-alert>, clr-alert is AppLevel. When alert closed array element removes. 

### Actual behavior
When all alerts closed, adding new don`t show it, in DOM tree whey added with hiden class.

### Reproduction of behavior
https://embed.plnkr.co/FjtBJtlnKni9aRtH4aoL/
### Environment details

* **Angular version:** 4.x.x

* **Clarity version: latest.

The trouble in file 
src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
on line 60: it
this.current = (this.current === 0) ? this.activeAlerts.length - 1 : this.current - 1;
must be
this.current = (this.current === 0) ? 0 : this.current - 1;

adityarb88:
@limited-ed: Thanks for reporting this issue. We will take a look and get back to you.

PR #2037, opened 2018-03-06
[NG] MultiAlertService fix for current index

bdunkz:
Prevents the current index from dropping below 0.
Fixes: #1857

Signed-off-by: Brian Duncan <bduncan@vmware.com>

### DIFF
--- a/src/clr-angular/emphasis/alert/providers/multi-alert.service.spec.ts
+++ b/src/clr-angular/emphasis/alert/providers/multi-alert.service.spec.ts
@@ -55,5 +55,9 @@ describe("Multi Alert provider", function() {
         anotherAlert.componentInstance.close();
         expect(multiAlertService.count).toBe(1);
         expect(multiAlertService.current).toBe(0);
+        // Ensure current alert does not drop below 0
+        alert.componentInstance.close();
+        expect(multiAlertService.count).toBe(0);
+        expect(multiAlertService.current).toBe(0);
     });
 });
--- a/src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
+++ b/src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
@@ -57,6 +57,9 @@ export class MultiAlertService {
     }
 
     previous() {
+        if (this.activeAlerts.length === 0) {
+            return;
+        }
         this.current = (this.current === 0) ? this.activeAlerts.length - 1 : this.current - 1;
     }
 


src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
```diff
@@ -57,6 +57,9 @@ export class MultiAlertService {
     }
 
     previous() {
+        if (this.activeAlerts.length === 0) {
```
gnomeontherun: The concern I see here is that we don't set the `current` property back to 0 when we close. So `current` could remain at a higher index, and when a new alert is added it would be out of bounds, if I'm reading it correctly here.

src/clr-angular/emphasis/alert/providers/multi-alert.service.spec.ts
```diff
@@ -55,5 +55,9 @@ describe("Multi Alert provider", function() {
         anotherAlert.componentInstance.close();
         expect(multiAlertService.count).toBe(1);
         expect(multiAlertService.current).toBe(0);
+        // Ensure current alert does not drop below 0
+        alert.componentInstance.close();
```
gnomeontherun: This test doesn't call `previous()` to verify behavior, and since the `current` value is already 0 I don't think it is testing the condition mentioned below if `current` is above 0.

gnomeontherun:
Hey @bdunkz I'm wondering about the behavior here can you verify my understanding and the conditions are handled, or if not safe guard against it?

src/clr-angular/emphasis/alert/providers/multi-alert.service.spec.ts
bdunkz: Calling close on the alert instance results in the the Alert calling previous.  To verify, I comment out the added code from multi-alert service and this test fails because -1 != 0.

src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
bdunkz: The only way we should be able to enter this block is if current already equals 0.  The service initializes at 0.  Once some number of alert instances are added, the only way for the length to get back to 0 is if each alert instance is closed.  When the last remaining alert instance is there, current will be 0.  Once it is closed, the block catches, and does not alter current, so it remains at 0, ready for any future alert instances that may be added.

PR #2040, opened 2018-03-07
fix: [NG] Changes Id Generator to be aot compatible.

hippee-lee:
The first attempt at a id-generator util was not AOT compatible. This fixes that. 


src/clr-angular/utils/id-generator/id-generator.service.ts
```diff
@@ -10,7 +10,11 @@ let NB_INSTANCES = 0;
 
 export const UNIQUE_ID = new InjectionToken<string>("UNIQUE_ID");
 
+export function UniqueIdProvider() {
```
adityarb88: should be `uniqueIdProvider`. lowercase `U`. you can fix that in your other PR. no need to update this or travis will have to run again.

src/ks-app/package.json
```diff
@@ -21,10 +21,10 @@
     "@angular/platform-browser": "^5.0.0",
     "@angular/platform-browser-dynamic": "^5.0.0",
     "@angular/router": "^5.0.0",
-    "@webcomponents/custom-elements": "1.0.0",
     "@clr/angular": "file:../../dist/clr-angular/clr-angular-0.11.7-patch.1.tgz",
     "@clr/icons": "file:../../dist/clr-icons",
     "@clr/ui": "file:../../dist/clr-ui",
+    "@webcomponents/custom-elements": "1.0.0",
```
adityarb88: this change is not related to this PR. if we revert the main fix, then this will revert too. this is not a major issue here but just something to note for the future.

src/clr-angular/utils/id-generator/id-generator.service.ts
hippee-lee: ok. will do.

src/ks-app/package.json
hippee-lee: Sorry, I missed it when I split out the changes for the other PR that I'm working on.

src/clr-angular/utils/id-generator/id-generator.service.ts
Shijir: Can we rename this `UniqueIdProvider()` function name? This function itself is not the provider. Secondly, we are exporting two different things with the same name. The only difference is one of them has uppercase letters with underscore (`UniqueIdProvider` & `UNIQUE_ID_PROVIDER`).  Maybe, `uniqueIdFactory()`?

PR #2041, opened 2018-03-07
fix: [CORE] Adds a check for AOT compilation on Clarity artifacts

### LINKED ISSUE #2028: Update Build to test Clarity artifacts with AOT build of ks-app
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
`npm run build` should be tested for AOT. 
The build script needs to be modified to:

- Install build artifacts in the ks-app
- run the `ng build --prod` for the ks-app
- Report if there are errors during the build phase of ks-app

### Actual behavior
The ks-app isn't automatically tested with a build for production.

### Reproduction of behavior
n/a

n/a

### Environment details
n/a

hippee-lee:
## Changes
- installs Clarity artifacts into the ks-app and builds that for production to test for AOT
- catches ks-app build errors and notifies on the command line
- closes #2028

## Notes
I kept it pretty simple but if there is something a script like this should do that I missed. Let me know. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
         "clean": "del ./dist",
         "cpy": "cpy",
         "build": "npm-run-all clean build:angular webpack build:icons",
+        "postbuild": "node ./scripts/aot-check.js",
         "pretest": "webpack",
         "test": "karma start ./karma.conf.js",
         "test:travis": "npm-run-all clang:check tslint:check tslint:spec test",
--- a/scripts/aot-check.js
+++ b/scripts/aot-check.js
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+const colors = require('colors/safe');
+const shell = require('shelljs');
+const VERSION = require('../package.json').version;
+
+const packagePath = `../../dist/clr-angular/clr-angular-${VERSION}.tgz`;
+const installCommand = `npm i ${packagePath} --no-save`;
+
+// Moving to the ks-app root to work.
+console.log(colors.yellow('Testing Clarity Artifacts for AOT'));
+process.chdir('./src/ks-app');
+
+// Do package installations.
+console.log(colors.yellow(`Installing latest @clr/angular with: ${packagePath}`));
+shell.exec('npm i');
+shell.exec(installCommand);
+
+// Compile ks-app for production.
+// NOTE: this must stay here (after the script cd's into src/ks-app)!!
+let status = shell.exec('npm run build --preserve-symlinks');
+console.log(colors.yellow('Testing Clarity AOT on ks-app'));
+if (status.code === 0) {
+    console.log(colors.green('Clarity passed AOT compilation in the ks-app.'));
+} else {
+    console.log(colors.red(`AOT compilation failed on ks-app!! Status: ${status.code}\n`));
+    process.exit(status.code);
+}


src/ks-app/package.json
```diff
@@ -21,7 +21,7 @@
     "@angular/platform-browser": "^5.0.0",
     "@angular/platform-browser-dynamic": "^5.0.0",
     "@angular/router": "^5.0.0",
-    "@clr/angular": "file:../../dist/clr-angular/clr-angular-0.11.7-patch.1.tgz",
+    "@clr/angular": "file:///Users/mhippely/Projects/clarity/fork/dist/clr-angular/clr-angular-0.11.7-patch.1.tgz",
```
hippee-lee: I just saw this and will push a fix for it.

package.json
```diff
@@ -7,6 +7,7 @@
         "clean": "del ./dist",
         "cpy": "cpy",
         "build": "npm-run-all clean build:angular webpack build:icons",
+        "postbuild": "node ./scripts/aot.spec.js",
```
gnomeontherun: If this fails, does the `npm run build` still return with an error state? I would expect, but wanted to be certain.

scripts/aot.spec.js
```diff
@@ -0,0 +1,31 @@
+/*
```
gnomeontherun: Let's not call this a spec file, some globs might match it based on `*.spec` to include in karma or something.

scripts/aot.spec.js
```diff
@@ -0,0 +1,31 @@
...
+const installCommand = `npm i ${packagePath} --save`;
+
+// Moving to the ks-app root to work.
+console.log(colors.yellow('Testing Clarity Artifacts for AOT'));
+process.chdir('./src/ks-app');
+
+// Do package installation.
+console.log(colors.yellow(`Installing latest @clr/angular with: ${packagePath}`));
+shell.exec(installCommand);
+shell.exec('npm i');
```
gnomeontherun: `npm ci`? Here and above?

scripts/aot.spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Compile ks-app for production.
+// NOTE: this must stay here (after the script cd's into src/ks-app)!!
+let status = shell.exec('npm run build --preserve-symlinks');
+console.log(colors.yellow('Testing Clarity AOT on ks-app'));
+if (status.code === 0) {
+    console.log(colors.green('Clarity passed AOT compilation in the ks-app.'));
+} else {
+    console.log(colors.red(`AOT compilation failed on ks-app!! Status: ${status.code}\n`));
```
gnomeontherun: I suspect we should explicitly throw an error or `process.exit(1)` to ensure its caught.

src/ks-app/package.json
gnomeontherun: Ok, I see another problem here. The file is always pointing to a specific version number, which might be changed but is sometimes not updated when we do a release. Package-lock also points to this.

Can we point to a directory instead? We build a tgz asset, but don't we also have a directory with the source files we could point to instead?

package.json
hippee-lee: I can rename.

hippee-lee: Yes. It was erroring out on the build and npm was exiting w/ error state.

scripts/aot.spec.js
hippee-lee: I will rename.

scripts/aot.spec.js
hippee-lee: I can change to ci but think that I should move the `installCommand` below so it doesn't get clobbered.

scripts/aot.spec.js
hippee-lee: Didn't see `process.exit` in my auto-completion list. Will revisit.

src/ks-app/package.json
hippee-lee: I thought I fixed this last night. 
The ks-app has issues linking to source files directly. This is why the tgz has been used since we switched to webpack.

PR #2047, opened 2018-03-07
[NG] Remove @private annotation

adityarb88:
Fixed an issue with the build.

screen shot 2018-03-07 at 2 32 37 pm

Without this fix `npm run build` fails.


gnomeontherun:
I advised we add that for IDEs, but it appears TS doesn't currently support it but this PR will https://github.com/Microsoft/TypeScript/pull/21455.

Issue #2062, opened 2018-03-11
Cannot close DatePicker popup with keyboard


Magister:
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Pressing `<Esc>` key should close popup and do not select any date.

### Actual behavior
`<Esc>` key do not close popup, the only way to do that with keyboard is to select some date.

### Environment details

* **Angular version:** 5.0.1

* **Clarity version:** 0.11.8

* **OS and version:** Debian 9

* **Browser:** Firefox 58.0.1

adityarb88:
@Magister: Thanks for raising this issue. I have added this enhancement to our backlog.

PR #2064, opened 2018-03-12
[NG] Closes a Popover on ESC

adityarb88:
Fixes: #2062

Closes a popover when the `ESC` key is pressed.

Tested on Chrome

Note: For some reason `abstract-popover.ts` didn't have any tests. I have added tests related to this change in this PR. Will add the remaining tests in a separate PR.

### DIFF
--- a/src/clr-angular/popover/common/abstract-popover.ts
+++ b/src/clr-angular/popover/common/abstract-popover.ts
@@ -7,6 +7,7 @@ import {
     AfterViewChecked,
     ElementRef,
     HostBinding,
+    HostListener,
     Injectable,
     Injector,
     OnDestroy,
@@ -16,6 +17,7 @@ import {
 import {Subscription} from "rxjs/Subscription";
 
 import {IfOpenService} from "../../utils/conditional/if-open.service";
+import {ESC} from "../../utils/key-codes/key-codes";
 
 import {Point, Popover} from "./popover";
 import {PopoverOptions} from "./popover-options.interface";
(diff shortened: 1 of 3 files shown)


src/clr-angular/popover/common/abstract-popover.ts
```diff
@@ -90,6 +92,13 @@ export abstract class AbstractPopover implements AfterViewChecked, OnDestroy {
         return this.ifOpenService.open ? false : true;
     }
 
+    @HostListener("keydown", ["$event"])
+    onKeyDown(event: KeyboardEvent) {
```
youdz: This event listener is not on the document, so it only triggers if the key is pressed while focus is inside of the popover. Is that what we want?
I don't have an opinion, but whichever the decision is, please add a comment explaining it or linking to the corresponding conversation on GitHub.

youdz: I don't believe dropdowns or tooltips should have a focus trap, unless I missed something. So you can't rely on that.
As I said, I'm fine either way, I just want the decision to be clear and explicitly commented. :)

adityarb88: ok. i will check with @reddolan once. Thanks @youdz

adityarb88:
@youdz: Per UX discussion, `dropdowns` and `tooltips` wont have `FocusTrap`, so that means we will have to add document level listeners like you suggested earlier. I have updated the PR to incorporate that change. Please take a look and let me know if this looks okay.

@hippee-lee: You might have to review the PR again.

PR #2068, opened 2018-03-13
[UI] new forms layouts for input, radio, checkbox, input-group, textarea, file controls

gnomeontherun:
This implements all new styles and layouts for the above form controls. They should not collide with existing forms as it requires new classes and DOM structure. All Angular form controls will be based upon this soon.

TODO

* [ ] Polish up variables and organization
* [ ] UX Review from @reddolan 
* [ ] Test in Chrome
* [ ] Test in Firefox
* [ ] Test in Safari
* [ ] Test in IE11
* [ ] Test in Edge


adityarb88:
image

just starting to review but noticed that the alignment of validation in compact forms is off.

adityarb88:
One important thing that we are missing is Vertical Rhythm.

Current Forms:
image


New Forms:
image


I am not sure if the spec has changed or not but the row height of 42.5 is definitely incorrect. It should be divisible by 24px (1.25, 1.5, 1.75rem and so on is fine too).

Also, note that the height of the label and corresponding input fields is 24px in the current forms:
image

image

That helps to vertically align the row. We just adjust the margins to differentiate between regular and compact forms.

The new forms fields have different heights:
image
image


One thing that is incorrect in the current forms is that it uses margin-bottom on each form-groups. We should use margin-top which is consistent with all other components in Clarity.
image

src/app/forms/forms.demo.ts
```diff
@@ -0,0 +1,25 @@
...
+@Component({
+    selector: "clr-forms-demo",
+    styles: [`.list-nav { columns: 3; margin-bottom: 20px; }`],
+    template: `
+        <h2>Forms</h2>
+        <ul class="list-nav">
+            <li><a [routerLink]="['./input-vertical']">Input Vertical</a></li>
+            <li><a [routerLink]="['./input-horizontal']">Input Horizontal</a></li>
+            <li><a [routerLink]="['./input-compact']">Input Compact</a></li>
+            <li><a [routerLink]="['./input-vertical']">Input Vertical Grid</a></li>
```
adityarb88: `routerLinks` are incorrect. should be `input-vertical-grid` here.

src/app/forms/forms.demo.ts
```diff
@@ -0,0 +1,25 @@
...
+            <li><a [routerLink]="['./input-horizontal']">Input Horizontal Grid</a></li>
+            <li><a [routerLink]="['./input-compact']">Input Compact Grid</a></li>
```
adityarb88: same for these two.

adityarb88: Would be able to update: https://clarity-forms.surge.sh/ with the correct demos so that I can review the correct demos?

gnomeontherun: Updated

src/app/forms/forms.demo.ts
gnomeontherun: Fixed

src/app/forms/forms.demo.ts
```diff
@@ -0,0 +1,25 @@
...
+    template: `
+        <h2>Forms</h2>
+        <ul class="list-nav">
+            <li><a [routerLink]="['./input-vertical']">Input Vertical</a></li>
+            <li><a [routerLink]="['./input-horizontal']">Input Horizontal</a></li>
+            <li><a [routerLink]="['./input-compact']">Input Compact</a></li>
+            <li><a [routerLink]="['./input-vertical-grid']">Input Vertical Grid</a></li>
+            <li><a [routerLink]="['./input-horizontal-grid']">Input Horizontal Grid</a></li>
+            <li><a [routerLink]="['./input-compact-grid']">Input Compact Grid</a></li>
+            <li><a [routerLink]="['./input-group']">Input Group</a></li>
```
adityarb88: Do you think we should add a demo here which tests if popovers are being clipped inside of forms? I remember 2 main issues:
1. Angular tooltip component in forms in a modal
2. Datepicker in forms in a modal.

 but the Forms Refactor Epic has more details.

gnomeontherun: Added modal examples, if you use a grid in a modal it still clips.

gnomeontherun:
Finally got it passing that pesky gemini test.

src/app/forms/forms.demo.ts
adityarb88: Thank you!

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
+<!--
+  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+
+<h1>Layout - {{layout}} - {{grid ? 'grid' : 'no grid'}}</h1>
+
+<form class="clr-form clr-form-{{layout}}">
```
adityarb88: The current forms have a title/label for the complete form. Not sure if we have that for the new forms or not but just wanted to note that here. cc: @reddolan 

image

image

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
...
+    <div class="clr-form-control" [ngClass]="{row: grid}">
+        <label for="{{layout}}-basic" class="clr-control-label" [ngClass]="{'col-xs-12': grid, 'col-lg-2': grid && layout !== 'vertical'}">
+            Basic input
+        </label>
+        <div class="clr-control-container" [ngClass]="{'col-xs-12': grid, 'col-lg-10': grid && layout !== 'vertical'}">
+            <div class="clr-input-wrapper">
+                <input type="text" id="{{layout}}-basic" placeholder="Enter value here" class="clr-input">
+                <clr-icon class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
+            </div>
+            <span class="clr-subtext">Helper Text</span>
```
adityarb88: I am trying to remember what we decided about the `.clr-subtext`. Was it going to be on the `input` level or the `.clr-control-container` level? Based on that it should be in or outside of the `.clr-input-wrapper` right?

From the current implementation, it looks like if there are multiple inputs in the same row, each input can't have a subtext for it. Please correct me if I am not understanding this correctly.

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
...
+        </div>
+    </div>
+
+    <div class="clr-form-control" [ngClass]="{row: grid}">
+        <label class="clr-control-label" [ngClass]="{'col-xs-12': grid, 'col-lg-2': grid && layout !== 'vertical'}">
+            Basic checkbox</label>
+        <div class="clr-control-container" [ngClass]="{'col-xs-12': grid, 'col-lg-10': grid && layout !== 'vertical'}">
+            <div class="clr-checkbox-wrapper">
```
adityarb88: Are we breaking the checkboxes too? The current checkboxes have this markup:

```
<div class="checkbox">
      <input type="checkbox" id="checkrads_1">
      <label for="checkrads_1">Checkbox 1</label>
</div>
```

The new ones though have `.clr-checkbox-wrapper`, wrapping multiple checkboxes and each individual checkbox not having a wrapper around it.

Is the reason for this change simplicity of the markup?

adityarb88: Are the new forms compatible with the old style of the markup? @youdz's implementation of the Angular component still has a wrapper around each individual checkbox: https://github.com/vmware/clarity/blob/master/src/clr-angular/forms/checkbox/checkbox-container.spec.ts#L17

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
...
+                <label for="{{layout}}-checkbox2" class="clr-control-label">
+                    option 2
+                </label>
+                <input type="checkbox" name="checkbox3" id="{{layout}}-checkbox3" placeholder="Enter value here" class="clr-checkbox"> 
+                <label for="{{layout}}-checkbox3" class="clr-control-label">
+                    option 3
+                </label>
+            </div>
+            <!-- IMPORTANT DIFFERENCE IN STRUCTURE! ICON IS NOT PART OF THE INPUT WRAPPER -->
+            <clr-icon class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
```
adityarb88: Does the icon have to be outside? What would happen here if we used inline checkboxes here? I think we should add a demo for inline checkboxes in all form layouts to test this out.

adityarb88: I mean inline checkboxes with validation.

adityarb88: saw the demo below

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
...
+            <span class="clr-subtext">Helper Text</span>
+        </div>
+    </div>
+    
+    <div class="clr-form-control" [ngClass]="{row: grid}">
+        <label class="clr-control-label" [ngClass]="{'col-xs-12': grid, 'col-lg-2': grid && layout !== 'vertical'}">
+            Inline checkbox</label>
+        <div class="clr-control-container" [ngClass]="{'col-xs-12': grid, 'col-lg-10': grid && layout !== 'vertical'}">
```
adityarb88: Missing `.clr-error` class. Noting this because you have the error icon at the end.

src/app/forms/input/input.html
```diff
@@ -0,0 +1,356 @@
...
+            Basic textarea, error</label>
+        <div class="clr-control-container clr-error" [ngClass]="{'col-xs-12': grid, 'col-lg-10': grid && layout !== 'vertical'}">
+            <div class="clr-textarea-wrapper">
+                <textarea id="{{layout}}-textarea-basic-error" rows="5" placeholder="Enter value here" class="clr-textarea"></textarea>
+                <clr-icon class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
+            </div>
+            <span class="clr-subtext">Helper Text</span>
+        </div>
+    </div>
+</form>
```
adityarb88: Could you please add toggles in the demo too?

adityarb88: Native select and multi select is missing too.

src/clr-angular/layout/_grid.clarity.scss
```diff
@@ -7,4 +7,10 @@
         margin-left: 0;
         margin-right: 0;
     }
+
+    // Override grid, helps with issues for popovers and such
+    // See https://github.com/twbs/bootstrap/pull/25255
+    .col-lg, .col-lg-1, .col-lg-2, .col-lg-3, .col-lg-4, .col-lg-5, .col-lg-6, .col-lg-7, .col-lg-8, .col-lg-9, .col-lg-10, .col-lg-11, .col-lg-12, .col-md, .col-md-1, .col-md-2, .col-md-3, .col-md-4, .col-md-5, .col-md-6, .col-md-7, .col-md-8, .col-md-9, .col-md-10, .col-md-11, .col-md-12, .col-sm, .col-sm-1, .col-sm-2, .col-sm-3, .col-sm-4, .col-sm-5, .col-sm-6, .col-sm-7, .col-sm-8, .col-sm-9, .col-sm-10, .col-sm-11, .col-sm-12, .col-xl, .col-xl-1, .col-xl-2, .col-xl-3, .col-xl-4, .col-xl-5, .col-xl-6, .col-xl-7, .col-xl-8, .col-xl-9, .col-xl-10, .col-xl-11, .col-xl-12, .col-xs, .col-xs-1, .col-xs-2, .col-xs-3, .col-xs-4, .col-xs-5, .col-xs-6, .col-xs-7, .col-xs-8, .col-xs-9, .col-xs-10, .col-xs-11, .col-xs-12 {
```
adityarb88: I know a few things break with `position: static` but could we just use something like:
https://stackoverflow.com/a/22509502/3538394

instead of writing all the classes down in case we decide to keep this style?

src/clr-angular/forms/styles/_form.clarity.scss
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-form {
```
adityarb88: Wow. Really appreciate you separating the styles into their individual files. Thank you!

BTW each file should use the exports mixin. Eg: https://github.com/vmware/clarity/blob/master/src/clr-angular/button/_buttons.clarity.scss#L92

src/clr-angular/forms/styles/_form.clarity.scss
```diff
@@ -0,0 +1,11 @@
...
+  padding: 6px;
```
adityarb88: rem

src/clr-angular/forms/styles/_file.clarity.scss
```diff
@@ -0,0 +1,42 @@
...
+        .btn {
+            margin: 0;
+        }
+    }
+}
+
+.clr-form-compact {
+    .clr-form-control {
+        .clr-control-container {
+            .clr-file-wrapper {
```
adityarb88: Does the nesting here have to be 4 levels deep? Could we just do? 
```
-clr-form-compact
--clr-file-wrapper
----styles
```?

File: src/clr-angular/forms/styles/_file.clarity.scss [hunk 14]
```diff
@@ -0,0 +1,42 @@
...
+                @include form-flatten-compact-adjustments();
+
+                & ~ .clr-validate-icon {
+                    margin-top: 0.1rem;
+                }
+                & ~ .clr-subtext {
+                    line-height: 24px;
```
adityarb88: 1rem

src/clr-angular/forms/styles/_file.clarity.scss
```diff
@@ -0,0 +1,42 @@
...
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-control-container {
+    .clr-file-wrapper {
+        @include form-flatten-validate-text();
+        position: relative;
+        margin-top: $clr-forms-baseline;
+
+        .clr-file {
+            width: 0.1px;
```
adityarb88: wouldn't `height: 0` and `width: 0` work?

adityarb88:
image

image


I see this happening in some screen sizes. I really think we should add responsiveness by default.
A fixed width on the form control label will also mitigate such issues, no?

src/app/forms/input/input.html
gnomeontherun: The `.clr-subtext` is always a child of `.clr-control-container`. You can have multiple `.clr-control-containers` in a single `.clr-form-control` to get multiple inputs into the same row. Each optionally has its own subtext that stays with it.

adityarb88: good to know. Thanks! Do you think we can add an example which demos this?

src/app/forms/input/input.html
gnomeontherun: I've looking into this a little bit, but need to do more prototyping to verify how they work together.

src/app/forms/input/input.html
gnomeontherun: I've put the error icon everywhere, even if its not active. The ones with the `.clr-error` class are the only ones that should show the error icon, so I needed to verify the icon was hidden by default.

src/clr-angular/layout/_grid.clarity.scss
gnomeontherun: Yeah, didn't get fancy because I just copied this from Bootstrap. We need to decide on our solution with grids first.

src/clr-angular/forms/styles/_form.clarity.scss
gnomeontherun: fixed

src/clr-angular/forms/styles/_form.clarity.scss
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.clr-form-group {
+  padding: 6px;
```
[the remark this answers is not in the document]
gnomeontherun: fixed

src/clr-angular/forms/styles/_file.clarity.scss
gnomeontherun: fixed

src/clr-angular/forms/styles/_file.clarity.scss
gnomeontherun: looks like, some research I saw said it had to be over 0, but need to verify cross-browser

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-control-container {
+    .clr-checkbox-wrapper {
+        @include form-flatten-validate-text();
+
+        position: relative;
```
adityarb88: I am assuming we added this because if someone had to add a tooltip, the would add it outside of the `.clr-checkbox-wrapper` right?

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-control-container {
```
adityarb88: Cant we extract the basic styles outside of the `.clr-control-container` selector? Positioning, margin styles can go inside of the container but other styles can be moved out of it right?

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,102 @@
...
+
+    .clr-checkbox-wrapper.clr-checkbox-inline {
+        @include form-inline-input();
+    }
+}
+
+.clr-form-compact {
+    .clr-form-control {
+        .clr-control-container {
+            .clr-checkbox-wrapper {
```
adityarb88: same here. does this have to be so deeply nested?

src/clr-angular/forms/styles/_form.clarity.scss
gnomeontherun: done

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-form-control {
+    padding-top: $clr-forms-baseline * 4;
+
+    .clr-control-label {
```
adityarb88: same nesting question here. can a few of these styles be moved out? I will just note this once here but this applies basically to all nested styles.

src/app/forms/input/input.html
gnomeontherun: Semantically, that label is just a piece of text like any header. I don't know why we'd need to do anything special to handle this case, just use header elements.

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,98 @@
...
+        display: block;
+        color: $clr-forms-label-color;
+        font-size: calc(#{$clr-forms-baseline} * 2 + 1px);
```
adityarb88: @mathisscott created a mixin and a map where all font styles are added. More info here:
https://github.com/vmware/clarity/blob/master/src/clr-angular/button/_buttons.clarity.scss#L96
https://github.com/vmware/clarity/blob/master/src/clr-angular/utils/_mixins.clarity.scss#L6
https://github.com/vmware/clarity/blob/master/src/clr-angular/utils/_maps.clarity.scss#L568

Can we move this there? Not sure if we can because you are using variables here but just making a note to check if we can.

adityarb88: The map makes sure that we are following the typography recommendations from UX.

src/clr-angular/forms/styles/_input-group.clarity.scss
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+
+.clr-control-container {  
+    .clr-input-group {
```
adityarb88: nesting. can be moved out of `.clr-control-container`

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-form-control {
+    padding-top: $clr-forms-baseline * 4;
```
adityarb88: Shouldn't this be `margin-top` instead?

src/clr-angular/forms/styles/_checkbox.clarity.scss
gnomeontherun: This looks like an error, removing it.

src/clr-angular/forms/styles/_checkbox.clarity.scss
gnomeontherun: I did this intentionally to ensure that people used the correct DOM structure. Let's chat about this more.

src/clr-angular/forms/styles/_containers.clarity.scss
gnomeontherun: fixed

src/clr-angular/forms/styles/_input.clarity.scss
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+.clr-control-container {
+    .clr-input-wrapper {
+        white-space: nowrap;
+        height: calc(#{$clr-forms-baseline} * 4 + 1px); // This is strange but makes the vertical rhythm work properly
```
adityarb88: I saw this issue in the demo. So weird. Not sure whats going on here.

Just a note that using display flex solves this issue. But we might have to test the whole thing again. We can then remove this hack.

gnomeontherun: Looks like it fixes things, but yes need to do validation.

gnomeontherun:
It looks like a `min-width` and `width` on the label in compact keeps the alignment. If the width of the row exceeds the page width it is scrollable.

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,101 @@
...
+        .clr-form-control {
+            display: flex;
+            flex-direction: row;
+            padding-top: $clr-forms-baseline * 2;
+
+            .clr-control-container {
+                display: flex;
+            }
+            
+            .clr-subtext {
```
youdz: This is the one I'm mentioning in my comment above. Wouldn't `clr-form-compact .clr-subtext` be just as good for us and way easier to override for our consumers?

src/clr-angular/forms/styles/_radio.clarity.scss
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    .clr-form-compact {
+        .clr-form-control {
+            .clr-control-container {
+                .clr-radio-wrapper {
+                    @include form-flatten-compact-adjustments();
+                    @include form-inline-input();
+
+                    .clr-control-label {
```
youdz: This one is pretty heavily nested too.

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.control.checkbox') {
+    .clr-control-container {
+        .clr-checkbox-wrapper {
```
mathisscott: Our rule of thumb is to have CSS selectors nested no more than two deep. We have some that are 3-4 deep but those are the exceptions.

Here `.clr-control-container .clr-checkbox-wrapper input[?]` could just as easily be `.clr-checkbox-wrapper input[?]`.

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,102 @@
...
+                }
+            }
+        }
+
+        .clr-checkbox-wrapper.clr-checkbox-inline {
+            @include form-inline-input();
+        }
+    }
+
+    .clr-form-compact {
```
mathisscott: `.clr-form-compact .clr-form-control .clr-control-container .clr-checkbox-wrapper ?`

...to...

`.clr-form-compact .clr-checkbox-wrapper ?`

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,101 @@
...
+            color: $clr-forms-label-color;
+            font-size: calc(#{$clr-forms-baseline} * 2 + 1px);
+            font-weight: 600;
+            line-height: $clr-forms-baseline * 3;
+        }
+
+        .clr-control-container {
+            display: inline-block;
+
+            .clr-subtext {
```
mathisscott: `.clr-form-control .clr-control-container .clr-subtext` et al.

...could be better as...

`.clr-control-container .clr-subtext` or `.clr-form-control .clr-subtext` et al.

No need to use css selectors for DOM mapping.

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        .clr-form-control {
+            display: flex;
+            flex-direction: row;
+        }
+    }
+
+    .clr-form-compact {
+        .clr-form-control > .clr-control-label {
```
mathisscott: Is the direct descendant selector here necessary?

src/clr-angular/forms/styles/_file.clarity.scss
```diff
@@ -0,0 +1,44 @@
...
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.control.file') {
+    .clr-control-container {
+        .clr-file-wrapper {
+            @include form-flatten-validate-text();
+            position: relative;
+            margin-top: $clr-forms-baseline;
+
+            .clr-file {
```
mathisscott: `.clr-control-container .clr-file-wrapper .clr-file`

...to...

`.clr-file-wrapper .clr-file`

src/clr-angular/forms/styles/_input-group.clarity.scss
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.control.input-group') {
+    .clr-control-container {  
+        .clr-input-group {
```
mathisscott: Same here `.clr-control-container` is an unnecessary nesting...

src/clr-angular/forms/styles/_containers.clarity.scss
youdz: Yes, that one is. It catches the "general control label" that is displayed on the left-hand side in horizontal forms, as opposed to more targeted per-input label that would be on the right-hand side in the case of multiple inputs (checkboxes, radios, multiple text inputs, ...)

src/clr-angular/utils/_components.clarity.scss
```diff
@@ -98,14 +98,25 @@
...
+@import "../forms/styles/form.clarity";
+@import "../forms/styles/checkbox.clarity";
+@import "../forms/styles/file.clarity";
+@import "../forms/styles/input.clarity";
+@import "../forms/styles/input-group.clarity";
+@import "../forms/styles/radio.clarity";
+@import "../forms/styles/textarea.clarity";
+
+//Forms Deprecated
+// @import "../forms-deprecated/utils/form-variables.clarity";
```
adityarb88: Why are we commenting this out? If we are going to merge this into master, we want the old forms to work too right?

src/clr-angular/forms/styles/_mixins.forms.scss
```diff
@@ -90,6 +90,7 @@
     //and so that the user's default input size doesn't affect it
     height: $size;
     width: $size;
+    display: none;
```
youdz: @adityarb88: Wasn't `display: none` a blocker for accessibility? I don't remember it it was screenreaders that couldn't work with it.

reddolan:
I couldn't find the comment, but I just saw the comment in a GitHub notification email. 
`display:none;` isn't accessible. It is recommended to push it off screen
> Your regular hide class should position the content off screen, which still leaves it screen reader accessible:
[https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/](https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/)

src/clr-angular/forms/styles/_mixins.forms.scss
adityarb88: Yes `display: none` is a blocked on accessibility.  I believe we used `opacity: 0` before. @gnomeontherun why did we change this to `display: none`? Are you still able to checkbox of the input using the keyboard?

youdz:
@reddolan: The comment is the one considered "outdated", right above you. I commented on a specific commit, but even though the line of code I pointed to didn't change in the latest version, GitHub marks it as outdated. Weird, and probably my mistake.

reddolan:
@youdz no worries. I just wanted to make sure I could help out seeing as how this is an important issue and didn't want it to get blocked by something I could answer

src/clr-angular/forms/styles/_mixins.forms.scss
gnomeontherun: This is a mistake, Firefox was still calculating a size for it so the tests failed. Will fix.

src/app/forms/controls/checkbox.html
```diff
@@ -0,0 +1,154 @@
...
+    </div>
+
+    <div class="clr-form-control">
+        <label class="clr-control-label">Multi wrappers</label>
+        <div class="clr-control-container">
+            <div class="clr-checkbox-wrapper">
+                <input type="checkbox" id="multi-wrappers-1" class="clr-checkbox">
+                <label for="multi-wrappers-1" class="clr-control-label">Label</label>
+            </div>
+            <div class="clr-checkbox-wrapper clr-error">
```
adityarb88: I see that you are using `.clr-error` here but the demo doesn't reflect that.
image

src/app/forms/controls/radio.html
```diff
@@ -0,0 +1,179 @@
...
+      <div class="clr-form-control">
+          <label class="clr-control-label">
+              Multi wrappers
+          </label>
+          <div class="clr-control-container">
+              <div class="clr-radio-wrapper">
+                  <input type="radio" id="multi-1" class="clr-radio">
+                  <label for="multi-1">Label with for</label>
+              </div>
+              <div class="clr-radio-wrapper clr-error">
```
adityarb88: Same question that I asked for checkboxes for `.clr-error` here. I don't see the error in the demo.

src/app/forms/controls/select.html
```diff
@@ -0,0 +1,123 @@
...
+<h2>Select</h2>
+
+<form class="clr-form">
+    <div class="clr-form-control">
+        <label class="clr-control-label">
+            Basic
+        </label>
+        <div class="clr-control-container">
+            <div class="clr-select-wrapper">
+                <div class="clr-select">
```
adityarb88: Do we need both `.clr-select-wrapper` and `.clr-select`? The error Icon can go inside of the `.clr-select` div after the `select`, no?

src/app/forms/controls/select.html
```diff
@@ -0,0 +1,123 @@
...
+        </div>
+    </div>
+
+    <div class="clr-form-control">
+        <label class="clr-control-label">
+            Basic multiple
+        </label>
+        <div class="clr-control-container">
+            <div class="clr-multiselect-wrapper">
+                <div class="clr-select">
```
adityarb88: Same question about multiple wrappers here.

src/app/forms/controls/text.html
```diff
@@ -0,0 +1,58 @@
...
+<h2>Text Input</h2>
+
+<form class="clr-form">
+    <div class="clr-form-control">
+        <label for="basic" class="clr-control-label">
+            Basic
+        </label>
+        <div class="clr-control-container">
+            <div class="clr-input-wrapper">
+                <input type="text" id="basic" placeholder="Enter value here" class="clr-input">
```
adityarb88: Did we try using the `input` without the wrapper to see if it renders correctly? Also it should render the same if the user doesn't add `type="text"` because the default type is `text`. This was a bug on our side before.

gnomeontherun: This doesn't work completely, but making a backlog note to address in a future iteration.

src/app/forms/controls/select.html
gnomeontherun: This was the only way to keep the `.clr-select` element from having sizing issues from other elements in the flex layouts.

src/clr-angular/data/stack-view/_variables.stack-view.scss
```diff
@@ -17,3 +17,4 @@ $clr-stack-view-stack-block-expanded-bg-color: $clr-global-selection-color;
 $clr-stack-view-stack-block-expandable-hover: $gray-light;
 $clr-stack-view-stack-block-content-text-color: inherit;
 $clr-stack-view-stack-block-expanded-text-color: $clr-black;
+$styledInputs: inputs("text", "password", "number", "email", "url", "tel", "date", "time", "datetime-local");
```
adityarb88: Why was this moved into StackView?

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,114 @@
...
+            border: $clr-default-borderwidth solid $clr-forms-checkbox-indeterminate-border-color;
+        }
+        input[type="checkbox"].clr-indeterminate + label::after {
+            border-left: none;
+            border-bottom-color: $clr-forms-checkbox-indeterminate-border-color;
+            display: inline-block;
+            transform: translate(0, $mark-half-size);
+        }
+
+        label > input[type="checkbox"]::after {
```
adityarb88: Something doesnt look right here. I dont think the `input` element has an `::after` pseudo element, no?

Is this `input[type="checkbox"] > label::after` instead of what you have here currently?

adityarb88: Wait the styles look repeated on line 35. Line 35 has the correct selector. This one looks incorrect.

adityarb88: https://github.com/vmware/clarity/pull/2068/files#diff-c6fa1884c13e1bcc0567ed5b96b0c75cR35

src/clr-angular/forms/styles/_checkbox.clarity.scss
```diff
@@ -0,0 +1,114 @@
...
+            border-color: $clr-forms-checkbox-mark-color;
+        }
+
+        &.clr-checkbox-inline {
+            @include form-inline-input();
+        }
+    }
+
+    .clr-form-control-disabled .clr-checkbox-wrapper {
+        input[type="checkbox"] + label::before, input[type="checkbox"]:checked + label::before {
```
adityarb88: Do we need 
```
.clr-form-control-disabled .clr-checkbox-wrapper {
      input[type="checkbox"] + label::before, input[type="checkbox"]:checked + label::before {
```
here?

We could just use:
```
.clr-form-control-disabled {
     input[type="checkbox"] + label::before, input[type="checkbox"]:checked + label::before {
```

no?

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,105 @@
...
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.containers') {
+    .clr-form-control {
+        margin-top: $clr-forms-baseline * 4;
+        display: flex;
+        flex-direction: column;
+
+        &.clr-form-control-disabled {
```
adityarb88: This could be out of the nesting. The styles assigned to this selector look common across forms, no?

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,105 @@
...
+            opacity: $clr-forms-disabled-opacity;
+        }
+
+        &.clr-form-control-multi {
```
adityarb88: Same question about nesting here.

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -0,0 +1,105 @@
...
+        color: $clr-forms-label-color;
+        @include clr-getTypePropertiesForDomElement(form_label, (font-size));
+        font-weight: 600;
+        line-height: $clr-forms-baseline * 3;
+    }
+
+    .clr-control-container {
+        display: inline-block;
+
+        .clr-subtext {
```
adityarb88: Here too. Do we need to target `.clr-subtext`, `.clr-validate-icon` and `.clr-error` inside of `.clr-control-container` only? Or are these styles common across forms. If they are common, we can remove the nesting.

src/clr-angular/forms/styles/_file.clarity.scss
```diff
@@ -0,0 +1,43 @@
...
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.control.file') {
+    .clr-file-wrapper {
+        @include form-flatten-validate-text();
+        position: relative;
+        margin-top: $clr-forms-baseline;
+
+        .clr-file {
```
adityarb88: Can move out of `.clr-file-wrapper`

src/clr-angular/forms/styles/_input.clarity.scss
```diff
@@ -0,0 +1,46 @@
...
+                border: none;
+            }
+            
+            &:not([readonly]) {
+                @include input-border-bottom-animation($clr-forms-focused-color);
+            }
+        }
+    }
+
+    .clr-control-container.clr-error .clr-input-wrapper .clr-input {
```
adityarb88: If we end up removing the above nesting, we could just use:

`.clr-control-container.clr-error .clr-input`

instead of `.clr-control-container.clr-error .clr-input-wrapper .clr-input`

src/clr-angular/forms/styles/_textarea.clarity.scss
```diff
@@ -0,0 +1,61 @@
...
+        white-space: nowrap;
+        display: flex;
+        margin-top: $clr-forms-baseline;
+        
+        .clr-validate-icon {
+            margin-left: 0;
+        }
+    }
+    
+    .clr-textarea {
```
adityarb88: This pattern looks good to me. Nesting only when its needed.

src/clr-angular/forms/styles/_file.clarity.scss
gnomeontherun: We can, but this does require having a wrapper for it to work.

src/clr-angular/data/stack-view/_variables.stack-view.scss
gnomeontherun: It is used here and in the old forms, but not used in the new forms. I moved it to keep it alive when the old forms are removed.

src/clr-angular/forms/styles/_file.clarity.scss
adityarb88: I think thats fine. The advantage of removing the nesting is that people can target `.clr-file` directly instead of having to write `.clr-file-wrapper .clr-file` if they want to customize something.

src/clr-angular/forms/styles/_checkbox.clarity.scss
gnomeontherun: This is leftover from attempting to support inputs inside of labels, and I'll remove it for now.

src/clr-angular/forms/styles/_checkbox.clarity.scss
gnomeontherun: Only if we unnest the input selectors in the top of this file. This is not an option at the moment, as it would collide with existing forms.

src/clr-angular/forms/styles/_containers.clarity.scss
gnomeontherun: Unnested, no functional change for the user.

src/clr-angular/forms/styles/_containers.clarity.scss
gnomeontherun: Can unnest, same reason.

src/clr-angular/forms/styles/_checkbox.clarity.scss
adityarb88: But the existing forms dont use `.clr-form-control-disabled` right? I am just saying that we can get rid of `.clr-checkbox-wrapper`.

src/clr-angular/forms/styles/_containers.clarity.scss
gnomeontherun: Unnested. They are common and unless we used this selector elsewhere it should be fine.

src/clr-angular/forms/styles/_file.clarity.scss
gnomeontherun: Done.

src/clr-angular/forms/styles/_input.clarity.scss
gnomeontherun: Done.

gnomeontherun:
@youdz That is correct, per the design only the underline is removed and the rest of the display remains the same. Do you have a concern about this? We can talk to UX on that.

reddolan:
@youdz @gnomeontherun yes that is correct. We wanted the read only to look like a form input that had a value so that it tied into the other form input value design

src/app/forms/controls/checkbox.html
gnomeontherun: You have to put `clr-error` on a `clr-control-container` for it to work, so this was a typo I fixed up. The example following is correct, since it shows the full container.

src/app/forms/controls/radio.html
gnomeontherun: Fixed.

src/app/forms/controls/select.html
gnomeontherun: As we reviewed, we got this to one level now :)

src/app/forms/controls/select.html
gnomeontherun: fixed

src/clr-angular/forms/styles/_input.clarity.scss
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('forms.control.input') {
+    .clr-input-wrapper {
+        white-space: nowrap;
+        height: $clr-forms-baseline * 4;
+        
+        .clr-input {
```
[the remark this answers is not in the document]
gnomeontherun: Fixed.

src/clr-angular/forms/styles/_checkbox.clarity.scss
gnomeontherun: No, because the specificity of the original values that I need to change have `.clr-checkbox-wrapper` in there, because without it new forms would conflict with old forms selectors.

src/clr-angular/forms/styles/_textarea.clarity.scss
gnomeontherun: 👍

src/clr-angular/forms/styles/_mixins.forms.scss
```diff
@@ -0,0 +1,144 @@
...
+    position: absolute;
+    opacity: 0;
+
+    //Display
+    //NOTE: Even though the input is invisible, we assign it a height
+    //and width so that it aligns right behind the fake checkbox
+    //and so that the user's default input size doesn't affect it
+    height: $size;
+    width: $size;
+    display: none;
```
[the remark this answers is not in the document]
gnomeontherun: Removed.

PR #2070, opened 2018-03-14
[ICONS] add vol.10 icons

Shijir:


src/clr-icons/interfaces/icon-alias.ts
```diff
@@ -1,6 +0,0 @@
-/*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
- * This software is released under MIT license.
- * The full license information can be found in LICENSE in the root directory of this project.
- */
-export interface IconAlias { [key: string]: string[]; }
```
adityarb88: I think `IconAlias` and `IconTemplate` interfaces are removed because they aren't being used anymore. Is that correct?

src/clr-icons/shapes/chart-shapes.ts
```diff
@@ -0,0 +1,238 @@
...
+  <circle class="clr-i-solid--badged clr-i-solid-path-2--badged clr-i-badge" cx="30" cy="6" r="5"/>
+  <path class="clr-i-outline--alerted clr-i-outline-path-1--alerted" d="M 34 29 C 34 30.105 33.105 31 32 31 L 4 31 C 2.895 31 2 30.105 2 29 L 2 7 C 2 5.895 2.895 5 4 5 L 21.958 5 L 20.786 7 L 4 7 L 4 29 L 32 29 L 32 15.357 L 34 15.357 Z"/>
+  <path class="clr-i-outline--alerted clr-i-outline-path-2--alerted" d="M 6 25 L 8 25 L 8 22 L 10 22 L 10 25 L 13 25 L 13 22 L 15 22 L 15 25 L 18 25 L 18 22 L 20 22 L 20 25 L 23 25 L 23 22 L 25 22 L 25 25 L 27.723 25 C 28.023 25.02 28.293 25.18 28.463 25.43 C 28.903 26.06 28.483 26.93 27.723 26.99 L 6 26.991 Z"/>
+  <path class="clr-i-outline--alerted clr-i-outline-path-3--alerted clr-i-alert" d="M 26.854 1.144 L 21.134 11.004 C 20.579 11.818 21.114 12.928 22.097 13.001 C 22.142 13.005 22.188 13.006 22.234 13.004 L 33.684 13.004 C 34.669 13.036 35.319 11.991 34.855 11.122 C 34.834 11.081
```
adityarb88: I think this should be `heat-map` and not `heat-chart`

adityarb88: should be `scatter-plot` and not `scatter-chart`

adityarb88: `tick-chart` instead of `ticks-chart`. Confirm with @lil-kim? I checked here: https://www.investopedia.com/articles/trading/10/data-based-intraday-chart-intervals.asp

adityarb88: should be `box-plot` instead of `box-chart`

src/clr-icons/shapes/essential-shapes.ts
```diff
@@ -813,7 +827,14 @@ export const EssentialShapes: any = {
     "accessibility-2": ClrShapeAccessibility2,
     "sort-by": ClrShapeSortBy,
     "collapse-card": ClrShapeCollapseCard,
-    "expand-card": ClrShapeExpandCard
+    "expand-card": ClrShapeExpandCard,
+    "briefcase": ClrShapeBriefcase,
+    "color-picker": ClrShapeColorPicker,
+    "copy-to-clipboard": ClrShapeCopyToClipboard,
```
adityarb88: shouldn't this just be `copy`?

src/clr-icons/shapes/essential-shapes.ts
```diff
@@ -813,7 +827,14 @@ export const EssentialShapes: any = {
...
+    "filter-2": ClrShapeFilter2,
```
adityarb88: Why `filter-2`? Are we adding icons like `name-1`, `name-2` etc?

src/clr-icons/shapes/media-shapes.ts
```diff
@@ -119,7 +127,11 @@ export const MediaShapes: any = {
     "headphones": ClrShapeHeadphones,
     "film-strip": ClrShapeFilmStrip,
     "music-note": ClrShapeMusicNote,
-    "wand": ClrShapeWand
+    "wand": ClrShapeWand,
```
adityarb88: Why is the `wand` icon in media shapes?

src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -1112,7 +1118,10 @@ export const TechnologyShapes: any = {
     "hard-drive": ClrShapeHardDrive,
     "hard-drive-disks": ClrShapeHardDriveDisks,
     "nvme": ClrShapeNVMe,
-    "ssd": ClrShapeSSD
+    "ssd": ClrShapeSSD,
+    "bluetooth": ClrShapeBluetooth,
+    "no-bluetooth": ClrShapeNoBluetooth,
```
adityarb88: isn't `bluetooth-off` better here? You do the same with `alarm-off`. Two different naming conventions. We need to choose one

src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -1112,7 +1118,10 @@ export const TechnologyShapes: any = {
...
+    "process-on-vm": ClrShapeProcessOnVM
```
adityarb88: I dont know what this icon represents. Is there a better name for this? @lil-kim

src/clr-icons/interfaces/icon-alias.ts
Shijir: They are moved to `icon-interfaces`.

src/clr-icons/shapes/chart-shapes.ts
Shijir: Agree, I think `heat-map-chart` is better.

adityarb88: we dont even need a `chart` at the end. its just called a `heat-map`

src/clr-icons/shapes/essential-shapes.ts
Shijir: We already have `copy`. Maybe, it should be `copy-2` instead?

src/clr-icons/shapes/essential-shapes.ts
Shijir: we have `accessibility-1` and `accessibility-2`.

src/clr-icons/shapes/chart-shapes.ts
lil-kim: --If a user searches for "chart" it would be nice to see all of these types of charts popup. That's the only reason I thought it should be in the name. 

--Heat-map-chart is better than heat-chart.

lil-kim: Aditya is right. It should be "tick"

src/clr-icons/shapes/essential-shapes.ts
lil-kim: This icon is a more modern update to the old style of filter we currently have. We have some other instances that called for a second version. name-1, name-2 is the convention we have been using so far.

src/clr-icons/shapes/chart-shapes.ts
Shijir: 🐞 🐞 🐞 🐞 🐞 💹

adityarb88: But we can add an alias to this right? So the name can be `heat-map` and the alias can contain `chart`, that way it shows up in the search. Is that right @Shijir?

src/clr-icons/shapes/media-shapes.ts
lil-kim: Yeah, that does seem odd. My original thinking was that it could be used for photo editing or something like that. It might make more sense to put it in Essentials - I'm OK with that.

src/clr-icons/shapes/technology-shapes.ts
lil-kim: I like Aditya's suggestions. 👍

Shijir:
@adityarb88 @lil-kim I have published the icons in KS-app: http://vol10-icons-dark-theme.surge.sh/iconography

src/clr-icons/shapes/technology-shapes.ts
adityarb88: 😎

src/clr-icons/shapes/technology-shapes.ts
lil-kim: That was for a VMW product. It was to be shown when there's a process running in a VM. We don't have to publish this to the external site.

src/clr-icons/shapes/chart-shapes.ts
Shijir: We are avoiding to ship aliases by default. So all aliases will be removed from 0.12.

src/clr-icons/shapes/essential-shapes.ts
Shijir: Ok, I will keep it `copy-to-clipboard`

src/clr-icons/shapes/media-shapes.ts
Shijir: I have moved the shape into essential-set.

Shijir:
With new naming changes: http://vol10-icons.surge.sh/iconography

adityarb88:
@shijir: Have we tested that the icons are published properly? We also need to make sure that the website is updated to add the new sets. We could push this change to 0.11.10 if all these changes are a bit too much to do before tomorrows release.

Shijir:
thanks @adityarb88. I will jump on the website update in minutes. You are right. Website doc will include the new set and new zip artifacts.

PR #2073, opened 2018-03-14
[CORE] add support for schematics to add Clarity to an Angular project

gnomeontherun:
This provides the base tooling and a single action (add) which allows us to take any Angular CLI project and add Clarity to it.

It creates a new package, which we need to test and publish before it can be consumed.


packages/schematics/src/add/index.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Schema as ComponentOptions } from "./schema";
+
+import * as ts from "typescript";
+import * as fs from "fs";
+
+// Looks up and finds the path to the app module (or other module if specified)
+function findModuleFromOptions(host: Tree, options: ComponentOptions): Path | undefined {
+    const modulePath = normalize(
+      '/' + options.sourceDir + '/' + (options.appRoot) + '/' + options.module);
+    const moduleBaseName = normalize(modulePath).split('/').pop();
```
Shijir: Regarding this `.split('/')`, on windows backslashes are used. Would that matter here?

gnomeontherun: As I understand it, we normalize to *nix format and later on its handled for Windows. There is a 'virtual filesystem' in play here, where we apply changes in a file tree and then later merge it with the real file tree (and I believe that is where they handle windows backslash).

packages/schematics/src/collection.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema":
+        "../node_modules/@angular-devkit/schematics/collection-schema.json",
+    "schematics": {
+        "add": {
```
gnomeontherun: I need to set this to `ng-add` to work with the latest CLI beta.

packages/schematics/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+`npm run test` will run the unit tests, using Jasmine as a runner and test framework.
+
+### Publishing
+
+To publish, simply do:
+
+```bash
+npm run build
+npm publish
```
hippee-lee: 1. I don't see a publish script in the `packages/schematics/package.json` file. 
2. Is this how an update for our @clr/schematics is created and pushed to npm?

gnomeontherun: `npm publish` is the normal publish command, so it will publish the files based on the package.json of the current directory. As long as you run the `build` command, it will compile the TypeScript and publish the resulting files. We can do something more automated in time, but its the minimal starting point.

Right now its not hooked into our release cycle, as I don't think we'll want to update this weekly.

packages/schematics/src/add/index.ts
```diff
@@ -0,0 +1,129 @@
...
+            }
+        });
+
+        // Chain a series of tasks to setup Clarity
+        // 1. Add ClarityModule to NgModule
+        // 2. Add BrowserAnimationsModule to NgModule
+        // 3. Run npm install
+        return chain([
+            addDeclarationToNgModule(options, "imports", "ClarityModule", "@clr/angular"),
+            addDeclarationToNgModule(options, "imports", "BrowserAnimationsModule", "@angular/platform-browser/animations"),
```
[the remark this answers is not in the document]
gnomeontherun: We don't need to, the downstream method handles this for us so we can run this command multiple times without issues.

https://github.com/angular/devkit/blob/ad9a30c2ea287d5166c1d997a5ddc8d4a8862d1d/packages/schematics/angular/utility/ast-utils.ts#L268

gnomeontherun:
I had a conversation with someone about the lack of dedicated utilities for schematics yet. They haven't built it up to support virtual file changes in JSON files yet, so this is the recommendation for now.

I actually need to copy any methods I'm borrowing from Angular, as they also mentioned they aren't meant to be a public API.

The long term plan is to improve the public API and make these repetitive tasks easier.

gnomeontherun:
I'm also wondering about our decision to use `@clr/schematics`. Imagine this is the command to add Clarity:

```ng add @clr/schematics```

I think I would prefer to see 

```ng add @clr/angular```

We can always make a separate @clr/schematics package at a later date if we want/need.

packages/schematics/README.md
gnomeontherun: Done

PR #2087, opened 2018-03-19
[WIP | NG] Drag and Drop - create module with event bus

Shijir:
- setup DnD module
- Implemented a service class for Drag and Drop Event Bus.
- Tested the Drag and Drop Event Bus service class.

UML Reference: 
[dnd-UML-class-diagram.pdf](https://github.com/vmware/clarity/files/1826687/dnd-UML-class-diagram.pdf)

### DIFF
--- a/src/clr-angular/utils/drag-and-drop/all.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/all.spec.ts
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import ClrDragAndDropEventBusSpecs from "./providers/drag-and-drop-event-bus.spec";
+describe("Drag And Drop", function() {
+    describe("Providers", function() {
+        ClrDragAndDropEventBusSpecs();
+    });
+    describe("Components And Directives", function() {
+        // TODO: test components and directives
+    });
+});
--- a/src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
+++ b/src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
@@ -1,0 +1,13 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {CommonModule} from "@angular/common";
+import {NgModule} from "@angular/core";
+
+import {ClrDragAndDropEventBus} from "./providers/drag-and-drop-event-bus";
+
+@NgModule({imports: [CommonModule], providers: [ClrDragAndDropEventBus]})
+export class ClrDragAndDropModule {}
--- a/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
+import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
+
+export default function(): void {
+    describe("Drag And Drop Event Bus Provider", function() {
+        const dndEventBus = new ClrDragAndDropEventBus();
+        let isEmitted: boolean;
+
+        beforeEach(function() {
+            isEmitted = false;
+        });
+
+        it("should broadcast from correct Observable on drag start event type", function() {
+            dndEventBus.dragStarted.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
+                expect(event.type).toBe(ClrDragEventType.DRAG_START);
+            });
+            dndEventBus.broadcast({type: ClrDragEventType.DRAG_START});
+            expect(isEmitted).toBeTruthy();
+        });
+
+        it("should broadcast from correct Observable on drag move event type", function() {
+            dndEventBus.dragMoved.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
+                expect(event.type).toBe(ClrDragEventType.DRAG_MOVE);
+            });
+            dndEventBus.broadcast({type: ClrDragEventType.DRAG_MOVE});
+            expect(isEmitted).toBeTruthy();
+        });
+
+        it("should broadcast from correct Observable on drag end event type", function() {
+            dndEventBus.dragEnded.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
+                expect(event.type).toBe(ClrDragEventType.DRAG_END);
+            });
+            dndEventBus.broadcast({type: ClrDragEventType.DRAG_END});
+            expect(isEmitted).toBeTruthy();
+        });
+
+        it("should broadcast from correct Observable on drop event type", function() {
+            dndEventBus.dropped.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
+                expect(event.type).toBe(ClrDragEventType.DROP);
+            });
+            dndEventBus.broadcast({type: ClrDragEventType.DROP});
+            expect(isEmitted).toBeTruthy();
+        });
+    });
+}
--- a/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
+++ b/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Injectable} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+
+import {ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
+
+@Injectable()
+export class ClrDragAndDropEventBus {
+    private dragStart: Subject<ClrDragEvent> = new Subject<ClrDragEvent>();
+    private dragMove: Subject<ClrDragEvent> = new Subject<ClrDragEvent>();
+    private dragEnd: Subject<ClrDragEvent> = new Subject<ClrDragEvent>();
+    private drop: Subject<ClrDragEvent> = new Subject<ClrDragEvent>();
+
+    get dragStarted(): Observable<ClrDragEvent> {
+        return this.dragStart.asObservable();
+    }
+
+    get dragMoved(): Observable<ClrDragEvent> {
+        return this.dragMove.asObservable();
+    }
+
+    get dragEnded(): Observable<ClrDragEvent> {
+        return this.dragEnd.asObservable();
+    }
+
+    get dropped(): Observable<ClrDragEvent> {
+        return this.drop.asObservable();
+    }
+
+
+    broadcast(event: ClrDragEvent): void {
+        switch (event.type) {
+            case ClrDragEventType.DRAG_START:
+                this.dragStart.next(event);
+                break;
+            case ClrDragEventType.DRAG_MOVE:
+                this.dragMove.next(event);
+                break;
+            case ClrDragEventType.DRAG_END:
+                this.dragEnd.next(event);
+                break;
+            case ClrDragEventType.DROP:
+                this.drop.next(event);
+                break;
+            default:
+                break;
+        }
+    }
+}
--- a/src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
+++ b/src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum ClrDragEventType {
+    DRAG_START,
+    DRAG_MOVE,
+    DRAG_END,
+    DROP
+}
+
+export interface ClrDragEvent {
+    // TODO: this interface will be expanded and customized as the implementation progresses.
+    type: ClrDragEventType;
+}


src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export interface ClrDraggableEvent {
+    // TODO: this interface will be expanded and customized as the implementation progresses.
+    dragDataStore: any;
```
adityarb88: What does `dragDataStore` do?

src/clr-angular/utils/drag-and-drop/all.spec.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import ClrDragAndDropEventBusSpecs from "./providers/drag-and-drop-event-bus.spec";
+describe("If Expand Directive", function() {
```
adityarb88: Copy Paste :trollface:

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
+
+export default function(): void {
+    describe("Drag And Drop Event Bus Provider", function() {
+        beforeEach(function() {
+            this.dndEventBus = new ClrDragAndDropEventBus();
+        });
+
+        it("notifyAllOnDragStart() should cast to dragStart subject", function() {
```
adityarb88: exposes an Observable to follow `dragStart` events?

would this be better? check with @youdz because I am terrible at this.

same for all messages below.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+            const testPayLoad: ClrDraggableEvent = {dragDataStore: "dragStartTest"};
+            this.dndEventBus.dragStarted.subscribe((value: any) => {
+                expect(value).toBe(testPayLoad);
+                expect(value.dragDataStore).toBe("dragStartTest");
+            });
+            this.dndEventBus.notifyAllOnDragStart(testPayLoad);
```
adityarb88: This test in particular is testing the `dragStarted` Observable but I am wondering if we need a separate test which just checks if the `notify*` methods are defined.

src/clr-angular/utils/drag-and-drop/all.spec.ts
Shijir: lol will fix this. 😛

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
Shijir: We basically test that by calling notify* methods :D

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: This will carry the payload from draggable to droppable. https://html.spec.whatwg.org/multipage/dnd.html#the-drag-data-store

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
adityarb88: Yes. I wanted to check if we need to separately test that too. Not requesting a change but just a question. Feel free to ignore this :D

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
adityarb88: would be beneficial if you add this as a comment in the code.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    get dragEnded(): Observable<ClrDraggableEvent> {
+        return this.dragEnd.asObservable();
+    }
+
+    get dropped(): Observable<ClrDraggableEvent> {
+        return this.drop.asObservable();
+    }
+
+    notifyAllOnDragStart(dragEvent: ClrDraggableEvent): void {
```
gnomeontherun: Nit: Prefer a name like `startDrag`, since this is calling the observable with this event.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
```diff
@@ -0,0 +1,50 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Injectable} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+import {ClrDraggableEvent} from "../interfaces/draggable-event";
+
+@Injectable()
+export class ClrDragAndDropEventBus {
```
gnomeontherun: The term `eventbus` seems out of place for me. Its really just a service. Shouldn't we name it `ClrDragEventService`?

src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
```diff
@@ -0,0 +1,13 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {CommonModule} from "@angular/common";
+import {NgModule} from "@angular/core";
+
+import {ClrDragAndDropEventBus} from "./providers/drag-and-drop-event-bus";
+
+@NgModule({imports: [CommonModule], providers: [ClrDragAndDropEventBus]})
+export class ClrDragAndDropModule {}
```
gnomeontherun: A general question for naming, `ClrDrag*` seems more concise. Everything revolves around the HTML `DragEvent`s. Open for debate, but feels better to me.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export interface ClrDraggableEvent {
```
gnomeontherun: Is there a reason we couldn't just use the native `DragEvent`s?

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
Shijir: I see. Yeah, that makes more sense. Will change accordingly.

src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
Shijir: I think ClrDragAndDrop* makes more sense for the classes/modules that controls both draggable and droppable. Classes that are only involved in the draggable and drag events will be named ClrDrag*.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: In our case, we are converting `MouseEvent` or `TouchEvent` into our custom drag event with more fields for custom ghost and its properties. So the API of our drag events will be almost completely different from the native DragEvent API.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
Shijir: @adityarb88 and @youdz bullied me and made me change this service's name. It was DragAndDropDispatcher initially. 😂 :trollface: 

Actually, this is the service class that coordinates events between draggables and droppables. So it carries events from draggables and transfer them to droppables. I guess that's why we've named it EventBus. 🚌 🤷‍♂️

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
Shijir: Actually, startDrag sounds good. But how about dragMove and dragEnd? I think following method names might work better?
`triggerDragStart()`
`triggerDragMove()`
`triggerDragEnd()`

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
gnomeontherun: 🎉 🚌

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
gnomeontherun: Its not really triggering it, its just a stream of those events. `dragStart()` `dragMove()` `dragEnd()` perhaps?

src/clr-angular/utils/drag-and-drop/drag-and-drop.module.ts
gnomeontherun: I can live with that.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
Shijir: I agree that they are not the source of the events so `trigger*` might not be suitable here. The subjects are already named `dragStart`, `dragMove`,  and `dragEnd`. 

Basically, the main purposes of these public methods are to be called when custom `dragStart`, `dragMove`,  and `dragEnd` events are registered on a draggable and notify all observers of the events. So I think the current names actually suggest their purposes well enough?

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
youdz: Still, Jeremy made me realize: Why isn't this called `ClrDragEvent`? It'd make more sense.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
youdz: You want to have a type parameter on the data type:
```typescript
export interface ClrDraggableEvent<T> {
  dragDataStore: T;
}
```
It'll allow proper type checking both for us and for people using the feature.

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: Yes, that made me realize the same thing. I've updated that now.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
youdz: Maybe `broadcast*` as in `broadcastStartDrag`, since that's what the service is here for?

youdz: Actually, how about this:

- The `ClrDragEvent` has a `type: EventType` property, where 
```typescript
enum EventType {
  DRAG_START,
  DRAG_MOVE,
  DRAG_END,
  DROP
}
```
- The components properly set the event type on the event when creating it
- `ClrDragAndDropEventBus` offers a single `broadcast(event: ClrDragEvent)` method, that checks the event type and fires the correct Observable.

I like this a lot better than 4 broadcast methods.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
youdz: This test passes even if the Observable never emits anything. You can try it by just commenting out
```typescript
// this.dndEventBus.notifyAllOnDragStart(testPayLoad);
```
you'll see the test still passes.

What you want is to set some variable in the subscription, and check that this variable was properly set on line 24, right after the call to `notify...()`.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
youdz: `it("exposes an Observable to follow dragStart events")` seems better to me, you're right.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.ts
Shijir: That's a great idea! 👍

src/clr-angular/utils/drag-and-drop/interfaces/draggable-event.ts
Shijir: I've removed this property for now. So instead, it has only a `type` property for now. We could decide on this property once we progress to its implementation.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
```diff
@@ -4,50 +4,60 @@
...
             this.dndEventBus = new ClrDragAndDropEventBus();
+            isEmitted = false;
+        });
+
+        it("should broadcast from correct Observable on drag start event type", function() {
+            this.dndEventBus.dragStarted.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
+                expect(event.type).toBe(ClrDragEventType.DRAG_START);
+            });
+            this.dndEventBus.broadcast({type: ClrDragEventType.DRAG_START});
```
youdz: Just a heads up: these tests will not compile when you add more required properties on the `ClrDragEvent` interface. But that's fine for this PR, we'll see when we get there.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
```diff
@@ -4,50 +4,60 @@
...
-            const testPayLoad: ClrDraggableEvent = {dragDataStore: "drop"};
-            this.dndEventBus.dropped.subscribe((value: any) => {
-                expect(value).toBe(testPayLoad);
-                expect(value.dragDataStore).toBe("drop");
+        it("should not broadcast anything on wrong event type", function() {
+            this.dndEventBus.dropped.subscribe((event: ClrDragEvent) => {
+                isEmitted = true;
             });
-            this.dndEventBus.notifyAllOnDrop(testPayLoad);
+            this.dndEventBus.broadcast({type: "if ever wrong type is passed"});
```
youdz: Wait, how does that even compile? Typescript should throw an error here. 😮

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
Shijir: Yup exactly, I have updated the properties here in my upcoming PR.

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
Shijir: Yes 😂 that's why I was asking whether I should throw an exception or not. In the test files, it's compiling. But when editing other files in IDE, it notifies me by showing that squiggly error on the type.

adityarb88: doesnt this mean that something is wrong with our `tslint` config?

youdz: It's just that `this` isn't typed. But there isn't a real point to using the user context from Jasmine here, so we should just use simple variables, that way they'll be properly typed.

Issue #1631, opened 2017-10-22
Underlying checkboxex are in front of Dropdown body


windischb:
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Use Dropdown in Datagrid ActionBar.

### Actual behavior
Checkboxes for selecting items are in front of dropdown body

### Reproduction of behavior
https://plnkr.co/edit/3nl7U7yjPXW9cHwI2gTr?p=preview

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 0.10.10

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@windischb one alternative is that you add the dropdown outside of the datagrid for now: https://plnkr.co/edit/5xmfT9M4QeY4Vt91Ie5R?p=preview let us know if that works for you. We will still look for a solution to the issue you posted.

windischb:
Yes, this is working for now!
Thanks

PR #2090, opened 2018-03-19
[UI] Datagrid Action Overflow Fix

adityarb88:
Fixes: #1631

`translateY` was causing issues with dropdown menus and modals. Replaced that by a negative margin. @hippee-lee's PR will fix this better but this can be a temp fix until that goes in.

Before:
screen shot 2018-03-19 at 4 18 51 pm

After:
screen shot 2018-03-19 at 4 18 33 pm

Before:
screen shot 2018-03-19 at 4 20 06 pm

After:
screen shot 2018-03-19 at 4 19 51 pm

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -16,6 +16,7 @@
     .datagrid-host {
         display: flex;
         flex-flow: column nowrap;
+        padding-top: 0.5rem; //TODO: Remove. Temp fix. Matt to follow new UX recommendations with pinned columns
     }
 
     .datagrid-overlay-wrapper {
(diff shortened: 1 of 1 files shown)


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -16,6 +16,7 @@
     .datagrid-host {
         display: flex;
         flex-flow: column nowrap;
+        padding-top: 0.5rem; //TODO: Remove. Temp fix. Matt to follow new UX recommendations with pinned columns
```
hippee-lee: Note: in the solution we discussed with @reddolan this is not needed. 
There is 0.5 rem above the datagrid itself and if the action-bar is present there is 1rem above that. 

I think this works until I get the pinned columns in there.

adityarb88: Yep. Thats why I added the comment. When you rebase, you can just accept your changes.

Issue #2093, opened 2018-03-20
Can't bind initial datepicker value using reactive forms


jansivans:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Initial date value should be bound to datepicker when using reactive forms.

### Actual behavior
Initial date value can not be bound to datepicker when using reactive forms.

### Reproduction of behavior
https://stackblitz.com/edit/clarity-light-theme-v11-icqw1x?file=app%2Fapp.component.ts

### Environment details

* **Angular version:** 5.2.9

* **Clarity version:** 0.11.9

* **OS and version:** Windows 10

* **Browser:** all

adityarb88:
@jansivans: Thanks for reporting this issue. I can confirm that this is a bug. The `NgControl.valueChanges` subscription does not fire for the initial value and thats why we don't write it back to the input or update the current selected day. I will spend some time today to investigate this and post more updates here.

adityarb88:
@youdz, @gnomeontherun: So I see 2 different behaviors when the initial value is not null:
1. Reactive Forms: `NgControl.valueChanges` does not fire but `NgControl.value` is not null and contains the initial value. I can use this value to update the input and the selected day in the Calendar.
2. Template Drive Forms: `NgControl.valueChanges` fires but `NgControl.value` is null.

Obviously, we have the second case covered but the first one is missing. I will add it today and test if that solves the issue.

PR #2097, opened 2018-03-20
[NG] Bug Fix for Datepicker in Reactive Forms

adityarb88:
Fixes: #2093

The initial value set by the user in Reactive Forms wasn't being processed by the Date Picker. Fixes that issue. For more details see #2093

Tested on Chrome

### DIFF
--- a/src/clr-angular/forms/datepicker/date-input.ts
+++ b/src/clr-angular/forms/datepicker/date-input.ts
@@ -61,7 +61,23 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
             this.populateContainerServices();
         }
         this.initializeSubscriptions();
+        this.processInitialInputs();
+    }
+
+    /**
+     * Process the inputs initialized by the user which were missed
+     * because of late subscriptions or lifecycle method calls.
+     */
+    private processInitialInputs(): void {
         this.processUserDateObject(this.dateValueOnInitialLoad);
+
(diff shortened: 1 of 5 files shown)


src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -61,7 +61,28 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
+        // Handle Inital Value from Reactive Forms
+        // TODO: We are repeating this logic at multiple places. This makes me think
+        // if this class should have implemented the ControlValueAccessor interface.
+        // Will explore that later and see if its a cleaner solution.
+        if (this._ngControl && this._ngControl.value) {
+            const date: Date = this._dateIOService.isValidInput(this._ngControl.value);
+            if (date) {
+                this._dateNavigationService.selectedDay =
+                    new DayModel(date.getFullYear(), date.getMonth(), date.getDate());
+                this.writeDateStrToInputField(this._ngControl.value);
```
youdz: That's a lot of code duplication with `processUserDateObject()`. It's actually the same method, but `processUserDateObject()` starts by converting the Date object to a string, then does exactly the same as this.

Please rewrite these two methods so that they share the common code.

adityarb88: @youdz: done.

PR #2102, opened 2018-03-21
[NG] Fix toggleAll selection tracking

### LINKED ISSUE #2083: Server-driven datagrid selection not preserved when all rows are selected via header checkbox
```
[ x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Header checkbox select should select currently displayed rows and track the selection of those rows.

### Actual behavior
Header checkbox select selects all rows currently displayed, but does not track them.  

### Reproduction of behavior
**Stackblitz:** https://stackblitz.com/edit/clarity-serverdatagrid-header-trackby

Check header checkbox, all rows should be selected.  Navigate to next page in grid, return to previous page. Items that were previously selected will no longer appear selected.

### Environment details

* **Angular version:** 5.2.0

* **Clarity version:** 0.11.4

* **OS and version:** Windows Server 2016

* **Browser:** all  

### Additional Comment
`toggleAll()` in selection.ts does a push of all non-selected items into current[], but does not update the selection[] in a similar manner to `setSelected`.  I believe updating selection[] would resolve the issue.

mviger:
Moved select and deselect into private methods to keep current and
selection in sync.
toggleAll calls select and deselect when adding or removing items.

closes #2083

Signed-off-by: Marc Viger <mviger@gmail.com>

### DIFF
--- a/src/clr-angular/data/datagrid/providers/selection.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.ts
@@ -189,6 +189,29 @@ export class Selection {
     }
 
     /**
+     * Selects an item
+     */
+    private selectItem(item: any): void {
+        this.current.push(item);
+        if (this._items.trackBy) {
+            // Push selected ref onto array
+            const lookup = this._items.all.findIndex(maybe => maybe === item);
+            this.selected.push(this._items.trackBy(lookup, item));
+        }
+    }
+
+    /**
+     * Deselects an item
+     */
+    private deselectItem(indexOfItem: number): void {
+        this.current.splice(indexOfItem, 1);
+        if (this._items.trackBy && indexOfItem < this.selected.length) {
+            // Keep selected refs array in sync
+            this.selected.splice(indexOfItem, 1);
+        }
+    }
+
+    /**
      * Selects or deselects an item
      */
     public setSelected(item: any, selected: boolean) {
@@ -200,21 +223,11 @@ export class Selection {
                 break;
             case SelectionType.Multi:
                 const index = this.current.indexOf(item);
-                const trackBy = this._items.trackBy;
                 if (index >= 0 && !selected) {
-                    this.current.splice(index, 1);
-                    if (trackBy) {
-                        // Keep selected refs array in sync
-                        this.selected.splice(index, 1);
-                    }
+                    this.deselectItem(index);
                     this.emitChange();
                 } else if (index < 0 && selected) {
-                    this.current.push(item);
-                    if (trackBy) {
-                        // Push selected ref onto array
-                        const lookup = this._items.all.findIndex(maybe => maybe === item);
-                        this.selected.push(this._items.trackBy(lookup, item));
-                    }
+                    this.selectItem(item);
                     this.emitChange();
                 }
                 break;
@@ -251,16 +264,19 @@ export class Selection {
          * If at least one item isn't selected, we select every currently displayed item.
          */
         if (this.isAllSelected()) {
-            this.current = this.current.filter(item => {
-                return this._items.displayed.indexOf(item) < 0;
+            this._items.displayed.forEach((item, displayIndex) => {
+                const currentIndex = this.current.indexOf(item);
+                if (currentIndex > -1) {
+                    this.deselectItem(currentIndex);
+                }
             });
         } else {
             this._items.displayed.forEach(item => {
                 if (this.current.indexOf(item) < 0) {
-                    this.current.push(item);
+                    this.selectItem(item);
                 }
             });
-            this.emitChange();
         }
+        this.emitChange();
     }
 }
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -352,6 +352,18 @@ export default function(): void {
                 expect(selectionInstance.isSelected(itemsC[0])).toEqual(stateC);
             }
 
+            function testToggleAllSelection(stateA, stateB, stateC) {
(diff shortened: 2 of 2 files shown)


mviger:
This is my first attempt at a PR in GitHub.  I apologize profusely if I've screwed something up, or messed up some etiquette.  I've tried to follow as closely as possible all the contribution guidelines.

adityarb88:
Thanks for the contribution @mviger! We will take a look and get back to you soon.

mviger:
I initially ran clang:check in my environment, saw there were errors, ran clang:format to fix everything and then re-ran clang:check.  Everything passed and I updated the PR.  I see the build CI has failed again on clang:check, but in my environment it is passing.  I'm not sure how to resolve the remaining issues currently.

Sorry for the trouble.

Shijir:
It might be a Travis CI clutch issue. I have restarted the failed job. Let's see if it passes or not.

adityarb88:
@mviger: Could you also try running `npm run tslint:fix`?

mviger:
@adityarb88 `tslint:fix` resulted in no changes on my machine.  I've also tried to run `clang:format` again, but no changes with that either.  I'm running on a Windows box currently, don't think that should matter though.   Potentially line break issue?

src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -77,8 +77,8 @@ export const ClrShapeWifi = clrIconSVG(
             <path class="clr-i-solid clr-i-solid-path-3" d="M28.82,15.44a20.59,20.59,0,0,0-21.7,0,1.51,1.51,0,0,0-.46,2.1l.11.17a1.49,1.49,0,0,0,2,.46,17.4,17.4,0,0,1,18.36,0,1.49,1.49,0,0,0,2-.46l.11-.17A1.51,1.51,0,0,0,28.82,15.44Z"/>
             <path class="clr-i-solid clr-i-solid-path-4" d="M24.88,21.49a13.41,13.41,0,0,0-13.82,0,1.5,1.5,0,0,0-.46,2.09l.1.16a1.52,1.52,0,0,0,2.06.44,10.27,10.27,0,0,1,10.42,0,1.52,1.52,0,0,0,2.06-.45l.1-.16A1.49,1.49,0,0,0,24.88,21.49Z"/>
 </svg>`);
-export const ClrShapeRackServer =
-    clrIconSVG(`<rect class="clr-i-outline--alerted clr-i-outline-path-1--alerted" x="10" y="17" width="14" height="2"/>
+export const ClrShapeRackServer = clrIconSVG(
```
hippee-lee: Is this part of your PR? 
I pulled the changes for your PR locally and `npm run clang:check` failes. `npm run clang:fix` reverts this back to what it used to be.

mviger: it wasn't originally, but was added when I did `npm run clang:format`.  I was surprised by it, but figured it was probably correct because it was a result from `npm run clang:format`.

hippee-lee: Hmmm, it sounds like your computer is doing the opposite of mine? 
Anyway, can you revert that and try again?

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -188,6 +188,29 @@ export class Selection {
         return false;
     }
 
+    /**
+     * Selects an item
+     */
+    private selectItem(item: any): void {
+        this.current.push(item);
+        if (this._items.trackBy) {
```
Shijir: @adityarb88 didn't you add an check on whether `this._items.all` is empty or not? Do we need that here?

src/clr-icons/shapes/technology-shapes.ts
hippee-lee: Looks like unit tests are running now. Sorry, I'm not sure exactly whats happening on your system. Thanks for working with us to get the tests to run

mviger: I was running the clang stuff through **git bash** on windows, don't know if that would result in any differences.  I can try pulling it onto a mac later and see what results from the same scripts there.

Thanks for all the help.

src/clr-angular/data/datagrid/providers/selection.ts
youdz: That one can't be called on initialization, so we're ok. The problem that Aditya fixed was when the selection was *initialized* before the items themselves were. Here, this method can only be called with an item that's being displayed, so the items have been initialized for sure.

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -251,16 +264,19 @@ export class Selection {
          * If at least one item isn't selected, we select every currently displayed item.
          */
         if (this.isAllSelected()) {
-            this.current = this.current.filter(item => {
-                return this._items.displayed.indexOf(item) < 0;
+            this._items.displayed.forEach((item, displayIndex) => {
+                const currentIndex = this.current.indexOf(this._items.trackBy(displayIndex, item));
```
youdz: You don't want the `trackBy` here to test against `current`. This should simply be:
```typescript
const currentIndex = this.current.indexOf(item);
```

mviger: You are correct.  I've updated the commit accordingly. 

Sorry, that was left over from an attempt at getting the index from `this.selected` instead of `this.current`.  That failed one of the other units tests. I realized that when `set current` is called, `selected` does not get updated at that time.

That might expose a separate issue, but outside of the current scope I believe.

PR #2106, opened 2018-03-23
[WIP | NG] ClrDragEventListener implementation

Shijir:

### DIFF
--- a/src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
+++ b/src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
@@ -3,7 +3,6 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-
 export enum ClrDragEventType {
     DRAG_START,
     DRAG_MOVE,
@@ -11,7 +10,11 @@ export enum ClrDragEventType {
     DROP
 }
 
-export interface ClrDragEvent {
+export interface ClrDragEvent<T> {
     // TODO: this interface will be expanded and customized as the implementation progresses.
     type: ClrDragEventType;
+    group?: string|string[];
+    draggableElement: Node;
+    dragPosition: {pageX: number, pageY: number};
+    dragDataTransfer?: T;
 }
--- a/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
@@ -3,13 +3,25 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-
 import {ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
 import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
+
+type DragTransfer = {
+    data: any
+};
 
 export default function(): void {
     describe("Drag And Drop Event Bus Provider", function() {
         const dndEventBus = new ClrDragAndDropEventBus();
+        const dragEventMockObj =
+            (dragEventType: ClrDragEventType, dragDataTransfer?: DragTransfer): ClrDragEvent<DragTransfer> => {
+                return {
+                    type: dragEventType,
+                    draggableElement: null,
+                    dragPosition: {pageX: 0, pageY: 0},
+                    dragDataTransfer: dragDataTransfer
+                };
+            };
         let isEmitted: boolean;
 
         beforeEach(function() {
@@ -17,38 +29,38 @@ export default function(): void {
         });
 
         it("should broadcast from correct Observable on drag start event type", function() {
-            dndEventBus.dragStarted.subscribe((event: ClrDragEvent) => {
+            dndEventBus.dragStarted.subscribe((event: ClrDragEvent<DragTransfer>) => {
                 isEmitted = true;
                 expect(event.type).toBe(ClrDragEventType.DRAG_START);
             });
-            dndEventBus.broadcast({type: ClrDragEventType.DRAG_START});
+            dndEventBus.broadcast(dragEventMockObj(ClrDragEventType.DRAG_START));
             expect(isEmitted).toBeTruthy();
         });
 
         it("should broadcast from correct Observable on drag move event type", function() {
-            dndEventBus.dragMoved.subscribe((event: ClrDragEvent) => {
+            dndEventBus.dragMoved.subscribe((event: ClrDragEvent<DragTransfer>) => {
                 isEmitted = true;
                 expect(event.type).toBe(ClrDragEventType.DRAG_MOVE);
             });
-            dndEventBus.broadcast({type: ClrDragEventType.DRAG_MOVE});
+            dndEventBus.broadcast(dragEventMockObj(ClrDragEventType.DRAG_MOVE));
             expect(isEmitted).toBeTruthy();
         });
 
         it("should broadcast from correct Observable on drag end event type", function() {
-            dndEventBus.dragEnded.subscribe((event: ClrDragEvent) => {
+            dndEventBus.dragEnded.subscribe((event: ClrDragEvent<DragTransfer>) => {
                 isEmitted = true;
                 expect(event.type).toBe(ClrDragEventType.DRAG_END);
             });
-            dndEventBus.broadcast({type: ClrDragEventType.DRAG_END});
+            dndEventBus.broadcast(dragEventMockObj(ClrDragEventType.DRAG_END));
             expect(isEmitted).toBeTruthy();
         });
 
         it("should broadcast from correct Observable on drop event type", function() {
-            dndEventBus.dropped.subscribe((event: ClrDragEvent) => {
+            dndEventBus.dropped.subscribe((event: ClrDragEvent<DragTransfer>) => {
                 isEmitted = true;
                 expect(event.type).toBe(ClrDragEventType.DROP);
             });
-            dndEventBus.broadcast({type: ClrDragEventType.DROP});
+            dndEventBus.broadcast(dragEventMockObj(ClrDragEventType.DROP));
             expect(isEmitted).toBeTruthy();
         });
     });
--- a/src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
@@ -1,0 +1,322 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component, ElementRef, NgZone, OnDestroy, OnInit, Renderer2, ViewChild} from "@angular/core";
+import {ComponentFixture, TestBed} from "@angular/core/testing";
+import {Subscription} from "rxjs/Subscription";
+
+import {ClrDragEvent} from "../interfaces/drag-event";
+import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
+import {ClrDragEventListener} from "./drag-event-listener";
+
+type DragTransfer = {
+    data: any
+};
+
+const emulateEventOn =
+    (eventName: string, pageX: number = 0, pageY: number = 0, element: Document|Node = document.body): void => {
+        const emulatedEvent: any = new CustomEvent(eventName, {bubbles: true});
+        if (eventName.startsWith("touch")) {
+            emulatedEvent.changedTouches = [{pageX: pageX, pageY: pageY}];
+        } else {
+            emulatedEvent.pageX = pageX;
+            emulatedEvent.pageY = pageY;
+        }
+
+        element.dispatchEvent(emulatedEvent);
+    };
+
+const expectEventPropValues = <T>(event: ClrDragEvent<T>) => {
+    return {
+        toBe: (element: Node, pageX: number, pageY: number, dragTransfer?: T, group?: string|string[]) => {
+            expect(event.draggableElement).toBe(element);
+            expect(event.dragPosition.pageX).toBe(pageX);
+            expect(event.dragPosition.pageY).toBe(pageY);
+
+            if (group) {
+                expect(event.group).toEqual(group);
+            } else {
+                expect(event.group).toBeUndefined();
+            }
+
+            if (dragTransfer) {
+                expect(event.dragDataTransfer).toEqual(dragTransfer);
+            } else {
+                expect(event.dragDataTransfer).toBeUndefined();
+            }
+        }
+    };
+};
+
+export default function(): void {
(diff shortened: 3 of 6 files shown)


src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private customDragEvent(element: Node, startOnEvent: string, moveOnEvent: string, endOnEvent: string): () => void {
+        let moveListener: () => void;
+        let endListener: () => void;
+
+        return this.renderer.listen(element, startOnEvent, () => {
+            moveListener = this.renderer.listen("document", moveOnEvent, (moveEvent: MouseEvent|TouchEvent) => {
+                moveEvent.preventDefault();
+                moveEvent.stopPropagation();
```
youdz: We mentioned this before, we can never stop propagation. I don't remember why you added it here, but it's simply something we cannot have at all in a library.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,122 @@
...
+
+                if (!this.hasDragStarted) {
+                    this.hasDragStarted = true;
+
+                    // attach drag end listener only if drag start was fired.
+                    endListener = this.renderer.listen("document", endOnEvent, (endEvent: MouseEvent|TouchEvent) => {
+                        moveListener();  // Unregistering from mouseMove or touchMove
```
youdz: I believe you have a bug here: if you click on a draggable element, here is what happens:
1) The `mousedown` listener is called, and attaches and `mousemove` listener.
2) The user releases the mouse without moving, nothing happens because there isn't a mouseup listener yet.
3) The user starts moving their mouse, the `mousemove` listener is called and starts doing all its things.
4) All the listeners keep firing when the user moves the mouse, even though they are not dragging at all since the mouse has been released in 2).

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,122 @@
...
+                        endListener();   // Unregistering from mouseUp or touchEnd
+
+                        this.hasDragStarted = false;
+                        const dragEndEvent: ClrDragEvent = this.generateDragEvent(endEvent, ClrDragEventType.DRAG_END);
+                        this.broadcast(dragEndEvent);
```
youdz: The utility method should generate the drag event _and_ broadcast it. You never generate an event without broadcasting it, no point in having two separate methods.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,122 @@
...
+            nativeEvent = (<TouchEvent>event).changedTouches[0];
+        } else {
+            nativeEvent = event;
+        }
+
+        return {
+            type: eventType,
+            draggableElement: this.draggabelEl,
+            dragPosition: {pageX: nativeEvent.pageX, pageY: nativeEvent.pageY},
+            dragDataTransfer: this.dragDataTransfer || null
```
youdz: Your `dragDataTransfer` here is a property on this service, that seems pretty static. I think it's okay right now, but be careful to make sure it can be updated at any time by the consumer when you write the `ClrDraggable` directive.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,122 @@
...
+import {Injectable, Renderer2} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+
+import {ClrDragDataTransfer, ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
+import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
+
+@Injectable()
+export class ClrDragEventListener {
+    private draggabelEl: Node;
```
youdz: Typo, `draggableEl`.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -3,15 +3,23 @@
...
     DRAG_START,
     DRAG_MOVE,
     DRAG_END,
     DROP
 }
 
+export interface ClrDragDataTransfer {
+    // TODO: groupLinks seems to be a good name?
+    groupLinks?: string|string[];
+    data: any;
```
youdz: Why declare a sub-interface for this? It seems a bit overkill, having the group and the payload at the root of the event seems easier, especially when you're going to have to update them independently sometimes.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -3,15 +3,23 @@
...
+}
+
 export interface ClrDragEvent {
     // TODO: this interface will be expanded and customized as the implementation progresses.
     type: ClrDragEventType;
+    draggableElement: Node;
+    dragPosition: {pageX: number, pageY: number};
+    dragDataTransfer?: ClrDragDataTransfer;
```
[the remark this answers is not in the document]
adityarb88: I think he went with what the native Drag and Drop API uses: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

Even I think payload is better though.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
adityarb88: I think he mentioned something about nested draggables. Not sure though. But yes, we can't have `stopPropagation`.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
Shijir: Addressed the request.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
Shijir: Addressed the request.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
Shijir: Addressed the request. Added in a test for this case.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -3,15 +3,18 @@
...
     DRAG_END,
     DROP
 }
 
-export interface ClrDragEvent {
+export interface ClrDragEvent<T> {
     // TODO: this interface will be expanded and customized as the implementation progresses.
     type: ClrDragEventType;
+    draggableElement: Node;
+    dragPosition: {pageX: number, pageY: number};
```
adityarb88: Should `DragPosition` be a separate type or interface? Feels like it will be reused in DnD a lot, no?

src/clr-angular/utils/drag-and-drop/providers/drag-and-drop-event-bus.spec.ts
```diff
@@ -3,52 +3,59 @@
...
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-
 import {ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
 import {ClrDragAndDropEventBus} from "./drag-and-drop-event-bus";
 
 export default function(): void {
     describe("Drag And Drop Event Bus Provider", function() {
         const dndEventBus = new ClrDragAndDropEventBus();
+        const dragEventMockObj = (dragEventType: ClrDragEventType, dragDataTransfer?: any): ClrDragEvent<null> => {
```
adityarb88: Shouldn't the type for `dragDataTransfer` match that of `ClrDragEvent`?

For eg: this is valid:
`const dragEventMockObj = (dragEventType: ClrDragEventType, dragDataTransfer?: string): ClrDragEvent<string>`

but you have `any` and `null` which do not match.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    private customDragEvent(element: Node, startOnEvent: string, moveOnEvent: string, endOnEvent: string): () => void {
+        let moveListener: () => void;
+        let endListener: () => void;
+
+        return this.renderer.listen(element, startOnEvent, () => {
+            moveListener = this.ngZone.runOutsideAngular(() => {
+                return this.renderer.listen("document", moveOnEvent, (moveEvent: MouseEvent|TouchEvent) => {
+                    moveEvent.preventDefault();
+                    moveEvent.stopPropagation();
```
adityarb88: Add a note as to why we are using this and why we shouldn't be using it elsewhere.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+            emulatedEvent.pageX = pageX;
+            emulatedEvent.pageY = pageY;
+        }
+
+        element.dispatchEvent(emulatedEvent);
+    };
+
+const expectEventPropValues = <T>(event: ClrDragEvent<T>) => {
+    return {
+        toBe: (element: Node, pageX: number, pageY: number, dragTransfer?: T, dropGroup?: string|string[]) => {
```
adityarb88: `dropScope` instead `dropGroup`

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    dragStartFired = false;
+    dragMoveFired = false;
+    dragEndFired = false;
+
+    dragEvent: ClrDragEvent<DragTransfer>;
+
+    nbDragMoveFired = 0;
+
+    constructor(renderer: Renderer2, ngZone: NgZone, private dragEventListener: ClrDragEventListener<DragTransfer>,
+                private eventBust: ClrDragAndDropEventBus<DragTransfer>) {}
```
adityarb88: `eventBus` instead of `eventBust`

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+            draggableButton = testComponent.draggableButtonRef.nativeElement;
+        });
+
+        afterEach(() => {
+            fixture.destroy();
+        });
+
+        function testCases(startEvent: string, moveEvent: string, endEvent: string) {
+            it(startEvent + " alone shouldn't fire any drag event", function() {
```
adityarb88: Fix the sentence. Discussed offline.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+                emulateEventOn(moveEvent, 22, 33);
+
+                expect(testComponent.dragStartFired).toBeTruthy();
+                expectEventPropValues(testComponent.dragEvent).toBe(draggableButton, 22, 33, null);
+
+                expect(testComponent.dragMoveFired).toBeFalsy();
+                expect(testComponent.dragEndFired).toBeFalsy();
+            });
+
+            it("multiple " + moveEvent + " after " + startEvent + " should fire dragstart event", function() {
```
adityarb88: looks like this test is redundant. the above test already tests this, no?

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+                   // mousedown+mouseup means it just ended prematurely
+                   emulateEventOn(startEvent, 0, 0, draggableButton);
+                   emulateEventOn(endEvent, 0, 0);
+
+                   // emulating twice
+                   emulateEventOn(startEvent, 0, 0, draggableButton);
+                   emulateEventOn(endEvent, 0, 0);
+
+                   // mousedown+mousemove shouldn't fire dragstart
```
adityarb88: `mousedown + mousemove` should fire `dragStart`. no?

adityarb88: You are testing it correctly. I think the comment is incorrect.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+                // drag end
+                emulateEventOn(endEvent, 22, 33);
+
+                expect(testComponent.dragStartFired).toBeTruthy();
+                expect(testComponent.dragMoveFired).toBeTruthy();
+                expect(testComponent.dragEndFired).toBeTruthy();
+            });
+
+            it("can transfer data on each drag events", function() {
+                const dataOnDragStart = {data: {test: "dataOnDragStart"}};
```
adityarb88: Do you think we should break this into 3 separate tests?

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
Shijir: this one is removed.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.spec.ts
Shijir: fixed the comment message.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -3,15 +3,18 @@
...
+    dragDataTransfer?: T;
+    dragScope?: string|string[];
```
youdz: As discussed, we're going to go with a `clrGroup` input, so you should update "scope" to "group" here and everywhere else.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    private customDragEvent(element: Node, startOnEvent: string, moveOnEvent: string, endOnEvent: string): () => void {
+        let moveListener: () => void;
+        let endListener: () => void;
+
+        return this.renderer.listen(element, startOnEvent, () => {
+            moveListener = this.ngZone.runOutsideAngular(() => {
+                return this.renderer.listen("document", moveOnEvent, (moveEvent: MouseEvent|TouchEvent) => {
```
youdz: As mentioned before, I believe `renderer.listen` completely voids the `ngZone.runOutsideAngular`. Because it's "free" and doesn't impact anything, keep it like that for now and we can test for performance later on in the topic branch.

Issue #2055, opened 2018-03-08
Support other date delimiters in new date control


rbirkgit:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x ] enhancement
```

### Expected behavior
Would be nice to support other date delimiters than "/". Sweden for example use "-" and Germany ".'

### Actual behavior
Right now only "/" is supported.

### Environment details

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web

adityarb88:
@rbirkgit: Thanks for filing the issue!

I see that the date format which is retrieved from Angular for locale `sv` is `y-MM-dd` (https://github.com/angular/angular/blob/master/packages/common/locales/sv.ts#L40)
and for `de` is `dd.MM.yy`
(https://github.com/angular/angular/blob/master/packages/common/locales/de.ts).

I want to verify if we can use the delimiter in this format to display dates in the date picker input. I think we can but I just need to verify it once.

I would also have to consult with out UX team and our internationalization team to check what they think about this. I will add more updates to this tickets when I have more info.

KANekT:
in russian date format dd.MM.y

[ru.ts](https://github.com/angular/angular/blob/master/packages/common/locales/ru.ts)

rameshyala:
Hi, @adityarb88 i am set Locale to my app like { provide: LOCALE_ID, useValue: 'en-IN' } in app module, but it still showing the MMDDYYYY format only, actually it need to change into DDMMYYY format right?.How i get this format.

and i have a one more doubt, now if i change the actual date in datePicker i will get any onChnage event  now (change detection)?

Thank you.

PR #2110, opened 2018-03-26
[NG] CLDR Delimiter Support

adityarb88:
Fixes: #2055

https://github.com/angular/angular/blob/master/packages/common/locales/de.ts
image
image

https://github.com/angular/angular/blob/master/packages/common/locales/kkj.ts
image
image

https://github.com/angular/angular/blob/master/packages/common/locales/hr.ts
image
image

https://github.com/angular/angular/blob/master/packages/common/locales/ar.ts
image
image

Tested on Chrome

### DIFF
--- a/src/clr-angular/forms/datepicker/_datepicker.clarity.scss
+++ b/src/clr-angular/forms/datepicker/_datepicker.clarity.scss
@@ -179,6 +179,7 @@
     .yearpicker-trigger {
         //Dimensions
         min-width: 2rem;
+        max-width: 2.5rem;
 
         //Other Props
         color: $clr-calendar-btn-color;
--- a/src/clr-angular/forms/datepicker/providers/date-io.service.ts
+++ b/src/clr-angular/forms/datepicker/providers/date-io.service.ts
@@ -9,13 +9,14 @@ import {Injectable} from "@angular/core";
 import {
     BIG_ENDIAN,
     DEFAULT_LOCALE_FORMAT,
+    DELIMITER_REGEX,
     InputDateDisplayFormat,
     LITTLE_ENDIAN,
     LITTLE_ENDIAN_REGEX,
     MIDDLE_ENDIAN,
     MIDDLE_ENDIAN_REGEX,
-    NON_NEGATIVE_NUMBER,
-    SEPARATORS
+    RTL_REGEX,
+    USER_INPUT_REGEX
 } from "../utils/constants";
 import {getNumberOfDaysInTheMonth, parseToFourDigitYear} from "../utils/date-utils";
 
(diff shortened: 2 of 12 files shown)


src/clr-angular/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -179,6 +179,7 @@
     .yearpicker-trigger {
         //Dimensions
         min-width: 2rem;
+        max-width: 2.5rem;
```
hippee-lee: Should the width always be 2rem?

src/clr-angular/forms/datepicker/providers/date-io.service.ts
```diff
@@ -41,6 +43,19 @@ export class DateIOService {
...
+
+    private extractDelimiters(): void {
+        if (this.cldrLocaleDateFormat) {
+            // Sanitize Date Format. Remove RTL characters.
+            // FIXME: When we support RTL, remove this and handle it correctly.
+            const localeFormat: string = this.cldrLocaleDateFormat.replace(RTL_REGEX, "");
+            const delimiters: string[] = localeFormat.split(DELIMITER_REGEX);
+            if (delimiters.length === 4) {
+                this.delimiters = [delimiters[1], delimiters[2]];
+            }
```
youdz: Just in case, do you mind logging or raising an error here? If we read more than 4 splits, then we have an issue somewhere and we'll want to fix it.

Shijir: I don't get what's going on here :P . Why is it including delimiter strings in an array? What value does it have over using just plain string delimiter? 

Why for example this:
`this.delimiters = [delimiters[1], delimiters[2]];` is not `this.delimeters = delimiters`?

src/clr-angular/forms/datepicker/_datepicker.clarity.scss
adityarb88: some locales require more space to display the month value. but we dont want it to extend more than 2.5rem.

src/clr-angular/forms/datepicker/providers/date-io.service.ts
adityarb88: You are right. I should have added a comment here. Let me update the PR with a comment.

Explanation:
https://github.com/angular/angular/blob/master/packages/common/locales/kkj.ts#L46

image

The date format received from CLDR is always in D..M..Y or M..D..Y or Y..M..D format. Our regex detects that and splits around the D, M and Y. The result is always an array with 4 strings.

adityarb88: Also, @youdz pointed out on our internal chat that there is missing null value check here. Will fix that too.

PR #2113, opened 2018-03-27
[NG] Clean up datagrid footer's flexbox

youdz:
The only child of the footer that now expands is the
`.datagrid-foot-description` which contains any custom text or buttons
the consumer decides to add to the footer. Since this element is always
here, it will naturally push its siblings to the left and to the right
of the footer, preserving the current styles.


adityarb88:
Wait I forgot to ask if this was tested on all browsers? IE11 usually behaves different than other browsers when it comes to flexbox. Actually when it comes to everything CSS.

youdz:
It does work on IE, this isn't even a gray area, I'm just swapping which element is flexible, so we have the exact same CSS as before except the second child stretches instead of the first. I did check. 👍 

But I just realized I'm assuming our users use the new-ish API for pagination, which isn't the case. Even our CSS tests don't use it yet, so it messes up the position of the pagination text. I'm pushing a quick fix for that right now.

Issue #2119, opened 2018-03-29
clrCurrentAlertIndex of Alerts component does not accept zero


gerinjacob:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
`clrCurrentAlertIndex` input attribute of `clr-alerts` should accept zero

### Actual behavior
In the punker provided, https://current-alert-index-zero.stackblitz.io
There are 2 buttons, `Go To First Alert` and `Go To Second Alert`.
I am able to go to second alert, but not to first alert.

**Root Cause:**
Inside ClrAlerts component there is an `if` check for index, which evaluates zero to false. Hence it is not getting set.
```javascript
public set _inputCurrentIndex(index: number) {
  if (index) {
    this.multiAlertService.current = index;
  }
}
```
**Possible Solution:**
Check for zero as well,
```javascript
public set _inputCurrentIndex(index: number) {
  if (index || index === 0) {
    this.multiAlertService.current = index;
  }
}
```

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

https://current-alert-index-zero.stackblitz.io
https://stackblitz.com/edit/current-alert-index-zero

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** v11

* **OS and version:** Any

* **Browser:** all

svvac:
duplicate of #1907

PR #2120, opened 2018-03-29
[NG] Bug fix for ClrAlerts

gerinjacob:
Fixes: #2119
Signed-off-by: Gerin Jacob <gerinjacob@gmail.com>

### DIFF
--- a/src/clr-angular/emphasis/alert/alerts.spec.ts
+++ b/src/clr-angular/emphasis/alert/alerts.spec.ts
@@ -68,6 +68,39 @@ describe("ClrAlerts component", function() {
             expect(testComponent.currentAlertIndex).toEqual(1);
         });
 
+        it("allows changing alert index from 1 to 0", function() {
+            this.create(TestComponent);
+            const testComponent = fixture.componentInstance;
+            fixture.detectChanges();
+            testComponent.currentAlertIndex = 1;
+            fixture.detectChanges();
+            expect(testComponent.alertsInstance.currentAlertIndex).toEqual(1);
+            testComponent.currentAlertIndex = 0;
+            fixture.detectChanges();
+            expect(testComponent.alertsInstance.currentAlertIndex).toEqual(0);
+        });
+
+        it("does not allow invalid indexes (undefined, null, -1, 1.5, string)", function() {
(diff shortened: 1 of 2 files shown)


youdz:
I agree with Scott, our policy is that if we fix a simple bug, we add a simple unit test for it to make sure it doesn't come back later. Thanks for the contribution. :)

gerinjacob:
@mathisscott, @youdz I have added unit test for this scenario

src/clr-angular/emphasis/alert/alerts.spec.ts
```diff
@@ -68,6 +68,18 @@ describe("ClrAlerts component", function() {
...
+            const testComponent = fixture.componentInstance;
+            fixture.detectChanges();
+            testComponent.currentAlertIndex = 1;
+            fixture.detectChanges();
+            expect(testComponent.alertsInstance.currentAlertIndex).toEqual(1);
+            testComponent.currentAlertIndex = 0;
+            fixture.detectChanges();
+            expect(testComponent.alertsInstance.currentAlertIndex).toEqual(0);
+        });
+
```
mathisscott: This is great. But could you also check against `null` and `-1`? Just to make sure that we are properly guarding (in the tests) against bad values coming through...

src/clr-angular/emphasis/alert/alerts.ts
```diff
@@ -29,7 +29,7 @@ export class ClrAlerts implements AfterContentInit {
      */
     @Input("clrCurrentAlertIndex")
     public set _inputCurrentIndex(index: number) {
-        if (index) {
+        if (index || index === 0) {
```
gnomeontherun: Can we also just simplify this to `index >= 0` instead of having an `||`?

gerinjacob: I didn't do that because `null >= 0` is `true`. With latest change I am checking Number.isInteger(index) && index >= 0
This should cover all the invalid values.

gerinjacob:
@mathisscott , I have added check for undefined, null, -1, 1.5, string

gnomeontherun:
Thanks @gerinjacob, this will go into the release today!

PR #2170, opened 2018-04-06
[NG] Issue with popover escaping angular zone in embedded apps

### LINKED ISSUE #2108: Issue with Popovers while hosted in sharepoint page
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Clicking outside should close the popover

### Actual behavior
The dropdown menu dosen't close and places itself in the top left corner

### Reproduction of behavior
Not really possible

### Environment details

* **Angular version:** 5.0.0

* **Clarity version:** 0.11.10

* **OS and version:** Windows 10

* **Browser:** All

headswe:
Fixes issue caused by hosting the app in a SharePoint

Fixes #2108

Signed-off-by: Sebastian Nygren <nygseb@gmail.com>

### DIFF
--- a/src/clr-angular/forms/datepicker/datepicker-view-manager.ts
+++ b/src/clr-angular/forms/datepicker/datepicker-view-manager.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import {Component, ElementRef, Injector, SkipSelf} from "@angular/core";
+import {Component, ElementRef, Injector, NgZone, SkipSelf} from "@angular/core";
 
 import {AbstractPopover} from "../../popover/common/abstract-popover";
 import {Point} from "../../popover/common/popover";
@@ -19,8 +19,9 @@ import {ViewManagerService} from "./providers/view-manager.service";
     host: {"[class.datepicker]": "true"}
 })
 export class ClrDatepickerViewManager extends AbstractPopover {
-    constructor(@SkipSelf() parent: ElementRef, _injector: Injector, private _viewManagerService: ViewManagerService) {
-        super(_injector, parent);
+    constructor(@SkipSelf() parent: ElementRef, _injector: Injector, private _viewManagerService: ViewManagerService,
+                ngZone: NgZone) {
+        super(ngZone, _injector, parent);
         this.configurePopover();
     }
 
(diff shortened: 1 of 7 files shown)


youdz:
Your PR clearly won't _break_ anything, but it adds code that will make things harder to maintain for no reason I can think of.

Basically, you're checking whether callbacks registered with `Renderer2.listen()` are executed inside of the `NgZone`, but by definition of the Renderer this is **always** the case. So this really looks like a bug in your Angular + SharePoint integration to me, and trying to fix it in Clarity will just hide part of the problem for now and you'll keep hitting it more in the future.

After a quick search on Google, I found this: http://www.sebastienlevert.com/2017/08/01/sharepoint-framework-and-angular-walkthrough-of-the-spfx-angular-boilerplate/
It states that:
>The last piece that needs to happen is to externalize the Zone.js component of the bundle. This is EXTREMELY IMPORTANT. Without this, the code will work without any problem but will break as soon as another webpart built with Angular but from another project will be dropped on the same page as yours. The key to this boilerplate is to keep a single Zone.js instance to be loaded so all the webparts share the same.
This really looks like your current problem to me, did you go through this process already?

headswe:
Thanks for looking it over!

I very much expected this PR to met with no since it's rather a special problem, The article to you refer to is sadly not applicable to my case since we don't use Sharepoint Framework nor do we host multiple angular apps in one page.

We have hosted several apps with the Angular Material in the same way and have not noticed the same behaviour by their components.

gnomeontherun:
Took a few minutes to review what material is doing here to get a sense what might be different. It looks like material tends to listen for events using `document.body.addEventListener` instead of `renderer.listen`. Its hard to debug something that I can't work with, but I agree there is something odd about having to do this workaround in Clarity without being certain there isn't an issue with the way its being deployed given Sharepoint's unique environment. 

https://github.com/angular/material2/blob/e60f5fc391c103d489db948234bb6e9aaee0856b/src/cdk/overlay/keyboard/overlay-keyboard-dispatcher.ts#L44

There are also examples of running it outside of zones https://github.com/angular/material2/blob/b38b966b925b95e62b674012c2951f93737d02c7/src/cdk/a11y/focus-monitor/focus-monitor.ts#L191

PR #2189, opened 2018-04-14
[NG] emitting on string filter value change

### LINKED ISSUE #2004: No clrFilterValueChange events?
```
[ x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Changing the value of a default string filter raises a clrFilterValueChange event.

### Actual behavior
No event is raised.

### Reproduction of behavior
https://stackblitz.com/edit/clarity-light-theme-v10-yh9hrb

This logs to the console if the clrFilterValueChange event is raised on any of the three columns. Use any of the filters with any string, nothing is logged. 

For comparison, it also logs on clrDgSortedChange events, which can be seen by sorting any of the columns.

### Environment details

* **Angular version:** 4.3.4

* **Clarity version:** 0.10.23, 0.10.24

* **OS and version:** Windows 10 Enterprise

* **Browser:** [Chrome  64.0.3282.167 | Firefox 47.0.2 | IE 11]

Shijir:
When a user types in to filter, `clr-dg-string-filter` emits value through the `clrFilterValueChange` output. However, `clr-dg-column` was merely setting its `filterValue` property on those filter value changes sourced from `clr-dg-string-filter`, but wasn't emitting them from itself. 

So this fix lets `clr-dg-column` emit the filter value changes, that are first emitted from `clr-dg-string-filter`.

The demo of this fix (in the "Sort and Filter" section): http://2004-after-fix.surge.sh/datagrids

- Fixes #2004

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.ts
@@ -20,7 +20,6 @@ import {DatagridFilterRegistrar} from "./utils/datagrid-filter-registrar";
 
 let nbCount: number = 0;
 
-
 @Component({
     selector: "clr-dg-column",
     template: `
@@ -31,7 +30,7 @@ let nbCount: number = 0;
             <clr-dg-string-filter
                     *ngIf="field && !customFilter"
                     [clrDgStringFilter]="registered"
-                    (clrFilterValueChange)="filterValueChange.emit($event)"></clr-dg-string-filter>
+                    [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
 
             <ng-template #columnTitle><ng-content></ng-content></ng-template>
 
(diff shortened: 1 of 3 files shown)


src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -31,7 +31,7 @@ let nbCount: number = 0;
             <clr-dg-string-filter
                     *ngIf="field && !customFilter"
                     [clrDgStringFilter]="registered"
-                    [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
+                    (clrFilterValueChange)="filterValueChange.emit($event)"></clr-dg-string-filter>
```
adityarb88: Why did we lose the `@Input` `clrFilterValue` here? With this change, we are only binding on the `Output`, no? Does pre-setting a filter value work after this change?

Shijir: We can still pre-set the filter value. I noticed that having two-way binding with `[(clrFilterValue)]` on `clr-dg-string-filter` had no impact. Inside the clr-`dg-column`, we actually set the filter value programmatically [here](https://github.com/Shijir/clarity/blob/bf77f817c6b86cb6ebc6740d1abbf134ab991653/src/clr-angular/data/datagrid/datagrid-column.ts#L318). 

To see whether it had impact or not, we actually tested the pre-setting filter value before. And I haven't made changes to those pre-existing tests and they are still passing with the changes I made. Also, you can see the filter was pre-set with a value, "d" in the [demo](http://2004-after-fix.surge.sh/datagrids) (in the "Sort and Filter" section).

So we didn't need to set the value to `clr-dg-string-filter` using template bindings.

PR #2202, opened 2018-04-19
[SCHEMATICS] update paths for new builds

gnomeontherun:
There are path changes in the latest CLI config files.

### DIFF
--- a/packages/schematics/src/add/index.ts
+++ b/packages/schematics/src/add/index.ts
@@ -135,18 +135,19 @@ export default function(options: ComponentOptions): Rule {
 
                 const scriptsSearch = scripts.join("|");
                 const stylesSearch = styles.join("|");
+                const pathPrefix = (json.apps) ? "../" : "";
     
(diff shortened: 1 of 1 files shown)


packages/schematics/src/add/index.ts
```diff
@@ -135,18 +135,19 @@ export default function(options: ComponentOptions): Rule {
 
                 const scriptsSearch = scripts.join("|");
                 const stylesSearch = styles.join("|");
+                const pathPrefix = (json.apps) ? "../" : "";
     
-                if (stylesSearch.search("../node_modules/@clr/ui/clr-ui") < 0) {
-                    styles.push("../node_modules/@clr/ui/clr-ui.min.css");
+                if (stylesSearch.search("node_modules/@clr/ui/clr-ui") < 0) {
```
Shijir: why not just `indexOf` instead of `search`? Looks like we are not searching for a regular expression... 🤔

PR #2223, opened 2018-04-27
[NG] Select 2.0 Initial Commit

adityarb88:
Select 2.0 PR#1:

**Scope:** Focus on input to open popover

**Other changes:**
1. Rearranged Demo File Structure
2. Created Component File Structure
3. Created CSS files

**Popover closed:**
image

**Popover open:**
image

**Tested on Chrome**

**Note:** We haven't finalized on `contenteditable`. There are some issues associated with it which we will try to resolve and take it from there.


src/app/selects/basic-ng-select.html
```diff
@@ -0,0 +1,11 @@
+<!--
+  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<clr-select>
+    <clr-select-menu>
+        Select Menu
+    </clr-select-menu>
```
youdz: Now that I see it, how about `<clr-options>` (with an "s") instead of `<clr-select-menu>`? It feels clearer, and "menu" is really off for a select.

adityarb88: Yes. I like that. Will update it to `clr-options`. Thanks!

src/app/selects/select-static.html
```diff
@@ -0,0 +1,151 @@
...
+        <div class="form-group">
+            <label for="selects_2">Text field paired with a select</label>
+            <input type="text" id="selects_2" placeholder="Test">
+            <div class="select">
+                <select id="selects_3">
+                    <option>New York</option>
+                    <option>San Francisco</option>
+                    <option>Austin</option>
+                    <option>Paris</option>
+                    <option>Seattle</option>
```
youdz: Thank you for not putting Jersey City in there. 👍

adityarb88: Just because of this I will make sure I add it in there. 😝

src/clr-angular/forms/select/all.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * The goal is to have the tests properly grouped in the reporter, instead of having them all
+ * over the place because we load them asynchronously.
+ */
+
+import SelectSpecs from "./select.spec";
+import SelectMenuSpecs from "./select-menu.spec";
+import {addHelpers} from "../../data/datagrid/helpers.spec";
+
+fdescribe("Select component", function() {
```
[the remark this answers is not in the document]
adityarb88: I wasnt. Thanks for pointing it out.

src/clr-angular/forms/select/_select-angular.clarity.scss
```diff
@@ -0,0 +1,33 @@
...
+        min-height: $clr-ng-select-min-height;
+        width: $clr-ng-select-min-width;
+
+        //Mixins
+        @include form-fields-appearance($clr-forms-border-color);
+        @include clr-getTypePropertiesForDomElement(form_label, (font-size));
+        @include input-border-bottom-animation($clr-forms-focused-color);
+
+        //Others
+        outline: none;
```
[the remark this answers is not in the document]
adityarb88: For regular inputs, the blue animation of the bottom border is the `focus` indicator. Thats how it is for all form inputs.

src/clr-angular/forms/select/_select-angular.clarity.scss
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@import "../../popover/dropdown/menu-mixins.clarity";
+
+@include exports('select-angular.clarity') {
+    .clr-select{
```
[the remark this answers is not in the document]
adityarb88: yep. will fix it.

src/clr-angular/forms/select/_select-angular.clarity.scss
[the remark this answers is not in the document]
Shijir: So no visual indicator here when it's focused?

src/clr-angular/forms/select/select-menu.html
```diff
@@ -0,0 +1,8 @@
+<!--
+* Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+* This software is released under MIT license.
+* The full license information can be found in LICENSE in the root directory of this project.
+-->
+
+<!--Option Groups and Options will be projected here-->
+<ng-content></ng-content>
```
Shijir: Are we expecting this template to be expanded in the future? Otherwise, can we have it in as `TemplateUrl:`?

CarreraPHP:
I am one step away from rewriting my current project with Clarity. Hope to see the select 2.0 implemented soon.

src/clr-angular/forms/select/_select-angular.clarity.scss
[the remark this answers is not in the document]
adityarb88: image

This visual indicator is the blue bottom border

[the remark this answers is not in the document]
Shijir: oh sorry, I must have forgotten this blue line animation.

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,54 @@
...
+@Component({
+    selector: "clr-select",
+    templateUrl: "select.html",
+    providers: [
+        IfOpenService,
+        {provide: POPOVER_HOST_ANCHOR, useExisting: ElementRef
+    }],
+    host: {"[class.clr-select]": "true"}
+})
+export class ClrSelect implements OnDestroy {
```
Shijir: no `AfterContentInit`?

adityarb88: forgot. will add. ty!

src/clr-angular/forms/select/select-menu.ts
```diff
@@ -0,0 +1,66 @@
...
+        this.closeOnOutsideClick = true;
+    }
+
+    private initializeSubscriptions(): void {
+        if (this._ifOpenService) {
+            const sub: Subscription = this._ifOpenService.ignoredElementChange.subscribe((el: ElementRef) => {
+                if (el) {
+                    this.ignoredElement = el.nativeElement;
+                }
+                sub.unsubscribe();
```
adityarb88: Now that I think about this, I should add the unsubscribe in the onDestroy itself. If by chance `open` doesnt change, this subscription will never unsubscribe.

adityarb88: cc: @youdz

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,54 @@
...
+    open: boolean;
```
Shijir: Should this be `private`? What's this property for? We assign to it in the `this._ifOpenService.openChange`, but didn't use it anywhere else.

adityarb88: Now I am thinking this variable and the subscription are actually not doing anything. They were required before but @youdz asked me to move things around. So I think I missed removing them during the cleanup. Will take a look at this. Thanks!

youdz:
On hold while we reach a consensus on the open/close interaction for keyboard users.

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,54 @@
...
+            this.open = open;
+        });
+    }
+
+    private registerPopoverIgnoredInput() {
+        if (this.ignoredInput) {
+            this._ifOpenService.registerIgnoredElement(this.ignoredInput);
+        }
+    }
+
```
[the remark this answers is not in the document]
adityarb88: actually I was planning on using `focusout` https://developer.mozilla.org/en-US/docs/Web/Events/focusout

src/clr-angular/forms/select/options.ts
```diff
@@ -0,0 +1,68 @@
...
+                    this.ignoredElement = el.nativeElement;
+                }
+                this.sub.unsubscribe();
+            });
+        }
+    }
+
+    // Lifecycle hooks
+    ngOnDestroy() {
+        if (!this.sub.closed) {
```
Shijir: Shouldn't this be just `if(this.sub)`?

adityarb88: no because I also unsubscribe on line 57. If that unsubscribe hasn't been triggered, only then should we unsubscribe here too.

Shijir: But what if this component gets destroyed before this `initializeSubscriptions()` method called? Also, if `this.ifOpenService` is `falsy`, then the subscription wouldn't be set. Then the `this.sub` will be `undefined` though, right?

adityarb88: I dont understand. If the component gets destroyed before the `initializeSubscriptions` then the `Subscription` wont be initialized in the first place. Why would we need to unsubscribe in that case? And I don't think thats a valid use case or else we would have to check if the subscription is null or not before unsubscribing in all our components.

Shijir: If we are subscribing under a certain condition, we should check if the subscription property is truthy before calling its `unsubscribe()` method in `ngOnDestroy()`. Otherwise, that would throw an error as we try to call a method of `undefined`. In this case, we are subscribing under this condition, `if(this.ifOpenService)`.  And, I believe we actually do this in some of our components. I found `label.ts` does just that.

adityarb88: wait are you talking about this condition? `if (this.ifOpenService) {`

If yes, yes I have to remove that condition. Its always going to be true.

Shijir: Yes, I was talking about that condition :D

src/clr-angular/utils/conditional/if-open.service.spec.ts
```diff
@@ -56,6 +61,23 @@ export default function(): void {
...
+
+            it("provides an observable for ignored element changes", () => {
+                let testEl: ElementRef;
+                const sub: Subscription = ifOpenService.ignoredElementChange.subscribe((el: ElementRef) => {
+                    testEl = el;
+                });
+
+                ifOpenService.registerIgnoredElement(new ElementRef(null));
+
+                expect(testEl).not.toBeUndefined();
```
youdz: Nitpick: wouldn't it be slightly better to do this, since it's pretty much free?
```typescript
const el = new ElementRef(null);
ifOpenService.registerIgnoredElement(el);
expect(testEl).toBe(el);
```

src/clr-angular/utils/conditional/if-open.service.ts
```diff
@@ -83,4 +83,18 @@ export class IfOpenService {
...
         delete this.originalEvent;
     }
+
+    /**
+     *  Popovers might need to ignore click events on an element
+     *  (eg: popover opens on focus on an input field. Clicks should be ignored in this case)
+     */
+    private _ignoredElementChange: Subject<ElementRef> = new Subject<ElementRef>();
+
+    get ignoredElementChange(): Observable<ElementRef> {
```
youdz: I'm guessing you need an Observable because this is set too late in the lifecycle of the select menu?

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,44 @@
...
+    constructor(private ifOpenService: IfOpenService) {}
+
+    private registerPopoverIgnoredInput() {
+        if (this.input) {
+            this.ifOpenService.registerIgnoredElement(this.input);
+        }
+    }
+
+    onClick(event: MouseEvent): void {
```
youdz: As a general rule, I don't like `onClick` or `onHover` as method names, it makes it very hard to read the template. The name conveys no information here. It's much better to call them by what they do, in your case it would be `toggleMenu` or something similar.

Note to self: add this to the coding guidelines.

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,44 @@
...
+        this.ifOpenService.toggleWithEvent(event);
+        if (this.input) {
+            this.input.nativeElement.focus();
```
youdz: I think you have a bug if the menu is open and you click on the caret again. The click will force re-focus on the input over and over.

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+                context.clarityDirective.onClick(new MouseEvent("click"));
+
+                expect(open).not.toBe(false);
+            });
+
+            it("provides a method to close the popover on focusout", () => {
+                expect(ifOpenService.open).toBeUndefined();
+
+                context.clarityDirective.onFocusOut();
+                expect(ifOpenService.open).toBe(false);
```
youdz: That's a very strange test, going from `undefined` to `false`.  But ok.

... I guess it'd be a bitt better if it _started_ with `ifOpenService.open = true` instead of the first `expect`.

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+                trigger.click();
+
+                focused = context.clarityElement.querySelector(":focus");
+
+                expect(focused.classList.contains("clr-select-input")).toBe(true);
+            });
+
+            it("keeps the options menu open when the input is clicked", () => {
+                const trigger = context.clarityElement.querySelector(".clr-select-trigger");
+                trigger.click();
```
youdz: You're replicating the `"opens the menu on the trigger click"` test here. Just set `ifOpenService.open = true` directly, then test what you really want to test here.

src/clr-angular/forms/select/options.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        // Initialize Services
+        this.ifOpenService = ifOpenService;
+
+        // Configure Popover
+        this.initializeSubscriptions();
+        this.configurePopover();
+    }
+
+    @HostBinding("class.open")
```
youdz: The other popover components all have the `.open` class on the parent component, not the popover itself. Any reason why you changed this for select? I'm okay either way, I'm mostly curious.

src/clr-angular/utils/conditional/if-open.service.spec.ts
adityarb88: done.

src/clr-angular/utils/conditional/if-open.service.ts
adityarb88: yes

src/clr-angular/forms/select/select.ts
adityarb88: good point. fixed.

src/clr-angular/forms/select/select.spec.ts
adityarb88: Here the `ifOpenService.open` was initialized and thats why I checked from `undefined` to `false` but I can see how this can be confusing. Changing to test from `true` to `false`

src/clr-angular/forms/select/select.ts
adityarb88: My thinking here was that even if you click on the caret the `input` should be focused because the native select too works the same way, no? In the native select, the trigger doesn't not function separately. Let me know what you think about this @youdz.

youdz: You're right, the focus should stay focused after closing the menu.
But as we saw during our call, we still have a focus/open/close bug.

src/clr-angular/forms/select/options.ts
adityarb88: This was a mistake but now that I think about it, if I move it to the parent it creates a nesting in the CSS. Even though its just 2 levels of nesting, I feel that its unnecessary.

So instead of:
```
.clr-select.open .clr-options {
    //do whatever
}
```

I can just do:
```
.clr-options.open {
    //do whatever
}
```

So i think we should keep this pattern instead of what we normally do. Do you think thats alright?

src/clr-angular/forms/select/options.ts
```diff
@@ -4,30 +4,30 @@
...
-                ifOpenService: IfOpenService) {
+                ifOpenService: IfOpenService, blurEventDisableService: BlurEventDisableService) {
         if (!parentHost) {
-            throw new Error("clr-select-menu should only be used inside of a clr-select");
+            throw new Error("clr-options should only be used inside of a clr-select");
         }
         super(injector, parentHost);
 
         // Initialize Services
         this.ifOpenService = ifOpenService;
```
youdz: This is still not needed, `AbstractPopover` already sets it. 😛 It also means you don't even need to inject the `IfOpenService` in the constructor here.

src/clr-angular/forms/select/options.spec.ts
```diff
@@ -31,6 +31,26 @@ export default function(): void {
             ifOpenService = context.getClarityProvider(IfOpenService);
         });
 
+        describe("Typescript API", function() {
+            it("has a host listener on mouseenter", () => {
```
youdz: Nitpick: `"Has a host listener"` isn't descriptive. It'd be better with `"disables blur events on mouseenter"`.
Same comment for the next test, obviously.

src/clr-angular/forms/select/providers/blur-event-disable.service.ts
```diff
@@ -0,0 +1,27 @@
...
+@Injectable()
+export class BlurEventDisableService {
+    private _blurEventDisableChange: Subject<boolean> = new Subject<boolean>();
+
+    get blurEventDisableChange(): Observable<boolean> {
+        return this._blurEventDisableChange.asObservable();
+    }
+
+    enableBlurEvent(): void {
+        this._blurEventDisableChange.next(true);
```
youdz: `enable()` sets the `DisableService` to `true`.
I'm pretty sure you want either to swap the `true/false`, or rename the service. 😉

src/clr-angular/forms/select/options.ts
youdz: Do we style the input in any way when the menu is open? Or do we plan on doing so?

I'm not even sure what you're going to use the selector you just showed for. I don't see any styles that apply to the menu only when it is open, since... well... when it's not open it's not displayed.

src/clr-angular/forms/select/providers/blur-event-disable.service.ts
youdz: You can keep the name if you want, but then just use `disabled` in your components too (like [here](https://github.com/vmware/clarity/pull/2223/files/f975e60091eac9ec64b988ea4726ceac58770e2f..ceaccae974b442aa297969ff0ac2d52099f438d4#diff-b8e1571d74b2bc9b2f8106a47b7802cbR25)) so that `true` means "disabled" everywhere.

I just laughed again at [this line](https://github.com/vmware/clarity/pull/2223/files/f975e60091eac9ec64b988ea4726ceac58770e2f..ceaccae974b442aa297969ff0ac2d52099f438d4#diff-f20561bd6e03550368d629f4c6cdd065R29):
```typescript
let blurEnabledService: BlurEventDisableService;
```

src/clr-angular/forms/select/select.ts
```diff
@@ -5,40 +5,59 @@
...
+    toggleOptionsMenu(event: MouseEvent): void {
         this.ifOpenService.toggleWithEvent(event);
         if (this.input) {
             this.input.nativeElement.focus();
         }
     }
 
-    onFocusOut(): void {
-        this.ifOpenService.open = false;
+    onInputBlur(event: FocusEvent): void {
```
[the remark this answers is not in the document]
youdz: Sure, sounds better. 👍

src/clr-angular/forms/select/providers/blur-event-disable.service.ts
adityarb88: doesnt matter now. removing the service completely. using the tab key event listener instead like we discussed.

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -99,6 +115,18 @@ export default function(): void {
...
+            it("focuses the input when the menu is clicked", () => {
+                ifOpenService.open = true;
+
+                const menu = context.clarityElement.querySelector(".menu");
+
+                menu.click();
+
+                const focused = context.clarityElement.querySelector(":focus");
+
+                expect(focused.classList.contains("clr-select-input")).toBe(true);
```
youdz: Small mistake: https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#unit-testing
> Unit tests should not duplicate coverage. In particular, avoid multiple unit tests failing for the same error.

You're indirectly re-testing the `focusInput()` method here. You want to spy on this method, and expect it to have been called.

EDIT: I just noticed you have the right test a few lines below. Not sure what this one is for then.

adityarb88: Yep. I added the Host click listener in the last commit and forgot to remove this test. Now the new test that you mentioned below tests for this. Will remove this test.

Shijir:
@adityarb88 looks good to me. I've also pulled this branch, tried it locally. The focus working now makes sense 👍

PR #2230, opened 2018-05-02
[ng] new spinner button (#1488)

jeeyun:
Implementation of the new spinner button. After reviewing with @youdz came to a couple of realization which led to the approach in this PR:

- the new implementation should be non-breaking, so need to continue to support [clrLoading] on button with a boolean value input, while allowing newer approach of taking in different button states
- we want to continue to use clrLoading / LoadingListener to not duplicate this logic in button when many other components already utilize this directive.
- clrLoading is now modified to handle more than two states (loading / not loading)

### DIFF
--- a/src/clr-angular/button/button-group/button.spec.ts
+++ b/src/clr-angular/button/button-group/button.spec.ts
@@ -7,7 +7,7 @@
 import {Component, DebugElement, ViewChild} from "@angular/core";
 import {ComponentFixture, TestBed} from "@angular/core/testing";
 
-import {ClrLoadingModule} from "../../utils/loading";
+import {ClrLoadingModule, ClrLoadingState} from "../../utils/loading";
 import {ButtonInGroupService} from "../providers/button-in-group.service";
 
 import {ClrButton} from "./button";
@@ -151,14 +151,13 @@ export default function(): void {
             });
 
             it("implements LoadingListener", () => {
-                expect(componentInstance.button1.startLoading).toBeDefined();
-                expect(componentInstance.button1.doneLoading).toBeDefined();
+                expect(componentInstance.button1.loadingStateChange).toBeDefined();
             });
 
             it("sets loading to true on start loading", () => {
                 expect(componentInstance.button1.loading).toBeUndefined();
 
-                componentInstance.button1.startLoading();
+                componentInstance.button1.loadingStateChange(ClrLoadingState.LOADING);
 
                 expect(componentInstance.button1.loading).toBe(true);
             });
@@ -166,7 +165,7 @@ export default function(): void {
             it("sets loading to false on done loading", () => {
                 expect(componentInstance.button1.loading).toBeUndefined();
 
-                componentInstance.button1.doneLoading();
+                componentInstance.button1.loadingStateChange(ClrLoadingState.DEFAULT);
 
                 expect(componentInstance.button1.loading).toBe(false);
             });
--- a/src/clr-angular/button/button-loading/loading-button.module.ts
+++ b/src/clr-angular/button/button-loading/loading-button.module.ts
@@ -7,16 +7,12 @@
 import {CommonModule} from "@angular/common";
 import {NgModule, Type} from "@angular/core";
 
-import {ClrLoadingModule} from "../../utils/loading/loading.module";
 import {ClrLoadingButton} from "./loading-button";
 
 export const CLR_LOADING_BUTTON_DIRECTIVES: Type<any>[] = [ClrLoadingButton];
 
-@NgModule({
-    imports: [CommonModule, ClrLoadingModule],
-    declarations: [CLR_LOADING_BUTTON_DIRECTIVES],
-    exports: [CLR_LOADING_BUTTON_DIRECTIVES, ClrLoadingModule]
-})
+@NgModule(
+    {imports: [CommonModule], declarations: [CLR_LOADING_BUTTON_DIRECTIVES], exports: [CLR_LOADING_BUTTON_DIRECTIVES]})
 export class ClrLoadingButtonModule {
 }
 
--- a/src/clr-angular/button/button-loading/loading-button.ts
+++ b/src/clr-angular/button/button-loading/loading-button.ts
@@ -5,26 +5,91 @@
  */
 
 
-import {Component} from "@angular/core";
-import {LoadingListener} from "../../utils/loading/loading-listener";
-
+import {animate, keyframes, style, transition, trigger} from "@angular/animations";
+import {Component, ElementRef, EventEmitter, Output, Renderer2} from "@angular/core";
+import {ClrLoadingState, LoadingListener} from "../../utils/loading";
 
 @Component({
     selector: "button[clrLoading]",
     template: `
-        <span class="spinner spinner-inline" *ngIf="loading"></span>
-        <ng-content></ng-content>
+        <ng-container [ngSwitch]="state">
+            <span *ngSwitchCase="buttonState.LOADING">
+                <span @spinner class="spinner spinner-inline"></span>
+            </span>
+            <span *ngSwitchCase="buttonState.SUCCESS">
+                <span @validated class="spinner spinner-inline spinner-check"></span>
+            </span>
+            <span *ngSwitchCase="buttonState.DEFAULT" @defaultButton>
+                <ng-content></ng-content>
+            </span>
+        </ng-container>
     `,
-    providers: [{provide: LoadingListener, useExisting: ClrLoadingButton}]
+    providers: [{provide: LoadingListener, useExisting: ClrLoadingButton}],
+    animations: [
+        trigger("defaultButton",
+                [
+                    transition(":enter", [style({opacity: 0}), animate("200ms 100ms ease-in", style({opacity: 1}))]),
+                    // TODO: see if we can get leave animation to work before spinner's enter animation
+                    transition(":leave", [style({opacity: 0})])
+                ]),
+        trigger("spinner",
+                [
+                    transition(":enter", [style({opacity: 0}), animate("200ms 100ms ease-in", style({opacity: 1}))]),
+                    transition(":leave", [style({opacity: 1}), animate("100ms ease-out", style({opacity: 0}))])
+                ]),
+        trigger("validated",
+                [
+                    transition(":enter",
+                               [animate("300ms", keyframes([
+                                            style({transform: "scale(0,0)"}), style({opacity: 1}),
+                                            style({transform: "scale(1.2,1.2)"}), style({transform: "scale(.9,.9)"}),
+                                            style({transform: "scale(1,1)"})
+                                        ]))]),
+                    transition(":leave", [style({opacity: 1}), animate("100ms ease-out", style({opacity: 0}))])
+                ])
+    ]
 })
 export class ClrLoadingButton implements LoadingListener {
-    public loading: Boolean;
+    public buttonState = ClrLoadingState;
+    public state: ClrLoadingState = ClrLoadingState.DEFAULT;
 
-    startLoading(): void {
-        this.loading = true;
+    @Output("clrLoadingChange")
+    public clrLoadingChange: EventEmitter<ClrLoadingState> = new EventEmitter<ClrLoadingState>(false);
+
+    constructor(private el: ElementRef, private renderer: Renderer2) {}
+
+    loadingStateChange(state: ClrLoadingState): void {
+        this.state = state;
+
+        switch (state) {
+            case ClrLoadingState.DEFAULT:
+                this.renderer.removeStyle(this.el.nativeElement, "width");
+                this.renderer.removeAttribute(this.el.nativeElement, "disabled");
+                break;
+            case ClrLoadingState.LOADING:
+                this.setExplicitButtonWidth();
+                this.renderer.setAttribute(this.el.nativeElement, "disabled", "");
+                break;
+            case ClrLoadingState.SUCCESS:
+                this.setExplicitButtonWidth();
+                setTimeout(() => {
+                    this.loadingStateChange(ClrLoadingState.DEFAULT);
+                }, 1000);
+                break;
+            case ClrLoadingState.ERROR:
+                this.loadingStateChange(ClrLoadingState.DEFAULT);
+                break;
+            default:
+                break;
+        }
+
+        this.clrLoadingChange.emit(state);
     }
 
-    doneLoading(): void {
-        this.loading = false;
+    private setExplicitButtonWidth() {
+        if (getComputedStyle) {
+            const width = getComputedStyle(this.el.nativeElement).getPropertyValue("width");
+            this.renderer.setStyle(this.el.nativeElement, "width", width);
+        }
     }
 }
--- a/src/clr-angular/image/_icons.clarity.scss
+++ b/src/clr-angular/image/_icons.clarity.scss
@@ -21,6 +21,14 @@ $svg_data: "data:image/svg+xml;charset=utf8";
     @return url("#{$svg_data},#{$warning-icon}");
 }
 
+@function generateCheckIcon($fillColor:$gray-dark) {
+    $check-icon: "%3Csvg%20version%3D%221.1%22%20viewBox%3D%220%200%2036%2036%22%20preserveAspectRatio%3D%22xMidYMid%20meet%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20xmlns%3Axlink%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%22%20focusable%3D%22false%22%20aria-hidden%3D%22true%22%20role%3D%22img%22%3E%3Cpath%20fill%3D%22%23"
+    + str-slice($fillColor + "",2)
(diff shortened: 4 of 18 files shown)


src/app/buttons/button-loading.ts
```diff
@@ -4,35 +4,45 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import {Component} from "@angular/core";
+import {ClrLoadingState} from "../../clr-angular/utils/loading";
```
youdz: The app should always import from `"../../clr-angular"`, since it acts as a consumer of Clarity and anything deeper is not public.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,84 @@
...
+import {Component, ElementRef, EventEmitter, Output, Renderer2} from "@angular/core";
+import {ClrLoadingState, LoadingListener} from "../../utils/loading";
 
 @Component({
     selector: "button[clrLoading]",
     template: `
-        <span class="spinner spinner-inline" *ngIf="loading"></span>
-        <ng-content></ng-content>
+        <ng-container [ngSwitch]="state">
+            <span *ngSwitchCase="buttonState.LOADING" class="spinner-btn-wrapper">
```
youdz: Did the switch solve your animation issue?

src/clr-angular/image/_icons.clarity.scss
```diff
@@ -21,6 +21,14 @@ $svg_data: "data:image/svg+xml;charset=utf8";
     @return url("#{$svg_data},#{$warning-icon}");
 }
 
+@function generateCheckIcon($fillColor:$gray-dark) {
```
youdz: Wait, why are we not using a `<clr-icon>` here? We're trying _not_ to have any encoded SVGs in the CSS anymore.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,84 @@
...
+
+    updateLoadingState(state: ClrLoadingState): void {
+        this.state = state;
+
+        switch (state) {
+            case ClrLoadingState.DEFAULT:
+                this.renderer.removeAttribute(this.el.nativeElement, "disabled");
+                break;
+            case ClrLoadingState.LOADING:
+                const width = getComputedStyle(this.el.nativeElement).getPropertyValue("width");
```
youdz: I'll let @gnomeontherun confirm, but I believe we want to wrap this in a check for server-side rendering.

src/clr-angular/utils/loading/loading-listener.ts
```diff
@@ -4,12 +4,10 @@
...
  * This is an abstract class because we need it to still be a valid token for dependency injection after transpiling.
  * This does not mean you should extend it, simply implementing it is fine.
  */
-export abstract class LoadingListener {
-    abstract startLoading(): void;
-
-    abstract doneLoading(): void;
-}
\ No newline at end of file
+export abstract class LoadingListener { abstract updateLoadingState(state: ClrLoadingState): void; }
```
youdz: I know I'm a PITA, but this "update" method feels like calling it will update the loading state itself, but it doesn't. It's what's called _when_ the loading state has been updated.

I feel like `loadingStateChange(state: ClrLoadingState)` shows better what the method should do: update the view based on a new loading state.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,84 @@
...
+                this.renderer.setStyle(this.el.nativeElement, "width", width);
+                this.renderer.setAttribute(this.el.nativeElement, "disabled", "");
+                break;
+            case ClrLoadingState.SUCCESS:
+                setTimeout(() => {
+                    this.updateLoadingState(ClrLoadingState.DEFAULT);
+                }, 1000);
+                break;
+            case ClrLoadingState.ERROR:
+                this.updateLoadingState(ClrLoadingState.DEFAULT);
```
youdz: You're not updating the actual Loading directive here. Maybe it doesn't matter, but it feels weird. Does that mean we need a second method on `LoadingListener` to allow for communication both ways?

src/clr-angular/utils/loading/loading.module.ts
```diff
@@ -6,13 +6,15 @@
...
 
-import {ClrLoading} from "./loading";
+import {ClrLoading, ClrLoadingState} from "./loading";
 
 export const CLR_LOADING_DIRECTIVES: Type<any>[] = [ClrLoading];
 
 @NgModule({imports: [CommonModule], declarations: [CLR_LOADING_DIRECTIVES], exports: [CLR_LOADING_DIRECTIVES]})
 export class ClrLoadingModule {}
 
+export const loadingState = ClrLoadingState;
```
youdz: Not sure what this renamed export is for. I might have missed it somewhere. But it looks like it can be safely removed.

src/clr-angular/utils/loading/loading.ts
```diff
@@ -7,32 +7,46 @@ import {Directive, Input, OnDestroy, Optional} from "@angular/core";
...
             return;
         }
-        this._loading = value;
-        if (this.listener) {
+
+        if (value === null) {
+            value = false;
+        }
+
+        if (typeof value === "boolean") {
```
youdz: This if should go _before_ the first one, so that setting `true` when the state is already `LOADING` is ignored too.

src/app/buttons/button-loading.ts
jeeyun: 👍

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: Nope, unfortunately the same results as using `*ngIf`.

src/clr-angular/image/_icons.clarity.scss
jeeyun: I'm making use of existing `spinner`, `spinner-inline` classes and it makes it a breaking change to get rid of using SVG completely in spinner (or alternatively, gets pretty messy trying to make it an exception for the checkbox case here only). Would it be ok to create a separate issue for getting rid of encoded SVGs in spinner-related components all together in the future?

src/clr-angular/utils/loading/loading-listener.ts
jeeyun: No problem. Will change the name.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -59,30 +59,38 @@ export class ClrLoadingButton implements LoadingListener {
 
     constructor(private el: ElementRef, private renderer: Renderer2) {}
 
-    updateLoadingState(state: ClrLoadingState): void {
+    loadingStateChange(state: ClrLoadingState): void {
         this.state = state;
 
         switch (state) {
             case ClrLoadingState.DEFAULT:
+                this.renderer.removeStyle(this.el.nativeElement, "width");
```
youdz: Why did you remove this? Does it mean the button never reverts to pure dynamic width?

src/clr-angular/image/_icons.clarity.scss
gnomeontherun: If there is a deprecation we want to make, let's get a ticket in to have that done for 0.12.

src/clr-angular/button/button-loading/loading-button.module.ts
```diff
@@ -7,15 +7,16 @@
...
-import {ClrLoadingModule} from "../../utils/loading/loading.module";
+import {ClrIconModule} from "../../icon";
+
 import {ClrLoadingButton} from "./loading-button";
 
 export const CLR_LOADING_BUTTON_DIRECTIVES: Type<any>[] = [ClrLoadingButton];
 
 @NgModule({
-    imports: [CommonModule, ClrLoadingModule],
+    imports: [CommonModule, ClrIconModule],
```
gnomeontherun: Why do you need `ClrIconModule`?

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,92 @@
...
+            <span *ngSwitchCase="buttonState.DEFAULT" @defaultButton>
+                <ng-content></ng-content>
+            </span>
+        </ng-container>
     `,
-    providers: [{provide: LoadingListener, useExisting: ClrLoadingButton}]
+    providers: [{provide: LoadingListener, useExisting: ClrLoadingButton}],
+    animations: [
+        trigger("defaultButton",
+                [
```
gnomeontherun: Is this formatting just what clang does automatically? I think its hard to read, if we can do anything about it.

src/clr-angular/button/button-loading/loading-button.module.ts
jeeyun: Oh this was when I was testing out using clr-icon for the spinner graphics. Will remove.

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: Some of it is clang. Let me see if I can improve by at least separating the comment into a new line.

src/app/buttons/button-loading.html
```diff
@@ -6,23 +6,23 @@
   -->
 
 <h4>Loading Buttons</h4>
-<button [clrLoading]="validateLoading" class="btn btn-info-outline" (click)="validateDemo()">Validate</button>
-<button [clrLoading]="submitLoading" type="submit" class="btn btn-success-outline" (click)="submitDemo()">Submit</button>
+<button [clrLoading]="validateState" class="btn btn-primary" (click)="validateDemo()">Validate</button>
+<button [clrLoading]="submitState" type="submit" class="btn btn-success-outline" (click)="submitDemo()">Submit</button>
 
 <pre>
```
adityarb88: Can we remove the code snippets from the demos? We have decided to do that across all demos now. We will also get rid of prismjs from clarity & our dev app.

src/app/buttons/button-loading.ts
```diff
@@ -4,35 +4,45 @@
...
+            this.validateState = ClrLoadingState.SUCCESS;
+        }, 1500);
     }
 
     submitDemo() {
-        this.submitLoading = true;
-        setTimeout(() => this.submitLoading = false, 1500);
+        this.submitState = ClrLoadingState.LOADING;
+        setTimeout(() => {
+            this.submitState = ClrLoadingState.DEFAULT;
```
adityarb88: What is the difference between changing from `LOADING` to `SUCCESS` like you did in the `validateDemo()` method vs changing from `LOADING` to `DEFAULT`?

src/app/buttons/button-loading.ts
```diff
@@ -4,35 +4,45 @@
...
+        }, 1500);
     }
 
     validateSmDemo() {
-        this.validateSmLoading = true;
-        setTimeout(() => this.validateSmLoading = false, 1500);
+        this.validateSmState = true;
+        setTimeout(() => {
+            this.validateSmState = false;
```
adityarb88: Why is this different than the other methods? I am assuming that we are testing that the old API still works.

src/clr-angular/button/button-loading/loading-button.module.ts
```diff
@@ -7,16 +7,12 @@
...
 
 export const CLR_LOADING_BUTTON_DIRECTIVES: Type<any>[] = [ClrLoadingButton];
 
-@NgModule({
-    imports: [CommonModule, ClrLoadingModule],
-    declarations: [CLR_LOADING_BUTTON_DIRECTIVES],
-    exports: [CLR_LOADING_BUTTON_DIRECTIVES, ClrLoadingModule]
-})
+@NgModule(
+    {imports: [CommonModule], declarations: [CLR_LOADING_BUTTON_DIRECTIVES], exports: [CLR_LOADING_BUTTON_DIRECTIVES]})
```
adityarb88: Is `ClrLoadingModule` no longer needed in the imports?

src/clr-angular/button/button-group/button.spec.ts
```diff
@@ -151,22 +151,21 @@ export default function(): void {
...
+                componentInstance.button1.loadingStateChange(ClrLoadingState.LOADING);
 
                 expect(componentInstance.button1.loading).toBe(true);
             });
 
             it("sets loading to false on done loading", () => {
                 expect(componentInstance.button1.loading).toBeUndefined();
 
-                componentInstance.button1.doneLoading();
+                componentInstance.button1.loadingStateChange(ClrLoadingState.DEFAULT);
```
adityarb88: Are we missing a test for `ClrLoadingState.SUCCESS`?

src/app/buttons/button-loading.ts
adityarb88: Nvm. I should have read `loading-button.ts` first :D

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,91 @@
...
+import {ClrLoadingState, LoadingListener} from "../../utils/loading";
 
 @Component({
     selector: "button[clrLoading]",
     template: `
-        <span class="spinner spinner-inline" *ngIf="loading"></span>
-        <ng-content></ng-content>
+        <ng-container [ngSwitch]="state">
+            <span *ngSwitchCase="buttonState.LOADING">
+                <span @spinner class="spinner spinner-inline spinner-loading"></span>
```
adityarb88: What does the `.spinner-loading` class do and where is it declared? I dont see it in the CSS.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,91 @@
...
+            </span>
+            <span *ngSwitchCase="buttonState.SUCCESS">
+                <span @validated class="spinner spinner-inline spinner-loading spinner-check"></span>
```
adityarb88: `spinner-check` seems a bit weird to me. Couldn't we have just extended the `.spinner` styles and created a different class or even better, used the `clr-icon` here? We would just have to align that icon correctly in the button, no? Is the `.spinner` class adding any styles to the `checkmark` icon?

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,91 @@
...
+                    transition(":enter", [style({opacity: 0}), animate("200ms 100ms ease-in", style({opacity: 1}))]),
+                    // TODO: see if we can get leave animation to work before spinner's enter animation
+                    transition(":leave", [style({opacity: 0})])
+                ]),
+        trigger("spinner",
+                [
+                    transition(":enter", [style({opacity: 0}), animate("200ms 100ms ease-in", style({opacity: 1}))]),
+                    transition(":leave", [style({opacity: 1}), animate("100ms ease-out", style({opacity: 0}))])
+                ]),
+        trigger("validated",
```
adityarb88: should we rename this state to `done` or `completed` instead of `validated`?

src/app/buttons/button-loading.ts
jeeyun: Yes that's correct. To make sure it's a non-breaking change for old API.

src/clr-angular/button/button-loading/loading-button.module.ts
jeeyun: Everything in `ClrLoadingModule` is deprecated. I believe it is exported thru `utils` now and you can simply import from `@clr/angular`.

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: Hmm.. maybe a class I was using while implementing and no longer needed. Will doublecheck and remove if that's the case.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -5,26 +5,91 @@
...
+                [
+                    transition(":enter",
+                               [animate("300ms", keyframes([
+                                            style({transform: "scale(0,0)"}), style({opacity: 1}),
+                                            style({transform: "scale(1.2,1.2)"}), style({transform: "scale(.9,.9)"}),
+                                            style({transform: "scale(1,1)"})
+                                        ]))]),
+                    transition(":leave", [style({opacity: 1}), animate("100ms ease-out", style({opacity: 0}))])
```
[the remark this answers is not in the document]
jeeyun: 😄

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: `spinner-check` leans on the styles of `.spinner` and `.spinner-inline` and just overrides the background and animation. Using `clr-icon` would require breaking away from using any spinner classes since the `animation` and `background`are used there. Changing `spinner` class to use `clr-icon` is a breaking change for static spinner markup, so that was not an option. I created a separate ticket to look into getting rid of embedded SVGs from our components in the future: https://github.com/vmware/clarity/issues/2241. Let me know if my explanation is confusing. Happy to discuss on a call further.

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: Hmm.. I'm torn about that. Depending on the use case going from SPINNING to DEFAULT is considered "done" in some cases even though we skip the `validated` state.

src/app/buttons/button-loading.html
jeeyun: Yup will remove!

src/clr-angular/button/button-group/button.spec.ts
jeeyun: More extensive testing is done in `loading-button.spec.ts` including test for `ClrLoadingState.SUCCESS` :)

src/clr-angular/image/_icons.clarity.scss
jeeyun: issue created, though not sure this is amust for 0.12: https://github.com/vmware/clarity/issues/2241

PR #2232, opened 2018-05-03
[NG] *clrIfDragged structural directive

Shijir:
- Created this simple `*clrIfDragged` structural directive, which projects its content only during dragging.
- Also, I have extracted `emulateEventOn()` function from `drag-event-listener.spec.ts`, renamed that as `emulateDragEvent()`, and moved that into `helpers.spec.ts` as it will be used in many other incoming testing files.


src/clr-angular/utils/drag-and-drop/if-dragged.ts
```diff
@@ -0,0 +1,28 @@
...
+import {Subscription} from "rxjs/Subscription";
+
+import {ClrDragEvent} from "./interfaces/drag-event";
+import {ClrDragEventListener} from "./providers/drag-event-listener";
+
+@Directive({selector: "[clrIfDragged]"})
+export class ClrIfDragged<T> implements OnDestroy {
+    private subscriptions: Subscription[] = [];
+    constructor(private template: TemplateRef<any>, private container: ViewContainerRef,
+                private dragEventListener: ClrDragEventListener<T>) {
```
adityarb88: Could you please remind me where `ClrDragEventListener` is declared?

src/clr-angular/utils/drag-and-drop/if-dragged.ts
```diff
@@ -0,0 +1,28 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Directive, OnDestroy, TemplateRef, ViewContainerRef} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
+
+import {ClrDragEvent} from "./interfaces/drag-event";
+import {ClrDragEventListener} from "./providers/drag-event-listener";
+
+@Directive({selector: "[clrIfDragged]"})
```
adityarb88: So from our call you mentioned that this directive will be placed on the ghost element. So thats why we render it only after `dragStart`. May be add a comment here to describe what this directive does.

src/clr-angular/utils/drag-and-drop/if-dragged.ts
adityarb88: So it looks like `clrIfDragged` has to be under `clrDraggable`. If thats the case, I think we need to warn the user only to use this directive under `clrDraggable`. If they don't we should throw an error. Check `dropdown-menu` for an example. It does something like this:

```
constructor(injector: Injector, @Optional() @Inject(POPOVER_HOST_ANCHOR) parentHost: ElementRef,
                @Optional() @SkipSelf() nested: ClrDropdownMenu) {
        if (!parentHost) {
            throw new Error("clr-dropdown-menu should only be used inside of a clr-dropdown");
        }
       super(...)
       ....
```

Do you think we need a check like this here too?

src/clr-angular/utils/drag-and-drop/if-dragged.ts
Shijir: Sure, will do that.

src/clr-angular/utils/drag-and-drop/if-dragged.ts
Shijir: I think we could/should have it. 🤔 Thanks for bringing this up!

src/clr-angular/utils/drag-and-drop/if-dragged.ts
adityarb88: @youdz suggested this for all popover implementations. You should check with him once before we implement something similar here with a new InjectionToken.

src/clr-angular/utils/drag-and-drop/if-dragged.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        afterEach(function() {
+            this.fixture.destroy();
+        });
+
+        it("should not display anything on normal state", function() {
+            expect(this.draggableEl.textContent.trim()).toBe("");
+        });
+
+        it("should not display anything on mousemove or touchstart events", function() {
+            emulateDragEvent("mousedown", 0, 0, this.draggableEl);
```
youdz: You're actually testing `clrDraggable` all over again here, which is problematic. You shouldn't have a  `clrDraggable` element at all in this test, just the `*clrIfDragged` directive, and you should simply manipulate the service directly. You can declare the service as a root provider in the testing module, it works absolutely fine for this spec.

src/clr-angular/utils/drag-and-drop/if-dragged.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+            this.fixture.detectChanges();
+            expect(this.draggableEl.textContent.trim()).toBe("");
+
+            emulateDragEvent("touchstart", 0, 0, this.draggableEl);
+            this.fixture.detectChanges();
+            expect(this.draggableEl.textContent.trim()).toBe("");
+        });
+
+        it("should project its content during dragging with mouse events", function() {
```
youdz: It's not projection by the way. It's template instantiation.

src/clr-angular/utils/drag-and-drop/if-dragged.ts
youdz: I don't see why there would be an InjectionToken here, the current structure/injection is fine.

adityarb88: I commented on the wrong thread. Please check the conversation below.

src/clr-angular/utils/drag-and-drop/if-dragged.spec.ts
Shijir: But if the requirement is I shouldn't have a `clrDraggable` element at all in the test, I can manage it :P

src/clr-angular/utils/drag-and-drop/if-dragged.spec.ts
```diff
@@ -74,7 +82,14 @@ export default function(): void {
...
 
-@Component({template: `<div clrDraggable #draggable><div *clrIfDragged>Test</div></div>`})
-class IfDraggedTest {
-    @ViewChild("draggable") draggableEl: ElementRef;
+@Component({providers: [ClrDragEventListener], template: `<div *clrIfDragged>Test</div>`})
+class IfDraggedTest<T> {
+    public draggableEl: Node;
+    constructor(private el: ElementRef, private dragEventListener: ClrDragEventListener<T>) {
+        this.draggableEl = el.nativeElement;
+        dragEventListener.attachDragListeners(this.draggableEl);
```
youdz: You misunderstood me, it's not that I didn't want to use the `clrDraggable` specifically, it's that I didn't want to test stuff that you already test in another spec, using low-level mouse and touch events. With this latest version, if `ClrDragEventListener` has a bug, all the tests from this spec will fail too...

I realize `DragEventListener` doesn't offer a way to force a dragStart or dragEnd broadcast directly, so you can write a dummy mock that just exposes the two Subjects for instance. The whole point is for you to test that:
> When `ClrDragEventListener` emits `dragStarted`, the ghost instantiates the template.
not that:
> When user presses the left mouse button and starts moving the mouse, the ghost instantiates the template.

Shijir: Implemented the requested change. Please take a look again.

adityarb88:
@shijir I think you need to run clang:format and tslint:fix before you can merge this in.

Shijir:
@adityarb88 I did ran them yesterday. But now one of the gemini screens is not matching. Will fix that.

Issue #2219, opened 2018-04-26
[UI] Nested Dropdown Menu Indicator not vertically centered when screen size is less than 576px


adityarb88:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Nested Menu indicator should be vertically centered.

### Actual behavior

![image](https://user-images.githubusercontent.com/1426805/39326035-d67486d0-4961-11e8-9a84-da62c84d3259.png)

### Reproduction of behavior

https://vmware.github.io/clarity/documentation/v0.11/dropdowns

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

PR #2248, opened 2018-05-08
[UI] Dropdown Nested Menu Arrow Alignment Fix

adityarb88:
Fixes: #2219 

Before:
image

After:
image


src/clr-angular/popover/dropdown/_dropdown.clarity.scss
```diff
@@ -130,7 +130,7 @@
             }
 
             &.expandable {
-                $clr-dropdown-caret-xOffset: 0.333rem;
+                $clr-dropdown-caret-margin-top: 0.375rem;
```
hippee-lee: Can we move this to `./_variables.dropdown.scss` ?

adityarb88: I don't think we can because its a local variable. It changes in different breakpoints. I would have to create multiple variables if we want to do that. For now, I think this is fine. The whole dropdown css needs a refactor and we can get this fixed that time. Does that sound okay?

PR #2249, opened 2018-05-09
[ICONS] addition of Vol.11 icons

Shijir:
**KS app with vol11 icons:** 
http://vol11-icons.surge.sh/iconography
### Text Edit Shapes
- bold
- bullet-list
- checkbox-list
- number-list
- font-size
- italic
- justify-text
- center-text
- align-left-text
- align-right-text
- paint-roller
- block-quote
- text
- underline
- align-center
- align-left
- align-right
- align-bottom
- align-middle
- align-top

### Commercial Shapes
- peso

### Essential Shapes
- cursor-move


src/clr-icons/clr-icons.spec.ts
```diff
@@ -5,32 +5,49 @@
  */
 import {itIgnore} from "../../tests/tests.helpers";
 
+import {ClarityIcons} from ".";
```
adityarb88: Why has the import changed from `./index` to `.`

Shijir: I used VScode plugin that organizes imports and it also turned `"./index"` to `"."`. But I can revert it back.

Issue #2256, opened 2018-05-10
v11 clr-datagrid height not calculated correctly on IE


sstrobel99:
**Select one ...**  (check one with "x")
```
[X ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
clr-datagrid (angular) height computation should be correct for all browsers including IE

### Actual behavior
Using Clarity v10, this seemed to display correctly, however when upgrading to v11 the effective height seems to be calculated incorrectly on IE.

### Reproduction of behavior
I've create two reproductions to demonstrate this problem:  one that uses v10 and works fine; another that uses v11 and demonstrates the problem.  See https://stackblitz.com/edit/clarity-light-theme-v11-hsebgg for the v11 version.  That plunk contains a link to the working v10 plunk.  Pull both of them up on IE11 and Chrome.  You should see that IE11 and Chrome show correct results for the v10 version, but only Chrome shows correctly for the v11 version; the v11 version using IE should demonstrate the problem.
I'm attaching a screen shot showing both IE versions side-by-side.  The v11 problem child is on the left side.  Notice the height and padding-top values.  The v10 version did not have padding-top at all.  The v11 version does.
<img width="1893" alt="clarity_11_dg_ie_bug" src="https://user-images.githubusercontent.com/9386562/39884262-82d4ceaa-5446-11e8-9bf3-2db20e508cc4.PNG">
The suspected area of code is in clr-angular/data/datagrid/main-renderer.ts, where it is attempting to compute the grid height.  This happens in ngAfterViewChecked() life-cycle callback, using the computedHeight(element) function in render/dom-adapter.ts (which ultimately calls getComputedStyle(element).getPropertyValue("height")).
The problem seems to be related to the possiblity that IE's implementation of getComputedStyle(element).getPropertyValue("height") subtracts out the value for "padding-top".
I think in the demo/repro, the ngAfterViewChecked() life-cycle method is being called 5 times (once for each row), so my theory is that for each of the 5 cylces, IE is returning 12px less than it should, and clarity is setting the new height based on this each time, resulting in 5*12px = 60px.  At the end, it should be 282px, but is ending up as 222px.  In our production use, I see ngAfterViewChecked() being called many more times, ultimately resulting in zero height and no rows visible.

### Environment details

* **Angular version:** 5.x.x

* **Clarity version:** 0.11.16

* **OS and version:** Windows 10

* **Browser:** IE 11

PR #2257, opened 2018-05-10
[NG] v11 clr-datagrid height not calculated correctly on IE

sstrobel99:
Use getBoundingClientRect().height to determine clr-datagrid height rather than getComputedStyle(element).getPropertyValue("height"), since IE11 returns the latter incorrectly.

Refer to issue #2256.

Signed-off-by: Your Name <sstrobel99@gmail.com>


sstrobel99:
Will redo with correct signature

PR #2261, opened 2018-05-11
[CORE] Additional coding guidelines

youdz:
- Do not name your methods `onClick()`, `onHover()`, etc.
- Use only the component you are currently testing.

Signed-off-by: Eudes Petonnet-Vincent <epetonnetvince@vmware.com>


youdz:
Travis :trollface:ing again. I don't even have a local version, I directly edited the guidelines on GitHub. 😂

gnomeontherun:
This is because it uses a local branch, and so it runs the tests on SauceLabs. There appears to be an issue with Firefox that has to be resolved for this to turn green.

PR #2271, opened 2018-05-14
[CORE] Fix master build

adityarb88:
Master build was failing because of an extra comma in webpack.icons.config.js.

After the fix, `npm run build` also generated the new `clr-angular/package.json` file.

### DIFF
--- a/src/clr-angular/package.json
+++ b/src/clr-angular/package.json
@@ -2,28 +2,28 @@
     "name": "@clr/angular",
     "version": "0.11.14",
     "description": "Angular components for Clarity",
-    "es2015": "clr-angular/esm2015/clr-angular.js",
+    "es2015": "esm2015/clr-angular.js",
     "homepage": "https://vmware.github.io/clarity/",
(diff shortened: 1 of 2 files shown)


src/clr-angular/package.json
```diff
@@ -2,28 +2,28 @@
...
+        "@angular/forms": "^6.0.0",
+        "@angular/platform-browser": "^6.0.0",
+        "@angular/platform-browser-dynamic": "^6.0.0",
+        "@angular/router": "^6.0.0",
         "@clr/ui": "0.11.14"
     },
     "dependencies": {
-        "@angular-devkit/core": "0.4.7",
-        "@angular-devkit/schematics": "0.4.7"
+        "@angular-devkit/core": "^0.5.12",
```
hippee-lee: @gnomeontherun - do we want to bump these to 6 here or in a seperate PR?

gnomeontherun: I'm actually hoping we can remove these, just need to do a little more work on this. I talked with Hans about it and we should not have to declare these if we check a few things.

PR #2274, opened 2018-05-14
[CORE] have textarea tests dynamically render size of textarea

gnomeontherun:
Something changed in a recent firefox release it seems where heights of the textarea may differ. This relaxes the test to calculate the heights of the textarea control pieces based on the height of the textarea (just like how multiselect cannot be expected to be a same height across browsers).


src/clr-angular/forms/styles/containers.spec.ts
```diff
@@ -281,11 +281,15 @@ describe("Form layouts", () => {
         });
 
         describe("textarea", () => {
+            let textarea;
+            beforeEach(() => {
+                textarea = height("#textarea textarea");
```
jeeyun: would make it slightly easier to follow to name this `textarea_height`. not a big deal though.

PR #2277, opened 2018-05-15
[NG] clrDragHandle directive

Shijir:
API for users will look like:
```html
<div class="card" clrDraggable>
     ...
    <button clrDragHandle><clr-icon shape="folder"></clr-icon></button>
    ...
</div>
```

Beside `ClrDragEventListener` now, `ClrDragHandleRegistrar` is also provided into `ClrDraggable`. So `ClrDragHandleRegistrar` has access to `ClrDragEventListener`. Now `ClrDraggable` itself doesn't directictly call the `.attachListeners()` method of `ClrDragEventListener` on its element. It provides for its element to `ClrDragHandleRegistrar`, which configures which element to be attached to the listeners.

### DIFF
--- a/src/clr-angular/utils/drag-and-drop/draggable.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/draggable.spec.ts
@@ -24,8 +24,8 @@ export default function(): void {
                 this.fixture = TestBed.createComponent(NestedHandleTest);
                 this.testComponent = this.fixture.componentInstance;
                 this.draggable = this.fixture.debugElement.query(By.directive(ClrDraggable));
-                this.dragEvenListener = this.draggable.injector.get(ClrDragEventListener);
-                this.dragDragRegistrar = this.draggable.injector.get(ClrDragHandleRegistrar);
+                this.dragEventListener = this.draggable.injector.get(ClrDragEventListener);
+                this.dragHandleRegistrar = this.draggable.injector.get(ClrDragHandleRegistrar);
             });
 
             afterEach(function() {
@@ -36,9 +36,9 @@ export default function(): void {
                 this.fixture.detectChanges();
 
                 expect(this.draggable.nativeElement.classList.contains("drag-handle")).toBeTruthy();
-                expect(this.dragEvenListener.draggableEl).toBe(this.draggable.nativeElement);
-                expect(this.dragDragRegistrar._defaultHandleEl).toBe(this.draggable.nativeElement);
-                expect(this.dragDragRegistrar.customHandle).toBeUndefined();
+                expect(this.dragEventListener.draggableEl).toBe(this.draggable.nativeElement);
+                expect(this.dragHandleRegistrar.defaultHandleEl).toBe(this.draggable.nativeElement);
+                expect(this.dragHandleRegistrar.customHandleEl).toBeUndefined();
             });
 
             it("should have its nested handle as drag handle if it is present", function() {
@@ -47,9 +47,9 @@ export default function(): void {
                 this.dragHandle = this.fixture.debugElement.query(By.directive(ClrDragHandle));
                 expect(this.draggable.nativeElement.classList.contains("drag-handle")).toBeFalsy();
                 expect(this.dragHandle.nativeElement.classList.contains("drag-handle")).toBeTruthy();
-                expect(this.dragEvenListener.draggableEl).toBe(this.dragHandle.nativeElement);
-                expect(this.dragDragRegistrar.customHandle).toBe(this.dragHandle.nativeElement);
-                expect(this.dragDragRegistrar._defaultHandleEl)
+                expect(this.dragEventListener.draggableEl).toBe(this.dragHandle.nativeElement);
+                expect(this.dragHandleRegistrar.customHandleEl).toBe(this.dragHandle.nativeElement);
+                expect(this.dragHandleRegistrar.defaultHandleEl)
                     .toBe(
                         this.draggable.nativeElement,
                         `The default handle should be still set to the draggable element even though it's not made an active drag handle.`);
--- a/src/clr-angular/utils/drag-and-drop/draggable.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/draggable.spec.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component} from "@angular/core";
+import {TestBed} from "@angular/core/testing";
+import {By} from "@angular/platform-browser";
+import {ClrDragHandle} from "./drag-handle";
+import {ClrDraggable} from "./draggable";
+import {ClrDragAndDropEventBus} from "./providers/drag-and-drop-event-bus";
+import {ClrDragEventListener} from "./providers/drag-event-listener";
+import {ClrDragHandleRegistrar} from "./providers/drag-handle-registrar";
+
+export default function(): void {
+    describe("ClrDraggable", function() {
+        describe("ClrDragHandle", function() {
+            beforeEach(function() {
+                TestBed.configureTestingModule({
+                    declarations: [NestedHandleTest, ClrDraggable, ClrDragHandle],
+                    providers: [ClrDragAndDropEventBus]
+                });
+
+                this.fixture = TestBed.createComponent(NestedHandleTest);
+                this.testComponent = this.fixture.componentInstance;
+                this.draggable = this.fixture.debugElement.query(By.directive(ClrDraggable));
+                this.dragEvenListener = this.draggable.injector.get(ClrDragEventListener);
+                this.dragDragRegistrar = this.draggable.injector.get(ClrDragHandleRegistrar);
+            });
+
+            afterEach(function() {
+                this.fixture.destroy();
+            });
+
+            it("should have its own element as default drag handle when there is no nested drag handle", function() {
+                this.fixture.detectChanges();
+
+                expect(this.draggable.nativeElement.classList.contains("drag-handle")).toBeTruthy();
+                expect(this.dragEvenListener.draggableEl).toBe(this.draggable.nativeElement);
+                expect(this.dragDragRegistrar._defaultHandleEl).toBe(this.draggable.nativeElement);
+                expect(this.dragDragRegistrar.customHandle).toBeUndefined();
+            });
+
+            it("should have its nested handle as drag handle if it is present", function() {
+                this.testComponent.display = true;
+                this.fixture.detectChanges();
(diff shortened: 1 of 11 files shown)


src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+            testComponent = fixture.componentInstance;
+            dragHandleRegistrar = fixture.debugElement.injector.get(ClrDragHandleRegistrar);
+            dragEventListener = fixture.debugElement.injector.get(ClrDragEventListener);
+        });
+
+        afterEach(() => {
+            fixture.destroy();
+        });
+
+        it("should be able to register element as default handle on assignment", function() {
```
youdz: Nitpick: "should be able to" is redundant with the fact that this is a spec. You can even see that by just looking at the other unit tests: every single one starts with "should be able to", because that's the entire point of a test. 😛 

Just `it("registers element as default handle on assignment", ...)`.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    public registerCustomHandle(handleElement: Node) {
+        this.dragEventListener.detachDragListeners();  // removes the listeners only when there are existing ones
+        this._customHandleEl = handleElement;
+        this.makeElementHandle(this._customHandleEl);
+    }
+
+    public unregisterCustomHandle() {
+        delete this._customHandleEl;
+        // if default handle is set, make that handle
```
youdz: You're missing the `removeClass()` and `detachDragListeners()`.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+            expect(customHandleEl.hasListener).toBeTruthy();
+            expect(customHandleEl.classList.contains("drag-handle")).toBeTruthy();
+        });
+
+        it("should be able to unregister custom handle", function() {
+            dragHandleRegistrar.registerCustomHandle(customHandleEl);
+            dragHandleRegistrar.unregisterCustomHandle();
+
+            expect(dragHandleRegistrar.customHandle).toBeUndefined();
```
youdz: You're missing a couple `expect()` (see comment on `ClrDragHandleRegistrar`):
- Listeners should be gone.
- CSS class should be gone.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+               expect(dragHandleRegistrar.customHandle).toBeUndefined();
+               expect(dragEventListener.draggableEl).toBe(draggableEl);
+               expect(draggableEl.hasListener).toBeTruthy();
+               expect(draggableEl.classList.contains("drag-handle")).toBeTruthy();
+           });
+
+        it("should be able to unregister custom handle and fall back to default handle if default handle is set after custom handle",
+           function() {
+               dragHandleRegistrar.registerCustomHandle(customHandleEl);
+               dragHandleRegistrar.defaultHandleEl = draggableEl;
```
youdz: I can see one missing test here. If you have:
```typescript
dragHandleRegistrar.registerCustomHandle(customHandleEl);
dragHandleRegistrar.defaultHandleEl = draggableEl;
```
You never check that the default handle doesn't override the listeners. The implementation has an `if` which is correct, but you don't have the corresponding unit test.

src/typings.d.ts
```diff
@@ -16,3 +16,8 @@ interface Window {
 interface ClarityIconElement extends HTMLElement {
     clrIconUniqId: string;
 }
+
+
+interface Node {
+    hasListener: boolean;
```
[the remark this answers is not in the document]
Shijir: Moved this to the mock provider.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
Shijir: Updated the messages.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.ts
Shijir: Added in and called the methods.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
Shijir: Added in the corresponding `expect()`s.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.spec.ts
Shijir: Indeed, I missed that test. Thanks! I added the test in now.

Shijir:
@youdz I have added in the relevant part of the `clrDraggable` in this PR and the tests for the part.

youdz:
Ugh, you amended your previous commit and rewrote the history on this PR, so checking for the recent changes is going to be a mess. :(

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            beforeEach(function() {
+                TestBed.configureTestingModule({
+                    declarations: [NestedHandleTest, ClrDraggable, ClrDragHandle],
+                    providers: [ClrDragAndDropEventBus]
+                });
+
+                this.fixture = TestBed.createComponent(NestedHandleTest);
+                this.testComponent = this.fixture.componentInstance;
+                this.draggable = this.fixture.debugElement.query(By.directive(ClrDraggable));
+                this.dragEvenListener = this.draggable.injector.get(ClrDragEventListener);
```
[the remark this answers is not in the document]
Shijir: Will fix this.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+                this.dragDragRegistrar = this.draggable.injector.get(ClrDragHandleRegistrar);
```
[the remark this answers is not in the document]
Shijir: Two drags... just in case 😛. Will fix this.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            afterEach(function() {
+                this.fixture.destroy();
+            });
+
+            it("should have its own element as default drag handle when there is no nested drag handle", function() {
+                this.fixture.detectChanges();
+
+                expect(this.draggable.nativeElement.classList.contains("drag-handle")).toBeTruthy();
+                expect(this.dragEvenListener.draggableEl).toBe(this.draggable.nativeElement);
+                expect(this.dragDragRegistrar._defaultHandleEl).toBe(this.draggable.nativeElement);
```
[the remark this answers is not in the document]
Shijir: I might expose it through `get defaultHandleEl() {}`?

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -3,24 +3,27 @@
...
 
-    constructor(private el: ElementRef, private dragEventListener: ClrDragEventListener<T>) {
+    constructor(private el: ElementRef, private dragEventListener: ClrDragEventListener<T>,
+                private dragHandleRegistrar: ClrDragHandleRegistrar<T>) {
         this.draggableEl = this.el.nativeElement;
     }
 
-    ngOnInit() {
-        this.dragEventListener.attachDragListeners(this.draggableEl);
+    ngAfterContentInit() {
```
adityarb88: Why delay from `ngOnInit` to `ngAfterContentInit`?

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.ts
```diff
@@ -0,0 +1,63 @@
...
+@Injectable()
+export class ClrDragHandleRegistrar<T> {
+    private _customHandleEl: Node;
+    private _defaultHandleEl: Node;
+
+    get defaultHandleEl() {
+        return this._defaultHandleEl;
+    }
+
+    set defaultHandleEl(value: Node) {
```
adityarb88: You are using three different naming conventions for the parameters: 
`set defaultHandleEl(value: Node)`
`private makeElementHandle(el: Node)`
`public registerCustomHandle(handleElement: Node)`

Lets just choose one? :D

Shijir: Will address this issue.

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: During `ngOnInit`, we don't know if there will be a nested `clrDragHandle` or not. So this is what happens there:
1. Register draggable element as the default drag handle and make it the active drag handle anyways.
2. if `clrDragHandle` is registered sometime later, override the current active drag handle and make the `clrDragHandle`'s element the active drag handle.

But during `ngOnAfterInit`, we know whether we have nested elements or not. And if one of those nested elements happen to be `clrDragHandle` element, make that the active drag handle. If there is no nested `clrDragHandle`, just make the `clrDraggable` element the active drag handle. So basically, using `ngOnAfterInit` prevents `clrDraggable` element from being registered too early and overriden later.

src/clr-angular/utils/drag-and-drop/providers/drag-handle-registrar.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import {ClrDragEventListener} from "./drag-event-listener";
+
+// This provider registers the drag handle element.
+// When it registers a element as a drag handle, it attaches that element to the listeners from ClrDragEventListener.
+// Also, it adds the "drag-handle" css class to the registered element through Renderer.
+@Injectable()
+export class ClrDragHandleRegistrar<T> {
+    private _customHandleEl: Node;
+    private _defaultHandleEl: Node;
```
jeeyun: I thought we were moving away from underscored private variables? :)

Shijir: I do avoid underscores for private properties. But when It's necessary to use setter/getter for a private property, I use an underscore to make it distinct.

jeeyun: Oh that makes sense. Thanks for clarifying! 👍

Issue #2151, opened 2018-04-05
Deprecate wizard ghost pages


gnomeontherun:
This is an undocumented feature and adds code we don't want to maintain.

* [ ] Mark as deprecated in code
* [ ] Verify no docs exist

PR #2279, opened 2018-05-15
[core] deprecate wizard ghost pages (#2151)

jeeyun:

### DIFF
--- a/src/clr-angular/modal/modal.html
+++ b/src/clr-angular/modal/modal.html
@@ -31,6 +31,7 @@
                     <ng-content select=".modal-footer"></ng-content>
                 </div>
             </div>
+            <!--todo: deprecate the modal-ghost-wrapper div below after 0.12-->
             <div class="modal-ghost-wrapper">
                 <div [@ghostPageOneState]="ghostPageState" class="modal-ghost modal-ghost-1"></div>
                 <div [@ghostPageTwoState]="ghostPageState" class="modal-ghost modal-ghost-2"></div>
--- a/src/clr-angular/modal/utils/ghost-page-animations.ts
+++ b/src/clr-angular/modal/utils/ghost-page-animations.ts
@@ -3,6 +3,10 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+
+/*
+ * @deprecated since 0.12
+ */
 export const GHOST_PAGE_ANIMATION = {
     STATES: {NO_PAGES: "inactive", ALL_PAGES: "ready", NEXT_TO_LAST_PAGE: "penultimateGhost", LAST_PAGE: "lastGhost"},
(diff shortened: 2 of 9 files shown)


src/clr-angular/modal/utils/ghost-page-animations.ts
```diff
@@ -3,7 +3,11 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-export const GHOST_PAGE_ANIMATION = {
+
+/*
+ * @deprecated since 0.12
+ */
+export const GHOST_PAGE_ANIMATION_DEPRECATED = {
```
gnomeontherun: No reason to change this name, its internal and should just have a deprecated comment.

src/clr-angular/modal/modal.html
```diff
@@ -32,8 +32,8 @@
                 </div>
             </div>
             <div class="modal-ghost-wrapper">
-                <div [@ghostPageOneState]="ghostPageState" class="modal-ghost modal-ghost-1"></div>
```
gnomeontherun: Let's put an HTML comment in here to mark this deprecation.

src/clr-angular/modal/modal.ts
```diff
@@ -88,7 +90,7 @@ export class ClrModal implements OnChanges, OnDestroy {
     @Input("clrModalSkipAnimation") skipAnimation: string = "false";
 
     // presently this is only used by wizards
-    @Input("clrModalGhostPageState") ghostPageState: string = "hidden";
+    @Input("clrModalGhostPageState") ghostPageStateDeprecated: string = "hidden";
```
gnomeontherun: I think all the changes in this file can be reverted to original names.

src/clr-angular/wizard/providers/wizard-navigation.service.ts
```diff
@@ -737,45 +737,51 @@ export class WizardNavigationService implements OnDestroy {
 
     /**
      * @memberof WizardNavigationService
+     * @deprecated since 0.12
      */
-    private _wizardGhostPageState: string = GHOST_PAGE_ANIMATION.STATES.NO_PAGES;
+    private _wizardGhostPageStateDeprecated: string = GHOST_PAGE_ANIMATION_DEPRECATED.STATES.NO_PAGES;
```
gnomeontherun: Keep the deprecation comments, revert the names throughout since they are private.

src/clr-angular/modal/modal.html
jeeyun: 👍

gnomeontherun:
@jeeyun I realized it would be good to add a quick test to the deprecated spec file that tests the presence of ghost pages, so we can keep track of this.

jeeyun:
@gnomeontherun took a look and there wasn't anything we exported to `public_api` for ghost pages. Other wizard directives have been deprecated for longer, and are already in the spec.

Issue #2125, opened 2018-04-02
Deprecate code highlight


gnomeontherun:
We need to deprecate the codehighlight component in the code and website.

PR #2281, opened 2018-05-15
[core] deprecate code highlight (#2125)

jeeyun:


adityarb88:
@jeeyun: How does this change affect the website?

gnomeontherun:
@jeeyun I realized it would be good to add a quick test to the deprecated spec file that tests the presence of code highlight, so we can keep track of this.

Issue #2263, opened 2018-05-11
Date Picker don't mark form as dirty


williangd:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When you select a date in the date picker, the form is marked as dirty.

### Actual behavior
When you select a date in the date picker, the form is not marked as dirty.

### Reproduction of behavior
Click the date picker icon to open the component and select a date.
The form is marked as dirty only when typing the date with the keyboard
https://stackblitz.com/edit/clarity-light-theme-v11-6quune

### Environment details

* **Angular version:** 5.2.1

* **Clarity version:** 0.11.16

* **OS and version:**

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@williangd: Thanks for reporting this issue. We will take a look and get back to you.

gnomeontherun:
Yup, that should do it. Same with `markAsTouched()`

PR #2284, opened 2018-05-16
[NG] Fix Date Picker State

adityarb88:
Fixes: #2263 

**Changes:**
1. Date Control marked as `touched` when the datepicker toggle is clicked.
2. Date Control marked as `dirty` when a date is selected from the popover.

**Tested on Chrome**

### DIFF
--- a/src/clr-angular/forms/datepicker/date-container.spec.ts
+++ b/src/clr-angular/forms/datepicker/date-container.spec.ts
@@ -14,6 +14,7 @@ import {IfOpenService} from "../../utils/conditional/if-open.service";
 import {FormControlService} from "../common/form-control.service";
 
 import {ClrDateContainer} from "./date-container";
+import {DateFormControlService} from "./providers/date-form-control.service";
 import {DateIOService} from "./providers/date-io.service";
 import {DateNavigationService} from "./providers/date-navigation.service";
 import {DatepickerEnabledService} from "./providers/datepicker-enabled.service";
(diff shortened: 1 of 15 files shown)


src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -94,6 +98,16 @@ export default function() {
 
                 sub.unsubscribe();
             });
+
+            // IE doesn't support MouseEvent constructors
+            itIgnore(["ie"], "marks the date control as touched when the datepicker popover is toggled", () => {
+                spyOn(dateFormControlService, "markAsTouched");
+
+                const fakeEvent: MouseEvent = new MouseEvent("fakeEvent");
+                context.clarityDirective.toggleDatepicker(fakeEvent);
```
youdz: I don't believe you need the MouseEvent. You could literally pass `null` and the test would still work.

src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -94,6 +98,15 @@ export default function() {
 
                 sub.unsubscribe();
             });
+
+            // IE doesn't support MouseEvent constructors
+            itIgnore(["ie"], "marks the date control as touched when the datepicker popover is toggled", () => {
```
gnomeontherun: This doesn't need to be wrapped so it skips IE.

adityarb88: Yep. Initially I was create a `MouseEvent` here and thats why I added the `itIgnore`. But then later @youdz correctly pointed out that we didn't have to create the `MouseEvent` for this test. Will remove the `itIgnore`.

PR #2291, opened 2018-05-17
[UI, NG] Select Option Component

adityarb88:
PR Scope:

1. Create `clr-option` component.
2. Style `clr-option` component.
3. Tests

Tested on Chrome

### DIFF
--- a/src/clr-angular/forms/select/option.spec.ts
+++ b/src/clr-angular/forms/select/option.spec.ts
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Component, ElementRef} from "@angular/core";
+import {TestBed} from "@angular/core/testing";
+
+import {TestContext} from "../../data/datagrid/helpers.spec";
+import {POPOVER_HOST_ANCHOR} from "../../popover/common/popover-host-anchor.token";
+import {IfOpenService} from "../../utils/conditional/if-open.service";
+
+import {ClrOption} from "./option";
+
+@Component({
+    template: `
+        <clr-option>
+            Test
+        </clr-option>
+    `,
+    providers: [IfOpenService, {provide: POPOVER_HOST_ANCHOR, useExisting: ElementRef}]
+})
+class TestComponent {
+}
+
+@Component({
+    template: `
+        <clr-option>
+            Test
+        </clr-option>
+    `,
+    providers: [IfOpenService]
+})
+class TestComponentWithError {
+}
+
+export default function(): void {
+    describe("Select Option Component", function() {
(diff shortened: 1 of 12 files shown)


src/clr-angular/forms/select/option.ts
```diff
@@ -0,0 +1,22 @@
...
+@Component({selector: "clr-option", templateUrl: "option.html", host: {"[class.clr-option]": "true"}})
+export class ClrOption {
+    constructor(private ifOpenService: IfOpenService, @Optional() @Inject(POPOVER_HOST_ANCHOR) parentHost: ElementRef) {
+        if (!parentHost) {
+            throw new Error("clr-option should only be used inside of a clr-select");
+        }
+    }
+
+    @HostListener("click")
+    closeMenuOnClick() {
```
youdz: Can you just add a short comment saying this "close on click" behavior should _not_ be kept for multi-selects? I'm just afraid we might forget. Thanks.

src/clr-angular/forms/select/option.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+class TestComponent {
+}
+
+export default function(): void {
+    describe("Select Option Component", function() {
+        let context: TestContext<ClrOption, TestComponent>;
+        let ifOpenService: IfOpenService;
+
+        beforeEach(function() {
+            context = this.create(ClrOption, TestComponent, [IfOpenService]);
```
Shijir: Do we need `IfOpenService` provided here again? I see it's provided at the component level above?

src/clr-angular/forms/select/option.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+import {Component, ElementRef, HostListener, Inject, Optional} from "@angular/core";
+import {POPOVER_HOST_ANCHOR} from "../../popover/common/popover-host-anchor.token";
+import {IfOpenService} from "../../utils/conditional/if-open.service";
+
+@Component({selector: "clr-option", templateUrl: "option.html", host: {"[class.clr-option]": "true"}})
+export class ClrOption {
+    constructor(private ifOpenService: IfOpenService, @Optional() @Inject(POPOVER_HOST_ANCHOR) parentHost: ElementRef) {
+        if (!parentHost) {
+            throw new Error("clr-option should only be used inside of a clr-select");
```
Shijir: I usually test if the expected error is being thrown or not if the `parentHost` is missing.

src/clr-angular/forms/select/option.spec.ts
adityarb88: Dont think we do. Will remove it. Thanks!

src/clr-angular/forms/select/option.ts
adityarb88: 😠. (:trollface: will test that.)

src/clr-angular/forms/select/option.ts
adityarb88: Yep. Forgot to note that.

Issue #1024, opened 2017-06-08
[NG] Create Angular Form Field Components


adityarb88:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Description:
Create the following Angular components for forms:

- [x] Checkbox - #319 
- [ ] Radio Button - #373
- [ ] Input Fields - #835
- [ ] Select - #248
- [ ] Textarea - #1023
- [x] Datepicker - #474

### Actual behavior

Form Field components do no exist

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [All]

jacobbutton:
Is there a time frame on these yet? They are sorely needed.  If not can we move them up the priority list possibly?

mathisscott:
Hi @jacobbutton 
Before we begin implementation, we need this epic to go through a full UX pass. So this issue is still a ways off, but it is a priority for us.

gnomeontherun:
All of these components have been released.

PR #2295, opened 2018-05-17
[NG] clrInput and form control validation components

gnomeontherun:
* Renamed FormControlService and WrappedFormControl
* Added ClrInput, ClrInputContainer for tracking input control states
* Added ClrControlHelper and ClrControlError for validation
* Added ClrIfError directive to conditionally display error messages

Links to #1878 #1024


src/clr-angular/forms/common/providers/control.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    // When the control is set, we want to subscribe to statusChanges and forward them to
+    // our observable when the control is invalid (dirty,touched)
+    private subscribe() {
+        if (this.subscription) {
+            this.subscription.unsubscribe();
+        }
+
+        if (this._control.statusChanges) {
+            this.subscription = this._control.statusChanges.subscribe(status => {
+                if (this._control.dirty && this._control.touched) {
```
youdz: I thought we said touched was enough for required fields to throw an error? Based on what I'm reading here, if I focus in then out of a required input, I do not get the error message.

I think this should be `if (this._control.dirty || this._control.touched)`, but maybe I missed something.

gnomeontherun: It doesn't really matter if the control is dirty or not, if it is touched is when we care to check it. This makes sense when a form is new or when the data loaded is valid, but might not make sense if the control could preload invalid data (not sure when this could be the case, but perhaps it could be, perhaps when a form is autosaved even if its incomplete). 

Perhaps this is something that can/should be configurable?

youdz: Invalid pre-loaded data has to come from the user anyway, right? So without Clarity, it seems logical to me that the way to do this in Angular is to load the previously entered user data on initialization, and to mark the corresponding fields as dirty since they were "entered" by the user. 
So I think Clarity would just do the same, we don't have to provide anything, they just get hold the the FormControls (which isn't even extra work with reactive forms) and mark the ones the pre-filled with user data as dirty.

src/clr-angular/forms/input/input-container.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+            inputContainer = inputContainerDE.componentInstance;
+            inputContainerEl = inputContainerDE.nativeElement;
+        });
+
+        it("injects the ControlService and subscribes in AfterContentInit", () => {
+            expect(inputContainerDE.injector.get(ControlService, null)).toBeTruthy();
+            expect(inputContainer.subscription).toBeTruthy();
+        });
+
+        it("projects the label", () => {
```
youdz: This test is `"projects the label as the first child"`, since you have that second `expect()`.

src/clr-angular/forms/input/input-container.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+            const label = inputContainerEl.querySelector("label");
+            expect(label).toBeTruthy();
+            expect(label.previousElementSibling).toBeFalsy();
+        });
+
+        it("projects the input", () => {
+            expect(inputContainerEl.querySelector("[clrInput]")).toBeTruthy();
+        });
+
+        it("projects the helper text", () => {
```
youdz: I would make this test slightly more complicated to test `it("projects the helper text when the input isn't invalid")`. Double negation, I know, but "when then input is valid" doesn't describe it correctly.

src/clr-angular/forms/input/input-container.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+        });
+
+        it("tracks the validity of the form control", () => {
+            const input = fixture.debugElement.query(By.directive(ClrInput)).nativeElement;
+            expect(inputContainer.invalid).toBeFalse();
+            // inputs must be both invalid and blurred to register the validity
+            input.value = "abc";
+            input.dispatchEvent(new Event("input"));
+            input.dispatchEvent(new Event("blur"));
+            fixture.detectChanges();
```
youdz: You're re-testing the blur binding of the input here. You should directly set the value of the FormControl (and make it dirty if needed) rather than "type" into the input.

src/clr-angular/forms/common/if-error.ts
```diff
@@ -0,0 +1,52 @@
...
+                "clrIfError can only be used within a form control container element like clr-input-container");
+        } else {
+            this.displayError(false);
+        }
+    }
+
+    @Input("clrIfError") error: string;
+
+    private subscription: Subscription;
+    private state: boolean = false;
```
youdz: Nitpick: after reading the code, I think `state` is too generic of a name for this property. I'd rather have `displayed`, or `currentError`, or anything that indicates this property tracks whether we are already displaying the error at the moment or not.
As I said it's a nitpick, nothing I feel strongly about or that I consider critical.

src/clr-angular/forms/common/if-error.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+const errorMessage = "ERROR_MESSAGE";
+
+@Component({template: `<div *clrIfError></div>`})
+class InvalidUseTest {
+}
+
+@Component({
+    template: `
+    <clr-input-container>
+        <input type="text" name="model" clrInput [(ngModel)]="model" required />
```
youdz: As mentioned before, these tests use too many components at the same time, they're not isolated at all. I'd much rather you just declared the `ControlService` as a provider and that you instantiated a `new FormControl()` for this spec. 
FormControls are quite simple to instantiate, and they offer methods like `setErrors()` for you to have full control of them to test your directive.

src/clr-angular/forms/common/if-error.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+                    imports: [ClrIconModule, FormsModule],
+                    declarations: [ClrInput, ClrControlError, ClrInputContainer, ClrIfError, SpecificErrorTest]
+                });
+                fixture = TestBed.createComponent(SpecificErrorTest);
+                fixture.detectChanges();
+                container = fixture.debugElement.query(By.directive(ClrInputContainer));
+                input = container.query(By.directive(ClrInput));
+            });
+
+            it("displays the message when control has an error", () => {
```
youdz: "displays the message when control has **the specified** error"

src/dev/src/app/forms/template-driven/template-driven.html
```diff
@@ -0,0 +1,26 @@
+<!--
+  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h2>Text Input</h2>
+
+<form class="clr-form" #form="ngForm" (submit)="submit()">
```
adityarb88: Just to be safe, can we test this for Reactive Forms too?

src/clr-angular/forms/common/providers/ng-control.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    private _control: NgControl;
+    get control(): NgControl {
+        return this._control;
+    }
+
+    set control(control: NgControl) {
+        this._control = control;
+        this._controlChanges.next(control);
```
adityarb88: Why do we need a setter and a getter when the `Subject` emits the `NgControl`?

Could we not have just used a method which registers the `NgControl` and all its subscribers know about it because the `Subject` emits the control on registration?

src/clr-angular/forms/common/providers/ng-control.service.ts
```diff
@@ -0,0 +1,28 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Injectable} from "@angular/core";
+import {NgControl} from "@angular/forms";
+import {Observable, Subject} from "rxjs";
+
+@Injectable()
+export class NgControlService {
```
adityarb88: Would it help to rename this service to `NgControlRegistrationService`?

src/clr-angular/forms/common/label.ts
```diff
@@ -3,22 +3,30 @@
...
 import {Subscription} from "rxjs";
 
-import {FormControlService} from "./form-control.service";
+import {IfErrorService} from "./if-error/if-error.service";
+import {ControlIdService} from "./providers/control-id.service";
 
 @Directive({selector: "label"})
 export class ClrLabel implements OnInit, OnDestroy {
-    constructor(@Optional() private formControlService: FormControlService) {}
+    labelClass: boolean;
```
adityarb88: Are you using this anywhere?

src/clr-angular/forms/common/label.ts
```diff
@@ -3,22 +3,30 @@
...
+    constructor(@Optional() private controlIdService: ControlIdService,
+                @Optional() private controlService: IfErrorService, private renderer: Renderer2,
```
adityarb88: Did you want to use the `NgControlService` here instead of the `IfErrorService`? I know it doesn't matter because the `input-container` declares both the services but the variable name you have chosen confused me.

src/clr-angular/forms/common/if-error/if-error.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import {Injectable, OnDestroy} from "@angular/core";
+import {NgControl} from "@angular/forms";
+import {Observable, Subject, Subscription} from "rxjs";
+import {filter} from "rxjs/operators";
+
+import {NgControlService} from "../providers/ng-control.service";
+
+@Injectable()
+export class IfErrorService implements OnDestroy {
+    // Implement our own since it doesn't fire on events like blur, and return the control instead of a string
```
adityarb88: Didn't understand this comment. `since it doesn't fire on events` do you mean the status isn't updated on events like `blur`

src/clr-angular/forms/common/if-error/if-error.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    private listenForChanges() {
+        this.subscriptions.push(
+            this.ngControlService.control.statusChanges.pipe(filter(() => this.ngControlService.control.touched))
+                .subscribe(() => {
+                    this._statusChanges.next(this.ngControlService.control);
+                }));
+    }
+
+    // Allows a control to push a status check upstream, such as on blur
+    checkStatus() {
```
adityarb88: `checkStatus` makes me think that this method would return a `boolean` but instead, it emits the `NgControl`.  May be we should rename this method? May be `updateControlStatus`?

src/clr-angular/forms/common/if-error/if-error.service.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            expect(testControl.statusChanges.subscribe).toHaveBeenCalled();
+        });
+
+        it("provides observable for statusChanges, passing the control", () => {
+            const cb = jasmine.createSpy("cb");
+            const sub = service.statusChanges.subscribe(control => cb(control));
+            ngControlService.control = testControl;
+            // Change the state of the input to trigger statusChange
+            testControl.markAsTouched();
+            testControl.updateValueAndValidity();
```
adityarb88: Would this trigger the `blur` event on the control?

adityarb88: I am not sure why we have to call `updateValueAndValidity` because you are just filtering for the `touch` event:

```
this.ngControlService.control.statusChanges.pipe(filter(() => this.ngControlService.control.touched))
                .subscribe(() => {
                  this._statusChanges.next(this.ngControlService.control);
                }));
```

src/clr-angular/forms/common/if-error/if-error.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    ngOnDestroy() {
+        this.subscription.unsubscribe();
+    }
+
+    private displayError(invalid: boolean) {
+        if (invalid && !this.displayed) {
+            this.container.createEmbeddedView(this.template);
+            this.displayed = true;
```
adityarb88: I see that we update `displayed` but I can't find where we use it?

src/dev/src/app/forms/template-driven/template-driven.html
gnomeontherun: Added new demos.

src/clr-angular/forms/common/if-error/if-error.service.spec.ts
adityarb88: I still haven't understood why we need to call `updateValueAndValidity` but I will ask you that question on a call separately.

src/clr-angular/forms/common/providers/ng-control.service.ts
```diff
@@ -0,0 +1,22 @@
...
```
youdz: Now that I see it implement, this is literally a `Subject`, or `BehaviorSubject` if you add the getter again.
You don't need to change it, and I get that we might want to add properties or methods to it in the future, I just thought it was worth pointing out this is a bit of a silly wrapper.

gnomeontherun: Yeah once I detangled it its pretty light. I do like the idea of having the method do the emit instead of exposing it.

PR #2297, opened 2018-05-18
[NG] Optional Select Menu

adityarb88:
**PR Scope:**

Added the ability to make the options menu optional. If the user doesn't provide the `clr-options` menu, then we create it for them.

Tested on Chrome.


src/clr-angular/forms/select/select.ts
```diff
@@ -18,6 +19,7 @@ import {TAB} from "../../utils/key-codes/key-codes";
 })
 export class ClrSelect implements AfterContentInit {
     @ViewChild("input") input: ElementRef;
+    @ContentChild(ClrOptions) options: ElementRef;
```
youdz: The type here is wrong. Either you want:
```typescript
@ContentChild(ClrOptions) options: ClrOptions;
```
or 
```typescript
@ContentChild(ClrOptions, {read: ElementRef}) options: ElementRef;
```
I believe in your case the first one is simpler and will work fine. But my guess is that the current typing would throw an error when compiled AoT. And if it doesn't, it should. 😛

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -139,5 +162,18 @@ export default function(): void {
...
+        describe("Select with Menu", () => {
+            beforeEach(function() {
+                context = this.create(ClrSelect, TestSelectWithMenu, [IfOpenService]);
+                ifOpenService = context.getClarityProvider(IfOpenService);
+            });
+
+            it("renders the menu projected by the consumer", () => {
+                const menus = context.clarityElement.querySelectorAll("clr-options");
+                expect(menus.length).toBe(1);
+                expect(menus[0].classList.contains("test")).toBe(true);
```
youdz: I like this sneaky class check to make sure we're not cheating. Awesome. 👍

src/clr-angular/forms/select/select.ts
adityarb88: fixed

src/clr-angular/forms/select/select.spec.ts
adityarb88: thanks!

PR #2310, opened 2018-05-22
[web] remove dependency on ClrCodeHighlight

jeeyun:
Only some of the wizard demos imported the exported `ClrCodeHighlight`. Website for the most part has been using its own implementation internally. Removing those imports in preparation for 0.12.


latest/src/app/documentation/demos/buttons/utils/example.ts
```diff
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import {Component, Input, ViewChild, AfterViewInit} from "@angular/core";
-import {CodeHighlight} from "@clr/angular";
+import {CodeHighlight} from "../../../../utils/code-highlight";
```
gnomeontherun: It seems like we have this, which duplicates the same code highlight functionality, and this appears to be used in a few random places. Can we use just the one implementation and remove this?

PR #2316, opened 2018-05-25
[NG] Set isRootLevelToggle to public

krotscheck:
When running a strict configuration of ngc (e.g. ng-packagr) against a library
which imports the ClrDropdownModule (e.g. ng-packagr), ngc errors out because
the dropdown's private property is used in the "host" component. Note that
even though the compiler appears to access clarity, the component isn't actually
bundled in the resulting output.

Signed-off-by: Michael Krotscheck <krotscheck@gmail.com>

### DIFF
--- a/src/clr-angular/popover/dropdown/dropdown-trigger.ts
+++ b/src/clr-angular/popover/dropdown/dropdown-trigger.ts
@@ -20,8 +20,7 @@ import {ClrDropdown} from "./dropdown";
     }
 })
 export class ClrDropdownTrigger {
-    /* tslint:disable-next-line:no-unused-variable */
-    private isRootLevelToggle: boolean = true;
+    public isRootLevelToggle: boolean = true;
 
     constructor(dropdown: ClrDropdown, private ifOpenService: IfOpenService) {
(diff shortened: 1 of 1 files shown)


hippee-lee:
@youdz  The `ks-app` uses `clr-dropdown` in a template and imports all of the pieces publicly declared by the Clarity API: `ClrDropdownTrigger` etc ... So I'm not sure why this wasn't caught. Need to look deeper and test this specific use case. 

We test buildin  the ks-app for production as most users will use clarity in a client app:  `ng build --prod`. Maybe re-packaging clarity with `ng-packgr` does something different with the installed Clarity codebase? I'm not sure with just a quick look at the ks-app. In this case, it seems like we are packaging Clarity and then _re-packaging_ it inside another module.

So to reproduce this I should set up ng-packgr for the `ks-app`  and use `ng-packgr` to publish it as if it is a library to be consumed in an app. Is this correct @krotscheck?

krotscheck:
@hippee-lee Yes, something like that. I've got an internal project if you'd like to take a look, just ping me on slack. From personal experience, it feels that the multi-configuration setup of ng-packagr is a bit more aggressive about its configuration checks.

PR #2324, opened 2018-05-30
[CORE] Adds Prettier.io

### LINKED ISSUE #2156: Replace clang with prettier
Setup prettier and only have it format changed lines.

hippee-lee:
- Removes clang scripts and configuration
- Updates package.json for pre-commit hook to format changed files
- Adds  and  for git hooks w/ npm scripts
-adds a  script to run prettier with the  option (this will be used after all files are formatted for both a pre-push git hook and a ci check on in the  script
- closes #2156


scripts/format-check.js
```diff
@@ -0,0 +1,30 @@
...
+
+const colors = require('colors/safe');
+const path = require('path');
+const prettier = path.join(process.cwd(), 'node_modules', '.bin', 'prettier');
+const shell = require('shelljs');
+const unPrettyFiles = [];
+
+console.log('Formatting Test');
+
+let result = shell.exec(`${prettier} --list-different ./**/*.{js,json,md,ts,scss}`, { silent: true });
```
gnomeontherun: Do we really need to have a script for this? Looks like `--list-different` will error out for you. https://prettier.io/docs/en/cli.html#list-different

scripts/format-check.js
```diff
@@ -0,0 +1,30 @@
...
+result.stdout.split('\n').forEach(file => {
+  unPrettyFiles.push(file);
+});
+
+if (unPrettyFiles.length > 1) {
+  console.log(colors.yellow.underline(`There are ${unPrettyFiles.length - 1} un-formatted files`));
+  unPrettyFiles.forEach(file => {
+    console.log(colors.red(file));
```
adityarb88: Nice! Printing the filenames which need to be formatted helps a lot instead of what we do right now which is just printing the number of files that need formatting.

prettier.config.js
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+module.exports = {
+  printWidth: 120,
+  tabWidth: 2,
+  useTabs: false,
+  semi: true,
+  singleQuote: true,
+  trailingComma: 'es5',
+  bracketSpacing: true,
```
gnomeontherun: I'm being picky here, but if a config file can be JSON that seems more sensible to me.

package.json
```diff
@@ -26,13 +26,13 @@
         "start": "ng serve dev",
         "clean": "del ./dist",
         "test:aot": "ng build --prod ks-app",
-        "test:format": "./scripts/clang-check.sh",
```
gnomeontherun: We should also remove the clang packages from `package.json`.

package.json
```diff
@@ -26,13 +26,13 @@
...
+        "test:format": "node ./scripts/format-check.js",
         "test:lint": "tslint -p . -c tslint.json \"src/**/*.ts\" -e \"src/ks-app/**\" -e \"src/schematics/**\"",
         "test:visual": "./scripts/docker-cdc.js -t",
         "test": "ng test clr-angular",
         "test:watch": "ng test clr-angular --watch",
-        "test:travis": "npm-run-all build:ui test:format test:lint test test:aot",
-        "format:fix": "node_modules/.bin/clang-format -style=file -i --glob=src/**/*.ts",
+        "test:travis": "npm-run-all build:ui test:lint test test:aot",
```
jeeyun: is there an equivalent of `test:format` in the prettier version? that is, make the travis build fail if the code isn't formatted to prettier's standard?

jeeyun: Oh I might have spoken too soon. Is that handled by husky?

scripts/format-check.js
hippee-lee: Right, but it won't give us a list of unformatted files (so I quickly know where to look when a push fails).

prettier.config.js
hippee-lee: I will turn this into a json file.

package.json
hippee-lee: I have to disable it until all of the files are formatted. So its not enabled for this commit. Changed files will be formatted with a `pre-commit` hook. 

And, once this goes in the bulk formatting  per file type will follow within an hour or two. I'll re-enable the `pre-push` hook that checks for correct formatting and add the script back to the test:travis job.

hippee-lee: @jeeyun - fyi husky only enables git hooks so that we can define them in the package.json as `npm scripts` instead of mucking with the `./.git/hooks` files themselves like we did w/ clang checking.

scripts/format-check.js
hippee-lee: Actually, my last comment is incorrect, yes it will give us a list, but I'd still like to keep the script so I can see a count and colors for quickly scanning the cli output on my local. 
\cc @gnomeontherun

package.json
hippee-lee: Will push a change to address this.

.prettierrc
```diff
@@ -0,0 +1,10 @@
+{
+  printWidth: 120,
+  tabWidth: 2,
+  useTabs: false,
+  semi: true,
+  singleQuote: true,
```
Shijir: nice 😄 👍

package.json
hippee-lee: This has been updated per the discussion.

Issue #2300, opened 2018-05-21
Date Picker not outputting dates in Reactive & Template Driven Forms


rameshyala:
[X] Question

Hi , i am set locale to my app in app.module file like ' providers: [{ provide: LOCALE_ID, useValue: 'en-IN' }]', but this ' en-IN' locale need to show format as DDMMYYYY but actually it showing MMDDYYYY format. check once [Stackblitz](https://stackblitz.com/edit/clarity-light-theme-v11-nknd98?file=app%2Fapp.module.ts).

One more doubt is now date picker have change detection option , if i change actual date is emit any event now, i am check clarity git hub code i am found for template-driven-forms have '(clrDateChange)=" "'  option, but in that for reactive forms does not see anything , how i will do that ?. 

Thank you.

brinton-chand:
https://angular.io/guide/i18n#i18n-pipes

> By default, Angular only contains locale data for en-US. If you set the value of LOCALE_ID to another locale, you must import locale data for that new locale.

Check the link for instructions in how to import. Hope this helps.

rameshyala:
Thank you @brinton-chand the locale issue is works fine, now this date picker is having a change detection option for reactive forms ?.

adityarb88:
@rameshyala: What exactly do you mean by change detection problems? Can you replicate this on a StackBlitz? In the StackBlitz that you posted above, I do not see Reactive Forms.

Our docs has an example of using Date Picker in reactive forms. May be that will help?
https://vmware.github.io/clarity/documentation/v0.11/datepicker
![image](https://user-images.githubusercontent.com/1426805/40426476-2c1fc54c-5e69-11e8-9861-69a68374ce01.png)

rameshyala:
Hi, @adityarb88 i am just asking, if change the date (assume the date picker have default today date), when i am modify that date in that time i will get any event now? 
Thank you

adityarb88:
@rameshyala: What I was saying is that your StackBlitz doesnt have reactive forms. Can you replicate your issue exactly. If you look at the link above, we have provided a demo for Reactive Forms. And when the user modifies the date, an event is emitted. If you can replicate the issue you are facing in a StackBlitz we can take a look and let you know what the issue is.

adityarb88:
@rameshyala: I created a StackBlitz which demos this: https://stackblitz.com/edit/clarity-light-theme-v11-u1lzvu?file=app/app.component.html

Hope this helps.

rameshyala:
[...]

PR #2334, opened 2018-06-04
[NG] Date Picker Output Bug Fix

adityarb88:
Fixes: #2300

Date Picker did not emit an output when the user manually changed Date in the input without using the Date Picker popover. This happened in Reactive and Template Driven Forms.

We want to output the date when the following conditions are true:
1. User has updated the input and focused out of the input.
2. Modified Date is not equal to the current Selected Date.

Since, in `date-input.ts`, we subscribe to `ngControl.valueChanges`, the selected date was updated before to the modified date before it could be compared to decide whether to emit the output or not.

This PR fixes the issue mentioned above.

Tested on Chrome.

### DIFF
--- a/src/clr-angular/forms/datepicker/date-input.ts
+++ b/src/clr-angular/forms/datepicker/date-input.ts
@@ -42,6 +42,22 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
    */
   private _subscriptions: Subscription[] = [];
 
+  //We need this variable because if the date input has a value initialized
+  //we do not output it. This variable is false during initial load. We make sure that
+  //during initial load dayModelOutputted is equal to the value entered by the user so that initialized
+  //value isn't emitted back to the user. After initial load,
+  //we set this to true and the dayModelOutputted is set only
+  //when the Output is emitted to the user.
+  private previousOutputInitializedFlag: boolean = false;
+  private previousOutput: DayModel;
+
+  private initializePreviousOutput(dayModel: DayModel) {
+    if (!this.previousOutputInitializedFlag) {
+      this.previousOutput = dayModel;
+      this.previousOutputInitializedFlag = true;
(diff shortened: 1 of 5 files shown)


src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -42,6 +42,22 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
    */
   private _subscriptions: Subscription[] = [];
 
+  //We need this variable because if the date input has a value initialized
+  //we do not output it. This variable is false during initial load. We make sure that
+  //during initial load dayModelOutputted is equal to the value entered by the user so that initialized
+  //value isn't emitted back to the user. After initial load,
+  //we set this to true and the dayModelOutputted is set only
+  //when the Output is emitted to the user.
+  private dayModelOutputtedInitialized: boolean = false;
```
hippee-lee: is there a reason for not naming it `dayModelOutputInitialized`? Its the `Outputted` part that seems odd to me.

src/dev/src/app/datepicker/datepicker-in-reactive-forms.ts
```diff
@@ -14,4 +14,8 @@ import { FormControl, FormGroup } from '@angular/forms';
 })
 export class DatepickerInReactiveForms {
   dateForm = new FormGroup({ date: new FormControl('03/05/2018'), name: new FormControl('Jane') });
+
+  dateChanged(date: Date) {
+    console.log(date);
```
hippee-lee: Should this still be here?

src/clr-angular/forms/datepicker/date-input.ts
gnomeontherun: I agree with @hippee-lee to keep the names consistent with `dayModelOutput`

PR #2335, opened 2018-06-04
[NG] clrDraggableGhost component

Shijir:

Demo: http://dnd-ghost.surge.sh/drag-and-drop

Tested in: Chrome, Firefox, Safari, Edge, IE11


src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
```diff
@@ -0,0 +1,22 @@
+@include exports("drag-and-drop.clarity") {
+    .drag-handle {
+        cursor: grab;
+    }
+    .draggable-ghost {
+        position: fixed;
+        visibility: hidden;
+        z-index: 2147483647; // max z-index;
```
adityarb88: There is a map called `$clr-layers`. Can we add the `z-index` to that and reference it from there?

src/clr-angular/utils/drag-and-drop/all.spec.ts
```diff
@@ -4,22 +4,25 @@
...
 import ClrDraggableSpecs from "./draggable.spec";
 import ClrIfDraggedSpecs from "./if-dragged.spec";
 import ClrDragAndDropEventBusSpecs from "./providers/drag-and-drop-event-bus.spec";
 import ClrDragEventListenerSpecs from "./providers/drag-event-listener.spec";
 import ClrDragHandleRegistrarSpecs from "./providers/drag-handle-registrar.spec";
+import ClrDraggableStateRegistrarSpecs from "./providers/draggable-state-registrar.spec";
 
-
-describe("Drag And Drop", function() {
+fdescribe("Drag And Drop", function() {
```
adityarb88: What the `fdescribe`?

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
youdz: Please remind me to add something like "All `z-index` properties must use the `$clr-layers` variable because blahblahblah" to our guidelines. You made a very good point raising this.

Or if anyone feels like adding this themselves right now, feel free to do it. 😄

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
```diff
@@ -0,0 +1,24 @@
...
+        pointer-events: none;
+        will-change: left, top;
```
youdz: Ooh fancy. I hadn't seen that in a while.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    constructor(private el: ElementRef, @Optional() private dragEventListener: ClrDragEventListener<T>,
+                private renderer: Renderer2, private ngZone: NgZone,
+                private draggableStateRegistrar: ClrDraggableStateRegistrar<T>) {
+        if (!this.dragEventListener) {
+            throw new Error("The clr-draggable-ghost component can only be used inside of a clrDraggable directive.");
+        }
+
+        this.draggableGhostEl = this.el.nativeElement;
+        this.renderer.addClass(document.body, "in-drag");
```
youdz: Nitpick: I feel like this doesn't belong here. Maybe in the draggable itself or in the structural directive? I'd almost want this to be in a service, since it's here to make sure the app itself has the right behavior (no selection) more than having the draggable look good.

Please comment here or talk about it with the team on Slack before changing this, I'm not sure where the best place for this line of code is and I don't want you to have to change it 5 times.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
```diff
@@ -0,0 +1,44 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Injectable} from "@angular/core";
+import {DomAdapter} from "../../dom-adapter/dom-adapter";
+import {ClrDragEvent} from "../interfaces/drag-event";
+
+
+@Injectable()
+export class ClrDraggableStateRegistrar<T> {
```
youdz: I'm still trying to understand what this does, but I can tell you for sure the name `ClrDraggableStateRegistrar` doesn't help at all. I have no idea what the "state" of a draggable is, or why you'd have to register it.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    private draggableEl: Node;
+    private draggableElClientRect: ClientRect;
+    private draggableElComputedStyle: CSSStyleDeclaration;
+    private stateEvent: ClrDragEvent<T>;
+
+    public register(el: Node, event: ClrDragEvent<T>) {
+        this.draggableEl = el;
+        this.draggableElClientRect = this.domAdapter.clientRect(this.draggableEl);
+        this.draggableElComputedStyle = getComputedStyle(<HTMLElement>this.draggableEl);
```
youdz: This should go through the DomAdapter.

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -3,40 +3,82 @@
...
+                private draggableStateRegistrar: ClrDraggableStateRegistrar<T>) {
         this.draggableEl = this.el.nativeElement;
+        this.componentFactory = this.cfr.resolveComponentFactory<ClrDraggableGhost<T>>(ClrDraggableGhost);
     }
 
+    @ContentChild(ClrIfDragged) customGhost: ClrDraggableGhost<T>;
+
+    private createDefaultGhost(event: ClrDragEvent<T>) {
+        this.draggableStateRegistrar.register(this.draggableEl, event);
+        this.viewContainerRef.createComponent(this.componentFactory, 0, this.injector,
```
youdz: Can you please just add a comment explaining that in this case, the ghost will be **outside** of the draggable in the DOM (which means a sibling)? I just want to have something written in case we forget and decide to base our styling based on parent selectors..

src/clr-angular/utils/drag-and-drop/draggable.ts
```diff
@@ -3,40 +3,82 @@
...
     constructor(private el: ElementRef, private dragEventListener: ClrDragEventListener<T>,
-                private dragHandleRegistrar: ClrDragHandleRegistrar<T>) {
+                private dragHandleRegistrar: ClrDragHandleRegistrar<T>, private viewContainerRef: ViewContainerRef,
+                private cfr: ComponentFactoryResolver, private injector: Injector,
+                private draggableStateRegistrar: ClrDraggableStateRegistrar<T>) {
         this.draggableEl = this.el.nativeElement;
+        this.componentFactory = this.cfr.resolveComponentFactory<ClrDraggableGhost<T>>(ClrDraggableGhost);
     }
 
+    @ContentChild(ClrIfDragged) customGhost: ClrDraggableGhost<T>;
```
youdz: You're querying `ClrIfDragged` but you're typing what you receive as `ClrDraggableGhost`. I have no idea why this doesn't throw an error in AoT, but it should. What you want is:
```typescript
@ContentChild(ClrIfDragged) customGhost: ClrIfDragged<T>;
```

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
```diff
@@ -0,0 +1,24 @@
...
+        opacity: 0.6;
+        transform: translate3d(0px, 0px, 0);
```
youdz: Why the `transform`? As far as I see, this is unused since you are animating `top` and `left`.

src/app/drag-and-drop/custom-ghost.demo.html
```diff
@@ -0,0 +1,15 @@
...
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h4>Draggable button with a custom ghost</h4>
+
+<div class="clr-example">
+  <button class="btn btn-primary" clrDraggable>draggable
+        <clr-draggable-ghost *clrIfDragged>
+            <button type="submit" class="btn btn-success">custom ghost</button>
```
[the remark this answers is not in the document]
Shijir: Sure, I am thinking of using `clr-icon`, probably hourglass shape or mail shape.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
Shijir: I pondered this too. But at the same, the purpose of applying this `"in-drag"` class' styles is to make the underlaying document possible to have the hovering draggable-ghost element over. It's facilitating draggable-ghost. So as I didn't see strong, compelling reasons to view this line of code outrageously out of place, I kept it here for now.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
Shijir: I am not fond of this name too, but this is what I have come up with tentatively. The goal of this service is to capture the state that the `draggable` element was in so that we could apply it to the `draggable-ghost` element. The captured state means draggable's BoundingClientRect, ComputeStyles, and the event that draggable element was captured in.

Maybe, I should also add in a comment explaining what this service is about too.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
Shijir: okay, I will add getComputedStyle to the DomAdapter service.

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: Sure, will do that.

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: Ooh right, good catch! 

On the side, I am also wondering it could be the way it is in Angular? Because `clrIfDragged` is placed on the `clr-draggable-ghost` component. So it's kind of both `IfDraggad` and `ClrDraggableGhost` at the same time. Nevertheless, I will change according to your suggestion.

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
Shijir: Yes, I can remove it now. It doesn't offer any benefits here.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+import {animate, style, transition, trigger} from "@angular/animations";
+import {Component, ElementRef, HostBinding, NgZone, OnDestroy, Optional, Renderer2} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
+
+import {ClrDragEvent} from "./interfaces/drag-event";
+import {ClrDragEventListener} from "./providers/drag-event-listener";
+import {ClrDraggableStateRegistrar} from "./providers/draggable-state-registrar";
+
+const ANIMATION_DURATION = "0.2s ease-in-out";
```
adityarb88: `ANIMATION_DURATION` is misleading. Has both the duration and the timing function. May be just call it `ANIMATION_PROPS` or a better name if you can think of one? I am really bad at this.

adityarb88: oooh how about `ANIMATION_CONFIG`?

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
adityarb88: Just a note here, we do a similar kind of thing for Responsive Navigation or Modals (not sure which one) with the `ScrollingService`.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,116 @@
...
+        const draggableClientRectWidth = this.draggableStateRegistrar.clientRect.width;
+        const draggableClientRectHeight = this.draggableStateRegistrar.clientRect.height;
+
+        const draggableMarginLeft = this.draggableStateRegistrar.computedStyle.marginLeft;
+        const draggableMarginTop = this.draggableStateRegistrar.computedStyle.marginTop;
+
+        this.renderer.setStyle(this.draggableGhostEl, "margin-left", `-${draggableMarginLeft}`);
+        this.renderer.setStyle(this.draggableGhostEl, "margin-top", `-${draggableMarginTop}`);
+
+        this.renderer.setStyle(this.draggableGhostEl, "width", `${draggableClientRectWidth}px`);
```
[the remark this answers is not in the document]
Shijir: Right, it comes with the same style as the original draggable has. But also we make it fixed element and its left position becomes the draggable's `.getBoundingClientRect().left`, which already accounts for the margin of the draggable margin space. So the ghost receives extra margin space. That's why I applied the margin spaces in opposite directions to compensate.

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
Shijir: This z-index has been moved to `clr-layers`.

src/app/drag-and-drop/custom-ghost.demo.html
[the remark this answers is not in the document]
Shijir: I've used the centered `event` shape icon now as the ghost now.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
Shijir: I've used `ClrDraggableSnapshot`.  The `register()` method has become `capture()`. And the `unregister()` has become `discard()`.

src/clr-angular/utils/drag-and-drop/providers/draggable-state-registrar.ts
Shijir: No longer needed to access `getComputedStyle` :-)

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: Added the comment.

src/clr-angular/utils/drag-and-drop/draggable.ts
Shijir: Addressed this issue.

src/clr-angular/utils/drag-and-drop/_drag-and-drop.clarity.scss
Shijir: Removed `transform`.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
Shijir: I shouldn't have set this animation config to a variable like this. So I've inlined this in the animations property of the component.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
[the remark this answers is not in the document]
Shijir: I no longer needed to subtract margins from the ghost calculations as I've just removed the original margins from the nested element inside the default ghost element through CSS.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
```diff
@@ -0,0 +1,116 @@
...
+    private initPosition: {pageX: number; pageY: number};
+    private dragPosition: {pageX: number; pageY: number};
+    private initDragDelta = {top: 0, left: 0};
+
+    @HostBinding("@leaveAnimation") leaveAnimConfig = {value: 0, params: {top: "0px", left: "0px"}};
+
+    constructor(private el: ElementRef, @Optional() private dragEventListener: ClrDragEventListener<T>,
+                private renderer: Renderer2, private ngZone: NgZone,
+                private draggableStateRegistrar: ClrDraggableStateRegistrar<T>) {
+        if (!this.dragEventListener) {
```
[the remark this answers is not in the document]
Shijir: Addressed this issue.

src/clr-angular/utils/drag-and-drop/draggable-ghost.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+import {MOCK_DRAG_EVENT_LISTENER_PROVIDER} from "./providers/drag-event-listener.mock";
+import {ClrDraggableStateRegistrar} from "./providers/draggable-state-registrar";
+
+export default function(): void {
+    describe("ClrDraggableGhost", function() {
+        describe("Without Wrapping ClrDraggable", function() {
+            it("should throw an error with a message", function() {
+                TestBed.configureTestingModule({
+                    declarations: [WithNoWrappingDraggable, ClrDraggableGhost],
+                    providers: [ClrDraggableStateRegistrar, DomAdapter]
```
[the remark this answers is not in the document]
Shijir: Addressed this issue.

src/clr-angular/utils/drag-and-drop/draggable-ghost.ts
adityarb88: @Shijir did you get a chance to look at the `ScrollingService` and check if we can do something like that here?

Shijir: Looks promising. I can incorporate something similar to that.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -6,32 +6,83 @@
...
 
             afterEach(function() {
                 this.fixture.destroy();
             });
 
+            it("should have draggable class", function() {
+                expect(this.draggable.nativeElement.classList.contains("draggable")).toBeTruthy();
+            });
+
+            it("should emit event on drag start and have being-dragged class", function() {
```
adityarb88: Shouldn't this be broken down into 2 tests? You are testing 2 different things here.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -6,32 +6,83 @@
...
+
+            it("should emit event on drag move", function() {
+                const mockDragMoveEvent = {dragPosition: {pageX: 33, pageY: 44}};
+                this.dragEventListener.dragMoved.next(mockDragMoveEvent);
+                expect(this.testComponent.dragStartEvent).toBeUndefined();
+                expect(this.testComponent.dragMoveEvent).toBe(mockDragMoveEvent);
+                expect(this.testComponent.dragEndEvent).toBeUndefined();
+            });
+
+            it("should emit event on drag end", function() {
```
adityarb88: Here the description is missing the class test that you are doing. If you are deciding to break the earlier test into 2 parts then this should be too. If not, the description should be updated to include the class details.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -6,32 +6,83 @@
...
+import {DomAdapter} from "../dom-adapter/dom-adapter";
+import {ClrDragAndDropModule} from "./drag-and-drop.module";
 import {ClrDragHandle} from "./drag-handle";
 import {ClrDraggable} from "./draggable";
-import {ClrDragAndDropEventBus} from "./providers/drag-and-drop-event-bus";
 import {ClrDragEventListener} from "./providers/drag-event-listener";
+import {MOCK_DRAG_EVENT_LISTENER_PROVIDER} from "./providers/drag-event-listener.mock";
 import {ClrDragHandleRegistrar} from "./providers/drag-handle-registrar";
+import {ClrDraggableSnapshot} from "./providers/draggable-snapshot";
+import {ClrGlobalDragMode} from "./providers/global-drag-mode";
```
adityarb88: For all new tests, the convention is to add the test components after the imports and then start testing those components. Could you move your test components right after the imports please?

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -41,24 +92,168 @@ export default function(): void {
                 expect(this.dragHandleRegistrar.customHandleEl).toBeUndefined();
             });
 
+            it("should instantiate cloned version of draggable as ghost on drag start", function() {
```
adityarb88: Aren't you also testing that the ghost is cloned right next to the draggable? We should update the description accordingly.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -41,24 +92,168 @@ export default function(): void {
...
+                this.dragEventListener.dragStarted.next();
+                expect(this.fixture.nativeElement.querySelectorAll("clr-draggable-ghost").length).toBe(1);
+                const draggableGhost = this.fixture.nativeElement.querySelector("clr-draggable-ghost");
```
adityarb88: nit: Instead of `querySelectorAll` and then `querySelector` to `const draggableGhost`, why not just `querySelectorAll` to `const draggableGhost` and use the first child of it everywhere after testing that the length of the result of `querySelectorAll` is 1

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
```diff
@@ -41,24 +92,168 @@ export default function(): void {
...
-                expect(this.dragHandleRegistrar.defaultHandleEl)
-                    .toBe(
-                        this.draggable.nativeElement,
-                        `The default handle should be still set to the draggable element even though it's not made an active drag handle.`);
+            });
+
+            it("should project custom ghost on drag start", function() {
+                this.dragEventListener.dragStarted.next();
+                expect(this.fixture.nativeElement.querySelectorAll("clr-draggable-ghost").length).toBe(1);
+                const draggableGhost = this.fixture.nativeElement.querySelector("clr-draggable-ghost");
```
adityarb88: Same comment about `querySelectorAll` as above.

src/clr-angular/utils/drag-and-drop/if-dragged.ts
```diff
@@ -17,9 +17,9 @@ import {ClrDragEventListener} from "./providers/drag-event-listener";
 @Directive({selector: "[clrIfDragged]"})
 export class ClrIfDragged<T> implements OnDestroy {
     private subscriptions: Subscription[] = [];
-    constructor(private template: TemplateRef<any>, private container: ViewContainerRef,
+    constructor(private template: TemplateRef<any>, @Optional() @SkipSelf() private container: ViewContainerRef,
```
adityarb88: I remember we discussed that we should add `@SkipSelf` but why do we need to add `@Optional()`?

src/clr-angular/utils/drag-and-drop/if-dragged.ts
```diff
@@ -17,9 +17,9 @@ import {ClrDragEventListener} from "./providers/drag-event-listener";
...
                 @Optional() private dragEventListener: ClrDragEventListener<T>) {
-        if (!this.dragEventListener) {
+        if (!this.dragEventListener && !this.container) {
```
adityarb88: Is there a use case where only one of these conditions is true?

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
```diff
@@ -0,0 +1,36 @@
...
+import {ClrDragEvent} from "../interfaces/drag-event";
+
+// This service is used to capture the state of clrDraggable element
+// at a certain event and passes it to clrDraggableGhost component.
+@Injectable()
+export class ClrDraggableSnapshot<T> {
+    constructor(private domAdapter: DomAdapter) {}
+
+    private draggableElClientRect: ClientRect;
+    private stateEvent: ClrDragEvent<T>;
```
adityarb88: shouldn't this be `dragEvent` instead of `stateEvent`? The naming seems off.

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
```diff
@@ -0,0 +1,36 @@
...
+        delete this.draggableElClientRect;
+        delete this.stateEvent;
+    }
+    get hasDraggableState() {
+        return !!this.stateEvent && !!this.draggableElClientRect;
+    }
+    get clientRect() {
+        return this.draggableElClientRect;
+    }
+    get event() {
```
adityarb88: same here. Should this be `dragEvent`? Also can we add the type for this getter?

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    public capture(el: Node, event: ClrDragEvent<T>) {
+        this.draggableElClientRect = this.domAdapter.clientRect(el);
+        this.stateEvent = event;
+    }
+    public discard() {
+        delete this.draggableElClientRect;
+        delete this.stateEvent;
+    }
+    get hasDraggableState() {
```
adityarb88: `boolean` type

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+        mockDraggable.style.width = "100px";
+        mockDraggable.style.height = "50px";
+        mockDraggable.style.left = "90px";
+        mockDraggable.style.top = "45px";
+
+        const domAdapter = new DomAdapter();
+        const draggableSnapshot = new ClrDraggableSnapshot(domAdapter);
+
+        it("registers element and sets clientRect and computedStyle", function() {
+            expect(draggableSnapshot.hasDraggableState).toBeFalsy();
```
adityarb88: Shouldn't you also test that `draggableSnapshot.clientRect` and `draggableSnapshot.event` are undefined before you run `capture`?

adityarb88: Seems odd to test just `hasDraggableState`

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+            expect(draggableSnapshot.clientRect).toEqual(domAdapter.clientRect(mockDraggable));
+            expect(draggableSnapshot.event).toEqual(mockDragMoveEvent);
+        });
+
+        it("unregisters element and deletes clientRect and computedStyle", function() {
+            expect(draggableSnapshot.hasDraggableState).toBeTruthy();
+            draggableSnapshot.discard();
+            expect(draggableSnapshot.hasDraggableState).toBeFalsy();
+            expect(draggableSnapshot.clientRect).toBeUndefined();
+            expect(draggableSnapshot.event).toBeUndefined();
```
adityarb88: nit: This `spec` deviates from all the other ones. We normally have `beforeEach` where we in this case you would create the `draggableSnapshot`. This tests has to run after the first test for it to pass, no?

IMO, it would have been better if you would run `capture` again here and then run `discard` directly to test if the last there `expect` are fine. You wouldn't have to test if the `hasDraggableState` is `true` and the other parts of the state are `undefined` because you already tested that in the first test.

src/clr-angular/utils/drag-and-drop/providers/global-drag-mode.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Injectable, Renderer2} from "@angular/core";
+
+@Injectable()
+export class ClrGlobalDragMode {
```
adityarb88: A comment to describe what this service does? Its not clear by the name of the class.

src/clr-angular/utils/drag-and-drop/providers/global-drag-mode.ts
```diff
@@ -0,0 +1,19 @@
...
+    constructor(private renderer: Renderer2) {}
+
+    enter(): void {
```
adityarb88: Do we want to raise an issue to fix `ScrollingService` too?

src/ks-app/src/app/containers/drag-and-drop/dnd.component.scss
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
```
adityarb88: If no CSS is required for the demo, then we should remove the file itself.

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
Shijir: this private property has become `snapshotEvent`, but gets accessed as `get dragEvent()` now.

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
Shijir: Addressed this.

src/ks-app/src/app/containers/drag-and-drop/dnd.component.scss
Shijir: Will keep it now because I will probably use to show the demos in KS-app.

src/clr-angular/utils/drag-and-drop/providers/global-drag-mode.ts
Shijir: Sure, I will bring the discussion to the team.

src/clr-angular/utils/drag-and-drop/if-dragged.ts
Shijir: This should have been `||` instead of `&&` here. If only any of them is undefined, should throw an error.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
Shijir: Thanks. Addressing this right now.

src/clr-angular/utils/drag-and-drop/draggable-ghost.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+            });
+
+            it("should have draggable-ghost class", function() {
+                expect(this.ghostElement.classList.contains("draggable-ghost")).toBeTruthy();
+            });
+
+            it("should remove in-drag class from document once destroyed", function() {
+                this.testComponent.display = false;
+                this.fixture.detectChanges();
+                expect(document.body.classList.contains("in-drag")).toBeFalsy();
```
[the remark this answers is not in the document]
Shijir: Forgot to remove it. Addressed this issue.

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.ts
Shijir: Addressed this.

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.spec.ts
Shijir: My bad. I made the tests independent here.

src/clr-angular/utils/drag-and-drop/providers/global-drag-mode.ts
Shijir: Addressed this issue

src/clr-angular/utils/drag-and-drop/providers/draggable-snapshot.spec.ts
Shijir: Addressed this issue.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
Shijir: Addressed this one

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
Shijir: Address this.

src/clr-angular/utils/drag-and-drop/draggable.spec.ts
Shijir: Addressed this.

Issue #2038, opened 2018-03-06
Datagrid clrDgSingleSelectedChange can emit wrong values


marcojahn:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
When using Datagrid with singleselection on a remote dataset I expect the clrDgSingleSelectedChange not to fire with an NOT existing entry (when the selected entry has been deleted).
Otherwise is would like to have an option to reset the selection (did not figure out a way to do this correctly)

_We are sure that this behaviour has changed within the last weeks because it was working as expected as far as we can say, but didn't have a UI test that has covered this situation (but now have)
The last "known good" behaviour was around 12/2017 // 01/2018_

### Actual behavior
Long story short (please see stackblitz with an implementation reproducing the behaviour).
I describe two scenarios that you can "replay" in the stackblitz example (I output `Date.now()` in the event so that you can see that it is really emited)

#### Scenario 1
1) Init a Datagrid with `[]`
2) Load remote data `ngOnInit() -> Observable.subscribe() -> write data``
3) Grid will emit `clrDgSingleSelectedChange` with `null` (nothing selected)
4) Do NOT select an entry and click "delete an item"
5) The item is removed from the remoteList, the grid is reloaded afterwards
6) Grid will emit `clrDgSingleSelectedChange` with `null` (nothing selected)
* this behaviour does not break our implementation*

#### Scenario 2
Steps 1-3 are the same as in _Scenario 1_
4) Select User ID *9876* and click "delete item" (the last item of the list is always deleted)
5) The item is removed from the remoteList, the grid is reloaded afterwards
6) Grid will emit `clrDgSingleSelectedChange` with `{ "id": "9876", "name": "bar", "mail": "bar@bar.com" }`
*The item that has been deleted (remote) and is not available anymore in the grid is emitted as selected*

### Reproduction of behavior
[working plunker](https://stackblitz.com/edit/clarity-light-theme-v11-d88ddq)

### Environment details

* **Angular version:** 5.2.x

* **Clarity version:** 0.11.5

* **OS and version:** all

* **Browser:** [all]

youdz:
Wow, I see 3 bugs here: 
- The first one is the initial output which should not happen, and was apparently introduced by #1272 where an empty selection is set to `null`. We end up with a bunch of problems because of poor `null` and `undefined` juggling.
[...]

PR #2344, opened 2018-06-06
[ng] Clean up datagrid selection preservation (#2038)

jeeyun:

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid.spec.ts
@@ -300,7 +300,7 @@ export default function(): void {
 
       describe('Template API', function() {
         it('sets the currentSingle binding', function() {
-          expect(selection.currentSingle).toBeNull();
+          expect(selection.currentSingle).toBeUndefined();
           context.testComponent.selected = 1;
           context.detectChanges();
           expect(selection.currentSingle).toEqual(1);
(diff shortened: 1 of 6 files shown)


src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -300,17 +301,47 @@ export default function(): void {
...
+          expect(selection.currentSingle).toBeUndefined();
+          context.testComponent.selected = 1;
+          context.detectChanges();
+          expect(selection.currentSingle).toEqual(1);
+          context.testComponent.selected = null;
+          context.detectChanges();
           expect(selection.currentSingle).toBeNull();
+        });
+
+        it('does not emit a change event for initial value of currentSingle that is undefined', function() {
```
gnomeontherun: I think we want to break this up into separate tests.


1. `it does not emit a change event on initialization, before selection`
2. `it does not emit a change event when selecting the same value`
3. `it does not emit a change event when setting selection to undefined if already null, and vice versa`
4. `it emits a change event when changing the selection`

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -39,44 +39,85 @@ export class Selection {
...
+            let newSingle: any;
+            const trackBy: TrackByFunction<any> = this._items.trackBy;
+            let selectionUpdated: boolean = false;
+
+            updatedItems.forEach((item, index) => {
+              const ref = trackBy(index, item);
+              // If one of the updated items is the previously selectedSingle, set it as the new one
+              if (this.prevSingleSelectionRef === ref) {
+                newSingle = item;
+                selectionUpdated = true;
```
gnomeontherun: Ok thinking about the cases here, and I _think_ this is what will happen. In all cases, if you don't select an item and reload the list, then it will not emit (as expected). However, if you select and reload the list, I believe the behaviors are as follows.

1. ngFor, no trackBy (or trackBy index). If you have already selected an item (which is by index), any reloading of the dataset would essentially select the new item at that index regardless if it was the original item or not. Do we address this by always clearing the selection in this case on reload, or is this not a supported use case? 
2. ngFor, trackBy. If you already have selected an item (which is track by the item), and reload the dataset, if the reference exists it will set and emit it again. If the reference doesn't exist, it will do nothing. Outside of the extra emit, this looks correct.
3. clrDgItems, no trackBy (or trackBy index). If you select an item (tracked by index), and you reload, we run into the same problem as 1 where the index will remain stored and it will select the new item at that index and emit it. If the reference doesn't exist (as in the index is larger than the length of the new list), no refs would match and it would be set to `undefined` and emitted. 
4. clrDgItems, trackBy. If you select an item (track by item) and reload, if the reference exists in the new list the ref will match and the selection will be emitted again. If the reference doesn't exist, it is cleared and set to `undefined`. Outside of the extra emit, this looks correct.

So the problems I see:

1. If a new set of items is loaded and the trackBy (index or item) matches an item it will be emitted again (since emit happens in the `currentSingle` setter). Do we have an expectation that the event will _only_ fire when a change actually happens, or is this an acceptable case? I think we can live with it, as a limitation where apps will have to handle possible multiple emits of the same selected value (or it should be an idempotent event handler). 
2. If we have a computed reference that is an index, then reloading the list will automatically set `currentSingle` to the new item at that index (regardless if it is the same item or not). This is a new issue not originally outlined. If you aren't using the smart iterator, then perhaps we should always clear the `currentSingle` value, and if you don't want that behavior then use a trackBy item calculation?

gnomeontherun: Talked with Jeeyun more, while both are quirky cases, they are generally no the ideal case and are best handled with having better documentation about how to use trackBy and selection modes.

src/clr-angular/data/datagrid/datagrid.spec.ts
jeeyun: Will break it up into four tests.

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -300,7 +300,7 @@ export default function(): void {
 
       describe('Template API', function() {
         it('sets the currentSingle binding', function() {
-          expect(selection.currentSingle).toBeNull();
+          expect(selection.currentSingle).toBeUndefined();
```
adityarb88: I don't see any changes to the `currentSingle` declaration. How did this test pass before?

PR #2347, opened 2018-06-07
[NG] Proper Datagrid typing

youdz:
This is a pass over our datagrid code to parametrize everything that needs to by the types of items we receive. It drastically reduces the number of `any` and improves the stability of our code.

I also removed old casts that are not needed anymore since both Angular and Typescript evolved and can now guess types better.

Deprecated exports are not type parametrized, so we simply use `<any>` for them.

### I just realized this is a breaking change for people who import our Datagrid classes in Typescript to interact dynamically with them.
I will mark this as "do not merge" until I either find a solution, we decide to squeeze it into 0.12 at the last minute, or we sit on it for long enough that it can go in the next major.

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -240,26 +240,26 @@ export declare class ClrControlError {
 export declare class ClrControlHelper {
 }
 
-export declare class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
+export declare class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, OnDestroy {
     SELECTION_TYPE: typeof SelectionType;
     allSelected: boolean;
-    columns: QueryList<ClrDatagridColumn>;
+    columns: QueryList<ClrDatagridColumn<T>>;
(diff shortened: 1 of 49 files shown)


youdz:
I forgot to do a pass on the app, silly me.

gnomeontherun:
I don't like the idea of throwing a breaking change in 0.12 when we're already at RC. I'd opt for this to go in when we branch for 0.13, otherwise it looks good to me.

m-tomczyk:
Hey @youdz I guess that breaking changes in this PR are because of introduction of generics? That can be made into non-breaking change using default generic type mechanism: https://github.com/Microsoft/TypeScript/issues/2175 . For newly added generics you should be able to simply `Foo<T = any>` and usage should default to just `Foo`
I don't know if 0.13 will be coming soon enough that this would be wasted effort tho.

youdz:
Wow @m-tomczyk. 😮 

Whether we decide to just wait for 0.13 or use the default generic now, we will _definitely_ use this in the future. Thank you so much for pointing a Typescript feature we didn't know about, this is awesome. You win the open-source. ❤️

PR #2354, opened 2018-06-12
[NG] Input working with grid automatically, and host wrapping on inputs

gnomeontherun:
This adds a few new directives and a service to manage the layout option for the form. It will use that to conditionally add the new grid classes that will auto-flip to vertical on narrow screens regardless of layout choice. It also adds host wrapping to the input so it will always be able to layout correctly.


src/clr-angular/forms/common/label.ts
```diff
@@ -29,6 +31,10 @@ export class ClrLabel implements OnInit, OnDestroy {
     if (this.ifErrorService) {
       this.renderer.addClass(this.el.nativeElement, 'clr-control-label');
     }
+    if (this.ifErrorService && !this.layoutService.isVertical()) {
+      this.renderer.addClass(this.el.nativeElement, 'clr-col-xs-12');
+      this.renderer.addClass(this.el.nativeElement, 'clr-col-md-2');
```
adityarb88: Why `clr-col-md-2` as the default? What happens if the user requires the label to occupy more width? 

For eg: What happens if someone uses a vertical form in a small modal and the breakpoint is greater than `md`?

src/clr-angular/forms/common/label.ts
```diff
@@ -8,12 +8,14 @@ import { Subscription } from 'rxjs';
 
 import { IfErrorService } from './if-error/if-error.service';
 import { ControlIdService } from './providers/control-id.service';
+import { LayoutService } from './providers/layout.service';
 
 @Directive({ selector: 'label' })
```
adityarb88: I know this was introduced before but won't the HTML tag selector `label` target all labels whether they are in or outside the form?

adityarb88: I think I missed asking about this in the earlier review.

gnomeontherun: Yes they do target all labels, but are largely inert unless they are inside a form control container that provides the services.

src/clr-angular/forms/common/label.ts
gnomeontherun: This is the default based on what aligns with the design best, but it does not have an option to override or change it.

I could have it first check if `clr-col-xxx` classes exist and only apply these if they are not present.

src/clr-angular/forms/common/layout.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(private el: ElementRef, private renderer: Renderer2, private layoutService: LayoutService) {}
+
+  ngAfterContentInit() {
+    this.renderer.addClass(this.el.nativeElement, 'clr-form-' + this.layoutService.layout);
+  }
+
+  ngOnInit() {
+    if (this.layout) {
+      this.layoutService.layout = this.layout;
```
adityarb88: Should we be able to change the layout from horizontal to vertical or vice-versa using the input? Here we are initializing the layout once.

src/clr-angular/forms/input/input-container.ts
```diff
@@ -23,19 +28,42 @@ import { NgControlService } from '../common/providers/ng-control.service';
...
     });
   }
 
+  controlClass() {
+    const controlClasses = [];
+    if (this.invalid) {
+      controlClasses.push('clr-error');
+    }
+    if (!this.isVertical()) {
+      controlClasses.push('clr-col-md-10 clr-col-xs-12');
```
adityarb88: Same question here. What happens if the user wants to override the classes?

gnomeontherun: I'll add a check to only add if there are no existing col- classes

src/clr-angular/forms/common/wrapped-control.ts
```diff
@@ -12,7 +12,7 @@ import { ControlIdService } from './providers/control-id.service';
 export class WrappedFormControl<W extends DynamicWrapper> implements OnInit {
   // I lost way too much time trying to make this work without injecting the ViewContainerRef and the Injector,
   // I'm giving up. So we have to inject these two manually for now.
-  constructor(protected wrapperType: Type<W>, protected vcr: ViewContainerRef) {}
+  constructor(protected wrapperType: Type<W>, protected vcr: ViewContainerRef, protected index: number = 0) {}
```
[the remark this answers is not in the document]
gnomeontherun: I can conjure one.

src/clr-angular/forms/common/layout.ts
gnomeontherun: This does seem like something you want to do once, and because changing the layout is a jarring experience for the user. Eudes and I talked about it briefly, and I believe we agreed on that point but happy to debate it further.

I'll add some check to ensure the input value is valid.

src/clr-angular/forms/common/layout.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive, ElementRef, Input, Renderer2 } from '@angular/core';
+import { Layouts, LayoutService } from './providers/layout.service';
+
+@Directive({
+  selector: '[clrLayout]',
+})
+export class ClrLayout {
```
Shijir: Missing Lifecycle hooks interfaces here?

src/clr-angular/forms/input/input-container.ts
```diff
@@ -23,19 +28,42 @@ import { NgControlService } from '../common/providers/ng-control.service';
...
 })
-export class ClrInputContainer implements OnDestroy {
+export class ClrInputContainer implements DynamicWrapper, OnDestroy {
   subscription: Subscription;
   invalid = false;
+  _dynamic = false;
+  @ContentChild(ClrLabel) label: ClrLabel;
 
-  constructor(private ifErrorService: IfErrorService) {
+  constructor(private ifErrorService: IfErrorService, @Optional() private layoutService: LayoutService) {
```
Shijir: Should this `LayoutService` be optional here? In case it's missing, can we throw an error that the component should be placed inside a clrForm.

src/clr-angular/forms/common/layout.ts
```diff
@@ -0,0 +1,27 @@
...
+  @Input('clrLayout') layout: Layouts;
+
+  constructor(private el: ElementRef, private renderer: Renderer2, private layoutService: LayoutService) {}
+
+  ngAfterContentInit() {
+    this.renderer.addClass(this.el.nativeElement, 'clr-form-' + this.layoutService.layout);
```
Shijir: I am not really sure why this line of code should use `Renderer2` and placed inside `ngAfterContentInit`. 

Also, I am just not really fond of accessing a part of a class name from another service and concatenating here. Can we access the entire class name from that same service? In that case, if this class name changes, we would need to make changes to only that service file.

src/clr-angular/forms/input/input-container.ts
gnomeontherun: It is possible to have a form control without the `clrForm`, so it is optional.

src/clr-angular/forms/common/layout.ts
gnomeontherun: To do this you have to know the host property value at compile time. Like this:

```
host: {
  '[class.clr-form-{layoutType}]: 'true'
}
```

As far as the class name, I could have the service generate it instead of here.

Shijir: We can do this:
```html
host: {
    "[class]": "layoutTypeClass",
 }
```
```ts
this.layoutTypeClass = 'clr-form-' + this.layoutService.layout;
```
With this, we don't even need to inject `Renderer2` and `ngAfterContentInit` hook

sdurnov:
@gnomeontherun any docs updates on this feature?

gnomeontherun:
It has been merged, docs are coming.

PR #2365, opened 2018-06-14
first pass at forms documentation

gnomeontherun:
This adds the first pass at forms documentation with samples. There are a few other details to add, but they tend to fit more into the specific form controls themselves, which will come in another PR. The UX guidelines are marked as a WIP.


gnomeontherun:
This is waiting on the other forms work to be merged so putting it in a holding pattern.

PR #2378, opened 2018-06-20
[CORE] feature: Adds automatic publishing to build repos

hippee-lee:
- This will only run on master branch.
- It published to [clr-team](https://github.com/clr-team) build repositories for each of the Clarity packages
- Automatic commit messages will include a link to the commit on master that triggered the build and the commit message.
- clr-angular has an example of the automatic commit message but its left over from testing I did on my fork so it doesn't actually link to a published commit (yet).
- the publish should not run if any of the test jobs fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,20 @@ cache:
   directories:
     - node_modules
 
-before_install:
-
 before_script:
   - npm install -g gemini@5.2.0 html-reporter@2.0.1
 
 script:
   - node scripts/travis
+
+jobs:
+  include:
+    - stage: publish
+      if: (NOT type IN (pull_request)) AND branch = master
+      before_script:
+        - npm run build
+      script:
+        - bash ./scripts/publish-build.sh
+
+
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ group: edge
 notifications:
   email:
   - wilkenj@vmware.com
+  - mhippely@vmware.com
   slack:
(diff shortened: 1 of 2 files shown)


scripts/publish-build.sh
```diff
@@ -0,0 +1,34 @@
...
+    cd "./${repo}"
+    # Delete all the files from last build
+    # A filepath to the dist folder in clraity project
+    DIST="$TRAVIS_BUILD_DIR/dist/${repo}/*"
+    # Copy all files into the build repo
+    cp -rf $DIST .
+    # Add any changes to git
+    git add -A
+    # Commit changes with a link to the vmware/clarity commit on master and its message
+    git commit -m "Clarity Build: http://github.com/vmware/clarity/commit/${SHA}\nMessage: ${MESSAGE}"
```
gnomeontherun: The `\n` doesn't make a newline. I like the link to the commit, but I think it should be secondary to the message.

```
git commit -m "${MESSAGE}

Clarity Commit: http://github.com/vmware/clarity/commit/${SHA}"
```

scripts/publish-build.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "Travis runs in the $(pwd) directory"
+# Set up the user for commiting and pushing.
+git config --global user.email "<mhippely@vmware.com>"
+git config --global user.name "clr-team"
+
+SHA=$(git rev-parse HEAD)
+MESSAGE=$(git log -1 --pretty=format:%s)
+
+# Clone repositories
+git clone https://${GITHUB_OAUTH_BUILD_TOKEN}@github.com/clr-team/clr-angular.git
```
gnomeontherun: Might as well move this inside of the for loop below. Also add `--depth 1` so it is faster and only clones the latest commit instead of the whole tree.

```
do
  git clone https://${GITHUB_OAUTH_BUILD_TOKEN}@github.com/clr-team/${repo}.git --depth 1
  ...
```

.travis.yml
```diff
@@ -43,10 +43,20 @@ cache:
...
 before_script:
   - npm install -g gemini@5.2.0 html-reporter@2.0.1
 
 script:
   - node scripts/travis
+
+jobs:
+  include:
+    - stage: publish
+      if: (NOT type IN (pull_request)) AND branch = master
```
gnomeontherun: Would we rather it be `if : type = push AND branch = master`? We only want to have this run on a code push to master right? The other types are api/cron, so we have a cron job doing daily testing now that would probably trigger this as well.

scripts/publish-build.sh
```diff
@@ -0,0 +1,34 @@
...
+    cp -rf $DIST .
+    # Add any changes to git
+    git add -A
+    # Commit changes with a link to the vmware/clarity commit on master and its message
+    git commit -m "Clarity Build: http://github.com/vmware/clarity/commit/${SHA}\nMessage: ${MESSAGE}"
+    # Push changes up the the build repo
+    git push
+    # Go back up to
+    cd ..
+done
```
gnomeontherun: General question, what about conditions like github is down or there is a problem with cloning/comitting? I suppose this would just error out, but can we get a notification?

scripts/publish-build.sh
hippee-lee: I'll move it into the loop and add the depth. I had it in an earlier version and forgot to add it in this one.

.travis.yml
hippee-lee: I'll test with your suggestion and  `if: (NOT type IN (api, cron, pull_request)) AND branch = master`

I'm not 100% sure if github does a push when you merge with the PR button from the web UI and I'm not seeing anything searching the docs. All I see is this
```
Pull requests are merged using the --no-ff option, except for pull requests with squashed or rebased commits, which are merged using the fast-forward option.
```

The intent here is to only run when changes are merged into the master branch.

scripts/publish-build.sh
hippee-lee: I can add error handling and notification(s) to slack.
### Error Notifications:
- github down, not able to clone issue
- commiting changes to local repo issue 
- pushing changes to remote repo issue
- dist files didn't build for some reason issue

### Success Notifications
- build success and link to the build repository

I was planning to add a success notification to a slack #clarity-build/#clarity-travis channel but it took me longer than anticipated to iron out running the script on travis. It might take me a little bit longer but I can add it in this PR instead of a different one later.

hippee-lee:
I still need to do some testing on my fork to make sure the build and notification behavior work as I expect them to so I added the `DO NOT MERGE` label for now. Code and config changes are ready for review.

Issue #2342, opened 2018-06-06
Datagrid: preserve selection when filtering


youdz:
**Select one ...** (check one with "x")

```
`[ ]` bug
`[ ]` feature request
`[x]` enhancement
```

### Expected behavior

I need to be able to incrementally build a selection of items in a datagrid through the following process:
1. filter the items of a datagrid down to a manageable number,
2. select one of these items, 
3. undo my filter,
4. repeat from step 1 for a new item.

### Actual behavior

Currently adding, removing or changing any filter simply erases the current selection, which makes the desired process impossible.

### Proposed solution

The overall selection we keep in memory should remain constant but private, while the  "public" selection should follow the current filters at all times. **Any actions performed through the action bar will apply only to the public selection** (this is **critical** to prevent destructive actions from happening on unwanted items). To ensure this, we simply have to make sure the private selection remains truly **private**, our consumers will never need it.

Here is a step-by-step use case example showing what the private and public selections become based on user action, and how many selected items the selection counter in the footer of the data grid displays:

1. My dataset is the list of states of the United States.
2. I select Alabama and California.
  Private selection: `["Alabama", "California"]`
  Public selection: `["Alabama", "California"]`
  Selection counter: 2
3. I apply the filter "Jer".
  Private selection: `["Alabama", "California"]`
  Public selection: `[]`
  Selection counter: 0
4. I select New Jersey.
  Private selection: `["Alabama", "California", "New Jersey"]`
  Public selection: `["New Jersey"]`
  Selection counter: 1
5. I delete my "Jer" filter.
  Private selection: `["Alabama", "California", "New Jersey"]`
  Public selection: `["Alabama", "California", "New Jersey"]`
  Selection counter: 3
6. I apply the filter "New".
  Private selection: `["Alabama", "California", "New Jersey"]`
  Public selection: `["New Jersey"]`
  Selection counter: 1
7. I select New York.
  Private selection: `["Alabama", "California", "New Jersey", "New York"]`
  Public selection: `["New Jersey", "New York"]`
  Selection counter: 2
8. I realize my terrible mistake and unselect New Jersey.
  Private selection: `["Alabama", "California", "New York"]`
  Public selection: `["New York"]`
  Selection counter: 1
9. I select New Hampshire.
  Private selection: `["Alabama", "California", "New York", "New Hampshire"]`
  Public selection: `["New York", "New Hampshire"]`
  Selection counter: 2
10. I delete me "New" filter.
[...]

PR #2381, opened 2018-06-21
[ng] Datagrid bug fix for when selection is cleared by applying filter

jeeyun:
The datagrid clears the current selection to empty array when filter is applied and we have a separate ticket (https://github.com/vmware/clarity/issues/2342) to preserve selection when filter is applied. 

My code was not accounting for this fact, and was trying to re-select the previous selections when data is refreshed in the case where filter is applied as well. This results in having inaccurate current selection data post-filter, since `prevSelectionRefs` still contained refs for selection even though current is cleared.

### DIFF
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -184,6 +184,31 @@ export default function(): void {
         expect(nbChanges).toBe(3);
       });
 
+      it('does not emit selection change twice after a filter is applied', function() {
+        let nbChanges = 0;
+        let currentSelection: any;
+        selectionInstance.change.subscribe((items: any) => {
+          nbChanges++;
+          currentSelection = items;
+        });
+
+        selectionInstance.selectionType = SelectionType.Multi;
+        expect(nbChanges).toBe(1); // current is initialized to [] at this point
+
+        selectionInstance.current = [4, 2];
+        expect(nbChanges).toBe(2);
+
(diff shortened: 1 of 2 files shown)


src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -124,6 +129,7 @@ export class Selection {
 
   public clearSelection(): void {
     this.current.length = 0;
+    this.prevSelectionRefs = [];
```
gnomeontherun: Approving but would like a test for this.

jeeyun: I don't have a super direct way of testing this bug fix, but added a test case that I think is ok-ish. Take a look and let me know if you think there is a better way.

src/clr-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -184,6 +184,27 @@ export default function(): void {
         expect(nbChanges).toBe(3);
       });
 
+      it('does not emit selection change after a filter is applied and selection is cleared', function() {
```
youdz: This test description doesn't match what the test actually checks. I believe the test is correct and the description is wrong, but I don't understand what we're trying to fix here so I might be off. 😛

jeeyun: So, normally you'd expect a 4th change to fire, since the code to preserve selection applies and updates the current. In the case of filters, we skip that. Is there a better way to test this or would it be sufficient to add a more clear test description and comment?

PR #2395, opened 2018-06-26
[NG] Select an Option

adityarb88:
**Scope: Select an option from the menu and render it on the Input**

New additions with this PR:
1. Provider to manage option selection.
2. Select DOM Adapter (This will be merged with the Datagrid DOM Adapter that @Shijir  is going to extract with the Drag and Drop PR. For now, its created separately and not tested.)


src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Injectable} from "@angular/core";
+import {Subject} from "rxjs";
+import {Observable} from "rxjs/internal/Observable";
```
gnomeontherun: import from `rxjs` only, collapse into above import.

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    private _selectionChanged: Subject<void> = new Subject<void>();
+
+    get selectionChanged(): Observable<void> {
+        return this._selectionChanged.asObservable();
+    }
+
+    // Currently only handles single selection
+    updateSelection(option: ClrOption): void {
+        if (this.currentSelection.indexOf(option) > -1) {
```
gnomeontherun: I'm struggling with this method, its written to support an array of selected items but then forced to be only single selection. I'd like this method (and service) to be either 1) written to just handle single selection or 2) written to be ready for multi-selection.

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,36 @@
...
+import {Injectable} from "@angular/core";
+import {Subject} from "rxjs";
+import {Observable} from "rxjs/internal/Observable";
+
+import {ClrOption} from "../option";
+
+@Injectable()
+export class OptionSelectionService {
+    // Array as we need support for selecting multiple options
+    currentSelection: Array<ClrOption> = [];
```
gnomeontherun: This is a public property, but it would be better for the Subject below to emit the current selection instead for consumers to access. 

The only place I see it now is inside of the ClrSelect which can just get the array when the emit happens.

src/clr-angular/forms/select/providers/option-selection.service.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+            optionSelectionService.updateSelection(fakeOption1);
+
+            expect(optionSelectionService.currentSelection.length).toBe(1);
+            expect(optionSelectionService.currentSelection[0]).toBe(fakeOption1);
+        });
+
+        it("replaces the current selection with the new selection", () => {
+            optionSelectionService.updateSelection(fakeOption1);
+
+            expect(optionSelectionService.currentSelection[0]).toBe(fakeOption1);
```
gnomeontherun: Obtain the current selection from the observable in these tests

src/clr-angular/forms/select/select.ts
```diff
@@ -4,24 +4,71 @@
...
     @ContentChild(ClrOptions) options: ClrOptions;
 
-    constructor(private ifOpenService: IfOpenService) {}
+    constructor(private ifOpenService: IfOpenService, private optionSelectionService: OptionSelectionService,
+                private renderer: Renderer2, private domAdapter: SelectDomAdapter) {
+        this.initializeSubscriptions();
+    }
+
+    private initializeSubscriptions(): void {
+        this.optionSelectionService.selectionChanged.subscribe(() => {
```
gnomeontherun: Need to track this subscription and unsubscribe on destroy.

src/clr-angular/forms/select/option.spec.ts
```diff
@@ -50,29 +52,53 @@ export default function(): void {
...
+
+            it("calls to update the selection when an item is clicked", () => {
+                spyOn(optionSelectionService, "updateSelection");
+
+                context.clarityDirective.updateSelectionAndCloseMenu();
+
+                expect(optionSelectionService.updateSelection).toHaveBeenCalled();
+            });
+
+            it("provides a ref to the ElementRef of the option", () => {
```
gnomeontherun: I don't see this used anywhere. Does it have a purpose or is it from trying approaches?

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: fixed

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: I understand the confusion. I started out with thinking that I can code for both multi and single select right now but then realized that clicking on an option again in multi selection toggles the selection, unlike in single selection. For now, I will just code it for single selection and we can handle multi selection after MVP.

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: I did this because if I remember our conversation on Slack correctly, we decided to follow this pattern rather than the `updateSelection` sending the selection over to the `select`, no? We discussed this I think during the forms PR review recently if I am not wrong.

src/clr-angular/forms/select/select.ts
adityarb88: Oh wow. Cant believe I missed this. Thanks!

src/clr-angular/forms/select/option.spec.ts
adityarb88: We do use this in `select.ts`. We need this to clone the `option` in the input when the user makes the selection.

gnomeontherun: Cool, page search did not show it when the file is collapsed... 😕

src/clr-angular/forms/select/providers/option-selection.service.ts
gnomeontherun: We did talk about it, but the decision was to expose the values through the observable instead of storing the property on the service. See here https://github.com/vmware/clarity/blob/master/src/clr-angular/forms/common/providers/ng-control.service.ts#L20

adityarb88: Oh cool. Will update this :-) Thanks @gnomeontherun!

src/clr-angular/forms/select/_select-angular.clarity.scss
```diff
@@ -52,6 +53,19 @@
 
     .clr-option {
         @include generate-dropdown-item();
+
+        white-space: nowrap;
```
youdz: I think this is a problem. Custom options are going to be multi-line, and we don't want them to have to overwrite this every single time. Why do you need it? Can't we simply let the option text wrap?

src/clr-angular/forms/select/utils/select-dom-adapter.ts
```diff
@@ -0,0 +1,26 @@
...
+import {Injectable} from "@angular/core";
+
+// TODO: Shijir is moving the DOM Adapter outside of the Datagrid. While rebase, move the functions
+// from this class to there.
+
+@Injectable()
+export class SelectDomAdapter {
+    clearChildren(element: any): void {
+        while (element.firstChild) {
+            element.removeChild(element.firstChild);
```
youdz: Why not just `element.innerHTML = ''`? That sounds way more efficient.

src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
```diff
@@ -78,12 +79,9 @@
...
 @mixin generate-dropdown-item-height() {
-
-    height: $clr-dropdown-item-height;
-    line-height: $clr-dropdown-item-height;
+    padding: 0.125rem 1rem;
 
     @media screen and (max-width: map-get($grid-breakpoints, sm)) {
-        height: $clr-dropdown-item-height-sm;
-        line-height: $clr-dropdown-item-height-sm;
+        padding: 0.25rem 1rem;
```
youdz: Shouldn't these be variables?

And I can't really check, but if the `$clr-dropdown-item-height` variable isn't used anywhere anymore, you should delete it.

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import {Injectable} from "@angular/core";
+import {Observable, Subject} from "rxjs";
+
+import {ClrOption} from "../option";
+
+@Injectable()
+export class OptionSelectionService {
+    private currentSelection: ClrOption;
```
youdz: I'm sensing a refactor as soon as we hit multi-selection. But I'd rather not have the same mess as datagrid...

Maybe always having an array, and single selection will always be an array of 0 or 1 option?

youdz: I just realized this is not going to work. Yo are using our `ClrOption` object to track the current selection. But you are going to have to provide an two-way binding for this, and the consumer will need to provide the **value** of the option as the input. As in, the model of the option. The only way to do this with the current structure is going to be to iterate over all the options and find the one whose model is the same, and that wouldn't even be possible for lazy-loaded options.

I know it's all features that haven't been implemented yet, but two-way binding is part of the MVP and we need to anticipate it.

The solution is to make all of select parametrized by the consumer's model type, and store that type for the current selection. `ClrSelect<T>`, `ClrOption<T>`, `OptionSelectionService<T>`, and then `private currentSelection: T;`.

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    private _selectionChanged: Subject<ClrOption> = new Subject<ClrOption>();
```
youdz: This would becomes a `Subject<T>`.

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    get selectionChanged(): Observable<ClrOption> {
+        return this._selectionChanged.asObservable();
+    }
+
+    // Currently only handles single selection
+    updateSelection(option: ClrOption): void {
+        if (this.currentSelection && this.currentSelection === option) {
```
youdz: When you move this to a model comparison, can you please add a comment to support trackBy in a later PR? This is going to break when the options come from the server and I pre-select one.

src/clr-angular/forms/select/providers/option-selection.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            optionSelectionService.updateSelection(fakeOption2);
+
+            expect(selectedOption).toBe(fakeOption2);
+
+            subscription.unsubscribe();
+        });
+
+        it("updates the selected flag of the option when it is selected", () => {
+            expect(fakeOption1.selected).toBe(false);
```
youdz: When you move to a model-based selection, the options will need to listen to selection changes to update their own flag.

src/clr-angular/forms/select/providers/option-selection.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            expect(fakeOption1.selected).toBe(false);
+            expect(fakeOption2.selected).toBe(true);
+        });
+
+        it("provides an observable to subscribe to change in option selection", () => {
+            expect(optionSelectionService.selectionChanged).toBeDefined();
+        });
+
+        it("notifies when a selection has been updated", () => {
```
youdz: I'm worried that all of your tests break if this one breaks. Not sure if you can improve on that, but all the tests above already subscribe to this observable and rely on it to check their own stuff. It's okay if you don't change anything, I'm just voicing mild concerns.

src/clr-angular/forms/select/option.ts
```diff
@@ -3,13 +3,26 @@
...
 import {IfOpenService} from "../../utils/conditional/if-open.service";
+import {OptionSelectionService} from "./providers/option-selection.service";
 
 @Component({selector: "clr-option", templateUrl: "option.html", host: {"[class.clr-option]": "true"}})
 export class ClrOption {
-    constructor(private ifOpenService: IfOpenService, @Optional() @Inject(POPOVER_HOST_ANCHOR) parentHost: ElementRef) {
+    private _selected: boolean = false;
+
+    @HostBinding("class.active")
+    get selected(): boolean {
```
youdz: You have a getter and a setter that don't do anything special, we avoid this to keep things simple. Just use a public property:
```typescript
@HostBinding("class.active") selected = false;
```

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -175,5 +189,41 @@ export default function(): void {
                 expect(menus[0].classList.contains("test")).toBe(true);
             });
         });
+
+        describe("Rendering Selected Option", () => {
```
youdz: Didn't we say integration specs go in their own file? 😝

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -175,5 +189,41 @@ export default function(): void {
...
+
+            it("renders the selected option in the input when it is clicked", () => {
+                const options = context.clarityElement.querySelectorAll(".clr-option");
+                let input: HTMLElement = context.clarityElement.querySelector(".clr-select-input");
+
+                expect(input.children.length).toBe(0);
+
+                options[0].click();
+
+                input = context.clarityElement.querySelector(".clr-select-input");
```
youdz: Why query again, you already got the input 4 lines above?

src/clr-angular/forms/select/_select-angular.clarity.scss
adityarb88: I believe this is what we do for dropdowns. Custom options can still be online though. They would be in a separate container of their own. That container by default can wrap unless specified not to by the user, no?

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: I spoke about this to you over chat. We decided to keep it this way and change things when we implement Multiple Selection

adityarb88: Fixed

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: Fixed

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: Done

src/clr-angular/forms/select/select.spec.ts
adityarb88: fixed

src/clr-angular/forms/select/utils/select-dom-adapter.ts
adityarb88: I read here that `element.innerHTML = ''` is not as fast as we think it is: https://stackoverflow.com/a/3955238

src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
adityarb88: The dropdowns CSS doesn't have any variables. We need to fix that but I thought I should do that in a separate PR and not pollute this one.

src/clr-angular/forms/select/providers/option-selection.service.spec.ts
adityarb88: done

src/clr-angular/forms/select/providers/option-selection.service.spec.ts
adityarb88: Sorry but I didn't understand the concern that you had here. The tests in this file have been updated to test the new pattern. Please let me know if you still feel that there is an issue here.

src/clr-angular/forms/select/option.ts
```diff
@@ -3,29 +3,40 @@
...
+                public elRef: ElementRef, private optionSelectionService: OptionSelectionService<T>) {
         if (!parentHost) {
             throw new Error("clr-option should only be used inside of a clr-select");
         }
+        this.initializeSubscription();
+    }
+
+    private initializeSubscription(): void {
+        this.subscription = this.optionSelectionService.valueChanged.subscribe((value: T) => {
+            if (value === null || value === undefined) {
```
adityarb88: I should have added a comment here but if I don't do the null and undefined check and if the users do not assign any value to the options, all options get highlighted.

src/clr-angular/forms/select/_select-angular.clarity.scss
youdz: `white-space` is inherited, so that's what I meant by having to override it. I think that's bad for DX.

src/clr-angular/forms/select/select.ts
```diff
@@ -4,24 +4,73 @@
...
+
+    private initializeSubscriptions(): void {
+        this.subscription = this.optionSelectionService.renderSelectionChanged.subscribe((option: ClrOption<T>) => {
+            this.renderSelection(option);
+        });
+    }
+
+    private renderSelection(selectedOption: ClrOption<T>): void {
+        if (this.input && selectedOption) {
+            this.domAdapter.clearChildren(this.input.nativeElement);
```
Shijir: Looking at this makes me wonder whether `initializeSubscriptions` should be called in AfterViewInit. What if DOM changes after we call clearChildren?

src/clr-angular/forms/select/select.ts
```diff
@@ -4,24 +4,73 @@
...
+            const clone: HTMLElement = this.domAdapter.cloneNode(selectedOption.elRef.nativeElement);
```
Shijir: Have we considered using a dynamic component creation based on the selected option's data? 

Also, if we clone the DOM element, we would need to also remove its old attributes and styles and add its styles to make it work in the input. We would need to manually handle classes/styles/events from now on.

I think the selected option needs its own component. That way, wouldn't it be more Angular appropriate and easy to scale once we add multi-select options?

@youdz what's your take on it?

src/clr-angular/forms/select/select.ts
```diff
@@ -4,24 +4,73 @@
...
+            this.renderer.setAttribute(clone, "contenteditable", "false");
```
Shijir: Do we have a test for this one? _The clone shouldn't have `contenteditable` attribute._

src/clr-angular/forms/select/providers/option-selection.service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import {Injectable} from "@angular/core";
+import {Observable, Subject} from "rxjs";
+
+import {ClrOption} from "../option";
+
+@Injectable()
+export class OptionSelectionService<T> {
+    private currentValue: T;
+    private currentOption: ClrOption<T>;
```
Shijir: Do we need to track `currentOption` here? `currentOption` is based on `currentValue`, right? When `currentValue` changes current option should change...?

src/clr-angular/forms/select/option.ts
```diff
@@ -3,29 +3,40 @@
...
+                this.selected = false;
+            } else if (this.value === value) {
+                this.selected = true;
```
youdz: You don't have an input for the selected option right now, so this works. Maybe put a comment for later that once we add an input, we should call `this.optionSelectionService.renderSelection(this)` here?

src/clr-angular/forms/select/select.ts
adityarb88: The `clearChildren` method though would only be called after the user selects an option. So I believe it wouldn't matter when we subscribe to it.

src/clr-angular/forms/select/select.ts
youdz: We have a "custom selected option template" as part of the MVP, but we want the API for the simple case to be just as simple as a native select, so we have to resort to the cloning trick in that case.

The selected option template directive will come in a future PR, but still before we ship.

src/clr-angular/forms/select/select.ts
adityarb88: Actually `contenteditable="false"` would still have the `contenteditable` attribute. It would just be disabled.

src/clr-angular/forms/select/providers/option-selection.service.ts
adityarb88: So I believe we assume that the option itself is static. The option list can change but once option content wouldn't. Thats why we do not need to track this info and thats why we directly clone the option on the input.

src/clr-angular/forms/select/select.ts
Shijir: Then, it would be still worth testing `contenteditable` has `false` value or not? If we ever modify something here and forget to change this attribute value, that test will fail and tell us why.

adityarb88: Absolutely. Will add that test.

src/clr-angular/forms/select/select.ts
Shijir: I see. In that case, we are safe here.

src/clr-angular/forms/select/option.ts
adityarb88: added a comment

src/clr-angular/forms/select/option.ts
adityarb88: added a comment

src/clr-angular/forms/select/select.ts
adityarb88: added the test. good catch @Shijir

src/clr-angular/forms/select/utils/select-dom-adapter.ts
adityarb88: @youdz: not sure if you read the above link but I am assuming you have and there is no change required here.

src/clr-angular/forms/select/_select-angular.clarity.scss
adityarb88: removed for now but we can discuss later if i decide to add it back.

PR #2398, opened 2018-06-27
[NG] fix layout to not use Object.values since its not ie11 supported

gnomeontherun:
This is basically a polyfill for Object.values(), which IE11 and Node <9 don't support :(

String enums cannot be reverse-mapped, meaning Layouts['COMPACT'] does not return 'compact' so this exists to deal with this little caveat to get the list of the values of a string enum.

Source: https://github.com/tc39/proposal-object-values-entries/blob/master/polyfill.js

### DIFF
--- a/src/clr-angular/forms/common/providers/layout.service.ts
+++ b/src/clr-angular/forms/common/providers/layout.service.ts
@@ -15,12 +15,20 @@ export enum Layouts {
 @Injectable()
 export class LayoutService {
   layout: Layouts = Layouts.VERTICAL;
+  // This is basically a replacement for Object.values(), which IE11 and Node <9 don't support :(
+  // String enums cannot be reverse-mapped, meaning Layouts['COMPACT'] does not return 'compact' so
+  // this exists to deal with this little caveat to get the list of the values as an array.
+  private layoutValues: string[] = Object.keys(Layouts).map(key => Layouts[key]);
 
-  isVertical() {
+  isVertical(): boolean {
     return this.layout === Layouts.VERTICAL;
   }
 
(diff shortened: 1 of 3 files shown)


src/clr-angular/forms/common/providers/layout.service.ts
```diff
@@ -16,11 +16,32 @@ export enum Layouts {
...
+
+  isValid(layout: string): boolean {
+    return this.layoutsAsList().indexOf(layout) > -1;
+  }
+
+  // This is basically a polyfill for Object.values(), which IE11 and Node <9 don't support :(
+  // String enums cannot be reverse-mapped, meaning Layouts['COMPACT'] does not return 'compact' so
+  // this exists to deal with this little caveat to get the list of the values of a string enum.
+  // Source: https://github.com/tc39/proposal-object-values-entries/blob/master/polyfill.js
+  private layoutsAsList(): string[] {
```
Shijir: @gnomeontherun I don't think just to determine `Object.values(Layouts).includes(this.layout)`, we should incorporate external polyfills. This chunk of code looks more like a compiled code. Unless I am overlooking anything, we could do it in simpler and more maintainable ways on our own. One example I could see is here:

```
private layoutValues: string[] = Object.keys(Layouts).map(key => Layouts[key]);

isValid(layout: string): boolean {
    return this.layoutValues.indexOf(layout) > -1;
}
```

gnomeontherun: Thanks, from reading I didn't expect Object.keys to work here either (due to string enum), so this is better.

Issue #1927, opened 2018-02-02
The tooltip “Lorem” for “info-circle” button is revealed only by mouse over and is not available to keyboard-only users in Tooltips tab


mathisscott:
URL: http://clarity-ks.surge.sh/tooltips

Tools: Web Accessibility Toolbar (WAT2.0) 

Reproduced Steps:
1. Login Clarity Kitchen Sink with web browser.
2. Go to Tooltips tab.
3. Use TAB key to check the tooltip of “info-circle” button in Tooltips tab.

Actual Results:
The tooltip “Lorem” for “info-circle” button is revealed only by mouse over and is not available to keyboard-only users in Tooltips tab.
For more details, please refer to "Tooltips-NotAvailableOnscreen.png".

Expected Results:
The tooltip “Lorem” for “info-circle” button is revealed only by mouse over and should be available onscreen in Tooltips tab.

Fix Suggestion:
Add equivalent text on screen and the text is consistent with tooltip.
Or the equivalent text can be displayed with TAB key.

Violation of Section 508 Standards: 
1194.21(a) When software is designed to run on a system that has a keyboard, product functions shall be executable from a keyboard where the function itself or the result of performing a function can be discerned textually. 

User Impact: Blind.

Screen shot:
1.	Tooltips-NotAvailableOnscreen.png

![image.png](https://images.zenhubusercontent.com/582f4cffb9dc3c6220222f57/7fdcee43-22c9-43ba-b25c-76347a0b4327)

mathisscott:
Verified that ANGULAR tooltips are no longer accessible. See website to test it out.

PR #2414, opened 2018-07-02
[NG] Fix tooltip visibility issue on trigger focus

adityarb88:
Fixes: #1927

This is not the perfect solution for this issue but we need to refactor the Angular component. Before refactoring the Angular component, we also need to solidify the tooltip design guidelines. As of now, we recommend tooltips to be used only on actionable icons but UX is looking in other valid use cases too. Until we can finalize that, this PR fixes but it solves the high priority accessibility bug.

Before:
image

image

After:
image

### DIFF
--- a/src/clr-angular/forms/common/layout.ts
+++ b/src/clr-angular/forms/common/layout.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { AfterContentInit, Directive, ElementRef, Input, OnInit, Renderer2 } from '@angular/core';
+import { Directive, Input, OnInit } from '@angular/core';
 import { Layouts, LayoutService } from './providers/layout.service';
 
 @Directive({
--- a/src/clr-angular/popover/tooltip/tooltip-trigger.ts
+++ b/src/clr-angular/popover/tooltip/tooltip-trigger.ts
@@ -6,7 +6,7 @@
 import { Directive, HostListener } from '@angular/core';
 import { IfOpenService } from '../../utils/conditional/if-open.service';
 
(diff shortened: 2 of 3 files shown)


src/clr-angular/forms/common/layout.ts
```diff
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { AfterContentInit, Directive, ElementRef, Input, OnInit, Renderer2 } from '@angular/core';
+import { Directive, Input, OnInit } from '@angular/core';
```
adityarb88: I dont know why `npm run lint:fix` caught this now but not earlier.

src/clr-angular/popover/tooltip/tooltip-trigger.ts
```diff
@@ -6,7 +6,7 @@
 import { Directive, HostListener } from '@angular/core';
 import { IfOpenService } from '../../utils/conditional/if-open.service';
 
-@Directive({ selector: '[clrTooltipTrigger]', host: { '[attr.tabindex]': '0' } })
+@Directive({ selector: '[clrTooltipTrigger]', host: { '[attr.tabindex]': '0', '[class.tooltip-trigger]': 'true' } })
```
Shijir: This class looks like a static class. Why just use `class: "tooltip-trigger"`

adityarb88: I don't do that because webstorm syntax highlighting fails on that and I find it very annoying.

youdz: And more importantly that would override the classes used by the consumer, etrase them to replace them with `tooltip-trigger`.

PR #2415, opened 2018-07-02
[NG] textarea component

gnomeontherun:
This introduces the new clrTextarea component, which handles the form validations for the textarea. It also extracts the control class logic into a service since it was being duplicated.

#1878


src/clr-angular/forms/textarea/textarea.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      it('should handle blur events', () => {
+        // textareas must be both invalid and blurred to register the validity
+        textarea.nativeElement.value = 'abc';
+        textarea.nativeElement.dispatchEvent(new Event('textarea'));
+        textarea.nativeElement.dispatchEvent(new Event('blur'));
+        fixture.detectChanges();
+        expect(ifErrorService.triggerStatusChange).toHaveBeenCalled();
+      });
+    });
```
youdz: You're missing a test checking that it properly sets `controlClassService.className`.

src/clr-angular/forms/textarea/textarea-container.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+        });
+        fixture = TestBed.createComponent(NoLabelTest);
+
+        textareaContainerDE = fixture.debugElement.query(By.directive(ClrTextareaContainer));
+        textareaContainerEl = textareaContainerDE.nativeElement;
+        layoutService = textareaContainerDE.injector.get(LayoutService);
+      });
+
+      it('adds an empty label when instantiated without vertical layout', () => {
+        layoutService.layout = Layouts.HORIZONTAL;
```
youdz: Maybe we should also test that the label **isn't** here when using a vertical layout?

src/clr-angular/forms/textarea/textarea-container.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+        textareaContainerDE = fixture.debugElement.query(By.directive(ClrTextareaContainer));
+        textareaContainer = textareaContainerDE.componentInstance;
+        textareaContainerEl = textareaContainerDE.nativeElement;
+        ifErrorService = textareaContainerDE.injector.get(IfErrorService);
+        layoutService = textareaContainerDE.injector.get(LayoutService);
+        fixture.detectChanges();
+      });
+
+      it('injects the NgControlService and subscribes in AfterContentInit', () => {
```
youdz: The description is off, it's the `IfErrorService ` there is no `AfterContentInit` involved.

src/clr-angular/forms/textarea/textarea-container.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      it('computes the grid class for the control container when not vertical', () => {
+        expect(textareaContainer.controlClass()).toEqual('');
+        layoutService.layout = Layouts.HORIZONTAL;
+        expect(textareaContainer.controlClass()).toContain('clr-col-xs-12');
+      });
+
+      it('tracks the validity of the form control', () => {
+        let control;
+        const sub = ifErrorService.statusChanges.subscribe(ctrl => (control = ctrl));
```
youdz: I don't understand this subscription, you're not using `control` at all in this test.

src/clr-angular/forms/textarea/textarea.ts
```diff
@@ -0,0 +1,50 @@
...
+      );
+    }
+    if (controlClassService) {
+      controlClassService.className = el.nativeElement.className;
+    }
+  }
+
+  ngAfterContentInit() {
+    if (this.ngControlService) {
+      this.ngControlService.setControl(this.control);
```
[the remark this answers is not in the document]
gnomeontherun: Yeah, I looked into this again and I think there _was_ a reason before the final design and now its not needed. Moving up to constructor.

src/clr-angular/forms/textarea/textarea.spec.ts
gnomeontherun: Added

src/clr-angular/forms/textarea/textarea-container.spec.ts
gnomeontherun: Removed

src/clr-angular/forms/textarea/textarea-container.spec.ts
gnomeontherun: Fixed

src/clr-angular/forms/textarea/textarea-container.spec.ts
gnomeontherun: Done

gnomeontherun:
@youdz I decided to consolidate the tests for input and textarea because they are practically identical. So the changeset is larger but reduces overall code (and will be better for future components).

Issue #2375, opened 2018-06-19
Spinner button remains active after state change when [disabled] is true


RobArbor:
**Select one ...** (check one with "x")

```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Spinner button should remain disabled when `[disabled]` is `true`.

### Actual behavior

After state change spinner button becomes active.

### Reproduction of behavior
https://stackblitz.com/edit/clarity-light-theme-v012-2ohocu

Steps to reproduce:
1. type some text
2. click save

### Environment details

* **Angular version:** 6.0.0

* **Clarity version:** 0.12.0

* **OS and version:** Windows 10

* **Browser:** Chrome 67 | Firefox 60

adityarb88:
@RobArbor: We will take a look and get back to you. Thanks!

PR #2421, opened 2018-07-03
[ng] Bug fix on spinner button to work with [disabled] input (#2375)

jeeyun:
- Now spinner button respects the [disabled] input after state change
- Using `getBoundingClientRect` instead of `getComputedStyle` to set the width of the button

### DIFF
--- a/src/clr-angular/button/button-loading/loading-button.ts
+++ b/src/clr-angular/button/button-loading/loading-button.ts
@@ -5,7 +5,7 @@
  */
 
 import { animate, keyframes, style, transition, trigger } from '@angular/animations';
-import { Component, ElementRef, EventEmitter, Output, Renderer2 } from '@angular/core';
+import { Component, ElementRef, EventEmitter, Input, Output, Renderer2 } from '@angular/core';
 import { ClrLoadingState } from '../../utils/loading/loading';
 import { LoadingListener } from '../../utils/loading/loading-listener';
 
@@ -17,7 +17,7 @@ import { LoadingListener } from '../../utils/loading/loading-listener';
                 <span @spinner class="spinner spinner-inline"></span>
             </span>
             <span *ngSwitchCase="buttonState.SUCCESS">
-                <span @validated class="spinner spinner-inline spinner-check"></span>
+                <span @validated (@validated.done)="this.loadingStateChange(this.buttonState.DEFAULT)" class="spinner spinner-inline spinner-check"></span>
             </span>
             <span *ngSwitchCase="buttonState.DEFAULT" @defaultButton>
                 <ng-content></ng-content>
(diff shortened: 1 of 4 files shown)


src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -38,61 +38,64 @@ import { LoadingListener } from '../../utils/loading/loading-listener';
...
     }
-
     this.clrLoadingChange.emit(state);
   }
 
   private setExplicitButtonWidth() {
-    if (getComputedStyle) {
-      const width = getComputedStyle(this.el.nativeElement).getPropertyValue('width');
-      this.renderer.setStyle(this.el.nativeElement, 'width', width);
+    if (this.el.nativeElement.getBoundingClientRect) {
```
adityarb88: Would this work when the server side rendering script runs?

adityarb88: Also I believe the `dom-adapter.ts` already has this, no?

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -38,61 +38,64 @@ import { LoadingListener } from '../../utils/loading/loading-listener';
...
         animate(
-          '300ms',
+          '600ms',
           keyframes([
-            style({ transform: 'scale(0,0)' }),
-            style({ opacity: 1 }),
-            style({ transform: 'scale(1.2,1.2)' }),
-            style({ transform: 'scale(.9,.9)' }),
-            style({ transform: 'scale(1,1)' }),
+            style({ transform: 'scale(0,0)', offset: 0 }),
```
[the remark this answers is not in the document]
jeeyun: Before, I was using `setTimeout`to change the animation back to default button. Now it's relying only on the animation and the animation callback to achieve this.

src/clr-angular/button/button-loading/loading-button.ts
jeeyun: Good point. I think `dom-adapter.ts` returns the integer value of the width though, which is going to be less precise, and I'm wondering if that's going to shrink the button a tiny bit and result in a tiny shrink/expand of the button. I also am not sure if we should use it outside of datagrid (I think it could be, just don't see an example of us using it so far outside of it)?

As for the server side rendering, I *think* `elementRef` works fine but will defer to @gnomeontherun to confirm if the above code is safe for SSR.

gnomeontherun: This would fail if someone set the default state to `SUCCESS`, which seems unlikely but possible. At the least, you would need to check if `nativeElement` is set in this conditional for SSR.

jeeyun: You mean if they started off the button state as `SUCCESS`? That is true, but not something we can accommodate as there would be no way to measure the width with the text without having rendered that first.

Will update to check if `this.el.nativeElement` exists for SSR.

src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -38,61 +38,64 @@ import { LoadingListener } from '../../utils/loading/loading-listener';
...
       transition(':leave', [style({ opacity: 1 }), animate('100ms ease-out', style({ opacity: 0 }))]),
     ]),
   ],
+  host: { '[attr.disabled]': "disabled? '' : null" },
 })
 export class ClrLoadingButton implements LoadingListener {
   public buttonState = ClrLoadingState;
   public state: ClrLoadingState = ClrLoadingState.DEFAULT;
 
+  @ContentChild(ClrLoading) public loadingDirective: ClrLoading;
```
youdz: `@ContentChild` is not what you want to do here. The loading directive and this component are on the same host, so you want to inject it directly in the constructor.

But I just realized you are not using this `loadingDirective` anywhere, did you forgot to remove it?

jeeyun: When I directly invoke the loadingStateChange method in the template, I see the change emit twice. Is this the intended behavior or am I doing something wrong?

PR #2422, opened 2018-07-03
[NG] clrDroppable and more tests

Shijir:
- This PR is for the `clrDroppable` directive and its tests.
- This PR also includes an integrated tests for D'nD  directives and components.


src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+            if (typeof this._group === "string") {
+                return this._group === draggableGroup;
+            } else {
+                return this._group.indexOf(draggableGroup) > -1;
+            }
+        } else {
+            if (typeof this._group === "string") {
+                return draggableGroup.indexOf(this._group) > -1;
+            } else {
+                return (this._group as string[]).filter(groupKey => draggableGroup.indexOf(groupKey) > -1).length > 0;
```
youdz: Nitpick: `array.filter(_ => condition).length > 0` is just `array.some(_ => condition)`

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    private isInDropArea(point: {pageX: number; pageY: number}): boolean {
+        if (!point) {
+            return false;
+        }
+
+        if (!this.clientRect) {
+            this.clientRect = this.domAdapter.clientRect(this.droppableEl);
```
youdz: I think there's a bug here. If I drag an element, stop dragging, then something changes on my page (content added, deleted, ...) `this.clientRect` will still be set to the old one and the computations will be off.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+            });
+        }
+    }
+
+    private onDragMove(dragMoveEvent: ClrDragEvent<T>): void {
+        if (!this._isDraggableOver && this.isInDropArea(dragMoveEvent.dropPointPosition)) {
+            this.isDraggableOver = true;
+            const dragEnterEvent = this.transformAndBroadcastEvent(dragMoveEvent, ClrDragEventType.DRAG_ENTER);
+            this.dragEnterEmitter.emit(dragEnterEvent);
+        } else if (this._isDraggableOver && !this.isInDropArea(dragMoveEvent.dropPointPosition)) {
```
youdz: Nitpick: you should just call `this.isInDropArea(dragMoveEvent.dropPointPosition)` at the beginning of this method and use the result for both `if`s. Right now you're caching the `clientRect` so the extra computation is almost nothing, but I'm scared we might make it heavier in the future and this is a free optimization.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+        this.dragMoveEmitter.emit(dragMoveEvent);
+    }
+
+    private onDragEnd(dragEndEvent: ClrDragEvent<T>): void {
+        if (this._isDraggableOver) {
+            if (dragEndEvent.ghostElement) {
+                // By this point draggable ghost component would have been destroyed,
+                // but the element would be active until its animation completes.
+                // As such, once the ghost is dropped over, we will give it "dropped" class.
+                this.renderer.addClass(dragEndEvent.ghostElement, "dropped");
```
youdz: I believe this should be in the ghost component, not here. You are broadcasting a DROP event, so you can have the ghost just listen for it and cancel the animation/disappear when it receives the event. You wouldn't have to pass around the native element for the ghost anymore.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+        this.dragEndEmitter.emit(dragEndEvent);
+        this.unsubscribeFrom(this.dragMoveSubscription);
+        this.unsubscribeFrom(this.dragEndSubscription);
+        this.isDraggableMatch = false;
+        delete this.clientRect;
+    }
+
+    private transformAndBroadcastEvent(event: ClrDragEvent<T>, newEventType: ClrDragEventType): ClrDragEvent<T> {
+        event.type = newEventType;
```
youdz: Wait, you're modifying the event in place! Isn't the event object shared with everyone? I don't know why you don't have bugs with this, maybe I'm missing something.
Anyway, I'm positive you should do:
```typescript
newEvent = {...event, type: newEventType}
this.eventBus.broadcast(newEvent);
```

src/clr-angular/utils/drag-and-drop/droppable/droppable.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+            expect(this.testComponent.dragMoveEvent).toEqual(dragMoveEvent);
+            expect(this.testComponent.dragEnterEvent).toEqual(dragMoveEvent);
+            this.eventBus.broadcast(dragEndEvent);
+            this.fixture.detectChanges();
+            expect(this.droppable.nativeElement.classList.contains("draggable-match")).toBeFalsy();
+            expect(this.droppable.nativeElement.classList.contains("draggable-over")).toBeFalsy();
+            expect(this.testComponent.dropEvent).toEqual(dragEndEvent);
+        });
+
+        it("should remove both draggable-match and draggable-over classes on drop", function() {
```
youdz: These unit tests are getting really, really big but I'm not sure what we can do about it.

Anyway, this one and the previous one have the same description and are indeed testing a lot of the same thing, but they have small differences (like testing the `.dropped` class). Can you please update the tests descriptions?

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+                // but the element would be active until its animation completes.
+                // As such, once the ghost is dropped over, we will give it "dropped" class.
+                this.renderer.addClass(dragEndEvent.ghostElement, "dropped");
+            }
+            const dropEvent = this.transformAndBroadcastEvent(dragEndEvent, ClrDragEventType.DROP);
+            this.dropEmitter.emit(dropEvent);
+            this.isDraggableOver = false;
+        }
+        this.dragEndEmitter.emit(dragEndEvent);
+        this.unsubscribeFrom(this.dragMoveSubscription);
```
gnomeontherun: Now I see why you have individual subscriptions. You could still put them on a map

```
this.subscriptions = {
  dragMove: null,
  dragEnd: null,
};
```

And later you can just do something like this in ondestroy.

```
Object.values(this.subscriptions).forEach(sub => { 
  if (sub) {
    sub.unsubscribe();
  }
});
```

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+    @Output("clrDragEnter") dragEnterEmitter: EventEmitter<ClrDragEvent<T>> = new EventEmitter();
+    @Output("clrDrop") dropEmitter: EventEmitter<ClrDragEvent<T>> = new EventEmitter();
+
+    private unsubscribeFrom(subscription: Subscription): void {
+        if (subscription) {
+            subscription.unsubscribe();
+        }
+    }
+
+    private checkGroupMatch(draggableGroup: string|string[]): boolean {
```
gnomeontherun: This logic is not too intuitive to me, could you annotate a little bit about these conditionals?

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+                private renderer: Renderer2) {
+        this.droppableEl = this.el.nativeElement;
+    }
+
+    private isDraggableMatch: boolean = false;
+    private _isDraggableOver: boolean = false;
+
+    set isDraggableOver(value: boolean) {
+        if (value) {
+            this.renderer.addClass(this.droppableEl, "draggable-over");
```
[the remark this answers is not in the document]
gnomeontherun: In that case, the setter becomes redundant and can be just a single private property.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+import {DomAdapter} from "../../dom-adapter/dom-adapter";
+
+import {ClrDragEvent, ClrDragEventType} from "../interfaces/drag-event";
+import {ClrDropTolerance} from "../interfaces/drop-tolerance";
+import {ClrDragAndDropEventBus} from "../providers/drag-and-drop-event-bus";
+
+@Directive({
+    selector: "[clrDroppable]",
+    providers: [DomAdapter],
+    host: {class: "droppable", "[class.draggable-match]": "isDraggableMatch"}
```
gnomeontherun: You're clearing any class from the droppable element, use`[class.droppable]: 'true',`

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Directive, ElementRef, EventEmitter, Input, OnDestroy, OnInit, Output, Renderer2} from "@angular/core";
+import {Subscription} from "rxjs/Subscription";
```
gnomeontherun: Just import from`rxjs`

src/clr-angular/utils/drag-and-drop/draggable/draggable.ts
```diff
@@ -14,8 +14,9 @@ import {
     Injector,
     OnDestroy,
     Output,
-    ViewContainerRef,
+    ViewContainerRef
 } from "@angular/core";
+import {Input} from "@angular/core";
 import {Subscription} from "rxjs/Subscription";
```
gnomeontherun: Just `rxjs`

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: @youdz I delete `this.clientRect` in the `onDragEnd` method. And `this.clientRect` get reassigned newly every time `dragMove` event is fired for the first time. Or am I missing something or misunderstanding what you are saying here?

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
youdz: Actually he's not, I was really surprised by that too. `"class": "droppable"` does not clear, but `"[class]": "'droppable'"` does.

Still, I told him I thought this was a bit too obscure and had a risk of breaking in the future because it is undocumented.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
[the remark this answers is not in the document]
Shijir: Yes, it's due to the deactivated NgZone.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+    selector: "[clrDroppable]",
+    providers: [DomAdapter],
+    host: {class: "droppable", "[class.draggable-match]": "isDraggableMatch"}
+})
+export class ClrDroppable<T> implements OnInit, OnDestroy {
+    private dragStartSubscription: Subscription;
+    private dragMoveSubscription: Subscription;
+    private dragEndSubscription: Subscription;
+
+    private droppableEl: Node;
```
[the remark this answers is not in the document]
Shijir: Changed to `any`

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: I believe that change is valid in Angular 6. But this topic branch is still running in Angular 5. Once I rebase to the latest master, I will make this suggested adjustments.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: I have to turn off this specific subscription in multiple different cases/places. That's why I haven't pushed it into an array of subscriptions.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: CoOoOOol! Thanks :)

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private transformAndBroadcastEvent(event: ClrDragEvent<T>, newEventType: ClrDragEventType): ClrDragEvent<T> {
+        event.type = newEventType;
+        this.eventBus.broadcast(event);
+        return event;
+    }
+
+    ngOnInit() {
+        this.dragStartSubscription = this.eventBus.dragStarted.subscribe((dragStartEvent: ClrDragEvent<T>) => {
```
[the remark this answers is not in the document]
Shijir: I have to turn off this specific subscription in multiple different cases/places. That's why I haven't pushed it into an array of subscriptions.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+        });
+    }
+
+    ngOnDestroy() {
+        this.unsubscribeFrom(this.dragStartSubscription);
+        this.unsubscribeFrom(this.dragMoveSubscription);
+        this.unsubscribeFrom(this.dragEndSubscription);
+        this.isDraggableMatch = false;
+        this.isDraggableOver = false;
+        delete this.clientRect;
```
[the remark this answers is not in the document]
Shijir: Right! Thanks. I removed them.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    ngOnInit() {
+        this.dragStartSubscription = this.eventBus.dragStarted.subscribe((dragStartEvent: ClrDragEvent<T>) => {
+            this.onDragStart(dragStartEvent);
+        });
+    }
+
+    ngOnDestroy() {
+        this.unsubscribeFrom(this.dragStartSubscription);
```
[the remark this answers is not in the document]
Shijir: We have this kind of individual subscriptions in many of our components, especially the components/directives that need only one subscription, etc. But this one has multiple subscriptions and they need to be unsubscribed at specific locations, so that's why I have this unsubscribe method just to make the unsubscription more concise.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: 🤔 Wouldn't this be kind of against our pattern? We actually don't subscribe/unsubscribe anywhere like this. IMO, `this.unsubscribeFrom` is very declartive. Also, `Object.values` is not supported in IE :( But I could still use `Object.keys`. But again, I am pondering if it will make any compelling improvements here.

gnomeontherun: We don't do this anywhere else because we always unsubscribe to all at the end, and this is the only case I think of where you sub/unsub to some while the component is still active. Yeah when I was thinking about this originally I didn't see the use of individual subscriptions on drag events, so I can let this go.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
gnomeontherun: Ok, you should get build errors off master, unless you add the rxjs-compat library (please dont).

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,205 @@
...
+            this.isDraggableOver = false;
+        }
+        this.dragEndEmitter.emit(dragEndEvent);
+        this.unsubscribeFrom(this.dragMoveSubscription);
+        this.unsubscribeFrom(this.dragEndSubscription);
+        this.isDraggableMatch = false;
+        delete this.clientRect;
+    }
+
+    private transformAndBroadcastEvent(event: ClrDragEvent<T>, newEventType: ClrDragEventType): ClrDragEvent<T> {
```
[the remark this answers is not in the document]
Shijir: Yes, I've now removed this method. Eudes suggested another way of doing this.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: This one cannot be done in the ghost component because any subscription to the `drop` event (or `dragEnd`) is ineffective or invalid as the component is already destroyed at this point. The ghost component can only subscribe to the `dragMove` event. I've annotated why it's here though. But also, I am open to other suggestions or improvements.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: Thanks, I haven't thought of this. I used this method in other areas too :)

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,246 @@
...
+    private _group: string|string[];
+    private _dropTolerance: ClrDropTolerance = {top: 0, right: 0, bottom: 0, left: 0};
+
+    @Input("clrGroup")
+    set group(value: string|string[]) {
+        this._group = value;
+    }
+
+    @Input("clrDropTolerance")
+    set dropTolerance(value: number|string|ClrDropTolerance) {
```
Shijir: @gnomeontherun this is the implementation I did as per our discussion on slack. All of the following options will work now.
```
[clrDropTolerance]="10"
clrDropTolerance="10"
clrDropTolerance="10"
clrDropTolerance="10 20"
clrDropTolerance="10 20 30"
```

src/clr-angular/utils/drag-and-drop/droppable/droppable.spec.ts
Shijir: Addressed this.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -7,14 +7,17 @@ export enum ClrDragEventType {
...
     DROP
 }
 
 export interface ClrDragEvent<T> {
-    // TODO: this interface will be expanded and customized as the implementation progresses.
     type: ClrDragEventType;
     group?: string|string[];
     draggableElement: Node;
-    dragPosition: {pageX: number, pageY: number};
+    ghostElement?: Node;
```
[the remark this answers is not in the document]
Shijir: Addressed this issue.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -53,12 +53,42 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
...
         if (typeof value === "number") {
             this._dropTolerance.top = value;
             this._dropTolerance.right = value;
             this._dropTolerance.bottom = value;
             this._dropTolerance.left = value;
+        } else if (typeof value === "string") {
+            const toleranceValues = value.trim().split(/\s+/).map((tolerance) => parseInt(tolerance, 10));
+            switch (toleranceValues.length) {
+                case 1:
```
youdz: Just showing off 😎:
```typescript
const nbValues = toleranceValues.length;
['top', 'right', 'bottom', 'left'].forEach((dir, i) => {
	this._dropTolerance[dir] = toleranceValues[Math.max(i%nbValues, i%2%nbValues)];
});
```

That can replace your entire switch.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
youdz: Yes, you understood what I was saying. I missed the `delete` in `dragEnd`, you're right.
We might still have the issue I described if the page updates **while I'm dragging**, but I'm okay if that's something we don't want to support. Moving objects around as a user is dragging is probably bad practice for a11y.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
youdz: My first intuition would be to just put that code in the ghost component's `ngOnDestroy()`, then. But I realize this wouldn't work since you only want the class to be applied if it was successfully dropped. This is fine for now. 👍

src/clr-angular/utils/drag-and-drop/drag-event-external.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+import {ClrDragEvent} from "./interfaces/drag-event";
+
+// This class is used to convert an internal event
+// to an external event to be emitted.
+export class ClrDragEventExt<T> {
+    public dragPosition: {pageX: number; pageY: number};
+    public group: string|string[];
+    public dragDataTransfer: T;
+    public dropPointPosition: {pageX: number; pageY: number};
```
youdz: Could you please add a quick comment explaining the difference between `dragPosition` and `dropPointPosition`?

EDIT: my bad, I just saw the comment in the private interface.

src/clr-angular/utils/drag-and-drop/drag-event-external.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {ClrDragEvent} from "./interfaces/drag-event";
+
+// This class is used to convert an internal event
+// to an external event to be emitted.
+export class ClrDragEventExt<T> {
```
youdz: Sorry, wrong name. 😁 

The public one needs to be `ClrDragEvent`, and the private one can have any name you want.

src/clr-angular/utils/drag-and-drop/index.ts
```diff
@@ -8,11 +8,11 @@ import {Type} from "@angular/core";
...
-import {ClrDroppable} from "./droppable";
+import {ClrDroppable} from "./droppable/droppable";
 import {ClrIfDragged} from "./if-dragged";
 
 export * from "./draggable/draggable";
-export * from "./droppable";
+export * from "./droppable/droppable";
 export * from "./if-dragged";
 export * from "./drag-handle";
 export * from "./draggable-ghost";
```
youdz: I just realized 2 things:
- We need an Angular module for drag and drop, which you're missing.
- You need to export the public event type (not the private one).

Shijir: >We need an Angular module for drag and drop, which you're missing

```
@NgModule({
    imports: [CommonModule],
    declarations: [DRAG_AND_DROP_DIRECTIVES],
    entryComponents: [ClrDraggableGhost],
    providers: [ClrDragAndDropEventBus],
    exports: [DRAG_AND_DROP_DIRECTIVES]
})
export class ClrDragAndDropModule {}
```
We do have a modul for drag and drop. Or are you talking about something else?

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
Shijir: Update as per our discussion on Slack.

src/clr-angular/utils/drag-and-drop/index.ts
youdz: Oops, I don't know what I was thinking. I forgot it was a partial contribution on an already well-advanced topic branch, sorry about that. My comment on the public event still stands though.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -45,72 +45,39 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
...
-            if (value.bottom) {
-                this._dropTolerance.bottom = value.bottom;
-            }
-            if (value.left) {
-                this._dropTolerance.left = value.left;
-            }
+            // The value could be passed in as {left: 20, top: 30 }
+            // In this case, the rest of the direction properties should be 0.
+            // That's why we initialize properties with 0 first, then override with user's given value.
+            this._dropTolerance = {...this.dropToleranceGenerator(0), ...value};
```
youdz: Cute. 😄

src/clr-angular/utils/drag-and-drop/index.ts
Shijir: No worries :D I added the public event to the `index.ts` file.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -45,72 +45,39 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
...
 
     @Input("clrGroup")
     set group(value: string|string[]) {
         this._group = value;
     }
 
+    private _dropTolerance: ClrDropTolerance = {top: 0, right: 0, bottom: 0, left: 0 };
+
+    private dropToleranceGenerator(top?: any, right?: any, bottom?: any, left?: any): ClrDropTolerance {
+        return {top: top, right: right || top, bottom: bottom || top, left: left || right || top };
```
youdz: Oops, just realized this doesn't work. `"20 0"` to have only vertical tolerance will set everything to 20 because 0 is falsy.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -0,0 +1,213 @@
...
+    private _group: string|string[];
+
+    @Input("clrGroup")
+    set group(value: string|string[]) {
+        this._group = value;
+    }
+
+    private _dropTolerance: ClrDropTolerance = {top: 0, right: 0, bottom: 0, left: 0};
+
+    private dropToleranceGenerator(top?: any, right?: any, bottom?: any, left?: any): ClrDropTolerance {
```
youdz: Ok, here is how you solve this issue:
```typescript
private dropToleranceGenerator(top = 0, right = top, bottom = top, left = right)
```

I ❤️ Typescript.

src/clr-angular/utils/drag-and-drop/droppable/droppable.spec.ts
```diff
@@ -424,6 +424,39 @@ export default function(): void {
...
+           });
+
+        it("can register dragEnter if dropPointPosition is within drop tolerance added as string of '0 0 20'",
+           function() {
+               const tolerance = {top: 0, right: 0, bottom: 20, left: 0};
+               this.testComponent.tolerance = "0 0 20";
+               this.fixture.detectChanges();
+               this.eventBus.broadcast(mockDragStartEventInt);
+               expect(this.testComponent.dragStartEvent).toEqual(mockDragStartEventExt);
+               this.checkDropTolerance(
```
youdz: Nitpick: if I'm not mistaken, you've rewritten the same test 6 times with slightly different parameters and expected results. You could easily just make a function that takes the tolerance string and the expected result object, that would save you dozens of lines of duplicated code.

src/clr-angular/utils/drag-and-drop/interfaces/drag-event.ts
```diff
@@ -7,14 +7,18 @@ export enum ClrDragEventType {
...
     DRAG_MOVE,
     DRAG_END,
+    DRAG_ENTER,
+    DRAG_LEAVE,
     DROP
 }
 
-export interface ClrDragEvent<T> {
-    // TODO: this interface will be expanded and customized as the implementation progresses.
+export interface ClrDragEventInternal<T> {
```
gnomeontherun: If this is the internal event now, then we shouldn't prefix it right? Just `DragEvent` or `DragEventInternal`

Issue #2357, opened 2018-06-13
Keyboard Interaction for a stack view component


dkanev:
[X ] feature request
[X ] enhancement

I would like to request a Keyboard Interaction for the stack view component
Probably they should be something like the on specified by the “WAI-ARIA Authoring Practices 1.1” for Accordion (https://www.w3.org/TR/wai-aria-practices/#accordion)
For example:
The stack view is good to support collapsed and expanded with a keyboard. 

We have a lot of Angular JS components which use the downgrade stack block. We would like if possible the above requested improvements to be made in such a way that to be available in the downgraded stack blocks.

Environment details
Angular version: 4.x.x

Clarity version: 11.0.1

OS and version:**

Browser:all

dkanev:
We have a lot of Angular JS components which uses the downgrade stack block. We would like if possible the above requested improvements to be made in such a way that to be available in downgraded stack blocks.

mathisscott:
Our plan is to enable this via tabindex. This should allow users to interact with stack views via the keyboard.

Issue #1946, opened 2018-02-02
The interactive elements “arrow” icons cannot be accessed by keyboard in Stack Views tab


mathisscott:
URL: http://clarity-ks.surge.sh/stackviews

Reproduced Steps:
1. Login Clarity Kitchen Sink with web browser.
2. Go to Stack Views tab.
3. Use the tab key to access all the interactive elements in Stack Views tab.

Actual Results:
The interactive elements “arrow” icons cannot be accessed by keyboard. 
For more details, please refer to “StackViews-CannotBeAccessed.png”.

Expected Results:
All interactive elements can be accessed or activated by keyboard.

Violation of Section 508 Standards: 
1194.21(a)When software is designed to run on a system that has a keyboard, product functions shall be executable from a keyboard where the function itself or the result of performing a function can be discerned textually.

User Impact: Blind.

Screen shot:
1.  StackViews-CannotBeAccessed.png

![image.png](https://images.zenhubusercontent.com/582f4cffb9dc3c6220222f57/0bb817fe-2a87-4847-b393-09dec7972b75)

PR #2433, opened 2018-07-06
[NG] StackView Accessibility Bug Fixes

adityarb88:
Fixes: #2357, #1946

**Demo:**
stackview demo

**NOTE:** 
1. The static markup cannot be fixed without a breaking change. This PR fixes the accessibility issues in the Angular component. We still have the generated icon in use for the static markup but the angular component uses `clr-icon`.
2. Tested on Chrome, Firefox and Safari. My VM isn't working properly so couldnt test on IE and Edge. 🤞 
3. @youdz: I am pretty sure you wont like the way the tests are structured, but I didn't want to fix that in this PR.

**Fixes:**
1. Adds a `button` `role` to the `.stack-block-label`
2. Adds a `tabindex` to the `.stack-block-label`
3. Opens the stack block on `enter` and `space`
4. Uses `clr-icon` instead of the generated icon.

### DIFF
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -186,10 +186,8 @@ export default function(): void {
 
       it('does not emit selection change twice after a filter is applied', function() {
         let nbChanges = 0;
-        let currentSelection: any;
         selectionInstance.change.subscribe((items: any) => {
           nbChanges++;
-          currentSelection = items;
         });
 
         selectionInstance.selectionType = SelectionType.Multi;
--- a/src/clr-angular/data/stack-view/_stack-view.clarity.scss
+++ b/src/clr-angular/data/stack-view/_stack-view.clarity.scss
@@ -90,6 +90,10 @@
       }
     }
 
+    .stack-block-label {
+      outline: none;
+    }
+
     .stack-block-label,
     .stack-block-content {
       padding: 0.25rem 0.5rem;
@@ -118,6 +122,14 @@
         margin: (0.25rem + $clr-rem-1px) 0.25rem 0 0;
         text-align: center;
       }
+    }
+
+    .stack-block-caret {
+      height: 0.5rem;
+      width: 0.5rem;
+      margin-right: 0.125rem;
+
+      fill: $clr-stack-view-stack-block-caret-color;
     }
 
     .stack-block-content {
@@ -171,6 +183,7 @@
         transition: background-color 0.2s ease-in-out, color 0.2s ease-in-out;
       }
 
+      &.on-focus:not(.stack-block-expanded),
       &:hover:not(.stack-block-expanded) {
         > .stack-block-label,
         > .stack-block-content {
@@ -251,6 +264,22 @@
     .modal & {
       height: 55vh;
       margin-bottom: 0;
+    }
+  }
+
+  //Disable the generated icon without a breaking change for the stackview static markup
+  //Only targets the Angular component as of now. Static will require a breaking change.
+  .stack-view {
+    clr-stack-block.stack-block-expandable {
+      .stack-block-label::before {
+        content: none;
+      }
+    }
+
+    .stack-children clr-stack-block {
+      .stack-block-label {
+        padding-left: 1.5rem;
+      }
     }
   }
 }
--- a/src/clr-angular/data/stack-view/stack-block.spec.ts
+++ b/src/clr-angular/data/stack-view/stack-block.spec.ts
@@ -110,6 +110,96 @@ export default function(): void {
       expect(getBlockInstance(fixture).expandable).toBeTruthy();
     });
 
+    it('displays a caret when the stack block is expandable', () => {
+      fixture = TestBed.createComponent(DynamicBlock);
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.stack-block-caret')).not.toBeNull();
+
+      getBlockInstance(fixture).expandable = false;
+
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.stack-block-caret')).toBeNull();
+    });
+
+    it('toggles the caret direction based on the expandable property', () => {
+      fixture = TestBed.createComponent(DynamicBlock);
+      fixture.detectChanges();
+
+      const caret: HTMLElement = fixture.nativeElement.querySelector('.stack-block-caret');
+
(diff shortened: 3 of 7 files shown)


src/clr-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -186,10 +186,8 @@ export default function(): void {
 
       it('does not emit selection change twice after a filter is applied', function() {
         let nbChanges = 0;
-        let currentSelection: any;
```
adityarb88: I dont know why the linter wakes up all of a sudden and starts to complain on random files.

gnomeontherun: I caught this in another situation, agree its confusing but could be a new version of the linter.

src/clr-angular/data/stack-view/_stack-view.clarity.scss
```diff
@@ -90,6 +90,10 @@
       }
     }
 
+    .stack-block-label {
+      outline: none;
```
Shijir: Hmm, I have UX concern over this as I think it would create focus state inconsistencies over elements/components. Once a user starts clicking the tab key to navigate through elements/components, native focus actually disappears but its hover state will kick in as a focus state.

src/clr-angular/data/stack-view/stack-block.spec.ts
```diff
@@ -110,6 +110,96 @@ export default function(): void {
       expect(getBlockInstance(fixture).expandable).toBeTruthy();
     });
 
+    it('displays a caret when the stack block is expandable', () => {
+      fixture = TestBed.createComponent(DynamicBlock);
```
Shijir: Every unit test is creating its component, why don't we just move these two lines into `beforeEach()`? It doesn't look like it will require significant changes to do that.

src/clr-angular/data/stack-view/_stack-view.clarity.scss
adityarb88: @shijir: Here the hover and the focus states are the same. We do this in a components like the sidenav, subnav, vertical nav etc too.

src/clr-angular/data/stack-view/stack-block.spec.ts
adityarb88: @Shijir: I noted the same exact point in the PR description. I didn't want to pollute this PR with other changes as we are trying to keep the PRs small. What do you think @youdz?

src/clr-angular/data/stack-view/_stack-view.clarity.scss
adityarb88: I should have added a note here though.

Shijir: Wait, we have a native focus for vertical nav and side nav, don't we?

adityarb88: Lets confirm how UX wants this

src/clr-angular/data/stack-view/stack-block.spec.ts
Shijir: Yes. I saw that. That's why I am not blocking this PR. But it seemed it won't pollute this PR much as it's evident that only two exact same lines are repeated in every single test.

adityarb88: I think its more than that. We would have to break the tests into `View` and `Typescript API`. Then we would also have to create a separate file for Integration tests. If we do start to fix things, I would assume we would fix all the above things, no? I don't mind doing this with a separate PR but adding it in this PR is something that I wasn't sure about.

adityarb88: As of now, I followed the same style that other tests in this spec follow.

src/clr-angular/data/stack-view/_stack-view.clarity.scss
adityarb88: discussed with @mathisscott and @Shijir. this is the behavior for now. we can discuss the focus styles later on as a separate change.

src/clr-angular/utils/_theme.dark.clarity.scss
```diff
@@ -624,6 +624,7 @@ $clr-stack-view-stack-block-expanded-bg-color: #324f61;
 $clr-stack-view-stack-block-expandable-hover: #324f61;
 $clr-stack-view-stack-block-content-text-color: #adbbc4;
 $clr-stack-view-stack-block-expanded-text-color: $clr-white;
+$clr-stack-view-stack-block-caret-color: $gray-dark-midtone;
```
youdz: Wait, is this the correct color for the dark theme? I'd expect it to be lighter.

adityarb88: I checked. We are using the same color for the icon. We didn't replace the generated icon in dark theme.

src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -98,4 +110,25 @@ export class ClrStackBlock implements OnInit {
...
+  get role(): string {
+    return this.expandable ? 'button' : null;
+  }
+
+  get tabIndex(): string {
+    return this.expandable ? '0' : null;
+  }
+
+  @HostBinding('class.on-hover')
+  get onHover(): boolean {
```
youdz: Naming is off. I know it's the same styles as the hover one, but the actual behavior here has nothing to do with hover, it's just focus.

src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -9,17 +9,28 @@ import { Component, EventEmitter, HostBinding, Input, OnInit, Optional, Output,
...
-        <!-- FIXME: remove this string concatenation when boolean states are supported -->
-        <div [@collapse]="''+!expanded" class="stack-children">
-            <ng-content select="clr-stack-block"></ng-content>
-        </div>
-    `,
+    <dt class="stack-block-label"
+        (click)="toggleExpand()"
+        (keyup.enter)="toggleExpand()"
+        (keyup.space)="toggleExpand()"
+        (focus)="onStackBlockFocus(true)"
```
youdz: Overkill method, it doesn't even do anything. Just `(focus)="focused = true"`.

src/clr-angular/data/stack-view/stack-block.ts
adityarb88: agreed. fixed.

Issue #2429, opened 2018-07-06
ARIA attributes for a stack view component


dkanev:
**Select one ...** (check one with "x")

```
[X ] feature request
[X] enhancement
```

I would like to request ARIA attributes for the stack view component
Probably they should be something like the on specified by the “WAI-ARIA Authoring Practices 1.1” for Accordion (https://www.w3.org/TR/wai-aria-practices/#accordion)
For example:
The stack view is good to have ARIA attributes when the stack block is collapsed and expanded. 

Also I would like if possible to be provided ARIA roles for this component and its parts.

We have a lot of Angular JS components which use the downgrade stack block. We would like if possible the above requested improvements to be made in such a way that to be available in the downgraded stack blocks.

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 11.0.18

* **OS and version:**

* **Browser:** all

adityarb88:
@dkanev: We have a general issue open to add ARIA attributes to all our components here: https://github.com/vmware/clarity/issues/1190 Please subscribe to https://github.com/vmware/clarity/issues/1190 for more updates. 

@smathis: Can we close this issue as a duplicate of https://github.com/vmware/clarity/issues/1190 or do we want to track this separately?

mathisscott:
@adityarb88 
No, we are keeping these separate so we can prioritize them according to a need of a VMware product. Let's leave them atm. Thanks.

dkanev:
@adityarb88, @mathisscott 
Is it possible this to be merge in 0.11.x version of clarity?
Thanks!

adityarb88:
Yes. This change will be merged in both 0.11 and 0.12

PR #2440, opened 2018-07-09
[NG] Add Stack View aria-expanded attribute

adityarb88:
Fixes: #2429

@youdz mentioned that only the `aria-expanded` attributes are needed because StackView is already semantic as we already use `dd` and `dt`

More info here: https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

### DIFF
--- a/src/clr-angular/data/stack-view/stack-block.spec.ts
+++ b/src/clr-angular/data/stack-view/stack-block.spec.ts
@@ -200,6 +200,30 @@ export default function(): void {
       expect(fixture.nativeElement.querySelector('.stack-block-label').getAttribute('tabindex')).toBeNull();
     });
 
+    it('adds the aria-expanded attribute when the stack block is expandable', () => {
+      fixture = TestBed.createComponent(DynamicBlock);
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.stack-block-label').getAttribute('aria-expanded')).not.toBeNull();
+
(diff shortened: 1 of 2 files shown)


src/clr-angular/data/stack-view/stack-block.spec.ts
```diff
@@ -200,6 +200,30 @@ export default function(): void {
...
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.stack-block-label').getAttribute('aria-expanded')).toBe('false');
+
+      getBlockInstance(fixture).expanded = true;
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.stack-block-label').getAttribute('aria-expanded')).toBe('true');
+    });
+
```
mathisscott: ❤️

src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -131,4 +128,12 @@ export class ClrStackBlock implements OnInit {
   get onStackLabelFocus(): boolean {
     return this.expandable && !this.expanded && this.focused;
   }
+
+  get ariaExpanded(): string {
+    if (!this.expandable) {
+      return null;
+    } else {
+      return this.expanded ? 'true' : 'false';
```
Shijir: This doesn't need to be a string, I believe. So `return this.expanded` could work as well.

PR #2453, opened 2018-07-11
[WEB] Adding hiring banner to latest dox

jeeyun:
Questions:
- Do I need to add this to v11 and v10 dox too? v10 dox already has a warning alert about the Clarity version not being supported
- Is the expectation that the view jobs redirects you to http://vmware.design/career.html? If not, let me know and I can update it.

screen shot 2018-07-11 at 11 41 47 am


mathisscott:
Hi @jeeyun 
Thanks for doing this! We don't want this in the documentation. The home page should be sufficient for now...

mathisscott:
I think the VMware Design jobs page is good for now. But could we add the Sofia dev positions to that page?

Could you contact whoever manages that Careers page to get those added?

jeeyun:
So we don't want it anywhere except on landing page? And yes, will contact the appropriate parties for adding the positions.

mathisscott:
Yup, only on the landing page for now

PR #2473, opened 2018-07-18
[NG] Change locale strings to nominative form

pawelbialaszczyk:
This PR changes the currently used CLDR data strings provided by Angular (month names and day names) from their genitive to nominative form.

The change is motivated by the current (incorrect) form of strings for the Polish locale. For example September is rendered as `września` while it should be `wrzesień`.

The change itself doesn't make a visible difference for most languages (i.e. English, German) but fixes the form for the Polish locale and possibly others.

Signed-off-by: Paweł Białaszczyk <pawelbialaszczyk@gmail.com>

### DIFF
--- a/src/clr-angular/forms/datepicker/providers/locale-helper.service.spec.ts
+++ b/src/clr-angular/forms/datepicker/providers/locale-helper.service.spec.ts
@@ -6,50 +6,158 @@
 
 import { registerLocaleData } from '@angular/common';
 import localeFr from '@angular/common/locales/fr';
+import localePl from '@angular/common/locales/pl';
 
 import { LocaleHelperService } from './locale-helper.service';
 
 registerLocaleData(localeFr);
+registerLocaleData(localePl);
 
 export default function() {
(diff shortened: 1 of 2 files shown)


adityarb88:
Thanks for the contribution. This change makes sense to me. More info here: http://cldr.unicode.org/translation/date-time

`Make sure that the correct forms are provided, especially for the months, and used in the patterns. That is, suppose that the language uses "Dezembro"  for December when standing alone (LLLL), but "Dezembru" when with a date (MMMM, when it means the nth day of that month)`

It looks like when the month is used alone, we should use the Stand-Alone style.

@youdz can you please verify?

pawelbialaszczyk:
Thanks for approving this PR!

In the meantime I've also updated the implementation for abbreviated month names accordingly (forgot last time) and updated test cases. Let me know if more changes are needed.

I've forgot to sign off my last commits. Is it OK to rebase & force push them again?

gnomeontherun:
Yes please rebase and force push into a single commit!

PR #2480, opened 2018-07-20
[DOX] Adding v0.12 stack blitz links to readme

mathisscott:


mathisscott:
Will do. Thanks!

Should I keep the 0.10 links? What are your thoughts on that?

gnomeontherun:
I'd drop them, we aren't doing any more releases for 0.10. I might suggest the same for the readme.

mathisscott:
Thanks, everyone. As soon as I get the golden file working, I'll merge.

PR #2469, opened 2018-07-17
[NG, INCOMPLETE, DONOTUSE] Select 2.0 Move to Master

adityarb88:
**Note: This feature is incomplete. Do not use yet.**

Signed-off-by: Aditya Bhandari <adityab@vmware.com>

PR #2485, opened 2018-07-23
clrSelect

gnomeontherun:
This adds the clrSelect component for standard HTML selects. This conflicts with the #2469 implementation and have talked with @youdz about renaming it to combobox.


src/clr-angular/forms/select/select-container.ts
```diff
@@ -0,0 +1,81 @@
...
+    ngControlService: NgControlService
+  ) {
+    this.subscriptions.push(
+      this.ifErrorService.statusChanges.subscribe(control => {
+        this.invalid = control.invalid;
+      })
+    );
+    this.subscriptions.push(
+      ngControlService.controlChanges.subscribe(control => {
+        this.multi = control.valueAccessor instanceof SelectMultipleControlValueAccessor;
```
youdz: I initially ticked at the fact that `multi`, which is used only for styling, is tied to Angular forms. But then I remembered we actually don't support our forms components to be used without Angular forms, so this is fine.

I don't know, I just felt like posting this. It felt weird to me that styling would break without Angular forms. But no change needed here.

gnomeontherun: The alternative was to inspect the element and check for the multi property. I thought this was a little more elegant and localized.

src/clr-angular/forms/select/select-container.ts
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+    `,
+  host: {
+    '[class.clr-form-control]': 'true',
+    '[class.clr-row]': 'addGrid()',
+  },
+  providers: [IfErrorService, NgControlService, ControlIdService, ControlClassService],
+})
+export class ClrSelectContainer implements DynamicWrapper, OnDestroy {
+  subscriptions: Subscription[] = [];
```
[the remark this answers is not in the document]
gnomeontherun: Updated.

gemini/tests/set3/selects.js
```diff
@@ -11,9 +11,9 @@ gemini.suite('selects', child => {
   child
     .setUrl('/selects')
     .before((actions, find) => {
-      actions.waitForElementToShow('.form-block', WAIT_TIME);
+      actions.waitForElementToShow('.content-area', WAIT_TIME);
```
hippee-lee: Any reason not to make the reference image smaller and more targeted?

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,51 @@
...
+import { WrappedFormControl } from '../common/wrapped-control';
+import { ClrSelectContainer } from './select-container';
+
+@Directive({ selector: '[clrSelect]', host: { '[class.clr-select]': 'true' } })
+export class ClrSelect extends WrappedFormControl<ClrSelectContainer> implements OnInit {
+  constructor(
+    vcr: ViewContainerRef,
+    @Optional() private ngControlService: NgControlService,
+    @Optional() private ifErrorService: IfErrorService,
+    @Optional() private control: NgControl,
```
hippee-lee: Is this really optional? (perhaps I misunderstand `@Optinoal()`) We throw an error below if there is `!control`.

gemini/tests/set3/selects.js
gnomeontherun: This required adding a new selector to the page in another test. I'm not sure why it renders so large, must be part of Gemini?

src/clr-angular/forms/select/select.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  template: `
+    <select clrSelect name="model" class="test-class" [(ngModel)]="model"></select>
+    `,
+})
+class TemplateDrivenTest {}
+
+@Component({
+  template: `
+  <div [formGroup]="example">
+    <select clrSelect name="model" class="test-class" formControlName="model"></select>
```
Shijir: QQ: does it matter it has both `name` and `formControlName`? I am assuming `formControlName` would override the static name property attribute. 

Also, it's `name="model"` in both `TemplateDrivenTest` and `ReactiveTest` templates, which is not binding. Was it for purpose?

src/clr-angular/forms/select/select.ts
```diff
@@ -0,0 +1,51 @@
...
+    @Optional() controlClassService: ControlClassService,
+    renderer: Renderer2,
```
Shijir: QQ: Where are we using Renderer? And some of these properties such as `el`, `vcr` are not declared as `private` for reason?

src/clr-angular/forms/select/select.spec.ts
Shijir: QQ: it has both `name` and `formControlName`. I am assuming it doesn't matter? As `formControlName` overrides static `name="model"`.

Btw, why it's not `[name]="model"` in the template of `TemplateDrivenTest`?. Is it meant to be like that?

src/clr-angular/forms/select/select.ts
Shijir: Are we using Renderer2 in this class? And some of these properties such as `vcr`, `el` are not private.

src/clr-angular/forms/select/select-container.ts
```diff
@@ -0,0 +1,81 @@
...
+import { DynamicWrapper } from '../../utils/host-wrapping/dynamic-wrapper';
+import { ControlIdService } from '../common/providers/control-id.service';
+import { ClrLabel } from '../common/label';
+import { ControlClassService } from '../common/providers/control-class.service';
+
+@Component({
+  selector: 'clr-select-container',
+  template: `    
+        <ng-content select="label"></ng-content>
+        <label *ngIf="!label && addGrid()"></label>
```
Shijir: `hasGrid` would be more appropriate compared to `addGrid`?

src/clr-angular/forms/select/select.ts
gnomeontherun: No Renderer2 is not used so I removed that, but `vcr` and `el` are not available outside of the constructor. Unless you define `public` `protected` or `private` on a constructor property, they are not stored on the class as properties.

src/clr-angular/forms/select/select-container.ts
gnomeontherun: This is the name I've given elsewhere, though its indicative of when the component needs to add the grid (horizontal and compact layouts only).

src/clr-angular/forms/select/select-container.ts
```diff
@@ -0,0 +1,81 @@
...
+        <div class="clr-control-container" [ngClass]="controlClass()">
+            <div [class.clr-multiselect-wrapper]="multi" [class.clr-select-wrapper]="!multi">
```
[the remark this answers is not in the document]
gnomeontherun: Actually the main reason I think is to reduce bindings by one.

PR #2494, opened 2018-07-25
[NG] clrRadio component for new forms

gnomeontherun:
Introduces a new 'wrapper' for the radio which deviates slightly from other controls. Since a radio input is really a set of inputs that resolve to one value, this wrapper helps organize the different options somewhat like select->option works.


src/clr-angular/forms/tests/wrapper.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+        providers: [NgControl, NgControlService, IfErrorService, LayoutService],
+      });
+      fixture = TestBed.createComponent(testComponent);
+
+      containerDE = fixture.debugElement.query(By.directive(testContainer));
+      containerEl = containerDE.nativeElement;
+      layoutService = containerDE.injector.get(LayoutService);
+    });
+
+    it('adds an empty label when instantiated without vertical layout', () => {
```
youdz: I don't understand this test. The empty label for the wrapper (not the container) has nothing to do with layout, it's necessary just to display a radio button. The implementation in the wrapper is correct, but this test has nothing to do with what we actually are testing.

src/dev/src/app/radios/radios.demo.html
```diff
@@ -1,65 +1,279 @@
...
-                <input type="radio" name="gridRadios2" id="checkrads_13" disabled>
-                <label for="checkrads_13">Radio option disabled and unchecked</label>
-            </div>
-            <div class="radio disabled">
-                <input type="radio" name="gridRadios2" id="checkrads_14" disabled checked>
-                <label for="checkrads_14">Radio option disabled and checked</label>
-            </div>
-        </div>
-    </section>
+<!--<h2>CSS Radio</h2>-->
```
youdz: Why is most of this demo commented out?

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
+      <ng-content select="clr-control-error" *ngIf="invalid"></ng-content>
+    </div>
     `,
-  host: { '[class.radio]': 'true' },
-  providers: [ControlIdService],
+  host: {
+    '[class.clr-form-control]': 'true',
+    '[class.clr-row]': 'addGrid()',
+  },
+  providers: [NgControlService, ControlClassService, IfErrorService, NgControlService],
```
youdz: You declare `NgControlService` twice.

src/clr-angular/forms/radio/radio.ts
```diff
@@ -4,17 +4,48 @@
...
+    @Optional() private ifErrorService: IfErrorService,
+    @Optional() private control: NgControl,
+    @Optional() controlClassService: ControlClassService,
+    renderer: Renderer2,
+    el: ElementRef
+  ) {
+    super(ClrRadioWrapper, vcr, 0);
+    if (!control) {
+      throw new Error(
+        'clrRadio can only be used within an Angular form control, add ngModel or formControl to the input'
```
youdz: The hard dependency on Angular forms is actually a problem for Radio and Checkbox. Because they are both entirely rendered in CSS, they need empty labels, matching `id` and `for` attributes, etc., to display correctly. This is a nightmare when generating dynamically many of them for non-form purposes (lists, trees, datagrids, ...) if you **don't** want to import `FormsModule`, which is a perfectly reasonable request.

The thing to note is that in these cases, there is no validation. No helper text. So I believe we really want to make an exception for radios and checkboxes and support them outside of Angular forms too.

We added the Angular components at the time so that people

src/dev/src/app/radios/radios.demo.html
```diff
@@ -1,65 +1,279 @@
...
+<!--<span class="clr-subtext">Helper Text</span>-->
+<!--</div>-->
+<!--</div>-->
+<!--</form>-->
+
+<h2>Angular Radio</h2>
+
+<h4>Vertical</h4>
+
+<label>One <input type="radio" name="one" value="one" [(ngModel)]="vertical.one" /></label>
```
youdz: Wait, this example doesn't display a Clarity radio button. There is no sibling label after the input, so our styles don't apply at all.

Why is this here? Something we want to support? Some old artifact forgotten here? I'm having a hard time understanding what's happening with this demo file overall.

src/dev/src/app/radios/radios.demo.html
```diff
@@ -1,65 +1,279 @@
...
+  </clr-radio-container>
+
+  <p>Selected: {{vertical.two}}</p>
+
+  <clr-radio-container>
+    <label>Third</label>
+    <clr-radio-wrapper>
+      <!--<label>One</label>-->
+      <!-- type="radio" required for angular to pick up this input correctly -->
+      <input clrRadio type="radio" [(ngModel)]="vertical.three" value="one" name="three" required />
```
youdz: This use case doesn't exist. You can't put 3 radio buttons without labels together, it makes no sense for the user.

src/clr-angular/forms/tests/container.spec.ts
```diff
@@ -53,13 +53,16 @@ export function ReactiveSpec(testContainer, testControl, testComponent, wrapperC
...
   describe(description, () => {
     let fixture, containerDE, container, containerEl, ifErrorService, layoutService;
+    if (!Array.isArray(directives)) {
+      directives = [directives];
+    }
     beforeEach(() => {
       TestBed.configureTestingModule({
         imports: [ClrIconModule, ClrCommonFormsModule, FormsModule, ReactiveFormsModule],
-        declarations: [testContainer, testControl, testComponent],
+        declarations: [testContainer, ...directives, testComponent],
```
youdz: You're doing a spread on something of type `any | any[]`. I have no idea what the expected result is here. If you get something of type `any[]`, it's going to spread as expected, but if you get something of type `any`, don't you risk is spreading the actual properties of the object you receive? I'm guessing it works right now since you submitted the PR, I'm just really curious about it.

src/clr-angular/forms/radio/radio.module.ts
```diff
@@ -6,17 +6,19 @@
...
 
 @NgModule({
-  imports: [CommonModule, ClrCommonFormsModule, ClrHostWrappingModule],
-  declarations: [ClrRadio, ClrRadioContainer],
-  exports: [ClrCommonFormsModule, ClrRadio, ClrRadioContainer],
-  entryComponents: [ClrRadioContainer],
+  imports: [CommonModule, ClrCommonFormsModule, ClrHostWrappingModule, ClrIconModule],
+  declarations: [ClrRadio, ClrRadioContainer, ClrRadioWrapper],
+  exports: [ClrCommonFormsModule, ClrRadio, ClrRadioContainer, ClrRadioWrapper],
+  entryComponents: [ClrRadioContainer, ClrRadioWrapper],
```
youdz: Why is `RadioContainer` still an entry component? You don't create it dynamically anymore, unless I missed it.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
-        <ng-content select="[clrRadio]"></ng-content>
-        <ng-content></ng-content>
-        <label *ngIf="_dynamic"></label>
+    <ng-content select="label"></ng-content>
+    <label *ngIf="!label && addGrid()"></label>
+    <div class="clr-control-container" [ngClass]="controlClass()">
+      <ng-content select="clr-radio-wrapper"></ng-content>
+      <ng-content select="clr-control-helper" *ngIf="!invalid"></ng-content>
+      <clr-icon *ngIf="invalid" class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
+      <ng-content select="clr-control-error" *ngIf="invalid"></ng-content>
```
youdz: So the error is per-radio. I know we keep having that discussion, but I really don't remember us going that way for radio buttons. It makes sense for checkboxes, but radios are a single input so it validates as one.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
+    </div>
     `,
-  host: { '[class.radio]': 'true' },
-  providers: [ControlIdService],
+  host: {
+    '[class.clr-form-control]': 'true',
+    '[class.clr-row]': 'addGrid()',
```
Shijir: Are we testing these conditionally applied classes?

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
+  providers: [NgControlService, ControlClassService, IfErrorService, NgControlService],
 })
-export class ClrRadioContainer implements DynamicWrapper {
-  // Indicates whether the container is dynamically created by the radio button itself
+export class ClrRadioContainer implements DynamicWrapper, OnDestroy {
+  subscriptions: Subscription[] = [];
   _dynamic = false;
+  invalid = false;
+  @ContentChild(ClrLabel) label: ClrLabel;
+  controls = [];
```
Shijir: What's this property for? I don't see it being used.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
+    private controlClassService: ControlClassService
+  ) {
+    this.subscriptions.push(
+      this.ifErrorService.statusChanges.subscribe(control => {
+        this.invalid = control.invalid;
+      })
+    );
+  }
+
+  controlClass() {
```
Shijir: The same comment as in the `clrSelect` PR. This method is just calling another method and has no dependency injections. It could be a simple getter property.

src/clr-angular/forms/radio/radio.ts
```diff
@@ -4,17 +4,48 @@
...
-  constructor(vcr: ViewContainerRef) {
-    super(ClrRadioContainer, vcr);
+export class ClrRadio extends WrappedFormControl<ClrRadioWrapper> implements OnInit {
+  constructor(
+    vcr: ViewContainerRef,
+    @Optional() private ngControlService: NgControlService,
+    @Optional() private ifErrorService: IfErrorService,
+    @Optional() private control: NgControl,
+    @Optional() controlClassService: ControlClassService,
+    renderer: Renderer2,
```
Shijir: Are we using `Renderer2` in this class? Also, `vcr` and `el` could be private.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,67 @@
...
+  host: {
+    '[class.clr-form-control]': 'true',
+    '[class.clr-row]': 'addGrid()',
+  },
+  providers: [NgControlService, ControlClassService, IfErrorService, NgControlService],
 })
-export class ClrRadioContainer implements DynamicWrapper {
-  // Indicates whether the container is dynamically created by the radio button itself
+export class ClrRadioContainer implements DynamicWrapper, OnDestroy {
+  subscriptions: Subscription[] = [];
```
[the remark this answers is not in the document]
gnomeontherun: Fixed

src/clr-angular/forms/radio/radio-wrapper.ts
```diff
@@ -0,0 +1,28 @@
...
+import { DynamicWrapper } from '../../utils/host-wrapping/dynamic-wrapper';
+import { ClrLabel } from '../common/label';
+import { ControlIdService } from '../common/providers/control-id.service';
+
+@Component({
+  selector: 'clr-radio-wrapper',
+  template: `
+    <ng-content select="[clrRadio]"></ng-content>
+    <ng-content select="label"></ng-content>
+    <label *ngIf="!label"></label>
```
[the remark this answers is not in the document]
gnomeontherun: Switched back to `_dynamic`

src/clr-angular/forms/tests/wrapper.spec.ts
gnomeontherun: This test should just check the host wrapping is wired up correctly in this case, since its a pure use of the host wrapper. I removed the layout part.

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: Fixed.

src/dev/src/app/radios/radios.demo.html
gnomeontherun: Reverted.

src/dev/src/app/radios/radios.demo.html
gnomeontherun: No, this was something I was tinkering with and forgot to remove. This was a bit of a draft to get some input.

src/dev/src/app/radios/radios.demo.html
gnomeontherun: Nope, again trial and error failure.

src/clr-angular/forms/tests/container.spec.ts
gnomeontherun: Line 59 does a check if `directives` is an array, if not converts it to always be an array. This was a backward compatible API change I made since in the case of radio and checkboxes they will need multiple directives (wrappers and containers) and didn't want to hack all the previous uses. Happy to discuss other approaches, but Shijir's suggestion is a more clever one liner for what I did in lines 59-61.

src/clr-angular/forms/radio/radio.module.ts
gnomeontherun: Right you are.

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: The error is per radio based on the form control, so if you have 3 radios by the same name they share the same error state. I think you have it backwards, having 3 radios each with `name="control"` will show only one validation error, which is most cases I can imagine is only that it is required. Some custom validation might be created where certain radios are invalid given other conditions, but either way the control for all radios inside of the container should be marked as `invalid` and therefore the error messages should appear.

I capture the last declared radio for their validity, because Angular updates all radio controllers with the same validity state and I only need one of them to do the checks. Checkboxes are a different beast, which is partially why I started with radios.

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: Axed!

src/clr-angular/forms/radio/radio-container.ts
youdz: Oh I misread this, my bad.

👍

src/clr-angular/forms/radio/all.spec.ts
```diff
@@ -3,17 +3,13 @@
...
- * The goal is to have the tests properly grouped in the reporter, instead of having them all
- * over the place because we load them asynchronously.
- */
 
 import RadioContainerSpecs from './radio-container.spec';
 import RadioSpecs from './radio.spec';
+import RadioWrapperSpecs from './radio-wrapper.spec';
 
-describe('Radio component', function() {
+fdescribe('Radio component', function() {
```
youdz: `fdescribe`

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -30,14 +32,31 @@ import { NgControlService } from '../common/providers/ng-control.service';
...
   invalid = false;
   @ContentChild(ClrLabel) label: ClrLabel;
-  controls = [];
+  private inline = false;
+
+  /*
+   * Here we want to support the following cases
+   * clrInline - true by presence
+   * clrInline="true|false" - unless it is explicitly false, strings are considered true
+   * [clrInline]="true|false" - expect a boolean
```
youdz: 👍 
So exhaustive.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,86 @@
...
+   */
+  @Input()
+  set clrInline(value: boolean | string) {
+    if (!isBoolean(value)) {
+      this.inline = value === 'false' ? false : true;
+    } else {
+      this.inline = !!value;
+    }
+  }
+  get clrInline() {
```
Shijir: Nit: Do we need `clr` prefix here for this internal property? I think It's inconsistent with what we do currently. If we use setter/getter, we make the private property have an underscore and expose it with getter without one instead of prefixing the getter with `clr`.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -4,23 +4,86 @@
...
+
+  addGrid() {
+    if (this.layoutService && !this.layoutService.isVertical()) {
+      return true;
+    }
+    return false;
+  }
+
+  ngOnDestroy() {
+    if (this.subscriptions) {
```
Shijir: This if check is redundant. 😛

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: I did this because the `clrInline` is the name of the Input, and the `inline` property is private so I needed something. I thought it was a little different from standard setter/getter examples because clrInline is now a property of the component that gets exposed publicly, and I didn't want to confuse with a second property of the same name.

PR #2495, opened 2018-07-25
Trap focus inside doc

### LINKED ISSUE #2396: Modals are losing focus when user presses tab
**Select one ...** (check one with "x")

```
[x] bug
[x] enhancement
```

### Expected behavior
I would like to request the following functionality:  
when I open a Clarity Modal and navigatе with the Tab button in it and when I focus out the last element in the modal the first focusable element of the modal should be focused like this example shows https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

### Actual behavior
When I focus out the last element of the modal the address bar is focused

EDIT by @youdz: I'm copy-pasting your detailed explanation here, because you nailed it and I want this to be visible immediately without reading all the comments. 😄 The important part for reproduction is that the modal or wizard needs to be **the last element in the page**.

> According to my observations if the wizard is the last focusable element in the DOM it happens all the time. This is a bug in the focus trap directive which you are using to keep the focus in the wizard. This is what happens according to my observations:
>
>1. When you press tab on the last button of the wizard the focus goes away from the wizard
>2. The browser tries to focus the next focusable element in the DOM, but since there is no such element the focus goes to the first in the browser , which is the address bar
>3. The focus trap directive does not catch the situation, because it listens the DOM three for focusin event which does not happen because the address bar is outside the DOM. on the next tab , when the focus enters inside the DOM(which may not be exactly the next one, if there are more controls on the browser), the focus trap catches the event.

### Reproduction of behavior
Go to https://clarity-light-theme-v11-9rttaf.stackblitz.io/ and click the Open Modal button
A modal will show
Start pressing tab button. After focus is removed from the OK button, the address bar of the browser is focused

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 0.11.21

* **OS and version:** OSX 10.12.6 | Windows 10

* **Browser:** [Chrome 67.0.3396.99 | Firefox 60.0.2 | IE 11 ]

Shijir:
Fixes #2396 

This solution adds offscreen tabbable dummy elements to document as the first and the last child elements. Multiple nested focus-trapper directives will share the same focus rebounding elements.

Tested on: Chrome, Safari, Firefox.

### DIFF
--- a/src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
+++ b/src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
@@ -3,7 +3,7 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, ViewChild } from '@angular/core';
+import { Component, DebugElement, ViewChild } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormsModule } from '@angular/forms';
 import { By } from '@angular/platform-browser';
@@ -12,7 +12,6 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { itIgnore } from '../../../../tests/tests.helpers';
 import { ClrModal } from '../../modal/modal';
 import { ClrModalModule } from '../../modal/modal.module';
-
 import { FocusTrapDirective } from './focus-trap.directive';
 import { ClrFocusTrapModule } from './focus-trap.module';
 
@@ -20,21 +19,26 @@ describe('FocusTrap', () => {
   let fixture: ComponentFixture<any>;
   let compiled: any;
   let component: TestComponent;
-  let directive: FocusTrapDirective;
+
   let lastInput: HTMLElement;
-  const tabEvent = { shiftKey: false, keyCode: 9, preventDefault: () => {} };
+
+  let directiveDebugElement: DebugElement;
+  let directiveInstance: FocusTrapDirective;
+  let directiveElement: HTMLElement;
 
   describe('default behavior', () => {
     beforeEach(() => {
       TestBed.configureTestingModule({ imports: [ClrFocusTrapModule], declarations: [TestComponent] });
-
       fixture = TestBed.createComponent(TestComponent);
-      component = fixture.componentInstance;
       fixture.detectChanges();
-      compiled = fixture.nativeElement;
-      directive = fixture.debugElement.query(By.directive(FocusTrapDirective)).injector.get(FocusTrapDirective);
 
+      component = fixture.componentInstance;
+      compiled = fixture.nativeElement;
       lastInput = compiled.querySelector('#last');
+
+      directiveDebugElement = fixture.debugElement.query(By.directive(FocusTrapDirective));
+      directiveElement = directiveDebugElement.nativeElement;
+      directiveInstance = directiveDebugElement.injector.get(FocusTrapDirective);
     });
 
     afterEach(() => {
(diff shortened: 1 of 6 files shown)


src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -13,50 +13,99 @@ import {
...
     this.focusTrapsTracker.current = this;
   }
 
   @HostListener('document:focusin', ['$event'])
   onFocusIn(event: any) {
-    const nativeElement: HTMLElement = this.elementRef.nativeElement;
+    const nativeElement: HTMLElement = this.el.nativeElement;
 
-    if (this.focusTrapsTracker.current === this && !nativeElement.contains(event.target)) {
+    if (this.focusTrapsTracker.current === this && event.target && !nativeElement.contains(event.target)) {
```
gnomeontherun: In what case is `event.target` not set?

Shijir: It's mostly for testing cases. We had a false positive test where we created a custom event, but we didn't specify `its event.target`. 
Once, that custom event is passed into `onFocusIn` method, `nativeElement.contains(even.target)` was returning `false` as of course event.target is `undefined` there. As such it satisfies this conditional check:

```
if (this.focusTrapsTracker.current === this && event.target && !nativeElement.contains(event.target))
```

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -13,50 +13,99 @@ import {
...
+    if (isPlatformBrowser(this.platformId)) {
+      this.previousActiveElement = <HTMLElement>this.document.activeElement;
+      this.renderer.setAttribute(this.el.nativeElement, 'tabindex', '0');
+
+      this.addReboundEls();
     }
   }
 
   ngOnDestroy() {
+    if (isPlatformBrowser(this.platformId)) {
```
gnomeontherun: I like the idea of putting the platform check inside of the method that actually uses the body or nativeElement directly, instead of in a higher caller, can we move that around?

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
```diff
@@ -12,50 +12,101 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
...
-      expect(element.getAttribute('tabindex')).toEqual('0');
+      expect(directiveElement.getAttribute('tabindex')).toEqual('0');
     });
 
-    it(`should focus on trappable element when tab key is pressed and last input is active`, () => {
-      const element = directive.elementRef.nativeElement;
-      lastInput.focus();
-      directive.onFocusIn(tabEvent);
-      expect(document.activeElement).toEqual(element);
+    it('should add its off-screen focus rebounder elements to parent element on instantiation', () => {
```
gnomeontherun: Description is misleading, it doesn't test that rebounder elements are in the parent. Can you update the description, and then have a second test to verify that it adds the rebounders as the first and last child of body?

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
```diff
@@ -12,50 +12,101 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
...
+      afterRebound.focus();
+      expect(document.activeElement).toBe(directiveElement);
+    });
+
+    it('should remove its off-screen focus rebounder elements from parent element on removal', () => {
+      let offScreenEls = document.body.querySelectorAll('span.offscreen-focus-rebounder');
+      expect(offScreenEls.length).toBe(2);
+      component.mainFocusTrap = false;
+      fixture.detectChanges();
+      offScreenEls = offScreenEls = document.body.querySelectorAll('span.offscreen-focus-rebounder');
```
gnomeontherun: Double assignment, but nitpick on this line and line 76-77.
```
expect(document.body.querySelectorAll('span.offscreen-focus-rebounder').length).toBe(0);
```

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -13,50 +13,99 @@ import {
...
+    // 2. We should add them more than once. Hence, we are counting a number of focus trappers
+    //    and only add on the first focus trapper.
+
+    if (this.focusTrapsTracker.nbFocusTrappers === 1) {
+      this.topReboundEl = this.createFocusableOffScreenEl('span', 'before-focus-trap');
+      this.bottomReboundEl = this.createFocusableOffScreenEl('span', 'after-focus-trap');
+
+      // Add reboundBeforeTrapEl to the document body as the first child
+      const bodyFirstChild = this.document.body.firstChild;
+      if (bodyFirstChild) {
```
gnomeontherun: This will always be true, since the body has to have at least one element for Angular. Don't need to capture it as a variable I think.

```
this.renderer.insertBefore(this.document.body, this.topReboundEl, this.document.body.firstChild);
```

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -13,50 +13,99 @@ import {
...
+
+  private addReboundEls() {
+    // We will add these focus rebounding elements only in the following conditions:
+    // 1. It should be running inside browser platform as it accesses document.body element
+    // 2. We should add them more than once. Hence, we are counting a number of focus trappers
+    //    and only add on the first focus trapper.
+
+    if (this.focusTrapsTracker.nbFocusTrappers === 1) {
+      this.topReboundEl = this.createFocusableOffScreenEl('span', 'before-focus-trap');
+      this.bottomReboundEl = this.createFocusableOffScreenEl('span', 'after-focus-trap');
```
youdz: You never actually use the `before-focus-trap` and `after-focus-trap` classes here. Only in unit tests, but even that part of the unit test isn't checking anything relevant. So I would remove them and remove the `optionalClass` argument of the creation method. This is all private and used only by the focus trap, let's keep it simple.

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
```diff
@@ -12,50 +12,101 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
...
 
     it('should create directive', () => {
-      expect(directive).toBeTruthy();
+      expect(directiveInstance).toBeTruthy();
     });
 
     it('should add tabindex attribute with value zero', () => {
-      directive.ngAfterViewInit();
-      const element: HTMLElement = directive.elementRef.nativeElement;
-      expect(element.getAttribute('tabindex')).toEqual('0');
```
youdz: Thanks for cleaning that one up, I don't remember why it was manually doing that.

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
```diff
@@ -148,7 +199,7 @@ describe('FocusTrap', () => {
 
 @Component({
   template: `
-        <a href="#">Not in form</a>
+        <ng-container *ngIf="mainFocusTrap">
```
youdz: Nitpick: You don't need the `ng-container`, you can just put the `*ngIf` on the host of the focus trap directly.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: You are right. I will update this. Thanks

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -13,50 +13,99 @@ import {
...
       nativeElement.focus();
     }
   }
 
-  ngAfterViewInit() {
-    if (isPlatformBrowser(this.platformId)) {
-      this._previousActiveElement = <HTMLElement>document.activeElement;
-      const nativeElement: HTMLElement = this.elementRef.nativeElement;
-      nativeElement.setAttribute('tabindex', '0');
+  private createFocusableOffScreenEl(nodeName: string, optionalClass?: string): any {
```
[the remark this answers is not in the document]
Shijir: Removed all argument passings now.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: Addressed this issue.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: Done

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
Shijir: Addressed this.

src/clr-angular/utils/focus-trap/focus-trap.directive.spec.ts
Shijir: You're right. I've updated this as you suggested.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: Now checking inside the method.

Issue #2431, opened 2018-07-06
ARIA attributes for a datagrid component


dkanev:
**Select one ...** (check one with "x")

```

[X ] feature request
[X ] enhancement
```

### Expected behavior
I would like to request ARIA roles, attributes and states for the clarity datagrid.
https://www.w3.org/TR/wai-aria-practices/#grid

The selected or expanded row is good to have appropriate attributes.
The cells to be map to the column and rows.
The expanded rows to have appropriate attributes. 
The datagrid to be recognizable by screen readers.

Environment details

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:** 11.0.18

* **OS and version:**

* **Browser:** all

adityarb88:
@dkanev: Thanks for filing the issue. We have a general issue open to add ARIA attributes to all our components here: https://github.com/vmware/clarity/issues/1190 Please subscribe to https://github.com/vmware/clarity/issues/1190 for more updates. 

@smathis: Can we close this issue as a duplicate of https://github.com/vmware/clarity/issues/1190 or do we want to track this separately?

mathisscott:
@adityarb88 
No, we are keeping these separate so we can prioritize them according to a need of a VMware product. Let's leave them atm. Thanks.

hippee-lee:
Recommendations for ARIA attributes on the `ClrDatagrid` component

## Easy Additions
- Add role="grid" to the `clr-datagrid` component
- Add role="row" to the `clr-dg-row` component
- Add role="columnheader" to the `clr-dg-column` component
- Add role="gridcell" to the `clr-dg-cell` component

## Complex Additions
- Add `aria-tabindex` to cells to enable arrow key navigation 
- Add `aria-rowindex` to the respective `clr-dg-row`'s 
- Add `aria-colindex` to the respective `clr-dg-column`'s and `clr-dg-cell`'s

## Others To Consider
- Add `aria-sort="ascending|descending|none" to `clr-dg-columns`
-  Expose an optional `@Input` for `aria-labelledby="grid1Label"` at the `clr-datagrid` level for users
- Add `aria-haspopup="true"` to the filters and hideable column toggler buttons
- Add `aria-rowcount="COUNT" to the `clr-datagrid` component
- Add `aria-colcount="COUNT"` to the `clr-datagrid-component

[...]

PR #2501, opened 2018-07-27
[ng] add ARIA attributes to datagrid (#2431)

jeeyun:
Highlights:
- Datagrid is now accessible by screen reader
- No longer using HostListener on click or keypress.
- We are wrapping the datagrid row's content inside a label for full row click mode

preview of demo-app with the implementation: http://dg-aria.surge.sh/datagrids

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-cell.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-cell.spec.ts
@@ -43,6 +43,10 @@ export default function(): void {
       context.detectChanges();
       expect(context.clarityElement.classList.contains('datagrid-signpost-trigger')).toBeTruthy();
     });
+
+    it('adds a11y roles to the cell', function() {
+      expect(context.clarityElement.attributes.role.value).toBe('cell');
+    });
   });
 }
 
--- a/src/clr-angular/data/datagrid/datagrid-column.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.spec.ts
@@ -346,6 +346,20 @@ export default function(): void {
         expect(context.clarityElement.classList.contains('desc')).toBeTruthy();
       });
 
+      it('adds a11y roles to the column', function() {
+        expect(context.clarityElement.attributes.role.value).toEqual('columnheader');
+        expect(context.clarityElement.attributes['aria-sort'].value).toBe('none');
+
+        context.clarityDirective.sortBy = new TestComparator();
+        context.clarityDirective.sort();
+        context.detectChanges();
+        expect(context.clarityElement.attributes['aria-sort'].value).toBe('ascending');
+
+        context.clarityDirective.sort();
+        context.detectChanges();
+        expect(context.clarityElement.attributes['aria-sort'].value).toBe('descending');
+      });
+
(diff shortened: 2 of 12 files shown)


src/clr-angular/data/datagrid/datagrid-cell.spec.ts
```diff
@@ -43,6 +43,10 @@ export default function(): void {
       context.detectChanges();
       expect(context.clarityElement.classList.contains('datagrid-signpost-trigger')).toBeTruthy();
     });
+
+    it('adds a11y roles to the rolumn', function() {
```
youdz: rolumn => cell


:joy: at "rolumn" by the way.

src/clr-angular/data/datagrid/datagrid-column.spec.ts
```diff
@@ -346,6 +346,20 @@ export default function(): void {
         expect(context.clarityElement.classList.contains('desc')).toBeTruthy();
       });
 
+      it('adds a11y roles to the rolumn', function() {
```
youdz: rolumn => column

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -211,6 +211,20 @@ export default function(): void {
...
+        const tableWrapper = context.clarityElement.querySelector('.datagrid-table-wrapper');
+        expect(tableWrapper.attributes.role.value).toEqual('grid');
+
+        const header = context.clarityElement.querySelector('.datagrid-head');
+        expect(header.attributes.role.value).toEqual('rowgroup');
+
+        const row = context.clarityElement.querySelector('.datagrid-row');
+        expect(row.attributes.role.value).toEqual('row');
+
+        const column = context.clarityElement.querySelector('.datagrid-column');
```
youdz: For rows and columns, we'd better be safe and check them all, it's barely more code:
```typescript
const columns = context.clarityElement.querySelectorAll('.datagrid-column');
columns.forEach(column => expect(column.attributes.role.value).toEqual('columnheader'));
```

src/clr-angular/data/datagrid/datagrid-cell.spec.ts
jeeyun: cost of copy-paste.. T_T thanks for catching!

src/clr-angular/data/datagrid/render/table-renderer.spec.ts
```diff
@@ -51,6 +51,11 @@ export default function(): void {
       expect(body.textContent).not.toMatch('Hello');
       expect(body.textContent).toMatch('World');
     });
+
+    it('adds a11y roles', function() {
+      const rowContent = context.clarityElement.querySelector('.datagrid-body');
+      expect(rowContent.attributes.role.value).toEqual('rowgroup');
```
Shijir: `.toEqual` works fine, but it should be used for objects. Here, we are checking a string value, which is primitive so `.toBe` is more appropriate, I think.

jeeyun: will update the tests. thanks!

Issue #248, opened 2016-12-14
Combobox component (new)


reddolan:
### Description
- a Select component that allows for typing as well as scrolling through options
- similar to autocomplete or combobox

### Use cases under consideration
- type into field and filter down to options matching query
- click something to show the options available
- scroll through list and select an option
- multi-select
- option grouping (or categories)
- accessible navigation, option selection and option deletion
- system status for slow responding option list
- highlight value matches in options

#### Basic
Default view
- will look like a standard Select
- caret will be grey and have an circle around it to distinguish it from a standard Select
- the input field will show the standard input focus style (blue underline) while being interacted with

![inactive](https://cloud.githubusercontent.com/assets/1858226/21270544/be955880-c37c-11e6-8163-1cb870d4e7de.png)

Active view
- clicking the caret will show the Options dropdown
- with the dropdown showing, the arrow keys can navigate up or down through options
- hitting Enter will select the option and display it in the input field
- show up to 10 options at a time
- if more than 10 options exist, user will be able to scroll down through them
- hover or highlight while scrolling will be the hover grey (#EEE)

![active - hover](https://cloud.githubusercontent.com/assets/1858226/21270765/a0f1fe4a-c37d-11e6-87f5-c715cf9eb7ce.png)

Selections
- selected option will be the row selection blue (#D9E4EA)

![active - selected](https://cloud.githubusercontent.com/assets/1858226/21271192/8fd47d34-c37f-11e6-8ab5-463e36072d45.png)

Typing
- tabbing into the field will show the cursor in the input field
- options are displayed after 1 character has been typed
- query values are highlighted bold within each matching option
- pressing the down arrow key shows the dropdown
- user can tab to the caret and activate the dropdown
- tabbing out, clicking outside of, or hitting Escape will close the dropdown

![active - typing](https://cloud.githubusercontent.com/assets/1858226/21270781/ad7628f8-c37d-11e6-8e6a-4fef8e868d2f.png)

Searching message
- if the options come back quickly (< 100ms), show the Options dropdown
- if the options load slowly (> 100ms), show the Searching dropdown
- Searching message will use the Select's label value

![searching](https://cloud.githubusercontent.com/assets/1858226/21270734/7f884ed0-c37d-11e6-94d0-f307cf1448eb.png)

#### Multiselect
- includes all Basic interactions
- user can select more than one option by clicking on it or hitting Enter
- holding Shift and selecting two separate options will select all options in between them
[...]

PR #2505, opened 2018-07-30
Combobox initial implementation

gnomeontherun:
This is the select2 component renamed to combobox for the time being until we finish this up and overwrite the default select element in HTML. This just gets it into the code base, without conflicts, and gives us start.

see #248 #2469


src/dev/src/app/combobox/combobox-static.html
```diff
@@ -0,0 +1,76 @@
...
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<form class="test-select">
+    <section class="form-block">
+        <label>Select Boxes</label>
+        <div class="form-group">
+            <label for="selects_1">This is a select box</label>
+            <div class="select">
+                <select id="selects_1">
```
youdz: Why do we still have all these select demos for combobox? We should just delete them.

src/dev/src/app/combobox/combobox-static.ts
```diff
@@ -5,5 +5,5 @@
  */
 import { Component } from '@angular/core';
 
-@Component({ selector: 'clr-basic-ng-select-demo', templateUrl: './basic-ng-select.html' })
-export class BasicNgSelectDemo {}
+@Component({ templateUrl: './combobox-static.html' })
+export class ComboboxStaticDemo {}
```
youdz: Yeah, that's youre other Select PR. We can just delete that.

Issue #250, opened 2016-12-15
Accordion Component


getcodenow:
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
I expected to find behavior for an accordion component within clarity UI so I don't have pull in accordion functionality in from bootstrap library which has conflicts with Clarity.  Want to also add that Clarity is awesome.  You guys have done a great job with these components and the packaging.

### Actual behavior
The closest thing I can find to an accordion component is the StackedView which is very close to the required functionality, but is designed to work with name/value pairs.  In my case I want to insert my content into a single column instead of being required to have two columns.

The plunk below shows the behavior I see, but what I would like to achieve is for the Card embedded within the Stacked View to take up the entire horizontal space.

### Reproduction of behavior
https://plnkr.co/edit/DwtjI0?p=preview

### Environment details

* **Angular version:** 2.2.3

* **Clarity version:** 0.7.3 

* **OS and version:** OSX 10.12.1

* **Browser:** Chrome 54.0.2840.98 (64-bit) | Firefox 50.0.2 | Safari 10.0.1

lfs1102:
I have a similar issue. 
I want to display only one data in a row and display it with full width. But Stack View works with name/value pairs and fixes the with of label to 40% and content to 60%. It create \<dd\> tag for label and \<dt\> tag for content dynamically so that I can't modify the style. 
<img width="1260" alt="screen shot 2016-12-19 at 3 41 21 pm" src="https://cloud.githubusercontent.com/assets/6532225/21305296/bdb1d58e-c605-11e6-8bb0-a612b4790d76.png">

Actually my issue is not for an Accordion Effect, but the problem is similar: Stack View should be more flexible to developers.

P.S. I was told about [Tree view](https://github.com/vmware/clarity/issues/239), it would solve my problem.

youdz:
It can be done with the current Stack View without waiting for a new feature, but it's not 100% perfect, it might need some extra styles tweaking. Put everything in the labels (you don't even have to put a single `<clr-stack-content>` in your HTML), then customize the styles of your `#single-column` stack view by adding this in your CSS:
```css
#single-column .stack-block-label {
    max-width: 100%;
    flex-basis: 100%;
}
#single-column .stack-block-content {
    display: none;
}
```
I edited your plunker to show it: https://plnkr.co/edit/tvWpP8pk5IfwG4Tv8Y04?p=preview
[...]

PR #2508, opened 2018-07-31
feature: Render Refactor

hippee-lee:
- Updates rendering for wrapped cells, rows and columns
- Updates calculations for autosizing (still some issues)
- Updates templates for fixed elements and vertical/horizontal scrolling
- Removed old rendering for body, table and associated tests
- Adjusts the size of the loading spinner

NOTE: This pr adds the sticky containers to datagrid header and rows but doesn't put any elements in there yet due to the popover issues that creates.


src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -39,28 +43,41 @@ export const domAdapterFactory = (platformId: Object) => {
...
     {selector: "clr-datagrid", providers: [{provide: DomAdapter, useFactory: domAdapterFactory, deps: [PLATFORM_ID]}]})
-export class DatagridMainRenderer implements AfterContentInit, AfterViewChecked, OnDestroy {
+export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, AfterViewChecked, OnDestroy {
     constructor(private organizer: DatagridRenderOrganizer, private items: Items, private page: Page,
-                private domAdapter: DomAdapter, private el: ElementRef, private renderer: Renderer2) {
-        this._subscriptions.push(organizer.computeWidths.subscribe(() => this.computeHeadersWidth()));
-        this._subscriptions.push(this.page.sizeChange.subscribe(() => {
+                private domAdapter: DomAdapter, private el: ElementRef, private renderer: Renderer2,
+                private tableSizeService: TableSizeService) {
+        this.subscriptions.push(this.organizer.renderStep.subscribe(step => {
```
gnomeontherun: Another place you could use `filter` for the steps.

src/clr-angular/data/datagrid/render/render-organizer.ts
```diff
@@ -7,65 +7,31 @@
 import {Injectable} from "@angular/core";
 import {Observable} from "rxjs/Observable";
 import {Subject} from "rxjs/Subject";
+import {DatagridRenderStep} from "../interfaces/render-step.interface";
+
 
 @Injectable()
 export class DatagridRenderOrganizer {
+    _renderStep: Subject<DatagridRenderStep> = new Subject<DatagridRenderStep>();
```
gnomeontherun: Needs to be `private`

src/clr-angular/data/datagrid/render/render-organizer.ts
```diff
@@ -7,65 +7,31 @@
...
-    }
-
-    public scrollbar = new Subject<any>();
-    public scrollbarWidth = new Subject<number>();
-
-    protected _done = new Subject<any>();
-    public get done(): Observable<any> {
-        return this._done.asObservable();
-    }
+    constructor() {}
```
gnomeontherun: Unnecessary

src/clr-angular/data/datagrid/render/render-organizer.ts
```diff
@@ -7,65 +7,31 @@
 import {Injectable} from "@angular/core";
 import {Observable} from "rxjs/Observable";
 import {Subject} from "rxjs/Subject";
+import {DatagridRenderStep} from "../interfaces/render-step.interface";
+
 
 @Injectable()
 export class DatagridRenderOrganizer {
```
gnomeontherun: I like this approach a lot better than so many observables, nice.

src/clr-angular/data/datagrid/render/render-organizer.spec.ts
```diff
@@ -31,13 +37,21 @@ export default function(): void {
...
             this.organizer.resize();
             let step = 0;
-            this.organizer.noLayout.subscribe(on => expect(step++).toBe(on ? 0 : 7));
-            this.organizer.clearWidths.subscribe(() => expect(step++).toBe(1));
-            this.organizer.detectStrictWidths.subscribe(() => expect(step++).toBe(2));
-            this.organizer.tableMode.subscribe(on => expect(step++).toBe(on ? 3 : 5));
-            this.organizer.computeWidths.subscribe(() => expect(step++).toBe(4));
-            this.organizer.alignColumns.subscribe(() => expect(step++).toBe(6));
-            this.organizer.scrollbar.subscribe(() => expect(step++).toBe(8));
+            this.organizer.renderStep.subscribe(renderStep => {
```
gnomeontherun: I believe this works because everything essentially becomes synchronous, but this test is written in a way that if an observable was async that it would pass even if it failed. Should we use `async` to help with this?

src/clr-angular/data/datagrid/render/render-organizer.spec.ts
```diff
@@ -31,13 +37,21 @@ export default function(): void {
...
+                if (renderStep === DatagridRenderStep.CALCULATE_MODE_ON) {
+                    expect(step++).toBe(0);
```
gnomeontherun: I know this was an existing test, but I'm not a big fan of this `step++` business. It makes me have to stop and think, oh yeah its postfixed so its still `0` but after this expression will be `1`. 

Suggestion: move the `step + 1` addition to the top of the subscription since its done EVERY time.

src/clr-angular/data/datagrid/render/row-renderer.ts
```diff
@@ -6,13 +6,19 @@
...
+import {DatagridRenderStep} from "../interfaces/render-step.interface";
+
 import {DatagridCellRenderer} from "./cell-renderer";
 import {DatagridRenderOrganizer} from "./render-organizer";
 
 @Directive({selector: "clr-dg-row, clr-dg-row-detail"})
 export class DatagridRowRenderer implements AfterContentInit, OnDestroy {
     constructor(private organizer: DatagridRenderOrganizer) {
-        this.subscription = organizer.alignColumns.subscribe(() => this.setWidths());
+        this.subscription = organizer.renderStep.subscribe(step => {
```
gnomeontherun: `filter` again

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -39,28 +43,41 @@ export const domAdapterFactory = (platformId: Object) => {
...
             // TODO: only re-stabilize if a column was added or removed. Reordering is fine.
             this.columnsSizesStable = false;
             this.stabilizeColumns();
         }));
     }
 
+    // Initialize and set Table width for horizontal scrolling here.
+    ngAfterViewInit() {
+        this.tableSizeService.table = this.el;
+        // this.tableSizeService.updateRowWidth();
```
gnomeontherun: Remove, no such method

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -152,28 +165,10 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewChecked,
...
-            // 3. While this is happening the datagrid is in the process of cleaning up the view. The view first
-            // renders the new displayed items and then cleans up the old items. But there is a point where the view
-            // contains the old items as well as the new items. So if the page size is 10 the view contains 20 items.
-            // This causes the datagrid body to overflow.
-            // Now since shouldStabilizeColumns was set to true, the scrollbar width is calculated
-            // and added to the datagrid header. Adding the setTimeout gives Angular time to clean up the view so that
-            // the scrollbar disappears.
-            // See this: https://github.com/angular/angular/issues/19094
-            // When the above issue is resolve, remove the setTimeout
-            setTimeout(() => {
```
gnomeontherun: This is good news!

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -214,14 +198,41 @@ export class ClrDatagridRow implements AfterContentInit {
...
+                this.displayCells = false;
+            } else {
+                this.dgCells.forEach(cell => {
+                    this.scrollableCells.insert(cell.view);
+                });
+                this.displayCells = true;
+            }
+        }));
+        // Insert views the first time? Fixed the pagination issue after discussing w/ Eude.
+        this.dgCells.forEach(cell => this.scrollableCells.insert(cell.view));
```
gnomeontherun: If the displaymode service was a behaviorsubject and sent the default mode as display, you wouldn't need this extra default thing here.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -26,23 +27,24 @@ import {StateDebouncer} from "./providers/state-debouncer.provider";
...
 import {DatagridRenderOrganizer} from "./render/render-organizer";
 
+
 const PROVIDERS = [
     Selection, Items, FiltersProvider, Sort, Page, RowActionService, ExpandableRowsCount, DatagridRenderOrganizer,
-    DomAdapter, HideableColumnService, DatagridWillyWonka, StateDebouncer
+    DomAdapter, HideableColumnService, DatagridWillyWonka, StateDebouncer, DisplayModeService
 ];
 
 export default function(): void {
```
gnomeontherun: Missing a number of tests

* handles expand changes
* host wrapping tests
* display mode changes

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -91,7 +99,30 @@ export class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
...
-    @ContentChild(ClrDatagridItems) public iterator: ClrDatagridItems;
+    // TODO - Add correct type - rebase with: https://github.com/vmware/clarity/pull/2347
+    @ContentChild(ClrDatagridItems) public iterator: ClrDatagridItems<RowContext<any>>;
+
+    /**
+     * Set the state of the pinned first column
+     * default is false.
+     */
+
+    public _pinnedColumn: boolean = false;
```
gnomeontherun: Matt and I have talked about this a little bit, while I think it would be nice to have the the pinned columns pieces separated from here, they are pretty minor and inert. What do you think @youdz ?

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -168,16 +199,25 @@ export class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
...
     ngAfterContentInit() {
+        if (!this.items.smart) {
+            this.items.all = this.rows.map((row: ClrDatagridRow) => row.item);
+        }
+
+        this.rows.forEach(row => {
+            this.displayedRows.insert(row.view);
+        });
+
         this._subscriptions.push(this.rows.changes.subscribe(() => {
```
gnomeontherun: Another place where perhaps we want the row changes to be a behavior subject so we trigger this on init and don't have duplicated logic in rows above.

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -201,8 +241,45 @@ export class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
                 this.selectedChanged.emit(s);
             }
         }));
+        this.displayMode.view.subscribe(viewChange => {
```
gnomeontherun: Same comments about view change mode as in row and the `while` blocks.

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -215,4 +292,10 @@ export class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
     resize(): void {
         this.organizer.resize();
     }
+
+    @ViewChild("projectedDisplayColumns", {read: ViewContainerRef}) projectedDisplayColumns: ViewContainerRef;
+    @ViewChild("projectedCalculationColumns", {read: ViewContainerRef}) projectedCalculationColumns: ViewContainerRef;
+    @ViewChild("displayedRows", {read: ViewContainerRef}) displayedRows: ViewContainerRef;
+    @ViewChild("calculationRows", {read: ViewContainerRef}) calculationRows: ViewContainerRef;
+    public showDisplayTable = false;  // Init to calculate when a datagrid is created.
```
gnomeontherun: This does not appear to be used anywhere.

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -201,8 +241,45 @@ export class ClrDatagrid implements AfterContentInit, AfterViewInit, OnDestroy {
...
+        });
+        // Make sure columns get added when datagrid is first created.
+        this.columns.forEach(column => {
+            this.projectedDisplayColumns.insert(column.view);
+        });
+
+        // Moving here to try fixing conditional pagination issue.
     }
 
+    public display = true;
```
gnomeontherun: This does not appear to be used anywhere.

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -44,13 +44,13 @@ export default function(): void {
 
             it("allows to manually resize the datagrid", function() {
```
gnomeontherun: Missing some tests, particularly around projection based on display modes and perhaps some around the modification to row changes.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -840,60 +874,83 @@
...
+
+            .datagrid-calculation-header {
+                display: table-header-group;
+
+                clr-dg-column {
+                    display: table-cell;
+                    min-width: 4rem;
+                }
+
+                .datagrid-column {
```
gnomeontherun: Are these duplicated styles? If so can we extend them instead?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -522,7 +551,7 @@
 
         .datagrid-placeholder {
             background: $clr-table-bgcolor;
-            border-top: $clr-default-borderwidth solid $clr-table-border-color;
+            //border-top: $clr-default-borderwidth solid $clr-table-border-color;
```
gnomeontherun: Remove?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -355,11 +386,10 @@
                     background-image: generateFilterIcon($action-blues-dark-midtone),
                     generateCircleIcon($action-blues-dark-midtone);
                 }
-
             }
 
             .datagrid-filter {
-                position: absolute;
+                //position: relative;
```
gnomeontherun: Remove?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -245,51 +270,57 @@
...
-            display: flex;
-            flex-flow: column nowrap;
-            flex: 1 1 auto;
-            min-height: 3rem;
-            height: 100%;
-            overflow-y: auto;
-            -ms-overflow-style: -ms-autohiding-scrollbar;
-
+            width: auto;
+            //min-height: 3rem; // Preserves original design but could this be updated to 1.5 rem for single row / null data states?
```
gnomeontherun: Remove?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -245,51 +270,57 @@
...
             }
         }
+
+        .datagrid-row-scrollable {
+            flex-direction: row;
+        }
     }
 
     .datagrid-table {
+        //flex: 1 1;
```
gnomeontherun: Remove all the commented out styles if they aren't needed

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -62,12 +71,13 @@ export class DatagridHeaderRenderer implements OnDestroy {
                 this.columnResizer.columnResizeBy = 0;
                 this.widthSet = false;
             }
+            // Don't set width if there is a user-defined one. Just add the strict width class.
             this.renderer.addClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
-            // We don't actually set the width if there already is a user-defined one, we just add the class
             return;
```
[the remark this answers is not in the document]
hippee-lee: This is here because we don't want the code outside the if block to run when there is a strict width. We are handling three use cases here:
1. Strict width and user resized to set the width
2. user defined strict width on the column via attr or css
3. columns and cells need to flex and adjust width.

src/clr-angular/data/datagrid/render/render-organizer.spec.ts
hippee-lee: I'm not sure but I think we should keep it synchronous. The render function we are testing absolutely depends on this exact order to get calculations correct. 

Would it make sense to instead test that the renderStep is synchronous?

src/clr-angular/data/datagrid/render/render-organizer.spec.ts
hippee-lee: Will move.

hippee-lee: Actually, trying to quickly move this means I have to add logic that handles the 0 value case  for the first step and the increment case for the rest of the steps. 

I'll add a note to revisit the pattern if you are ok with it.

src/clr-angular/data/datagrid/providers/display-mode.service.ts
```diff
@@ -0,0 +1,37 @@
...
+                this.update(DatagridDisplayMode.DISPLAY);
+            }
+        });
+    }
+
+    public get view(): Observable<DatagridDisplayMode> {
+        return this._view.asObservable();
+    }
+
+    public update(view: DatagridDisplayMode) {
```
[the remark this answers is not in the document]
hippee-lee: I will update this.

src/clr-angular/data/datagrid/providers/display-mode.service.ts
```diff
@@ -0,0 +1,37 @@
...
+import {Injectable} from "@angular/core";
+import {Observable} from "rxjs/Observable";
+import {Subject} from "rxjs/Subject";
+
+import {DatagridDisplayMode} from "../interfaces/display-mode.interface";
+import {DatagridRenderStep} from "../interfaces/render-step.interface";
+import {DatagridRenderOrganizer} from "../render/render-organizer";
+
+@Injectable()
+export class DisplayModeService {
```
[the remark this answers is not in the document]
hippee-lee: I added a test for this.

src/clr-angular/data/datagrid/interfaces/render-step.interface.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum DatagridRenderStep {
```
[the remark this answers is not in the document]
hippee-lee: I moved these to an `enums` folder.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -11,10 +11,16 @@ import {ElementRef, Injectable} from "@angular/core";
...
  */
 @Injectable()
-export class TableHeightService {
-    public tableRef: ElementRef;
+export class TableSizeService {
+    private tableRef: Element;
 
-    getHeight(): string {
-        return `${this.tableRef.nativeElement.clientHeight}px`;
+    constructor() {}
```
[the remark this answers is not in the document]
hippee-lee: Removed.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -11,10 +11,16 @@ import {ElementRef, Injectable} from "@angular/core";
...
+    public set table(table: ElementRef) {
+        this.tableRef = table.nativeElement.querySelector(".datagrid-table");
```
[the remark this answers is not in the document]
hippee-lee: I added a guard for ssr.

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -6,13 +6,19 @@
...
+import {DatagridRenderStep} from "../interfaces/render-step.interface";
+
 import {STRICT_WIDTH_CLASS} from "./constants";
 import {DatagridRenderOrganizer} from "./render-organizer";
 
 @Directive({selector: "clr-dg-cell"})
 export class DatagridCellRenderer implements OnDestroy {
     constructor(private el: ElementRef, private renderer: Renderer2, organizer: DatagridRenderOrganizer) {
-        this.subscription = organizer.clearWidths.subscribe(() => this.clearWidth());
+        this.subscription = organizer.renderStep.subscribe(step => {
```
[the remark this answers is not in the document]
hippee-lee: I updated the render-organizer with a util function that takes an array of `DatagridRenderStep`'s `filterRenderSteps` and will add tests for that.

src/clr-angular/data/datagrid/render/cell-renderer.spec.ts
```diff
@@ -22,25 +23,23 @@ export default function(): void {
...
         });
 
         it("resets the cell to default width when notified", function() {
             context.clarityDirective.setWidth(true, 42);
             expect(context.clarityElement.style.width).toBe("42px");
             expect(context.clarityElement.classList).toContain("datagrid-fixed-width");
-            organizer.clearWidths.next();
+            organizer._renderStep.next(DatagridRenderStep.CLEAR_WIDTHS);
             expect(context.clarityElement.style.width).toBeFalsy();
             expect(context.clarityElement.classList).not.toContain("datagrid-fixed-width");
```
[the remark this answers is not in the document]
hippee-lee: I updated this to use the `STRICT_WIDTH_CLASS` const.

src/app/datagrid/column-sizing/column-sizing.ts
```diff
@@ -16,6 +16,7 @@ import {User} from "../inventory/user";
 })
 export class DatagridColumnSizingDemo {
     users: User[];
+    selected = [];
```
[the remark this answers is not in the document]
hippee-lee: Will remove.

src/app/datagrid/column-sizing/column-sizing.html
```diff
@@ -28,14 +28,15 @@ <h2>Smart column sizing</h2>
 <h2>Forcing columns width</h2>
 
 <clr-datagrid>
-    <clr-dg-column>User ID</clr-dg-column>
+    <clr-dg-column [style.min-width.px]="120">User ID</clr-dg-column>
```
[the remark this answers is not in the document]
hippee-lee: Will remove

src/clr-angular/data/datagrid/render/main-renderer.ts
hippee-lee: Will remove.

src/clr-angular/data/datagrid/render/row-renderer.ts
hippee-lee: Will be handled with the `filterRenderSteps` util function in the `render-organizer.

src/clr-angular/data/datagrid/render/dom-adapter.spec.ts
```diff
@@ -12,13 +13,15 @@ import {DomAdapter} from "./dom-adapter";
 interface UserContext {
     domAdapter: DomAdapter;
     element: HTMLElement;
+    elementRef: ElementRef;
```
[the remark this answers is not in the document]
hippee-lee: I'm not 100% sure why I made this change. I think that initially, I had an issue with testing. Reverting things back doesn't have any errors in the implementation of the tests. 

I will revert these changes.

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -38,37 +39,42 @@ export default function(): void {
...
 
         it("resets the column to default width when notified", function() {
-            context.clarityDirective.setWidth(123);
-            expect(context.clarityElement.style.width).toBe("123px");
-            organizer.clearWidths.next();
+            // context.clarityDirective.setWidth(123);
+            // expect(context.clarityElement.style.width).toBe("123px");
+
+
+            // spy on header-renderer.clearWidth()
```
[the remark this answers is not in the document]
hippee-lee: I think I addressed this and reverted the test back to the original after I fixed the auto-sizing issue.

src/app/datagrid/pagination/pagination.html
```diff
@@ -79,9 +79,6 @@ <h2>Pagination</h2>
         <clr-dg-cell>
             <span class="color-square" [style.backgroundColor]="user.color"></span>
         </clr-dg-cell>
-        <clr-dg-row-detail *clrIfExpanded="expand">
```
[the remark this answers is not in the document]
hippee-lee: Extraneous. Will remove.

src/app/datagrid/test-cases/test-cases.html
```diff
@@ -34,7 +34,7 @@ <h4>Zero Rows + Fixed Height</h4>
     <p>Datagrid height should be fixed irrespective on the number of rows displayed.</p>
     <button class="btn btn-outline-primary" (click)="toggle()">Toggle Spinner</button>
     <div style="height: 300px">
-        <clr-datagrid style="height: 100%" [clDgRowSelection]="true">
+        <clr-datagrid style="height: 100%" [clrDgLoading]="loading">
```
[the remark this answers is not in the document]
hippee-lee: It should toggle loading state with the rest of the test cases.

src/clr-angular/data/datagrid/wrapped-cell.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {AfterViewInit, Component, EmbeddedViewRef, TemplateRef, ViewChild} from "@angular/core";
+
+import {DynamicWrapper} from "../../utils/host-wrapping/dynamic-wrapper";
+
+@Component({
+    selector: "clr-wrapped-cell",
```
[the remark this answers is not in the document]
hippee-lee: Will update this and the other wrapper components to use `dg-`

src/clr-angular/data/datagrid/datagrid-cell.ts
```diff
@@ -43,8 +56,7 @@ export class ClrDatagridCell {
     private hiddenStateSubscription: Subscription;
 
     constructor(public hideableColumnService: HideableColumnService, private _el: ElementRef,
-                private _renderer: Renderer2) {}
-
+                private _renderer: Renderer2, private vcr: ViewContainerRef, el: ElementRef) {}
```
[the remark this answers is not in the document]
hippee-lee: Will add it.

src/clr-angular/data/datagrid/wrapped-cell.ts
```diff
@@ -0,0 +1,27 @@
...
+    template: `        
+        <ng-template #cellPortal>
+            <ng-content></ng-content>
+        </ng-template>
+    `,
+})
+export class ClrWrappedCell implements DynamicWrapper, AfterViewInit {
```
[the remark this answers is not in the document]
hippee-lee: Will address with the same fix.

src/clr-angular/utils/_layers.clarity.scss
```diff
@@ -16,3 +18,75 @@ $clr-layers: (
     modal: 1050,
     tooltips: 1070
 ) !default;
+
+/*
+Elements with z-index
```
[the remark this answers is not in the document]
hippee-lee: No, it was a leftover note to myself when I was working on the row-action popover issue. 
Will remove.

src/clr-angular/utils/conditional/if-open.service.ts
```diff
@@ -19,6 +19,7 @@ import {Subject} from "rxjs/Subject";
  *
  */
 export class IfOpenService {
+    constructor() {}
```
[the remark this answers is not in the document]
hippee-lee: Will remove.

src/clr-angular/utils/host-wrapping/host-wrapper.ts
```diff
@@ -17,6 +17,15 @@ import {
...
 import {EmptyAnchor} from "./empty-anchor";
 
+
+/**
+ * @class HostWrapper
+ *
+ * @description
+ * HostWrapper must be called in OnInit to ensure that the Views are ready. If its called in a constructor the view is
+ * still undefined.
+ * TODO - make sure these comment annotations do not break ng-packgr.
```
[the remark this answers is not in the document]
hippee-lee: Will remove both.

src/clr-angular/utils/expand/providers/expand.spec.ts
```diff
@@ -12,8 +12,12 @@ export default function(): void {
         });
 
         it("starts with the correct default settings", function() {
+            let isExpanded = null;
+            this.expand.replace.subscribe(expandChange => {
```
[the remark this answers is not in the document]
hippee-lee: Will add an `it spec` for _notifies when cells are replaced_

src/clr-angular/data/datagrid/datagrid-action-popover.ts
```diff
@@ -0,0 +1,35 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import {Component, ElementRef, Inject, Injector, Optional} from "@angular/core";
+
+import {AbstractPopover} from "../../popover/common/abstract-popover";
+import {Point} from "../../popover/common/popover";
+import {POPOVER_HOST_ANCHOR} from "../../popover/common/popover-host-anchor.token";
+
+@Component({
```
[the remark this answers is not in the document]
hippee-lee: Was part of the investigation. I was trying to see if the new Popover would help with the z-index issues the row action popovers are having with the sticky positioning of row elements and sibling rows. 

Will remove.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Internal datagrid service that holds a reference to the clr-dg-table element and exposes a method to get height.
+ */
+@Injectable()
+export class TableSizeService {
+    private tableRef: Element;
+
+    constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+    public set table(table: ElementRef) {
+        if (isPlatformBrowser(this.platformId) && table.nativeElement) {
+            this.tableRef = table.nativeElement.querySelector(".datagrid-table");
```
Shijir: I just noticed that we actually don't need query DOM to find this table ref. The table service could be called inside `DatagridTableRenderer` and its `elementRef` is assigned to `TableSizeService.ref`. So we would no longer need to even inject this `PLATFORM_ID`.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+
+    // Used when resizing columns to show the column border being dragged.
+    getColumnDragHeight(): string {
+        return `${this.tableRef.clientHeight}px`;
```
Shijir: The DnD topic branch has DomAdapter changes. So once that's merged in, we could use it here.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -12,24 +12,39 @@ import {
...
     private _differ: IterableDiffer<any>;
-
-    constructor(public template: TemplateRef<any>, private _differs: IterableDiffers, private _items: Items) {
+    displayed: Array<any> = [];
+    constructor(public template: TemplateRef<RowContext<T>>, private _differs: IterableDiffers, private _items: Items,
+                private vcr: ViewContainerRef) {
         _items.smartenUp();
+        _items.change.subscribe(items => {
+            this.displayed = items;
+            this.vcr.clear();  // It is additive if I don't clear the datagrids vcr
```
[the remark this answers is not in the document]
hippee-lee: I'll change this to a local const to hold the items.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -12,24 +12,39 @@ import {
...
+            // Will need to use IterableDiff to properly account for trackBy
+            // TODO revisit detectChanges() here, even thought it fixes conditionalPaginition, it isn't ideal.
+            for (const item of this.displayed) {
+                this.vcr.createEmbeddedView(this.template, {$implicit: item}).detectChanges();
```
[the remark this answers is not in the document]
hippee-lee: Yes, lets talk about this one. @youdz helped me find this solution to a lifecycle issue with conditional pagination. e.g when consumer wants to enable/disable pagination in the datagrid. 

what was happening in that case was that the row changes were heard in `datagrid.ts` and that was calling the `row.view` getter on a row that hadn't been initialized yet. 

I hav not looked for a way to batch up the detectChanges() but can look into it if this is a performance concern. I'd prefer to handle that optimization as a separate PR though.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -12,24 +12,39 @@ import {
...
         this._rawItems = items ? items : [];
     }
     private _differ: IterableDiffer<any>;
-
-    constructor(public template: TemplateRef<any>, private _differs: IterableDiffers, private _items: Items) {
+    displayed: Array<any> = [];
+    constructor(public template: TemplateRef<RowContext<T>>, private _differs: IterableDiffers, private _items: Items,
+                private vcr: ViewContainerRef) {
         _items.smartenUp();
+        _items.change.subscribe(items => {
```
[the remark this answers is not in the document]
hippee-lee: I added a test for the `Items` change subscription.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -21,18 +22,30 @@ import {Selection, SelectionType} from "./providers/selection";
...
     template: `
-        <ng-container *ngIf="!replace">
-            <clr-dg-cell class="datagrid-fixed-column"
-                *ngIf="selection.selectionType === SELECTION_TYPE.Multi 
-                    || selection.selectionType === SELECTION_TYPE.Single"></clr-dg-cell>
-            <clr-dg-cell *ngIf="rowActionService.hasActionableRow" class="datagrid-fixed-column"></clr-dg-cell>
-            <clr-dg-cell class="datagrid-fixed-column"></clr-dg-cell>
-        </ng-container>
+        <!-- space for multiselection state -->
+        <div class="datagrid-cell datagrid-select datagrid-fixed-column"
```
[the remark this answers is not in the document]
hippee-lee: Because I moved the check/radio's and action overflow buttons out of the sticky container (due to the z-index issue) I need to have space for them in the details view for the use case when one or more is present and cells are not replaced. 

tbh, I copied these changes from the header though and we use three placeholder elements there. Either way the conditional logic should be the same for header, row and detail elements. I'll change this or the header to be the same. 

fyi - The UX recommendation is not to have all three in a row at the same time iir correctly so consumers should only have 2 of these elements at a time.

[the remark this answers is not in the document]
hippee-lee: checking the templates, I made them all individual elements with the correct logic for each element on the `*ngIf`

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -0,0 +1,48 @@
...
+    <div class="datagrid-row-scrollable">
+        <div class="datagrid-scrolling-cells">
+            <div *ngIf="selection.selectionType === SELECTION_TYPE.Multi"
+                          class="datagrid-select datagrid-fixed-column datagrid-cell">
+                <clr-checkbox [clrChecked]="selected" (clrCheckedChange)="toggle($event)"></clr-checkbox>
+            </div>
+            <div *ngIf="selection.selectionType === SELECTION_TYPE.Single"
+                         class="datagrid-select datagrid-fixed-column datagrid-cell">
+                <div class="radio">
+                    <input type="radio" [id]="id" [name]="selection.id + '-radio'" [value]="item"
```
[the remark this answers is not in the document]
hippee-lee: If we want to get them in there I could either a) submit a change for that after the topic branch is rebased with master or b) create an issue now and address it more atomically after the topic branch is pushed to master.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -100,11 +66,31 @@ export class ClrDatagridRow implements AfterContentInit {
...
-                public hideableColumnService: HideableColumnService) {
+                public hideableColumnService: HideableColumnService, private displayMode: DisplayModeService,
+                private vcr: ViewContainerRef, private renderer: Renderer2, private el: ElementRef) {
         this.id = "clr-dg-row" + (nbRow++);
         this.role = selection.rowSelectionMode ? "button" : null;
+
+        // Listen to both the expand.replace and the expand.expandChange observables
+        this.combinedLatestExpand = combineLatest(this.expand.replace, this.expand.expandChange);
+
+        this.subscriptions.push(this.combinedLatestExpand.subscribe(latestValues => {
```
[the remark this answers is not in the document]
hippee-lee: That works. ty.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -1,90 +1,56 @@
...
+    templateUrl: "./datagrid-row.html",
     host: {
         "[class.datagrid-row]": "true",
         "[class.datagrid-selected]": "selected",
         "[attr.tabindex]": "selection.rowSelectionMode ? 0 : null"
     },
     providers: [Expand, {provide: LoadingListener, useExisting: Expand}]
 })
-export class ClrDatagridRow implements AfterContentInit {
+export class ClrDatagridRow {
```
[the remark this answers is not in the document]
hippee-lee: Doh! Will fix.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -100,11 +66,31 @@ export class ClrDatagridRow implements AfterContentInit {
...
         this.id = "clr-dg-row" + (nbRow++);
         this.role = selection.rowSelectionMode ? "button" : null;
+
+        // Listen to both the expand.replace and the expand.expandChange observables
+        this.combinedLatestExpand = combineLatest(this.expand.replace, this.expand.expandChange);
+
+        this.subscriptions.push(this.combinedLatestExpand.subscribe(latestValues => {
+            const [expandReplaceValue, expandChangeValue] = latestValues;
+
+            if (expandReplaceValue && expandChangeValue) {  // replaced and expanding
```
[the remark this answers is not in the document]
hippee-lee: wow, to many days staring at the sun. 
I updated this but noticed an issue I missed in the DOM changes. Will also add  code to display the collapse button when in the replace condition and add css to position it properly.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -214,14 +198,41 @@ export class ClrDatagridRow implements AfterContentInit {
...
+            this.scrollableCells.insert(cell.view);
         });
     }
 
+    ngAfterViewInit() {
+        this.subscriptions.push(this.displayMode.view.subscribe(viewChange => {
+            // Listen for view changes and move cells around depending on the current displayType
+            while (this.scrollableCells.detach()) {
+            }  // remove cells containers
+            while (this.calculatedCells.detach()) {
```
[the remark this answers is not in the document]
hippee-lee: I'm open to that, way back when I got help with the prototype this was the way we did it. @youdz - any objections here before I change everything like this?

src/clr-angular/data/datagrid/datagrid-row.ts
hippee-lee: BehaviorSubject ftw.

hippee-lee: I was able to remove similar code in `datagrid.ts`.

src/clr-angular/data/datagrid/datagrid.ts
hippee-lee: Updated when I added behavior subject for similar code in datagrid-row.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
hippee-lee: Added tests for these three areas.

src/clr-angular/data/datagrid/datagrid.ts
hippee-lee: This I forgot to remove from before starting the render refactor. Will remove it for now. 

Pulling the pinned column demos out is easy - they are just a simple html/ts demo. Either way is fine with me.

src/clr-angular/data/datagrid/datagrid.ts
hippee-lee: It is used in the displayMode subscription.  line #250/#259
I removed the initialization value since display mode service switched to a behavior subject.

src/clr-angular/data/datagrid/datagrid.ts
hippee-lee: Will remove.

src/clr-angular/data/datagrid/datagrid.spec.ts
hippee-lee: Will add tests for display mode detection and projection for display and calculate modes.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: I don't thing I can extend here. I want to only override the display when `clr-dg-column` etc are in the calculation table. If I understand extension id have to do something like

```
.datagrid-calculation-header {
    display: table-header-group;
    @extend clr-dg-column; // but these styles will target the datagrid-calculation-header, not columns
}
```

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: Removed.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: Yes.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: done.

src/clr-angular/data/datagrid/providers/table-size.service.ts
hippee-lee: The table renderer is going away with this PR.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -332,4 +349,14 @@ export class ClrDatagridColumn extends DatagridFilterRegistrar<DatagridStringFil
...
      */
     public hideable: DatagridHideableColumnModel;
+
+    private wrappedInjector: Injector;
+
+    ngOnInit() {
+        this.wrappedInjector = new HostWrapper(ClrWrappedColumn, this.vcr);
+    }
+
+    public get view() {
```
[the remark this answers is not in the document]
youdz: According to our guidelines, this is the one case where you should name it `get _view()` with an underscore. It has to be public for component communication reasons, but is not part of our public API. I think that would be enough.

src/clr-angular/data/datagrid/datagrid-column.spec.ts
```diff
@@ -328,7 +334,7 @@ export default function(): void {
         });
 
         describe("View filters", function() {
-            it("doesn't display any filter by default", function() {
+            it("doesn't displayType any filter by default", function() {
```
youdz: Oops. 😛

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -12,24 +12,38 @@ import {
...
     TemplateRef,
-    TrackByFunction
+    TrackByFunction,
+    ViewContainerRef
 } from "@angular/core";
 
 import {Items} from "./providers/items";
 
+export type RowContext<T> = {
+    $implicit: T
```
youdz: As I mentioned when we were discussing prototypes, we need this to extends `NgForOfContext<T>`. That means you'll have to fill out a few boolean properties and an index when creating the context.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -12,24 +12,38 @@ import {
...
     private _differ: IterableDiffer<any>;
-
-    constructor(public template: TemplateRef<any>, private _differs: IterableDiffers, private _items: Items) {
+    constructor(public template: TemplateRef<RowContext<T>>, private _differs: IterableDiffers, private _items: Items,
+                private vcr: ViewContainerRef) {
         _items.smartenUp();
+        _items.change.subscribe(items => {
+            const displayed = items;
+            this.vcr.clear();  // It is additive if I don't clear the datagrids vcr
+            // Will need to use IterableDiff to properly account for trackBy
```
youdz: Yeah, this is going to be a performance issue. My best recommendation is to check out the code for `NgForOf` itself, and I'm okay with you copying most of it. 

Or even better, maybe you can see if you can simply proxy to an `NgForOf` that you instantiate manually, that would be ideal. Ping me if you want to discuss it, I'd be happy to pair for a little bit to help you get that done.

src/clr-angular/data/datagrid/datagrid-items.spec.ts
```diff
@@ -45,6 +45,13 @@ export default function(): void {
             expect(this.itemsProvider.displayed).toEqual([1, 2, 3, 4, 5]);
         });
 
+        it("subscribes to changes in the datagrid Items (_items)", function() {
+            // testing the observable subscription in the constructore
+            this.testComponent.numbers.push(6);
+            this.fixture.detectChanges();
+            expect(this.itemsProvider.displayed).toEqual([1, 2, 3, 4, 5, 6]);
```
youdz: This doesn't test the subscription, you are checking the provider's array here. You could remove the subscription in the directive and this test would still pass.

You simply want to check that the test component now displays the numbers. That's what your new code is doing, giving the responsibility of view creation to the `*clrDgItems` directive.

hippee-lee: I moved `.datagrid-row-replaced` up to the top level for the export.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -65,25 +57,55 @@
...
+                }
+            }
+
+          .datagrid-row-detail {
+              .datagrid-cell {
+                  display: block;
+                  padding-top: 11px;
+                  border-top: $clr-default-borderwidth solid $clr-table-border-color;
+              }
+              .datagrid-expandable-caret {
```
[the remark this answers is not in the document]
hippee-lee: I moved `.datagrid-row-replaced` up to the top level for the export.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -245,51 +297,53 @@
...
-        flex-flow: column nowrap;
-        /**
-            Important - the flex setting here allows headers/body rows to expand to full width of their container
-            Its needed to enable both full row hover and the details content when expand id declared.
-         */
-        flex: 1 1 auto;
+        flex-direction: column;
+        align-content: flex-start;
+        width: auto;
+        min-height: 3rem; // Includes header and space for a single (non existing) row
```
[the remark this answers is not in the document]
hippee-lee: @reddolan - can you weigh in here. Previously we had spaces for two rows when the datagrid was in loading state and didn't have any rows yet.  The visual change I'm proposing here would use space for one row. Pics are what this looks like with the loading spinner overlayed on top. Datagrids with action rows will be slightly different as the loading element now covers those as well. I can mock a quick demo if you want to see that use case ... 

### Two row empty state (previous implementation)
screen shot 2018-08-20 at 11 35 35 am

### One row w/ empty state (proposed change for rendering refactor implementation)
screen shot 2018-08-20 at 11 36 38 am

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -840,60 +894,82 @@
...
 
-        .datagrid-body {
+        .datagrid-calculation-table {
             display: table;
             table-layout: auto;
+
+            .datagrid-calculation-header {
+                display: table-header-group;
+
+                clr-dg-column {
```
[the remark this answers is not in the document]
hippee-lee: fixed

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -840,60 +894,82 @@
...
+                    color: $clr-p6-color;
+                    font-weight: $clr-typography-boldfont;
+                    font-size: $clr-typography-xxsmalltext;
+                    letter-spacing: $clr-kerning-tinytext;
+                    line-height: 0.5rem;
+                    margin-top: 1rem;
+                    margin-bottom: 0;
+                }
+            }
+            clr-dg-row {
```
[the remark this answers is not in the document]
hippee-lee: fixed.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -840,60 +894,82 @@
...
+                display: table-row;
+
+                clr-dg-cell {
```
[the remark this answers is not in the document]
hippee-lee: and fixed.

src/clr-angular/data/datagrid/wrapped-row.ts
```diff
@@ -0,0 +1,28 @@
...
+        <ng-template #rowPortal>
+            <ng-content></ng-content>
+        </ng-template>
+    `,
+})
+export class ClrWrappedRow implements DynamicWrapper, AfterViewInit {
+    _dynamic = false;
+
+    @ViewChild("rowPortal") templateRef: TemplateRef<void>;
+    rowView: EmbeddedViewRef<void>;  // the columns projected view (in memory)
```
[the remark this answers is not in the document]
hippee-lee: Yes, I meant row. Fixed.

src/clr-angular/utils/expand/providers/expand.spec.ts
```diff
@@ -12,12 +12,26 @@ export default function(): void {
         });
 
         it("starts with the correct default settings", function() {
+            let isExpanded = null;
```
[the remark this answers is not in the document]
hippee-lee: I changed the name to `doesReplace`.

src/clr-angular/data/datagrid/datagrid-column.spec.ts
hippee-lee: IntelliJ stopped working for me and I tried to replace a var name in one file with VSCode. Clearly it did not go well. I almost can't believe that this is the only one that I didn't find. 😳

Fixed.

src/clr-angular/data/datagrid/datagrid-items.ts
hippee-lee: Thanks for the `NgForOf` idea. Will address this with that.

src/clr-angular/data/datagrid/datagrid-items.ts
hippee-lee: Thanks for the `NgForOf` idea. Will address this with that.

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -44,13 +46,13 @@ export default function(): void {
...
-                    resizeDone = true;
+                let resizeSteps = 0;
+                organizer.renderStep.subscribe(() => {
+                    resizeSteps++;
                 });
-                expect(resizeDone).toBe(false);
+                expect(resizeSteps).toBe(0);
                 context.clarityDirective.resize();
-                expect(resizeDone).toBe(true);
+                expect(resizeSteps).toBe(6);
```
youdz: Not fond of the idea that this test is going to break every time we change the internal logic of our rendering steps, but it's an easy fix every time. So it's fine for now.

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -211,16 +213,89 @@ export default function(): void {
...
+                    context = this.create(ClrDatagrid, FullTest, [HideableColumnService]);
+                    displayMode = context.getClarityProvider(DisplayModeService);
+                });
+
+                it("moves columns into the display container", function() {
+                    context.clarityDirective.displayMode.view.subscribe(mode => {
+                        displayHeader = context.clarityElement.querySelector(".datagrid-header");
+                        displayColumns = displayHeader.querySelectorAll(".datagrid-column");
+                        if (mode === DatagridDisplayMode.DISPLAY) {
+                            expect(displayColumns.length).toBe(2);
```
youdz: Why is all of this in the subscription? You can just inline it after `displayMode.updateView(DatagridDisplayMode.DISPLAY);` and it's much safer. As usual, tests that put `expect` statements in the subscription will always pass if the Observable simply doesn't fire.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -214,12 +195,33 @@ export class ClrDatagridRow implements AfterContentInit {
...
+
+    ngAfterViewInit() {
+        this.subscriptions.push(this.displayMode.view.subscribe(viewChange => {
+            // Listen for view changes and move cells around depending on the current displayType
+            while (this.scrollableCells.detach()) {
+            }  // remove cells containers
+            while (this.calculatedCells.detach()) {
+            }  // remove cells containers
+            if (viewChange === DatagridDisplayMode.CALCULATE) {
+                this.dgCells.forEach(cell => {
```
youdz: Serious performance question: does this run **before or after** we set the `display:none` on the display div in `datagrid.ts`?

It depends on who subscribes first I guess, but if for some reason this cell-moving happens before the `IE performance again.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -214,12 +195,33 @@ export class ClrDatagridRow implements AfterContentInit {
...
+                    this.calculatedCells.insert(cell.view);
+                });
+                this.displayCells = false;
```
youdz: Same here, I don't see this used anywhere. Is it still needed?

src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -246,25 +271,22 @@ export default function(): void {
                 expect(context.clarityElement.querySelector(".spinner")).not.toBeNull();
             });
 
-            it("doesn't display the details when collapsed", function() {
-                expect(context.clarityElement.textContent).toMatch("Hello world");
```
youdz: Why delete these `expect`s? Looks like they should still work, and they test that we still display the original content of the row. Or are we just not rendering anything in these tests? If that's the case, then the test doesn't actually mean anything since "always displaying nothing" clearly satisfies "does not display the details".

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -215,24 +237,54 @@
...
+        }
+    }
 
+    .datagrid-row-flex {
+        flex: 1 1 auto;
+        display: flex;
+        flex-flow: column nowrap;
+
+        .datagrid-row-detail {
+            display: fles;
```
youdz: `display: fles`? 😛 

But seriously, how is it even rendering correctly right now?

src/clr-angular/data/datagrid/all.spec.ts
```diff
@@ -34,35 +34,39 @@ import DatagridRowSpecs from "./datagrid-row.spec";
...
-import DatagridTableRendererSpecs from "./render/table-renderer.spec";
+import WrappedCellSpec from "./wrapped-cell.spec";
+import WrappedColumnSpec from "./wrapped-column.spec";
+import WrappedRowSpec from "./wrapped-row.spec";
 
 describe("Datagrid", function() {
     addHelpers();
 
-    describe("Providers", function() {
+    fdescribe("Providers", function() {
```
youdz: Forgot an `fdescribe`.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -4,25 +4,37 @@
...
+    private _tableRef: HTMLElement;
+    private columns: Element[];
 
-    public get tableRef(): Element {
+    public get tableRef(): HTMLElement {
         return this._tableRef;
     }
 
-    public set tableRef(element: Element) {
+    public set tableRef(element: HTMLElement) {
```
youdz: Why getter and setter here?

src/clr-angular/data/datagrid/providers/table-size.service.spec.ts
```diff
@@ -3,32 +3,76 @@
...
+            this.sizeService = this.fixture.debugElement.injector.get(TableSizeService);
+            this.fixture.detectChanges();
+            this.table = this.fixture.elementRef.nativeElement.children[0];  // reference to the TestComponnt table
+            this.sizeService.tableRef = this.table;  // setting service up with the component table for testing
+        });
+
+        it("sets a tableRef property with an elementReference", function() {
+            let testTable: Element;
+            testTable = this.sizeService.tableRef;  // service getter
+            expect(testTable).toEqual(this.table);  // should equal TestComponent element
```
youdz: This test, given the `beforeEach`, is not checking anything at all. It checks the `beforeEach`, nothing related to the directive.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -4,25 +4,37 @@
...
         this._tableRef = element;
     }
 
-    constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+    constructor(@Inject(PLATFORM_ID) private platformId: Object, renderOrganizer: DatagridRenderOrganizer,
+                private renderer: Renderer2) {
+        this.subscriptions.push(renderOrganizer.renderStep.subscribe(step => {
```
youdz: You forgot to use your filter helper here.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -33,4 +45,20 @@ export class TableSizeService {
...
+    }
+
+    updateRowWidth(): void {
+        let newWidth: number = 0;
+        this.renderer.removeStyle(this.tableRef, "width");
+        this.columns = Array.from(this.tableRef.querySelectorAll(".datagrid-column"));
+        this.columns.forEach(item => {
+            newWidth += item.clientWidth;
+        });
+        this.renderer.setStyle(this.tableRef, "width", newWidth + "px");
```
youdz: I don't understand where this is needed at the moment. Is this another inert artifact from pinned columns? If so, that one actually sets a hard width on the datagrid so it's not as inert as we would like, we probably want it out of this PR.

Or I forgot about something you explained to me some time ago and it is needed. 🤔

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: Will fix and confirm the functionality.

src/clr-angular/data/datagrid/all.spec.ts
hippee-lee: Will remove.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -21,18 +22,35 @@ import {Selection, SelectionType} from "./providers/selection";
...
-            <clr-dg-cell class="datagrid-fixed-column"
-                *ngIf="selection.selectionType === SELECTION_TYPE.Multi 
-                    || selection.selectionType === SELECTION_TYPE.Single"></clr-dg-cell>
-            <clr-dg-cell *ngIf="rowActionService.hasActionableRow" class="datagrid-fixed-column"></clr-dg-cell>
-            <clr-dg-cell class="datagrid-fixed-column"></clr-dg-cell>
+        <ng-container *ngIf="!replacedRow">
+            <!-- space for multiselection state -->
+            <div class="datagrid-cell datagrid-select datagrid-fixed-column"
+                *ngIf="selection.selectionType === SELECTION_TYPE.Multi">
+                <span class="datagrid-column-title"></span>
```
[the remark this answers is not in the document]
hippee-lee: I went back to the old way of projecting the details template but the spaces are still needed in there for the row action items. They ensure the details columns/default lines up with the cells. 

I copied this from the header. Will remove all of the extra elements.

src/clr-angular/data/datagrid/providers/table-size.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+    public set table(table: ElementRef) {
+        if (isPlatformBrowser(this.platformId) && table.nativeElement) {
+            this.tableRef = table.nativeElement.querySelector(".datagrid-table");
+        }
+    }
+
+    // Used when resizing columns to show the column border being dragged.
+    getColumnDragHeight(): string {
+        return `${this.tableRef.clientHeight}px`;
```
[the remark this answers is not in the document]
hippee-lee: I'll add guards for both methods to check if `tableRef` is set.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -62,12 +69,20 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
         }
         this.renderer.removeClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
+        // Why is a width set, isn't this column suppsed to be flex: 1 1 auto?
         this.renderer.setStyle(this.el.nativeElement, "width", width + "px");
         this.widthSet = true;
     }
+
+    ngAfterViewInit() {
+        this.subscriptions.push(
```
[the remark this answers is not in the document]
hippee-lee: I'm not sure what my reason for moving the subscriptions was. Moved them back.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -62,12 +69,20 @@ export class DatagridHeaderRenderer implements OnDestroy {
                 this.columnResizer.columnResizeBy = 0;
                 this.widthSet = false;
             }
+            // Don't set width if there is a user-defined one. Just add the strict width class.
             this.renderer.addClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
-            // We don't actually set the width if there already is a user-defined one, we just add the class
             return;
         }
         this.renderer.removeClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
+        // Why is a width set, isn't this column suppsed to be flex: 1 1 auto?
```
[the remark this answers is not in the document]
hippee-lee: 👍 - I'll remove my question.

src/clr-angular/data/datagrid/providers/display-mode.service.ts
```diff
@@ -0,0 +1,35 @@
...
+        new BehaviorSubject<DatagridDisplayMode>(DatagridDisplayMode.DISPLAY);
+
+    constructor(renderOrganizer: DatagridRenderOrganizer) {
+        renderOrganizer.filterRenderSteps(DatagridRenderStep.CALCULATE_MODE_ON)
+            .subscribe(() => this._view.next(DatagridDisplayMode.CALCULATE));
+        renderOrganizer.filterRenderSteps(DatagridRenderStep.CALCULATE_MODE_OFF)
+            .subscribe(() => this._view.next(DatagridDisplayMode.DISPLAY));
+    }
+
+    public updateView(mode: DatagridDisplayMode) {
```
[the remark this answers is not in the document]
hippee-lee: Yes, I only added this for testing. Perhaps my test is wrong. In order to test things that are reactive to the view mode changes, I need a single instance of the service. e.g I can't have an instance of `DisplayModeService` and an instance of `MockDisplayModeService`. So far the way I've addressed this so far is to create a mock that extends the `DisplayModeService` and move the updateView there. Does that seem correct?

src/clr-angular/data/datagrid/providers/display-mode.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+            expect(currentView).toBe(DatagridDisplayMode.DISPLAY);
+        });
+
+        it("updates the view for DatagridDisplayMode.DISPLAY", function() {
+            const viewObservable = this.displayService.view;
+            let currentView = null;
+            viewObservable.subscribe(viewChange => {
+                currentView = viewChange;
+            });
+            this.displayService.updateView(DatagridDisplayMode.DISPLAY);
```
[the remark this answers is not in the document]
hippee-lee: I see what you mean. I'll fix this.

src/clr-angular/data/datagrid/providers/display-mode.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        it("updates the view for DatagridDisplayMode.CALCULATE", function() {
+            const viewObservable = this.displayService.view;
+            let currentView = null;
+            viewObservable.subscribe(viewChange => {
+                currentView = viewChange;
+            });
+            expect(currentView).not.toBe(DatagridDisplayMode.CALCULATE);
+            this.displayService.updateView(DatagridDisplayMode.CALCULATE);
```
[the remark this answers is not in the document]
hippee-lee: Yes, will update with same from above.

src/clr-angular/data/datagrid/datagrid-items.spec.ts
hippee-lee: I updated this test to test the `testContent` of the test element.

src/clr-angular/data/datagrid/render/render-organizer.ts
hippee-lee: Will remove.

src/clr-angular/data/datagrid/providers/table-size.service.ts
hippee-lee: This is in there because of how hovering and the expandable row columns is behaving. Without an explicit width, the content / hover area is limited to whats visible even when the columns are wide enough for overflow on the x-axis.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -18,56 +20,50 @@ import {
...
-            }
-        }
-    }
-
-    @Input("clrDgItemsTrackBy")
-    set trackBy(value: TrackByFunction<any>) {
-        this._items.trackBy = value;
+        // TODO: REMOVE WHEN THIS IS UPGRADED TO ANGULAR 6.1 (consumers will have compile errors)
+        changes.ngForOf = new SimpleChange(undefined, this._items.displayed, true);
+        this.iterableProxy.ngOnChanges(changes);
```
youdz: Depending on which Angular versions we have to support, there are ways for us to do this and still support future versions without throwing Typescript errors. Mostly has to do with casting to `any` and testing if `this.iterableProxy` has an `ngOnChanges` method.

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -18,56 +20,50 @@ import {
...
     ngDoCheck() {
+        if (!this._differ) {
+            this._differ = this._differs.find(this._rawItems).create(this.iterableProxy.ngForTrackBy);
+        }
         if (this._differ) {
             const changes = this._differ.diff(this._rawItems);
             if (changes) {
-                // TODO: not very efficient right now,
-                // but premature optimization is the root of all evil.
                 this._items.all = this._rawItems;
```
youdz: Do you mind keeping this todo? I want this to raise a flag if we ever run into performance issues with a large number of items.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -206,20 +206,24 @@ export class ClrDatagridRow implements AfterContentInit, AfterViewInit, OnDestro
     ngAfterViewInit() {
         this.subscriptions.push(this.displayMode.view.subscribe(viewChange => {
             // Listen for view changes and move cells around depending on the current displayType
-            while (this.scrollableCells.detach()) {
-            }  // remove cells containers
-            while (this.calculatedCells.detach()) {
-            }  // remove cells containers
+            // remove cell views from display view
+            for (let i = this.scrollableCells.length; i > 0; i--) {
```
youdz: ❤️ 🙈

src/clr-angular/data/datagrid/datagrid-items.ts
hippee-lee: Yes. WIll keep.

src/clr-angular/data/datagrid/providers/table-size.service.spec.ts
hippee-lee: Will change this to make sure this.sizeService.tableRef is defined.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
hippee-lee: I'm not sure what I was thinking. I'll put it back and updated the test component.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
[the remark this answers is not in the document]
hippee-lee: After discussion - we will change the default (large) size to the `spinner-md` size.

src/clr-angular/data/datagrid/datagrid.module.ts
```diff
@@ -42,31 +43,31 @@ import {ClrDatagridComparatorInterface} from "./interfaces/comparator.interface"
...
-    DatagridMainRenderer, DatagridTableRenderer, DatagridHeadRenderer, DatagridHeaderRenderer, DatagridBodyRenderer,
-    DatagridColumnResizer, DatagridRowRenderer, DatagridCellRenderer,
+    DatagridMainRenderer, DatagridHeaderRenderer, DatagridColumnResizer, DatagridRowRenderer, DatagridCellRenderer,
 
     // Chocolate
     DatagridWillyWonka, ActionableOompaLoompa, ExpandableOompaLoompa,
 
     // Animation hack
-    DatagridRowExpandAnimation,
+    // DatagridRowExpandAnimation, // TODO: fix animation issues related to render refactor
```
gnomeontherun: Is this to be done in this PR or later?

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -16,60 +16,66 @@ import {Page} from "./providers/page";
...
      * Subscription to the Page service for page changes.
      * Note: this only emits after the datagrid is initialized/stabalized and the page changes.
      */
     ngOnInit() {
+        /*
+         * Default page size is 10.
+         * The reason we set it in this constructor and not in the provider itself is because
+         * we don't want pagination (page size 0) if this component isn't present in the datagrid.
+         */
+        if (this.defaultSize) {
```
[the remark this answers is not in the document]
gnomeontherun: Do you have an update on this @hippee-lee?

src/clr-angular/data/datagrid/datagrid-items.ts
```diff
@@ -3,50 +3,62 @@
...
+            this.iterableProxy.ngForOf = items;
+            this.iterableProxy.ngDoCheck();
+        });
     }
 
-    @Input("clrDgItemsTrackBy")
-    set trackBy(value: TrackByFunction<any>) {
-        this._items.trackBy = value;
+    ngOnChanges(changes: SimpleChanges): void {
+        // TODO: REMOVE WHEN THIS IS UPGRADED TO ANGULAR 6.1 (consumers will have compile errors)
```
gnomeontherun: When we do this, does that mean we'll require Angular 6.1+? Thats kind of a breaking change?

src/clr-angular/data/datagrid/datagrid-cell.ts
```diff
@@ -15,7 +28,7 @@ import {HideableColumnService} from "./providers/hideable-column.service";
     `,
     host: {"[class.datagrid-cell]": "true", "[class.datagrid-signpost-trigger]": "signpost.length > 0"}
 })
-export class ClrDatagridCell {
+export class ClrDatagridCell implements OnInit {
```
gnomeontherun: add OnDestroy too I guess

src/clr-angular/data/datagrid/wrapped-cell.ts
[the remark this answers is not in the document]
hippee-lee: Sorry, missed it. I will update this.

src/clr-angular/data/datagrid/datagrid-items.ts
hippee-lee: Since it is acting as a proxy for `NgForOf` this note is there to remind me to remove it when the Clarity Angular version moves ahead of angular 6.1. I believe it that is the version this was added: https://github.com/angular/angular/pull/23378 and I see it in the `6.1.0-beta.0` tag: https://github.com/angular/angular/commits/6.1.0-beta.0

src/clr-angular/data/datagrid/datagrid-cell.ts
hippee-lee: Will add.

PR #2516, opened 2018-08-03
Website forms docs updates

gnomeontherun:
This adds the UX docs for forms, and the password docs.


latest/src/app/documentation/demos/forms/forms-do-block.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) 2016 - 2017 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import {Component, Input} from "@angular/core";
+
+@Component({
+    selector: "clr-forms-do-block",
+    styles: [
+        `
```
Shijir: why is it empty style like this?

latest/src/settings/componentlist.json
```diff
@@ -284,6 +284,16 @@
             "ng": 5,
             "dox": 5
         },
+        {
```
Shijir: Should we need to add `"newLayout": true` as it's updated?

Issue #2493, opened 2018-07-25
[Edge] Wizard Navigation shorter than Wizard


Gesh4o:
Hello guys,

**I am submitting a ...** (check one with "x")

```
[*] bug
[ ] feature request
[ ] enhancement
```

I have a problem when working with wizard pages in **Edge (W10)** - the **navigation side** of the wizard is **shorter** than the **wizard** itself which results into the following:
<img width="1680" alt="screen shot 2018-07-25 at 3 09 07 pm" src="https://user-images.githubusercontent.com/13942085/43200679-f735d9ec-901e-11e8-94b2-dc3a228a3a54.png">

### Steps to reproduce

1. Download the example - https://stackblitz.com/edit/clarity-light-theme-v11-z4ibyz
2. Run it **locally** (since with stackblitz on Edge you may not)
3. **Minimize** the browser **height** until the above error occurs 

### Root cause/fix (friendly suggestion)
There is this selector which controls the height: 
```css
.clr-wizard.wizard-xl .clr-wizard-stepnav-wrapper, .clr-wizard.wizard-xl .modal-content {
    max-height: 75vh
}
```

If I disable the above rule it will look like this:

<img width="1680" alt="screen shot 2018-07-25 at 3 37 06 pm" src="https://user-images.githubusercontent.com/13942085/43201265-ab36f510-9020-11e8-9a0b-70e1531f4e1c.png">

**NOTE**: As it is said this is just a **suggestion** - I assume there must be reason for this selector to exist but I believe that this is a good starting point.

**NOTE**: Keep in mind that the page that is shown in the example overflows but in the domain I work I can reproduce it on almost every wizard and with no matter of page size.

### Environment details

* **Angular version:**
4.x.x 

* **Clarity version:**
0.11.*
* **OS and version:**
Windows 10

* **Browser:** [ Edge  ]

mathisscott:
@Gesh4o 
Thanks for catching this. That CSS is there to prevent the wizard for expanding beyond all sense and reason on large screens. We should guard against this styling on smaller screens.

PR #2518, opened 2018-08-06
[BUG] Odd wizard vertical height issue when resizing browser

mathisscott:
• This was caused by a combination of justify-content flexbox settings that Microsoft browsers didn't care for complicated by Microsoft browsers being confused over the max-height vs. height settings when the content in either the stepnav or content panel exceeded the available height

• I did some whack-a-mole through the SASS to find the guilty parties and removed/amended them

• I'm marking this as Low priority and only back porting it to 0.12 because it only happens on XL wizards when the vertical height of the screen goes below a certain range determined by the height of EITHER the stepnav or the content area

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ Safari
✔︎ IE11
✔︎ Edge

Closes: #2493

### DIFF
--- a/src/clr-angular/wizard/_wizard.clarity.scss
+++ b/src/clr-angular/wizard/_wizard.clarity.scss
@@ -11,11 +11,8 @@
 @include exports('wizard.clarity') {
   .clr-wizard {
     .modal-dialog {
-      // removed for IE10... maybe remove altogether? or only remove for IE10?
-      // position: static;
       display: flex;
       flex-direction: column;
-      justify-content: space-between;
       align-items: center;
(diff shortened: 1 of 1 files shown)


src/clr-angular/wizard/_wizard.clarity.scss
```diff
@@ -141,6 +136,7 @@
       font-size: 0.583333rem;
       color: $clr-wizard-sidenav-text;
       width: 100%;
+      flex: 1 1 auto;
```
hippee-lee: What about putting this next to the other flexbox properties>

Issue #2500, opened 2018-07-27
Passing a value to clrDgSingleSelected  doesn't make the row appear as selected.


dennitsa:
```
[ X] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Passing value into clrDgSingleSelected should make the row appear selected.

### Actual behavior
Passing a value to clrDgSingleSelected  doesn't make the row appear as selected.

My workaround for the time being is to use v0.11.21, where this is not happening, but this definitely needs a fix.

### Reproduction of behavior

This is reproducible in v0.11.25( See https://stackblitz.com/edit/clarity-light-theme-v11-rakdgv)
As well as in the latest v0.12.5 (See https://stackblitz.com/edit/clarity-light-theme-v11-nbupfw)

### Environment details

* **Angular version:** 5.x.x & 6.x.x

* **Clarity version: ^v0.11.25

* **OS and version: Any

* **Browser:** [all]

youdz:
To help out whoever is going to pick this up, I'm copy-pasting here what I wrote in the issue this was split from:
> I remember we had a very similar issue a while back which we solved with debouncing. The debouncing became useless around 0.10.x, but it seems like latest refactors might have brought the issue back. Your bug clearly needs proper investigation, so creating a separate issue for it means we can consider it in our planning sessions. 👍

I think the cause of the issue is that due to the default instantiation order of our components by Angular, the current selection is set before the list of items and their `trackBy` function is available to us, and when we finally receive the list of items it ends up canceling the pre-selection because it could not find the correct item.

ivanbest:
Hi team, this impacts my team too. Could you prioritize accordingly ?

jordanovi:
Hi team, this impacts my team too. Could you prioritize accordingly ?

dkanev:
I hit the same issue. The  Selection service (https://github.com/vmware/clarity/blob/master/src/clr-angular/data/datagrid/providers/selection.ts) emits undefined.
See:
https://stackblitz.com/edit/clarity-light-theme-v11-26gr

Is there a workaround for this?

PR #2519, opened 2018-08-06
[ng] allow pre-selection of currentSingle when `all` hasn't been set

jeeyun:
This is a fix for #2500. 

Here are my observations of current (broken) behavior:
- In the bug scenario, setting of the `clrDgSingleSelected` happens before setting of the `clrDgItems`.
- Usually, setting the `clrDgSingleSelected` updates the `prevSingleSelectionRef` IF the datagrid's `all` (i.e. `clrDgItems`) has been defined. But since this hasn't happened yet, `prevSingleSelectionRef` remains undefined.
- The setting of the `clrDgItems` triggers an update, where our code looks for the `newSingle` by comparing it against `prevSingleSelectionRef`. Since `prevSingleSelectionRef` remains undefined, this updates the `newSingle` to `undefined`. This overrides the `clrDgSingleSelected` that was set by the user in the beginning.

Proposed fix:
- We should only update `newSingle` to `undefined` if 1. It's a smart datagrid where all data is present, and 2. if the selection changed from defined (with a set value) to undefined.
- Check for condition 1 is already there; I just made sure the check for condition 2 is correct with this bug fix.

I think the comment is pretty self explanatory, but let me know if I could improve its clarity. I added unit tests for both single and multi selection to account for above scenario. Multi selection test passes before and after this bug fix. The single selection test only passes with this bug fix.

### DIFF
--- a/src/clr-angular/data/datagrid/providers/selection.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.ts
@@ -49,6 +49,14 @@ export class Selection<T = any> {
             const trackBy: TrackByFunction<T> = this._items.trackBy;
             let selectionUpdated: boolean = false;
 
+            // if the currentSingle has been set before data was loaded, we look up and save the ref from current data set
+            if (this.currentSingle && !this.prevSingleSelectionRef) {
+              if (this._items.all && this._items.trackBy) {
+                const lookup = this._items.all.findIndex(maybe => maybe === this.currentSingle);
+                this.prevSingleSelectionRef = this._items.trackBy(lookup, this.currentSingle);
+              }
+            }
+
             updatedItems.forEach((item, index) => {
               const ref = trackBy(index, item);
               // If one of the updated items is the previously selectedSingle, set it as the new one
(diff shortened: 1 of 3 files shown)


src/dev/src/app/datagrid/selection-single/selection-single.html
```diff
@@ -126,7 +126,7 @@ <h3>Server-driven, trackBy item</h3>
   <p class="card-text username-list">
     Selected user:
     <em *ngIf="!trackByIdSingleSelected">No user selected.</em>
-    <span class="username" *ngIf="trackByIdSingleSelected">{{trackByIdSingleSelected.name}}</span>
+    <span class="username" *ngIf="trackByIdSingleSelected">{{trackByIdServerSingleSelected.name}}</span>
```
gnomeontherun: You need to update the other properties here on lines 128-129 to `trackByIdServerSingleSelected`

jeeyun: Thank you!

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -49,6 +49,14 @@ export class Selection<T = any> {
             const trackBy: TrackByFunction<T> = this._items.trackBy;
             let selectionUpdated: boolean = false;
 
+            // if the currentSingle has been set before data was loaded, we look up and save the ref from current data set
+            if (this.currentSingle && !this.prevSingleSelectionRef) {
+              if (this._items.all && this._items.trackBy) {
+                const lookup = this._items.all.findIndex(maybe => maybe === this.currentSingle);
+                this.prevSingleSelectionRef = this._items.trackBy(lookup, this.currentSingle);
```
youdz: I was about to ask you to change something, but then I realized this actually uses `trackBy` with index `-1` if we can't find the exact same reference as the pre-selected item in the overall list. That's perfect, it allows us to support proper trackBy (like by id) even for pre-selection.

Nothing to change here.

Issue #2152, opened 2018-04-05
Remove wizard ghost pages


gnomeontherun:
Remove the deprecated wizard ghost pages features.

* [ ] Code
* [ ] Docs if necessary
* [ ] List in release notes

PR #2531, opened 2018-08-09
[ng] Remove wizard ghost pages (#2152)

jeeyun:


src/clr-angular/modal/modal.ts
```diff
@@ -41,36 +39,6 @@ import { GHOST_PAGE_ANIMATION } from './utils/ghost-page-animations';
       transition('void => *', [style({ opacity: 0 }), animate('0.2s ease-in-out', style({ opacity: 0.85 }))]),
       transition('* => void', [animate('0.2s ease-in-out', style({ opacity: 0 }))]),
     ]),
```
mathisscott: do we care about trailing commas in arrays?

src/clr-angular/wizard/wizard.spec.ts
```diff
@@ -19,26 +19,6 @@ import { ClrWizard } from './wizard';
...
-        xit('should set proper state on last page', () => {});
-
-        xit('should set proper state when on next to last page', () => {});
-
-        xit('should fall through to proper state', () => {});
-
-        xit('should set expected state when first page is last page', () => {});
-
-        xit('should set expected state when first page is next to last page', () => {});
-      });
```
mathisscott: awesome. unit tests FTW!

src/clr-angular/modal/modal.html
```diff
@@ -31,11 +31,6 @@
                     <ng-content select=".modal-footer"></ng-content>
                 </div>
             </div>
-            <!--todo: deprecate the modal-ghost-wrapper div below after 0.12-->
-            <div class="modal-ghost-wrapper">
-                <div [@ghostPageOneState]="ghostPageState" class="modal-ghost modal-ghost-1"></div>
-                <div [@ghostPageTwoState]="ghostPageState" class="modal-ghost modal-ghost-2"></div>
-            </div>
```
mathisscott: i thought the DOM edits would be more severe than this. i wonder if the outer-wrapper is still necessary?

gnomeontherun:
Do we want to include a schematic that scans for ghost pages and alerts the user they are removed?

mathisscott:
That's a good suggestion, @gnomeontherun

src/clr-angular/modal/modal.html
jeeyun: I'll investigate if the removal of outer-wrapper can be done without breaking anything.

src/clr-angular/modal/modal.ts
jeeyun: At times I take them off, but I've seen them in our code base elsewhere too so I assume we don't care?

mathisscott: I don't think the browsers we support have a problem with them. Still feels messy. I've already approved, so my feelings on it are clear. It pings my OCD. But not worth blocking progress.

hippee-lee: This is something we could enforce / standardize the codebase to with prettier: https://prettier.io/docs/en/options.html#trailing-commas

gnomeontherun:
We can do the schematic later, just want to track it. I just merged #2526 so the schematics are in place to extend easily. I don't want to put the effort into building the actual migration for this, just scan for some feature of ghost pages that is easy to find and show a warning message.

gnomeontherun:
I'm fine with the schematic coming later, I can make a PR for it and use that for devshare content.

jeeyun:
@gnomeontherun I'll leave it up to you for the schematic change for ghost pages. Merging this to master.

PR #2546, opened 2018-08-14
[ng] Remove OK/Apply button from column hide/show panel in datagrid

jeeyun:

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
@@ -19,21 +19,13 @@ export default function(): void {
         component = new ClrDatagridColumnToggleButton(columnToggleButtons);
       });
 
-      it('knows if the button isOk()', function() {
-        component.clrType = 'ok';
-        expect(component.isOk()).toBe(true);
-        component.clrType = 'selectAll';
-        expect(component.isOk()).toBe(false);
-      });
-
-      it('gets the correct classes', function() {
-        component.clrType = 'ok';
-        expect(component.getClasses()).toEqual('btn btn-primary');
-        component.clrType = 'selectAll';
-        expect(component.getClasses()).toEqual('btn btn-sm btn-link p6 text-uppercase');
-      });
-
       it('calls the click method', function() {
+        spyOn(columnToggleButtons, 'buttonClicked');
+        component.click();
+        expect(columnToggleButtons.buttonClicked).toHaveBeenCalledWith(component.clrType);
+      });
+
+      it('calls the click method when clrType is `ok`', function() {
         component.clrType = 'ok';
         spyOn(columnToggleButtons, 'buttonClicked');
         component.click();
@@ -55,9 +47,6 @@ export default function(): void {
 
       it('has a button', function() {
         expect(button).toBeDefined();
-        expect(button.className).toContain('btn btn-primary');
-        context.testComponent.type = 'selectAll';
-        context.detectChanges();
         expect(button.className).toContain('btn btn-sm');
       });
 
@@ -81,5 +70,5 @@ export default function(): void {
     `,
 })
 class ButtonTest {
-  type = 'ok';
+  type = 'selectAll';
 }
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
@@ -10,34 +10,19 @@ import { ColumnToggleButtons, ColumnToggleButtonsService } from './providers/col
 @Component({
   selector: 'clr-dg-column-toggle-button',
   template: `
-        <button
+        <button class="btn btn-sm btn-link"
             (click)="click()"
-            [disabled]="toggleButtons.selectAllDisabled && !isOk()"
-            [ngClass]="getClasses()"
+            [disabled]="toggleButtons.selectAllDisabled"
             type="button">
             <ng-content></ng-content>
         </button>
     `,
-  host: { '[class.action-right]': 'isOk()' },
 })
 export class ClrDatagridColumnToggleButton {
-  @Input() clrType: ColumnToggleButtons;
+  /** @deprecated since 0.13 */
+  @Input() clrType: ColumnToggleButtons = 'selectAll';
 
   constructor(public toggleButtons: ColumnToggleButtonsService) {}
-
-  getClasses() {
-    let classes = 'btn ';
-    if (this.isOk()) {
-      classes += 'btn-primary';
-    } else {
-      classes += 'btn-sm btn-link p6 text-uppercase';
-    }
-    return classes;
-  }
-
-  isOk() {
-    return this.clrType === 'ok';
-  }
 
   click() {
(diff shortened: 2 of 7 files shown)


src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
```diff
@@ -10,35 +10,19 @@ import { ColumnToggleButtons, ColumnToggleButtonsService } from './providers/col
 @Component({
   selector: 'clr-dg-column-toggle-button',
   template: `
-        <button
+        <button class="btn btn-sm btn-link p6 text-uppercase"
```
mathisscott: Why did we add `.p6` and `.text-uppercase` here?

Aren't those implicit in the styling of a `.btn-sm`?

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
```diff
@@ -10,35 +10,19 @@ import { ColumnToggleButtons, ColumnToggleButtonsService } from './providers/col
...
+            [disabled]="toggleButtons.selectAllDisabled"
             type="button">
             <ng-content></ng-content>
         </button>
     `,
-  host: { '[class.action-right]': 'isOk()' },
 })
 export class ClrDatagridColumnToggleButton {
-  @Input() clrType: ColumnToggleButtons;
+  @Input() clrType: ColumnToggleButtons = 'selectAll';
```
mathisscott: It seems weird to have this defined as a type `ColumnToggleButtons` and then hardcoded as the `selectAll` string. Can `clrType` ever be something that _isn't_ `"selectAll"`?

If it can, then it seems we would have a typing issue. If not, it seems weird to have it be an input.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
```diff
@@ -7,28 +7,20 @@ import { Injectable, TemplateRef } from '@angular/core';
...
-  }
-
   private _selectAllButtonClicked = new Subject<void>();
   public get selectAllButtonClicked(): Observable<void> {
     return this._selectAllButtonClicked.asObservable();
   }
 
   public buttonClicked(type: ColumnToggleButtons): void {
     switch (type.toLowerCase()) {
-      case 'ok':
```
mathisscott: Waaaat?

A switch statement with only one case? Why?

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
jeeyun: Discussed this briefly with the dev team. We thought it'd be good to preserve backward compatibility even though we're removing the other type so that this is still a valid mark up:

`<clr-dg-column-toggle-button clrType="selectAll">Alle auswählen!</clr-dg-column-toggle-button>`

As well as one without which would default to the selectAll type:
`<clr-dg-column-toggle-button>Alle auswählen!</clr-dg-column-toggle-button>`

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
jeeyun: Same comment as above.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
jeeyun: I copied over the class we had for the "select all" button. Let me evaluate if those classes (`.p6` and `.text-uppercase` are really needed.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
youdz: What Jeeyun said. Please just make sure to:
- Add an `@deprecated` JSDoc annotation,
- Remove the `clrType` completely from the docs.

gnomeontherun:
This is intended as a non-breaking change right? If so, have we tested in a demo with the `clrType="ok"` still enabled to see we don't get any errors (and on --prod builds)?

jeeyun:
@gnomeontherun actually I think this would be considered a breaking change (both visually and also in that it won't support `clrType="ok"` going forward. Is this not what we want?

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
```diff
@@ -7,28 +7,20 @@ import { Injectable, TemplateRef } from '@angular/core';
 import { Observable } from 'rxjs';
 import { Subject } from 'rxjs';
 
-export type ColumnToggleButtons = 'ok' | 'selectAll';
```
youdz: Oh right, I missed this. We want to still allow for `'ok'` and just ignore it. Not project it, not throw an error, it's just dead HTML on their side. That way it's non-breaking, they can remove it later.

Of course, removing it from the docs like you did in this demo. 👍

src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
```diff
@@ -19,22 +19,7 @@ export default function(): void {
         component = new ClrDatagridColumnToggleButton(columnToggleButtons);
       });
 
-      it('knows if the button isOk()', function() {
```
gnomeontherun: I'd like to have a test here that shows that an `ok` button type is inert and doesn't throw an error.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
```diff
@@ -10,35 +10,20 @@ import { ColumnToggleButtons, ColumnToggleButtonsService } from './providers/col
 @Component({
   selector: 'clr-dg-column-toggle-button',
   template: `
-        <button
+        <button class="btn btn-sm btn-link"
```
gnomeontherun: We are losing the `text-uppercase` class here, is it not needed?

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
gnomeontherun: 👍

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
gnomeontherun: Should be a basic `if (type.toLowerCasee() === 'selectall') { ... }`

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
youdz: Nope, both uppercase an `p6` styles are already part of button styles.

src/clr-angular/data/datagrid/providers/column-toggle-buttons.service.ts
jeeyun: I think that is better. Will update.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.spec.ts
jeeyun: Good idea. Will add this test back in and modify.

PR #2551, opened 2018-08-15
[NG] fix tabs role from presentation to tab

gnomeontherun:
Our tabs were previously given a `role="presentation"` treatment instead of `role="tab"` which flags aria scanners and is incorrect.


gnomeontherun:
@Shijir I added a couple tests and also made the changes for wizards

Shijir:
@gnomeontherun Great! Thanks for doing that 👍

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -85,7 +85,7 @@ export class ClrTabLink {
 
   @HostBinding('attr.role')
   get role(): string {
-    return 'presentation';
+    return 'tab';
```
youdz: Note: We have a [guideline](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#code-style) to put host bindings that are *not* dynamic as part of the `host` config in the component metadata, and use `@HostBinding` for the ones that are dynamic.

This component is doing the exact opposite. You can either fix this now, or we make a note of it somewhere either in comments or as a new issue.

src/clr-angular/layout/tabs/tabs.spec.ts
```diff
@@ -132,6 +132,10 @@ describe('Tabs', () => {
       context.fixture.destroy();
     });
 
+    it('sets the tab role on buttons', () => {
+      expect(compiled.querySelector('button.nav-link').getAttribute('role')).toEqual('tab');
```
youdz: You're only checking the first `button.nav-link` here, based on the description I don't think that's what you wanted to do.

I'm guessing you want to test that the overflow toggle has the tab role, in which case you have the wrong selector.

gnomeontherun:
I'm rethinking this a bit, so please do not merge for the moment.

src/clr-angular/layout/tabs/tabs.spec.ts
gnomeontherun: Actually no, this is checking that the buttons get the correct role, not the overflow button. I've added another test to check the overflow button's role.

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -25,14 +24,16 @@ let nbTabLinkComponents: number = 0;
 @Directive({
   selector: '[clrTabLink]',
   host: {
-    '[id]': 'tabLinkId',
+    id: 'tabLinkId',
```
youdz: Doesn't that mean the id is now `"tabLinkId"` as a string, instead of the actual dynamic id?

gnomeontherun:
Ugh, I didn't mean to keep that change. Fixing up. 

Also tested with screen reader. The quirk that I see is that if we hide tabs in an overflow that it only tells you there are 2 tabs instead of 4. However, I think thats a different problem from this.

Issue #2147, opened 2018-04-05
Clean up deprecated icons


gnomeontherun:
Remove the previously announced deprecated icons and clean up as necessary.

* [ ] Code
* [ ] Website

Shijir:
The milestone is correctly set. We have a few deprecated icons in 0.12 that can be removed once we release 0.13.

PR #2552, opened 2018-08-15
[ICONS] Clean up deprecated icons (#2147)

jeeyun:
Removing icons that were deprecated in 0.12. Leaving a couple that are slated for removal at 1.0.


src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -1168,7 +1163,6 @@ export const TechnologyShapes: any = {
   'process-on-vm': ClrShapeProcessOnVM,
 };
 
-Object.defineProperty(TechnologyShapes, 'analytics', descriptorConfig(TechnologyShapes['line-chart']));
```
Shijir: @jeeyun can you make sure we have this `"analytics"` alias in the `"chart-shapes"`? We need to remove the deprecated icons, but we shouldn't remove their aliases. The aliases should go into their new sets too.

jeeyun: Yeah I have already added it in this PR! :)

Shijir: oops I missed that sorry.

Shijir: And thanks again working on this :)

Issue #2143, opened 2018-04-05
Remove push/pull from old grid


gnomeontherun:
We need to remove the push/pull options from the old grid, as per migration strategy.

* [ ] Remove code
* [ ] Documentation
* [ ] `ng update` migration tooling

PR #2559, opened 2018-08-17
[UI] Remove push/pull from old grid (#2143)

jeeyun:
Let me know if I should be removing all demo code from deprecated grid (and not just push/pull).


gnomeontherun:
I remembered there are some gemini tests for this that need to be removed.

PR #2567, opened 2018-08-21
[UI] make progress element block element

Shijir:


youdz:
@hippee-lee: The old height was a bug and did not respect the spec, I believe. Or height set in CSS was ignored because the element were inline elements.

PR #2579, opened 2018-08-23
[NG] Schematics update for importing deprecated non-Clr classes

jeeyun:
I'm not very great with RegEx so any suggestion to improve it is greatly welcome. Basically the goal was to catch any imports from `@clr/angular` where the classname didn't start with `Cl`. I used `Cl` instead of `Clr` for simplicity so that `ClarityModule` would be included in that.

Here are some of the test strings I've used for the regex:
```
import { foo } from '@clr/angular';

import { 
	foo 
} from '@clr/angular';

import { 
	foo,
	bar,
} from '@clr/angular';

import { foo, bar, baz, bat } from '@clr/angular';

import { clrfoo, bar, baz, bat } from '@clr/angular';

import { foo, clrbar, baz, bat } from '@clr/angular';

import { foo, bar, clrbaz, bat } from '@clr/angular';

import { foo, bar, baz, clrbat } from '@clr/angular';

import { clrfoo, clrbar, baz, bat } from '@clr/angular';

```

### DIFF
--- a/src/schematics/src/update/update-0.13.0.ts
+++ b/src/schematics/src/update/update-0.13.0.ts
@@ -21,6 +21,39 @@ Start changes for Clarity 0.13.0
       messageFlags[message] = true;
     }
 
+    function getNewClassname(oldClassname: string) {
+      switch (oldClassname) {
+        case 'Comparator':
+          return 'ClrDatagridComparatorInterface';
+        case 'Filter':
+          return 'ClrDatagridFilterInterface';
+        case 'SortOrder':
+          return 'ClrDatagridSortOrder';
+        case 'State':
+          return 'ClrDatagridStateInterface';
+        case 'StringFilter':
+          return 'ClrDatagridStringFilterInterface';
+        case 'Checkbox':
+          return 'ClrCheckboxDeprecated';
+        case 'DatagridHideableColumnDirective':
+          return 'ClrDatagridHideableColumn';
+        case 'NavLevelDirective':
+          return 'ClrNavLevel';
+        case 'TabLinkDirective':
+          return 'clrTabLink';
+        case 'WizardPageButtonsDirective':
+          return 'ClrWizardPageButtons';
+        case 'WizardPageHeaderActionsDirective':
+          return 'ClrWizardPageHeaderActions';
+        case 'WizardPageNavTitleDirective':
+          return 'ClrWizardPageNavTitle';
+        case 'WizardPageTitleDirective':
+          return 'ClrWizardPageTitle';
+        default:
+          return 'Clr' + oldClassname;
+      }
+    }
+
     // Recurse through the directory scanning and changing text as we go
     function replaceContent(path: string) {
       const dir = tree.getDir(path);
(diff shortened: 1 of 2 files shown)


gnomeontherun:
I'm wondering if we can actually do this transformation for them. I may tinker a bit more with this, but it would require thinking about it differently so we can capture each individual error and replace in the entire document. We can chat later.

jeeyun:
@gnomeontherun sure, I'm open to ideas on this. Replace might be tricky depending on the case too. Will hold off on this one.

youdz:
It's actually not that hard, you can get the nice replace behavior with 2 regex much, much simpler than the one you have.
1. Exec `/import\s*\{\s*([^}]*)\}\s*from\s*['"]@clr\/angular['"]/g` for as long as you find matches.
2. For each match, get the captured group which is the comma-separated list of all classes imported in that statement.
3. On the captured group, replace `/\b(?!Cl)\w+\b/g` by `"Clr$&"`.

That's it, and it even sounds more readable than that giant regex. Please forgive potential typos in the regexps, I type them live here without testing. 😛

src/schematics/src/update/update-0.13.0.ts
```diff
@@ -76,6 +76,17 @@ Start changes for Clarity 0.13.0
             );
           }
 
+          // Change non-Clr prefixed classes to include the Clr prefix.
+          // We are using Cl instead of Clr in the regex for simplicity, because we export
+          // `ClarityModule,` which would be covered by this regex.
+          const reg = /import\s*{([^}]*)}\s*from\s*['"]@clr\/angular['"]/g;
+          let result = reg.exec(updated);
+          while (result !== null) {
+            const newString = result[1].replace(/\b(?!Cl)\w+\b/g, 'Clr$&');
```
gnomeontherun: This doesn't handle the cases where we renamed like the datagrid interfaces. We need to have a list of those known name changes (mostly Datagrid interfaces, enums, but can't recall if there were others) and ensure we catch those.

src/schematics/src/update/update-0.13.0.ts
```diff
@@ -76,6 +76,17 @@ Start changes for Clarity 0.13.0
...
+            updated = updated.replace(result[1], newString);
```
gnomeontherun: Yes, as @youdz says we're not capturing all instances from the file. What we should be doing is for any match in the import that is missing `Clr`, we should search for it and replace it in the whole file.

src/schematics/src/update/update-0.13.0.ts
```diff
@@ -76,6 +107,55 @@ Start changes for Clarity 0.13.0
...
+          let result = reg.exec(updated);
+          while (result !== null) {
+            const classRegex = /\b(?!Cl)\w+\b/g;
+            let importedClass = classRegex.exec(result[1]);
+
+            while (importedClass !== null) {
+              // Check for ones that were renamed
+              if (
+                importedClass[0].search(
+                  /Comparator|Filter|State|StringFilter|Checkbox|DatagridHideableColumnDirective|NavLevelDirective|TabLinkDirective|WizardPageButtonsDirective|WizardPageHeaderActionsDirective|WizardPageNavTitleDirective|WizardPageTitleDirective/g
```
youdz: You can easily avoid this:
- In `getNewClassName`, make the default:
```typescript
default:
  return 'Clr'+oldClassname;
```
This makes it so that your function works for both hardcoded renames and generic prefixing.
- Remove this first `if()`, keep the other ones.
- In the final `else`, replace
```typescript
updated = updated.replace(re, 'Clr$&');
```
by
```typescript
updated = updated.replace(re, getNewClassname(importedClass[0]));
```

gnomeontherun:
We'll need to change the following (after the other PR is merged)

1. Update the release version of the `package.json` to `0.13.0-rc.1`
2. Update the `version` property of the `migration-collection.json` to 0.13.0-rc.1
3. Reorder so that all transformations happen first, then alerts are at the end

jeeyun:
@gnomeontherun I'll squash, reorder, and update as per your comment after you've had a chance to review my latest changes.

PR #2588, opened 2018-08-27
[NG] Schematics update for using clrType (datagrid show/hide column)

jeeyun:

### DIFF
--- a/src/schematics/src/update/update-0.13.0.ts
+++ b/src/schematics/src/update/update-0.13.0.ts
@@ -76,6 +76,16 @@ Start changes for Clarity 0.13.0
             );
           }
 
+          // Alert to using clrType in hide/show panel in datagrid (https://github.com/vmware/clarity/pull/2546)
+          if (updated.search(/clrType=['|"]selectAll|ok['|"]/g) > -1) {
+            logMessage(
+              'warn',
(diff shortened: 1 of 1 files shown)


src/schematics/src/update/update-0.13.0.ts
```diff
@@ -76,6 +76,16 @@ Start changes for Clarity 0.13.0
             );
           }
 
+          // Alert to using clrType in hide/show panel in datagrid (https://github.com/vmware/clarity/pull/2546)
+          if (updated.search(/clrType/g) > -1) {
+            logMessage(
+              'error',
+              'ACTION REQUIRED: Clarity has removed clrType from the column hide/show panel in datagrid, please remove the attribute from your template.',
```
gnomeontherun: The reality is that we've removed `clrType` _and_ the OK button variant. That means not just removing the attribute, but also removing the OK button.

src/schematics/src/update/update-0.13.0.ts
```diff
@@ -76,6 +76,16 @@ Start changes for Clarity 0.13.0
             );
           }
 
+          // Alert to using clrType in hide/show panel in datagrid (https://github.com/vmware/clarity/pull/2546)
+          if (updated.search(/clrType/g) > -1) {
```
gnomeontherun: This seems too generic, let's make it match something like `clrType="(.*?)"` (not correct regex)

jeeyun: Sounds reasonable. Will update the regex.

Issue #2592, opened 2018-08-28
clr-stack-block - when not expanded keyboard navigation using Tab key shows hidden element


bfortes-vmware:
## Describe the bug

clr-stack-block - when not expanded, then keyboard navigation using Tab key shows hidden element from  stack body and hides stack header

## How to reproduce

Two or more expandable stackblocks clr-stack-block. Some focusable input as nested block content inside. Using keyboard and press "Tab" key several times for navigation.
Expected pressing tab will move focus from one clr-stack-block to next one skipping all hidden content.
Actuall behaviour is that 
(a) focus goes to element inside content of internal nested block
(b) content of header of the block  get's hidden

## Expected behavior

Expected pressing tab will move focus from one clr-stack-block header to next one skipping all hidden (collapsed) content.

## Versions

**App**

* Angular: 1.5
* Node:
* Clarity: 0.11.20

**Device:**

* Type: iMac
* OS: mac OS 10.11.6 El Capitan
* Browser chrome
* Version [e.g. 22]

## Additional notes
None

PR #2604, opened 2018-08-29
[BUG] Fixing keyboard navigation in the stack view

mathisscott:
• previously tabbing through a stack view would bring form fields inside of stack children into focus and into view even when they were hidden
• this change fixes that
• I've also added input fields into our demo application's stack view so that we can test this intermittently

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ Safari
✔︎ IE11
✔︎ Edge

Closes: #2592


src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -42,9 +42,9 @@ import { Component, EventEmitter, HostBinding, Input, OnInit, Optional, Output,
   host: { '[class.stack-block]': 'true' },
   animations: [
     trigger('collapse', [
-      state('true', style({ height: 0 })),
-      transition('true => false', [animate('0.2s ease-in-out', style({ height: '*' }))]),
-      transition('false => true', [style({ height: '*' }), animate('0.2s ease-in-out')]),
+      state('true', style({ height: 0, display: 'none' })),
```
Shijir: @mathisscott have we considered using `*ngIf` on the stack children like this:
```
<div @collapse *ngIf="expanded" class="stack-children" >
```
With this, we just need to make tiny changes to this current animation spec to have the same effect while making the solution even more robust.

With current approach, the solution itself is contained in the animation's spec, which makes animation responsible for hiding the nested content as well. So I think we should have a unit test for it. The unit test will help us to figure out why if we break it while trying to retouch the animation effect itself.

Shijir: Oops too late 😬it's merged in. But it's okay. I had a comment only 😄

mathisscott: @Shijir 
Yes, this is a good point. Ultimately, we would want (I think) to use the clrIfExpanded directive in stack block. But we could do that without a breaking change.

Given that this fix was worthless for the vSphere team if it introduced a breaking change, I went for the lightest and most non-intrusive change I could manage.

PR #2609, opened 2018-08-30
[NG] Internationalizing accessibility

### LINKED ISSUE #2597: [Accessibility] Clarity datagrid VPAT compliant
## Summary

Our product which uses clarity has VPAT assessment report which contains several issues relate to the clarity datagrid.

**Issue 1: The radio buttons on the grids, have labels that are properly associated with the fields, but the LABEL element is empty.**  

**Instance Severity: 10**

claritydatagrid-radiobuttons

**Impact**
When form fields do not explicitly include a label, assistive technologies may incorrectly render the label or provide no label at all to users. When labels are not present or are incorrect, users of assistive technologies may not be able to complete a form.

**Code Reference**
```
<div class="radio">
<input id="clr-dg-row0" class="ng-untouched ng-pristine ng-valid" type="radio">
<label for="clr-dg-row0"></label>
</div>
```

**Propose Solution from VPAT assessment team**
Provide valid textual information within the LABEL element. 

**Code Example**
```
<div class="radio">
<input id="clr-dg-row0" class="ng-untouched ng-pristine ng-valid" type="radio">
<label for="clr-dg-row0">Host Connection</label>
</div>

```

**Issue 2: The only text within each button to expand the row is "angle".** 

**Instance Severity: 10**

datagrid-expandbutton

**Code Reference**
```
<button class="datagrid-expandable-caret-button ng-star-inserted" type="button">
	<clr-icon class="datagrid-expandable-caret-icon" shape="caret" dir="down">
		<svg version="1.1" viewBox="0 0 36 36" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" focusable="false" role="img">
		<title>angle</title>
		<path class="clr-i-outline clr-i-outline-path-1" d="M29.52,22.52,18,10.6,6.48,22.52a1.7,1.7,0,0,0,2.45,2.36L18,15.49l9.08,9.39a1.7,1.7,0,0,0,2.45-2.36Z"></path>
		</svg>
	</clr-icon>
</button>
```

**Propose Solution from VPAT assessment team**
There are several solutions in this particular case including modifying the title, adding aria-label, and others. In the example below, aria-hidden is added to the graphic and a screen reader only text is put in place. 

**Code Example**
```
<style type="text/css">
.sr-only-clip {
  position: absolute !important;
  height: 1px; 
  width: 1px; 
  overflow: hidden;
  padding: 0 !important;
  border: 0! important;
  white-space: nowrap !important;
  clip: rect(1px 1px 1px 1px) !important; /* IE6, IE7 */
  clip: rect(1px, 1px, 1px, 1px) !important;
  clip-path: inset(50%) !important;
}
</style>

<button class="datagrid-expandable-caret-button ng-star-inserted" type="button">
	<clr-icon class="datagrid-expandable-caret-icon" shape="caret" dir="down">
		<svg [...snip...] role="img" aria-hidden="true">
		[... snip ...]
		</svg>
	</clr-icon>
	<span class="sr-only-clip">Expand Statistics Section</span>
</button>

```

Probably the same issues can apply for filter column button, and buttons in the pagination component

Please evaluate how this issues can be address.

youdz:
This is a rough first draft to label every icon button and
every label-less checkbox and radio button we use in our
components with a very generic screenreader-only text.

Because this text has to be internationalized for the
components to be accessible in every language, we use a
provider to declare these labels. That way our consumers can
declare their own "CommonStringsService" to override ours with
the labels they want in any language. And their provider can
obviously inject the app's current locale with Angular so they
can fetch the correct strings dynamically.

Fixes #2597 so it is now VPAT-compliant, even though it's not the perfect UX for users with disabilities. We will hopefully improve this in the future, but this is good enough for an initial fix.

Additional notes:
- For invalid form inputs, the actually error message should be
enough for screen readers. I added `aria-hidden="true"` to them
for now, but I'm open to discussion on this. One more involved
solution would be to use `aria-labelledby` on the exclamation
circle icon to label it with the actual error message. But that
requires trading `id`s and it's out of the scope of this PR.
- No unit tests have been implemented for these labels (except
the alert ones) because we want this to be part of a broader
a11y testing tool.
- Upgrade Typescript version for a cool feature, but realized Angular 6 uses Typescript 2.8+ anyway.
- The CSS change on alerts is to fix a weird screenreader issue
due to the lack of display: block. No idea why.

Side bonus: our dev app now has an awesome URL (which doesn't show on the linux used by our Docker setup).
screen shot 2018-08-30 at 11 16 32


src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -15,17 +15,22 @@ import { DateIOService } from './providers/date-io.service';
...
 import { DatepickerEnabledService } from './providers/datepicker-enabled.service';
 import { LocaleHelperService } from './providers/locale-helper.service';
+import { ClrCommonStrings } from '../../utils/i18n';
 
 @Component({
   selector: 'clr-date-container',
   template: `
         <ng-content></ng-content>
+        <!--
+          Isn't this button supposed to be aria-hidden="true"?
```
mathisscott: No idea. Might be worth another github issue to investigate at a later date.

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -15,7 +15,8 @@
                         <div role="columnheader" class="datagrid-column datagrid-select datagrid-fixed-column"
                              *ngIf="selection.selectionType === SELECTION_TYPE.Multi">
                         <span class="datagrid-column-title">
-                            <clr-checkbox [(ngModel)]="allSelected"></clr-checkbox>
+                            <input clrCheckbox type="checkbox" [(ngModel)]="allSelected"
```
mathisscott: Thanks for updating to the new forms!

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -220,7 +222,7 @@ export default function(): void {
         iconBtn.click();
         context.detectChanges();
 
-        const columnCheckboxes = context.clarityElement.querySelectorAll('.switch-content > li > clr-checkbox > input');
+        const columnCheckboxes = context.clarityElement.querySelectorAll('.switch-content input[type="checkbox"]');
```
mathisscott: Selector much improved here!

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -171,9 +170,7 @@ export default function(): void {
         hideableColumnService.updateColumnList(updatedColumns);
         context.detectChanges();
 
-        const updatedRenderedTemplates = context.clarityElement.querySelectorAll(
-          '.switch-content > li > clr-checkbox > label'
-        );
+        const updatedRenderedTemplates = context.clarityElement.querySelectorAll('.switch-content label');
```
mathisscott: not sure why we went on such a kick with selector specificity a while back...

src/clr-angular/clr-angular.module.ts
```diff
@@ -35,5 +37,6 @@ import { ClrWizardModule } from './wizard/wizard.module';
     ClrPopoverModule,
     ClrWizardModule,
   ],
+  providers: [COMMON_STRINGS_PROVIDER],
```
mathisscott: trailing comma

golden/clr-angular.d.ts
```diff
@@ -194,6 +218,7 @@ export declare class ClrDatagrid<T = any> implements AfterContentInit, AfterView
     SELECTION_TYPE: typeof SelectionType;
     allSelected: boolean;
     columns: QueryList<ClrDatagridColumn<T>>;
+    commonStrings: ClrCommonStrings;
```
mathisscott: Oh wow. This had to be painful. 😬

src/dev/src/app/i18n-a11y/users.ts
```diff
@@ -0,0 +1,188 @@
...
+    name: 'Marcella',
+    creation: '2014-12-06T05:25:58.070Z',
+    pokemon: { number: 57, name: 'Primeape' },
+  },
+  {
+    id: 40230,
+    name: 'Alica',
+    creation: '2016-08-01T21:40:18.966Z',
+    pokemon: { number: 149, name: 'Dragonite' },
+  },
```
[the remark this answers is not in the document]
youdz: Our prettier config has `trailingComma: 'es5',`.

So these are added by prettier, not my choice. 🤷‍♂️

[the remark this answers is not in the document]
mathisscott: wow.

src/clr-angular/data/datagrid/datagrid.html
gnomeontherun: We haven't done the new checkboxes yet...

package-lock.json
```diff
@@ -1898,6 +1898,12 @@
                     "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
                     "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
                     "dev": true
+                },
+                "typescript": {
```
gnomeontherun: This appears to be due to @angular-devkit/core. It would be nice if we can dedupe typescript to have only one instance, or there tends to be build issues. Perhaps we can bump the devkit packages too?

src/clr-angular/data/datagrid/datagrid-footer.ts
```diff
@@ -15,9 +15,10 @@ import { Selection, SelectionType } from './providers/selection';
   template: `
         <ng-container
             *ngIf="(selection.selectionType === SELECTION_TYPE.Multi) && (selection.current.length > 0)">
-            <clr-checkbox [clrDisabled]="true" [clrChecked]="true" class="datagrid-foot-select">
-                {{selection.current.length}}
-            </clr-checkbox>
+            <clr-checkbox-container class="datagrid-foot-select disabled">
```
gnomeontherun: Seems like it might be better to use the CSS only version of this since its just visual?

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -153,6 +155,10 @@ export class ClrTreeNode extends AbstractTreeSelection implements OnDestroy {
     return this.nodeExpand.expanded ? 'down' : 'right';
   }
 
+  public get caretTitle(): string {
+    return this.nodeExpand.expanded ? this.commonStrings.collapse : this.commonStrings.expand;
```
gnomeontherun: We've got this logic in the template in most other places. I'd like it to be consistent either way, though it is nice to have it in the controller for testability.

src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -8,13 +8,20 @@ import { ChangeDetectorRef, Component, EventEmitter, Input, Optional, Output } f
...
 import { AlertIconAndTypesService } from './providers/icon-and-types.service';
 import { MultiAlertService } from './providers/multi-alert.service';
+import { ClrCommonStrings } from '../../utils/i18n';
 
-@Component({ selector: 'clr-alert', providers: [AlertIconAndTypesService], templateUrl: './alert.html' })
+@Component({
+  selector: 'clr-alert',
+  providers: [AlertIconAndTypesService],
+  templateUrl: './alert.html',
+  styles: [':host { display: block; }'],
```
gnomeontherun: What was this for?

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  warning = 'Warning';
+  danger = 'Error';
+  rowActions = 'Available actions';
+  pickColumns = 'Show or hide columns';
+}
+
+export function commonStringsFactory(existing?: ClrCommonStrings): ClrCommonStrings {
+  const defaults = new ClrCommonStringsService();
+  if (existing) {
+    return { ...defaults, ...existing };
```
gnomeontherun: Nicely done.

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  open = 'Open';
+  close = 'Close';
+  show = 'Show';
+  hide = 'Hide';
+  expand = 'Expand';
+  collapse = 'Collapse';
+  more = 'More';
+  select = 'Select';
+  selectAll = 'Select All';
+  previous = 'Previous';
```
gnomeontherun: Wouldn't we want more descriptive terms, like 'Previous page' or 'Close alert'. If so, we wouldn't want to reuse these strings and they would be specific to each icon use. When you translate a string I'm not certain that in each language or situation that it could be generalized. 

I'm thinking we'd want a property for each instance, and be more descriptive by default.

src/clr-angular/data/datagrid/datagrid-footer.ts
youdz: I started doing so, yes, but then I realized I would need to generate a unique ID per datagrid for the footer checkbox, and it felt like a pain.

Unless we consider _not_ linking the checkbox and the label is ok because the checkbox is disabled, but I'm pretty sure a11y tools will complain.

src/clr-angular/data/tree-view/tree-node.ts
youdz: I decided to not refactor anything to keep the PR focused. You'll notice the caret class `this.nodeExpand.expanded ? 'down' : 'right';` is in the template often, and in the controller sometimes. To minimize the footprint of this PR, I went with the convention in place *for that component* whenever I added the caret title.

I'm happy to open an issue to unify all this in a future PR, if you feel it's a clean-up we need to do.

src/clr-angular/emphasis/alert/alert.ts
youdz: I wrote about it in the commit message and description of this PR:
>The CSS change on alerts is to fix a weird screenreader issue
due to the lack of display: block. No idea why.

We always want to do this anyway, we forget about it most the time these days but it leads to very strange styling behavior. The problem is that custom elements (our Angular component hosts) are always `display: inline` by default, when most of them contain block elements.

gnomeontherun: 👍

src/clr-angular/data/tree-view/tree-node.ts
gnomeontherun: That sounds fair, and we can decide to clean up later. I hadn't really thought about this before, might be something for convention guidelines in the future.

src/clr-angular/utils/i18n/common-strings.service.ts
youdz: I considered it a long time as I was implementing this, I agree I'd much rather have more specific labels. It's clearly better UX for people with disabilities.

The problem is that it bloats our i18n API to a point where you have dozens and potentially hundreds of strings in the future, which becomes horrible to provide for our consumers. I think even right now we have around 50 of them, on a first crude pass.

It's still an interesting topic and I would like to find a better solution, whether we implement it right now or postpone it because of the workload. Let's talk about this offline.

Issue #2616, opened 2018-09-04
Allow for tooltips to be used inside button group Angular component


ChrisKaun:
## Describe the bug

When trying to use a tooltip inside a ClarityButton it won't be displayed in Firefox (61.0.2).
In Chrome (68.0.3440.106) the tooltip appears. Couldn't test the stackblitz posted below in Edge (42.17134.1.0) because it didn't load but in our application the tooltips are working with that markup.

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v012-kiyg7a

## Expected behavior

Don't know if that it is supported by clarity to use tooltips inside a button.
Either way the tooltip shouldn't be displayed in Chrome/Edge or it should appear in Firefox so that the browser handle this equaly.

## Versions

**App**

* Angular: 5.2.10
* Node: 9.11.2
* Clarity: 0.11.1

mathisscott:
Hi @ChrisKaun 

Once [this pull request](https://github.com/vmware/clarity/pull/2621) goes in tooltips should work fine in Firefox when using the static HTML/CSS button group.

I've pasted [a working example with the code](https://stackblitz.com/edit/clarity-light-theme-v012-2ycqqu?file=src%2Fapp%2Fapp.component.html) we needed to add to allow for tooltips inside static button groups on stackblitz.

Fixing this for the angular component will be more involved. Because it's Firefox and using tooltips in button groups is an edge case, it will not be high in priority. But I will change this issue to reflect the angular piece that is still ahead of us.

collins1830207:
Hi @mathisscott,

I need to use clr-button-group, how to add tooltip to the buttons in it? 
Here is the code: https://stackblitz.com/edit/clarity-light-theme-v012-xnfx6n
Thanks in advance!

gnomeontherun:
There is a workaround available for this issue, and we recommend using it for Clarity Angular. As we build Clarity Core implementations, we expect that this issue won’t be an issue or will be configurable on the application side. To help us clean up our backlog, we are going to close this with a functional workaround available and suggest you follow updates for Clarity Core for enhancements that can support your use case with Clarity Core components.

PR #2621, opened 2018-09-05
[ENH] Allowing tooltips to be used inside static button groups

mathisscott:
This will enable a workaround for people who want to use tooltips in their button groups. It is a fix specific to Firefox. To enable the angular tooltips in Firefox we would need to do a much more involved refactor of the API.

Tested in:
✔︎ Chrome
✔︎ Firefox

### DIFF
--- a/src/clr-angular/button/button-group/_button-group.clarity.scss
+++ b/src/clr-angular/button/button-group/_button-group.clarity.scss
@@ -2,6 +2,11 @@
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
+@mixin square-off-btngroup-btns($dir: left) {
+  border-top-#{$dir}-radius: 0;
+  border-bottom-#{$dir}-radius: 0;
+}
+
 @include exports('button-group.clarity') {
   .btn-group {
     display: inline-flex;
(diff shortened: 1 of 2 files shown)


mathisscott:
This isn't going to fix #2616 but it will allow people to use static button groups for the time being.

I will post the workaround and change #2616 to be focused on refactoring the angular button-group component.

src/clr-angular/button/button-group/_button-group.clarity.scss
```diff
@@ -10,23 +15,36 @@
...
+        @include square-off-btngroup-btns(right);
       }
     }
 
-    &.btn-primary .btn:not(:last-child),
-    &.btn-success .btn:not(:last-child),
-    &.btn-danger .btn:not(:last-child),
-    &.btn-warning .btn:not(:last-child) {
-      margin: 0 $clr-default-borderwidth 0 0;
+    clr-tooltip {
```
youdz: We don't style Angular components with element selectors, we target the class they add to the host. So here, you want `.tooltip`.

And I just realized we apparently don't have this as part of our coding guidelines, do you mind adding it with this PR? I would say it goes under "Typescript and Angular" > "Code style".

mathisscott: Sure. No problem.

mathisscott:
😂

PR #2631, opened 2018-09-07
Update datepicker to support new forms layout

gnomeontherun:
This change is a special case since we want the datepicker to support both new and old layouts for the time being. I tried to comment where things can be removed, but basically there is a new provider in the new forms module that is used to detect if new forms are used. There is also a new input that allows the datepicker to explicitly set its own display. 

Some of the tests are duplicated from the shared tests, but getting these tests to run in the same way proved to be a time sink. We'll need to refactor the tests sometime to ensure consistency, but would be easier after removal of old layouts.


src/clr-angular/forms/datepicker/all.spec.ts
```diff
@@ -26,7 +26,7 @@ import LocaleHelperServiceSpecs from './providers/locale-helper.service.spec';
 import ViewManagerServiceSpecs from './providers/view-manager.service.spec';
 import YearpickerSpecs from './yearpicker.spec';
 
-describe('Datepicker', function() {
+fdescribe('Datepicker', function() {
```
Shijir: fdescribe

src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -70,6 +131,7 @@ export default function() {
       });
 
       it('projects the date input', () => {
+        context.detectChanges();
```
Shijir: why is this extra change detector here? Is it because it relies on the previous test?

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -28,35 +42,107 @@ import { LocaleHelperService } from './providers/locale-helper.service';
...
+        <ng-content select="clr-control-helper" *ngIf="!invalid"></ng-content>
+        <ng-content select="clr-control-error" *ngIf="invalid"></ng-content>
+      </div>
+    </ng-template>
+    
+    <ng-template #clrDate>
+      <ng-content select="[clrDate]"></ng-content>
+    </ng-template>
+    
+    <ng-container *ngIf="newFormsLayout; then newLayout else oldLayout"></ng-container>
```
Shijir: oh I didn't know this could work like this too 👍

src/clr-angular/forms/datepicker/date-container.spec.ts
gnomeontherun: This is an artifact before I moved the newFormsLayout bit to a beforeEach, can be removed.

src/clr-angular/forms/datepicker/all.spec.ts
gnomeontherun: Yup, was rebasing onto latest to resolve merge conflicts to address this.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -28,35 +42,107 @@ import { LocaleHelperService } from './providers/locale-helper.service';
...
     private _dateNavigationService: DateNavigationService,
     private _datepickerEnabledService: DatepickerEnabledService,
-    private dateFormControlService: DateFormControlService
+    private dateFormControlService: DateFormControlService,
+    private ifErrorService: IfErrorService,
+    private focusService: FocusService,
+    private controlClassService: ControlClassService,
+    @Optional() private layoutService: LayoutService,
+    @Optional()
+    @Inject('IS_NEW_FORMS_LAYOUT')
```
youdz: I feel bad requiring this because it seems like such a small thing, but typing is 👑: can you please switch to an `InjectionToken<boolean>` rather than a plain string, so we get proper typing when injecting this?

PR #2645, opened 2018-09-12
[NG] fix index for radios and rows to be consistent

### LINKED ISSUE #2644: Datagrid (ClrDgSingleSelected) generated radio button id doesn't match row index
## Describe the bug

Currently, the nbRow value is incremented twice for each row ... see the Clarity code snippet below.

Can you make sure the this.radioId matches the row index?

This becomes problematic when writing tests with Selenium and the id is checked ... 

To be clear 
on row 1 the radio button's id = clr-dg-row-rd1
on row 2 the radio button's id = clr-dg-row-rd3
on row 3 the radio button's id = clr-dg-row-rd5
... and so on

Due to this double-increment of the nbRow variable, 
when trying to automate selecting the radio button, 
I have to do silly things to determine what the correct id should be, such as:
row# - 1 + row#

---

src\clr-angular\data\datagrid\datagrid-row.ts

```
  /**
   * Model of the row, to use for selection
   */
  @Input('clrDgItem') item: T;

  constructor(
    public selection: Selection<T>,
    public rowActionService: RowActionService,
    public globalExpandable: ExpandableRowsCount,
    public expand: Expand,
    public hideableColumnService: HideableColumnService
  ) {
    this.id = 'clr-dg-row' + nbRow++;
    this.radioId = 'clr-dg-row-rd' + nbRow++;
  }
```

## How to reproduce

Run the Clarity Datagrid ClrDgSingleSelected demo ...
The id values for the generated radio buttons can be seen with an "Inspect Element" tool 

## Expected behavior

In a perfect world, the row # (nbRow) would match the "index" of the row 
For example, if I were to use the following ngFor loop to generate rows ... 
```
    <clr-dg-row *clrDgItems="let server of servers; let i = index" [clrDgItem]="server">
        <clr-dg-cell id="row{{i}}_name">{{server.name}}</clr-dg-cell>
```
then I could easily match a row found by server name with the radio button on the same row using the given "index" value

## Versions

**App**

* Angular CLI: 6.1.3 
* Node: 8.11.2
* Clarity: 0.12.9

**Device:**

* Type: HP Z230
* OS: Windows 7 64bit
* Browser: Chrome - Version: 69.0.3497.92 (Official Build) (64-bit)
* Browser: Firefox - Version: 63.0b4 (64-bit)

## Additional notes

One more thing just noticed - the nbRow value is not reset if you navigate to another page and then back to the one showing the datagrid - therefore, what used to be 'clr-dg-row-rd3' becomes 'clr-dg-row-rd43' or some other random number depending on the number of rows shown the first time the datagrid is rendered.

gnomeontherun:
fixes #2644

We are double bumping the index for each row, so this does it only once.

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-row.ts
+++ b/src/clr-angular/data/datagrid/datagrid-row.ts
@@ -114,8 +114,9 @@ export class ClrDatagridRow<T = any> implements AfterContentInit {
     public hideableColumnService: HideableColumnService,
     public commonStrings: ClrCommonStrings
   ) {
-    this.id = 'clr-dg-row' + nbRow++;
-    this.radioId = 'clr-dg-row-rd' + nbRow++;
+    nbRow++;
+    this.id = 'clr-dg-row' + nbRow;
+    this.radioId = 'clr-dg-row-rd' + nbRow;
   }
 
   private _selected = false;


bret-bailey:
Did you address the "Additional Notes" issue from https://github.com/vmware/clarity/issues/2644 as well?

"the nbRow value is not reset if you navigate to another page and then back to the one showing the datagrid - therefore, what used to be 'clr-dg-row-rd3' becomes 'clr-dg-row-rd43' or some other random number depending on the number of rows shown the first time the datagrid is rendered."

gnomeontherun:
@bret-bailey That is not something we can fix, because when the route is destroyed that id gets incremented the next time it's generated. We don't want to ever reuse an ID since we could get duplicates on the page which is not valid. If you need to retain the same ID for test purposes, that is somewhat brittle. My suggestion is to either not destroy the datagrid, or use a selector that targets the row without the id.

bret-bailey:
@gnomeontherun - Understood, but ...

I wasn't retaining the ID for reuse; I was just trying to "find" the radio button based on the row # ... which doesn't appear to be possible with this datagrid implementation - at least for subsequent visits to the datagrid

Since the id attribute is set, the CSS selector essentially is that same inconsistent value ... even my initial plan of counting the rows and # of visits to the page doesn't work to determine the value used

I don't understand your final suggestion of not destroying the datagrid - if you navigate away from the page that it exists on ... how would you control whether it was destroyed or not?  could you elaborate on that?

Thanks

bret-bailey:
fair enough

PR #2656, opened 2018-09-14
[NG] fix form controls grid override retaining col classes

### LINKED ISSUE #2647: Horizontal form layout with longer labels are wrapped in multi lines
## Describe the bug

Horizontal form layout with longer labels are wrapped in multi lines, and in smaller view the inputs will be placed in the next line, however there is enough space to place in the same line.

## How to reproduce

https://stackblitz.com/edit/clr-form-layout-horizontal-issue?file=src/app/app.component.html

1.  Go to the link
2.  The form is only horizontal when you expand the view, if you make it smaller then itt won't be horizontal, however it could be (inconsistent usege of col-2 on labels)

## Expected behavior

Use more space, stay horizontal if there is enough space for the label and input.
Remove col-sm-2 constraints.

## Versions

App

Angular CLI: 6.1.3
Node: 8.11.2
Clarity: 0.13
Device:

Type: Dell latitude E5550
OS: Windows 10 64bit
Browser: Chrome - Version: 69.0.3497.92 (Official Build) (64-bit)

gnomeontherun:
If you wanted to change the default widths on the label and controls in a grid, it would apply the correct classes to the wrapper but not remove them from the control itself.

fixes #2647


speti43:
Hi Guys,

@gnomeontherun thanks for the fix! 
What do you think, when will this be released, because our app with the clarity form upgrade cannot be released, due to this bug.
All the other reviewers are busy right now?

Thanks,
Peter

src/clr-angular/forms/common/providers/control-class.service.ts
```diff
@@ -20,4 +20,15 @@ export class ControlClassService {
     }
     return controlClasses.join(' ');
   }
+
+  // We want to remove the column classes from the input up to the container
+  initControlClass(renderer, element: HTMLElement) {
+    this.className = element.className;
+    const klasses = element.className.split(' ');
```
youdz: This breaks when rendering server-side. Immediate fix would be to guard it, but that means you'll get the wrong classes when rendering on the server side.

This is all very static, can't you just have an `@Input() @HostBinding() class` on your control components? That should let you do the same thing without accessing native element properties.

gnomeontherun: I've been trying to prove you right about SSR, but everything I've tried builds just fine.

Using `@Input() @HostBinding() class;` clobbers the `host: { '[class.clr-input']: 'true'}` declaration, so I'm using the element directly. However, to assuage concerns about SSR I'll add a check so it doesn't throw undefined errors.

PR #2669, opened 2018-09-20
[Web] Updating login page

Jeff-Kumar:
Updating login page with new clarity components.


mathisscott:
@gnomeontherun 
Is this something we want to merge?

PR #2673, opened 2018-09-21
Fix for "Multiple App level Alerts have incorrect rendering on low resolution screen" Adding left margin to alert-actions for smaller screens

GeekAb:
Signed-off-by: Abhishek Srivastava <nit.abhi85@gmail.com>


src/clr-angular/emphasis/alert/_alert.clarity.scss
```diff
@@ -355,7 +355,7 @@
 
     .alert-app-level {
       .alert-actions {
-        margin-left: 0;
+        margin-left: 45px;
```
hippee-lee: @mathisscott - should this be the equivalent REM unit? e.g `1.875rem` for ~45px?

mathisscott: Ah, yes it should. Good catch.

mathisscott: Please update this value to rems...

GeekAb: Done. But for some reason, I am getting dco-required event when I am adding "Name and e-mail id" as the last line in the commit. 

Am I missing something?

hippee-lee:
As far as signing the commit ... Once we have the correct value for `margin-left`, as @mathisscott said, all you have to do is squash your commits into the one that is signed.

PR #2676, opened 2018-09-22
Adding left margin to alert-actions for smaller screens as rem

GeekAb:
Signed-off-by: Abhishek Srivastava <nit.abhi85@gmail.com>


GeekAb:
@hippee-lee @mathisscott fixed the issue. Created this new pull request. 

@hippee-lee you were right, value should be 1.875rem I was looking at wrong base value.

PR #2687, opened 2018-09-26
Dev app: import only feature modules in lazy-loaded modules

youdz:
The correct way for real-life application to import Clarity in production is to only import the feature modules they need, in order to reduce the size of their bundle. Components declared in an NgModule cannot be treeshaken.

Our dev app did not follow this pattern, which means it wasn't representative of actual applications using Clarity. This change fixes that to have an accurate sandbox when developing.
Also note that the kitchen sink app still imports `ClarityModule`, so the main module is still tested as part of our suite.

This allowed me to catch a few of our modules that didn't correctly re-export their obvious dependencies, which I fixed as part of this change: `ClrLoadingButtonModule` exports `ClrLoadingModule`, `ClrSignpostModule` exports `ClrIconModule` for the recommended icon trigger, etc.


gnomeontherun:
Components are tree-shaken from a project if they are not consumed, so this is no longer necessary (at least if you have the right setup with the CLI, custom builds might be failing to do this properly). The exception appears to be ClrFormsNextModule since its not part of the main ClarityModule, its being included even if you don't use it (strange, not certain why yet).

I'm not opposed to this change since it makes sense for testing cases and we can keep it this way to catch reexport issues. I wonder if there is a better way to catch those automatically though.

gnomeontherun:
To update the comment, we believe the _modules_ are dropped which means if you don't use a component inside of a module that whole module is dropped from the final build. That is a decent result for most cases, but yes for testing we should recommend importing just the module that contains the pieces you need in your test rather than all of Clarity (speed and isolation).

mathisscott:
@youdz 
Are we going to merge this one day?

gnomeontherun:
I think we can drop this PR, it isn't necessary since Angular drops the unused modules and the rebase is probably a pain.

youdz:
The problem wasn't optimization for performance, the problem was that we had no way of testing our individual modules exported all the pieces they needed. We've had broken modules shipped because they forgot to export some common/util directive that consumers couldn't even import themselves because it was private.

But yeah, I dropped the ball hard on this one, we can just close this PR.

PR #2693, opened 2018-09-27
 [NG] improve radio inline with error formatting

gnomeontherun:
Inline radios with errors were also inlining the errors instead of breaking to a new line based on the design.


src/dev/src/app/radios/radios.demo.html
```diff
@@ -33,32 +38,42 @@ <h4>Vertical</h4>
...
     </div>
   </div>
 
   <div class="clr-form-control">
     <label class="clr-control-label">
       Inline radio</label>
-    <div class="clr-control-container">
-      <div class="clr-radio-wrapper clr-radio-inline">
+    <div class="clr-control-container clr-control-inline">
+      <div class="clr-radio-wrapper">
```
mathisscott: So we now require a `.clr-radio-wrapper` around each radio and input?

Is this a breaking change?

gnomeontherun: This is actually what is in the docs, the demo just didn't have it right.

src/dev/src/app/radios/radios.demo.html
```diff
@@ -15,14 +15,19 @@ <h4>Vertical</h4>
...
         <label for="vertical-radio2" class="clr-control-label">option 2</label>
+      </div>
+      <div class="clr-radio-wrapper">
         <input type="radio" id="vertical-radio3" name="radio-basic" class="clr-radio">
         <label for="vertical-radio3" class="clr-control-label">option 3</label>
       </div>
-      <!-- IMPORTANT DIFFERENCE IN STRUCTURE! ICON IS NOT PART OF THE INPUT WRAPPER -->
-      <clr-icon class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
-      <span class="clr-subtext">Helper Text</span>
+      <div class="clr-subtext-wrapper">
```
gnomeontherun: @mathisscott This is actually what is different, but not documented yet. It works without this wrapper unless you have inline, so it should probably just be the default option regardless.

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -48,7 +47,7 @@ export class ClrRadioContainer implements OnDestroy {
    */
   @Input()
   set clrInline(value: boolean | string) {
-    if (!isBoolean(value)) {
+    if (typeof value === 'string') {
```
Shijir: So it looks like it's here to check whether the value is true or false regardless of its boolean or string. But if the string value is just a random string like "cat", `this.inline` would become true. Is that expected a behavior here?

src/clr-angular/forms/radio/radio-container.ts
```diff
@@ -19,13 +18,13 @@ import { NgControlService } from '../common/providers/ng-control.service';
   template: `
     <ng-content select="label"></ng-content>
     <label *ngIf="!label && addGrid()"></label>
-    <div class="clr-control-container" [ngClass]="controlClass()">
-      <div [class.clr-radio-inline]="clrInline">
-        <ng-content select="clr-radio-wrapper"></ng-content>
+    <div class="clr-control-container" [class.clr-control-inline]="clrInline" [ngClass]="controlClass()">
```
youdz: Don't `[class.xxx]` and `[ngClass]` conflict on the same element? I remember having this discussion, I think the conclusion was that even if `ngClass` doesn't overwrite existing classes today, it's pretty unstable and we shouldn't rely on it.

src/clr-angular/forms/radio/radio-container.ts
youdz: It's the expected behavior, yes. Normally `"false"` evaluates to `true` too, we just want to make that one exception to simplify the API a tiny bit if people don't use a binding here.

src/clr-angular/forms/radio/radio-container.ts
Shijir: Since we moved `[class.clr-control-inline]="clrInline"`, don't we need to add tests for this?

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: `clrInline=""` is basically `true` state. This is about handling how HTML attributes work and not JavaScript, plus handling an edge case. 

In HTML the presence of an attribute determines it's boolean state, but with Angular binding we want to handle the case where `[clrInline]="false"` or clrInline="false" to make it more fault tolerant for typos.

src/clr-angular/forms/radio/radio-container.ts
gnomeontherun: They do not conflict in this case, though I didn't want to modify the controlClass service. I don't remember having any discussions about this, but the alternative is to move the logic up to the `controlClass()` logic.

gnomeontherun: Added a test

gnomeontherun: Yes except I'll need this logic in checkboxes too so I've already consolidated it in service.

gnomeontherun:
Sorry about the explosion but I realized there were a few places in our dev app that needed the updated styles as well. Plus I expanded the unit tests on the form controls to include a version using the grid.

src/clr-angular/forms/radio/radio-container.spec.ts
```diff
@@ -73,5 +83,28 @@ export default function(): void {
...
+          providers: [NgControl, NgControlService, IfErrorService, LayoutService],
+        });
+        fixture = TestBed.createComponent(TemplateDrivenTest);
+
+        containerDE = fixture.debugElement.query(By.directive(ClrRadioContainer));
+        containerEl = containerDE.nativeElement;
+        fixture.detectChanges();
+      });
+
+      it('sets the inline class', () => {
```
Shijir: Awesome 👌

PR #2695, opened 2018-09-28
[NG] fix(datagrid-renderer):

### LINKED ISSUE #2692: Datagrids with clrDgSingleSelected and width constraints do not resize their columns correctly
## Describe the bug

When a datagrid uses the Single Select option (clrDgSingleSelected) and also has data which needs to have word-breaks applied to fit in the container size, the datagrid will be automatically sized to include a horizontal scrollbar. This happens even if the version of the datagrid without clrDgSingleSelected works correctly at a similar size.

## How to reproduce
[Stackblitz Example](https://stackblitz.com/edit/clarity-light-theme-v013-vj33sm)

1.  Create a datagrid within a div with a width constraint (say 500px)
2.  Add 5 columns with headers Column 1, 2, 3, 4, 5
3.  Add a single row with data which will require a line break. This can be data which requires a mid-word break, or data that simply results in long lines.
4.  Do not add any kind of Single Select options (i.e. do not include the clrDgSingleSelected attribute)
5. Load the page and observe that the datagrid is rendered correctly, and no horizontal scrollbar is introduced
6. Add the clrDgSingleSelected attribute to the datagrid
7. Increase the size of the div by 50px (this should be enough to accommodate the new column for the radio button, which as of 0.13.2 is 38px)
8. Reload the page and observe that even though the data columns in the grid have the same amount of space as before, a horizontal scrollbar is now shown.

## Expected behavior

If a datagrid can be displayed without a horizontal scrollbar when the clrDgSingleSelected option is not used, then that same datagrid should be displayed without a horizontal scrollbar when clrDgSingleSelected is used and it is given additional width to accommodate the radio buttons.

## Versions

**App**

* Angular: 6
* Node: 10.5.0
* Clarity: 0.13.2

**Device:**

* Type: MacBook
* OS: macOS High Sierra
* Browser: Chrome
* Version: 69

hippee-lee:
- Added  fix that properly calculates row width when row actions are present
- Added tests for each possible row action that could be declared
- fixes #2692

### DIFF
--- a/src/clr-angular/data/datagrid/render/main-renderer.spec.ts
+++ b/src/clr-angular/data/datagrid/render/main-renderer.spec.ts
@@ -3,10 +3,9 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component } from '@angular/core';
+import { Component, ElementRef, ViewChild } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations'; // Needed to recreate issue #1084
-
 import { ClrDatagridModule } from '../datagrid.module';
 import { TestContext } from '../helpers.spec';
 import { Page } from '../providers/page';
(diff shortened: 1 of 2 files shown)


src/clr-angular/data/datagrid/render/main-renderer.spec.ts
```diff
@@ -107,6 +131,62 @@ export default function(): void {
...
+          providers: PROVIDERS,
+        });
+        context = TestBed.createComponent(RenderWidthTest);
+        context.detectChanges();
+      });
+
+      it('calculates the correct width with no row options', () => {
+        context.componentInstance.currentTest = 'defaultDatagridTest';
+        context.detectChanges();
+        const containerWidth = context.componentInstance.container.nativeElement.clientWidth;
```
gnomeontherun: nit: Move this line to before each

hippee-lee: Done.

PR #2698, opened 2018-10-01
[NG] support disabled states in form controls

### LINKED ISSUE #2665: New Radio button control does not have disabled design
## Describe the bug

New Radio button control does not have disabled design

## How to reproduce

https://stackblitz.com/edit/clr-radio-missing-disabled-design

1.  Go to Stackblitz
2. The radiobutton itself is disabled, if I add the disabled attribute, but the css and the UI does not reflect the state, as you see I've tried every possibility, the documentation does not describe, how to do it:
https://vmware.github.io/clarity/documentation/v0.13/radio

The old version can be seen at the bottom, we need some way to pass the control container that this is disabled, and apply a css class.

## Expected behavior

Let the user see, this control is disabled

## Versions
App

Angular CLI: 6.1.3
Node: 8.11.2
Clarity: 0.13
Device:

Type: Dell latitude E5550
OS: Windows 10 64bit
Browser: Chrome - Version: 69.0.3497.92 (Official Build) (64-bit)

gnomeontherun:
fixes #2665

This adds support to check for disabled states on controls and lift that state up to the container.


gnomeontherun:
I will update this to apply the same treatment to datepicker (just merged).

PR #2708, opened 2018-10-06
[WEB] Unset max-height on pre of code-snippet

imcotton:
It's too often that default `pre { max-height: 15rem }` gets in the way while reading sample code on the doc site, given that other sites (e.g. MDN / Bootstrap / Angular.io) would just show code snippet in full-size, suggest to take out the squeezing to make it more readable.

### DIFF
--- a/latest/src/app/utils/code-snippet.ts
+++ b/latest/src/app/utils/code-snippet.ts
@@ -20,6 +20,7 @@ import {CodeHighlight} from "./code-highlight";
         pre {
             background: transparent;
             padding: 12px;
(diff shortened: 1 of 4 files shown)


imcotton:
After reading issue **1138**, it's more likely a UX adjustment on CSS level of `<pre>` tag, to that I'm not the expert and will wait on its development indeed.

The changing I'm proposing here only alter to `clr-code-snippet` component which to being used on the doc site only, so it has no impact on Clarity as a whole, but brings readability to the doc site immediately I think.

PR #2711, opened 2018-10-08
[CORE] add dark theme testing

gnomeontherun:
This sets up a way to build dev app as dark theme, and captures the initial gemini testing for dark theme.


src/dev/src/app/app.component.ts
```diff
@@ -3,11 +3,26 @@
...
+  templateUrl: './app.component.html',
+  encapsulation: ViewEncapsulation.None,
+})
 export class AppComponent {
   public routes: Route[] = APP_ROUTES;
+
+  constructor() {
+    if (environment.dark) {
+      // tslint:disable-next-line
+      require('style-loader!./../../../clr-angular/dark-theme.scss');
```
youdz: I'm just wondering if there is any particular reason this is in the app component with no encapsulation. It feels more like `main.ts`'s job to me. Probably a nitpick.

PR #2714, opened 2018-10-09
[UI] `clr-` prefix for css classes

jeeyun:
Using postcss to generate two versions of css files:
- one where every classname is prefixed with `clr-`
- one where old classname and prefixed version coexist to allow the client app to migrate at their own pace

Note that the script order in `build:ui` matters, since one of them change the css files in place as opposed to generating a new file. 


package-lock.json
```diff
@@ -13848,6 +13844,69 @@
         }
       }
     },
+    "postcss-prefix-selector": {
```
gnomeontherun: The package-lock changes are more extensive than just this new package. Can you revert the package-lock.json changes, except for this new piece?

jeeyun: I'll take a look and revert.

gnomeontherun:
Just to follow up, we discussed keeping this open until Matt has the other PRs ready so we can see them all together first, but otherwise this is good to me.

mathisscott:
this solution is problematic. we are pushing past 1.0 to find a better way to do this.

PR #2726, opened 2018-10-12
[NG] Switch our global providers to the tree-shakeable syntax

youdz:
The main reason for this change is not really for them to be
tree-shakeable (which doesn't change much for us, to be honest),
but rather to ensure we only get a single instance even with lazy-loaded feature modules.


src/clr-angular/layout/nav/providers/responsive-navigation.service.ts
```diff
@@ -10,7 +10,7 @@ import { Subject } from 'rxjs';
 import { ResponsiveNavCodes } from '../responsive-nav-codes';
 import { ResponsiveNavControlMessage } from '../responsive-nav-control-message';
 
-@Injectable()
+@Injectable({ providedIn: 'root' })
```
hippee-lee: in this context, `root` is the application consuming Clarity, right?

youdz: Exactly. It declares the service in the root injector for the application.

PR #2729, opened 2018-10-12
Angular 7 support

gnomeontherun:
This updates our package for Angular 7 rc.1.

* Build uses latest versions of Angular, ng-packagr, Typescript
* Let Angular run migration scripts (only removed a line from polyfills and bumped versions)
* Updated any dependencies that had reported dependency vulnerabilities (except gemini-coverage still has a few)
* Removed the `package.json` file which is generated during build and shouldn't be tracked
* Removed prismjs files from dev app and unit tests
* Changed the modal max height gemini test so it doesn't auto open, which was causing some issues with gemini/docker/selenium/something
* Fixed a few linting issues that came up after the dependency changes
* Changed gemini to be part of the repo instead of expecting it to be a globally installed node module


.gitignore
```diff
@@ -8,6 +8,7 @@
 src/clr-angular/.ng_build
 src/clr-angular/.ng_pkg_build
 /tests/tslint/dist
+src/clr-angular/package.json
```
hippee-lee: This file is getting deleted. Is there a reason this needs to be here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "clr-all",
-  "version": "0.13.4",
+  "name": "@clr/clarity",
+  "version": "1.0.0-alpha.0",
```
hippee-lee: 🙌🏼  🎉  🎊 💥

src/ks-app/src/polyfills.ts
```diff
@@ -29,7 +29,6 @@ import 'core-js/es6/set';
 
 /* Evergreen browsers require these. */
 import 'core-js/es6/reflect';
-import 'core-js/es7/reflect';
```
hippee-lee: Just a question about this, I don't think its a blocker. 
Why is this being removed? I don't see why es6 would be kept instead of es7. Doesn't the es7 polyfill have everything in es6 plus new things?

gnomeontherun: The Angular update tool removed this automatically, because this is only used in JIT. I believe the devkit now includes it as necessary instead of it being at the app level.

gnomeontherun: See https://github.com/angular/angular-cli/commit/30f8352ca2b3cccd72f5fcff8e09a543f9c70a8d

.gitignore
gnomeontherun: It was removed because its automatically generated during the build. Sometimes it was getting committed after being updated during releases, sometimes not. Point is, we don't need it except for build time and its generated. ng-packagr needs to have a version of the file for its build process.

src/clr-angular/button/button-loading/loading-button.spec.ts
```diff
@@ -33,15 +33,15 @@ describe('Loading Buttons', () => {
   });
 
   it('displays spinner when [(clrButtonState)] value is LOADING', () => {
-    fixture.componentInstance.buttonState = ClrLoadingState.LOADING;
+    fixture.componentInstance.buttonState = <ClrLoadingState>ClrLoadingState.LOADING;
```
youdz: This is probably only needed because our unit tests typing is a bit shaky. But it really doesn't matter for this upgrade.

gnomeontherun: It was due to TS version bump. It was expecting the `buttonState` to be typed as `ClrLoadingState.DEFAULT`.

.travis.yml
```diff
@@ -36,8 +36,6 @@ cache:
   apt: true
   directories:
   - node_modules
-before_script:
-- npm install -g gemini@5.2.0 html-reporter@2.0.1
```
jeeyun: Did we change something in travis builds? How come this is not needed anymore??

gnomeontherun: I moved `gemini` and `html-reporter` to the `devDependencies` in package.json, so it is one less thing to do when setting up Clarity. The scripts were updated to run against the version installed in node_modules instead of globally, which we should really avoid using global packages anyways.

jakehockey10:
Is this something I can use right now?  If so, how do I go about doing that?  I don't see any alpha releases for 1.0.0 on npm so I wasn't sure if I could upgrade to Angular 7 today or not

gnomeontherun:
Clarity 0.12-0.13 can work with Angular 7, but you'll get NPM warnings if you try to update (which you can ignore for the time being). When Clarity 1.0 is out, you can update it as well separately.

whizkidwwe1217:
What do I need to change so my custom themes will still work? My approach to theming (where the user can switch themes on runtime) is similar to https://github.com/d-m-s/clarity-addons. I tried out clarity 1.0 beta, seems fine but it broke my data grid. Data grid cells and headers are broken although the functionality still works.

gnomeontherun:
@whizkidwwe1217 why don't you post on stack overflow with the `vmware-clarity` tag with more details and a reproduction, as I don't know what your custom runtime theme process is.

PR #2730, opened 2018-10-13
[NG] Replace old forms module

hippee-lee:
This change removes and renames the old forms module.

- The `ClrFormsModule` becomes `ClrFormsDeprecatedModule`
- The `ClrFormsNextModule`becomes `ClrFormsModule`
- `ClrFormsDeprecatedModule` is no longer published in `ClarityModule` and will need to be explicetly imported if an app still has old forms. (TODO: Document this on the website)


hippee-lee:
I still need to update the documentation for this issue when the website is updated for the 1.0 releases.

PR #2735, opened 2018-10-16
[NG] support reset and forced validation

### LINKED ISSUE #2678: [NG] Support form reset for Clarity's form controls
## Describe the bug

Currently, our form controls all ignore status changes due to a full or partial form reset. This is due to us filtering the control status changes to ignore them until the input has been touched. See https://github.com/vmware/clarity/blob/master/src/clr-angular/forms/common/if-error/if-error.service.ts#L39:
```
this.control.statusChanges.pipe(filter(() => this.control.touched)).subscribe(() => {
  this._statusChanges.next(this.control);
})
```
It makes sense for a new form, but this prevents us from properly resetting a used form.

## How to reproduce

[This StackBlitz](https://stackblitz.com/edit/clarity-light-theme-v013-6s2qtq) someone posted [on StackOverflow](https://stackoverflow.com/questions/52459708/how-to-reset-error-state-of-clarity-forms) show the problem.

1. Click in the input
2. Click outside the input
3. Click the "Reset Form" button
The input error is still displayed, even though the input is back to pristine and untouched.

## Expected behavior

`FormGroup.reset()` and other ways of resetting a form should properly reset our controls to their initial state (in particular, no error displayed).

## Versions

**App**

* Angular: 6+
* Clarity: 0.13+

gnomeontherun:
The forms now handle form resets (which will clear validation errors) by calling `reset()` on a form group or control.

The new API for forcing validation is to get ahold of a `ClrForm` instance and call `markAsDirty()` which will mark every control as dirty in the form. I chose this naming to be very explicit about what happens (instead of saying 'validate' which is a side effect).

This also introduces some internal refactorings for the controls to remove the duplicate logic from individual controls and put it into the WrappedFormControl. Most of the changes are from carrying this change through all the controls. It also refactors the private ifErrorService to return a boolean instead of the control itself so that logic is contained in one place.

closes #2678

### DIFF
--- a/src/clr-angular/forms/checkbox/checkbox.ts
+++ b/src/clr-angular/forms/checkbox/checkbox.ts
@@ -4,43 +4,23 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Directive, Renderer2, ElementRef, HostListener, OnInit, Optional, ViewContainerRef } from '@angular/core';
+import { Directive, Renderer2, ElementRef, Injector, Self, Optional, ViewContainerRef } from '@angular/core';
 import { NgControl } from '@angular/forms';
 import { ClrCheckboxWrapper } from './checkbox-wrapper';
 
-import { IfErrorService } from '../common/if-error/if-error.service';
-import { ControlClassService } from '../common/providers/control-class.service';
-import { NgControlService } from '../common/providers/ng-control.service';
 import { WrappedFormControl } from '../common/wrapped-control';
 
 @Directive({ selector: '[clrCheckbox]' })
-export class ClrCheckbox extends WrappedFormControl<ClrCheckboxWrapper> implements OnInit {
+export class ClrCheckbox extends WrappedFormControl<ClrCheckboxWrapper> {
   constructor(
     vcr: ViewContainerRef,
-    @Optional() private ngControlService: NgControlService,
-    @Optional() private ifErrorService: IfErrorService,
-    @Optional() private control: NgControl,
-    @Optional() controlClassService: ControlClassService,
-    el: ElementRef,
-    renderer: Renderer2
+    injector: Injector,
+    @Self()
+    @Optional()
+    control: NgControl,
+    renderer: Renderer2,
+    el: ElementRef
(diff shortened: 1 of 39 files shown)


src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -4,17 +4,30 @@
...
-    super(ClrCheckboxContainer, vcr);
+  protected wrapperType = ClrCheckboxContainer;
+
+  constructor(
+    protected vcr: ViewContainerRef,
+    @Optional() ngControlService: NgControlService,
+    @Optional() ifErrorService: IfErrorService,
+    @Optional() control: NgControl,
+    @Optional() controlClassService: ControlClassService,
+    @Optional() markControlService: MarkControlService,
```
youdz: Nitpick: when we inherit and have that many injections, I prefer by far just injecting `Injector`, passing that to the parent constructor, and letting it query all the services it needs.

The reason we don't do this for some instances is that you can't query for `ViewContainerRef`, `ElementRef` or similar DOM-related references through the injector, they have some kind of special status.

mathisscott:
@gnomeontherun + @youdz 
When do we want this to go in?

gnomeontherun:
We should get the new checkboxes in before this, since there is going to be a conflict.

src/clr-angular/forms/common/if-error/if-error.service.ts
```diff
@@ -7,16 +7,15 @@
...
 import { NgControl } from '@angular/forms';
 import { Observable, Subject, Subscription } from 'rxjs';
-import { filter } from 'rxjs/operators';
 
 import { NgControlService } from '../providers/ng-control.service';
 
 @Injectable()
 export class IfErrorService implements OnDestroy {
   // Implement our own status changes observable, since Angular controls don't
   // fire on events like blur, and we want to return the control instead of a string
```
hippee-lee: Update the comment here?

gnomeontherun: done

PR #2738, opened 2018-10-16
Add two-way binding to ClrDatagridHideableColumn

Wallabeng:
Signed-off-by: Wallabeng <marcel@ring-mail.at>

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
@@ -56,6 +56,34 @@ export default function(): void {
         expect(context.clarityDirective.columnId).toEqual(context.clarityDirective.hideable.id);
       });
     });
+
+    describe('TypeScript Output API', function() {
+      let context: TestContext<ClrDatagridColumn<void>, HideableOutputTest>;
+
+      beforeEach(function() {
+        context = this.create(ClrDatagridColumn, HideableTest, PROVIDERS_NEEDED);
+      });
+
+      it('creates a DatagridHideableColumn instance on the DatagridColumn', function() {
+        expect(context.clarityDirective.hideable).toBeDefined();
+      });
+
+      it('defaults the HideableColumn.hidden property to false', function() {
+        expect(context.clarityDirective.hideable.hidden).toBe(false);
+      });
+
+      it('correctly populates the DatagridHideableColumn instance with an id', function() {
+        expect(context.clarityDirective.columnId).toEqual(context.clarityDirective.hideable.id);
+      });
+
+      it('emits column state change', function() {
+        let hiddenState = false;
+        context.testComponent.directive.hiddenChange.subscribe(state => (hiddenState = state));
+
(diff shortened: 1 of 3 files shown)


src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
```diff
@@ -56,6 +56,34 @@ export default function(): void {
...
+        expect(context.clarityDirective.hideable.hidden).toBe(false);
+      });
+
+      it('correctly populates the DatagridHideableColumn instance with an id', function() {
+        expect(context.clarityDirective.columnId).toEqual(context.clarityDirective.hideable.id);
+      });
+
+      it('emits column state change', function() {
+        let hiddenState = false;
+        context.testComponent.directive.hiddenChange.subscribe(state => (hiddenState = state));
```
youdz: Since you're using the two-way binding syntax in your test component, testing that the output emits correctly is simply testing that your test component's value is up to date. This means you don't need a subscription:
```typescript
it('emits column state change', function() {
  context.clarityDirective.hideable.hidden = true;
  expect(context.testComponent.hidden).toBeTrue();
});
```

src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
```diff
@@ -56,6 +56,34 @@ export default function(): void {
...
+
+      it('defaults the HideableColumn.hidden property to false', function() {
+        expect(context.clarityDirective.hideable.hidden).toBe(false);
+      });
+
+      it('correctly populates the DatagridHideableColumn instance with an id', function() {
+        expect(context.clarityDirective.columnId).toEqual(context.clarityDirective.hideable.id);
+      });
+
+      it('emits column state change', function() {
```
youdz: I feel bad because I asked you to add an input, and now I'm pointing out that you're missing a unit test for that input. 😂 
Well, here is the test, that way I don't feel so bad 😛:
```typescript
it('emits column state change', function() {
  context.testComponent.hidden = true;
  context.detectChanges();
  expect(context.clarityDirective.hideable.hidden).toBeTrue();
});
```

src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
Wallabeng: you're completely right, could've though of this on my own. thanks! fixed.

src/clr-angular/data/datagrid/datagrid-hideable-column.spec.ts
Wallabeng: No problem at all, thanks for providing it 😄 It's included now

Wallabeng:
Golden file needs some love too I guess, but there I need some instructions how to handle it. Just copy the compiled file over?

Wallabeng:
Everything updated and green now! Is it possible to add this to todays release?

Shijir:
Hi @Wallabeng, v0.13.6 and v0.12.6 are already out today. But we can include your fix in our next week's release 🙂

PR #2750, opened 2018-10-18
[ICONS] remove deprecated shapes for 1.0

gnomeontherun:
This removes the `eye` and `eye-hide` alias since they were moved to the core.


jeeyun:
We should also remove them from `essential-shapes.ts`.

PR #2751, opened 2018-10-18
[NG] remove support for deprecated `alert-X` variations of alert types

gnomeontherun:
This removes support for `alert-info` types and updates to use `clr-alert-item` for proper icon support.

### DIFF
--- a/src/clr-angular/emphasis/alert/alert.spec.ts
+++ b/src/clr-angular/emphasis/alert/alert.spec.ts
@@ -79,25 +79,25 @@ export default function(): void {
     });
 
     it('extends the alert type classes when clrAlertType is set', () => {
-      // default alert-info class
+      // default info class
       expect(compiled.querySelector('.alert-info')).not.toBeNull();
 
-      // set alert-danger
-      fixture.componentInstance.type = 'alert-danger';
+      // set danger
+      fixture.componentInstance.type = 'danger';
       fixture.detectChanges();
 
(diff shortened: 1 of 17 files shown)


src/clr-angular/emphasis/alert/alert.spec.ts
```diff
@@ -79,25 +79,25 @@ export default function(): void {
...
-      fixture.componentInstance.type = 'alert-warning';
+      // set warning
+      fixture.componentInstance.type = 'warning';
       fixture.detectChanges();
 
       expect(compiled.querySelector('.alert-danger')).toBeNull();
       expect(compiled.querySelector('.alert-warning')).not.toBeNull();
 
-      // set alert-success
-      fixture.componentInstance.type = 'alert-success';
```
youdz: Straight from our docs:
screen shot 2018-10-18 at 12 38 35

It uses every single deprecated syntax. I'm okay still cutting all this for 1.0, but we need to at least update the docs before we can actually cut them.

gnomeontherun: Yes we need to fix docs, but thats an oversight when it was deprecated to not update the docs accordingly.

mathisscott: Not sure it was an oversight or not. We had no timeline for the deprecation and back then had no strategy for handling deprecations like this.

PR #2752, opened 2018-10-18
[a11y] Updating alerts dox

mathisscott:
• Added examples of using ARIA role "alert" to the static alert examples
• Removed instances of using div.alert-item in Angular examples to using the preferred cir-alert-item
• Did not add ARIA roles to examples on page in order to prevent screenreaders from freaking out on the number of example alerts

Tested in:
✔︎ Chrome


mathisscott:
I verified these changes locally. I only updated `clr-alert-item` in latest dox.

PR #2754, opened 2018-10-18
[NG] Pagination 2.0

jeeyun:
I still need to edit this slightly after the icon for the first/past page has been released. Screenshot for what it's supposed to look like, based on implementing with the unreleased svg:

light:
screen shot 2018-10-25 at 4 30 56 pm

dark:
screen shot 2018-10-25 at 4 31 22 pm


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -862,6 +907,22 @@
     justify-content: center;
     align-items: stretch;
 
+
+
+    input[type='text'] {
+      background: none !important;
+      background-color: #ffffff;
```
mathisscott: how does this translate in the dark theme?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -871,29 +932,37 @@
...
       height: $clr-datagrid-icon-size;
       background-repeat: no-repeat;
       background-size: contain;
     }
 
+    .pagination-first:disabled,
+    .pagination-last:disabled,
+    .pagination-previous:disabled,
+    .pagination-next:disabled {
+      clr-icon {
```
mathisscott: generally, we try not to target `clr-icon` elements directly. Also we have a number of hardcoded color values in here we should clean up

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -871,29 +932,37 @@
...
+        fill: #737373;
+        opacity: 0.54;
```
mathisscott: generally speaking it's preferable to use a color value than opacity. opacity can have unintended side effects if users change the colors here via theming.

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -3,58 +3,45 @@
...
+      <button class="pagination-previous" [disabled]="page.current === 1" (click)="page.current = page.current - 1">
+        <clr-icon shape="angle left"></clr-icon>
+      </button>
+      <input type="text" class="pagination-current" [size]="page.last.toString().length" [(ngModel)]="page.current" />&nbsp;/&nbsp;<span>{{page.last}}</span>
+      <button class="pagination-next" [disabled]="page.current === page.last" (click)="page.current = page.current + 1">
+        <clr-icon shape="angle right"></clr-icon>
+      </button>
+      <button class="pagination-last" [disabled]="page.current === page.last" (click)="page.current = page.last">
+        <clr-icon shape="angle-double right"></clr-icon>
+      </button>
```
mathisscott: I dig this. Nice.

src/clr-angular/data/datagrid/datagrid-page-size.ts
```diff
@@ -0,0 +1,28 @@
...
+    <ng-content></ng-content>
+    <div class="clr-select-wrapper">
+      <select [(ngModel)]="page.size">
+        <option *ngFor="let option of pageSizeOptions" [ngValue]="option">{{option}}</option>
+      </select>
+    </div>
+  `,
+})
+export class ClrDatagridPageSize {
+  private _pageSizeOptions: number[] = [];
```
mathisscott: Does it make sense to offer default pageSizeOptions so users don't _have_ to supply them if they want the easy way out?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -862,6 +907,22 @@
     justify-content: center;
     align-items: stretch;
 
+
+
+    input[type='text'] {
```
mathisscott: I'm not sure what it is. But the input field doesn't seem properly aligned. It looks too close to the `/` to me and the space between the left angle looks too big. Do you see that as well?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
jeeyun: Ah, good point! I will ask @lil-kim for dark theme colors and update.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
jeeyun: What is the general approach we recommend in this case? Use classnames on these icons?

src/clr-angular/data/datagrid/datagrid-page-size.ts
jeeyun: If they don't want to have page size options, they can opt to not use it in their markup. 

With:
```
    <clr-dg-pagination #pagination [clrDgPageSize]="20">
      <clr-dg-page-size [clrPageSizeOptions]="[10,20,50,100]">Users per page</clr-dg-page-size>
      {{pagination.totalItems}} users
    </clr-dg-pagination>
```
Without:
```
    <clr-dg-pagination #pagination [clrDgPageSize]="20">
      {{pagination.totalItems}} users
    </clr-dg-pagination>
```

I would assume that they would want to supply them if they are choosing to use the component. What do you think?

mathisscott: So if the array is empty the select box doesn't show up?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
jeeyun: After internal discussion, we are making `color` and `opacity` for disabled state both variables that can be overwritten via theming. I also moved out out of targeting `clr-icon` as per your suggestion.

src/clr-angular/data/datagrid/datagrid-page-size.ts
jeeyun: Yeah it shows up but is blank. If you think it's worthwhile, I can default it to include one option with the value that is passed by `[clrDgPageSize]`?

mathisscott: Why would we want it to show up blank? 

If you had the pagination select box when the array is empty, then I think it's fine the way it is. Basically that's a way for people to turn off that feature if it's not useful for them (and by default it's turned off).

But if we are going to show the select box no matter what, we should have a set of default options. Which is problematic if our defaults are larger than their dataset.

jeeyun: I'll set it so that it shows up with one option of value passed in the `clrDgPageSize`. I'm afraid that if we hide the select when array is empty, it'll just cause more confusion.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
jeeyun: Yes! I updated the css and this should be fixed with latest changes.

Shijir:
`greater-than-one` icon is on the way :D

src/clr-angular/data/datagrid/_variables.datagrid.scss
```diff
@@ -9,6 +9,9 @@ $clr-datagrid-fixed-column-size: (2 * $clr-table-cellpadding) + $clr-table-lineh
 $clr-datagrid-font-color: #adbbc4;
 $clr-datagrid-row-hover: $clr-light-gray;
 $clr-datagrid-action-toggle: $clr-dark-gray;
+$clr-datagrid-pagination-btn-color: #565656;
+$clr-datagrid-pagination-btn-disabled-color: #737373;
```
mathisscott: These colors should be in our `$clr-gray` family already. We should use those SASS variables for now.

src/clr-angular/utils/_theme.dark.clarity.scss
```diff
@@ -341,6 +341,9 @@ $clr-datagrid-row-hover: #29414e;
 $clr-datagrid-row-selected: $clr-white;
 $clr-datagrid-popover-bg-color: #22343c;
 $clr-datagrid-action-toggle: $clr-datagrid-font-color;
+$clr-datagrid-pagination-btn-color: #e9ecef;
```
mathisscott: I know we have some special colors in our dark theme but is `#e9ecef` one of them?

mathisscott: It looks like it. We should consider using some sort of SASS variable for it, though. It shows up in this file 9 times already.

src/clr-angular/data/datagrid/datagrid-page-size.ts
youdz: I agree with the behavior Jeeyun has on this one. Components with a "mandatory" input are not hard to grasp, the basic example in every tutorial is a "greeter" component that utterly fails if you don't give it a `name` input.

I like the gracious fallback to a select that only has a single option, the current page size. So 👍 on that.
To address Scott's concern, see my next comment.

src/clr-angular/data/datagrid/datagrid-page-size.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor(public page: Page) {}
+
+  ngOnInit() {
+    if (this.pageSizeOptions.length === 0) {
+      this.pageSizeOptions = [this.page.size];
```
youdz: @mathisscott: This is a nice fallback, a select with a single option. If you're still worried, how about we log a warning inside of this if:
`console.warn('You are using the <clr-dg-page-size> component but forgot to provide a list of acceptable page sizes.');`

src/clr-angular/data/datagrid/datagrid-page-size.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Input('clrPageSizeOptions')
+  public set pageSizeOptions(options: number[]) {
+    if (options) {
+      this._pageSizeOptions = options;
+    }
+  }
+
+  public get pageSizeOptions(): number[] {
```
youdz: Why the getter and setter here? Looks like
```typescript
@Input('clrPageSizeOptions') pageSizeOptions: number[]
```
would work just fine.

src/clr-angular/data/datagrid/datagrid-page-size.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        it('updates the page size upon pageSizeOption selection', function() {
+          pageProvider.size = 10;
+          pageProvider.totalItems = 100;
+          pageProvider.current = 1;
+          context.testComponent.pageSizeOptions = [10, 20, 50, 100];
+          context.detectChanges();
+
+          const select = context.clarityElement.querySelector('select');
+          expect(select).not.toBeNull();
+          select.value = '2: 20'; // setting it to the second option, with the expected value syntax when we use [ngValue]
```
youdz: That's a bit brittle, using the `NgControl` or something would maybe work better, but don't worry about it. Just a general note, this is fine right now.

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -3,58 +3,45 @@
...
-            </li>
-        </ul>
+    <div class="pagination-size" *ngIf="pageSizeComponent">
+      <ng-content select="clr-dg-page-size"></ng-content>
+    </div>
+    <div class="pagination-description">
+      <ng-content></ng-content>
+    </div>
+    <div class="pagination-list" *ngIf="page.last > 1">
+      <button class="pagination-first" [disabled]="page.current === 1" (click)="page.current = 1">
```
youdz: The cynical engineer in me shivers at the sight of `page.current === 1` instead of `page.current <= 1`. But I'm sure it's fine, there is now way underflow would ever happen. 😬

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -3,58 +3,45 @@
...
+        <clr-icon shape="angle-double left"></clr-icon>
+      </button>
+      <button class="pagination-previous" [disabled]="page.current === 1" (click)="page.current = page.current - 1">
+        <clr-icon shape="angle left"></clr-icon>
+      </button>
+      <input type="text" class="pagination-current" [size]="page.last.toString().length" [(ngModel)]="page.current" />&nbsp;/&nbsp;<span>{{page.last}}</span>
```
youdz: `[size]` works? Cool. I was expecting `[attr.size]`, but it's cool if Angular lets you bind directly to the property of the HTMLElement on this one.

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -3,58 +3,45 @@
...
+      <button class="pagination-next" [disabled]="page.current === page.last" (click)="page.current = page.current + 1">
+        <clr-icon shape="angle right"></clr-icon>
+      </button>
+      <button class="pagination-last" [disabled]="page.current === page.last" (click)="page.current = page.last">
```
youdz: Actually, on the last page I *will* ask you to use `page.current >= page.last`. I really don't want us to lose time fixing edgy edge cases of server-driven datagrids where pages disappear as you navigate to them.

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -3,58 +3,45 @@
...
+        <clr-icon shape="angle-double right"></clr-icon>
+      </button>
+    </div>
     `,
   host: { '[class.pagination]': 'true' },
 })
 export class ClrDatagridPagination implements OnDestroy, OnInit {
-  constructor(public page: Page) {}
+  @ContentChild(ClrDatagridPageSize) pageSizeComponent: ClrDatagridPageSize;
```
youdz: [Our naming conventions](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#properties-and-methods) say this is `_ pageSizeComponent`. Public because the way Angular works requires it, but prefixed to denote people shouldn't be messing with this.

src/clr-angular/data/datagrid/datagrid-pagination.ts
youdz: And... 🌮. I could see this comment requiring large changes. 😢 

You're using `[(ngModel)]` on a text input. 
1. What happens if the user types "hello"?
2. What happens if the user types a negative number? A page that's after the last page? I don't believe the pagination service puts guards against this.
3. This is the scary one: If I want to go to page 123, `[(ngModel)]` fires **as I type**. So it will move to page 1, then 12, then finally 123. If each of these triggers an expensive HTTP call to the server, it's bad. Not to mention the fact that if people didn't implement their fetch with proper cancellation, you'll have race conditions and might end up on page 123 with the items of page 12. Admittedly that's the app's fault, but I'd rather avoid having to document how to properly fire HTTP requests to the server and cancel them when they are in quick succession.

src/dev/src/app/datagrid/pagination/pagination.html
```diff
@@ -81,7 +81,9 @@ <h2>Pagination</h2>
   </clr-dg-row>
 
   <clr-dg-footer>
-    {{pagination.firstItem + 1}} - {{pagination.lastItem + 1}} of {{pagination.totalItems}} users
-    <clr-dg-pagination #pagination [clrDgPageSize]="10"></clr-dg-pagination>
+    <clr-dg-pagination #pagination [clrDgPageSize]="20">
+      <clr-dg-page-size [clrPageSizeOptions]="[10,20,50,100]">Users per page</clr-dg-page-size>
+      {{pagination.totalItems}} users
```
youdz: Nitpick, but can you put back the:
```
{{pagination.firstItem + 1}} - {{pagination.lastItem + 1}} of {{pagination.totalItems}} users
``` 
Or is that just not the recommended text anymore?

src/clr-angular/data/datagrid/datagrid-pagination.ts
jeeyun: 1 and 2: It doesn't blow up, but the Page service will set it to whatever it's passed. We can use the input `type=number` and play around with css for calculating proper width (although I think Edge didn't support `type=number` perfectly?). So maybe a better approach is to enforce validation at the component-level (e.g. keyup event?) and also add a safeguard to the Page's set current method. Thoughts?

3: Instead of using ngModel could we use a custom input/output here, and we use debounce to ignore rapid succession of changes in input value when we emit the output?

src/clr-angular/utils/_theme.dark.clarity.scss
jeeyun: That can be said about a lot of colors in this file though. Would it be ok to punt this by creating a generic issue to create SASS variables for dark theme colors?

mathisscott: Absolutely.

src/clr-angular/data/datagrid/datagrid-page-size.ts
mathisscott: I'm just wondering if there's an easier way for people to hide the dropdown if they don't want it. If they don't set the pageSizeOptions, it's pretty clear they don't want it. So why do they need to write CSS for every time they want to hide it?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -633,13 +633,58 @@
       display: flex;
       align-items: center;
 
+      &-size {
+        display: block;
+        flex: 1 1 auto;
+        white-space: nowrap;
```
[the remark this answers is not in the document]
jeeyun: I can remove the `nowrap` and leave it up to the client app to respond to smaller width. Or do you think we should be offering something by default?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -633,13 +633,58 @@
...
+
+        .clr-select-wrapper {
+          &::after {
+            top: 0.5rem;
+          }
+        }
+
+        select {
+          background: none !important;
+          background-color: #ffffff;
```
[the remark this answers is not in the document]
jeeyun: Yeah I think it's cleaner to use a specific class so I'll follow your suggestion.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -862,43 +907,59 @@
     justify-content: center;
     align-items: stretch;
 
-    > * {
-      padding: 0 0.1rem;
-      margin-left: 0.3rem;
+    input[type='text'] {
```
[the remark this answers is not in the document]
jeeyun: Yeah, no strong reason for the specificity. I'll just add a class to this.

src/dev/src/app/datagrid/pagination/pagination.html
jeeyun: Yes, we can do that. I was just following the sample text from the new pagination design spec.

src/clr-angular/data/datagrid/datagrid-page-size.ts
youdz: Oh, that's the misunderstanding. If they don't put the `<clr-dg-page-size>` component, everything works with the page size they defined themselves, and there is just no dropdown or page size selection at all. The component itself is opt-in.

src/clr-angular/data/datagrid/datagrid-pagination.ts
youdz: For 3, I would ignore everything until the user presses Enter or blurs the input.

This means that for 1 and 2, I would probably let the user type stupid things, and fallback on page 1 if they enter an invalid value (or on the last page if they enter a value greater than the number of pages).

src/clr-angular/data/datagrid/datagrid-page-size.ts
mathisscott: Oh. Got it now. I'm sorry for the confusion.

src/clr-angular/utils/_theme.dark.clarity.scss
jeeyun: Created an issue for this: https://github.com/vmware/clarity/issues/2783

PR #2757, opened 2018-10-18
[UI] Adds a backward compatibility solution for deprecated-grid

### LINKED ISSUE #2144: Backwards compatibility solutions for grid
We need to provide a backwards compatibility solution for the grid. This should be exporting the old grid into a separate file as a compatibility layer for optional inclusion.

hippee-lee:
* Breaking change: stylesheets for deprecated grid must now be included in the app
* Closes #2144


src/clr-angular/layout/grid-deprecated.clarity.scss
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+// @deprecated Bootstrap dependencies are deprecated since 1.0. Remove before 2.0.
+
+$enable-grid-classes: true;
+$imported-modules: () !default;
+
+@mixin exports($name) {
```
mathisscott: i thought `exports` was something we used everywhere. are we re-defining it here? or did we pull it out by accident?

hippee-lee: Its defined in the `_mixins.clarity.scss`. I was trying to keep this deprecated sctylesheet as light as possible. You are right though and I'll just include the global mixins.

gnomeontherun:
I compared what was removed with the output from the new build, and there are a few discrepancies. 

1. The `.container` and `.container-fluid` classes were removed and are missing from new bundle.
2. The deprecated file needs to be run through `csso` and/or `postcss` to have it add the `-ms` prefixed flex properties. I'm not sure how critical those prefixed rules are, but we have them now so seems prudent to retain them for the time being.

PR #2761, opened 2018-10-19
Fix modal boolean testing

### LINKED ISSUE #2760: Pressing ESC key throws console errors when there's a hidden (clrModalOpen is initially false) clr-modal element
## Describe the bug

_A clear and concise description of what the bug is, and the conditions when it occurs._

I have a clr-modal component and it's not shown initially. When I press the ESC key, it throws console errors as shown below:
image

However, if the modal is visible and if I press the ESC key it doesn't throw an error.
## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

I can't put this to stackblitz, the project is huge and needs a server-side api to run it.
This is how I code it:
image

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

## Expected behavior

## Versions

**App**

* Angular: [e.g. 6] 6
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5] 0.13.5

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

youdz:
Our boolean testing in modals was very weird. Being too strict could be a problem, so I'm changing it back to a more standard syntax.

Yes, this commit is only here for me to catch up to Scott.

Fixes #2760.

### DIFF
--- a/src/clr-angular/modal/modal.ts
+++ b/src/clr-angular/modal/modal.ts
@@ -85,7 +85,7 @@ export class ClrModal implements OnChanges, OnDestroy {
   }
 
   open(): void {
-    if (this._open === true) {
+    if (this._open) {
       return;
     }
     this._open = true;
@@ -98,7 +98,7 @@ export class ClrModal implements OnChanges, OnDestroy {
       this.altClose.emit(false);
       return;
     }
-    if (!this.closable || this._open === false) {
+    if (!this.closable || !this._open) {
       return;
     }
     this._open = false;


src/clr-angular/modal/modal.ts
```diff
@@ -85,7 +85,7 @@ export class ClrModal implements OnChanges, OnDestroy {
   }
 
   open(): void {
-    if (this._open === true) {
+    if (this._open) {
```
mathisscott: I thought using Typescript meant we could explicitly type properties as boolean? what's the benefit of typescript if we have to check for truthiness like it's 2009?

(you will _never_ catch me) 😈

youdz: With `strictNullChecks` set to `true`, the both are the same. Except that *we* can't enforce anything, so we have to be flexible and make our library work for apps that are coded like it's 2009.

PR #2773, opened 2018-10-23
[NG] Check listener is defined before invoking it

HitkoDev:
`listener` has `@Optional()` decorator, therefore we have to check it's not null before calling it.


HitkoDev:
@youdz Done

gnomeontherun:
@HitkoDev Can you please squash your commits into just one? Thanks!

HitkoDev:
@gnomeontherun Okay, did it.

Issue #725, opened 2017-04-14
Improve documentation on loading clarity-icons in typescript


Shijir:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x ] enhancement
```

### Expected behavior

**Improve documentation on loading `clarity-icons` shape modules:**

Currently, when users load `clarity-icons` in Typescript, only `core-shapes` gets loaded. This may lead them to confusion as they expect all shape modules such as `core-shapes`, `essential-shapes`, `social-shapes`, and `technology-shapes` will be loaded together by default. The reason why only `core-shapes` loaded by `import 'clarity-icons'` is that we shouldn't dump all the shape markups into users' app. Instead, we wanted users to leverage the modularity of typescript so they can load only shape modules they would use.

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

PR #2782, opened 2018-10-25
[NG] new checkbox implementation

gnomeontherun:
This removes the old clr-checkbox implementation for clrCheckbox to align with new forms.

It also replaces a few uses of old checkbox internally in tree view and datagrid.

Part of #1878


hippee-lee:
Does `package-lock.json` need to change or can we revert it?

package.json
```diff
@@ -70,11 +70,11 @@
...
-    "@angular-devkit/build-angular": "^0.10.1",
-    "@angular-devkit/build-ng-packagr": "^0.10.1",
-    "@angular-devkit/build-optimizer": "^0.10.1",
-    "@angular-devkit/core": "^7.0.0-rc.2",
-    "@angular/cli": "7.0.1",
+    "@angular-devkit/build-angular": "^0.10.3",
+    "@angular-devkit/build-ng-packagr": "^0.10.3",
+    "@angular-devkit/build-optimizer": "^0.10.3",
+    "@angular-devkit/core": "^7.0.3",
+    "@angular/cli": "^7.0.3",
```
hippee-lee: Do these changes depend on thes devDep versions?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -719,29 +707,19 @@
...
-      line-height: inherit;
-    }
-
-    input[type='checkbox'] + label {
-      &::before,
-      &::after {
-        top: 0.333333rem;
-      }
+  .clr-form-control-disabled .datagrid-footer-select.clr-checkbox-wrapper input[type='checkbox']:checked + label {
+    top: calc(1rem / 6);
```
hippee-lee: Did this get tested in IE11? (referencing line #725 from the deleted code)

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -78,6 +79,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
     nbRow++;
     this.id = 'clr-dg-row' + nbRow;
     this.radioId = 'clr-dg-row-rd' + nbRow;
+    this.checkboxId = 'clr-dg-row-cb' + nbRow;
```
hippee-lee: Does this need a spec?

src/clr-angular/data/tree-view/_tree-view.clarity.scss
```diff
@@ -148,17 +148,16 @@
...
     width: $clr-tree-node-touch-target;
     padding-top: $clr-tree-checkbox-padding-top;
     padding-left: $clr-tree-checkbox-padding-left;
+  }
 
-    //States
-    &:first-child {
-      margin-left: $clr-tree-node-touch-target;
-    }
+  .clr-tree-node-content-container > .clr-checkbox-wrapper:first-child {
```
hippee-lee: I know we want to keep nesting down but re-declaring the selectors but this seems like something un-necesarily duplicated.

package.json
gnomeontherun: So I was having issues with running tests and found that some of these didn't get bumpped to latest with Angular 7 updates. That explains the package and package-lock changes, though I can pull them into a separate PR.

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -4,23 +4,120 @@
...
-export class ClrCheckboxContainer implements DynamicWrapper {
-  // Indicates whether the container is dynamically created by the checkbox input itself
-  _dynamic = false;
+export class ClrCheckboxContainer implements OnDestroy {
+  private subscriptions: Subscription[] = [];
+  invalid = false;
+  @ContentChild(ClrLabel) label: ClrLabel;
+  private inline = false;
+  control: NgControl;
+  // private formGroup: AbstractControl;
```
hippee-lee: Remove?

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -4,23 +4,120 @@
...
 })
-export class ClrCheckboxContainer implements DynamicWrapper {
-  // Indicates whether the container is dynamically created by the checkbox input itself
-  _dynamic = false;
+export class ClrCheckboxContainer implements OnDestroy {
+  private subscriptions: Subscription[] = [];
+  invalid = false;
+  @ContentChild(ClrLabel) label: ClrLabel;
+  private inline = false;
+  control: NgControl;
```
hippee-lee: Should this be explicitly set public since its used in the template?

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -4,23 +4,120 @@
...
+      this.inline = value === 'false' ? false : true;
+    } else {
+      this.inline = !!value;
+    }
+  }
+  get clrInline() {
+    return this.inline;
+  }
+
+  // @TODO Solve for group validation, which doesn't work now with ngModelGroup
```
hippee-lee: I'm assuming this is here so it can be used once the blocker is resolved. Correct?

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -4,23 +4,120 @@
...
+  ) {
+    this.subscriptions.push(
+      this.ngControlService.controlChanges.subscribe(control => {
+        this.control = control;
+      })
+    );
+  }
+
+  ngOnInit() {
+    // @TODO put a solution in for form group validation
```
hippee-lee: Does this still need to be done?

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  providers: [ControlIdService],
+})
+export class ClrCheckboxWrapper implements DynamicWrapper {
+  // We need both _dynamic for HostWrapper and ContentChild(ClrLabel) in cases where
+  // the user puts a radio inside a wrapper without a label, host wrapping doesn't apply
+  // but we'd still need to insert a label
+  _dynamic = false;
+  @ContentChild(ClrLabel) label: ClrLabel;
+  hasContainer = false;
```
hippee-lee: Two things here. 1) I can't find anything that uses this. 2) If it needs to be here, should it be marked as public so its visiblitly is explicit?

src/clr-angular/forms/common/if-error/if-error.service.ts
```diff
@@ -27,8 +27,10 @@ export class IfErrorService implements OnDestroy {
     // Wait for the control to be available
     this.subscriptions.push(
       this.ngControlService.controlChanges.subscribe(control => {
-        this.control = control;
-        this.listenForChanges();
+        if (control) {
```
hippee-lee: Is this guard redundant? What I mean by that is I haven't heard of a subscribe callback being called without an argument. Is that possible?

src/clr-angular/forms/common/label.spec.ts
```diff
@@ -43,9 +49,10 @@ export default function(): void {
     });
 
     it("doesn't set the the class unless its inside of a container", function() {
-      TestBed.configureTestingModule({ declarations: [ClrLabel, NoForTest], providers: [NgControlService] });
+      TestBed.configureTestingModule({ declarations: [ClrLabel, NoForTest] });
       const fixture = TestBed.createComponent(NoForTest);
       fixture.detectChanges();
+      console.log(fixture.debugElement.query(By.css('label')).nativeElement);
```
hippee-lee: Remove?

package.json
hippee-lee: No worries. imo - its ok to have this type of change here when it addresses/fixes something for the PR. Just wanted to confirm that it was intended.

src/clr-angular/data/tree-view/_tree-view.clarity.scss
gnomeontherun: It was necessary to target with the > selector or it was not specific enough.

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -331,7 +331,7 @@ export default function(): void {
       });
 
       it('displays a checkbox when selectable is true', () => {
-        const checkbox: HTMLElement = compiled.querySelector('clr-checkbox');
+        const checkbox: HTMLElement = compiled.querySelector('[clrCheckbox]');
```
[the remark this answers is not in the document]
gnomeontherun: Isn't necessary, and we do this elsewhere so I'm going to leave it.

src/clr-angular/deprecations.spec.ts
```diff
@@ -7,4 +7,8 @@
 describe('Deprecations', () => {
   // When we deprecate some code, we should write a test to verify it is still in the bundle
   // and keep track of when it was deprecated, and when we plan to remove it.
+
+  describe('1.0', () => {
+    it('should no longer support clr-checkbox in button-groups');
```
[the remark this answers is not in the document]
gnomeontherun: This is to help me keep track of it yes. This is a breaking change we cannot avoid right now without working on button groups, and so there isn't much to test since its just removed. Maybe we'll support it again though after 1.0.

src/clr-angular/forms/checkbox/checkbox-container.ts
gnomeontherun: We don't have a strict convention to mark as public because by default it is. It will get refactored soon in #2678 anyways.

src/clr-angular/forms/checkbox/checkbox-container.ts
gnomeontherun: This was to keep track of my progress for another day, yes.

src/clr-angular/forms/checkbox/checkbox-container.ts
gnomeontherun: This is part of the future work we need to do when we support checkbox group validation.

src/clr-angular/forms/common/if-error/if-error.service.ts
gnomeontherun: In the control components, the NgControl is injected with `@Optional()` and if there was no control it will set a `null` value through the observable. So yes this is needed, but should probably guard as well at the observable level. Since injection is runtime we can't be certain that we'll get the value or not.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
gnomeontherun: `calc()` can be used in IE11 for `top` so should be ok. will do some testing to confirm

src/clr-angular/data/datagrid/datagrid-row.ts
gnomeontherun: There weren't any for the radio before, so added both.

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
gnomeontherun: This appears to be carried over incorrectly, removing.

src/clr-angular/forms/common/label.spec.ts
gnomeontherun: Yup

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: Sounds good.

src/clr-angular/forms/checkbox/checkbox-container.ts
gnomeontherun: I kept this also because its part of the formGroup implementation, the other similar parts are commented out as well.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -719,29 +707,19 @@
...
-      }
+    margin-right: 0.375rem;
+
+    &::before {
+      background-color: $clr-gray;
+    }
+
+    &::after {
+      border-left-color: $clr-white;
+      border-bottom-color: $clr-white;
```
jeeyun: Are these to remain the same for dark theme?

gnomeontherun: I don't believe so, but checking again with @hippee-lee

gnomeontherun: There doesn't appear to be any breaking changes for existing datagrid, but we did see a bug in master that is unrelated to these changes.

JoshuaNitschke:
The documentation still needs to be updated: https://vmware.github.io/clarity/documentation/v1.0/checkboxes

gnomeontherun:
It hasn't been released yet, but will be updated.

farlock85:
@gnomeontherun do you know which release this will go in? We're waiting on checkboxes to be restored again.

farlock85:
@mcosta74 Sorry I might have misread this issue. I was talking about the checkboxes on forms, at the moment they are not implemented.

gnomeontherun:
@farlock85 Checkboxes are part of 1.0.0-rc.1 and above, see https://vmware.github.io/clarity/documentation/v1.0/checkboxes.

PR #2786, opened 2018-10-27
Added date-picker cancel icon. Functionality yet to be added.

AbhiAgarwal192:


gnomeontherun:
We don't have any UX specification for a close button in the date picker, so I'm not sure what your goal is. What is your intention with this PR?

youdz:
We cannot ship features without functionality, if this were to go in every UI using Clarity would have a non-functional buttons in their Datepickers.

Please only contribute fully-implemented features.

PR #2797, opened 2018-10-30
[NG] set default forms layout to horizontal in Angular forms

gnomeontherun:
This changes the default form layout to horizontal for Angular forms, to match our recommendation.


src/dev/src/app/datepicker/css-regression.html
```diff
@@ -5,11 +5,9 @@
...
 <div class="clr-example-1" clrForm>
-    <input type="date" clrDate>
+  <input type="date" clrDate>
 </div>
 
 <div class="clr-example-2" clrForm>
-    <div class="clr-example-input">
-        <input type="date" [(clrDate)]="date">
-    </div>
+  <input type="date" [(clrDate)]="date">
```
hippee-lee: Do you have a change for the gemini tests with this change?

src/clr-angular/forms/tests/container.spec.ts
```diff
@@ -127,25 +127,25 @@ function fullSpec(description, testContainer, directives: any | any[], testCompo
...
       expect(containerEl.classList).toContain('clr-row');
+      layoutService.layout = Layouts.VERTICAL;
+      fixture.detectChanges();
+      expect(containerEl.classList).not.toContain('clr-row');
       layoutService.layout = Layouts.COMPACT;
       fixture.detectChanges();
       expect(containerEl.classList).toContain('clr-row');
     });
 
     it('computes the error class for the control container', () => {
```
hippee-lee: Personal nit, not a blocker.
Instead of `computes` could this be changed to `adds`? Reading that makes more sense to me for what is being tested.

src/clr-angular/forms/common/providers/layout.service.ts
```diff
@@ -14,7 +14,7 @@ export enum Layouts {
 
 @Injectable()
 export class LayoutService {
-  layout: Layouts = Layouts.VERTICAL;
+  layout: Layouts = Layouts.HORIZONTAL;
```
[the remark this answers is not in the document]
gnomeontherun: It is internal to forms, so I didn't move it out from when it was originally done.

src/dev/src/app/datepicker/css-regression.html
gnomeontherun: Yes, I threw it up to see what tests failed on Gemini instead of doing it locally, faster.

src/clr-angular/forms/tests/container.spec.ts
gnomeontherun: Sure

gnomeontherun:
FYI - the reason there are so many gemini changes is because I uncovered a small bug when using horizontal/compact layouts that the form didn't apply the `.clr-form` class. The changes are simply adjustments for the padding of `.clr-form` to the layout.

PR #2803, opened 2018-11-01
Datepicker bugfixes

### LINKED ISSUE #2474: DatePicker reset with Form usage
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

When I use FormControl on my DatePicker field, if I apply a reset on my form, I didn't want DatePicker remember the old date. I should enter the same date as before.

### Actual behavior

When I reset my FormControl, I can't choose the same date as before. The only way I found is to use the "Two-way binding" and declare var for each datePicker and set manually "undefined" for each var when I reset my form.

* **Angular version:** 5.1.0

* **Clarity version:** 0.11.17

### LINKED ISSUE #2651: New from layout Date / select / text input design bugs
## Describe the bug

New from layout Date / select / text input design bugs

## How to reproduce

https://stackblitz.com/edit/clr-date-form-issue

1.  Go to stackblitz
2.  See the input fields, they are not below each others, date container uses different spacing then the others.

I've tried to use `<clr-date-container>`, but it was worse.

The first cause is related to the angular `<clr-input-container>` bug, which places the clr-col-8 to the wrong container, but the spacing of date container is a different issue.

## Expected behavior

All inputs and labels are in place, no different spacing, it worked in the previous version well.
The native html with clr-control-container, clr-input-wrapper works fine, 
 `<clr-input-container>`, `<clr-select-container>`, `<clr-date-container>` components should be fixed.
Please provide a documentation, how to use datepicker with new containers. 

## Versions

**App**

- Angular CLI: 6.1.3
- Node: 8.11.2
- Clarity: 0.13
- Device:
- 
- Type: Dell latitude E5550
- OS: Windows 10 64bit
- Browser: Chrome - Version: 69.0.3497.92 (Official Build) (64-bit)

gnomeontherun:
This includes two fixes for Datepicker.

fixes #2474 - This is a bug that if you reset the control that you cannot reselect the same value
fixes #2651 - This bug was that the date input control didn't properly allow a grid override setting

I added a new gemini test to capture all the layout options for datepicker (like the other form controls).


src/dev/src/app/datepicker/datepicker.demo.html
```diff
@@ -15,7 +15,8 @@ <h2>Datepicker</h2>
             <li><a [routerLink]="['./disabled']">Disabled</a></li>
         </ul>
     </div>
-    <div class="col-xs-12 col-sm-6 col-md-4">
+    <div class="clro-col-xs-12 clr-col-sm-6 clr-col-md-4">
```
mathisscott: typo at `clro-col-xs-12`

PR #2805, opened 2018-11-01
fix(dark-theme): Selected datagrid row font color should be white

hippee-lee:


mathisscott:
Do we need a gemini test for this?

gnomeontherun:
We have datagrid gemini tests off for now, they were being flaky. That needs more investigation.

PR #2811, opened 2018-11-05
[WEB] fix radio button example

mcosta74:
Small fix for the "Radio Buttons: Demo page.

Selecting a radio button from the last "inline radio example" group will select a radio button from the above group.

Signed-off-by: Massimo Costa <costa.massimo@gmail.com>


mcosta74:
If you already fixed (and maybe not yet released) feel free to close this PR

PR #2812, opened 2018-11-05
[NG] fix memory leak in focus trap

Shijir:
**Before**:

screenshot 2018-11-05 10 38 10

Angular Render2's `createElement` keeps the traces of the detached doms. Each time I open a modal component, the JS heap size kept getting the retained size of 40 bytes.

**After**:

After resorting the native DOM's `createElement` method, I found no traces of left over objects and the retained size became 0 bytes.

screenshot 2018-11-05 10 36 51

Also, I communicated with Mustafa and received his confirmation on the solution.

### DIFF
--- a/src/clr-angular/utils/focus-trap/focus-trap.directive.ts
+++ b/src/clr-angular/utils/focus-trap/focus-trap.directive.ts
@@ -49,7 +49,8 @@ export class FocusTrapDirective implements AfterViewInit, OnDestroy {
   }
 
   private createFocusableOffScreenEl(): any {
-    const offScreenSpan = this.renderer.createElement('span');
+    // Not using Renderer2's createElement method because that leads to DOM leakage.
+    const offScreenSpan = this.document.createElement('span');
     this.renderer.setAttribute(offScreenSpan, 'tabindex', '0');
     this.renderer.addClass(offScreenSpan, 'offscreen-focus-rebounder');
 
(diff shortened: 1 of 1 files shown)


src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -49,7 +49,8 @@ export class FocusTrapDirective implements AfterViewInit, OnDestroy {
   }
 
   private createFocusableOffScreenEl(): any {
-    const offScreenSpan = this.renderer.createElement('span');
+    // Not using Renderer2's createElement method because that leads to DOM leakage.
```
youdz: I already approved, but actually can you add here a link to the Angular issue you're about to open. It'll be good to go back to the renderer once they fix the memory leak on their side.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
```diff
@@ -49,7 +49,8 @@ export class FocusTrapDirective implements AfterViewInit, OnDestroy {
...
+    const offScreenSpan = this.document.createElement('span');
```
gnomeontherun: Does this work right with SSR? To be safe, I think we should update `ngAfterViewInit` so it doesn't try run this code branch on the server.

```
    ngAfterViewInit() {
     if (isPlatformBrowser(this.platformId)) {
       this.previousActiveElement = <HTMLElement>this.document.activeElement;
       this.addReboundEls();
     }
   }
```

Shijir: The `addReboundEls()` method itself checks whether it's running in the browser or not. So we only add these off-screen elements when the app is running in a browser.

src/clr-angular/utils/focus-trap/focus-trap.directive.ts
Shijir: Added the URL to the GitHub issue.

PR #2813, opened 2018-11-05
[NG] Models for the new tree

youdz:
Just introducing the models for now. They share the selection logic and propagation, and the specifics of each are that:
- The declarative model is pretty dumb since it just waits for the components to create it. It just registers/unregisters to the parent to make sure our structure is up to date.
- The recursive model is smarter and allows us to build the whole tree in memory (by calling `fetchChildren()` recursively, hence the name) even if we don't have the corresponding components. We will use that model to instantiate the node components ourselves.

Also note that I'm abandoning the double, tightly-couple, binding for both `selected` and `indeterminate`. It created way too many issues both for us and for the apps. I'm using a single binding with a three-values enum, which ends up being much more natural to use and cuts down a lot of the internal logic.

Travis is going to crash and burn on this preliminary PR, obviously, but I don't really care yet.


src/clr-angular/data/tree-view/models/tree-node-model.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export abstract class TreeNodeModel<T> {
+  selected = new BehaviorSubject<ClrSelectedState>(ClrSelectedState.UNSELECTED);
+  model: T | null;
+  // For some reason the polymorphic this doesn't work for either of my use cases. Dammit, Typescript.
+  // So I'm resorting to forcing override with more precise types by marking these abstract.
+  abstract parent: TreeNodeModel<T> | null;
+  abstract children: TreeNodeModel<T>[];
+
+  loading = false;
```
gnomeontherun: Seems this is only needed in recursive lazy loaded tree, should we move this up there?

src/clr-angular/data/tree-view/models/selected-state.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+// TODO: I'd like this to be a CheckedState enum for the checkboxes in the future.
+export enum ClrSelectedState {
```
gnomeontherun: Unless we are expecting external use of this enum, we should name it `SelectedState`. I also suggest `CheckboxSelectedState` or something a little more precise.

src/clr-angular/data/tree-view/models/tree-node-model.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+    if (propagateUp && this.parent) {
+      this.parent._updateSelectionFromChildren();
+    }
+  }
+
+  toggleSelection(propagate: boolean) {
+    // NOTE: we always propagate selection up in this method because it is only called when the user takes an action.
+    // It should never be called from lifecycle hooks or app-provided inputs.
+    switch (this.selected.value) {
```
gnomeontherun: Just preference, but seems like easier to do an `if/else` here since there are functionally only two cases.

src/clr-angular/data/tree-view/models/tree-node-model.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    it('starts unselected', function() {
+      expect(root.selected.value).toBe(ClrSelectedState.UNSELECTED);
+    });
+
+    it('allows to set the selection state of a node without propagating', function() {
+      // Selected
+      child.setSelected(ClrSelectedState.SELECTED, false, false);
+      expect(child.selected.value).toBe(ClrSelectedState.SELECTED);
+      [root, ...child.children].forEach(n => expect(n.selected.value).toBe(ClrSelectedState.UNSELECTED));
```
gnomeontherun: 🥇

src/clr-angular/data/tree-view/models/tree-node-model.ts
```diff
@@ -0,0 +1,100 @@
...
+  selected = new BehaviorSubject<ClrSelectedState>(ClrSelectedState.UNSELECTED);
+  model: T | null;
+  // For some reason the polymorphic this doesn't work for either of my use cases. Dammit, Typescript.
+  // So I'm resorting to forcing override with more precise types by marking these abstract.
+  abstract parent: TreeNodeModel<T> | null;
+  abstract children: TreeNodeModel<T>[];
+
+  loading = false;
+
+  destroy() {
```
gnomeontherun: Why wouldn't you want to recursively complete the children observables here? Otherwise you have to call each individually.

Also this is only tested in the other extended classes.

src/clr-angular/data/tree-view/models/tree-node-model.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Just to be safe
+    this.selected.complete();
+  }
+
+  // Propagate by default when smart, don't propagate in the lazy-loaded tree.
+  setSelected(state: ClrSelectedState, propagateUp: boolean, propagateDown: boolean) {
```
gnomeontherun: Should this be private?

src/clr-angular/data/tree-view/models/declarative-tree-node-model.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    afterEach(function() {
+      parent.destroy();
+    });
+
+    it('extends TreeNodeModel', function() {
+      expect(parent instanceof TreeNodeModel).toBeTrue();
+    });
+
+    it('registers to its parent on creation', function() {
```
gnomeontherun: This doubles as  a test for setting up children at the same time, can you update the description?

src/clr-angular/data/tree-view/models/declarative-tree-node-model.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      const child1 = new DeclarativeTreeNodeModel<void>(parent);
+      const child2 = new DeclarativeTreeNodeModel<void>(parent);
+      expect(parent.children).toEqual([child1, child2]);
```
gnomeontherun: You aren't calling `child1.destroy()` and `child2.destroy()` so memory leak?

src/clr-angular/data/tree-view/models/tree-node-model.ts
```diff
@@ -0,0 +1,100 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { ClrSelectedState } from './selected-state';
+import { BehaviorSubject } from 'rxjs';
+
+export abstract class TreeNodeModel<T> {
+  selected = new BehaviorSubject<ClrSelectedState>(ClrSelectedState.UNSELECTED);
+  model: T | null;
```
gnomeontherun: Only used in recursive?

src/clr-angular/data/tree-view/models/recursive-tree-node-model.ts
```diff
@@ -0,0 +1,78 @@
...
+        this._children = this.wrapChildren(raw);
+        this.loading = false;
+      });
+    } else if (asyncChildren) {
+      // Synchronous case
+      this._children = this.wrapChildren(asyncChildren);
+    } else {
+      this._children = [];
+    }
+    this.childrenFetched = true;
```
gnomeontherun: What if they want to re-fetch the children for a given node (perhaps the backend refreshed)?

src/clr-angular/data/tree-view/models/selected-state.ts
youdz: - It needs to be public, since we're going to offer a `[(clrSelected)]` two-way binding accepting and emitting this enum. This replaces the double two-way binding of `[(clrSelected)]` and `[(clrIndeterminate)]`.
- It is not tied to checkbox, it just describes a selection state. It turns out our implementation uses checkboxes, but Until we move checkboxes to this pattern, I'd like to keep the name abstract in that it just describes if a consumer's object is selected or not. And it's used in Tree here, but it could make sense in combobox or complex list components.

src/clr-angular/data/tree-view/models/tree-node-model.ts
youdz: Yes, I'm missing a unit test here. As I said on Slack, I was just trying to have minimal test coverage to meet the deadline, but I can add one for this.

I initially had it recursively detroy, but not only will the components do it for you (which isn't really an argument for this PR), even in the pure model it makes sense to want to destroy a parent without destroying the children. So I decided to keep it local. I don't feel strongly either way.

src/clr-angular/data/tree-view/models/tree-node-model.ts
youdz: `Toggle` doesn't let you control the target state, and doesn't let you set to indeterminate. That's why this is public.

src/clr-angular/data/tree-view/models/declarative-tree-node-model.spec.ts
youdz: I was being overly cautious in the other places, but for this one the Observables are never subscribed, so there is no memory leak. I can still destroy them if you want.

hippee-lee:
General question for the `model` files. We have conventions for Filenames for providers, interfaces and enums that seperate the _thing_ like this `my-file.thing.ts` e.g `sort-order.enum.ts` Does it make sense to do that for the model files as well? FOr example `recursive-tree-node.model.ts` instead of `recursive-tree-node-model.ts`? 

[Coding guidlines](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#file-names)

youdz:
@hippee-lee: I don't know why I didn't think of it, I like it. Unless someone feels strongly against it, I'll rename the files. 👍

src/clr-angular/data/tree-view/models/recursive-tree-node-model.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private wrapChildren(rawModels: T[]) {
```
hippee-lee: Is return type useful here? e.g 
```private wrapCHildren(rawModels: T[]): RecursiveTreeNodeModel<T>[]```

File: src/clr-angular/data/tree-view/models/tree-node-model.spec.ts [hunk 12]
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { TreeNodeModel } from './tree-node-model';
+import { BehaviorSubject } from 'rxjs';
+import { ClrSelectedState } from '@clr/angular';
```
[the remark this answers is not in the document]
hippee-lee: To make sure the universe is balanced, I've had similar behavior happen with VSCode.

mathisscott:
@youdz 
@hippee-lee [wrote a great blogpost that may be helpful](https://medium.com/claritydesignsystem/a-clean-contribution-ec9e459ce46) for the DCO requirement.

youdz:
screen shot 2018-11-08 at 17 44 33

I have no idea what you're talking about. 😇

src/clr-angular/data/tree-view/models/recursive-tree-node-model.ts
youdz: That's a greater question. 😛 

Because of time constraints, I considered they could refresh the tree itself if children have changed. Not ideal obviously, but not a blocker as far as I could tell. I believe the next feature we could add is that if the children received are an Observable, we keep up to date with it. That would be a way for them to push dynamic changes.

Are you okay with us moving forward without it for now? If yes I'll create an issue that we can address in 1.x.

gnomeontherun: I'm ok with this being a future enhancement, doesn't sound like a breaking change.

gnomeontherun:
Think it makes sense and next commit will give this stuff additional context.

PR #2820, opened 2018-11-07
[INTERNAL] Quick incremental improvement of our test helpers

youdz:
As my comment in the code says, the generic helpers in `utils/` were initially a copy-paste of the datagrid helpers, but when we modularized Clarity they got updated to something that's way too rigid and doesn't help for complex specs anymore. As proof, only 2 specs were able to use it, and several others around the project have a comment along the lines of "Screw it, I'm using the datagrid ones".
To make it even worse, these early Datagrid test helpers were written at a time were the Typescript Language Service didn't exist and IDEs didn't provide the latest and greatest typing tools, so they were heavily compromising on the way to use them.

I renamed some pieces of the context to use clearer property names, which lead to a dumb but sizeable change in the wizard spec.
I also wrote a new version of the utility that finally fully embraces the typing of `this`, now that we can. To show a little bit how to use it, I updated the Signpost spec to the new version. It honestly doesn't change much for that spec, but the added flexibility it's one step towards being able to unify our unit tests.

PS: I'm using this in the Tree unit tests, yes.

### DIFF
--- a/src/clr-angular/utils/testing/helpers.spec.ts
+++ b/src/clr-angular/utils/testing/helpers.spec.ts
@@ -8,38 +8,49 @@
  * when we have the time. This will be very helpful in future refactors due to Angular upgrades, or simply
  * just to avoid leaks since destroying fixtures is automatic with this.
  */
-import { DebugElement, ModuleWithProviders, Type } from '@angular/core';
-import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { DebugElement, InjectionToken, ModuleWithProviders, Type } from '@angular/core';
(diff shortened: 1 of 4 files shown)


src/clr-angular/utils/testing/helpers.spec.ts
```diff
@@ -8,38 +8,49 @@
  * when we have the time. This will be very helpful in future refactors due to Angular upgrades, or simply
```
hippee-lee: It doesn't seem like the above comment (line #7) still applies here.

>These helpers are local to Datagrid at the moment ...

Issue #2599, opened 2018-08-28
Adding one icon adds all shapes from set


mme-private:
## Describe the bug

Adding one clarity icon increases main.js bundle size (ng build --prod) by ~260KB

## How to reproduce

1. Create new project and add new dependencies
```bash
ng new my-app --style=scss
cd my-app
yarn add @clr/angular @clr/icons @clr/ui bootstrap@4.0.0-alpha.5 @webcomponents/custom-elements@1.0.0
```

2. Change angular.json to
```json
"styles": [
  "node_modules/@clr/icons/clr-icons.min.css",
  "src/styles.scss"
],
"stylePreprocessorOptions": {
  "includePaths": [
  "node_modules"
  ]
},
  "scripts": [
   "node_modules/@webcomponents/custom-elements/custom-elements.min.js"
  ]
```
3. Change style.css to
```scss
@import "@clr/ui/src/utils/dependencies.clarity";
@import "@clr/ui/src/utils/components.clarity";
```
4. Change app-module.ts to
```typescript
import { BrowserModule } from '@angular/platform-browser';
import { NgModule } from '@angular/core';

import { ClarityModule } from '@clr/angular';
import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
import { AppComponent } from './app.component';

@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    BrowserModule,
    ClarityModule,
    BrowserAnimationsModule
  ],
  providers: [],
  bootstrap: [AppComponent]
})
export class AppModule { }
```

5. Run build for the first time
```bash
ng build --prod
```
6. **Check ./dist/my-app/main.js size (main.d75896a6224663291cac.js (main) 308 kB)**
7. Change app.component.ts to
```typescript
import { Component } from '@angular/core';

import { ClarityIcons } from '@clr/icons';
import { ClrShapeTrash } from '@clr/icons/shapes/essential-shapes';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.scss']
})
export class AppComponent {
  title = 'my-app';

  constructor() {

    ClarityIcons.add({
      trash: ClrShapeTrash,
    });
  }
}
```
8. Run build second time
```bash
ng build --prod
```
9. **Check ./dist/my-app/main.js size (main.f524c5eda53ee31939a4.js (main) 574 kB)** 

So size increment is more than 260KB.
[...]

PR #2821, opened 2018-11-08
[ICONS] Refactor to not auto-include the whole set when importing one icon

jeeyun:
With this refactor, importing one icon will not result in adding all shapes from the set as per #2599 .

Now we have `-umd.js` and `-umd.min.js` files for each shape, that are meant to be included in script tags in static sites. Including those scripts will result in adding all shapes from that file.

To summarize usage:
## Angular clients using Typescript

```
import { ClarityIcons } from '@clr/icons';
import { EssentialShapes } from '@clr/icons/shapes/essential';
import { ClrShapeStar } from '@clr/icons/shapes/social';

ClarityIcons.init(); // this should be done once at some root level of the application
ClarityIcons.add(EssentialShapes); // adds the whole set
ClarityIcons.add(ClrShapeStar); // adds a single icon from this set
```

## Static sites loading javascript files via script tag
To include all icons:
```
<script src="clr-icons/clr-icons-all-shapes.min.js"></script>
```
Or to include specific icon sets:
```
<script src="clr-icons/clr-icons-base.min.js"></script>
<script src="clr-icons/shapes/core-umd.min.js"></script>
<script src="clr-icons/shapes/trave-umd.min.js"></script>
```


src/clr-icons/shapes/chart-shapes-umd.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { ChartShapes, ClrShapeLineChartSVG } from './chart-shapes';
+import { descriptorConfig } from '../utils/descriptor-config';
+
+/* tslint:disable:variable-name */
```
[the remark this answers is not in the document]
jeeyun: The file name changes in #2 make sense to me. For #1, are you suggesting that we change the name in both ts and js versions? So our imports will be: 

```import { EssentialShapes } from '@clr/icons/shapes/essential';```

And our script tags:

```<script src="clr-icons/shapes/essential-umd.min.js"></script>```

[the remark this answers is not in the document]
Comment (gnomeontherun) [reply src/clr-icons/shapes/chart-shapes-umd.ts]: Correct, seems simpler to me.

File: scripts/clr-icons-svg.js [hunk 2]
```diff
@@ -7,19 +7,10 @@
...
-  'core-shapes',
-  'commerce-shapes',
-  'essential-shapes',
-  'media-shapes',
-  'social-shapes',
-  'technology-shapes',
-  'travel-shapes',
-  'chart-shapes',
-];
+const SHAPE_SETS = ['core', 'commerce', 'essential', 'media', 'social', 'technology', 'travel', 'chart'];
```
Shijir: Could you add `text-edit-shapes` here?

src/clr-icons/clr-icons-api.ts
```diff
@@ -63,6 +64,16 @@ export class ClarityIconsApi {
     }
   }
 
+  init(options: any = { includeCore: true }) {
```
Shijir: I think it's better to use an interface here instead of this `any` type.

scripts/clr-icons-svg.js
jeeyun: Thanks for catching that. Should be fixed now.

src/clr-icons/clr-icons-api.ts
jeeyun: Yeah. We made a change to actually pass in any set optionally for init(). You can take a look at the latest changes. The reason for this change is.. if we explicitly have a shapeSet like the way I had it initially it doesn't tree-shake the way we want it to.

scripts/write-svg-icons.js
```diff
@@ -177,12 +177,14 @@ let makeSVG = (shapeTitle, shapeContent) => {
...
-  // will convert core-shapes into CoreShapes
-  return kebabCase
-    .split('-')
-    .map(splitPart => splitPart.charAt(0).toUpperCase() + splitPart.slice(1))
-    .join('');
+let convertToExportedName = setName => {
+  // will convert core into CoreShapes
+  return (
+    setName
+      .split('-')
```
gnomeontherun: I don't think we need to split anymore, just uppercase the first letter and append `Shapes`.

src/clr-icons/api-autoload.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { ClarityIconsApi } from './api';
```
gnomeontherun: Let's rename this to `api-autoload.umd.ts`for consistency with naming umd files.

src/clr-icons/clr-icons.spec.ts
```diff
@@ -35,18 +35,23 @@ import { clrIconSVG } from './utils/svg-tag-generator';
...
+  { name: 'Commerce shapes', shapes: CommerceShapes, randomShape: ClrShapeECheck },
+  { name: 'Essential shapes', shapes: EssentialShapes, randomShape: ClrShapePencil },
+  { name: 'Social shapes', shapes: SocialShapes, randomShape: ClrShapeStar },
+  { name: 'Media shapes', shapes: MediaShapes, randomShape: ClrShapePlay },
+  { name: 'Travel shapes', shapes: TravelShapes, randomShape: ClrShapeCar },
+  { name: 'Technology shapes', shapes: TechnologyShapes, randomShape: ClrShapeCPU },
+  { name: 'Chart shapes', shapes: ChartShapes, randomShape: ClrShapeBarChart },
+  { name: 'Text Edit shapes', shapes: TextEditShapes, randomShape: ClrShapeBold },
 ];
 /* tslint:enable:no-string-literal */
```
gnomeontherun: Can we add a test case scenario where we don't run `init()` and to see that it doesn't set the global object?

src/clr-icons/clr-icons.spec.ts
```diff
@@ -391,7 +399,7 @@ describe('ClarityIcons', () => {
 
       const clrIconUniqId = clarityIcon.clrIconUniqId;
       const testShape = giveAngleShapeTitle(clrIconUniqId, customTitle);
-
+      console.log(clarityIcon);
```
gnomeontherun: Remove

src/clr-icons/shapes/essential.umd.ts
```diff
@@ -0,0 +1,48 @@
...
+  ClrShapeShrinkSVG,
+  ClrShapeTimesCircleSVG,
+  ClrShapeUsersSVG,
+  EssentialShapes,
+} from './essential';
+import { descriptorConfig } from '../utils/descriptor-config';
+
+export * from './essential';
+
+Object.defineProperty(EssentialShapes, 'edit', descriptorConfig(ClrShapePencilSVG));
```
gnomeontherun: I believe we can remove all of these aliases from the umd files right?

src/dev/src/main.ts
```diff
@@ -3,8 +3,8 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import '@clr/icons';
-import '@clr/icons/shapes/essential-shapes';
+import { ClarityIcons } from '@clr/icons';
+import { AllShapes } from '@clr/icons';
```
gnomeontherun: Can combine into one import.

src/ks-app/src/app/containers/iconography/icon-selection.component.ts
```diff
@@ -6,15 +6,15 @@
...
-import { ChartShapes } from '@clr/icons/shapes/chart-shapes';
-import { CommerceShapes } from '@clr/icons/shapes/commerce-shapes';
-import { CoreShapes } from '@clr/icons/shapes/core-shapes';
-import { EssentialShapes } from '@clr/icons/shapes/essential-shapes';
-import { MediaShapes } from '@clr/icons/shapes/media-shapes';
-import { SocialShapes } from '@clr/icons/shapes/social-shapes';
-import { TechnologyShapes } from '@clr/icons/shapes/technology-shapes';
-import { TextEditShapes } from '@clr/icons/shapes/text-edit-shapes';
-import { TravelShapes } from '@clr/icons/shapes/travel-shapes';
+import { ChartShapes } from '@clr/icons/shapes/chart';
```
gnomeontherun: Can we just move these imports up to `main.ts` like in the dev app and do `AllShapes` instead?

webpack.icons.config.js
```diff
@@ -10,33 +10,31 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
...
-    'shapes/social-shapes.min': './src/clr-icons/shapes/social-shapes.ts',
-    'shapes/travel-shapes': './src/clr-icons/shapes/travel-shapes.ts',
-    'shapes/travel-shapes.min': './src/clr-icons/shapes/travel-shapes.ts',
-    'shapes/technology-shapes': './src/clr-icons/shapes/technology-shapes.ts',
-    'shapes/technology-shapes.min': './src/clr-icons/shapes/technology-shapes.ts',
-    'shapes/chart-shapes': './src/clr-icons/shapes/chart-shapes.ts',
-    'shapes/chart-shapes.min': './src/clr-icons/shapes/chart-shapes.ts',
-    'shapes/text-edit-shapes': './src/clr-icons/shapes/text-edit-shapes.ts',
-    'shapes/text-edit-shapes.min': './src/clr-icons/shapes/text-edit-shapes.ts',
+    index: './src/clr-icons/api.ts',
```
gnomeontherun: Remove this line.

webpack.icons.config.js
```diff
@@ -10,33 +10,31 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
...
+    'clr-icons-api': './src/clr-icons/api.ts',
```
gnomeontherun: We don't want this line anymore.

webpack.icons.config.js
```diff
@@ -10,33 +10,31 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
...
+    'clr-icons-base': './src/clr-icons/api-autoload.ts',
+    'clr-icons': './src/clr-icons/autoload.ts',
+    'clr-icons.min': './src/clr-icons/autoload.ts',
```
gnomeontherun: Lines 15-17 lines should be replaced with

```
'clr-icons-all': './src/clr-icons/autoload-all.umd.ts',
'clr-icons-all.min': './src/clr-icons/autoload-all.umd.ts',
'clr-icons': './src/clr-icons/autoload.umd.ts',
'clr-icons.min': './src/clr-icons/autoload.umd.ts',

File: scripts/write-svg-icons.js (hunk 4 above)
Comment (jeeyun) [reply scripts/write-svg-icons.js]: We do.. because `text-edit` :)

File: src/clr-icons/element.ts [hunk 14]
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { ClarityIconsApi } from './api';
+import { ShapeTemplateObserver } from './utils/shape-template-observer';
+
+export namespace ClarityIcon {
```
gnomeontherun: I'm not happy with the namespace names I gave. Let's change this to `export namespace ElementModule`.

src/clr-icons/utils/shape-template-observer.ts
```diff
@@ -5,51 +5,43 @@
  */
 import { ShapeTemplateObservables } from '../interfaces/icon-interfaces';
 
-export const changeHandlerCallbacks: ShapeTemplateObservables = {};
-
-export class ShapeTemplateObserver {
-  private static singleInstance: ShapeTemplateObserver;
-
-  private callbacks: ShapeTemplateObservables = changeHandlerCallbacks;
+export namespace ShapeTemplateObserver {
```
gnomeontherun: I'm not happy with the namespace names I gave. Let's change this to `export namespace ShapeTemplateObserverModule`

src/clr-icons/utils/shape-template-observer.ts
```diff
@@ -5,51 +5,43 @@
...
+  export const changeHandlerCallbacks: ShapeTemplateObservables = {};
+
+  class ShapeTemplateObserverImplementation {
```
gnomeontherun: Let's rename this back to `ShapeTemplateObserver`.

src/clr-icons/api.ts
```diff
@@ -0,0 +1,129 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { IconAlias, IconShapeSources } from './interfaces/icon-interfaces';
+import { ShapeTemplateObserver } from './utils/shape-template-observer';
+import { ClarityIcon } from './element';
+
+export namespace ClarityIconsApi {
+  export class ClarityIconsApiImplementation {
```
gnomeontherun: Let's change this to
```
export namespace ApiModule {
  export class ClarityIconsApi {
```

src/clr-angular/data/datagrid/datagrid-page-size.spec.ts
```diff
@@ -37,7 +37,8 @@ export default function(): void {
         it('displays a select with a default pageSize if no input is given', function() {
           const select = context.clarityElement.querySelector('select');
           expect(select).not.toBeNull();
-          expect(select.innerText).toMatch('0');
+          expect(select.options.length).toBe(1);
```
Shijir: Are these changes related to this PR?

src/clr-icons/shapes/all.umd.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { AllShapes } from './all';
+
+export * from './all';
```
Shijir: Do we need this export?

src/clr-angular/data/datagrid/datagrid-page-size.spec.ts
jeeyun: No, but the build doesn't pass without it. It's already a part of master branch. I just added those fix for now, but when I rebased it should already be incorporated into it.

src/clr-icons/shapes/essential.umd.ts
Shijir: Yeah, right. It looks like we already created the aliases and assigned them to `EssentialShapes ` in the `essential`.ts file.

src/clr-icons/shapes/core.ts
```diff
@@ -288,72 +288,108 @@ export const ClrShapeEvent = clrIconSVG(
...
             <path d="M34.29,17.53c-3.37-6.23-9.28-10-15.82-10a16.82,16.82,0,0,0-5.24.85L14.84,10a14.78,14.78,0,0,1,3.63-.47c5.63,0,10.75,3.14,13.8,8.43a17.75,17.75,0,0,1-4.37,5.1l1.42,1.42a19.93,19.93,0,0,0,5-6l.26-.48Z" class="clr-i-solid clr-i-solid-path-2"/>
             <path d="M4.87,5.78l4.46,4.46a19.52,19.52,0,0,0-6.69,7.29L2.38,18l.26.48c3.37,6.23,9.28,10,15.82,10a16.93,16.93,0,0,0,7.37-1.69l5,5,1.75-1.5-26-26Zm8.3,8.3a6.85,6.85,0,0,0,9.55,9.55l1.6,1.6a14.91,14.91,0,0,1-5.86,1.2c-5.63,0-10.75-3.14-13.8-8.43a17.29,17.29,0,0,1,6.12-6.3Z" class="clr-i-solid clr-i-solid-path-3"/>`
 );
-export const ClrShapeStepForward2 = clrIconSVG(
+export const ClrShapeStepForward2SVG = clrIconSVG(
   `<path d="M7.08,6.52a1.68,1.68,0,0,0,0,2.4L16.51,18,7.12,27.08a1.7,1.7,0,0,0,2.36,2.44h0L21.4,18,9.48,6.47A1.69,1.69,0,0,0,7.08,6.52Z" class="clr-i-outline clr-i-outline-path-1" /><path d="M26.49,5a1.7,1.7,0,0,0-1.7,1.7V29.3a1.7,1.7,0,0,0,3.4,0V6.7A1.7,1.7,0,0,0,26.49,5Z" class="clr-i-outline clr-i-outline-path-2" />`
 );
 
+export const ClrShapeUnknownStatus = { 'unknown-status': ClrShapeUnknownStatusSVG };
+export const ClrShapeHome = { home: ClrShapeHomeSVG };
```
Shijir: I think the aliases should be assigned to their corresponding icons instead of the sets.

For example, currently we have this:
```
Object.defineProperty(CoreShapes, 'house', descriptorConfig(ClrShapeHomeSVG));
```
We users import individual shapes, their aliases would be missing. To have the default aliases, they have to import the whole sets.

So I think it should be like this instead: 
```
Object.defineProperty(ClrShapeHome, 'house', descriptorConfig(ClrShapeHomeSVG));
```
I think this would make sense to users. What do you think?

CC: @gnomeontherun

src/clr-icons/shapes/all.umd.ts
gnomeontherun: Technically no, except to make the svg builder script work.

src/clr-icons/shapes/core.ts
gnomeontherun: This is a fair question, I even talked with @jeeyun about why we even bother having them. 

To me, it does makes sense that the alias would be tied to the shape rather than the set. I'd support that change.

src/clr-icons/shapes/all.umd.ts
jeeyun: Are you asking for `all.umd.ts` specifically or `<setName>umd.ts` files in general? For each shape set, the export was necessary for our script that writes out the individual svg files.

src/clr-icons/shapes/essential.umd.ts
jeeyun: Yes. Just confirmed and will remove.

src/clr-icons/shapes/all.umd.ts
Shijir: I see. That makes sense.

src/clr-icons/shapes/core.ts
jeeyun: It's an interesting idea. Depends on if we expect the aliases to be available when importing individual shape vs importing and adding a shape set. If we want both to work, we could do it by defining it twice.. so in your example above, once for `CoreShapes` and once for `ClrShapeHome`.

Shijir: I don't think we need to define the aliases twice. Assigning the aliases to its corresponding icon object once could work. As we use the spread operator to dump individual shapes in the set object, the aliases would be dropped in just as well.

tslint.json
```diff
@@ -38,6 +37,6 @@
     "semicolon": [true, "always", "ignore-bound-class-methods"],
     "switch-default": true,
     "triple-equals": [true, "allow-null-check"],
-    "variable-name": [true, "check-format", "allow-leading-underscore", "ban-keywords"]
+    "variable-name": [true, "check-format", "allow-leading-underscore", "allow-pascal-case", "ban-keywords"]
```
youdz: Any chance we could allow this only for icons? It saved me a couple times on our Angular components when declaring constants for providers.

src/clr-icons/shapes/core.ts
```diff
@@ -342,6 +342,20 @@ export const ClrShapeEye = { eye: ClrShapeEyeSVG };
 export const ClrShapeEyeHide = { 'eye-hide': ClrShapeEyeHideSVG };
 export const ClrShapeStepForward2 = { 'step-forward-2': ClrShapeStepForward2SVG };
 
+Object.defineProperty(ClrShapeHome, 'house', descriptorConfig(ClrShapeHomeSVG));
```
Shijir: Thank you, Jeeyun! :D Have we tested it by importing an individual shape? For example, if we import an individual shape, we expect both its icon name and aliases are added? 

I am a bit concerned that had the object been already exported, the alias would be actually missing because we are assigning the alias after the export.
```
export const ClrShapeHome = { home: ClrShapeHomeSVG };
Object.defineProperty(ClrShapeHome, 'house', descriptorConfig(ClrShapeHomeSVG));
```

tslint.json
```diff
@@ -28,7 +28,6 @@
     "no-switch-case-fall-through": true,
     "no-this-assignment": [true, { "allow-destructuring": true }],
     "no-unused-expression": true,
-    "no-unused-variable": true,
```
youdz: Ok, this is a real problem for the Angular code. Without this, `tslint:fix` doesn't clean up unused imports anymore. 

I don't want to put 1.0 RC on hold for this, but I really don't agree with relaxing our TSLint rules for the `@clr/angular` just because they don't work for our icons. I believe having an extra tslint config in the `icons/` folder overrides the main config.

src/clr-icons/shapes/core.ts
gnomeontherun: The old way worked too where aliases were added after the setShapes was exported. This is because the Object.defineProperty method mutates via the reference so it trickles to the exported value as well.

tslint.json
gnomeontherun: This is actually deprecated by tslint as a rule in general, it wants you to use the tsconfig flags for this instead.

tslint.json
jeeyun: Yup makes sense and I was prob being too lazy. I'll change to allow only for icons.

src/clr-icons/shapes/core.ts
Shijir: You might be correct there. But also I think that could have been case due to this block of code. We had this in every set of icons file. We did all our alias assignments before this:
```ts
if (typeof window !== 'undefined' && window.hasOwnProperty('ClarityIcons')) {
    window.ClarityIcons.add(ChartShapes);
}
```
Now icons are separate and individually inserted so I think it's worth testing icon and its alias by importing by their individual host object.

Shijir: Yesss! Awesome 👍

Issue #2822, opened 2018-11-08
The Checkbox example on Button Group page has a display issue (v1.0)


h657070128:
## Describe the bug

![screen shot 2018-11-08 at 4 37 52 pm](https://user-images.githubusercontent.com/6769465/48186994-b0fecc00-e374-11e8-8d23-694d7105cc88.png)

## How to reproduce

On the page https://vmware.github.io/clarity/documentation/v1.0/button-group, there is a display issue for the Checkbox example.

## Versions

**Device:**

* Type: MacBook Pro
* Browser: chrome
* Version: 70.0

Jinnie:
Reproduced locally. Only happens with v1.0.
I already have the fix. Want to check how this happened and I'm submitting the PR.

PR #2825, opened 2018-11-09
[WEB] Checkboxes in wrong format (#2822)

Jinnie:
Between 0.13 and 1.0 the checkbox text was extracted to label. This requires
a parent div.btn.checkbox though, which was missing.


gnomeontherun:
You'll just need to squash your commits and fix DCO.

Issue #2535, opened 2018-08-12
Wrong colors in components


sdurnov:
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

Colours in components should match corresponding functional colours

### Actual behavior

Warning alert in light theme v.0.12 has real bg colour #FEECB5, where as in functional colours we have a value of #FEF3B5 in standard alerts colours.

Same thing for badges:
Purple badge has #9460b8 color in reality, where as in functional colours for badges there's only #9B56BB color. #9460b8 is not present in color palette at all.
Orange badge has #eb8d00 color in reality, where as in functional colours for badges there's only #FF8400 color. #eb8d00 is not present in color palette at all.

Same thing for buttons:
Primary/Info outline button and flat button has #0077b7 color in reality, should be #007CBB ? Or there should be some level of opacity applied?
Success outline button has #308300 color in reality, should be #308700

Cards borders and dividers are also have color out of palette. Same for Datagrid dividers.

### Reproduction of behavior
Compare [Functional colors v.0.12](https://vmware.github.io/clarity/documentation/v0.12/color)
with warning alert here - [Alerts v.0.12](https://vmware.github.io/clarity/documentation/v0.12/alerts) and other listed components.

### Environment details

* **Browser:** [Chrome 68.0.3440.106]

youdz:
Thanks for reporting this, I will let UX determine which color is the correct one, and we can then either fix the color palette docs or fix our CSS.

colinreedmiller:
Instructions for correct colors.

![2535 color pal issue.png](https://images.zenhubusercontent.com/5bc78fe0ba9b001064ad7062/66177b89-4f1d-4e91-b254-fbca04529a3f)

mathisscott:
Also #2784

Jinnie:
Hi all, @colinreedmiller,
I have some questions about Issue 3 (b):

1. In the textual description we see a request to change #308300 to #308700. But on the mock image I see that #318700 has to be changed to #2F8400. I can do it, but I do not understand how this relates to the original request.
2. This color (#318700) is seen at 3 places on the website - general palette, buttons, badges. Do I need to change all 3 of them?
3. This probably will be answered by answering the above question (2.), but let me ask explicitly - where in code I also fix this color? Buttons, app-level alerts, labels, badges? Or anywhere I detect the old color is used.

Thanks you!

mathisscott:
Hey @colinreedmiller 
[...]

PR #2831, opened 2018-11-12
[UI] Updated colors in alerts, badges, labels (#2535)

Jinnie:
Update of styles, dev app and ks-app


Jinnie:
Strange... datagrid tests failing, while I'm only submitting css and not in the datagrid at all.

PR #2838, opened 2018-11-13
[NG] Tree view: Lazy declarative case

youdz:
The nodes are declared in HTML on the app's side, but the app uses `*clrIfExpanded` to only instantiate them (and maybe fetch them) when needed. Because of this, we cannot propagate selection down so we simply notify the app when the user selects a node and let them figure out what it means for their model.
In other words, "lazy" is the equivalent of server-driven datagrid: if we have imperfect information on the dataset, we just do not try to be smart and become a dumb UI, the logic is on the app's side or on their server.

This PR contains very little new code for the component itself, but a lot of demo to check our various supported use cases work fine.


src/dev/src/app/tree-view/tree-view.demo.scss
```diff
@@ -24,3 +24,31 @@
   overflow-y: auto;
   overflow-x: hidden;
 }
+
+.christmas ::ng-deep {
```
hippee-lee: This looks like [Christmas creep](https://en.wikipedia.org/wiki/Christmas_creep) to me.

src/clr-angular/data/tree-view/tree.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { Component, Input } from '@angular/core';
+import { TREE_FEATURES_PROVIDER, TreeFeaturesService } from './tree-features.service';
+
+@Component({
+  selector: 'clr-tree',
+  template: `
+    <ng-content></ng-content>
+  `,
+  providers: [TREE_FEATURES_PROVIDER],
```
hippee-lee: Is it useful to leave a comment here (and / or on line#51 of `tree-node.ts`) that is explicit in explaining how the service is instantiated and injected as a singlton only at the root of the tree (if declared) or at the root TreeNode?

src/clr-angular/data/tree-view/tree-features.service.ts
```diff
@@ -8,6 +8,7 @@ import { Injectable, Optional, SkipSelf } from '@angular/core';
 @Injectable()
 export class TreeFeaturesService<T> {
   selectable = false;
+  smart = true;
```
gnomeontherun: Calling this `smart` bothered me in Datagrid. We also use `clrLazy` to say it isn't `smart`, which requires a little bit of mental calculation. Do other features depend on this flag being set or could we `s/smart/lazy` and keep the correlation between the input and the service?

src/dev/src/app/tree-view/pre-selection/pre-selection.html
```diff
@@ -0,0 +1,176 @@
...
+    <clr-tree-node [(clrSelected)]="eager.childrenOnly.indeterminate.child1">
+      Selected
+    </clr-tree-node>
+    <clr-tree-node [(clrSelected)]="eager.childrenOnly.indeterminate.child2">
+      Not selected
+    </clr-tree-node>
+  </clr-tree-node>
+</div>
+
+<h4>Eager tree, inconsistent pre-selection (Christmas Edition)</h4>
```
gnomeontherun: For future folks, can you annotate here briefly _why_ this is broken?

src/dev/src/app/tree-view/tree-view.demo.scss
```diff
@@ -24,3 +24,31 @@
...
+  & > .clr-tree-node,
+  .clr-treenode-children > .clr-tree-node {
+    &:nth-child(even) {
+      input[type='checkbox']:checked + label::before {
+        background-color: #3c8d0d;
+      }
+      input[type='checkbox']:indeterminate + label {
+        &::before,
+        &::after {
+          border-color: #3c8d0d;
```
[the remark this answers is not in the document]
youdz: These are demo styles added just for laughs. But you've got a point, I'll check the contrast.

src/clr-angular/data/tree-view/tree-features.service.ts
youdz: As discussed, changing to `lazy` means every time we test it in our logic, it would be:
```typescript
if (!lazy) {
  // Propagate selection
}
```
```typescript
if (!lazy) {
  // Build children nodes
}
```
```typescript
if (!lazy) {
  // Add some enhanced feature on our side
}
```

We agreed this would force the "flip" gymnastic every time we right code, which is not desirable. To keep the property closer to the corresponding input, we agreed to rename it `eager`, which is more understandable.

youdz:
I can't believe I took your troll seriously. 🤦‍♂️ 

Merging in and preparing the next PR.

Issue #1752, opened 2017-12-06
[Datagrid] Manually updating clrDgPageSize no longer results in clrDgRefresh being triggered


jbgarr:
```
[X] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Manually updating the clrDgPageSize should trigger a clrDgRefresh event to be fired on the table (this was how it worked in Clarity v0.9.x). 

### Actual behavior
Manually updating clrDgPageSize no longer triggers a clrDgRefresh event.

### Reproduction of behavior
Here is a working version (v0.9.x): https://plnkr.co/edit/7PBBUqnPSopd21NCyfxU?p=preview
And here is the same code running in v0.10.x where the event is no longer fired: https://plnkr.co/edit/cl9WfX76lYlbVTh6nm2u?p=preview

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** MacOS Sierra

* **Browser:** Chrome 62.0.3202.94 (Official Build) (64-bit)

hippee-lee:
[StateProvider.change](https://github.com/vmware/clarity/blob/master/src/clr-angular/data/datagrid/providers/state.provider.ts#L31) is not broadcasting events to `Datagrid` for some reason. I traced it into the settor for clrDgPageSize and stopped there for now.

dragon503:
@jbgarr Have you fixed it? 
I met the same issue and use "clrDgStateChange" as mentioned in #1099 but do not work.

stefnoten:
We have the same problem. The problem seems to be that the Page service does not trigger the StateDebouncer inside the [size setter](https://github.com/vmware/clarity/blob/a2bc32dc49bc63a968983a6a00f94d74282f0b1d/src/clr-angular/data/datagrid/providers/page.ts#L22). It only does this for changes of the [current page](https://github.com/vmware/clarity/blob/a2bc32dc49bc63a968983a6a00f94d74282f0b1d/src/clr-angular/data/datagrid/providers/page.ts#L98).

mathisscott:
@m-tomczyk 
This is on our radar but we do not currently have it assigned to anyone to be fixed. If it is a potential blocker for your team, we are certainly open to contributions.

Jinnie:
Btw, plunkr reproduction no longer working.
Attaching the stackblitz one I did while researching the issue, if needed for future reference:
https://stackblitz.com/edit/clarity-issue-1752

Jinnie:
[...]

PR #2841, opened 2018-11-14
[NG] Page size change was not triggering a state change (#1752)

Jinnie:
During 0.9 -> 0.10 refactoring, the refresh has no longer been managed locally in the datagrid component, but delegated to the state provider service. Thus, we're no longer listening to page.change events directly.
StateDebouncer calls added to the _size_ setter, similar to the _current_ setter.
Tests are checking initial size change (from not set), then transition from one size to another, finally a size reset.


mathisscott:
@Jinnie 
This is ready to go! Can you backport to 0.13?

PR #2851, opened 2018-11-15
[NG] Tree view: Asynchronous cases

youdz:
This commit adds the integration of our loading directive to the tree, to display a nice spinner instead of a caret when waiting for the children to be fetched. Works for both the declarative and recursive cases.

No gemini tests because spinners...

### DIFF
--- a/src/clr-angular/data/tree-view/tree-node.html
+++ b/src/clr-angular/data/tree-view/tree-node.html
@@ -6,7 +6,7 @@
 
 <div class="clr-tree-node-content-container">
   <button
-    *ngIf="isExpandable()"
+    *ngIf="isExpandable() && !_model.loading && !expandService.loading"
     type="button"
     class="clr-treenode-caret"
     (click)="expandService.toggle()"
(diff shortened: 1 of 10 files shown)


src/clr-angular/data/tree-view/tree-node.html
```diff
@@ -6,7 +6,7 @@
 
 <div class="clr-tree-node-content-container">
   <button
-    *ngIf="isExpandable()"
+    *ngIf="isExpandable() && !_model.loading && !expandService.loading"
```
hippee-lee: To mirror how the `clr-tree-spinner-container` is displayed, should this be `isExpandable() && (!_model.loading || !expandService.loading)`?

youdz: It would have to be `isExpandable() && !(_model.loading || expandService.loading)`. If you feel it's more readable, I can do that, but I tend to avoid mixing `&&` and `||` in boolean conditions for clarity.

hippee-lee: Right I missed the `!`.
I don't feel strong about how it is now. Thats why I already approved it.

src/dev/src/app/tree-view/utils/async-infinite-tree.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import { ClrSelectedState } from '@clr/angular';
+
+import { InfiniteTree } from './infinite-tree';
+import { Observable, timer } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+export class AsyncInfiniteTree {
+  constructor(width: number, latency = 0) {
```
gnomeontherun: nit: Can we put a 100ms delay or something very small here just so it feels a little bit async?

youdz: Sure, I'll make the default delay 100ms. 👍 

Note that the demos actually set the latency to 500ms, so the spinners are clearly visible.

Issue #2849, opened 2018-11-15
[WEB] Login example code missing select close tag (v1.0)


sivaah:
## Describe the bug

The login example code provided is missing the close tag for select which results in throwing error in the browser when using the example code in my app

## How to reproduce

Open the page https://vmware.github.io/clarity/documentation/v1.0/login and use the code example provided

`      <clr-select-container>
                <select clrSelect name="type" [(ngModel)]="form.type">
                <option value="local">Local Users</option>
                <option value="admin">Administrator</option>
            </clr-select-container>
`

![image](https://user-images.githubusercontent.com/6573919/48556168-2d5f5500-e909-11e8-9026-bf961d539d24.png)

## Versions

**App**

* Angular: 7.0.0
* Node: 10.10.0
* Clarity: 1.0

**Device:**

* Type: MacBook
* OS: iOS
* Browser chrome
* Version 70.0

sivaah:
The issue has been fixed by commit 5187638e89b25d4ecc9fb025af1597fac422f57f

PR #2853, opened 2018-11-16
[WEB] Add missing select close tag in login example code

sivaah:
The login example code provided is missing the close tag for select which results in throwing error in the browser when using the example code in my app

More info can be found in the issue #2849


sivaah:
Just found this was fixed by commit 5187638e89b25d4ecc9fb025af1597fac422f57f

PR #2858, opened 2018-11-16
A11y fixes - add unique names to items

Jinnie:
Added unique names to [SHOW] buttons and sidebar links.

### DIFF
--- a/src/ks-app/src/app/containers/progress/progress-bars.component.html
+++ b/src/ks-app/src/app/containers/progress/progress-bars.component.html
@@ -24,14 +24,14 @@
         <h4>In Sidenav</h4>
         <ul class="sidenav list-unstyled col-xs-2">
             <li class="nav-item">
-                <a class="nav-link" href="javascript://">Link</a>
+                <a class="nav-link" href="javascript://">Sidenav item 1 (w/ progress)</a>
                 <div class="progress loop"><progress></progress></div>
             </li>
             <li class="nav-item">
-                <a class="nav-link" href="javascript://">Link</a>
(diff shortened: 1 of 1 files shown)


src/ks-app/src/app/containers/progress/progress-bars.component.html
```diff
@@ -18,20 +18,20 @@ <h4>{{e.title}}</h4>
...
                 (click)="e.start()"
-                [ngClass]="e.demoId">Show</button></p>
+                [ngClass]="e.demoId">Show {{e.title}}</button></p>
     </div>
     <div class="col-xs-12 col-lg-4">
         <h4>In Sidenav</h4>
         <ul class="sidenav list-unstyled col-xs-2">
             <li class="nav-item">
-                <a class="nav-link" href="javascript://">Link</a>
+                <a class="nav-link" href="javascript://">Sidenav item 1 (w/ progress)</a>
```
gnomeontherun: I think based on my reading of the issue, this is still too generic. We'd be better to put something like `Products` for these links.

Jinnie: These are a series of example links. "Product" may look more like a real world example, but I'm not sure it will fit the context of the mock, where we have several similar links - Sidenav item ${i} - and one which should be different and should stand out - Sidenav item 1 (w/ progress).
When I was choosing the name, I researched other similar items, where we have repetitive entities, and I saw this pattern already used in the ks-app.
See the radio buttons, for example:
https://github.com/vmware/clarity/blob/master/src/ks-app/src/app/containers/forms/radios.component.html

PR #2862, opened 2018-11-19
Make vertical-nav hover color different than active/selected

Jinnie:
The first commit adds the visual tests that fix the current state, to avoid regressions. They are important, as I need to redefine color definitions used in different cases.
The second commit makes the actual hover/active color changes for both light and dark themes.


Jinnie:
I did now squash these commits intentionally, as the tests are meaningful without the subsequent color fix, and may be separately merged to other versions if needed.

PR #2866, opened 2018-11-20
[web] updating icons dox for 1.0

jeeyun:
Just getting it started. We might need additional updates if we change some things before 1.0.

Also need to update the external plunker examples for 1.0.


gnomeontherun:
I'm pulling this into the website manually since a few more tweaks are needed that weren't finalized when this was made.

PR #2872, opened 2018-11-21
[WEB] Documentation for the new Tree

youdz:
Trying out a small navigation list and "Back to top" links. Not promising it's great,
but we still have no idea how to handle these types on long docs.


youdz:
Closing, @gnomeontherun manually added this to the website as part of the release.

PR #2873, opened 2018-11-21
Icon tweaks for API simplicity

gnomeontherun:
remove need for `init()` method, and allow `add()` to accept as many arguments as you like

The `add()` method will call `init()` internally and setup icons if they weren't already.


README.md
```diff
@@ -136,6 +134,7 @@ If you already have an Angular application, you can follow the installation step
     	map: {
     	   ...
     	   '@clr/angular': 'node_modules/@clr/angular/bundles/clr-angular.umd.js',
+    	   '@clr/icons': 'node_modules/@clr/icons/bundles/clr-icons.umd.js',
```
Shijir: We have `clr-icons-all.min.js` instead of `clr-icons.umd.js`, right?

webpack.icons.config.js
```diff
@@ -10,12 +10,10 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 module.exports = {
   mode: 'development',
   entry: {
-    'clr-icons-all': './src/clr-icons/autoload-all.umd.ts',
-    'clr-icons-all.min': './src/clr-icons/autoload-all.umd.ts',
-    'clr-icons': './src/clr-icons/api-autoload.umd.ts',
-    'clr-icons.min': './src/clr-icons/api-autoload.umd.ts',
-    'shapes/all': './src/clr-icons/shapes/all.umd.ts',
-    'shapes/all.min': './src/clr-icons/shapes/all.umd.ts',
+    'clr-icons': './src/clr-icons/api.umd.ts',
```
Shijir: Do we need these non-minified bundles `clr-icons` and `clr-icons-all`?

src/clr-icons/all.umd.ts
```diff
@@ -4,9 +4,12 @@
...
 import { ClrAllSet } from './shapes/all';
+import { ApiModule } from './api';
 
 /* tslint:disable:variable-name */
-const ClarityIcons = ApiModule.instance;
-ClarityIcons.init(ClrAllSet);
+ApiModule.instance.init();
+
+if (window && window.hasOwnProperty('ClarityIcons')) {
+  window.ClarityIcons.add(ClrAllSet);
```
youdz: I'm not sure I understand why you still need the `init()`. Can't you just do:
``
`typescript
ApiModule.instance.add(ClrAllSet);
```

File: README.md (hunk 1 above)
Comment (gnomeontherun) [reply README.md]: This is systemjs, so we ship a UMD bundle separately from the other ones that are specific to what systemjs would use here.

File: src/clr-icons/api.ts [hunk 4]
```diff
@@ -66,22 +66,30 @@ export namespace ApiModule {
...
+      icons.forEach(iconInstance => {
+        if (typeof iconInstance !== 'object') {
+          throw new Error(
+            `The argument(s) must be an object literal passed in the following pattern: { "shape-name": "shape-template" }`
+          );
         }
-      }
+        for (const shapeName in iconInstance) {
+          if (iconInstance.hasOwnProperty(shapeName)) {
+            console.log(shapeName, iconInstance[shapeName]);
```
youdz: Forgotten log.

src/clr-icons/api.ts
```diff
@@ -66,22 +66,30 @@ export namespace ApiModule {
...
-    add(icons?: IconShapeSources): void {
-      if (typeof icons !== 'object') {
-        throw new Error(`The argument must be an object literal passed in the following pattern: 
-                  { "shape-name": "shape-template" }`);
+    add(...icons: IconShapeSources[]): void {
+      this.init();
+
+      if (!icons.length) {
+        throw new Error(
+          `The argument(s) must be an object literal passed in the following pattern: { "shape-name": "shape-template" }`
```
youdz: This error message feels off for this case.

Honestly, I think I would just not put a guard on `icons.length` at all, `ClarityIcons.add()` with no arguments would simply not add any icons. It's silly, but I don't see a problem with it.

src/clr-icons/api.ts
Shijir: left over console.log

src/clr-icons/api.ts
```diff
@@ -66,22 +66,30 @@ export namespace ApiModule {
...
     }
 
-    add(icons?: IconShapeSources): void {
-      if (typeof icons !== 'object') {
-        throw new Error(`The argument must be an object literal passed in the following pattern: 
-                  { "shape-name": "shape-template" }`);
+    add(...icons: IconShapeSources[]): void {
+      this.init();
+
+      if (!icons.length) {
```
Shijir: I don't see the value of having this if block here?

webpack.icons.config.js
gnomeontherun: It can be useful for someone trying to read the source code, and it is pretty common to ship both a min and non-min version. If we didn't ship these as min, then we shouldn't ship any of the shapes either.

src/clr-icons/all.umd.ts
```diff
@@ -9,7 +9,4 @@ import { ApiModule } from './api';
 
 /* tslint:disable:variable-name */
 ApiModule.instance.init();
```
youdz: What I meant is, you can remove this `ApiModule.instance.init();` now, since it's included in the next line.
So `init()` can become truly private if you want it to.

src/clr-icons/all.umd.ts
```diff
@@ -4,9 +4,9 @@
...
  */
 
-import { ApiModule } from './api';
 import { ClrAllSet } from './shapes/all';
+import { ApiModule } from './api';
 
 /* tslint:disable:variable-name */
-const ClarityIcons = ApiModule.instance;
-ClarityIcons.init(ClrAllSet);
+ApiModule.instance.init();
```
Shijir: Nit: Since `ApiModule.instance.add` calls `init()`, we don't need to initialize, I think.

PR #2882, opened 2018-11-26
[icons] reverting icon changes 

gnomeontherun:
These changes supported proper treeshaking and builds, but it also breaks SSR. We will need to investigate other options and dig deeper, but don't want to hold up 1.0 any further.


src/clr-icons/clr-icons-element.ts
```diff
@@ -0,0 +1,201 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { ClarityIconsApi } from './clr-icons-api';
+import { ShapeTemplateObserver } from './utils/shape-template-observer';
+
+/* CLR-ICON CUSTOM ELEMENT */
+
+let clrIconId = 0;
+const offScreenSpan = document.createElement('span');
```
Shijir: Weird considering that this `document` has no guard but it was working?

gnomeontherun: Strange yes, but the contents of this file do no make it into the SSR bundles.

PR #2883, opened 2018-11-26
[core] have NPM use exact versions and bump some dependencies

gnomeontherun:
This bumps `npm-run-all` which removes the `events-stream` package (which had an older version with security issues but not the version we used), and locks versions to specific numbers. 

Unfortunately, `ng update` sometimes sets a range in the package.json file, so this will only save exact versions when using NPM to update or install directly.


gnomeontherun:
No I decided not to spend that time and get this done before trying to look into it further.

PR #2884, opened 2018-11-26
[NG] Multi global alerts can handle dynamic removal

### LINKED ISSUE #2878: Programmatically closing a clr-alert wrapped in clr-alerts causes all alerts to disappear
## Describe the bug

Programmatically removing the second alert from a list of two alerts managed by `<clr-alerts>` causes both alerts to be invisible.

## How to reproduce

https://clarity-rrbfud.stackblitz.io/

1.  Go to the second alert
2.  Click on 'This is alert #2. Remove me'
3.  See error

## Expected behavior

After clicking the button I would expect the first alert to become visible. Instead, no alerts are visible anymore.

A workaround is to inject 'clr-alerts' as a ViewChild and call `this.clrAlerts.multiAlertService.previous();` before splicing the alerts array. This causes the bug to not manifest itself.

Please note that the `alerts` array itself still contains the remaining alert.

## Versions

This happens in StackBlitz which contains clr 0.11.1, but on my local machine I have the latest version (1.0.0-beta-2) where this also occurs.

youdz:
We used a QueryList to track currently displayed alerts, but for some reason never subscribed to it.
I just went and made sure at least one of the alerts was always displayed, any advanced behavior
trying to keep the currently alert displayed would require hacking into `ngForTrackBy` which
isn't possible in this situation.

Fixes #2878.


youdz:
@Jinnie: My bad, you're absolutely right. I somehow missed it, and that would re-introduce some variant of #1857. Updating the PR, thanks.

PS: If you see an issue please check "Request changes", it's better to remove it if it turns out not to be a real issue than miss it and have to rush a second fix in. You're part of the team now, you get to block PRs. 😂

Jinnie:
@youdz Wasn't sure if it's important. -1 is a normal "not found" result. My actual thought was that a test for this case would make the intention clear.

src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
```diff
@@ -50,6 +50,11 @@ export class MultiAlertService {
 
   manage(alerts: QueryList<ClrAlert>) {
     this.allAlerts = alerts;
+    this.allAlerts.changes.subscribe(() => {
```
gnomeontherun: Do you need to unsubscribe to an observable that is on the same object? I suspect so?

src/clr-angular/emphasis/alert/providers/multi-alert.service.ts
```diff
@@ -4,14 +4,15 @@
...
-import { Observable } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 import { Subject } from 'rxjs';
 
 import { ClrAlert } from '../alert';
 
 @Injectable()
 export class MultiAlertService {
   private allAlerts: QueryList<ClrAlert> = new QueryList<ClrAlert>();
+  private _subscription: Subscription;
```
Shijir: ### A violation of our coding guidelines. ☠️ 

😂 Was there a specific reason you need to prefix these properties with underscores

youdz: 😭

Issue #2852, opened 2018-11-15
Modal close default


colinreedmiller:
## Describe the bug

_Default behavior for user closing a modal is specified as requiring explicit click - clicking outside to close is configurable but should not be the default._

## Expected behavior

_Click outside the modal should not cause it to close as default._

Documentation reads: "Clarity Modals are not dismissed when clicking on the background overlay. This prevents losing information or data. An option to override this is available."

[https://vmware.github.io/clarity/documentation/v0.11/modals](https://vmware.github.io/clarity/documentation/v0.11/modals)

Shijir:
Closed by #2885

Shijir:
Sorry closed wrong issue 😛

Jinnie:
Now we can close it officially :)

PR #2889, opened 2018-11-28
[NG] Modal close on backdrop click default changed (#2852)

Jinnie:
Click on modal backdrop now does not close the modal by default.
Changes of the website documentation coming in a separate PR.


src/clr-angular/modal/modal.spec.ts
```diff
@@ -206,6 +223,21 @@ describe('Modal', () => {
...
+    'should not be closed on backdrop click by default',
+    fakeAsync(() => {
+      const defaultsFixture = TestBed.createComponent(TestDefaultsComponent);
+      defaultsFixture.detectChanges();
+      compiled = defaultsFixture.nativeElement;
+
+      const backdrop: HTMLElement = compiled.querySelector('.modal-backdrop');
+
+      backdrop.click();
+      flushAndExpectOpen(defaultsFixture, true);
```
Shijir: I think we should trigger detect changes before this expect.

src/clr-angular/modal/modal.spec.ts
```diff
@@ -206,6 +223,21 @@ describe('Modal', () => {
...
+      defaultsFixture.destroy();
```
Shijir: Could this go run in after every unit test? Or it should run in specific tests?

src/clr-angular/modal/modal.spec.ts
youdz: That's what the `flushAndExpectOpen` function is for. It detects changes, waits for animations to finish, and _then_ checks the result.

Jinnie: It is the first thing that happens inside `flushAndExpectOpen` helper function.
It wraps several lifecycle calls, `detectChanges` being one of them. The `tick()` for the fakeAsync is also encapsulated in the `flushAndExpectOpen`.

src/clr-angular/modal/modal.spec.ts
Jinnie: It is only needed here, because we're using a different test component, where the template is as stripped of inputs, as possible, so we can test defaults. That's why we have the `defaultsFixture` only here. All the other tests are using the normal `fixture` with the extended template, which has all needed inputs.

src/clr-angular/modal/modal.spec.ts
Shijir: ah gotcha.

PR #2890, opened 2018-11-28
[WEB] Fix docs about default modal backdrop click (#2852)

Jinnie:
Now the default is to ignore backdrop clicks


Jinnie:
This illustrates changes introduced with PR #2889, and would better be merged only after the latter is merged.

PR #2903, opened 2018-12-02
[WEB] Fix links to "forms"

Jinnie:
Was leading to 404


Jinnie:
I have absolutely no clue why it requires three times "../"
The expected single "../forms" did not work.
Saw this was working in a navigation inside tables to the datagrid.
Any comment will be appreciated.

Issue #2915, opened 2018-12-06
Data grid header overlaps other data grid components.


spitfjre:
## Describe the bug

The data-grid header z-index (it is 500) is too high and overlaps other data-grid components.

![screenshot](https://i.imgur.com/LXS1eMX.png)

## Versions

**App**

* Angular: 7.1.0
* Node: 11.3.0
* Clarity: 1.0.0

Jinnie:
Thanks for the feedback.
Actually, the column-switch popup is missed in our z-index accounting. The datagrid header z-index is fine, as we are aligning other absolute positioned items against this 500 value.
I'll commit a fix soon.

Jinnie:
Fixed

PR #2916, opened 2018-12-06
[NG] Data grid header overlaps column switch popup (#2915)

Jinnie:
z-index added for column switch


Jinnie:
Thanks to @spitfjre for catching it. Thanks for the ship-it
I'm still 1 review short of merging :) It's an easy one, someone please, take a look

Issue #2912, opened 2018-12-05
[KS-APP] Radios in the Radios page not rendering


Jinnie:
## Describe the bug

Most of the radio buttons expected to be seen in the _Angular component_ section are not displayed

## How to reproduce

1.  Go to http://clarity-ks.surge.sh/radios
2.  Scroll down to _Angular component_
3.  It looks like this:
<img width="385" alt="screen shot 2018-12-05 at 1 47 23 pm" src="https://user-images.githubusercontent.com/1798828/49511861-fd1a2f00-f894-11e8-9ad8-46a323a75422.png">

## Expected behavior

Looking at the code, radios should be seen in 4 groups:
- Full user control
- User can't be bothered to give an ID
- Radio with no label
- Radio with no label, but the user still wants a specific ID

## Versions

**App**

* Angular: 7
* Node: v10.13.0
* Clarity: 1.0

**Device:**

* Type: Macbook
* OS: OSX
* Browser Chrome
* Version 70.0.3538.102

Jinnie:
Actually, the 2 visible radios are for the out-of-control (user control) radios - cases 3) and 4) of the Expected behavior section.
The two full-control examples were lacking a clr-radio-wrapper and they were not getting projected inside the clr-radio-component.
I have a fix for this, with h4 header descriptions for each of them, as we have examples without radio label and they are just hanging in the void now. PR coming soon.

Jinnie:
That's how I suggest to rework it:
(Compare with the screenshot from the issue description)
<img width="384" alt="screen shot 2018-12-05 at 3 33 52 pm" src="https://user-images.githubusercontent.com/1798828/49516860-483b3e80-f8a3-11e8-8ab1-83490a132a6e.png">

PR #2917, opened 2018-12-06
[KS-APP] Radios not rendering (#2912)

Jinnie:
Missing clr-radio-wrapper. Added headers for better readability.

### DIFF
--- a/src/ks-app/src/app/containers/forms/radios.component.html
+++ b/src/ks-app/src/app/containers/forms/radios.component.html
@@ -47,19 +47,27 @@
 <h2>Angular component</h2>
 
 <!-- Full user control -->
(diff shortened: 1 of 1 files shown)


Jinnie:
@hippee-lee, I am personally used to the 2 spaces, too. But I'm doing my best here to keep indent of my changes same as in the file I am editing. And from what I see, it's 4 spaces almost everywhere for html templates.

youdz:
Prettier isn't configured to format HTML in our project, for some reason. It targets only js, json, md, ts and scss. I'm fine if someone adds HTML to this in a dedicated PR, not sure why we didn't do it in the first place. Maybe Prettier doesn't handle HTML? I don't remember.

PR #2918, opened 2018-12-06
[NG] fix signpost close issue (#2819)

### LINKED ISSUE #2819: Signpost close button doesn't work if the signpost is inside of a label
Signpost close button doesn't work if the signpost is inside of a label.

https://clarity-rkqmjc.stackblitz.io

In the following structure the X icon doesn't close the signpost:
screen shot 2018-11-07 at 5 19 45 pm

 
```
<label>Some label
   <clr-signpost>
      <clr-signpost-content *clrIfOpen>
           <span>some content</span>
      </clr-signpost-content>
   </clr-signpost>
</label>
```

The X icon in the signpost content should hide the signpost even if it is inside of a label.

Angular: 6
Clarity: 1.0, 0.13, 0.12, 0.11

coryrylan:
Chrome & Safari signposts do not close properly if nested in a label. Default behavior in Chrome and Safari is to trigger a click to the first button in a label element. This behavior causes the signpost to open again after closing.

closes #2819


src/clr-angular/popover/signpost/signpost-content.ts
```diff
@@ -67,10 +67,12 @@ export class ClrSignpostContent extends AbstractPopover {
   /**********
    *
    * @description
-   * Close function that uses the signpost instance to toggle the state of the content popover.
+   * Close function that uses the IfOpenService instance to toggle the state of the content popover.
    *
    */
-  close() {
+  close(event: MouseEvent) {
+    event.preventDefault();
```
hippee-lee: Depending on how the consumer is using the signpost I don't think we can use `preventDefault` here. I remember it has always been an issue in the past when proposed as a solution.

src/clr-angular/popover/signpost/signpost-content.ts
```diff
@@ -67,10 +67,12 @@ export class ClrSignpostContent extends AbstractPopover {
   /**********
    *
    * @description
-   * Close function that uses the signpost instance to toggle the state of the content popover.
+   * Close function that uses the IfOpenService instance to toggle the state of the content popover.
    *
    */
-  close() {
+  close(event: MouseEvent) {
```
Shijir: You might need `TouchEvent` as well?

src/clr-angular/popover/signpost/signpost-content.ts
Shijir: As a library/framework, we try to avoid changing the default behavior of the browsers. The reason is that our users expect those default behaviors and their app functionalities could very well depend on them. So we try to avoid disabling or changing them. Unfortunately, both `event. preventDefault` and `event.stopPropagation` changes the default behaviors. If necessary, we could use only one of them, maybe? Based on that, I see that we might not need `event.preventDefault` here becuase the error was due to event bubbling.

coryrylan: It looks like `event.preventDefault` is needed to fix the issue, removing `event.stopPropagation` works but I'll need to adjust the tests a bit. I'm going to keep looking to see if there is any other ways I can get it to work without using either.

Jinnie: I am not really sure this is actually a Clarity issue. Any button in a label will be exchanging events:
https://stackblitz.com/edit/button-in-a-label-issue
On the other hand, we definitely do not want this behavior, nor it is good to blame it on the user.
Preventing default may be justified here. Or we might want to reconsider if we need our signpost to be implemented as button in the 1st place.

youdz:
I haven't read the code yet, but based on the comment I can confirm the fix isn't to make the DOM from the original issue work. Buttons inside of labels mean interactive elements inside of interactive elements, which at best is problematic for accessibility, and at worst is just invalid HTML. Here is a link to MDN's warning about buttons in labels: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Interactive_content

This means the fix is to make sure we support putting the signpost outside of the label and display it correctly. For instance, for text inputs the DOM would be:
```html
<clr-input-container>
  <label>Some label</label>
  <clr-signpost>
    <clr-signpost-content *clrIfOpen>
      <span>some content</span>
    </clr-signpost-content>
  </clr-signpost>
  <input clrInput type="text" />
</clr-input-container>
```
This is a pretty complex issue that requires non-trivial changes on probably all of our form components. Not sure it's the best one for early ramp-up. 😁

src/clr-angular/popover/signpost/signpost-content.ts
```diff
@@ -67,10 +67,12 @@ export class ClrSignpostContent extends AbstractPopover {
...
+    event.preventDefault();
+    event.stopPropagation();
```
youdz: `event.stopPropagation()` is a hard no in a library. The reason is that it completely breaks any listener that is added somewhere higher in the tree, whether it's been added by us or by the application.

Here are a couple examples:
- Consumers listening to clicks on the container of `event.stopPropagation()` element to trigger some action will never get the event. Typically this is done for performance reasons, adding one listener to a list of 1000 items instead of adding 1000 listeners.
- Elements that appear on top of the page, and close when you click anywhere. They will not close if you click on the element that has `event.stopPropagation()`.
- Global action trackers for analytics or business insights. They work by adding global `click`/`mousemove` listeners on the whole document, and `event.stopPropagation()` creates "dark zones" without anyone knowing.

Shijir:
Are we certain that signpost should be triggered like native HTML form elements if clicked on its containing label? I don't think signposts/tooltips/modals should behave like that if they are inside a label. In that case, we could check the event's origin with `event.target` and only trigger the signpost when the event origin is the signpost trigger element itself or its child element, and shouldn't trigger if it's from the containing `<label>` element.

CC: @hippee-lee @youdz

Jinnie:
Why do we need to implement the signpost with a button inside? Semantics? It's a general question. I also saw we implemented dg column separator with a button, for example?

mathisscott:
@Jinnie 
It's for accessibility primarily. Screenreaders and keyboard navigation work as expected when HTML is used correctly. When you start making clickable divs inside of other items that browsers register as clickable, you really open a Pandora's Box of javascript event handling that gets harder and harder to close.

From experience, we found it's almost always a better strategy to allow HTML to do what it is intended to do.

Jinnie:
The really weird thing is that clicking on the popup content triggers a click on the information icon.
I've updated my demo a bit: https://stackblitz.com/edit/button-in-a-label-issue
And with this I'm stopping the spam :)

PR #2919, opened 2018-12-06
[NG] rehauled old column resizing files and added resizing component …

Shijir:
…and services

### **Part 1:**

- Removed `DatagridColumnResizer` directive and its corresponding test file. The reason for the removal was that it was handling both changing resize model and view. Instead, now we have `column-resizer.service.ts` that handles the model changes and `datagrid-column-separator.ts` component that handles the view changes. 
- Having this dedicated component ,`datagrid-column-separator.ts` for column separators make `datagrid-column.ts` a bit cleaner.
- Tests are coming in the next PR (Part 2).

No visual changes, but just in case you would like to check out the changes in a live demo, here it is: http://column-resize-refactor.surge.sh/datagrids


src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { ColumnResizerService } from './providers/column-resizer.service';
+import { TableSizeService } from './providers/table-size.service';
+
+let nbCount: number = 0;
+
+@Component({
+  selector: 'clr-dg-column-separator',
+  template: `
+    <button class="datagrid-column-handle" tabindex="-1" type="button"
```
Shijir: @Jinnie I think your question regarding whether this should be a button or not was a good inquiry. I don't find strong compelling reasons for it to be a button element. I am still contemplating on the question. Hence, I am open to ideas and suggestions to answer the question.

Jinnie: I agree that things which are clickable, and lead to some actions, should have the corresponding semantics. But in this specific case, the column separator is only for updating the visual display of some data. I even think we should deliberately try to keep it out of the screenreaders' radars.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -468,20 +468,23 @@
...
           width: calc(0.5rem + #{$clr-rem-1px});
-          right: -0.5rem;
+          right: -0.25rem;
           top: -0.25rem;
           cursor: col-resize;
           height: calc(100% + 0.5rem - #{$clr-default-borderwidth});
           z-index: map-get($clr-layers, datagrid-header);
         }
-        .datagrid-column-handle-tracker {
+        .datagrid-column-resize-tracker {
```
gnomeontherun: Isn't this a breaking change?

src/clr-angular/data/datagrid/datagrid-column-separator.ts
gnomeontherun: I agree with @Jinnie, it can be just a `div`. There is simply no benefit for a screen reader to resize columns, the content is not visible. In other cases, I did see `aria-grabbed` for draggable operations.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -0,0 +1,70 @@
...
+    '[class.datagrid-column-separator]': 'true',
+  },
+})
+export class ClrDatagridColumnSeparator {
+  // Every column draggable separator should have its own unique ID
+  // in order to not conflict with other draggables/droppables.
+  public columnSeparatorId: string;
+
+  constructor(
+    public columnResizerService: ColumnResizerService,
```
gnomeontherun: Make this private and call the resizer service inside of your show/move/hide tracker methods instead. Don't want to expose this internal service if we don't need to.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
youdz: This is an element that's in our template. We tell people not to mess with the styles of internal datagrid stuff, so I don't consider this a breaking change.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
youdz: What about low-mobility users that use a regular browser but keyboard navigation? They don't get to resize the columns and they can't read the actual cells content?

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -0,0 +1,70 @@
...
+
+let nbCount: number = 0;
+
+@Component({
+  selector: 'clr-dg-column-separator',
+  template: `
+    <button class="datagrid-column-handle" tabindex="-1" type="button"
+      clrDraggable 
+      [clrGroup]="columnSeparatorId" 
+      (clrDragStart)="columnResizerService.startResize(); showTracker(resizeTrackerEl)"
```
youdz: `showTracker()` should handle the call to the service. Having more than one operation inline in an HTML output is generally a red flag.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
gnomeontherun: Shouldn't columns always be displaying their full content by default? If that isn't the case, then it would be important to support keyboard column resizing (though the behaviors are something to think through more). We don't have keyboard support built in for this yet, so this seems beyond scope for the current capability.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -68,9 +67,9 @@ export class DatagridHeaderRenderer implements OnDestroy {
 
   public setWidth(width: number) {
     if (this.strictWidth) {
-      if (this.columnResizer.columnResizeBy) {
+      if (this.columnResizerService.resizedBy) {
         this.renderer.setStyle(this.el.nativeElement, 'width', width + 'px');
-        this.columnResizer.columnResizeBy = 0;
+        this.columnResizerService.resizedBy = 0;
```
youdz: Can you move this to the `startResize()` method of the service? That's where you compute `widthBeforeResize`, so it makes sense to reset both of them at the same time.

src/clr-angular/data/datagrid/providers/column-resizer.service.ts
```diff
@@ -20,7 +20,11 @@ export class ColumnResizerService {
 
   private widthBeforeResize: number;
 
-  public resizedBy: number = 0;
+  private _resizedBy: number = 0;
```
youdz: Nice touch, making it read-only. 👍

PR #2928, opened 2018-12-10
[NG] fix button animation issue

### LINKED ISSUE #2700: ClrLoadingState.LOADING hidden for .btn-sm .btn-link in .card .card-footer within .card-columns
## Describe the bug

_A clear and concise description of what the bug is, and the conditions when it occurs._

My component has a save button that I'd like to have showing the loading state when making an http call.  When the button enters the loading state, the text disappears, but a spinner doesn't show up.  The button is disabled but does not contain any content during the loading.

I struggled to reproduce this in stackblitz until I started using the card-columns.  I noticed that the button does enter the loading state and does have a spinning spinner, but the width is being set to 2px via some javascript that I can't identify.

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

https://stackblitz.com/edit/clarity-dark-theme-v013-kdd7ob

1.  Click on 'SAVE'
2.  Observe hidden loading indicator

## Expected behavior

The button should maintain enough width to be visible when entering the loading state when in the `.card-footer` of a `.card` in a `.card-columns`.

## Versions

**App**

* Angular: 6.2.1
* Node: 10.9.0
* Clarity: 0.13

**Device:**

* OS: Windows 10
* Browser: Chrome
* Version: 69

## Additional notes

I noticed, too, that if you don't allow the card columns to display in more than one column, the button works fine.  Its only when the card columns are more than 1.

coryrylan:
In Chromium browsers button loading animation would not be visible with CSS multi-columns. Here is a minimal CSS/HTML only reproduction of the bug.

https://stackblitz.com/edit/js-bbvdrm?file=style.scss

It looks to be some kind of rendering bug. Changing certain properties cause the loader to render properly. Using `transform: translatez(0);` seems to fix the issue while not altering any of the existing rendered styles. This might also help the animation render performance as translatez will create a new layer for the browser to render with. We would want to use this dynamically as it's not ideal for every button to create its own layer.

closes #2700

### DIFF
--- a/src/clr-angular/button/button-loading/loading-button.ts
+++ b/src/clr-angular/button/button-loading/loading-button.ts
@@ -73,12 +73,14 @@ export class ClrLoadingButton implements LoadingListener {
     switch (state) {
       case ClrLoadingState.DEFAULT:
         this.renderer.removeStyle(this.el.nativeElement, 'width');
+        this.renderer.removeStyle(this.el.nativeElement, 'transform'); // for chromium render bug see issue https://github.com/vmware/clarity/issues/2700
         if (!this.disabled) {
           this.renderer.removeAttribute(this.el.nativeElement, 'disabled');
         }
         break;
       case ClrLoadingState.LOADING:
         this.setExplicitButtonWidth();
+        this.renderer.setStyle(this.el.nativeElement, 'transform', 'translatez(0)'); // for chromium render bug see issue https://github.com/vmware/clarity/issues/2700
         this.renderer.setAttribute(this.el.nativeElement, 'disabled', '');
         break;
       case ClrLoadingState.SUCCESS:


src/clr-angular/button/button-loading/loading-button.ts
```diff
@@ -73,12 +73,14 @@ export class ClrLoadingButton implements LoadingListener {
...
       case ClrLoadingState.DEFAULT:
         this.renderer.removeStyle(this.el.nativeElement, 'width');
+        this.renderer.removeStyle(this.el.nativeElement, 'transform');
         if (!this.disabled) {
           this.renderer.removeAttribute(this.el.nativeElement, 'disabled');
         }
         break;
       case ClrLoadingState.LOADING:
         this.setExplicitButtonWidth();
+        this.renderer.setStyle(this.el.nativeElement, 'transform', 'translatez(0)');
```
Shijir: Can we apply these styles to the animation specs or through the host binding instead of using Renderer2? Is that possible? Renderer2 ought to be used in cases where we need to directly manipulate DOM styles without dealing with Angular change detection.

youdz: Cory asked me this question, it's a matter of synrhonicity to interact with animations, and leftover attributes.
For instance, you can't do `this.renderer.removeAttribute(this.el.nativeElement, 'disabled')` with `@HostBinding`. Any value you use would leave a disabled attribute on the element no matter what.
And for the synchronicity, remember that HostBinding doesn't do anything until the view step of change detection, which can be problematic for our integration with animations.

I think this is 👌.

youdz: I just realized something after approving, do you mind adding a comment explaining briefly why we need this, maybe adding a link to this PR or the original issue? Thanks.

Shijir: Assigning`null` or `undefined` to the value can remove the `disabled` attribute though 😛 

I am just a bit concerned we might adopt a habit of resorting to doing things manually controlling DOM attributes/styles while there is an more "Angular way" option.

But this fix is fine by me as we had already been using Renderer2 in the component.

PR #2929, opened 2018-12-10
[UI/NG] Focus state for button group

jeeyun:
I think I found a solution that works for all browsers that we support. Let me know if you have concerns with this approach. Attaching screenshots of what they look like in each browser:

Chrome
chrome2
chrome

Firefox
ff2
ff

Safari
safari2
safari

Edge
edge2
edge

IE11
ie112
ie11


jeeyun:
Just noticed the visual impact of `overflow: visible;`. Let me investigate and update the PR.

src/clr-angular/button/button-group/_button-group.clarity.scss
```diff
@@ -15,6 +15,27 @@
...
+      }
+
+      input[type='radio']:focus,
+      input[type='checkbox']:focus {
+        appearance: none;
+        opacity: 1;
+        top: 0;
+        height: 100%;
+        width: 100%;
+        box-shadow: 0 0 5px rgba(81, 203, 238, 1);
```
Shijir: Don't we need `outline:none` as we are providing our own custom focus outline style here? 
Also, the color used seems to be very specific. Do we care whether this color is accessible or not?

jeeyun: That's a great question. I'm a bit torn on this, because the native outline provides the familiar visual indicator for focused state in chrome, safari, and IE so it's more recognizable even though we now have the added box shadow on top of that. In IE11, the box shadow isn't that prominent without the native outline (dotted line). If I change the box shadow property to make it more noticeable in IE11, chrome and safari look very different from the native outline style. I guess I'm not opposed to that but it'd seem like a bigger visual change and departure from the native outline.

I copied the actual color used here from a gist, and the idea is to mimic the glow color with it. I'm open to it being a different color (cc @lil-kim, @colinreedmiller) but we should keep it close to the focus color that's used throughout our components. 

I believe the color contrast ratio requirement for a11y is for distinguishing text against a background color.. For example, safari's focus color is super faint by default and what you see in the screenshot is that.

src/clr-angular/button/button-group/_button-group.clarity.scss
```diff
@@ -15,6 +15,27 @@
...
       vertical-align: top;
+      overflow: visible;
+
+      label {
+        height: 100%;
+      }
+
+      input[type='radio']:focus,
+      input[type='checkbox']:focus {
+        appearance: none;
```
Jinnie: The idea is great. You may also need a `-webkit-appearance: none;`. I tried to reproduce it in a stackblitz and got the native input remaining too: 
screen shot 2018-12-11 at 11 29 39 am

Please, check it here - https://stackblitz.com/edit/jeeyuns-awesome-focus-fix?file=src/app/app.component.css

nit: Some screens with the active/primary inputs focused would be helpful, too.

src/clr-angular/button/button-group/_button-group.clarity.scss
Jinnie: What do we need the explicit border for?
I tried changing it to red and this was the result:
screen shot 2018-12-11 at 11 41 14 am
(chrome on mac)

src/clr-angular/button/button-group/_button-group.clarity.scss
youdz: We use autoprefixer to not have to worry about all these vendor prefixes. That's why you see the problem in StackBlitz but Jeeyun doesn't have it in the project.

src/clr-angular/button/button-group/_button-group.clarity.scss
```diff
@@ -15,6 +15,27 @@
...
+        opacity: 1;
+        top: 0;
```
youdz: I don't see any positioning absolute or relative. What's this `top: 0` doing?

src/clr-angular/button/button-group/_button-group.clarity.scss
youdz: I'm with Jeeyun on this one, keeping the browser's default focus (when it exists) on top of ours is better for people who rely on it. From the various screenshots posted, I don't think the double outline looks bad in any case.

mathisscott: +1

Jinnie: I'm also with Jeeyun on this one. The question was only for my information. Sorry if it sounded like a request for change. Thanks for the details :D

src/clr-angular/button/button-group/_button-group.clarity.scss
jeeyun: We use a mixin that sets the position to absolute with a non-zero top value:

https://github.com/vmware/clarity/blob/7562bbff2ddd251393cc652af9905766f3ca7f11/src/clr-angular/forms-deprecated/utils/_mixins.clarity.scss#L10-L23

src/clr-angular/button/button-group/_button-group.clarity.scss
jeeyun: @Shijir I think it's ok for now. It's confined to one section of our scss file and not really re-usable anywhere else. Might be worth it in future if we decide to have custom focus across all our components.

Shijir: Sounds good 👍

PR #2938, opened 2018-12-13
[NG] fix(datagrid): Removes width calculation from datagrid render cycle

### LINKED ISSUE #2926: [WEB] Datagrid initial width not evaluated
## Describe the bug

Going to the datagrid page, the grid is not fully initialized.
screen shot 2018-12-10 at 1 30 49 pm

I am still not sure if this is website problem, or datagrid problem, after the initialization is now semi-manual via the `resize()` method.

## How to reproduce

1.  Go to https://vmware.github.io/clarity/documentation/v1.0
2.  Click on the "Datagrid" nav option
3.  Verify that the datagrid is as displayed in the screenshot from the description. NOTE: if you go directly to the page, or change through the local datagrid links, the datagrid is fine. You need to navigate to the "Datagrid" from another page.

## Expected behavior

Columns should be sized accordingly to the datagrid width

## Versions

**App**

* Angular: 7
* Node: 10.1
* Clarity: 1.0

**Device:**

* Type: Mac
* OS: OSX
* Browser chrome
* Version 70.0.3538.102

hippee-lee:
- fixes #2926

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -24,7 +24,7 @@
     border-bottom-right-radius: 0;
     -ms-overflow-style: -ms-autohiding-scrollbar;
     overflow: auto;
-    height: 100%;
+    flex: 1 1 auto;
     margin-top: 0.5rem;
   }
 
@@ -73,13 +73,11 @@
   }
 
   .datagrid-row {
-    display: block;
-    width: auto;
-
     &:hover {
       .datagrid-row-sticky,
       .datagrid-row-scrollable,
-      .datagrid-row-detail {
+      .datagrid-row-detail,
+      .datagrid-cell {
         background-color: $clr-datagrid-row-hover;
       }
     }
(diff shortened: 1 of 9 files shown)


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -843,6 +846,10 @@
         margin: 0;
         outline-offset: $clr-datagrid-compact-outline-offset;
       }
+
+      .clr-form-control-disabled .datagrid-footer-select.clr-checkbox-wrapper input[type='checkbox']:checked + label {
+        top: 0;
```
youdz: Just curious, why did the footer "selection count" change? I wouldn't expect it to be impacted.

hippee-lee: The inherited forms control is overridden above (when not in compact mode). I tried to only use the class but it is so specific above I had to duplicate the same selector here to override this in compact mode. 

Technically its not related to the width calculation but I saw it when working through all of our demos. I can separate it out into a different PR if you want me to.

youdz: That's okay, you can leave it here. :)

PR #2939, opened 2018-12-13
[NG] tests are added and updated

Shijir:

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-separator.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-separator.spec.ts
@@ -19,6 +19,13 @@ import { MOCK_TABLE_SIZE_PROVIDER } from './providers/table-size.service.mock';
 import { DatagridRenderOrganizer } from './render/render-organizer';
 import { generateDragPosition } from '../../utils/drag-and-drop/helpers.spec';
 
+@Component({
+  template: `<clr-dg-column-separator></clr-dg-column-separator>`,
+  styles: [':host { position: position; width: 200px; height: 400px;}'],
+  providers: [ColumnResizerService, MOCK_TABLE_SIZE_PROVIDER],
+})
+class TestComponent {}
+
 export default function(): void {
   describe('ClrDatagridColumnSeparator component', function() {
     let fixture: ComponentFixture<any>;
@@ -136,9 +143,3 @@ export default function(): void {
     });
   });
 }
-@Component({
-  template: `<clr-dg-column-separator></clr-dg-column-separator>`,
-  styles: [':host { position: position; width: 200px; height: 400px;}'],
-  providers: [ColumnResizerService, MOCK_TABLE_SIZE_PROVIDER],
-})
-class TestComponent {}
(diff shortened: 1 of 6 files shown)


src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -93,7 +99,190 @@ export default function(): void {
...
+    it('emits new header width once resizing ends', function() {
+      expect(context.testComponent.newWidth).toBeUndefined();
+      const resizeBy = 20;
+      emulateResizeOnColumn(resizeBy, columnHeader3DraggableDirective);
+      expect(context.testComponent.newWidth).toBe(column3InitialWidth + resizeBy);
+    });
+  });
 }
 
 @Component({ template: `<clr-dg-column>Hello world</clr-dg-column>` })
```
hippee-lee: Minor request: Can we move these to the top of the file before the exported function? IIR we are going to organize test files like that moving forward.

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -93,7 +99,190 @@ export default function(): void {
...
+
+    beforeEach(function() {
+      context = this.create(ClrDatagrid, HeaderResizeTestComponent);
+
+      columnHeader1DebugElement = context.fixture.debugElement.queryAll(By.directive(DatagridHeaderRenderer))[0];
+      columnHeader2DebugElement = context.fixture.debugElement.queryAll(By.directive(DatagridHeaderRenderer))[1];
+      columnHeader3DebugElement = context.fixture.debugElement.queryAll(By.directive(DatagridHeaderRenderer))[2];
+      columnHeader4DebugElement = context.fixture.debugElement.queryAll(By.directive(DatagridHeaderRenderer))[3];
+
+      columnHeader1ResizerService = columnHeader1DebugElement.injector.get(ColumnResizerService);
```
hippee-lee: Does there need to be one of these for each column? What happens if there is just one columnHeaderResizerService and all of the columns use it?

src/clr-angular/data/datagrid/datagrid-column-separator.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      columnResizerService.isWithinMaxResizeRange = false;
+      columnSeparatorComponent.moveTracker(mockExtDragEventExceededRange, resizeTrackerEl);
+      expect(resizeTrackerEl.classList.contains('exceeded-max')).toBeTrue();
+      columnResizerService.isWithinMaxResizeRange = true;
+      columnSeparatorComponent.moveTracker(mockExtDragEventWithinRange, resizeTrackerEl);
+      expect(resizeTrackerEl.classList.contains('exceeded-max')).toBeFalse();
+    });
+  });
+}
+@Component({
```
hippee-lee: I think this should be above the test suite.

hippee-lee:
The tests themselves look ok to me. I didn't review the components and services and will need some more time to look at this in context of these tests.

PR #2941, opened 2018-12-13
[website] Add more tags for icon search

jeeyun:
Adding more tags to icons for better search + minor fixes to routing issue in icons page.


gnomeontherun:
I'm not sure what happened but this isn't a clean commit so can you fix that up?

PR #2944, opened 2018-12-14
[NG] fix aot errors in dev app

coryrylan:
Fixes aot errors as well as minor code clean up in related components for the dev app.

closes vmware#2932


src/dev/src/app/checkboxes/checkboxes.demo.ts
```diff
@@ -119,6 +120,7 @@ export class CheckboxesDemo {
     name: new FormControl('', [Validators.required]),
     vmsSelection: new FormArray((() => this.vms.map(() => new FormControl(false)))(), [minimumSelection(1)]),
   });
+  vmsSelection: AbstractControl[] = (this.reactiveForm.controls.vmsSelection as FormArray).controls;
```
coryrylan: This was weird to get AOT to be ok with, any suggestions? The issue was AOT saw `controls.vmsSelection` as type `AbstractControl` and not a `FormArray`

src/dev/src/app/wizard/wizard-alt-next.demo.html
```diff
@@ -26,7 +26,7 @@
 
   <clr-wizard-page (clrWizardPageNext)="pageCustomNext()" [clrWizardPagePreventDefaultNext]="true">
     <ng-template clrPageTitle>Page level alt-next</ng-template>
-    <p *ngIf="showCancelConfirm">Complete this fibonacci sequence</p>
+    <p>Complete this fibonacci sequence</p>
```
mathisscott: Why are we getting rid of the `ngIf` if we are keeping the toggling of the `showCancelConfirm` property in `wizard-alt-cancel.demo.ts`?

coryrylan: I removed `showCancelConfirm` in `wizard-alt-next.demo.html` since it was not referenced anywhere else in the template or `wizard-alt-next.demo.ts` so it was always hidden. The `showCancelConfirm` property in `wizard-alt-cancel.demo.ts` was being used in multiple places to demo the cancel functionality.

src/dev/src/app/checkboxes/checkboxes.demo.ts
gnomeontherun: I had several issues with typing when working in forms, this seems unfortunate.

src/dev/src/app/datagrid/kitchen-sink/kitchen-sink.ts
```diff
@@ -74,7 +77,7 @@ export class DatagridKitchenSinkDemo {
     this.cleanUp();
   }
 
-  cleanUp(): void {
+  cleanUp() {
```
jeeyun: Curious.. how does having the return type of `void` here impact/block AOT?

youdz: It doesn't impact, but having it explicit is just extra verbosity that doesn't do anything for the compiler. So my guess is Cory might have used a slightly stricter linting config or something similar, that removed "useless" typings. I'm 👍 with that.

coryrylan: It doesn't apply to AOT but I believe TypeScript will usually will infer void on a method without a return statement. I removed it as the rest of the methods did not use it. This was something I noticed going through the dev app, some methods use void while others did not. I was not sure what style the team prefers so I just made it consistent to the rest of the file.

PR #2948, opened 2018-12-17
replace uses of old grid in website and documentation

### LINKED ISSUE #2927: [DOCS] Cards documentation still uses the old grid
## Describe the bug

Our cards documentation examples still uses the unprefixed grid class names. We need to move them to our new grid.

## How to reproduce

https://vmware.github.io/clarity/documentation/v1.0/cards

## Versions

* Angular: 7.x
* Clarity: 1.0

gnomeontherun:
This was tedious, but important.

closes #2927


gnomeontherun:
@jeeyun The challenge is writing a schematic that does this correctly, because these classes are found in many places (could be in a template (html or ts file), or a host binding, or a ngClass binding, or other places). I ultimately didn't feel confident we could do that correctly.

jeeyun:
Makes sense. Thanks for the feedback.

PR #2958, opened 2018-12-19
[NG] Column Resize With DnD

Shijir:


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -467,24 +467,25 @@
         }
 
         .datagrid-column-handle {
-          @include clr-no-styles-button();
-          display: block;
           position: absolute;
           width: calc(0.5rem + #{$clr-rem-1px});
-          right: -0.5rem;
+          right: -0.25rem;
```
hippee-lee: Did removing the button change the offset values?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -467,24 +467,25 @@
...
           cursor: col-resize;
           height: calc(100% + 0.5rem - #{$clr-default-borderwidth});
           z-index: map-get($clr-layers, datagrid-header);
         }
-        .datagrid-column-handle-tracker {
+        .datagrid-column-resize-tracker {
           position: absolute;
-          right: 0;
-          top: $clr-datagrid-column-handle-tracker-top;
+          top: -0.5rem;
```
hippee-lee: This changed to a hard coded value (which I'm ok if its only ever used for this feature). However, I dont see its removal from the `_variables.datagrid.scss` file.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -467,24 +467,25 @@
         }
 
         .datagrid-column-handle {
```
hippee-lee: Can you confirm that the last column doesn't affect the width of the `.datagrid-table` element? I recall needing to rework the layout for this to keep it from impacting the table width.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
Shijir: This was wrong value initially. The separator underneath the resize handle should be exactly aligned at the middle.

**Before:**
screenshot 2018-12-21 14 54 12

**After:**
screenshot 2018-12-21 14 56 13

src/clr-angular/data/datagrid/_datagrid.clarity.scss
Shijir: You're right! I will remove it from those variables.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -0,0 +1,76 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { DOCUMENT } from '@angular/common';
+import { Component, Inject, Renderer2 } from '@angular/core';
+
+import { ClrDragEvent } from '../../utils/drag-and-drop/drag-event';
+import { ColumnResizerService } from './providers/column-resizer.service';
+import { TableSizeService } from './providers/table-size.service';
+
+let nbCount: number = 0;
```
hippee-lee: Is there a reason not to use UNIQUE_ID_PROVIDER here instead?  We use it in tree-node and modals to set the instance id.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
Shijir: Do you remember the cases in which that specific behavior should be tested? I could cover that in `header-render.spec.ts` which checks all Datagrid headers together in an integrated manner. But as far as I see it visually on our demo app, it's behaving as expected. 

I have also a demo here: http://dnd-resize-header-demo.surge.sh/datagrids.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
Shijir: @hippee-lee I wasn't really aware of that provider. But here we are not generating a unique id for the instance itself, but for a resize draggable's `clrGoup` input. Though I think I could still use a unique id generated by UNIQUE_ID_PROVIDER for the `clrGoup` key as it provides unique id/key. But I have one hesitation to hold off of doing so right now. So I will ask some questions below. I would love to hear you and other reviewers perspectives on the question.

We will need to generate unique id/key for draggable columns, rows., etc as well eventually. On top of that, we will have many other directives and components that need to generate some kind of unique key or id. In those cases, do we need to utilize UNIQUE_ID_PROVIDER? Because by injecting the provider, all those directives/components become dependent on one more dependency injection. Isn't it better to have fewer dependencies for directives/components if possible? And on top of that, to generate a unique id, the provider needs to get instantiated in each directive/component instances (columns, rows, cells, etc.,) while we could generate that unique id just by incrementing a integer value in the constructor. So my question is: When do we need to use UNIQUE_ID_PROVIDER? What does it do better in terms of generating a unique id/key than incrementing a integer value in the constructor?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
hippee-lee: There was an issue I saw a while ago where the column handle element was positioned outside the column and added some width to the overall header width when dragging. I looked at the demo and it looks ok.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
hippee-lee: I think its better to have one place for generating and tracking unique ides instead of duplicating that code in different components. I think this has more value than  adding fewer dependencies to components/directives, imo. 
I believe that `useFactory` just injects the function and `let NB_INSTANCES = <value>;` is global to the Clarity app. This means one variable that stores the current `state` of uniqueness and then the function gets injected into things that need to increment it.

Shijir: I see. Thanks for the explanation. I will go with the provider then 👌

PR #2962, opened 2018-12-20
add back pagination separator

Shijir:
screenshot 2018-12-20 08 44 06

screenshot 2018-12-20 08 45 09

screenshot 2018-12-20 08 46 08

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -2,6 +2,16 @@
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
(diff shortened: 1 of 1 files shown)


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -643,23 +653,26 @@
...
+          &:before {
+            @include pagination-separator();
+          }
+        }
       }
 
       &-description {
         white-space: nowrap;
+
+        & + .pagination-list {
```
youdz: I thought you were missing the case with `.pagination-size` and `.pagination-list` and no `.pagination-description`, but I went back and realized that we actually will display a weird empty separator all the time anyway. So I guess we just don't support the case with no description.

That's fine by me.

PR #2964, opened 2018-12-20
fix import path in datepicker module

Shijir:


gnomeontherun:
Hmmm, why didn't our tslint catch this?

Shijir:
@gnomeontherun I was surpirsed as well. But this one is tricky I think. We have 
```
"no-barrel-imports": true,
```
But this one only catches paths like `./`

sdurnov:
Updated to clr 1.0.3 and got a build error, related to this. Weird that your CI didn't catch that..

Issue #2965, opened 2018-12-21
Tabs content DOM has extra nodes


Jinnie:
## Describe the bug

Currently the DOM generated by the tabs component looks like this:

> clr-tabs(
>  ul (contains all tab-links)
>  clr-tab - EMPTY
>  clr-tab - EMPTY
>  clr-tab - clr-tab-content
>  clr-tab - EMPTY
> )

The empty clr-tab nodes make it very difficult for the clr-tab-content to successfully obtain 100% parent width/height, because all clr-tab nodes are CSS-indistinguishable and can only be styled in bulk

If they don't have width, then the clr-tab-content can't have width/height:100%.
<img width="406" alt="screen shot 2018-12-19 at 11 19 26 pm" src="https://user-images.githubusercontent.com/1798828/50345097-7d4bc000-0536-11e9-868d-747a1bc657e3.png">

If we set width to them, or add them in a flexbox, then they all get some width/height, even the empty ones, and break the layout.

<img width="398" alt="screen shot 2018-12-19 at 11 14 50 pm" src="https://user-images.githubusercontent.com/1798828/50345098-7d4bc000-0536-11e9-95b7-8b6b15476449.png">

## How to reproduce

1.  Go to https://clarity.website/documentation/tabs
2.  Scroll down to Angular Component
3.  Inspect the DOM

## Expected behavior

The generated DOM should have only one content, which is not nested in a clr-tab, but in a section, compatible with the Static definition of the Docs website and the Clarity UI definitions.

<img width="635" alt="screen shot 2018-12-21 at 3 49 45 pm" src="https://user-images.githubusercontent.com/1798828/50345524-16c7a180-0538-11e9-950c-d0c0da185c04.png">

## Versions

**App**

* Angular: 7
* Node: n/a
* Clarity: 1.0.x

## Additional notes

Prerequisite for #452 Vertical tab component

PR #2966, opened 2018-12-21
[NG] Remove extra nodes from the Tabs DOM (#2965)

Jinnie:
Selectively projecting only the content. Actually, I'm restoring
the template from git logs, as it's been like this before, but
removed for simplification. This way I'm populating some leftover
templateref vars, that were undefined before.


src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -12,7 +12,14 @@ let nbTabContentComponents: number = 0;
 @Component({
   selector: 'clr-tab-content',
   template: `
+    <ng-template #tabContentProjectedRef>
```
Jinnie: Restored from git log. It corresponds to a "template" var below, that was previously empty.

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -12,7 +12,14 @@ let nbTabContentComponents: number = 0;
...
+      <section [id]="tabContentId" role="tabpanel" [class.active]="active"
+               [attr.aria-labelledby]="ariaLabelledBy"
+               [attr.aria-hidden]="!active" [attr.data-hidden]="!active">
         <ng-content></ng-content>
+      </section>
+    </ng-template>
+    <ng-container [ngTemplateOutlet]="tabContentProjectedRef"></ng-container>
```
Jinnie: I am adding a projection of the the template here, so the component can be instantiated and tested in isolation. Alternatively, I can change some tests that we have...

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -45,14 +46,17 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
             </ng-container>
         </ul>
         <!--tab content-->
-        <ng-content></ng-content>
+        <ng-container [ngTemplateOutlet]="tabContent.first?.templateRef" ]></ng-container>
```
Jinnie: I am projecting only a single content instance, opposed to many clr-tab elements, one of which holds a tab-content.

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -12,7 +12,14 @@ let nbTabContentComponents: number = 0;
 @Component({
   selector: 'clr-tab-content',
   template: `
+    <ng-template #tabContentProjectedRef>
+      <section [id]="tabContentId" role="tabpanel" [class.active]="active"
```
youdz: If you add this, you need to remove the identical bindings on the host.

Jinnie: Right... I also need to add the '[attr.aria-expanded]': 'active' that appears on the host, but not here. I'll do it.

src/clr-angular/layout/tabs/tab-content.ts
youdz: Since you're using the template in the tabs component HTML, this is clearly not needed. I get that it simplifies tests, but it also runs the risk of instantiating too many contents by mistake in the future, impacting performance.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -45,14 +46,17 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
     `,
   providers: [IfActiveService, IfOpenService, TabsService, TABS_ID_PROVIDER],
 })
 export class ClrTabs implements AfterContentInit {
   @ContentChildren(ClrTabLink, { descendants: true })
   tabLinkDirectives: QueryList<ClrTabLink>;
 
+  @ContentChildren(ClrTabContent, { descendants: true })
+  tabContent: QueryList<ClrTabContent>;
```
youdz: If you're always only using the first one, why did you go with `@ContentChildren` instead of `@ContentChild`?

src/clr-angular/layout/tabs/tabs.ts
youdz: Here is the thing: as much as we tell people to use `*clrIfActive`, we know of at least a handful of internal cases at VMware that don't, and rely on show/hide. With this solution, we can never display the content of anything except the first tab.

src/clr-angular/layout/tabs/tabs.ts
Jinnie: Added tests and code for the case without *clrIfActive.
@youdz, the other changes we discussed remain absolutely valid. My suggestion is to finalize this one, so it can allow some immediate vertical style changes and then prioritize the other changes in a separate task, which can get proper prioritization in a normal planning session.

:santa:

src/clr-angular/layout/tabs/tab-content.ts
gnomeontherun: Can we include the `hidden` html attribute here?

jeeyun:
I'm still bothered by the fact that I don't remember in full detail why we had `#tabContentProjectedRef` template and removed it. This isn't a specific comment to your code changes, but can you make sure that tab contents still work as expected with overflow tabs, and in particular with dynamically adding an overflow tab at runtime? I just remember those being challenging when it came to rendering, and wondering if some of those are the reason why we have what we have today. Either way, thanks for your work on this!

src/clr-angular/layout/tabs/tab-content.ts
Jinnie: Yes, it was leading to issues already when trying to support the no-clrIfActive case and I removed it from the latest code.

src/clr-angular/layout/tabs/tabs.ts
Jinnie: Acknowledged. no-clrIfActive case implemented and added tests to cover it.

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -12,16 +12,15 @@ let nbTabContentComponents: number = 0;
...
+               [attr.aria-labelledby]="ariaLabelledBy"
+               [attr.aria-expanded]="active"
+               [attr.aria-hidden]="!active" [attr.data-hidden]="!active">
```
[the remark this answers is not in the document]
Jinnie: @jeeyun thanks for the comment!
This is the commit that introduced this change:
https://github.com/vmware/clarity/pull/1236
I don't see anything specific around the template code there, discussions or comments.
I checked the dynamic move of tab label to overflow menu list - it seems fine. I also added a few clicks to the Gemini tests, to verify it looks as expected.
Nevertheless, it's in the tab-content code, so it should have no effect on the tab labels at all. It's only about how the content DOM is structured. Should be pretty safe.

@gnomeontherun fixed it, thanks.

src/clr-angular/layout/tabs/tab-content.ts
Jinnie: Added both for now.

gemini/tests/set4/tabs.js
```diff
@@ -27,6 +27,16 @@ gemini.suite('tabs', child => {
         actions.wait(WAIT_LOAD_TIME);
       })
       .setCaptureElements('.clr-example')
-      .capture('default');
+      .capture('default')
+      .capture('overflowing', (actions, find) => {
+        actions.executeJS(function(window) {
+          document.getElementById('toggleOverflow').click();
+        });
+        actions.wait(WAIT_LOAD_TIME);
```
gnomeontherun: This is probably unnecessary.

src/clr-angular/layout/tabs/tab-content.spec.ts
```diff
@@ -14,6 +14,7 @@ import { ClrTabContent } from './tab-content';
 @Component({
   template: `
         <clr-tab-content>Content1</clr-tab-content>
+        <ng-container [ngTemplateOutlet]="tabContent.templateRef"></ng-container>
```
gnomeontherun: I think we need a comment here _why_ this is here. Its here because we don't have a clr-tabs container which hosts this container for us.

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -12,16 +12,17 @@ let nbTabContentComponents: number = 0;
 @Component({
   selector: 'clr-tab-content',
   template: `
+    <ng-template #tabContentProjectedRef>
+      <section [id]="tabContentId" role="tabpanel" [class.active]="active"
+               [hidden]="!active"
+               [attr.aria-labelledby]="ariaLabelledBy"
+               [attr.aria-expanded]="active"
+               [attr.aria-hidden]="!active" 
+               [attr.data-hidden]="!active">
```
gnomeontherun: Do we need `data-hidden`? Thats just something you see more with Bootstrap and isn't necessary I believe.

src/clr-angular/layout/tabs/tab-content.spec.ts
Jinnie: You're right. Even I needed some time to remember why we do it this way.

src/clr-angular/layout/tabs/tab-content.ts
Jinnie: I restored it from git history like this. Even if it had been needed some time ago, there is no obvious reason to keep it now. Removing it.

Jinnie:
Thanks for checking this case, Shijir!

PR #1530, opened 2017-09-29
[NG] Adds alerts pager and container to show multiple app level alerts

bdunkz:
- Adds clr-alerts container to hold clr-alerts
    - Adds clr-alerts-pager to page through multiple alerts
    - closes #295

Signed-off-by: Brian Duncan <bduncan@vmware.com>

PR #2971, opened 2018-12-28
[NG] fix alert visibility issue

### LINKED ISSUE #2430: Clarity alert should be visible, when a new alert is pushed to an empty list of alerts
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

A Clarity alert should be visible, when a new alert is pushed to an empty list of alerts
This happens in production mode only. It works as expected in development mode.

### Actual behavior

Clarity alert is not visible, the first time an alert is pushed. 
On the next push of an alert, both alerts show up.
After deleting all alerts, this behavior can be reproduced!

Stackblitz url: https://stackblitz.com/edit/clarity-light-theme-v11-6npkw

### Reproduction of behavior

It works fine in Stackblitz, since this uses development mode.
The Stackblitz environment does not allow to enable production mode by inserting the statement

'enableProdMode();' into main.ts right after the last import statement, because of

Error in 5.2.1/bundles/core.umd.js (5296:15)
Cannot enable prod mode after platform setup.
Evaluating main.ts
Booting application
 
However, outside Stackblitz the production mode can be set this way and the malfunction can be reproduced

Looking at the Elements in Inspect Browser, you can see the difference:  
In production mode a  'class="alert alert-info alert-hidden"' is generated, which cause the alert not to be displayed.
'alert-hidden' is NOT generated in development mode.

### Environment details

* **Angular version:** 5
with
* **Clarity version:** 0.11

as well as 

* **Angular version:** 6
with
* **Clarity version:** 0.12.0

* **OS and version:** Ubuntu 18.04 LTS

* **Browser:** tested with
Chrome  Version 67.0.3396.99 (Official Build) (64-bit)
Mozilla Firefox for Ubuntu 60.0.2 (64-bit)

coryrylan:
In prod mode the first alert in alert list is not visible to the user. This happens when using `enableProdMode()` and is unrelated to AOT.

The check `isHidden()` only happens once during prod mode while in dev happens multiple times. In production it looks like `isHidden` is called before the `QueryList` of the parent alerts component has updated. So when `isHidden` checks to see if there are any alerts in the `QueryList` is empty the first time. Every subsequent check seems to work normally. 

I found a "fix" but I don't think its a good solution. The code change checks if the list is empty then it can assume its the first alert in the alert list and can fall into the correct if block. Any feedback or ideas would be appreciated.

closes #2430

### DIFF
--- a/src/clr-angular/emphasis/alert/alert.ts
+++ b/src/clr-angular/emphasis/alert/alert.ts
@@ -60,7 +60,9 @@ export class ClrAlert {
 
   get isHidden() {
     if (this.multiAlertService) {
-      if (this.multiAlertService.currentAlert === this) {
+      // change detection issue in production mode causes currentAlert to be undefined when only the first alert exists
+      // https://github.com/vmware/clarity/issues/2430
+      if (this.multiAlertService.currentAlert === this || this.multiAlertService.count === 0) {
         if (this.hidden === true) {
           this.previouslyHidden = true;
           this.hidden = false;


Shijir:
Actually, I just tracked the component's history and the PR that introduced this component. Eudes actually commented on the very code block here: https://github.com/vmware/clarity/pull/1530#discussion_r143845467
So you might also want to take a look into using Oompa-Loompa to solve the issue.

gnomeontherun:
This logic is strange, and I think we can do better. If we agree to this fix for now, we should open a new ticket to refactor this logic and get away from running change detection this way. I'd be ok with running that fix through at the moment, if we commit to a follow up in short order.

PR #2982, opened 2019-01-03
[NG] make datagrid header draggable

Shijir:
This PR includes changes that sets draggable directives to our datagrid headers. 

There are two issues that I found while testing that haven't been solved in this PR. I would like to hear reviewers' takes on them:

- On iOS devices, the scrolling is being triggered during dragging. Only trick that was solving this issue was using `preventDefault` on `touchstart` event.

- The draggable ghost that's resetting/returning to its origin position is being clipped if the header gets dragged outside the datagrid. This is only specific to Chrome. 

 


src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -197,8 +196,8 @@ export default function(): void {
       column3InitialWidth = widthOf(columnHeader3Element);
       column4InitialWidth = widthOf(columnHeader4Element);
 
-      columnHeader1DraggableDebugElement = context.fixture.debugElement.queryAll(By.directive(ClrDraggable))[0];
-      columnHeader3DraggableDebugElement = context.fixture.debugElement.queryAll(By.directive(ClrDraggable))[2];
+      columnHeader1DraggableDebugElement = context.fixture.debugElement.queryAll(By.directive(ClrDraggable))[1];
+      columnHeader3DraggableDebugElement = context.fixture.debugElement.queryAll(By.directive(ClrDraggable))[5];
```
gnomeontherun: `columnHeader5DraggableDebugElement`?

Shijir: This tells me need to improve these variable names. :)

Shijir:
@gnomeontherun those are good points. Currently, once we place the `clrDraggable` directive on any element, that element will be draggable permanently. So eventually as you mentioned, we might want to add that option of locking columns or making columns non-draggable.

Another point you have raised is closing a filter panel. And I believe it should be closed once dragging starts. I think that could be addressed in this PR.

PR #2988, opened 2019-01-03
[NG] datepicker fixes

coryrylan:
- After date selected by user refocus the input.

- If updating the value from user then programmatically and back to the user, the output event
should fire properly for two way binding.

closes vmware#2792

closes vmware#2993

### DIFF
--- a/src/clr-angular/forms/datepicker/date-input.spec.ts
+++ b/src/clr-angular/forms/datepicker/date-input.spec.ts
@@ -130,9 +130,7 @@ export default function() {
           const input: HTMLInputElement = context.testElement.querySelector('input');
           expect(document.activeElement).not.toBe(input);
 
-          ifOpenService.open = true;
-          context.detectChanges();
-          ifOpenService.open = false;
+          dateNavigationService.notifySelectedDayChanged(new DayModel(2019, 1, 1));
           context.detectChanges();
 
           expect(document.activeElement).toBe(input);
--- a/src/clr-angular/forms/datepicker/date-input.spec.ts
+++ b/src/clr-angular/forms/datepicker/date-input.spec.ts
@@ -196,7 +196,7 @@ export default function() {
           expect(context.testComponent.date.getDate()).toBe(1);
         });
 
-        it('outputs the date when the user changes the date manually in th input', () => {
+        it('outputs the date when the user changes the date manually in the input', () => {
           expect(context.testComponent.date).toBeUndefined();
 
           const input: HTMLInputElement = context.testElement.querySelector('input');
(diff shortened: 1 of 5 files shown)


src/clr-angular/forms/datepicker/providers/datepicker-focus.service.ts
```diff
@@ -15,20 +15,23 @@ import { first } from 'rxjs/operators';
...
     });
   }
+
+  focusInput(element: HTMLInputElement): void {
+    this._ngZone.runOutsideAngular(() => this.ngZoneIsStableInBrowser().subscribe(() => element.focus()));
+  }
+
+  private ngZoneIsStableInBrowser() {
+    // Credit: Material: https://github.com/angular/material2/blob/master/src/lib/datepicker/calendar.ts
+    return this._ngZone.onStable.asObservable().pipe(first(), map(() => isPlatformBrowser(this.platformId)));
```
youdz: I think you wanted to use `filter()`, not `map()`. With your code, you lost the `if (isPlatformBrowser(this.platformId))` from before, so you might crash server-side rendering.

src/clr-angular/forms/datepicker/date-input.spec.ts
```diff
@@ -118,6 +126,18 @@ export default function() {
           sub.unsubscribe();
         });
 
+        it('should refocus input after selecting a date for a11y', () => {
+          const input: HTMLInputElement = context.testElement.querySelector('input');
+          expect(document.activeElement).not.toBe(input);
+
+          ifOpenService.open = true;
+          context.detectChanges();
+          ifOpenService.open = false;
```
youdz: The name of this test, with this code, made me realize we force focus on the input even if I close the dropdown **without** selecting a date. It might create some weird behavior with someone opening the dropdown, then changing their mind and clicking on a link or button somewhere else in the page. 

Is this something we care about (cc @mathisscott)? 

If it is, how hard would it be to tie the focus to actual date selection using `DateNavigationService.selectedDayChange`? After a quick look it seems easy enough, but I may be missing something.

src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -296,44 +301,58 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
+          this.control.control.markAsDirty();
+        }
+      })
+    );
+  }
+
+  private listenForInputRefocus() {
+    this.subscriptions.push(
+      this.ifOpenService.openChange
+        .pipe(skip(1), filter(open => !open))
```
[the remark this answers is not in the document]
coryrylan: Using `DateNavigationService.selectedDayChange` made the `skip(1)` unnecessary.

src/clr-angular/forms/datepicker/date-input.spec.ts
coryrylan: Agree with the UX concern. Updated to use `DateNavigationService.selectedDayChange` instead.

Issue #2908, opened 2018-12-04
Datagrid Numeric Filter


IoannesBracciano:
## Summary

Data-grid columns with numeric values should have a dedicated numeric filter. We plan to build this for our team but since it is such a common feature we felt like we should propose a contribution so that more people can benefit from it and not have to write the same filter over and over again.

## Use case

Numeric data-grid columns will make use of the filter. Since it is specific to numbers, the filter can provide comparison options like _=_ equals, _>_ greater than, _<_ less than, between, etc.

## API

The filter can be implemented as an angular component and then be used inside `<clr-dg-filter>` tags:
```
<clr-datagrid>
    <clr-dg-column clrDgField="numericField">
        <clr-dg-filter>
            <clr-numeric-filter></clr-numeric-filter>
        </clr-dg-filter>
    </clr-dg-column>
</clr-datagrid>
```

youdz:
Thanks for offering to contribute to Clarity, as I said before filters are perfect candidates for first contributions.

Before we can go further, do you mind explaining or showing what the filter will be like? You say it will provide comparison options, but how will that be presented to the user? It could be a select element for the comparator, and a number input next to it. It could be multiple number inputs to allow for "between 10 and 20" filtering. It could be a text input that parses some mathematical syntax (please don't do that, it's a terrible idea 😂). It could be many other different interactions for the user.

Second, I recommend looking at our [existing string filter](ithub.com/vmware/clarity/blob/master/src/clr-angular/data/datagrid/built-in/filters/datagrid-string-filter.ts) and follow the same pattern, this way you'll obtain the much nicer syntax:
```html
<clr-datagrid>
    <clr-dg-column clrDgField="numericField">
        Column title
        <clr-numeric-filter></clr-numeric-filter>
    </clr-dg-column>
</clr-datagrid>
```
No need for that extra `<clr-dg-filter>`, since this will be one of Clarity's built-in filters.

amellnik:
@IoannesBracciano -- I'm also interested in this, and if there's anything I can do to help with this please let me know.

IoannesBracciano:
Greetings,
I have been a bit busy the last days but I took some time today to create a simple template of how we imagine the filter would look like.

<img width="920" alt="screen shot 2018-12-06 at 3 53 20 pm" src="https://user-images.githubusercontent.com/6152794/49588401-4474ee00-f96f-11e8-9bf5-a7cdd89d44b5.png">

[...]

Issue #2683, opened 2018-09-25
Datagrid: Various popovers get cut off depending on datagrid size and placement


hippee-lee:
## Describe the bug

This is a list of datagrid issues related to the lack of smart popovers in the datagrid:

1. Filters: When the datagrid is to short, the filter popover overflows with the rest of the header/table content underneath the footer. This can be worked around by setting a min height on the datagrid.
2. Hideable Column Toggle: Depending on the height of the datagrid, the list of hideable columns can be shown above the fold 
3. signposts / tooltips in rows can be `clipped` depending on the location and need to be manually set by the consumer to ensure onscreen and visible placement. 

## Expected behavior
After development of https://github.com/vmware/clarity/issues/1262#issuecomment-320699953 the issues with these popovers in datagrid should not occur.

amellnik:
One other issue that may already be solved: opening a filter on the first column can fall off the screen if the column is narrow.

hippee-lee:
Hey Alex, I don't think the issue you describe is already fixed. But, it should be fixed once I develop the fix for this issue.

sdurnov:
@hippee-lee do you have it planned for release anytime soon?

hippee-lee:
Hi @sdurnov - I have one more item that needs to be done for the upcoming v1.0 release and then I will start on the smart popover enhancement that will address these issues. 

I hope to start on it by end of this week or early next week.

hippee-lee:
Hey @JoshuaNitschke - thats one of the things the smart popover fix needs to address. For now, your best bet is to set a min height on the datagrid so that the filter popover doesn't overflow.

sdurnov:
@hippee-lee but that won't help with 1st columns filter popup. Btw, this issue is for 2.0 now? Meaning it is planned to be released in half a year? This is bad news :(

BramMeerten:
> 1. Filters: When the datagrid is to short, the filter popover overflows with the rest of the header/table content underneath the footer. This can be worked around by setting a min height on the datagrid.


I don't agree this should be part of an enhancement ticket. I think it should be a bug because it worked in the previous version. You can test this in the Custom Filtering documentation page of v.0.13:
https://vmware.github.io/clarity/documentation/v0.13/datagrid/custom-filtering
![image](https://user-images.githubusercontent.com/1921407/50888841-d355aa00-13f6-11e9-8e58-c6424489d978.png)

[...]

Issue #1632, opened 2017-10-23
Datagrid Enhancement: Filtering


gracesnoh:
## Description
Filtering in the datagrid component has been enhanced to include:
- Cross-column filtering
- Ability to easily clear filters
- Key:value filtering behavior
- Filter visibility

![documentaiton](https://user-images.githubusercontent.com/8237984/31907735-8a2e13e0-b7e9-11e7-9a72-c7d46eeae99f.png)

### Filters (new section)
Filters are text-input based and are represented by dynamic labels above the datagrid. Filtering can be performed both on a per-column and cross-column level. Since this method of filtering is unique, explanations of how to add, clear, and edit filters are detailed below. Built-in filters currently include string filters—see Custom Filters [link to Custom Filters section] to find out how to build your own filters.

![demoo](https://user-images.githubusercontent.com/8237984/31911592-114e3056-b7f6-11e7-999c-25dc1789cf25.png)

#### Adding filters
Filters can be added on a per-column or cross-column level. A dropdown of filtering suggestions will appear when the input is activated.

##### Per-column filters
![per_column](https://user-images.githubusercontent.com/8237984/31902453-c6fd3c8e-b7d9-11e7-8677-40b2cbcad1f5.gif)

Per-column filters are added by first typing or selecting the desired column to filter by, adding a “Key”. Any input typed or selected after the “Key:” is the “value” that will filter the column. Selecting an item from the dropdown or pressing enter will generate a filter.

##### Cross-column filters
![cross_column](https://user-images.githubusercontent.com/8237984/31902457-cca4e4f2-b7d9-11e7-972c-7e7241c1725e.gif)

Cross-column filters are added by directly typing the desired string to filter by into the input. Selecting an item from the dropdown or pressing enter will generate a filter. These filters are indicated with a globe icon.

#### Clearing filters
![clear_filters](https://user-images.githubusercontent.com/8237984/31905465-2023df26-b7e3-11e7-8d34-cc8ae0e34530.gif)

Individual filters can be cleared by clicking on the close icon of each filter. 

When more than three filters are applied, a “Clear all” option will appear at the end of the filters to clear all active filters at once. A modal asking to confirm this destructive action will help prevent accidentally clearing all filters.

> Info: Filters that exceed the width of the datagrid will overflow to a new row, which will push down the entire datagrid. While this behavior is not optimal, this design decision was made on the grounds that most users typically filter by only 1-3 filters.

#### Editing filters
[...]

Issue #1262, opened 2017-08-05
clr-dropdown should support a dynamic clrPosition


justindsmith:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[X] enhancement
```

### Expected behavior
When a `clrDropdownToggle` is not positioned so that the associated `clr-dropdown-menu`'s position (`[clrPosition]`) will not be visible in the page, the menu should draw in a direction that _will_ show fully on the page.

### Actual behavior
The dropdown will draw in the specified direction no matter where the menu is going to end up on the page.

This is especially tricky when iterating over a list as the main content of the page, with dropdowns included in the list items. In this case the dropdown menus for the last few items cause layout issues as they try to gain space for the menu.

### Reproduction of behavior
https://plnkr.co/edit/Jlmxcqe9419Nvgpxhkny?p=preview

### Environment details

* **Angular version:** 2.0.X
4.x

* **Clarity version:** 
0.10.0-rc

* **OS and version:** 

* **Browser:**
all

youdz:
This is what we've been categorizing as "collision detection" or "smart popovers" for our dropdowns, tooltips, signposts aand any other kind of popovers.

It has not been considered a priority so far. The amount of work required to handle collision detection for every possible use case out there, while integrating with Angular _and_ not slowing down performances, is gigantic. Knowing that this can be solved in all cases by adding minimal extra work on the app side (in your case, keep track of the index in your list's `*ngFor`, and position the last few dropdowns to the top rather than the bottom), the (work required / value provided) ratio is pretty low at the moment.

Thanks for filing this, I will raise it during our next planning session to see if anything has changed.

timvl:
is there a workaround possible?

juanmendes:
We are running into this problem by having a lot of options in the menu. Using the ngFor index to decide the position is not really feasible, you have to combine it with offsetTop, screen height to get decent results that don't scroll off the top. At the very least, Clarity should make a suggestion as to how longs menus from cards should work. Isn't there a fix like the one suggested at https://github.com/vmware/clarity/issues/1213? We were able to create a directive that implemented the suggested workaround.

KevinABoucher:
A similar issue occurs when a signpost is at the end of a paragraph that can flow to different positions on the screen.

marvinosswald:
is there any progress on this issue ? Or maybe a dirty workaround for the moment ?

timvl:
[...]

PR #2996, opened 2019-01-06
[NG] Datagrid numeric filter

amellnik:
This is a first pass at https://github.com/vmware/clarity/issues/2908.  It's not yet complete (no tests, minimal documentation, some bugs), but I was hoping to solicit some early feedback.  

image

The idea is to add a second built-in filter type that's similar in scope to the existing string filter.  Rather than a single input this second filter has two input fields which can optionally contain a lower and upper limit respectively.  This is currently called a numeric filter and uses number input fields, but a better name for it might be a range filter, since it would be easy for it to apply to non-numerical types (for example, a user could input letters to select a range alphabetically).  

There's demos for it in the datagrid section of the KS app in the "binding model properties" and the "built-in filters" sections.  

Open issues:
-  ~Rename to "range" filters and allow non-numeric inputs?~
- [x] Why is the filter icon on the wrong side?!?
- [x] Make the template look nicer
- [x] `ClrDatagridColumn` currently is closely tied to string filters ([example](https://github.com/amellnik/clarity/blob/topic/datagrid-numeric-filter/src/clr-angular/data/datagrid/datagrid-column.ts#L84))
- [x] Tests, specs
- [x] Docs
- [x] No hardcoded strings in template
- [x] Actual UI design

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8235,7 +8235,7 @@
       "dependencies": {
         "source-map": {
           "version": "0.2.0",
-          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.2.0.tgz",
+          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.2.0.tgz",
           "integrity": "sha1-2rc/vPwrqBm03gO9b26qSBZLP50=",
           "dev": true,
           "optional": true,
@@ -10856,7 +10856,7 @@
     },
     "handle-thing": {
       "version": "1.2.5",
-      "resolved": "http://registry.npmjs.org/handle-thing/-/handle-thing-1.2.5.tgz",
+      "resolved": "https://registry.npmjs.org/handle-thing/-/handle-thing-1.2.5.tgz",
       "integrity": "sha1-/Xqtcmvxpf0W38KbL3pmAdJxOcQ=",
       "dev": true
     },
@@ -24582,7 +24582,7 @@
         },
         "source-map": {
           "version": "0.4.4",
-          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
+          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
           "integrity": "sha1-66T12pwNyZneaAMti092FzZSA2s=",
           "dev": true,
           "requires": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@clr/clarity",
-  "version": "1.0.5",
+  "version": "1.0.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -8235,7 +8235,7 @@
       "dependencies": {
         "source-map": {
           "version": "0.2.0",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.2.0.tgz",
+          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.2.0.tgz",
           "integrity": "sha1-2rc/vPwrqBm03gO9b26qSBZLP50=",
           "dev": true,
           "optional": true,
@@ -10856,7 +10856,7 @@
     },
     "handle-thing": {
       "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/handle-thing/-/handle-thing-1.2.5.tgz",
+      "resolved": "http://registry.npmjs.org/handle-thing/-/handle-thing-1.2.5.tgz",
       "integrity": "sha1-/Xqtcmvxpf0W38KbL3pmAdJxOcQ=",
       "dev": true
     },
@@ -24582,7 +24582,7 @@
         },
         "source-map": {
           "version": "0.4.4",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
+          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
           "integrity": "sha1-66T12pwNyZneaAMti092FzZSA2s=",
           "dev": true,
           "requires": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@clr/clarity",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
+++ b/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
@@ -3,12 +3,11 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import { Observable } from 'rxjs';
 import { ClrDatagridNumericFilterInterface } from '../../interfaces/numeric-filter.interface';
-
 import { DatagridNumericFilterImpl } from './datagrid-numeric-filter-impl';
 import { DatagridPropertyNumericFilter } from './datagrid-property-numeric-filter';
 import { ClrDatagridFilterInterface } from '../../interfaces/filter.interface';
-import { Observable } from 'rxjs';
 
 export default function(): void {
   describe('DatagridNumericFilterImpl', function() {
@@ -57,10 +56,9 @@ export default function(): void {
       expect(fullFilter.accepts(12)).toBe(false);
     });
 
-    // TODO: why does this throw an error on build?
-    // it('exposes state', function() {
-    //   expect(fullFilter.state).toBe(fullFilter);
-    // });
+    it('exposes state', function() {
+      expect(fullFilter.state).toBe(fullFilter);
+    });
 
     it('compares filters', function() {
       let otherFilter = fullFilter;
--- a/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
+++ b/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { ClrDatagridNumericFilterInterface } from '../../interfaces/numeric-filter.interface';
+
+import { DatagridNumericFilterImpl } from './datagrid-numeric-filter-impl';
+import { DatagridPropertyNumericFilter } from './datagrid-property-numeric-filter';
+import { ClrDatagridFilterInterface } from '../../interfaces/filter.interface';
+import { Observable } from 'rxjs';
+
+export default function(): void {
+  describe('DatagridNumericFilterImpl', function() {
+    let fullFilter: DatagridNumericFilterImpl<number>;
+
+    beforeEach(function() {
+      const numericFilter = new TestFilter();
+      fullFilter = new DatagridNumericFilterImpl(numericFilter);
+    });
+
+    it('becomes active when either high or low limits are set', function() {
+      expect(fullFilter.isActive()).toBe(false);
+      fullFilter.high = 10;
+      expect(fullFilter.isActive()).toBe(true);
+      fullFilter.high = null;
+      expect(fullFilter.isActive()).toBe(false);
+      fullFilter.low = 1;
+      expect(fullFilter.isActive()).toBe(true);
+      fullFilter.low = null;
+      expect(fullFilter.isActive()).toBe(false);
+    });
+
+    it('filters numbers that are above the high limit', function() {
+      expect(fullFilter.accepts(11)).toBe(true);
+      fullFilter.high = 10;
+      expect(fullFilter.accepts(11)).toBe(false);
+      expect(fullFilter.accepts(9)).toBe(true);
+      fullFilter.high = null;
+      expect(fullFilter.accepts(11)).toBe(true);
+    });
+
+    it('filters numbers that are below the low limit', function() {
+      expect(fullFilter.accepts(0)).toBe(true);
+      fullFilter.low = 1;
+      expect(fullFilter.accepts(0)).toBe(false);
+      expect(fullFilter.accepts(2)).toBe(true);
+      fullFilter.low = null;
+      expect(fullFilter.accepts(0)).toBe(true);
+    });
+
+    // TODO: why does this throw an error on build?
+    // it('exposes state', function() {
+    //   expect(fullFilter.state).toBe(fullFilter);
+    // });
+
+    it('compares filters', function() {
+      let otherFilter = fullFilter;
+      expect(fullFilter.equals(otherFilter)).toBe(true);
+      // Reference only comparison should be enough for the common case
+      otherFilter = new DatagridNumericFilterImpl(new TestFilter());
+      expect(fullFilter.equals(otherFilter)).toBe(false);
+    });
+
+    describe('with DatagridPropertyNumericFilter', function() {
+      beforeEach(function() {
+        const propFilter = new DatagridPropertyNumericFilter('a.b.c');
+        fullFilter = new DatagridNumericFilterImpl(propFilter);
+      });
+
+      it('exposes state', function() {
+        fullFilter.high = 10;
+        fullFilter.low = null;
+        expect(fullFilter.state).toEqual({ property: 'a.b.c', high: 10, low: null });
+      });
+
+      it('compares filters', function() {
+        let otherFilter: ClrDatagridFilterInterface<any> = fullFilter;
+        expect(fullFilter.equals(otherFilter)).toBe(true);
+        // In the specific case we can compare different filter instances
+        otherFilter = new DatagridNumericFilterImpl(new DatagridPropertyNumericFilter('a.b.c'));
+        expect(fullFilter.equals(otherFilter)).toBe(true);
+        // Incompatible inner function type
+        otherFilter = new DatagridNumericFilterImpl(new TestFilter());
+        expect(fullFilter.equals(otherFilter)).toBe(false);
+        // Incompatible filter object
+        otherFilter = new IncompatibleFilter();
+        expect(fullFilter.equals(otherFilter)).toBe(false);
+      });
+    });
+  });
+}
+
+class TestFilter implements ClrDatagridNumericFilterInterface<number> {
+  accepts(item: number, low: number, high: number) {
+    return (
+      // Make sure the limits are set and not null before checking them
+      low === null || item >= low || (high === null || item <= high)
+    );
+  }
+}
+
+class IncompatibleFilter implements ClrDatagridFilterInterface<number> {
+  accepts(item: number): boolean {
+    return true;
+  }
+
+  changes: Observable<any>;
+
+  isActive(): boolean {
+    return true;
+  }
+}
--- a/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
+++ b/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
@@ -49,6 +49,14 @@ export default function(): void {
       expect(fullFilter.accepts(0)).toBe(true);
     });
 
+    it('only allows values within the range when both filters are set', function() {
+      fullFilter.low = 1;
+      fullFilter.high = 10;
+      expect(fullFilter.accepts(0)).toBe(false);
+      expect(fullFilter.accepts(5)).toBe(true);
+      expect(fullFilter.accepts(12)).toBe(false);
+    });
+
     // TODO: why does this throw an error on build?
     // it('exposes state', function() {
     //   expect(fullFilter.state).toBe(fullFilter);
--- a/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.ts
+++ b/src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Observable } from 'rxjs';
+import { Subject } from 'rxjs';
+import { ClrDatagridFilterInterface } from '../../interfaces/filter.interface';
+import { ClrDatagridNumericFilterInterface } from '../../interfaces/numeric-filter.interface';
+
+export class DatagridNumericFilterImpl<T = any> implements ClrDatagridFilterInterface<T> {
+  constructor(public filterFn: ClrDatagridNumericFilterInterface<T>) {}
+
+  /**
+   * The Observable required as part of the Filter interface
+   */
+  private _changes = new Subject<[number, number]>();
+  // We do not want to expose the Subject itself, but the Observable which is read-only
+  public get changes(): Observable<[number, number]> {
+    return this._changes.asObservable();
+  }
+
+  /**
+   * Internal values and accessor
+   */
+  private _low: number = null;
+  private _high: number = null;
+
+  /**
+   * Common setters for the input values, including individual limits and
+   * both at the same time.  Value is singular to make the interface similar
+   * to the built-in string filter.
+   */
+
+  public get value(): [number, number] {
+    return [this._low, this._high];
+  }
+
+  public set value(vals: [number, number]) {
+    const low = vals[0];
+    const high = vals[1];
+    if (low !== this._low || high !== this._high) {
+      this._low = low;
+      this._high = high;
+      this._changes.next([this._low, this._high]);
+    }
+  }
+
+  public set low(low: number) {
+    if (low !== this._low) {
+      this._low = low;
+      this._changes.next([this._low, this._high]);
+    }
+  }
+
+  public set high(high: number) {
+    if (high !== this._high) {
+      this._high = high;
+      this._changes.next([this._low, this._high]);
+    }
+  }
+
+  public get low() {
+    return this._low;
+  }
+
+  public get high() {
+    return this._high;
+  }
+
+  /**
+   * Indicates if the filter is currently active, (at least one input is set)
+   */
+  public isActive(): boolean {
+    return this._low !== null || this.high !== null;
+  }
+
+  /**
+   * Tests if an item matches a search text
+   */
+  public accepts(item: T): boolean {
+    // We have a filter function in case someone wants to implement a numeric
+    // filter that always passes nulls or similar
+    return this.filterFn.accepts(item, this._low, this._high);
+  }
+}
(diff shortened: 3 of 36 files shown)


amellnik:
Hi @gnomeontherun,

Thanks for the early feedback!  Two further questions:

* Do you think it would be worth generalizing this to other types of range filters or would it be better to stick with numeric values only?  
* Would you be open to adding a `value(): any` method to `ClrDatagridFilterInterface`?  If we expect all filters to expose some indication of their state I that would allow me to get rid of a lot of the `if` blocks to special case the two built-in filters.

src/dev/src/app/datagrid/binding-properties/binding-properties.html
```diff
@@ -16,20 +16,27 @@ <h2>Binding model properties to columns</h2>
...
+  By default, bound columns are assumed to contain string-like contents and the user is presented with the normal string filter.
+  If you know that the contents of the column will be numeric, you can instead use the built-in numeric range filter by adding
+  <code class="clr-code">[clrDgColType]=&quot;'number'&quot;</code>.  You can see an example of this in the ID column.
 </p>
 
 <clr-datagrid>
-  <clr-dg-column>User ID</clr-dg-column>
-  <clr-dg-column [clrDgField]="'name'" [(clrFilterValue)]="nameFilter">Name</clr-dg-column>
   <clr-dg-column [clrDgField]="'creation'">Creation date</clr-dg-column>
+  <clr-dg-column [clrDgField]="'id'" [clrDgColType]="'number'">User ID</clr-dg-column>
```
youdz: I forgot these weren't cleaned up. Do you mind using the "simpler" syntax without `[...]` binding?
For this column, it would be:
```html
<clr-dg-column clrDgField="id" clrDgColType="number">User ID</clr-dg-column>
```
It gets rid of these nasty nested quotes.

src/clr-angular/data/datagrid/index.ts
```diff
@@ -21,11 +21,14 @@ export * from './datagrid-placeholder';
 
 export * from './interfaces/state.interface';
 export * from './enums/sort-order.enum';
+export * from './enums/column-type.enum';
```
youdz: I don't see the consumer using this enum in any of your examples, can we just keep it private?

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -144,6 +153,30 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
...
+  /*
+     * What type is this column?  This defaults to STRING, but can also be
+     * set to NUMBER.  Unsupported types default to STRING. Users can set it
+     * via the [clrDgColType] input by setting it to 'string' or 'number'.
+     */
+
+  private _colType: ClrDatagridColumnType = ClrDatagridColumnType.STRING;
+  public get colType() {
+    if (this._colType === ClrDatagridColumnType.NUMBER) {
+      return 'number';
```
youdz: No offense, but this looks a bit silly to me. You receive a string as an input, convert it to an enum, and then you re-convert it to a string when you use it in your own template? Either just use strings everywhere, or embrace enums and don't convert back to strings.

To be clear, I think the best solution is simply to use a string union type rather than an enum:
```
@Input('clrDgColType') colType: "string" | "number" = "string";
```
No getter or setter needed, and the default works fine. If the consumer gives us anything other than these two strings, AoT compilation would fail with a clear message.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -341,25 +378,38 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
...
   public get filterValue() {
-    return this.filter.value;
+    if (this.filter instanceof DatagridStringFilterImpl || this.filter instanceof DatagridNumericFilterImpl) {
+      return this.filter.value;
+    }
   }
 
   @Input('clrFilterValue')
-  public set updateFilterValue(newValue: string) {
+  public set updateFilterValue(newValue: string | [number, number]) {
```
youdz: Note to self for future updates: we'll need to find a way to type the column so that we can infer the type of this setter based on the `colType` of the column.

Don't worry about it for this PR.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -44,10 +48,15 @@ let nbCount: number = 0;
             <ng-content select="clr-dg-filter, clr-dg-string-filter"></ng-content>
```
youdz: Just realized, you need to update this selector too to add `clr-dg-numeric-filter`. Otherwise you won't project custom filters in the right place.

youdz:
To answer your question of "Why is the filter icon on the wrong side?!?", I looked quickly at our CSS and realized we have some CSS from an external contribution that uses flex order to put the filter last:
```css
clr-dg-filter,
clr-dg-string-filter {
  display: flex;
  order: 99;
  margin-left: auto;
}
```

This doesn't respect our guidelines, styling Angular elements directly, so I'm okay with you stripping it and reordering the DOM so that the DOM order is the same as the display order. Or whatever solution you think is cleanest.

amellnik:
Would it be possible to create a new feature branch for this?  I'm planning on working through the suggestions above this week and it seems like it would be nice to build it on top of https://github.com/vmware/clarity/commit/6310d4b79adad171c167e863d32870b43fc0cbb3.

youdz:
Since you haven't committed anything to it yet, I'll just update the existing topic branch so it's based on latest master. Please make sure you rebase your changes on top of today's master before updating the PR. 👍

amellnik:
I'm working on the documentation and the most natural way to demo the numeric filters often runs into https://github.com/vmware/clarity/issues/2683 since they are wide and you often want to add them to the first column.  Should I go ahead and add them normally or try to work around the issue by reordering columns to put the filters in the middle of the grid?

hippee-lee:
Hey @amellnik - it looks like your changes affect the public Clarity API. You should be able to fix it by running `npm run golden:fix`.

amellnik:
I just noticed that while the internationalization appears to work locally for testing, I might have messed something up since it shows up like this when I build, publish, then use from a different project:  

screen shot 2019-02-27 at 1 47 05 pm

amellnik:
I just discovered one other issue that probably needs to be addressed -- in `clrDgRefresh` the `filter` array looks like this:

screen shot 2019-02-27 at 4 16 50 pm

for a built-in string filter, a built-in numeric filter, and a custom string filter respectively.  We should probably include the value of `clrDgColType` to make it clear what type the built-in filters are.

gnomeontherun:
@amellnik Just checking in, how are things?

amellnik:
@gnomeontherun I will fix the two minor issues I mentioned above, but I'm mostly just waiting on the design.  Is there anything else I should be working on?

We're using this for an internal tool now and haven't ran into any other issues with it.

mathisscott:
@colinreedmiller and @lil-kim 
Could one of you take a look at this to get it unblocked? Thanks.

sdurnov:
@amellnik, given that Clarity filters aren't closing automatically and you can open filters for several columns simultaneously, wouldn't it be better to place fields vertically, so they won't overlap with other filters?

amellnik:
@sdurnov - I don't have any strong feelings about the layout, and would be open to stacking the inputs vertically.  I think the function of the filter is a bit more clear when they are horizontal since many users may be used to seeing ranges written as `[lower_limit, upper_limit]` or similar, but having the wide filters is annoying especially given https://github.com/vmware/clarity/issues/1262#issuecomment-420265815.

amellnik:
image

This now matches @colinreedmiller's design except that it still has a X to close the filter and does show the up/down arrows.  I saw a few examples for removing the arrows in various browsers [here](https://stackoverflow.com/questions/3790935/can-i-hide-the-html5-number-input-s-spin-box) but they seem a bit hacky.  Do we want to do this, and if so, is there a recommended method?

colinreedmiller:
We do not want to attempt to manage the increment / de-increment arrows. They will be provided or not depending on the browser.

Design spec here:

colinreedmiller:
also: This design is for the numeric range implementation only. We are going to hold off on the operator / value version for now in deference to the advanced datagrid filtering currently in the cue: https://github.com/vmware/clarity/issues/1632

amellnik:
I think this now matches the design, let me know if I missed anything or there's anything else I should do!

amellnik:
@gnomeontherun @colinreedmiller Just checking in -- is there anything else I should do on this?

gnomeontherun:
I've started a review, but have not completed it yet. I'm tied up at ng-conf right now but am trying to get it done asap!

src/website/src/app/documentation/demos/datagrid/built-in-filters/built-in-filters.html
```diff
@@ -37,6 +36,18 @@ <h4>String filter</h4>
     entering the exact number of the pokemon he's interested in.
 </p>
 
+<h4>Numeric filter</h4>
+
+<p>
+    The "Wins" column demonstrates a built-in numeric filter.  You provide the function logic and
```
gnomeontherun: Update to

> Another built-in filter is the numeric filter, which allows you to filter a column by a minimum and/or maximum numeric value. The "Wins" column demonstrates the numeric filter.  You provide the function logic and the user can optionally enter high and low limits for elements in the column.  In this case, use a <code class="clr-code">&lt;clr-dg-numeric-filter&gt;</code> component and pass the filter to the <code class="clr-code">[clrDgNumericFilter]</code> property.

src/website/src/app/documentation/demos/datagrid/built-in-filters/built-in-filters.html
```diff
@@ -37,6 +36,18 @@ <h4>String filter</h4>
...
+    the user can optionally enter high and low limits for elements in the column.  In this case,
+    use a <code class="clr-code">&lt;clr-dg-numeric-filter&gt;</code> component and pass the filter
+    to the <code class="clr-code">[clrDgNumericFilter]</code> property.
+</p>
+<clr-code-snippet [clrCode]="examples.numericFilterInterface" clrLanguage="typescript"></clr-code-snippet>
+<!-- TODO: Is this next example really needed? -->
+<!-- <clr-code-snippet [clrCode]="examples.numericFilterInput" clrLanguage="html"></clr-code-snippet> -->
```
gnomeontherun: Yes I think we should include this line, please display it.

src/dev/src/app/datagrid/string-filtering/string-filtering.html
```diff
@@ -30,13 +30,24 @@ <h4>String filter</h4>
...
+<p>
+    The ID column demonstrates a built-in numeric filter.  You provide the function logic and the user can
+    optionally enter high and low limits for elements in the column.  In this case, pass a <code class="clr-code">&lt;clr-dg-numeric-filter&gt;</code> component.
 </p>
 
 <clr-datagrid>
-  <clr-dg-column>User ID</clr-dg-column>
   <clr-dg-column>Name
   </clr-dg-column>
+  <!-- <clr-dg-column [clrDgColType]="'number'" [clrDgField]="'id'">User ID</clr-dg-column> -->
```
gnomeontherun: Clean up.

src/dev/src/app/datagrid/string-filtering/string-filtering.html
```diff
@@ -18,9 +18,9 @@ <h2>Built-in filters</h2>
 <h4>String filter</h4>
```
gnomeontherun: Would you mind doing the same file renames like you did on the website from `string-filtering` to `built-in-filters`?

src/dev/src/app/datagrid/binding-properties/binding-properties.html
```diff
@@ -16,20 +16,27 @@ <h2>Binding model properties to columns</h2>
...
+  <clr-dg-column clrDgField="id" clrDgColType="number">User ID</clr-dg-column>
+  <clr-dg-column clrDgField="name" [(clrFilterValue)]="nameFilter">Name</clr-dg-column>
+  <clr-dg-column clrDgField="pokemon.name">Pokemon</clr-dg-column>
+  <clr-dg-column>Favorite color</clr-dg-column>
 
   <clr-dg-row *clrDgItems="let user of users">
+    <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
     <clr-dg-cell>{{user.id}}</clr-dg-cell>
     <clr-dg-cell>{{user.name}}</clr-dg-cell>
-    <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
```
gnomeontherun: Can you put the `creation` column back in the same place?

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -29,6 +29,8 @@ export class ClrCommonStringsService implements ClrCommonStrings {
   danger = 'Error';
   rowActions = 'Available actions';
   pickColumns = 'Show or hide columns';
+  min = 'Min';
```
gnomeontherun: I think we need to make these more clear `Minimum value` and `Maximum value`.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.ts
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Observable } from 'rxjs';
+import { Subject } from 'rxjs';
```
gnomeontherun: Merge these two imports.

gnomeontherun: Merge these two imports

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.ts
```diff
@@ -0,0 +1,111 @@
...
+    // We have a filter function in case someone wants to implement a numeric
+    // filter that always passes nulls or similar
+    return this.filterFn.accepts(item, this._low, this._high);
+  }
+
+  public get state() {
+    if (this.filterFn instanceof DatagridPropertyNumericFilter) {
+      return {
+        property: this.filterFn.prop,
+        // TODO: Should this return value: [this._low, this._high] instead?
```
gnomeontherun: I think this is good as it is, that gives an explicit property for the consumer to have instead of having to unpack an array.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    it('filters numbers that are below the low limit', function() {
+      expect(fullFilter.accepts(0)).toBe(true);
+      fullFilter.low = 1;
+      expect(fullFilter.accepts(0)).toBe(false);
+      expect(fullFilter.accepts(2)).toBe(true);
+      fullFilter.low = null;
+      expect(fullFilter.accepts(0)).toBe(true);
+    });
```
gnomeontherun: Can we add another test that has both high and low limits set and check the bounds?

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
```diff
@@ -0,0 +1,154 @@
...
+                it prevents NgControlStatus and other directives automatically added by Angular
+                on inputs with NgModel from freaking out because of their host binding changing
+                mid-change detection when the input is destroyed.
+            -->
+            <input #input_low type="number" name="low" [(ngModel)]="low" *ngIf="open"
+                (keyup.enter)="close()" (keyup.escape)="close()" style="width: 78px"
+                [placeholder]="commonStrings.min"/>
+                <span class="datagrid-filter-input-spacer"></span>
+            <input #input_high type="number" name="high" [(ngModel)]="high" *ngIf="open"
+                (keyup.enter)="close()" (keyup.escape)="close()" style="width: 78px"
```
gnomeontherun: Can we rework the styling so there isn't an explicit width set on the input? That would move some CSS into the datagrid file. I would rather see us have something like a flex element that can scale in case popover sizes are ever changed and it surprises me that we don't have the `clr-input` class on here either for the styling.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
```diff
@@ -0,0 +1,154 @@
...
+  providers: [{ provide: CustomFilter, useExisting: DatagridNumericFilter }],
+  template: `
+        <clr-dg-filter [clrDgFilter]="registered" [(clrDgFilterOpen)]="open">
+            <!--
+                Even though this *ngIf looks useless because the filter container already has one,
+                it prevents NgControlStatus and other directives automatically added by Angular
+                on inputs with NgModel from freaking out because of their host binding changing
+                mid-change detection when the input is destroyed.
+            -->
+            <input #input_low type="number" name="low" [(ngModel)]="low" *ngIf="open"
```
gnomeontherun: I believe we'll need to add `[aria-label]="commonStrings.min"` to this and the next input (max for that one of course) for a11y. There is no label otherwise and VoiceOver just says 'Stepper' but without context to what it does.

src/clr-angular/utils/i18n/common-strings.interface.ts
```diff
@@ -85,4 +85,12 @@ export abstract class ClrCommonStrings {
    * Datagrid: pick columns
    */
   pickColumns?: string;
+  /**
+   * Datagrid numeric filter: min
+   */
+  min?: string;
```
gnomeontherun: When you rebase onto `master`, we'll need to add these into the internationalization documentation as well.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -122,6 +130,14 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
     this._sortSubscription.unsubscribe();
   }
 
+  /*
+     * What type is this column?  This defaults to STRING, but can also be
+     * set to NUMBER.  Unsupported types default to STRING. Users can set it
+     * via the [clrDgColType] input by setting it to 'string' or 'number'.
+     */
+
+  @Input('clrDgColType') colType: 'string' | 'number' = 'string';
```
gnomeontherun: This is fine for now, thinking we'd want to setup an enum to track these types better in the future. Maybe make a comment todo, but fine for now.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -319,25 +339,38 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
...
+        this.filter.value = newValue;
+      }
     }
   }
 
-  public set filterValue(newValue: string) {
-    this.updateFilterValue = newValue;
-    this.filterValueChange.emit(this.filter.value);
+  public set filterValue(newValue: string | [number, number]) {
+    if (this.filter instanceof DatagridStringFilterImpl || this.filter instanceof DatagridNumericFilterImpl) {
```
gnomeontherun: Why was this changed?

src/clr-angular/data/datagrid/built-in/filters/datagrid-property-numeric-filter.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
```
gnomeontherun: Noticed we'll need all the copyrights updated for these files to 2019. Current master has a script to do this for you though, so you could wait until rebasing.

src/dev/src/app/datagrid/binding-properties/binding-properties.html
amellnik: I rearranged it because otherwise we run into this bug (https://github.com/vmware/clarity/issues/1262#issuecomment-420265815 and https://github.com/vmware/clarity/issues/2683)

image

It looks like this may take a while to get resolved -- should I add one more additional column in the first position or what?

src/clr-angular/utils/i18n/common-strings.service.ts
amellnik: At the current input widths these placeholders would get cut off.  Is that OK?

amellnik: `Min value` and `Max value` mostly fit if that works:

image

src/clr-angular/data/datagrid/datagrid-column.ts
amellnik: `value` is not part of `ClrDatagridFilterInterface` but is part of the individual string and numeric filters.  Originally I was going to have the state of the numeric filter be called `values` but I switched it to `value` to match.  Should we make this part of the interface?  I assume any future built-in filter will have some sort of internal state that we can refer to with `value`.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
amellnik: I moved the fixed width to the CSS.  I don't know much about design -- should we have breakpoints for this now or is that a TODO?

src/clr-angular/utils/i18n/common-strings.interface.ts
amellnik: Keeping open for the internationalization todo.

src/clr-angular/data/datagrid/built-in/filters/datagrid-property-numeric-filter.ts
amellnik: Cool, holding off for now.

src/dev/src/app/datagrid/binding-properties/binding-properties.html
gnomeontherun: Oh yes, that is something that is being actively addressed. I see why you did it now, and you can leave it.

src/clr-angular/utils/i18n/common-strings.service.ts
gnomeontherun: I can agree with that. It might also depend on the browser.

src/clr-angular/data/datagrid/datagrid-column.ts
gnomeontherun: Ok, makes sense. I think we can leave it as is. We have some tech debt on making filters more generically supportable but that is a separate task!

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
gnomeontherun: No breakpoint at the moment, just would like to get the style attribute out from here.

amellnik:
I think this should be ready for rebasing but let me know if there's anything else I should do.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    it('only allows values within the range when both filters are set', function() {
+      fullFilter.low = 1;
+      fullFilter.high = 10;
+      expect(fullFilter.accepts(0)).toBe(false);
+      expect(fullFilter.accepts(5)).toBe(true);
+      expect(fullFilter.accepts(12)).toBe(false);
+    });
+
+    // TODO: why does this throw an error on build?
+    // it('exposes state', function() {
```
coryrylan: What was the error in the build?

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.ts
```diff
@@ -0,0 +1,110 @@
...
```
coryrylan: nit: remove outdated comment

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      context.clarityDirective.value = [null, 10];
+      expect(filtersInstance.getActiveFilters().length).toBe(1);
+      expect(filtersInstance.getActiveFilters()[0]).toBe(context.clarityDirective.filter);
+    });
+
+    it('registers itself as a CustomFilter provider', function() {
+      expect(context.testComponent.customFilter).toBe(context.clarityDirective);
+    });
+
+    // TODO: Should this count the inputs?
```
coryrylan: I don't think that is necessary, we can use gemini to make a screenshot of the filter to make sure everything is looks correct.

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
```diff
@@ -0,0 +1,154 @@
...
+   * We need the actual input element to automatically focus on it
+   */
+  @ViewChild('input_low') public input: ElementRef;
+
+  /**
+   * We grab the ClrDatagridFilter we wrap to register this StringFilter to it.
+   */
+  @ViewChild(ClrDatagridFilter) public filterContainer: ClrDatagridFilter<T>;
+  ngAfterViewInit() {
+    this.filterContainer.openChanged.subscribe((open: boolean) => {
```
coryrylan: Should unsubscribe in ngOnDestroy

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter-impl.spec.ts
amellnik: There was actually nothing wrong with the test, I missed something in the `state` function.  The test now works.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@clr/clarity",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
```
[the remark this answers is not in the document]
amellnik: I think I actually corrected these now!  (They were left over from a hacky publish to an internal npm proxy so that we could start using the feature).

src/clr-angular/data/datagrid/built-in/filters/datagrid-numeric-filter.ts
amellnik: Like this?  https://github.com/vmware/clarity/pull/2996/files#diff-69d97a05a7a4f27670ea86537ac65355R83

Jinnie: We're usually using the unsubscribe on the Subscription object
```
this.subscription = this.filterContainer.openChanged.subscribe(...
...
ngOnDestroy() {
  this.subscription.unsubscribe();
}
```
and lately we use array of subscriptions even for single subscriptions, like this:
```
....
this.subscriptions.push(this.filterContainer.openChanged.subscribe(...
...
ngOnDestroy() {
  this.subscriptions.forEach(sub => {
    sub.unsubscribe();
  });
}
```
so that it's easy to manage multiple subscriptions if some are added in future.

amellnik: I'll switch it to the array version this evening and also re-run `golden:fix` (why the tests failed).  Thanks!

coryrylan:
This looks good, I think it could be rebased and then once rebased update the copyright dates on the altered files to 2019, as well as documenting the new entries for internationalization.

gnomeontherun:
From a pragmatic standpoint, it might be easiest to just rebase and make a new PR to master and close this out. That way we don't have to do too much git fu to move forward.

amellnik:
Closing in favor of #3397

PR #2998, opened 2019-01-08
Reorder column droppable

Shijir:
This PR solves a problem of activating a correct reorder droppable with a proper amount of drop tolerance. In this current implementation, every column header has two `clr-dg-column-reorder-droppable`s on its right and left sides. 

**This PR contains:**
- Partial implementation of two major services: `ColumnOrdersCoordinatorService` and `ColumnOrderModelService`
- Implementation of `ClrDatagridColumnReorderDroppable`

**The next PR will contain:**
- the business logic of dropping and swapping headers
- the animation of columns changing their positions.

Preview: http://gimme-drop-area.surge.sh/datagrids

### DIFF
--- a/src/clr-angular/data/datagrid/all.spec.ts
+++ b/src/clr-angular/data/datagrid/all.spec.ts
@@ -61,10 +61,10 @@ import WrappedCellSpec from './wrapped-cell.spec';
 import WrappedColumnSpec from './wrapped-column.spec';
 import WrappedRowSpec from './wrapped-row.spec';
 
-describe('Datagrid', function() {
+fdescribe('Datagrid', function() {
   addHelpers();
 
-  describe('Providers', function() {
+  fdescribe('Providers', function() {
     SortProviderSpecs();
     FiltersProviderSpecs();
     PageProviderSpecs();
--- a/src/clr-angular/data/datagrid/all.spec.ts
+++ b/src/clr-angular/data/datagrid/all.spec.ts
@@ -61,10 +61,10 @@ import WrappedCellSpec from './wrapped-cell.spec';
 import WrappedColumnSpec from './wrapped-column.spec';
 import WrappedRowSpec from './wrapped-row.spec';
 
-fdescribe('Datagrid', function() {
+describe('Datagrid', function() {
   addHelpers();
 
-  fdescribe('Providers', function() {
+  describe('Providers', function() {
     SortProviderSpecs();
     FiltersProviderSpecs();
     PageProviderSpecs();
--- a/src/clr-angular/data/datagrid/all.spec.ts
+++ b/src/clr-angular/data/datagrid/all.spec.ts
@@ -23,6 +23,7 @@ import DatagridColumnToggleButtonSpecs from './datagrid-column-toggle-button.spe
 import DatagridColumnToggleSpecs from './datagrid-column-toggle.spec';
 import DatagridColumnSpecs from './datagrid-column.spec';
 import DatagridColumnSeparatorSpecs from './datagrid-column-separator.spec';
+import DatagridColumnReorderDroppableSpecs from './datagrid-column-reorder-droppable.spec';
 import DatagridFilterSpecs from './datagrid-filter.spec';
 import DatagridFooterSpecs from './datagrid-footer.spec';
 import DatagridHideableColumnSpec from './datagrid-hideable-column.model.spec';
@@ -45,6 +46,10 @@ import PageProviderSpecs from './providers/page.spec';
 import SelectionProviderSpecs from './providers/selection.spec';
 import SortProviderSpecs from './providers/sort.spec';
 import TableSizeServiceSpec from './providers/table-size.service.spec';
+import ColumnResizerServiceSpecs from './providers/column-resizer.service.spec';
+import ColumnOrdersCoordinatorServiceSpecs from './providers/column-orders-coordinator.service.spec';
+import ColumnOrderModelServiceSpecs from './providers/column-order-model.service.spec';
+
 import DatagridCellRendererSpecs from './render/cell-renderer.spec';
 import DomAdapterSpecs from '../../utils/dom-adapter/dom-adapter.spec';
 import DatagridHeaderRendererSpecs from './render/header-renderer.spec';
@@ -55,7 +60,6 @@ import DatagridRowRendererSpecs from './render/row-renderer.spec';
 import WrappedCellSpec from './wrapped-cell.spec';
 import WrappedColumnSpec from './wrapped-column.spec';
 import WrappedRowSpec from './wrapped-row.spec';
-import ColumnResizerServiceSpecs from './providers/column-resizer.service.spec';
 
 describe('Datagrid', function() {
   addHelpers();
@@ -71,6 +75,8 @@ describe('Datagrid', function() {
     DisplayModeServiceSpecs();
     TableSizeServiceSpec();
     ColumnResizerServiceSpecs();
+    ColumnOrdersCoordinatorServiceSpecs();
+    ColumnOrderModelServiceSpecs();
   });
   describe('Components', function() {
     DatagridActionBarSpecs();
@@ -79,6 +85,7 @@ describe('Datagrid', function() {
     DatagridFilterSpecs();
     DatagridColumnSpecs();
     DatagridColumnSeparatorSpecs();
+    DatagridColumnReorderDroppableSpecs();
     DatagridItemsSpecs();
     DatagridItemsTrackBySpecs();
     DatagridRowSpecs();
--- a/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
@@ -13,6 +13,7 @@ import { DomAdapter } from '../../utils/dom-adapter/dom-adapter';
 import { MOCK_TABLE_SIZE_PROVIDER, MockTableSizeService } from './providers/table-size.service.mock';
 
 import {
+  destroyMockHeaderEl,
   MOCK_COLUMN_ORDER_MODEL_PROVIDER,
   MockColumnOrderModelService,
   populateMockProps,
@@ -55,10 +56,21 @@ export default function(): void {
 
     afterEach(function() {
       context.fixture.destroy();
+
+      destroyMockHeaderEl(columnOrderModelService.headerEl);
+
+      // the last column model service wouldn't have the next column model so we have to check.
+      if (columnOrderModelService.nextVisibleColumnModel) {
+        destroyMockHeaderEl(columnOrderModelService.nextVisibleColumnModel.headerEl);
+      }
+      // the first column model service wouldn't have the previous column model so we have to check.
+      if (columnOrderModelService.previousVisibleColumnModel) {
+        destroyMockHeaderEl(columnOrderModelService.previousVisibleColumnModel.headerEl);
+      }
     });
 
     it('has column group id', function() {
-      expect(context.clarityDirective.columnOrderDropKey).toBe('dg-mock-group-id');
+      expect(context.clarityDirective.columnGroupId).toBe('dg-mock-group-id');
     });
 
     it('both droppables should have no drop area for draggable from same column', function() {
--- a/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
@@ -83,12 +83,28 @@ export default function(): void {
       expect(context.clarityDirective.dropTolerance).toBe(-1);
     });
 
+    it('right droppable should have no drop area for draggable from next column', function() {
+      const dragEvent = { dragDataTransfer: columnOrderModelService.nextVisibleColumnModel };
+      componentInstance.side = ColumnHeaderSides.RIGHT;
+      context.detectChanges();
+      reorderDroppable.setDropTolerance(dragEvent);
+      expect(context.clarityDirective.dropTolerance).toBe(-1);
+    });
+
     it('left droppable should have proper drop area for draggable from next column', function() {
       const dragEvent = { dragDataTransfer: columnOrderModelService.nextVisibleColumnModel };
       componentInstance.side = ColumnHeaderSides.LEFT;
       context.detectChanges();
       reorderDroppable.setDropTolerance(dragEvent);
-      expect(context.clarityDirective.dropTolerance).toEqual({ left: 200, right: 100 });
+      expect(context.clarityDirective.dropTolerance).toEqual({ left: 0, right: 200 });
+    });
+
+    it('right droppable should have proper drop area for draggable from previous column', function() {
+      const dragEvent = { dragDataTransfer: columnOrderModelService.previousVisibleColumnModel };
+      componentInstance.side = ColumnHeaderSides.RIGHT;
+      context.detectChanges();
+      reorderDroppable.setDropTolerance(dragEvent);
+      expect(context.clarityDirective.dropTolerance).toEqual({ left: 200, right: 0 });
     });
 
     it('far left droppable should have proper drop area for draggable from next column', function() {
@@ -98,23 +114,7 @@ export default function(): void {
       componentInstance.side = ColumnHeaderSides.LEFT;
       context.detectChanges();
       reorderDroppable.setDropTolerance(dragEvent);
-      expect(context.clarityDirective.dropTolerance).toEqual({ left: 0, right: 100 });
-    });
-
-    it('right droppable should have no drop area for draggable from next column', function() {
-      const dragEvent = { dragDataTransfer: columnOrderModelService.nextVisibleColumnModel };
-      componentInstance.side = ColumnHeaderSides.RIGHT;
-      context.detectChanges();
-      reorderDroppable.setDropTolerance(dragEvent);
-      expect(context.clarityDirective.dropTolerance).toBe(-1);
-    });
-
-    it('right droppable should have proper drop area for draggable from previous column', function() {
-      const dragEvent = { dragDataTransfer: columnOrderModelService.previousVisibleColumnModel };
-      componentInstance.side = ColumnHeaderSides.RIGHT;
-      context.detectChanges();
-      reorderDroppable.setDropTolerance(dragEvent);
-      expect(context.clarityDirective.dropTolerance).toEqual({ left: 100, right: 150 });
+      expect(context.clarityDirective.dropTolerance).toEqual({ left: 0, right: 200 });
     });
 
     it('far right droppable should have proper drop area for draggable from previous column', function() {
@@ -124,7 +124,7 @@ export default function(): void {
       componentInstance.side = ColumnHeaderSides.RIGHT;
       context.detectChanges();
       reorderDroppable.setDropTolerance(dragEvent);
-      expect(context.clarityDirective.dropTolerance).toEqual({ left: 100, right: 0 });
+      expect(context.clarityDirective.dropTolerance).toEqual({ left: 200, right: 0 });
     });
 
     it('gives dropLine height of datagrid table if showHighlight is called', function() {
--- a/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
@@ -61,18 +61,6 @@ export default function(): void {
       expect(context.clarityDirective.columnOrderDropKey).toBe('dg-mock-group-id');
     });
 
-    it('can return width of its own column', function() {
-      expect(context.clarityDirective.headerWidth).toBe(200);
-    });
-
-    it('can return width of previous column', function() {
-      expect(context.clarityDirective.previousVisibleHeaderWidth).toBe(400);
-    });
-
-    it('can return width of next column', function() {
-      expect(context.clarityDirective.nextVisibleHeaderWidth).toBe(300);
-    });
-
     it('both droppables should have no drop area for draggable from same column', function() {
       const dragEvent = { dragDataTransfer: columnOrderModelService };
 
--- a/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component } from '@angular/core';
+import { ClrDatagridColumnReorderDroppable, ColumnHeaderSides } from './datagrid-column-reorder-droppable';
+import { TestContext } from '../../utils/testing/helpers.spec';
+import { ColumnOrderModelService } from './providers/column-order-model.service';
+import { TableSizeService } from './providers/table-size.service';
+import { DomAdapter } from '../../utils/dom-adapter/dom-adapter';
+import { MOCK_TABLE_SIZE_PROVIDER, MockTableSizeService } from './providers/table-size.service.mock';
+
+import {
+  MOCK_COLUMN_ORDER_MODEL_PROVIDER,
+  MockColumnOrderModelService,
+  populateMockProps,
+} from './providers/column-order-model.service.mock';
+
+@Component({
+  template: `<clr-dg-column-reorder-droppable [side]="side"></clr-dg-column-reorder-droppable>`,
+})
+class TestComponent {
+  side: ColumnHeaderSides;
+}
+
+export default function(): void {
+  describe('ClrDatagridColumnReorderDroppable component', function() {
+    let context: TestContext<ClrDatagridColumnReorderDroppable, TestComponent>;
+    let componentInstance: TestComponent;
+    let reorderDroppable: ClrDatagridColumnReorderDroppable;
+    let columnOrderModelService: MockColumnOrderModelService;
+    let tableSizeService: MockTableSizeService;
+
+    beforeEach(function() {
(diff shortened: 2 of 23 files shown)


src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,118 @@
...
+      return 0;
+    }
+  }
+
+  get previousVisibleHeaderWidth(): number {
+    if (this.columnOrderModel.previousVisibleColumnModel) {
+      const headerEl = this.columnOrderModel.previousVisibleColumnModel.headerEl;
+      return this.domAdapter.clientRect(headerEl).width;
+    } else {
+      return 0;
```
gnomeontherun: Your tests don't try this path in the width methods here.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -34,7 +34,12 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
     this.droppableEl = this.el.nativeElement;
   }
 
-  private isDraggableMatch: boolean = false;
+  private _isDraggableMatch: boolean = false;
+
+  get isDraggableMatch(): boolean {
```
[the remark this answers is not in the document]
Shijir: Oh this is actually what Travis caught for me. There was a compile error because it's used in the template even though it's private. So I had to make it public.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,118 @@
...
+        this.dropTolerance = -1; // a negative drop tolerance means no drop area
+      }
+    } else {
+      // A reorder droppable shouldn't have a droppable area for the draggable header from the same column.
+      // So if the draggable is from the same header, disable the dropTolerance
+      this.dropTolerance = -1; // a negative drop tolerance means no drop area
+    }
+  }
+
+  get headerWidth(): number {
```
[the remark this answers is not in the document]
Shijir: Are they are compelling reasons to make them private though? They are read-only currently. I think making read-only properties private is kinda overkill, and the tests would become unnecessarily convoluted/complicated.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,118 @@
...
+
+  public setDropTolerance(event: any): void {
+    const draggedFrom: number = event.dragDataTransfer.flexOrder;
+
+    if (draggedFrom < this.columnOrderModel.flexOrder) {
+      // if the dragged header is from the left side, the droppable at the right side in the current header
+      // would get a proper dropTolerance value and the left side one would be disabled with value of -1.
+
+      if (this.side === ColumnHeaderSides.Right) {
+        this.dropTolerance = { left: this.headerWidth / 2, right: this.nextVisibleHeaderWidth / 2 };
```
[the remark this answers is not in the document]
Shijir: I think it works as you are describing here.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -354,7 +378,17 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
...
    */
-  public hideable: DatagridHideableColumnModel;
+
+  private _hideable: DatagridHideableColumnModel;
+
+  get hideable(): DatagridHideableColumnModel {
+    return this._hideable;
+  }
+
+  set hideable(value: DatagridHideableColumnModel) {
```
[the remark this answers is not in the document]
Shijir: Will write tests for it 😛

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,118 @@
...
+  public get columnOrderDropKey(): string {
+    return this.columnOrderModel.columnGroupId;
+  }
+
+  public dropTolerance: any;
+
+  // Each column headers will have this clr-dg-column-reorder-droppable component on each of its sides.
+  // We need a way to distinguish which side of the header that this droppable is on.
+
+  @Input('side') side: ColumnHeaderSides;
```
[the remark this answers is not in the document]
Shijir: This is covered in the tests. But this had to be tested indirectly because assigning a value alone to this property doesn't make any direct changes to the component itself as the value is used to calculate the drop tolerance.

src/clr-angular/data/datagrid/providers/column-orders-coordinator.service.ts
```diff
@@ -0,0 +1,35 @@
...
+let nbColumnGroup = 0;
+
+/**
+ * This service is responsible for:
+ * 1. Sharing order model data across headers and cells
+ * 2. Broadcasting order model update
+ */
+
+@Injectable()
+export class ColumnOrdersCoordinatorService {
```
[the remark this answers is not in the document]
Shijir: I was actually thinking about the same question. Perhaps we need to create a common model for the state of each column? I am curious about everyone's take on it.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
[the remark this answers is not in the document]
gnomeontherun: When I tested it, you have to drag the column so the whole column is beyond the edge. If I have three columns, say they are 100px wide each, if I drag the middle one to the right or left I have to go 101px to the right or left before I can drop. That means the column is actually being dragged past the boundaries of the datagrid at that point. I thought before we had a demo of it where it became droppable after you dragged the column over half way over the column. In this example dragging 51px to the right or left and letting go would reorder, rather than having to go 101px.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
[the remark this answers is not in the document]
gnomeontherun: I think we should avoid having any public methods that aren't necessary, and making it public just for testing isn't a good reason. I understand its harder to test, but not dramatically if you create a component with 3 columns with dynamic widths you can manipulate during the test.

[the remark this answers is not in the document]
Shijir: These are not methods though, but just read-only getter and constant properties. As such, trying to set or manipulate their values would throw compile errors.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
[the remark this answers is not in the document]
Shijir: > I thought before we had a demo of it where it became droppable after you dragged the column over half way over the column

Yes, it currently works just like that. It's only enough to drag the column only over half way over its neighboring columns.

Take a look into the demo: http://gimme-drop-area.surge.sh/datagrids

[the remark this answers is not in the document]
gnomeontherun: That is not what I'm seeing, I have to drag past the edge of the column like you see in the screenshots below.

drag-columns

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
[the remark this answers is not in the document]
gnomeontherun: They are public properties, that is the concern. If it becomes public, then its part of our API (hence the golden file changes). If it isn't required to be public, we should have it private.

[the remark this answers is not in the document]
Shijir: Sorry, I see your point of view now. I agree with you that we shouldn't expose internally used properties as public. Hence, I've moved these properties to the internal service class: https://github.com/vmware/clarity/pull/2998/files#diff-cbf32820148eb4c8057355807c27ec09R94

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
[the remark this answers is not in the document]
Shijir: In the first screenshot, the midpoint of the column being dragged hasn't yet passed the midpoint of the column that it's hovering over. Also, in the case of dragging a column header over a neighboring column header, we don't activate re-order droppable between them. So as shown in the first screenshot, until the dragged header doesn't pass the midpoint of the neighboring header, the re-order droppable won't be activated. 

In the second screenshot, the midpoint of the column is dragged a bit more to the right and has actually passed the midpoint of the column that it's hovering over. This is for the purpose of not activating re-order droppables too early or too eagerly, and users should indicate clearly which crack between columns they want to drop the column whilst still offering them ample space of drop tolerance. And this is implemented based on the discussions I had with the UX designers during POC.

If we actually divide the width by 1.5, the drop tolerance areas of right and left reorder droppables would overlap on each other. 

However, I think it's still a good idea to try different ways of activating reorder-droppables and request people to chime in on them again and pick the one that offers the best experience.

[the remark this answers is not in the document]
Shijir: @gnomeontherun what do you think about this one? The reorder droppable are now activated a bit earlier than the previous one. http://reorder-droppable.surge.sh/datagrids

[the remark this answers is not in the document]
gnomeontherun: Yes this is more what I was thinking, when we say midpoint I think we're not being clear. I think the edge of the column being dragged should become droppable soon after reaching the midpoint of the next/prev column. In other words, like your demo here, so maybe 1.6?

src/clr-angular/data/datagrid/all.spec.ts
```diff
@@ -55,12 +60,11 @@ import DatagridRowRendererSpecs from './render/row-renderer.spec';
 import WrappedCellSpec from './wrapped-cell.spec';
 import WrappedColumnSpec from './wrapped-column.spec';
 import WrappedRowSpec from './wrapped-row.spec';
-import ColumnResizerServiceSpecs from './providers/column-resizer.service.spec';
 
-describe('Datagrid', function() {
+fdescribe('Datagrid', function() {
```
youdz: You f'd up.

src/clr-angular/data/datagrid/all.spec.ts
```diff
@@ -55,12 +60,11 @@ import DatagridRowRendererSpecs from './render/row-renderer.spec';
...
   addHelpers();
 
-  describe('Providers', function() {
+  fdescribe('Providers', function() {
```
youdz: Forgot the f here too.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,96 @@
...
+    private tableSizeService: TableSizeService,
+    private columnOrderModel: ColumnOrderModelService,
+    private renderer: Renderer2
+  ) {}
+
+  public get columnOrderDropKey(): string {
+    return this.columnOrderModel.columnGroupId;
+  }
+
+  public dropTolerance: any;
```
youdz: `public dropTolerance: -1 | ClrDropToleranceInterface

src/clr-angular/data/datagrid/providers/column-orders-coordinator.service.ts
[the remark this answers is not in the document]
youdz: I think a centralized service for columns and their states would be good, but we don't even have a model for the columns. So we can start to set it up here, but it'll start with none of the information you need and will grow organically from there. Which can turn into a mess.

@Shijir: you can try to create the model if you want. The main hurdle is going to be merging the model with the hideable column one, which is overdue for a cleanup. The column width/order should be much easier.

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -189,4 +194,16 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
...
   }
+
+  private setHeaderOrders(): void {
+    this.headers.forEach((header, index) => {
+      // set initial flex order
+      header.setFlexOrder(index);
+    });
+
+    // set orders array with headers ColumnOrder
+    this.columnOrderCoordinatorService.orderModels = this.headers.map((header, index) => {
```
youdz: You're only setting `ColumnOrdersCoordinatorService.orderModels` on initialization. This breaks for dynamic columns.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  get columnGroupId() {
+    return this.columnOrderCoordinatorService.columnGroupId;
+  }
+
+  get isAtFirst(): boolean {
+    return this.flexOrder === 0;
+  }
+
+  get isAtEnd(): boolean {
+    return this.flexOrder === this.columnOrderCoordinatorService.orderModels.length - 1;
```
youdz: What if the last column is hidden? Then the one before last is visually the last one. Shouldn't `isAtEnd` account for that?

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
```diff
@@ -0,0 +1,107 @@
...
+    // find the model with the smallest flexorder value from nextVisibleColumnModels
+    return nextVisibleColumnModels.reduce((smallestFlexOrderModel, currentModel) => {
+      if (smallestFlexOrderModel.flexOrder > currentModel.flexOrder) {
+        return currentModel;
+      }
+      return smallestFlexOrderModel;
+    });
+  }
+
+  get previousVisibleColumnModel(): ColumnOrderModelService {
```
youdz: Nitpick: that's a bit of code duplication with the `nextVisibleColumnModel` method. If you want a fun little exercise, try to refactor this going a bit more functional in order to have less code and no duplication. 😉

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
```diff
@@ -0,0 +1,107 @@
...
+    return previousVisibleColumnModels.reduce((largestFlexOrderModel, currentModel) => {
+      if (largestFlexOrderModel.flexOrder < currentModel.flexOrder) {
+        return currentModel;
+      }
+      return largestFlexOrderModel;
+    });
+  }
+
+  get headerWidth(): number {
+    return this.headerEl ? this.domAdapter.clientRect(this.headerEl).width : 0;
```
youdz: Making this one and the following two a lazy getter is risky for performance. Right now I can see it's called on `dragStart` so it's not that often, but you are re-computing 3 `clientRect`s for every single column at the same time. Depending on the browser and the size of the datagrid, this could get a bit slow.

src/clr-angular/data/datagrid/providers/column-order-model.service.mock.ts
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { ColumnOrderModelService } from './column-order-model.service';
+
+export class MockColumnOrderModelService {
```
youdz: `export class MockColumnOrderModelService implements ColumnOrderModelService`
It will ensure we keep the mock in sync with the actual service in case of future updates.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -91,6 +99,22 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
...
+
+  public get rightReorderDroppable() {
+    return ColumnHeaderSides.RIGHT;
+  }
+
+  public get columnDropData() {
+    return this.columnOrderModel;
+  }
+
+  public get columnOrderDropKey() {
```
youdz: Naming nitpick: you're using group/groupId everywhere, "dropKey" is inconsistent.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,96 @@
...
 })
-export class ClrDatagridColumnReorderDroppable {}
+export class ClrDatagridColumnReorderDroppable {
+  constructor(
+    private tableSizeService: TableSizeService,
+    private columnOrderModel: ColumnOrderModelService,
+    private renderer: Renderer2
+  ) {}
+
+  public get columnOrderDropKey(): string {
```
youdz: Same "dropKey" naming here.

src/clr-angular/data/datagrid/providers/column-order-model.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      columnOrderModelServicePrev.headerEl = mockHeaderEl(400, 40);
+
+      columnOrderModelServiceNext.flexOrder = 2;
+      columnOrderModelServiceNext.headerEl = mockHeaderEl(300, 40);
+
+      columnOrdersCoordinatorService.orderModels.push(columnOrderModelService);
+      columnOrdersCoordinatorService.orderModels.push(columnOrderModelServicePrev);
+      columnOrdersCoordinatorService.orderModels.push(columnOrderModelServiceNext);
+    });
+
```
youdz: You're missing an `afterEach()` to clean up the elements you created with `mockHeaderEl()`.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+      populateMockProps(columnOrderModelService.previousVisibleColumnModel, 'dg-mock-group-id', 0, 400);
+      populateMockProps(columnOrderModelService, 'dg-mock-group-id', 1, 200);
+      populateMockProps(columnOrderModelService.nextVisibleColumnModel, 'dg-mock-group-id', 2, 300);
+
+      context.detectChanges();
+    });
+
+    afterEach(function() {
+      context.fixture.destroy();
```
youdz: You are calling `populateMockProps` in your `beforeEach()`, which create DOM elements that Angular doesn't know about. You need to remove them yourself in this `afterEach()`.

src/clr-angular/data/datagrid/render/main-renderer.ts
Shijir: Yes, I have implemented a method that responds and addresses dynamic columns in the next upcoming PR, which also includes changes that actually reorders columns.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: I have changed the `isAtEnd` property to `isLast` in the next PR. I have also wrote another property called `isLastVisible` that basically addresses this issue you are raising here. And this property needed to figure out whether it needs to project a separator in a column header or not (as we know the last visible header shouldn't have a separator in itself).  I haven't really needed to implement `isFirstVisible` as I haven't faced a need to do so 🤔

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: ok will try out a few things :)

src/clr-angular/data/datagrid/providers/column-order-model.service.spec.ts
Shijir: Now cleaning up the elements in `afterEach()`.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.spec.ts
Shijir: Now removing the elements in `afterEach()`.

src/clr-angular/data/datagrid/datagrid-column.ts
Shijir: Changed to `columnGroupId` everywhere.

src/clr-angular/data/datagrid/providers/column-order-model.service.mock.ts
Shijir: `ColumnOrderModelService` is a class though. We could only implement interfaces though, no? Extending a class wouldn't serve the purpose you are describing here, I think.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: Did a bit of refactoring here 😛 Please take a look.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
Shijir: Updated accordingly.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: oh, I might have misunderstood your comment here. I could do a bit more than the current optimization I did.

src/clr-angular/data/datagrid/providers/column-order-model.service.mock.ts
youdz: Yes, you can use any class or or abstract class as an interface if you want.
So the point here really is to `implement`, not `extend`. It just ensures that if we add a new method to the `ColumnOrderModelService`, the mock will fail to compile because it doesn't implement all the methods of the original service. Failing at compile time is way better than running 15 minutes of test to see it fail with an obscure error message.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
youdz: That's better, but you went waaaay too flexible. Remember, this helper method you wrote is private so it only really needs to work for your two use case.

I would have gone for something like this:
```typescript
private sortAndFindNext(desc = false): ColumnOrderModelService | null {
  const sortedVisibleColumnModels = this.columnOrderCoordinatorService.orderModels
    .filter(model => !model.isHidden)
    .sort(model => model.flexOrder);
  const index = sortedVisibleColumnModels.findIndex(this);
  if (index >= 0 && index < sortedVisibleColumnModels.length - 1) {
    return sortedVisibleColumnModels[index+1];
  } else {
    return null;
  }
}

get nextVisibleColumnModel(): ColumnOrderModelService {
  return this.sortAndFindNext(false);
}

get previousVisibleColumnModel(): ColumnOrderModelService {
  return this.sortAndFindNext(true);
}
```

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
youdz: I get that you can't store a width forever. But in the same computation, aren't you re-computing every header width at least twice? Once for the drop tolerance of each side.

src/clr-angular/data/datagrid/datagrid-column-reorder-droppable.ts
```diff
@@ -4,12 +4,97 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Component } from '@angular/core';
+import { Component, Input, Renderer2 } from '@angular/core';
+import { TableSizeService } from './providers/table-size.service';
+import { ColumnOrderModelService } from './providers/column-order-model.service';
+import { ColumnHeaderSides } from './enums/header-sides.enum';
+import { ClrDropToleranceInterface } from '@clr/angular';
```
gnomeontherun: Fix import path.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: I see. Yes, using `.sort()` is smart! I have opted for the one below:
```ts
 private findAdjacentVisibleModel(prev = false): ColumnOrderModelService {
    const filteredVisibleColumnModels = this.columnOrderCoordinatorService.orderModels
      .filter(model => (!model.isHidden && prev ? model.flexOrder < this.flexOrder : model.flexOrder > this.flexOrder))
      .sort(model => model.flexOrder);

    return prev ? filteredVisibleColumnModels[filteredVisibleColumnModels.length - 1] : filteredVisibleColumnModels[0];
  }
```

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
Shijir: Once we compute the width of a header, we recycle that for both sides' droppable.

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
```diff
@@ -47,40 +47,20 @@ export class ColumnOrderModelService {
...
-        return currentModel;
-      }
-      return reducedModel;
-    });
+  private findAdjacentVisibleModel(prev = false): ColumnOrderModelService {
+    const filteredVisibleColumnModels = this.columnOrderCoordinatorService.orderModels
+      .filter(model => (!model.isHidden && prev ? model.flexOrder < this.flexOrder : model.flexOrder > this.flexOrder))
+      .sort(model => model.flexOrder);
+
+    return prev ? filteredVisibleColumnModels[filteredVisibleColumnModels.length - 1] : filteredVisibleColumnModels[0];
```
youdz: Doesn't this crash when getting next of the last column or prev of the first column?

src/clr-angular/data/datagrid/providers/column-order-model.service.ts
```diff
@@ -47,40 +47,20 @@ export class ColumnOrderModelService {
...
-
-    return filteredVisibleColumnModels.reduce((reducedModel, currentModel) => {
-      if (reduceCondition(reducedModel, currentModel)) {
-        return currentModel;
-      }
-      return reducedModel;
-    });
+  private findAdjacentVisibleModel(prev = false): ColumnOrderModelService {
+    const filteredVisibleColumnModels = this.columnOrderCoordinatorService.orderModels
+      .filter(model => (!model.isHidden && prev ? model.flexOrder < this.flexOrder : model.flexOrder > this.flexOrder))
```
youdz: `&&` takes precedence over the ternary operator `? :`, as far as I know. So this doesn't do what you want at all, you need to:
```typescript
!model.isHidden && (prev ? model.flexOrder < this.flexOrder : model.flexOrder > this.flexOrder)
```

Shijir: Fixed :p

PR #3005, opened 2019-01-08
clarity.design domain cutover fixes

gnomeontherun:
Fixes the analytics tracking, updates any references to domain URL, and points to 1.0.0 icons library.


src/website/src/app/documentation/documentation.component.html
```diff
@@ -3,7 +3,7 @@
     <section class="sidenav-content">
         <version-switcher></version-switcher>
 
-        <a class="nav-link" routerLink="documentation" routerLinkActive="active"
+        <a class="nav-link" routerLink="/documentation" routerLinkActive="active"
            [routerLinkActiveOptions]="{exact:true}">Component Status</a>
 
         <a class="nav-link" routerLink="get-started" routerLinkActive="active"
```
youdz: Any reason why `/documentation` gets a root slash, but `get-started` and the other links don't?

gnomeontherun: I believe since /documentation is a lazy loaded, but also root route and the rest are relative to it.

PR #3025, opened 2019-01-11
Topic/smart popover

hippee-lee:
These changes are all from a rebase of master onto the topic branch. It brings it up to date with master. I'm doing this first b/c I mistakenly rebased my upcoming PR with the `ClrSmartOpen` directive from master.


hippee-lee:
Ok, this clearly isn't going to work like I thought it would. 
I'll use a different approach.

PR #3050, opened 2019-01-16
[NG] fix input reset on datepicker/refactor

coryrylan:
When using clrDate input, setting input to null would not clear the input.

closes vmware#3012

### DIFF
--- a/src/clr-angular/forms/datepicker/date-input.ts
+++ b/src/clr-angular/forms/datepicker/date-input.ts
@@ -25,7 +25,7 @@ import {
   ViewContainerRef,
 } from '@angular/core';
 import { NgControl } from '@angular/forms';
-import { filter } from 'rxjs/operators';
+import { filter, switchMap } from 'rxjs/operators';
 
 import { FocusService } from '../common/providers/focus.service';
 import { WrappedFormControl } from '../common/wrapped-control';
@@ -37,6 +37,14 @@ import { DateNavigationService } from './providers/date-navigation.service';
 import { DatepickerEnabledService } from './providers/datepicker-enabled.service';
 import { IS_NEW_FORMS_LAYOUT } from '../common/providers/new-forms.service';
 import { DatepickerFocusService } from './providers/datepicker-focus.service';
+import { datesAreEqual } from './utils/date-utils';
+import { of } from 'rxjs';
+
+// There are four ways the datepicker value is set
+// 1. Value set by user typing into text input as a string ex: '01/28/2015'
+// 2. Value set explicitly by Angular Forms APIs as a string ex: '01/28/2015'
+// 3. Value set by user via datepicker UI as a Date Object
+// 4. Value set via `clrDate` input as a Date Object
 
 @Directive({
   selector: '[clrDate]',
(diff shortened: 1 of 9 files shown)


src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -37,6 +37,14 @@ import { DateNavigationService } from './providers/date-navigation.service';
 import { DatepickerEnabledService } from './providers/datepicker-enabled.service';
 import { IS_NEW_FORMS_LAYOUT } from '../common/providers/new-forms.service';
 import { DatepickerFocusService } from './providers/datepicker-focus.service';
+import { datesAreEqual } from './utils/date-utils';
+import { of } from 'rxjs';
```
coryrylan: This should be moved to the top group of imports

src/clr-angular/forms/datepicker/utils/date-utils.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect(getDay(2000, 1, 1)).toBe(2);
+    expect(getDay(2000, 1, 2)).not.toBe(2);
+  });
+
+  it('should determine if two dates are equal', () => {
+    const date1 = new Date(2000, 1, 1);
+    const date2 = new Date(2000, 1, 1);
+    const date3 = new Date(2000, 1, 2);
+
+    expect(datesAreEqual(date1, date2)).toBe(true);
```
youdz: Nitpick: you have the `if` for it in the code, can you just add an `expect(datesAreEqual(date1, null)).toBe(false);`?

src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -87,268 +93,180 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
-    this.processInitialInputs();
     this.setFormLayout();
-  }
+    this.populateServicesFromContainerComponent();
 
-  ngAfterViewInit() {
-    this.writeInitialInputFromUserInputField();
+    this.subscriptions.push(
+      this.listenForUserSelectedDayChanges(),
+      this.listenForControlValueChanges(),
```
youdz: 👍

src/clr-angular/forms/datepicker/utils/date-utils.ts
```diff
@@ -41,3 +41,15 @@ export function parseToFourDigitYear(year: number): number {
   }
   return result;
 }
+
+export function datesAreEqual(date1: Date, date2: Date) {
+  if (date1 && date2) {
```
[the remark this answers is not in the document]
coryrylan: I lean towards it should throw to prevent the function from masking a mistake, ex I pass in a string/number accidentally and still got a value back as if nothing was wrong. What does the team/codebase tend to lean towards? Throw with invalid types or defend/check against invalid types on params? Actually makes me kinda question why I had the if block to begin with... I'll double check that.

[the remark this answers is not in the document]
gnomeontherun: I'd like it to just return `false` in this case at least, since this is really a runtime check. If the developer explicitly calls it incorrectly TS will warn them, but we can't always be 100% certain of the values passed into this function.

Issue #2846, opened 2018-11-15
[datagrid] expose a "state" method on custom filters to simplify clDgRefresh


awdyson:
## Summary

* What is the change?
  - `ClrDatagridFilterInterface` gets an optional `state` (or whatever) method with a `{ property: string; value: any }` return type
  - `StateProvider` checks for the existence of this method before/instead-of returning the whole custom filter class
* Why should it go in Clarity?
It'd make dealing with the refresh output easier.
* Does this change impact existing behaviors? If so how?
Not that I see on quick inspection.
* If this change introduces a new behavior, is this behavior accessible?

## Use case

Here's my custom filter -- the template is basically `<checkbox *ngFor="options">`:
```
export class ClrDgCheckFilterComponent implements ClrDatagridFilterInterface<object> {
  @Input() options: Option[] = [];
  @Input() property = '';

  changes = new EventEmitter<true>();
  selected = new Set();

  // My request is that StateProvider knows to call this method if it exists
  get state() {
    if (!this.dirty) return null;
    return { propery: this.property, value: Array.from(this.selected) };
  }

  private get dirty() {
    return !!this.selected.size && this.selected.size !== this.options.length;
  }

  constructor(private dgFilter: ClrDatagridFilter) {
    this.dgFilter.setFilter(this);
  }

  accepts(entry: object) {
    if (!this.dirty) return true;

    const selected = this.selected.has(deepGet(entry, this.property));
    return selected;
  }

  isActive() {
    return this.dirty;
  }

  onChange({ val }: Option) {
    this.selected.has(val) ? this.selected.delete(val) : this.selected.add(val);
    this.changes.emit(true);
  }
}
```

Here's the current output from a built-in filter and this filter:
![image](https://user-images.githubusercontent.com/5612248/48573991-d5056480-e8c2-11e8-9611-d90d0c019d51.png)
Ideally, I'd like to be able to have the output from my filter look just like the built-in. @amellnik if https://github.com/vmware/clarity/issues/2513 requires a more complex representation, I'd love to be able to keep in step with that as well.

## Workarounds

Yep, handle the transformation in a service you provide to the component. Also open to suggestions 😄

amellnik:
By property you mean the property of the row object associated with that column, right?  If I'm understanding this correctly, this partially solves https://github.com/vmware/clarity/issues/2513.

awdyson:
[...]

PR #3066, opened 2019-01-23
[NG] Expose a "state" property on custom filters (#2846)

Jinnie:
Also exposed an "equals" method, for future needs.

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -349,9 +349,11 @@ export declare class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<
     toggle(): void;
 }
 
-export interface ClrDatagridFilterInterface<T> {
+export interface ClrDatagridFilterInterface<T, S = any> {
     changes: Observable<any>;
+    state?: S;
     accepts(item: T): boolean;
+    equals?(other: ClrDatagridFilterInterface<any, any>): boolean;
     isActive(): boolean;
 }
 
@@ -471,10 +473,7 @@ export declare enum ClrDatagridSortOrder {
 }
 
 export interface ClrDatagridStateInterface<T = any> {
-    filters?: ({
-        property: string;
-        value: string;
-    } | ClrDatagridFilterInterface<T>)[];
+    filters?: any[];
     page?: {
         from?: number;
         to?: number;
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -351,9 +351,9 @@ export declare class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<
 
 export interface ClrDatagridFilterInterface<T, S = any> {
     changes: Observable<any>;
-    state?: S;
+    readonly state?: S;
     accepts(item: T): boolean;
-    equals?(other: ClrDatagridFilterInterface<any, any>): boolean;
+    equals?(other: ClrDatagridFilterInterface<T, any>): boolean;
     isActive(): boolean;
 }
 
--- a/src/clr-angular/data/datagrid/built-in/filters/datagrid-string-filter-impl.spec.ts
+++ b/src/clr-angular/data/datagrid/built-in/filters/datagrid-string-filter-impl.spec.ts
@@ -6,6 +6,9 @@
 import { ClrDatagridStringFilterInterface } from '../../interfaces/string-filter.interface';
 
 import { DatagridStringFilterImpl } from './datagrid-string-filter-impl';
+import { DatagridPropertyStringFilter } from './datagrid-property-string-filter';
+import { ClrDatagridFilterInterface } from '../../interfaces/filter.interface';
+import { Observable } from 'rxjs';
 
 export default function(): void {
   describe('DatagridStringFilterImpl', function() {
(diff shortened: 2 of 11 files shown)


golden/clr-angular.d.ts
```diff
@@ -471,10 +473,7 @@ export declare enum ClrDatagridSortOrder {
 }
 
 export interface ClrDatagridStateInterface<T = any> {
-    filters?: ({
```
youdz: As discussed, this is a bummer but I'm not sure we can do better right now. 😞

golden/clr-angular.d.ts
```diff
@@ -349,9 +349,11 @@ export declare class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<
     toggle(): void;
 }
 
-export interface ClrDatagridFilterInterface<T> {
+export interface ClrDatagridFilterInterface<T, S = any> {
     changes: Observable<any>;
+    state?: S;
     accepts(item: T): boolean;
+    equals?(other: ClrDatagridFilterInterface<any, any>): boolean;
```
youdz: Thinking about it, I think we can get away with `ClrDatagridFilterInterface<T, any>`. The datagrid items are the same for all filters in that datagrid.

src/clr-angular/data/datagrid/datagrid-filter.spec.ts
```diff
@@ -58,6 +58,15 @@ export default function(): void {
         filter.active = false;
         expect(component.active).toEqual(false);
       });
+
+      it('compares filters', function() {
+        const testFilter = new TestCustomStateFilter('test');
+        const anotherTestFilter = new TestCustomStateFilter('test');
+        const nonTestFilter = new TestCustomStateFilter('nonTest');
+        expect(testFilter.equals(anotherTestFilter)).toBeTruthy();
```
youdz: I'm sorry, I don't understand this test. You are not testing any code from Clarity here, you're creating test objects and you're calling their method directly, that's not accomplishing anything.

src/clr-angular/data/datagrid/interfaces/filter.interface.ts
```diff
@@ -5,10 +5,14 @@
...
 
-export interface ClrDatagridFilterInterface<T> {
+export interface ClrDatagridFilterInterface<T, S = any> {
   isActive(): boolean;
 
   accepts(item: T): boolean;
 
   changes: Observable<any>;
+
+  state?: S;
```
youdz: `readonly state?: S;`, to allow for a getter only if they want. Specifying `readonly` in the interface still lets them have an actual property if they want to, we just won't write to it.

src/clr-angular/data/datagrid/providers/state.provider.ts
```diff
@@ -59,13 +59,16 @@ export class StateProvider<T> {
     if (activeFilters.length > 0) {
       state.filters = [];
       for (const filter of activeFilters) {
-        if (filter instanceof DatagridStringFilterImpl) {
+        if (filter.state) {
+          state.filters.push(filter.state);
+          continue;
+        } else if (filter instanceof DatagridStringFilterImpl) {
```
youdz: This if should be removed, instead just implement `state` and `equals()` in our built-in filters. It'll simplify everything.

golden/clr-angular.d.ts
Jinnie: Done

golden/clr-angular.d.ts
Jinnie: 😢

src/clr-angular/data/datagrid/datagrid-filter.spec.ts
Jinnie: My initial idea was to play the scenario. But in the end even at that moment I had doubts...
You are right, it's totally pointless the way it is. Reverting.
Won't resolve the conversation in case someone can suggest a more meaningful test scenario for this case.

src/clr-angular/data/datagrid/interfaces/filter.interface.ts
Jinnie: Done

src/clr-angular/data/datagrid/datagrid-filter.spec.ts
Jinnie: Removed

src/clr-angular/data/datagrid/providers/state.provider.ts
Jinnie: Done

src/clr-angular/data/datagrid/built-in/filters/datagrid-string-filter-impl.spec.ts
```diff
@@ -48,6 +51,44 @@ export default function(): void {
...
+        const propFilter = new DatagridPropertyStringFilter('a.b.c');
+        fullFilter = new DatagridStringFilterImpl(propFilter);
+      });
+
+      it('exposes state', function() {
+        fullFilter.value = 'test';
+        expect(fullFilter.state).toEqual({ property: 'a.b.c', value: 'test' });
+      });
+
+      it('compares filters', function() {
```
youdz: Great test. 👍

PR #3067, opened 2019-01-23
feat(toggle) new toggle implementation for form controls

### LINKED ISSUE #2953: Toggler component with new forms
## Summary

We have existing toggle buttons, but want to support them in new forms properly.

* Extend the checkbox functionality to support toggler (its largely a class to add)

## Use case

Toggles often need to be in a form like a checkbox, so use cases are largely the same.

## Examples

Nope.

## Workarounds

You can use it without this, but it doesn't get the niceties and layouts.

gnomeontherun:
This component is largely just a piggy back off of the checkbox component because toggles are checkboxes with a different visual treatment. That is why the checkbox and toggle selectors are bundled together and there is a service to track if the checkbox or toggle was defined, and display accordingly.

closes #2953

### Release Notes

This release contains a new ClrToggle component that works very similarly to the ClrCheckbox, but has the visual display of a toggle switch instead of a typical checkbox. The control also has support for validation. For the most basic use, you define a checkbox with the toggle directive like this: `<input type="checkbox" clrToggle />`. To see the full functionality, view the [Toggle Switches](/documentation/toggle-switches) documentation. We also are deprecating the old toggle switches, since they are coupled with the old forms and are slated for removal in 2.0.


src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -4,23 +4,47 @@
...
+ * visual styling. The challenge is that the container needs to know which selector was used, which the @Attribute
+ * decorator gets for us to determine if the toggle is used, and emits a value to the wrapper container to tell it
+ * there is a toggle switch instead.
+ */
+@Directive({ selector: '[clrCheckbox],[clrToggle]' })
 export class ClrCheckbox extends WrappedFormControl<ClrCheckboxWrapper> {
   constructor(
     vcr: ViewContainerRef,
-    injector: Injector,
+    private injector: Injector,
```
gnomeontherun: Need to revert this.

src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -19,8 +26,25 @@ export class ClrCheckbox extends WrappedFormControl<ClrCheckboxWrapper> {
     @Optional()
     control: NgControl,
     renderer: Renderer2,
-    el: ElementRef
+    el: ElementRef,
+    @Attribute('clrToggle') private toggle: string
```
Shijir: Why this is not `@Input('clrToggle')`? Why treating `clrToggle` as an attribute?

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
```diff
@@ -4,34 +4,58 @@
...
-import { OnInit, Component, ContentChild } from '@angular/core';
+import { Component, OnInit, ContentChild, Inject, InjectionToken, OnDestroy } from '@angular/core';
+import { BehaviorSubject, Subscription } from 'rxjs';
 
 import { DynamicWrapper } from '../../utils/host-wrapping/dynamic-wrapper';
 import { ControlIdService } from '../common/providers/control-id.service';
 import { ClrLabel } from '../common/label';
 
+/* tslint:disable-next-line:variable-name */
+export const TogglerService = new InjectionToken<any>(undefined);
```
Shijir: Isn't it cleaner to have a designated service class instead of using `InjectionToken`?

youdz: Actually this is fine, but can you use `InjectionToken<Subject<boolean>` instead, so we get proper type checking? Thanks.

I also just realized I missed the same comment on the `ToggleService` for the password input. If you want to squeeze it in the PR (even though it has nothing to do with it) you can, otherwise we'll just fix it with a one-line PR later.

Also, you wouldn't have to disable TSLint if we matched this to our conventions. We do this in a couple places, and we use actual constant names since we're not truly defining a service but just an injection token and a provider. In this case, you'd have:
```typescript
export const IS_TOGGLE = new InjectionToken<Subject<boolean>>("IS_TOGGLE");
export function isToggleFactory() {
  return new BehaviorSubject<boolean>(false);
}
export const IS_TOGGLE_PROVIDER = { provide: IS_TOGGLE, useFactory: isToggleFactory };
```
and then your component simply declares: `providers: [ControlIdService, IS_TOGGLE_PROVIDER]`.
Note that the string argument in the `InjectionToken` constructor is here purely for error messages and debugging purposes, so it's good practice to always set it. It's not public anyway.

package.json
```diff
@@ -58,6 +58,7 @@
     "publish:next": "npm publish dist/clr-ui --tag next; npm publish dist/clr-angular --tag next; npm publish dist/clr-icons --tag next",
     "publish:local": "npm publish dist/clr-ui --registry http://localhost:4873; npm publish dist/clr-angular --registry http://localhost:4873; npm publish dist/clr-icons --registry http://localhost:4873",
     "ks:publish": "ng build --prod ks-app; cat ./dist/ks-app/index.html > ./dist/ks-app/404.html; surge -p ./dist/ks-app",
+    "build:dev": "ng build --prod",
```
youdz: This line made me laugh.

src/clr-angular/forms/checkbox/checkbox-container.ts
```diff
@@ -15,12 +15,13 @@ import { LayoutService } from '../common/providers/layout.service';
...
 
 @Component({
-  selector: 'clr-checkbox-container',
+  selector: 'clr-checkbox-container,clr-toggle-container',
   template: `
     <ng-content select="label"></ng-content>
     <label *ngIf="!label && addGrid()"></label>
     <div class="clr-control-container" [class.clr-control-inline]="clrInline" [ngClass]="controlClass()">
       <ng-content select="clr-checkbox-wrapper"></ng-content>
+      <ng-content select="clr-toggle-wrapper"></ng-content>
```
youdz: Nitpick: I'd go with `<ng-content select="clr-checkbox-wrapper, clr-toggle-wrapper"></ng-content>` to better show that we're really using them in the same way. But I don't feel strongly about this.

src/clr-angular/forms/checkbox/checkbox.ts
```diff
@@ -19,8 +26,25 @@ export class ClrCheckbox extends WrappedFormControl<ClrCheckboxWrapper> {
...
   ) {
     super(vcr, ClrCheckboxWrapper, injector, control, renderer, el);
   }
+
+  ngOnInit() {
+    super.ngOnInit();
+
+    // This works fine normally, but with testing it throws unless its wrapped manually or has try/catch
+    // It basically doesn't seem to wire up the host wrapper container in time for this specific call, but the wrapper is there
+    try {
```
youdz: The fact that you need an empty try/catch is pretty scary. I'm okay(-ish) shipping this blindly if it works, but it might be hiding a real issue somewhere that I can't see right now.

src/clr-angular/forms/checkbox/checkbox.ts
gnomeontherun: I only wanted it to bind once, not be an actual input that could toggle.

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
gnomeontherun: Updated.

src/clr-angular/forms/checkbox/checkbox-container.ts
gnomeontherun: Sure, updated.

package.json
gnomeontherun: This is for the netlify script, wanting to keep it consistent.

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
```diff
@@ -11,12 +11,11 @@ import { DynamicWrapper } from '../../utils/host-wrapping/dynamic-wrapper';
 import { ControlIdService } from '../common/providers/control-id.service';
 import { ClrLabel } from '../common/label';
 
-/* tslint:disable-next-line:variable-name */
-export const TogglerService = new InjectionToken<any>(undefined);
-/* tslint:disable-next-line:variable-name */
-export function TogglerServiceProvider() {
+export const IS_TOGGLE = new InjectionToken<boolean>('IS_TOGGLE');
```
youdz: Careful, this is `InjectionToken<Subject<boolean>>` or `InjectionToken<BehaviorSubject<boolean>>`, depending on how specific you want to be.

I would expect AoT compilation to fail on this, not sure if it does.

src/clr-angular/forms/checkbox/checkbox.ts
gnomeontherun: As we discussed, I moved this up into the wrapped control directive with a comment about it.

src/clr-angular/forms/checkbox/checkbox-wrapper.ts
gnomeontherun: Updated

Issue #2856, opened 2018-11-16
Sidenav progress bar missing in kitchen sink app v1.0


Jinnie:
## Describe the bug

In old versions of the KS-APP, there is small, inline, progress bar in the Sidenav section. In the 1.0 KS app it is offscreen (see screenshots).
Found:
<img width="365" alt="screen shot 2018-11-16 at 16 58 57" src="https://user-images.githubusercontent.com/1798828/48629218-bcc83f00-e9c1-11e8-8234-51f37eb34174.png">
Expected:
<img width="450" alt="screen shot 2018-11-16 at 16 58 38" src="https://user-images.githubusercontent.com/1798828/48629264-d6698680-e9c1-11e8-877e-a5811bc6bc83.png">

## How to reproduce
Start the ks-app locally from master branch. Go to:
`http://localhost:4200/progress-bars
`
## Expected behavior

To be able to see a progress bar between the links on the page

## Versions

**App**

* Angular: [7.0]
* Node: [10.0]
* Clarity: [1.0-beta.2]

**Device:**

* Type: [MacBook]
* OS: [OSX]
* Browser [chrome]
* Version [70.0.3538.77]

## Additional notes
Removing the position:absolute style on the progressbar holder seems to fix it. Need to be checked if the problem is not with the parent, though, as this position is available on the working versions too.

mathisscott:
@Jinnie 
We can't really add `position: relative` to the new grid. Because that was what was causing tooltips and popovers to get cutoff in the old forms and whatnot.

To fix this we will likely need to update the DOM structure of the sidebar progress elements. If removing `position: absolute` does fix the problem, though, that may be a lighter fix for it.

Jinnie:
We should also generally research where else we might be using old grid classes.
We can consider it part of this task, unless you spawn a different issue to track it.

gnomeontherun:
There are a lot of places still using old grid classes, and thats something we'll need to fix by 2.0. I think we should make that a separate ticket and assign it to 2.0 work.

PR #3089, opened 2019-01-31
[UI][KS-APP] Fix inline progress-bar (#2856)

Jinnie:
The new grid classes don't impose position: relative on the container.


Jinnie:
Thanks for bringing this to attention, Eudes.
I verified it by running an old version of the ks-app (0.13) and applying this change there.
It seems to work fine. It gets position:relative from the .progress.loop rule and looks okay.
screen shot 2019-02-01 at 4 15 08 pm
So we are not breaking layouts with the old grid by applying this change.

PR #3102, opened 2019-02-12
[NG] Vertical tabs (#2856)

didkobravo:
Ability to display vertical tabs. Overflow is disabled in vertical, replaced with scroll.

Signed-off-by: Deyan Angelov <dangelov@vmware.com>


src/dev/src/app/tabs/tabs-static.demo.html
```diff
@@ -3,45 +3,94 @@
...
+    <section id="panelSettings" role="tabpanel" aria-labelledby="tabSettings" aria-hidden="true">
+      <p>Content for Settings</p>
+    </section>
+  </div>
+
+  <h3>Vertical tabs</h3>
+  <div class="demo tabs-vertical">
+    <ul id="verticalDemoTabs" class="nav" role="tablist">
+      <li role="tab" class="nav-item">
+        <button id="verticalTabDashboard" class="btn btn-link nav-link active" aria-controls="panelDashboard" aria-selected="true" type="button">Dashboard</button>
```
gnomeontherun: Do we need the `id` attribute on all these buttons? People tend to copy/paste from our demos so would like them to be clean.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +54,28 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
     `,
   providers: [IfActiveService, IfOpenService, TabsService, TABS_ID_PROVIDER],
+  host: {
+    '[class.tabs-vertical]': 'tabsService.orientation === "vertical"',
+  },
 })
 export class ClrTabs implements AfterContentInit {
-  @ContentChildren(ClrTabLink, { descendants: true })
-  tabLinkDirectives: QueryList<ClrTabLink>;
+  @Input('orientation')
```
gnomeontherun: I'm wondering if we want to call this `layout` instead? Orientation to me is like portrait/landscape.

src/clr-angular/layout/tabs/_tabs.clarity.scss
```diff
@@ -27,3 +27,48 @@ button.nav-link {
     margin-right: 0;
   }
 }
+
+.tabs-vertical {
```
gnomeontherun: So the big hangup here is that if you do static tabs with vertical, they require an extra container. We can mask that with Angular but not in static HTML. Is my understanding correct?

src/dev/src/app/tabs/tabs-static.demo.html
```diff
@@ -3,45 +3,94 @@
...
-        <button class="btn" aria-controls="panelInfrastructure">Infrastructure</button>
-        <button class="btn" aria-controls="panelSettings">Settings</button>
+      <li role="tab" class="nav-item">
+        <button id="tabManagement" class="btn btn-link nav-link" aria-controls="panelManagement" aria-selected="false" type="button">Management</button>
+      </li>
+      <li role="tab" class="nav-item">
+        <button id="tabCloud" class="btn btn-link nav-link" aria-controls="panelCloud" aria-selected="false" type="button">Cloud</button>
+      </li>
+      <div class="tabs-overflow open bottom-right">
+        <li class="nav-item">
```
hippee-lee: I know its a demo but shouldn't this button have role="tab" still?

src/clr-angular/layout/tabs/tabs.ts
hippee-lee: In addition to @gnomeontherun's comment this input needs to be prefixed with `clr` per Clarity naming conventions: https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md#naming-conventions

src/clr-angular/layout/tabs/providers/tabs.service.ts
```diff
@@ -26,7 +28,7 @@ export class TabsService {
 
   get overflowTabs() {
     return this.children.filter((tab: ClrTab) => {
-      return tab.tabLink.inOverflow === true;
+      return this.orientation !== 'vertical' && tab.tabLink.inOverflow;
```
hippee-lee: I know that previously we didn't have a spec for this service. Would you mind adding a spec for this service that tests the two changes here?

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -29,9 +30,8 @@ let nbTabLinkComponents: number = 0;
     '[attr.aria-hidden]': 'false',
     '[attr.aria-controls]': 'ariaControls',
     '[class.btn]': 'true',
-    '[class.btn-link]': '!inOverflow',
-    '[class.nav-link]': '!inOverflow',
-    '[class.nav-item]': '!inOverflow',
+    '[class.btn-link]': 'tabsService.orientation === "vertical" || !inOverflow',
+    '[class.nav-link]': 'tabsService.orientation === "vertical" || !inOverflow',
```
hippee-lee: Would you mind adding a test that checks for the correct class on the button when the orientation changes?

src/clr-angular/layout/tabs/_tabs.clarity.scss
Jinnie: Yes, Jeremy, absolutely correct. The static HTML has not a single root, but is handled as a collection of elements, all on root level. Although it's not needed for the component functioning, I think it is not a bad idea if we aim for all our components to have a single root.
The website and dev/ks-apps are updated with appropriate examples, which you can see later in this review.

src/clr-angular/layout/tabs/providers/tabs.service.ts
```diff
@@ -10,6 +10,8 @@ import { ClrTab } from '../tab';
 export class TabsService {
   private _children: ClrTab[] = [];
 
+  orientation: 'horizontal' | 'vertical';
```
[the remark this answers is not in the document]
Jinnie: done!
- enum
- horizontal set explicitly as default

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -29,9 +30,8 @@ let nbTabLinkComponents: number = 0;
     '[attr.aria-hidden]': 'false',
     '[attr.aria-controls]': 'ariaControls',
     '[class.btn]': 'true',
-    '[class.btn-link]': '!inOverflow',
-    '[class.nav-link]': '!inOverflow',
-    '[class.nav-item]': '!inOverflow',
+    '[class.btn-link]': 'tabsService.orientation === "vertical" || !inOverflow',
```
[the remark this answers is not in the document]
Jinnie: done

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -48,6 +48,7 @@ export class ClrTabLink {
     private el: ElementRef,
     private cfr: ComponentFactoryResolver,
     private viewContainerRef: ViewContainerRef,
+    public tabsService: TabsService,
```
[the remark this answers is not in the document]
Jinnie: done.

src/clr-angular/layout/tabs/tabs.ts
Jinnie: Both comments addressed. Using clrTabsLayout now.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +54,28 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
+  set orientation(orientation: 'horizontal' | 'vertical') {
+    this.tabsService.orientation = orientation;
```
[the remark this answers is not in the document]
Jinnie: Absolutely, we should not silently ignore such errors.
Fixed. Throwing an exception on wrong input from now on.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -76,8 +94,8 @@ export class ClrTabs implements AfterContentInit {
   }
 
   ngAfterContentInit() {
-    if (typeof this.ifActiveService.current === 'undefined') {
-      this.tabLinkDirectives.first.activate();
+    if (typeof this.ifActiveService.current === 'undefined' && this.tabLinkDirectives[0]) {
```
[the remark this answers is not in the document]
Jinnie: Unless we have tabs component without tabs inside them it will not be a problem. It's not a very realistic case, but I hit this issue once while testing so I decided it won't hurt to add a check that what we're trying to activate actually exists.

src/dev/src/app/tabs/tabs-static.demo.html
```diff
@@ -3,45 +3,94 @@
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
+<p>You cannot switch between tabs in this static demo because it does not contain the JavaScript to handle this behavior. The
```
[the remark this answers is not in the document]
Jinnie: It's not about switching modes, but about clicking on tabs.
This text existed before in the HTML, but was not visible when rendered as it was out of the layout. I decided to bring it back.
See line 46 that was removed from this diff.

src/dev/src/app/tabs/tabs-static.demo.html
Jinnie: It's negotiable. It's both a tab and a menu item. I'd be happy to get a bit more discussion around the accessibility of the overflow. Maybe involve the new a11y engineer once he joins the team?

src/dev/src/app/tabs/tabs-static.demo.html
Jinnie: I think I intended to use them in some gemini tests about hovering vs selected style, but I then simplified them and now the ids may be useless. Will check it and update here soon.
Though this is the dev-app, not the website and whoever can get to here should be advanced enough to be able to copy smartly.

src/dev/src/app/tabs/tabs.demo.scss
```diff
@@ -3,5 +3,13 @@
...
 .clr-example {
-  height: 360px;
+  height: 590px;
+
+  .demo {
+    min-height: 120px;
+  }
+}
+
+.tabs-vertical {
```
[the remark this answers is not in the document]
Jinnie: exactly!

src/clr-angular/layout/tabs/providers/tabs.service.ts
Jinnie: Will add a test!

src/clr-angular/layout/tabs/tab-link.directive.ts
Jinnie: Will add a test

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +54,28 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
+  get orientation(): 'horizontal' | 'vertical' {
+    return this.tabsService.orientation;
+  }
 
-  @ContentChildren(ClrTabContent, { descendants: true })
-  tabContents: QueryList<ClrTabContent>;
+  @ContentChildren(ClrTab) private tabs: QueryList<ClrTab>;
+
+  get tabLinkDirectives(): ClrTabLink[] {
+    return this.tabs.map(tab => tab.tabLink);
```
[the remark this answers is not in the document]
Jinnie: TBD

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +54,28 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
+  }
+
+  get tabContents(): ClrTabContent[] {
+    return this.tabs.filter(tab => !!tab.tabContent).map(tab => tab.tabContent);
```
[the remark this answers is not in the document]
Jinnie: TBD

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
Jinnie: Done

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
Jinnie: Actually, this array is pretty fine as it is. Due to the *clrIfActive it is far more dynamic than the clrTabLinks case. Now this call is carried only on active tab change, which is the desired behavior. It is not feasible to try to cover the case without *clrIfActive, the mixed case, etc.

src/dev/src/app/tabs/tabs-static.demo.html
Jinnie: No, it's not me, I've just conformed to the style of the other elements we already had on this page. Nevertheless, these IDs are not used anywhere, so I've removed them altogether.
Fixed.

src/dev/src/app/tabs/tabs-static.demo.html
Jinnie: Looking more closely at it, this is only the [...] overflow activation button.
It has not role of a tab, it only opens the context menu, where the buttons do have the role=tab attribute. See lines 29, 30 bellow.

src/clr-angular/layout/tabs/providers/tabs.service.ts
Jinnie: On a second check, there is a reason not to have a spec for this service. To be properly initialized it will requires a list of ClrTab items. This can be done most easily with a dummy test component. If we do this here, it will become a duplicate of what we have in tabs.spec.ts.
The service is already tested there, directly and indirectly, so I added my tests there instead of creating a new spec here.
I test the service state changes, and how the overflow is handled in vertical mode. Should be enough.

src/clr-angular/layout/tabs/tab-link.directive.ts
Jinnie: Done. Added tests for overflow too.

src/clr-angular/layout/tabs/index.ts
```diff
@@ -10,3 +10,4 @@ export * from './tab-content';
 export * from './tab-overflow-content';
 export * from './tab-link.directive';
 export * from './tabs.module';
+export * from './enums/tabsLayout';
```
gnomeontherun: Don't export this publicly.

src/clr-angular/layout/tabs/enums/tabsLayout.ts
```diff
@@ -0,0 +1,10 @@
+/*
```
gnomeontherun: Filename should be `tabs-layout.enum.ts`

src/clr-angular/layout/tabs/tab-link.directive.spec.ts
```diff
@@ -13,19 +13,21 @@ import { TabsService } from './providers/tabs.service';
 import { ClrTabLink } from './tab-link.directive';
 import { TABS_ID_PROVIDER } from './tabs-id.provider';
 import { ClrTabsModule } from './tabs.module';
+import { TabsLayout } from '@clr/angular';
```
gnomeontherun: Wrong path

src/website/src/app/documentation/demos/tabs/tabs-angular-vertical.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component } from '@angular/core';
+
+const EXAMPLE = `
+<clr-tabs clrTabsLayout="vertical">
```
gnomeontherun: @youdz do we want this, or `clrTabLayout` or `clrLayout`? We use `clrLayout` with forms.

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
gnomeontherun: I kind of like the `tabLinkDirectives.first` over using the array index. As long as it is consistent I guess.

src/dev/src/app/tabs/tabs-angular.ts
```diff
@@ -4,12 +4,14 @@
...
 import { Component } from '@angular/core';
+import { TabsLayout } from '@clr/angular';
 
 @Component({
   selector: 'clr-modal-tabs-angular',
   styleUrls: ['./tabs.demo.scss'],
   templateUrl: './tabs-angular.demo.html',
 })
 export class TabsAngularDemo {
+  layout: TabsLayout = TabsLayout.VERTICAL;
```
gnomeontherun: We really don't want to use the `TabsLayout` in our demos directly, so this should just be the string value.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +56,35 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
-export class ClrTabs implements AfterContentInit {
-  @ContentChildren(ClrTabLink, { descendants: true })
-  tabLinkDirectives: QueryList<ClrTabLink>;
+export class ClrTabs implements AfterContentInit, OnDestroy {
+  private subscription: Subscription;
+
+  @Input('clrTabsLayout')
+  set layout(layout: TabsLayout) {
+    if (!Object.values(TabsLayout).includes(layout)) {
+      throw `Unsupported layout ${layout}`;
```
gnomeontherun: I don't think we want to throw an error here, I think we just want to default to horizontal. @youdz?

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -76,12 +103,27 @@ export class ClrTabs implements AfterContentInit {
...
   toggleOverflow(event: any) {
     this.ifOpenService.toggleWithEvent(event);
   }
+
+  isVertical() {
+    return this.layout === TabsLayout.VERTICAL;
+  }
+
+  ngOnDestroy() {
+    if (this.subscription) {
```
gnomeontherun: Our pattern has been to push subscriptions onto an array (even if there is only one for now, could always expand) and then here we just loop over each and unsubscribe. Saves this if block and expands more easily.

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
gnomeontherun: I would need to inspect this again more closely, but can't we just get this anytime the `tabs.changes` subscription fires like the link directives? It might be firing now only when tabs change, (for now) but this is still a public property that could be optimized if I'm reading it right. I think we'd rather put it in the right place and ensure it fires on tab changes, unless that doesn't function.

src/clr-angular/layout/tabs/enums/tabsLayout.ts
Jinnie: omg! of course!

src/dev/src/app/tabs/tabs-angular.ts
Jinnie: Not using TabsLayout makes things a lot simpler - removing!

And about the tabLinkDirectives.first, yes - it is cooler, but tabLinkDirectives is not a QueryList anymore, but an array, so it has to be [0] instead. We can't use the QueryList returned by the previous @ContentChildren (descendants: true), because it also collects child nodes of nested tabs.

src/clr-angular/layout/tabs/tab-link.directive.spec.ts
Jinnie: oops! fixed.
should I also change it to simple string, or using the import (correct one) in unit tests is ok?

src/clr-angular/layout/tabs/tabs.ts
Jinnie: Done. I just noticed we have both patterns in our code and decided to go for the short one. Good to know.

src/clr-angular/layout/tabs/providers/tabs.service.ts
```diff
@@ -26,7 +29,7 @@ export class TabsService {
 
   get overflowTabs() {
     return this.children.filter((tab: ClrTab) => {
-      return tab.tabLink.inOverflow === true;
+      return this.layout !== TabsLayout.VERTICAL && tab.tabLink.inOverflow;
```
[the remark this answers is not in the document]
Jinnie: Done. Far more maintainable...

About testing, it's tested indirectly in tabs.spec > Tabs > Projection > "does not activate overflow in vertical mode". To test this here, I'd have to nearly duplicate the initialization done it tabs.spec. Or mock some ClrTab items. Matt also asked about a spec for this service. Do we need to have one, as code coverage is assured by tabs.spec anyway?

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -84,4 +86,8 @@ export class ClrTabLink {
   get active() {
     return this.ifActiveService.current === this.id;
   }
+
+  isVertical() {
```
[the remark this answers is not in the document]
Jinnie: Done. I was being lazy finding the proper naming here.

src/clr-angular/layout/tabs/tabs.spec.ts
```diff
@@ -228,6 +242,34 @@ describe('Tabs', () => {
...
+    let tabsService: TabsService;
+
+    beforeEach(function() {
+      context = this.create(ClrTabs, TestComponent);
+      tabsService = context.fixture.componentInstance.tabsInstance.tabsService;
+      compiled = context.fixture.nativeElement;
+    });
+
+    it('service defaults to horizontal', function() {
+      expect(tabsService.layout).toEqual(TabsLayout.HORIZONTAL);
```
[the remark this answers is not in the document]
Jinnie: Right. It was working well in the beginning, but I changed the template afterwards to fit other test cases. 
Done. Switched to other template that does not have a layout binding.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -3,25 +3,30 @@
...
 
 @Component({
   selector: 'clr-tabs',
   template: `
         <ul class="nav" role="tablist" [attr.aria-owns]="tabIds">
             <!--tab links-->
             <ng-container *ngFor="let link of tabLinkDirectives">
-                <ng-container *ngIf="link.tabsId === tabsId && !link.inOverflow"
-                              [ngTemplateOutlet]="link.templateRefContainer.template">
+                <ng-container *ngIf="link.tabsId === tabsId && !(!isVertical() && link.inOverflow)">
```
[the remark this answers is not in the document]
Jinnie: :) now it's just `&& !link.inOverflow`
Fixed.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -51,13 +56,35 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
         </ng-container>
     `,
   providers: [IfActiveService, IfOpenService, TabsService, TABS_ID_PROVIDER],
+  host: {
+    '[class.tabs-vertical]': 'isVertical()',
```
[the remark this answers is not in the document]
Jinnie: Started discussion on this. For the time being, transitioned all non-static metas in this module to fit the accepted code style.

src/clr-angular/layout/tabs/tabs.ts
youdz: You're right, I didn't pick up on this but some people don't have the cleanest lifecycles and we do receive the occasional `undefined` before the actual correct value. This would throw errors too often. Just rely on Typescript, you already made the argument of type `TabsLayout`, so Typescript will throw an error in AoT compilation if they give us something else.

TL;DR: Keep the if, but just ignore instead of throwing an error.

src/website/src/app/documentation/demos/tabs/tabs-angular-vertical.ts
youdz: I don't think we have a conflict risk here, so I like `clrLayout` better. Like you I guess, since you raised it.

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
gnomeontherun: This is my last question, just wanting to bring it back to attention.

src/website/src/app/documentation/demos/tabs/tabs-angular-vertical.ts
Jinnie: Done

src/clr-angular/layout/tabs/tabs.ts
Jinnie: Done

src/clr-angular/layout/tabs/providers/tabs.service.ts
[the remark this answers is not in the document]
Jinnie: Resolving as we agreed that the integration test in the main spec should be okay for this case.

src/clr-angular/layout/tabs/tabs.ts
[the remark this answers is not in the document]
Jinnie: To cover the `*clrIfActive` case we have to subscribe not only to `tabs.changes`, but to `IfActiveService#currentChange`. It might be possible to make some optimization for the `no-clrIfActive` case, but the tabs component will need to know if `*clrIfActive` has been applied to it. I'm still not sure that the added complexity and risk for breaking corner cases is worth the optimization of a possible, though improbable, future case.
Also the way that the tabs component is used does not imply neither huge amounts of tabs, nor frequent repetitive calls of this method.
These are my reasons to prefer a mildly suboptimal approach, which is safer and easier to maintain.

src/website/src/app/documentation/demos/tabs/tabs-angular-vertical.ts
Jinnie: Oops, I'll have to unresolve this. The clrLayout we use in the tabs is a `@Directive`, not an `@Input` as it it here. So it tries to get applied on the tabs too, leading to many errors.
I guess we can't use clrLayout then.

Jinnie: Done again after supplying a more specific selector to the forms clrLayout directive.

PR #3117, opened 2019-02-15
[NG] fix DataGrid clearSelection for SingleSelect mode

### LINKED ISSUE #3106: ClrDataGrid clearSelection does not work when using Single Selection mode
## Describe the bug
Selection is not cleared when applying filters if Single Section mode is used. (for Multiple selection mode the selection is cleared)

I will start with the fact that this is a bit confusing: 
- there is an issue to preserve selection when filtering (https://github.com/vmware/clarity/issues/2342)
- there is bug fix  for when selection is cleared by applying filter (https://github.com/vmware/clarity/pull/2381)

The bug fix works for the multi selection but does not fix the problem for the SingleSelection.

It looks like clearSelection only clears for MultiSelection and not for SingleSelection
https://github.com/vmware/clarity/blob/master/src/clr-angular/data/datagrid/providers/selection.ts#L149
```
public clearSelection(): void {
    this.current.length = 0;
    this.prevSelectionRefs = [];
    this.emitChange();
  }
```

I don't understand what is the expected behaviour: should the selection be cleared or should it not when applying filters.

If the selection should be cleared then it should probably also work for Single selection mode.
Probably: 
```
public clearSelection(): void {
    this.current.length = 0;
    this.prevSelectionRefs = [];
    this._currentSingle = null;
    this.prevSingleSelectionRef = null;
    this.emitChange();
  }
```
Thanks for looking into this!

sergiupantiru:
Selection was not cleared when Filters were applied for SingleSelect mode

closes #3106

Signed-off-by: Sergiu Pantiru <12131904+sergiupantiru@users.noreply.github.com>

### DIFF
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -235,6 +235,59 @@ export default function(): void {
         expect(nbChanges).toBe(2);
       });
 
+      it('clears selection when a filter is added (selectionType single)', function() {
+        selectionInstance.selectionType = SelectionType.Single;
+        selectionInstance.currentSingle = 2;
+
+        const evenFilter: EvenFilter = new EvenFilter();
+
+        filtersInstance.add(<ClrDatagridFilterInterface<number>>evenFilter);
+
+        evenFilter.toggle();
+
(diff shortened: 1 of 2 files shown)


sergiupantiru:
Ha, thanks. It's good to have some points in the beginning, this way I can loose a few along the way 😄

PR #3124, opened 2019-02-20
adding a column service to datagrid to track widths

gnomeontherun:
This will eventually be used to store several other stateful properties of columns, such as hidden and ordering information. For now, it is to give us a starting implementation point to store column states in a way that is consolidated and easily cached.

Internally, this lifts the state of the width calculations out of the render organizer into a centralized column service. This array of behaviorsubjects get passed into the column cells based on their index, and each cell subscribes to column changes. This privatizes some of the connection between the main/header and row/cell renderers. More work is to be done here, this is just a first pass at the basic architecture.

I also moved the SelectionType enum to its own file, since that is our pattern.

**Internal change - no release notes necessary**

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-separator.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-separator.ts
@@ -61,6 +61,7 @@ export class ClrDatagridColumnSeparator {
 
   private redFlagTracker(resizeTrackerEl: HTMLElement) {
     let isWithinMaxResizeRange: boolean;
+    // @TODO(JEREMY) Review this, it will always be true because above is always null
     if (isWithinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
       isWithinMaxResizeRange = this.columnResizerService.isWithinMaxResizeRange;
       if (!isWithinMaxResizeRange) {
--- a/src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
+++ b/src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
+
+export interface DatagridColumnState {
+  changes?: DatagridColumnChanges[]; // This is an array of change types to update
+  width?: number; // This is the width calculated for the column
+  strictWidth?: number; // This is the strict width if defined in styles/css
+}
--- a/src/clr-angular/data/datagrid/providers/columns.service.ts
+++ b/src/clr-angular/data/datagrid/providers/columns.service.ts
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Injectable, OnDestroy } from '@angular/core';
+import { BehaviorSubject, Subscription } from 'rxjs';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridRenderStep } from '../enums/render-step.enum';
+import { DatagridRenderOrganizer } from '../render/render-organizer';
+
+@Injectable()
+export class ColumnsService implements OnDestroy {
+  subscriptions: Subscription[] = [];
+  columns: BehaviorSubject<DatagridColumnState>[] = [];
+
+  constructor(private organizer: DatagridRenderOrganizer) {
+    this.subscriptions.push(
+      this.organizer.filterRenderSteps(DatagridRenderStep.CLEAR_WIDTHS).subscribe(() => this.reset())
+    );
+  }
+
+  ngOnDestroy() {
+    this.subscriptions.forEach(sub => sub.unsubscribe());
+  }
+
+  private reset() {
+    this.columns.forEach((column, index) => {
+      this.emitStateChange(index, { width: 0 });
+    });
+  }
+
+  // Helper method to emit a change to a column only when there is an actual diff to process for that column
+  emitStateChange(columnIndex: number, diff: Partial<DatagridColumnState>) {
+    if (!this.columns[columnIndex]) {
+      return;
+    }
+    const current = this.columns[columnIndex].value;
+    const hasChange = Object.keys(diff).filter(key => diff[key] !== current[key]);
+
+    if (hasChange) {
+      this.columns[columnIndex].next({ ...current, ...diff });
+    }
+  }
+}
--- a/src/clr-angular/data/datagrid/render/cell-renderer.ts
+++ b/src/clr-angular/data/datagrid/render/cell-renderer.ts
@@ -4,15 +4,27 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import { Directive, ElementRef, OnDestroy, Renderer2 } from '@angular/core';
-import { Subscription } from 'rxjs';
+import { BehaviorSubject, Subscription } from 'rxjs';
 
 import { DatagridRenderStep } from '../enums/render-step.enum';
 
 import { STRICT_WIDTH_CLASS } from './constants';
 import { DatagridRenderOrganizer } from './render-organizer';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 @Directive({ selector: 'clr-dg-cell' })
 export class DatagridCellRenderer implements OnDestroy {
+  private stateSubscription: Subscription;
+
+  // @TODO(JEREMY) Work out how to dedupe some of this code between header and cell renderers
+  set columnState(columnState: BehaviorSubject<DatagridColumnState>) {
+    if (this.stateSubscription) {
+      this.stateSubscription.unsubscribe();
+    }
+    this.stateSubscription = columnState.subscribe(state => this.stateChanges(state));
+  }
+
   constructor(private el: ElementRef, private renderer: Renderer2, organizer: DatagridRenderOrganizer) {
     this.subscriptions.push(
       organizer.filterRenderSteps(DatagridRenderStep.CLEAR_WIDTHS).subscribe(() => this.clearWidth())
@@ -22,6 +34,23 @@ export class DatagridCellRenderer implements OnDestroy {
   private subscriptions: Subscription[] = [];
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
+    if (this.stateSubscription) {
+      this.stateSubscription.unsubscribe();
+    }
+  }
+
+  private stateChanges(state: DatagridColumnState) {
+    if (state.changes && state.changes.length) {
+      state.changes.forEach(change => {
+        switch (change) {
+          case DatagridColumnChanges.WIDTH:
+            this.setWidth(state);
+            break;
+          default:
+            break;
+        }
+      });
+    }
   }
 
   private clearWidth() {
@@ -29,12 +58,12 @@ export class DatagridCellRenderer implements OnDestroy {
     this.renderer.setStyle(this.el.nativeElement, 'width', null);
   }
 
-  public setWidth(strict: boolean, value: number) {
-    if (strict) {
+  private setWidth(state: DatagridColumnState) {
+    if (state.strictWidth) {
       this.renderer.addClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
     } else {
       this.renderer.removeClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
     }
-    this.renderer.setStyle(this.el.nativeElement, 'width', value + 'px');
+    this.renderer.setStyle(this.el.nativeElement, 'width', state.width + 'px');
   }
 }
--- a/src/clr-angular/data/datagrid/render/constants.ts
+++ b/src/clr-angular/data/datagrid/render/constants.ts
@@ -4,6 +4,8 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+// @TODO The top two are not used now, which is probably a performance drag that was broken along the way.
+// There was a previous pattern to hide everything to do computation then display, for Firefox, needs revisiting.
 export const NO_LAYOUT_CLASS = 'datagrid-no-layout';
 export const COMPUTE_WIDTH_CLASS = 'datagrid-computing-columns-width';
 export const STRICT_WIDTH_CLASS = 'datagrid-fixed-width';
--- a/src/clr-angular/data/datagrid/render/header-renderer.spec.ts
+++ b/src/clr-angular/data/datagrid/render/header-renderer.spec.ts
@@ -5,6 +5,7 @@
  */
 import { Component, DebugElement, Renderer2 } from '@angular/core';
 import { By } from '@angular/platform-browser';
+import { BehaviorSubject } from 'rxjs';
 
 import { DomAdapter } from '../../../utils/dom-adapter/dom-adapter';
 import { MOCK_DOM_ADAPTER_PROVIDER, MockDomAdapter } from '../../../utils/dom-adapter/dom-adapter.mock';
@@ -24,6 +25,8 @@ import { STRICT_WIDTH_CLASS } from './constants';
 import { DatagridHeaderRenderer } from './header-renderer';
 import { DatagridRenderOrganizer } from './render-organizer';
 import { MOCK_ORGANIZER_PROVIDER, MockDatagridRenderOrganizer } from './render-organizer.mock';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 @Component({ template: `<clr-dg-column>Hello world</clr-dg-column>` })
 class SimpleTest {}
@@ -61,6 +64,7 @@ export default function(): void {
     let context: TestContext<DatagridHeaderRenderer, SimpleTest>;
     let domAdapter: MockDomAdapter;
     let organizer: MockDatagridRenderOrganizer;
+    let stateSub: BehaviorSubject<DatagridColumnState>;
 
     beforeEach(function() {
       context = this.create(DatagridHeaderRenderer, SimpleTest, [
@@ -75,20 +79,22 @@ export default function(): void {
       ]);
       domAdapter = <MockDomAdapter>context.getClarityProvider(DomAdapter);
       organizer = <MockDatagridRenderOrganizer>context.getClarityProvider(DatagridRenderOrganizer);
+      stateSub = new BehaviorSubject<DatagridColumnState>({});
+      context.clarityDirective.columnState = stateSub;
     });
 
     it('computes the width of header based on its scrollWidth', function() {
       domAdapter._scrollWidth = 123;
-      expect(context.clarityDirective.computeWidth()).toBe(123);
+      expect(context.clarityDirective.getColumnWidthState()).toEqual({ width: 123, strictWidth: 0 });
     });
 
     it('can set the width of a column', function() {
-      context.clarityDirective.setWidth(123);
+      stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123 });
       expect(context.clarityElement.style.width).toBe('123px');
     });
 
     it('resets the header to default width when notified', function() {
-      context.clarityDirective.setWidth(123);
+      stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123 });
       expect(context.clarityElement.style.width).toBe('123px');
       organizer.updateRenderStep.next(DatagridRenderStep.CLEAR_WIDTHS);
       expect(context.clarityElement.style.width).toBeFalsy();
@@ -97,10 +103,10 @@ export default function(): void {
     it('uses the width declared by the user if there is one', function() {
       domAdapter._userDefinedWidth = 123;
       organizer.updateRenderStep.next(DatagridRenderStep.DETECT_STRICT_WIDTHS);
-      expect(context.clarityDirective.strictWidth).toBe(123);
+      expect(context.clarityDirective.getColumnWidthState().strictWidth).toEqual(123);
       domAdapter._userDefinedWidth = 0;
       organizer.updateRenderStep.next(DatagridRenderStep.DETECT_STRICT_WIDTHS);
-      expect(context.clarityDirective.strictWidth).toBeFalsy();
+      expect(context.clarityDirective.getColumnWidthState().strictWidth).toEqual(0);
     });
 
     it('does not remove the width defined by the user', function() {
@@ -109,23 +115,15 @@ export default function(): void {
       organizer.updateRenderStep.next(DatagridRenderStep.CLEAR_WIDTHS);
       organizer.updateRenderStep.next(DatagridRenderStep.DETECT_STRICT_WIDTHS);
       expect(context.clarityElement.style.width).toBe('123px');
-      // One extra cycle to be sure, because clearing widths before computing them
-      // might have a special case handling
-      context.clarityDirective.computeWidth();
-      organizer.updateRenderStep.next(DatagridRenderStep.CLEAR_WIDTHS);
-      expect(context.clarityElement.style.width).toBe('123px');
     });
 
     it('does not set the width when the user declared a strict one', function() {
       domAdapter._scrollWidth = 123;
-      context.clarityDirective.strictWidth = 24;
-      expect(context.clarityDirective.computeWidth()).toBe(24);
-      context.clarityDirective.setWidth(24);
-      expect(context.clarityElement.style.width).toBeFalsy();
+      stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123, strictWidth: 24 });
       expect(context.clarityElement.classList).toContain('datagrid-fixed-width');
-      delete context.clarityDirective.strictWidth;
-      expect(context.clarityDirective.computeWidth()).toBe(123);
-      context.clarityDirective.setWidth(123);
+      expect(context.clarityElement.style.width).toBeFalsy();
+
+      stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123, strictWidth: 0 });
       expect(context.clarityElement.style.width).toBe('123px');
       expect(context.clarityElement.classList).not.toContain('datagrid-fixed-width');
     });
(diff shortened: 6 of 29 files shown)


src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -100,6 +109,15 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
     }
   }
 
+  private setupColumns() {
+    this.columnsService.columns = this.headers.toArray().map(() => new BehaviorSubject<DatagridColumnState>({}));
```
Shijir: I think you can directly transform `QueryList` with `map()` without turning it into array.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -4,16 +4,33 @@
...
 import { DatagridRenderStep } from '../enums/render-step.enum';
 import { ColumnResizerService } from '../providers/column-resizer.service';
 import { STRICT_WIDTH_CLASS } from './constants';
 import { DatagridRenderOrganizer } from './render-organizer';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 @Directive({ selector: 'clr-dg-column', providers: [ColumnResizerService] })
 export class DatagridHeaderRenderer implements OnDestroy {
+  stateSubscription: Subscription;
```
Shijir: they should be private, right?

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -0,0 +1,40 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Injectable, OnDestroy } from '@angular/core';
+import { BehaviorSubject, Subscription } from 'rxjs';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridRenderStep } from '../enums/render-step.enum';
+import { DatagridRenderOrganizer } from '../render/render-organizer';
+
+@Injectable()
+export class ColumnsService implements OnDestroy {
```
Shijir: @gnomeontherun IMO, we are making it too complex. I was imagining a central state mangager class to be much simpler as holding properties for a very few things such as an array of column state objects and observable that any column, cell or whatever that need to subscribe to in order to get state changes. 

Currently, this class holds `BehaviorSubject` for each columns state, and also looks like it's trying to manage the subscriptions in itself too. Also, any column or cell that tries to subsribe to the state changes needs to instantiate another `BehaviorSubject` as it looks in `header-renderer.ts` as well as `cell-renderer.ts`. 

My questions is, why did we need to take this route instead of the way I mentioned in the beginning?

src/clr-angular/data/datagrid/enums/render-step.enum.ts
```diff
@@ -11,4 +11,5 @@ export enum DatagridRenderStep {
   CLEAR_WIDTHS, // Note this is listened to by both cells and columns
   COMPUTE_COLUMN_WIDTHS,
   DETECT_STRICT_WIDTHS,
+  RESET,
```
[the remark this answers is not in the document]
gnomeontherun: Previously, the render organizer reset the widths when a full resize was triggered. This is a new step to accomplish the same task by alerting the columns service to reset the calculated widths.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: The idea is that a column is a thing that has state but that state matters to all of the cells in . that column, and only cells in that column need to know the current state of that column. There are only as many `BehaviorSubject`s as there are columns, and when the list of columns changes the corresponding column observable is passed to the cells (and subscribed to at the cell level). If there are 101 cells (header and 100 rows) in a column, they all subscribe to the same subject and it becomes simple to alert to a single column about a state change. For width changes from resizing, we only need to change the affected column instead of recompute the whole datagrid.

Having just a central service with an array of widths, an array of orderings, and so forth does not keep the column state as a single unit (which is basically what the current render organizer was doing for widths). 

Maybe I'm not clear on your question, if so we can talk about it separately.

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
+
+export interface DatagridColumnState {
```
hippee-lee: Is the intent with this to add other things here?  e.g hidden state, master detail mode (if thats the correct description).

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
```diff
@@ -0,0 +1,12 @@
...
+  actions?: DatagridColumnChanges[];
```
hippee-lee: This looks like it saves the specific states to replay them on demand. Two comments. 

1. They way Header/Cell renderers react to changes is switching on each action. Is there any concern about that growing to verbose and or complicated requireing many helper utility methods in both renderers? 
2. naming nit: is actions the best way to descibe these changes that need to be applied?

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -4,16 +4,33 @@
...
+  get columnState(): BehaviorSubject<DatagridColumnState> {
+    return this._columnState;
+  }
+
+  set columnState(columnState: BehaviorSubject<DatagridColumnState>) {
+    if (this.stateSubscription) {
+      this.stateSubscription.unsubscribe();
+    }
+    this._columnState = columnState;
+    this.stateSubscription = columnState.asObservable().subscribe(state => this.stateChanges(state));
```
hippee-lee: Would it be more natural to add a getter to the columnsState instead of directly exposing the BehaviorSubject? Same comment on `cell-renderer`

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -42,6 +59,23 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
+    if (this.stateSubscription) {
+      this.stateSubscription.unsubscribe();
+    }
+  }
+
+  private stateChanges(state: DatagridColumnState) {
+    if (state.actions && state.actions.length) {
+      state.actions.forEach(action => {
```
hippee-lee: Would it make sense to add a util method to the columns service that can centralize the case detection for DatagridColumnsChanges type instead of repeating them in all of the renderers?

src/clr-angular/data/datagrid/providers/columns.service.ts
hippee-lee: This approach makes sense to me. It collects the column states into one place. It also allows the main renderer to create/init the states when it it knows everything is ready for it. 

It seems like a good approach to have cells/headers subscribe to the changes I wasn't sure about exposing the BehaviorSubjects directly as commented.

src/clr-angular/data/datagrid/render/row-renderer.ts
```diff
@@ -3,46 +3,27 @@
...
     this.cells.forEach((cell, index) => {
-      const width = this.organizer.widths[index];
-      cell.setWidth(width.strict, width.px);
-    });
-  }
-
-  ngAfterContentInit() {
-    this.cells.changes.subscribe(() => {
-      this.setWidths();
+      cell.columnState = this.columnsService.columns[index];
```
youdz: Do we have a case where the cells of a row can change (I mean `this.cells.changes` fires) without the headers of the datagrid changing? If so, then you're missing a subscription here for cell ContentChildren changes.

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -4,15 +4,32 @@
...
+  get columnState(): BehaviorSubject<DatagridColumnState> {
+    return this._columnState;
+  }
+
+  set columnState(columnState: BehaviorSubject<DatagridColumnState>) {
+    if (this.stateSubscription) {
+      this.stateSubscription.unsubscribe();
+    }
+    this._columnState = columnState;
+    this.stateSubscription = columnState.asObservable().subscribe(state => this.stateChanges(state));
```
youdz: `asObservable()` doesn't do anything here. It's only useful for type safety if you want to expose a public `Observable` and keep a private `Subject`. But the code for `asObservable()` in RxJS is really just:
```typescript
asObservable() {
  return this as Observable<T>;
}
```

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -22,19 +39,36 @@ export class DatagridCellRenderer implements OnDestroy {
...
   }
 
   private clearWidth() {
     this.renderer.removeClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
     this.renderer.setStyle(this.el.nativeElement, 'width', null);
   }
 
-  public setWidth(strict: boolean, value: number) {
-    if (strict) {
+  private setWidth(state) {
```
youdz: `state: DatagridColumnState`

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -4,16 +4,33 @@
...
 @Directive({ selector: 'clr-dg-column', providers: [ColumnResizerService] })
 export class DatagridHeaderRenderer implements OnDestroy {
+  stateSubscription: Subscription;
+  _columnState: BehaviorSubject<DatagridColumnState>;
+
+  get columnState(): BehaviorSubject<DatagridColumnState> {
+    return this._columnState;
+  }
+
+  set columnState(columnState: BehaviorSubject<DatagridColumnState>) {
```
youdz: Can you add a comment that we should figure out a way to avoid the code duplication between `DatagridCellRenderer` and `DatagridHeaderRenderer` in the future? They started with a few lines of common code, but it's now grown to dozens.

Not something we have to do now, but a `TODO` comment lets us remember it next time we add another feature in here.

src/clr-angular/data/datagrid/render/main-renderer.ts
youdz: Dynamic columns with `*ngIf` or changing `*ngFor` will lose all information of the current column state. I'm pretty sure in the current version this would be pretty bad anyway, but I figured I'd point it out if anyone has a good idea. I don't think it's critical, except maybe for columns that the user manually resized.

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -22,19 +39,36 @@ export class DatagridCellRenderer implements OnDestroy {
...
+    if (state.strictWidth) {
       this.renderer.addClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
     } else {
       this.renderer.removeClass(this.el.nativeElement, STRICT_WIDTH_CLASS);
     }
-    this.renderer.setStyle(this.el.nativeElement, 'width', value + 'px');
+    this.renderer.setStyle(this.el.nativeElement, 'width', state.width + 'px');
```
youdz: In your new version, you're using a number `strictWidth` that overrides the `width` of the column state. Yet you're still using this logic as if `strictWidth` was a boolean, ignoring the actual number value and still setting the `width` value.

I would go back to a boolean because I think it's clearer, but right now you're in between and it's very confusing.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -61,6 +61,7 @@ export class ClrDatagridColumnSeparator {
 
   private redFlagTracker(resizeTrackerEl: HTMLElement) {
     let isWithinMaxResizeRange: boolean;
+    // @TODO(JEREMY) Review this, it will always be true because above is always null
     if (isWithinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
```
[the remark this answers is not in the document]
Shijir: Hee sorry, overlooked this one. it should have been a property instead of a variable. It will work but we are setting style through `Renderer2` constantly during resizing.

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
gnomeontherun: Would you prefer `changes` instead of `actions`?

src/clr-angular/data/datagrid/render/header-renderer.ts
gnomeontherun: There are some duplicated lines in the renderers now that I am thinking I'll extract over time as the need arises, for now I wasn't too worried about it as it makes this change more complex.

src/clr-angular/data/datagrid/render/header-renderer.ts
gnomeontherun: I'm not sure I understand, `columnState` is a BehaviorSubject. This is to autosubscribe to it anytime the state observable is passed back down into the cell (like if a column was hidden).

src/clr-angular/data/datagrid/render/row-renderer.ts
gnomeontherun: True, not a usual case I would guess but I'm sure it happens.

src/clr-angular/data/datagrid/render/main-renderer.ts
gnomeontherun: I had thought about this, but the current behavior will lose the column state too. We'd have to somehow know to cache that state and carry it over, something I can't see a good solution for without the app passing some configuration. Something for another day I guess.

src/clr-angular/data/datagrid/render/cell-renderer.ts
gnomeontherun: Ah this is a tricky one. The cell previously had a `strictWidth` property that got populated as well as a `strict` boolean. I was trying to condense two properties into one and move that up into the state instead. I haven't pulled this through yet fully (was trying to limit changes) but can work on that too.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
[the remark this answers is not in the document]
gnomeontherun: I figured it was meant to be a property, but didn't spend the time to dive deeper.

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
hippee-lee: `changes` makes more sense to me since its an array of `DatagridColumnChanges`'s. It wouldn't be something that I would feel strong enough about to block if actions is more natural to everyone else.

src/clr-angular/data/datagrid/render/header-renderer.ts
hippee-lee: That makes sense when I see that columnsState comes from the array of columnService.columns.

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    });
+  }
+
+  // Helper method to emit a change to a column only when there is an actual diff to process for that column
+  emitStateChange(columnIndex: number, diff: Partial<DatagridColumnState>) {
+    if (!this.columns[columnIndex]) {
+      this.columns[columnIndex] = new BehaviorSubject<DatagridColumnState>({});
+    }
+    const current = this.columns[columnIndex].value;
+    const hasChange = Object.keys(diff).find(key => diff[key] !== current[key]);
```
Shijir: I don't think `.find()` works in IE11. I've used `filter()` instead.

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -109,61 +115,46 @@ export default function(): void {
...
-      expect(context.clarityDirective.computeWidth()).toBe(24);
-      context.clarityDirective.setWidth(24);
-      expect(context.clarityElement.style.width).toBeFalsy();
+      stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123, strictWidth: 24 });
+      expect(context.clarityElement.style.width).toBe('');
       expect(context.clarityElement.classList).toContain('datagrid-fixed-width');
-      delete context.clarityDirective.strictWidth;
-      expect(context.clarityDirective.computeWidth()).toBe(123);
-      context.clarityDirective.setWidth(123);
+      expect(context.clarityElement.style.width).toBeFalsy();
```
youdz: Duplicate of line 123.

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -30,6 +30,9 @@ export class ColumnsService implements OnDestroy {
 
   // Helper method to emit a change to a column only when there is an actual diff to process for that column
   emitStateChange(columnIndex: number, diff: Partial<DatagridColumnState>) {
+    if (!this.columns[columnIndex]) {
+      this.columns[columnIndex] = new BehaviorSubject<DatagridColumnState>({});
```
youdz: I'm trying to understand the point of this. I'm guessing there might be a race condition somewhere where we try to change the state of a column before the datagrid's main renderer has the actual list of all the columns.

But shouldn't we just return in this case? We're creating a BehaviorSubject that no one subscribes to, since the `setupColumns()` method of the main renderer just replaces this subject with a new one.

src/clr-angular/data/datagrid/render/main-renderer.spec.ts
```diff
@@ -95,14 +99,15 @@ export default function(): void {
         expect(computeWidthSpy.calls.count()).toBe(0);
         // Too lazy to do something other than casting right now.
```
youdz: I like old comments that don't make sense anymore. 😁

src/clr-angular/data/datagrid/render/row-renderer.spec.ts
```diff
@@ -3,94 +3,51 @@
...
-      expect(cellWidthSpy.calls.allArgs()).toEqual([[false, 42], [true, 24]]);
-    });
-
-    it("doesn't set the width when the organizer doesn't have them yet", function() {
-      organizer.widths = [];
-      organizer.updateRenderStep.next(DatagridRenderStep.ALIGN_COLUMNS);
-      expect(cellWidthSpy).not.toHaveBeenCalled();
-    });
-
-    it('sets the widths of the cells when created after the widths have been computed', function() {
```
youdz: Losing this test seems like a pretty big deal to me. It's important that rows added _after_ the width have been computed correctly render with the right width. It's the basis of how our rendering works with pagination, for instance.

To be honest, right now I'm not even sure this works in the current state of this refactor. I do not see how moving to the next page preserves the computed columns width and renders them on the new rows.

src/clr-angular/data/datagrid/render/constants.ts
```diff
@@ -4,6 +4,4 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-export const NO_LAYOUT_CLASS = 'datagrid-no-layout';
```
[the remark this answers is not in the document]
gnomeontherun: Do you want me to keep this with a TODO?

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -53,17 +53,22 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
+
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
     if (this.stateSubscription) {
       this.stateSubscription.unsubscribe();
     }
   }
 
+  ngOnInit() {
+    this.detectStrictWidth();
```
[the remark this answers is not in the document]
gnomeontherun: This is an artifact of the refactoring that can go away.

src/clr-angular/data/datagrid/enums/render-step.enum.ts
youdz: The more I read this, the more I think `RESET` and `CLEAR_WIDTHS` are exactly the same step. Not sure why we still have the two of them.

gnomeontherun: I added it per your request 🥇 but will remove.

src/clr-angular/data/datagrid/render/row-renderer.ts
```diff
@@ -3,46 +3,20 @@
...
+  setupColumns() {
     this.cells.forEach((cell, index) => {
-      const width = this.organizer.widths[index];
-      cell.setWidth(width.strict, width.px);
+      cell.columnState = this.columnsService.columns[index];
     });
   }
-
-  ngAfterContentInit() {
-    this.cells.changes.subscribe(() => {
```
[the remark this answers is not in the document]
gnomeontherun: This needs to happen now in the main renderer.

src/clr-angular/data/datagrid/render/constants.ts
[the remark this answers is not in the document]
youdz: Sure, I think that's be helpful. Thanks.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: Yeah we can just `return`. Fixed.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: Fixed

src/clr-angular/data/datagrid/render/row-renderer.spec.ts
gnomeontherun: I've put a test back in for something similar.

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -4,15 +4,33 @@
...
 import { DatagridRenderOrganizer } from './render-organizer';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 @Directive({ selector: 'clr-dg-cell' })
 export class DatagridCellRenderer implements OnDestroy {
+  private stateSubscription: Subscription;
+  private _columnState: BehaviorSubject<DatagridColumnState>;
+
+  get columnState(): BehaviorSubject<DatagridColumnState> {
```
Shijir: Do we need to have a getter for this? Nothing should be subscribing to from outside, right?

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -0,0 +1,45 @@
...
+      this.emitStateChange(index, { width: 0 });
+    });
+  }
+
+  // Helper method to emit a change to a column only when there is an actual diff to process for that column
+  emitStateChange(columnIndex: number, diff: Partial<DatagridColumnState>) {
+    if (!this.columns[columnIndex]) {
+      return;
+    }
+    const current = this.columns[columnIndex].value;
```
Shijir: I think we will try to access the `.value` many times from the `ColumnService`. So I think it's ideal to have a method for this like:

```ts
inspectStateOf(columnIndex: number) {
   return this.columns[columnIndex].value
}
```

src/clr-angular/data/datagrid/render/row-renderer.ts
```diff
@@ -3,46 +3,32 @@
...
-    this.subscriptions.forEach(sub => sub.unsubscribe());
-  }
+export class DatagridRowRenderer implements AfterViewInit {
+  @ContentChildren(DatagridCellRenderer) private cells: QueryList<DatagridCellRenderer>;
 
-  @ContentChildren(DatagridCellRenderer) cells: QueryList<DatagridCellRenderer>;
+  constructor(private columnsService: ColumnsService) {}
 
-  private setWidths() {
-    if (this.organizer.widths.length !== this.cells.length) {
```
Shijir: Have you checked it with dynamic columns? I think we need this guard. This will make datagrid widths broken in the case dynamics columns.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -4,16 +4,33 @@
...
 import { DatagridRenderOrganizer } from './render-organizer';
+import { DatagridColumnState } from '../interfaces/column-state.interface';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 @Directive({ selector: 'clr-dg-column', providers: [ColumnResizerService] })
 export class DatagridHeaderRenderer implements OnDestroy {
+  private stateSubscription: Subscription;
+  private _columnState: BehaviorSubject<DatagridColumnState>;
+
+  get columnState(): BehaviorSubject<DatagridColumnState> {
```
Shijir: Looks like we are not using this `columnState` property here either. Do we even need to save the `BehaviorSubject ` to a property?

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -4,16 +4,33 @@
...
+    return this._columnState;
+  }
+
+  set columnState(columnState: BehaviorSubject<DatagridColumnState>) {
```
Shijir: In dynamic columns, if we toggle additional columns, we will run this method for every column. Should we have a guard to check their references, exit out early if the `columnState` argument has the same reference as whatever saved to `_columnState`?

src/clr-angular/data/datagrid/render/cell-renderer.ts
gnomeontherun: The reason for this was that you could get the column state from here, though it is not in use right now. I can remove this for now unless it becomes needed.

src/clr-angular/data/datagrid/render/row-renderer.ts
gnomeontherun: Yes I've tested with `*ngIf` on columns and toggling them. We reset the state when a column is removed and reattached, which is true before this change as well. It does calculate an initial strict width if it was set on the column like `[style.width.px]="100"`, otherwise reiszed widths are cleared.

src/clr-angular/data/datagrid/render/header-renderer.ts
gnomeontherun: When columns change, we do need to run this method to ensure they sync their latest states. If a new column arrives it might change the widths of other columns (such as last column has the flex column, so another one could take on a specific width defined). The frequency of which columns change is not a concern in performance, as the rows are more likely to change with pagination or data refreshing.

src/clr-angular/data/datagrid/render/row-renderer.ts
Shijir: Sounds good. Thanks for confirming.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: I think we can optimize this later if we do, otherwise its premature at this stage what we need here.

PR #3135, opened 2019-02-22
[NG] fix a11y issue password input

### LINKED ISSUE #3122: Password field: "reveal/hide password" is not keyboard accessible
## Describe the bug
Password field: "reveal/hide password" is not keyboard accessible
(the icon cannot be navigated with the keyboard)

## How to reproduce
1.  Go to https://clarity.design/documentation/password
2.  Click in the basic password example
3.  Press TAB-key

Result: Focus navigates to the next password field.

## Expected behavior
Focus should move to the eye-icon
`<space>, <enter>` key should reveal/hide password.
-> Similar to Datepicker https://clarity.design/documentation/datepicker

## Versions
* Clarity: 1.0
* Device: Desktop, Notebook
* OS: Windows 10
* Browser: Chrome

coryrylan:
Make show/hide icon button/focusable

closes #3122

### DIFF
--- a/src/clr-angular/forms/password/password-container.ts
+++ b/src/clr-angular/forms/password/password-container.ts
@@ -33,16 +33,22 @@ export const TOGGLE_SERVICE_PROVIDER = { provide: TOGGLE_SERVICE, useFactory: To
       <div class="clr-input-wrapper">
         <div class="clr-input-group" [class.clr-focus]="focus">
           <ng-content select="[clrPassword]"></ng-content>
-          <clr-icon *ngIf="!show && clrToggle"
-            shape="eye" 
-            class="clr-input-group-icon-action"
+          <button
+            *ngIf="!show && clrToggle"
+            (click)="toggle()"
             [attr.title]="commonStrings.show"
-            (click)="toggle()"></clr-icon>
-          <clr-icon *ngIf="show && clrToggle" 
-            shape="eye-hide"
             class="clr-input-group-icon-action"
+            type="button">
+            <clr-icon shape="eye"></clr-icon>
+          </button>
+          <button
+            *ngIf="show && clrToggle"
+            (click)="toggle()"
             [attr.title]="commonStrings.hide"
-            (click)="toggle()"></clr-icon>
+            class="clr-input-group-icon-action"
+            type="button">
+            <clr-icon shape="eye-hide"></clr-icon>
+          </button>
         </div>
         <clr-icon *ngIf="invalid" class="clr-validate-icon" shape="exclamation-circle" aria-hidden="true"></clr-icon>
       </div>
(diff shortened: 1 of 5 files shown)


src/clr-angular/forms/password/password-container.ts
```diff
@@ -33,16 +33,22 @@ export const TOGGLE_SERVICE_PROVIDER = { provide: TOGGLE_SERVICE, useFactory: To
       <div class="clr-input-wrapper">
         <div class="clr-input-group" [class.clr-focus]="focus">
           <ng-content select="[clrPassword]"></ng-content>
-          <clr-icon *ngIf="!show && clrToggle"
-            shape="eye" 
-            class="clr-input-group-icon-action"
+          <button
```
Jinnie: I think we could benefit an aria-label attribute here. A visually impaired user can now focus and toggle a button that he won't know what's it for, I even don't think he can recognize the current state.
Then comes the question about i18n of the message :) 
So I'm not really sure that we're not opening more problems than solving by choosing to implement this fix.
Would be great to get some more feedback from the team before proceeding.

youdz: I second this. Regarding the message, it's one of the few that needs to go into our i18n strings service, no way around it.

src/clr-angular/forms/password/password-container.ts
```diff
@@ -33,16 +33,22 @@ export const TOGGLE_SERVICE_PROVIDER = { provide: TOGGLE_SERVICE, useFactory: To
...
+            (click)="toggle()"
             [attr.title]="commonStrings.show"
-            (click)="toggle()"></clr-icon>
-          <clr-icon *ngIf="show && clrToggle" 
-            shape="eye-hide"
             class="clr-input-group-icon-action"
+            type="button">
+            <clr-icon shape="eye"></clr-icon>
+          </button>
+          <button
```
youdz: You're duplicating the button, which is a bit silly. You should have:
```<button>
  <clr-icon *ngIf>
  <clr-icon *ngIf>
</button>```

src/clr-angular/forms/password/password-container.ts
Jinnie: Doesn't the `Giant Warning` you talk about in #3136 apply to this case too? 😃 
https://www.w3.org/TR/html/dom.html#the-title-attribute
If not, where's the difference?

coryrylan: Seems adding aria-label is a good idea. I tested the title on both cases with a couple of screen reader with the same concerns and it was somewhat inconclusive. ChromeVox and Voice Over both do read titles of buttons but researching JAWS does not without a special setting. For this case drop the title altogether? 

You could argue title is to inconsistent for a11y and just not support it in any of the clarity components as @youdz suggested in #3136

youdz: Our icons cheat, we **did** make the title accessible. They're custom elements and we control what they do, so setting the title on them actually writes the title in the `<title>` element of the icon's SVG. It's not just and attribute.

Nice try though. :trollface:

coryrylan: Sounds good, I'll make the change to move the title back to the icon as the fix.

PR #3136, opened 2019-02-22
[NG] fix button title for button group

### LINKED ISSUE #3132: Unable to add title attribute to clr-button.
## Describe the bug

I need to add title attribute to clr-button, so I use ```<clr-button class="btn" title="title">Add</clr-button>```, it doesn't work.
And I tried to use custom attribute directive to complete it, but it still failed.

## How to reproduce
Reproduce this issue [here](https://stackblitz.com/edit/custom-title-directive?embed=1&file=src/app/app.component.html)

## Expected behavior
1. it would be nice if I could add title by using ```<clr-button class="btn" title="title">Add</clr-button>```.
2. custom directive works on ```<clr-button></clr-button>```

## Versions

**App**

* Angular: [6.0.0]
* Node: [8.12.0]
* Clarity: [0.12.0]

**Device:**

* Type: [MacBook]
* OS: [iOS]
* Browser [chrome]

coryrylan:
Properly set title attribute for button in button group.

closes #3132

### DIFF
--- a/src/clr-angular/button/button-group/button.ts
+++ b/src/clr-angular/button/button-group/button.ts
@@ -20,7 +20,7 @@ import { ButtonInGroupService } from '../providers/button-in-group.service';
                 [attr.type]="type"
                 [attr.name]="name"
                 [attr.disabled]="disabled"
-                [attr.title]="title">
+                [attr.id]="id">
                 <span class="spinner spinner-inline" *ngIf="loading"></span>
                 <ng-content></ng-content>
             </button>
@@ -101,16 +101,16 @@ export class ClrButton implements LoadingListener {
     }
   }
 
-  private _title: string = null;
+  private _id: string = null;
 
-  get title(): string {
-    return this._title;
+  get id(): string {
+    return this._id;
   }
 
-  @Input('title')
-  set title(value: string) {
+  @Input('id')
+  set id(value: string) {
     if (typeof value === 'string') {
-      this._title = value;
+      this._id = value;
     }
   }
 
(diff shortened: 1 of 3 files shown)


coryrylan:
I tested the title on both cases with a couple of screen readers with the same concerns and it was somewhat inconclusive. ChromeVox and Voice Over both do read titles of buttons but researching JAWS does not without a special setting (wasnt able actually to test with windows, just from research). Agree we may not want to merge this in. Related issue #3135

coryrylan:
Stated in issue #3132 ids are not passed in either which can cause issues for e2e automated testing. `id` support sounds like a valid use case and I think it might make sense to change this PR to `id` instead of `title` support and drop `title` support all together. Any thoughts @youdz

youdz:
Yes, `id` or other attributes that it makes sense to proxy to the button itself are a valid use case I think.

That's also why we try to keep these interactive elements on the app's side, but I guess this one slipped through. So updating this PR to `id` is 👌 by me.

src/clr-angular/button/button-group/button.ts
```diff
@@ -19,7 +19,8 @@ import { ButtonInGroupService } from '../providers/button-in-group.service';
                 (click)="emitClick()"
                 [attr.type]="type"
                 [attr.name]="name"
-                [attr.disabled]="disabled">
+                [attr.disabled]="disabled"
+                [attr.id]="id">
```
youdz: Nitpick: you can just bind to `[id]` on a native element. That sets the id property instead of the attribute, which in turn sets the attribute. Honestly, I'm not even sure which is supposedly better.

coryrylan: Missed this, I prefer properties by default. I think regular html elements will reflect them back and forth but custom elements you have to manually handle this. Angular I think reflects the property to the attribute but not the other way around. I'll get this updated, I think this might change the setter signature a bit.

PR #3146, opened 2019-02-27
[CORE] core(Angular v8beta.5): upgrade dependencies for Angular 8.

hippee-lee:
This change will add the angular v8.beta.5 (and associated dev dependencies for) for the Clarity v2.0 release. 


package.json
```diff
@@ -119,7 +119,7 @@
     "husky": "1.0.0-rc.8",
     "jasmine-core": "2.99.1",
     "jasmine-expect": "3.8.3",
-    "jasmine-spec-reporter": "4.2.1",
+    "jasmine-spec-reporter": "4",
```
Jinnie: How does this version work!?

gnomeontherun: You can put any valid tag or semver version, which will be resolved here as anything 4.x. We should fix this though to be specific.

hippee-lee: yeah, I missed putting that back when I was looked at what testing deps could also be upgraded. I'll revert and submit a separate commit for testing deps if they are applicable.

PR #3171, opened 2019-03-05
feat(schematics) automatically support installing @next

gnomeontherun:
This will setup the Clarity @next version when trying to use a n+1 version of Angular support.

### Release Notes

Updated `ng add @clr/angular` to detect if you are using a version of Angular that is ahead of our latest version (such as beta releases) and automatically install our beta/rc version as well to avoid mismatches.

### DIFF
--- a/src/schematics/src/add/index.ts
+++ b/src/schematics/src/add/index.ts
@@ -4,6 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+// @TODO REWRITE THIS
+// This is a mess, because it was written early on and could be much better.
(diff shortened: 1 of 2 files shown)


src/schematics/src/add/index.ts
```diff
@@ -4,6 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+// @TODO REWRITE THIS
+// This is a mess, because it was written early on and could be much better.
+// Also, no real docs existed then or now, so good luck brave soldier.
```
jeeyun: 😆

PR #3175, opened 2019-03-05
[NG] Fix pagination init lifecycle

### LINKED ISSUE #3082: Datagrid high records delay to render
## Describe the bug

When the Datagrid is processing client side, even with pagination, it is taking up to 3 seconds to render datagrid.

this is happening with my app that has the data stored in a store, i have timestamped the output of the last bit of processing on the component, to when the screen becomes interactive at 3-4 seconds.

## How to reproduce

1.  Go to https://clarity.design/documentation/datagrid/full
2.  set the number of users to 3100
3.  Click Apply
4.  the browser will become render blocked.

## Expected behavior

This did not happen with the previous version:
https://vmware.github.io/clarity/documentation/v0.13/datagrid/full
The Datagrid should be rendered with the 10 (page size) items displayed with similar speed to a datagrid with similar size.

## Versions

**App**

* Angular: v7.1.4
* Node: v8.11.1
* Clarity: v1.0.3-patch

**Device:**

* Type: HP Desktop
* OS: Windows 10
* Browser: Chrome 
* Version: 71.0.3578.98 (Official Build) (64-bit)

youdz:
Because of a lifecycle quirk, we would instantiate as many ClrDatagridRow components
as we would have in the entire dataset. They wouldn't go throught their full Angular
lifecycle, but they would be instantiated nonetheless.
This commit fixes this by making the Items provider wait for the page size to be
available before emitting any changes.

Fixes #3082.

### DIFF
--- a/src/clr-angular/data/datagrid/providers/items.ts
+++ b/src/clr-angular/data/datagrid/providers/items.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -178,7 +178,8 @@ export class Items<T = any> {
    * Extracts the current page from the sorted list
    */
   private _changePage() {
-    if (this.uninitialized) {
+    // If we know we have pagination but the page size hasn't been set yet, we wait for it.
+    if (this.uninitialized || (this._page.activated && this._page.size === 0)) {
       return;
     }
     if (this._page.size > 0) {
--- a/src/clr-angular/data/datagrid/providers/page.ts
+++ b/src/clr-angular/data/datagrid/providers/page.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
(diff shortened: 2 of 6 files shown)


src/clr-angular/data/datagrid/providers/page.ts
```diff
@@ -12,6 +12,8 @@ import { StateDebouncer } from './state-debouncer.provider';
 export class Page {
   constructor(private stateDebouncer: StateDebouncer) {}
 
+  public activated = false;
```
hippee-lee: Since this is public, should it have a test?

src/clr-angular/data/datagrid/providers/items.ts
```diff
@@ -178,7 +178,8 @@ export class Items<T = any> {
    * Extracts the current page from the sorted list
    */
   private _changePage() {
-    if (this.uninitialized) {
+    // If we know we have pagination but the page size hasn't been set yet, we wait for it.
+    if (this.uninitialized || (this._page.activated && this._page.size === 0)) {
```
gnomeontherun: Do we have a test to cover this? I haven't looked recently, but this is a new code path that we could extend an existing test to validate.

src/clr-angular/data/datagrid/providers/page.ts
gnomeontherun: The property itself is not important to test, but the side effects of it being set. I think the test covers the cases except when changePage is triggered.

youdz:
@hippee-lee: You're right, in this case the selector overload is pointless, my bad. I'm switching to an attribute directive since I control the test component anyway.

For our components though, we sadly don't have much choice since Angular doesn't offer a good composition pattern yet. But I understand the concern, maybe we should think of a way to document overloads somewhere in the code.

PR #3198, opened 2019-03-08
chore(dev-dependencies): Upgrades non-angular dev-deps

hippee-lee:
- All dev-deps that get generated by the angular cli were only updated to the latest patch version (Angular cli dev-deps use tilde(~) )
- All other dependencies were updated to the latest version as published on npm

Ran build, test:travis, unit tests and all visual regression tests. 


hippee-lee:
> Let's bump things so we get rid of as many messages from `npm audit`

Second commit address most of the warnings. 

## Summary of investigation for the remaining items:
### High Warnings (2)
*The two high warnings are in gemini* b/c it uses handlebars in gemini-coverage. I presume when it creates the html reports.
- Prototype Pollution - https://npmjs.com/advisories/755: Templates may alter an Objects’ prototype, thus allowing an attacker to execute arbitrary code on the server.
- XSS - https://npmjs.com/advisories/61: Versions of handlebars prior to 4.0.0 are affected by a cross-site scripting vulnerability when attributes in handlebar templates are not quoted

_I didn’t look to see how gemini-coverage uses it in its templates but will if you think its needed._

### Low warnings (6)
- uglify-js is used in handlebars (2 low warnings)
- braces is used in *csso-cli dependency (micromatch)* and *gemini && gemini-core dependency (micromatch)* and *lite-server dependency (micromatch)* (4 warnings in total)

src/website/src/app/documentation/demos/checkboxes/ng/disabled.html
```diff
@@ -1,4 +1,4 @@
-<clr-checkbox-container>
+`<clr-checkbox-container>
```
gnomeontherun: Do you know what dependency changed this formatting? Prettier remained the same.

hippee-lee: I updated it when I was getting website build issues. 
Now, I don't seem to be able to revert them back to master, prettier seems to change the files to something different from master. 

Do you think I should just close this PR and submit another one sans formatting changes?

gnomeontherun:
You should be able to revert those changes and force push, no need to make a new PR.

hippee-lee:
@gnomeontherun - I got the html files reverted. The only significant change, besides the version bumps,  is the use of `raw-loader` in the demos.

Issue #3184, opened 2019-03-06
Button without accessible name


chlane:
## Description
The button that expands and collapses the "inbox" example on https://clarity.design/documentation/badges does not have an accessible name. Assistive technology users will not be able to identify this button or understand its purpose. 

See the button in question below.

<button class="clr-treenode-caret ng-tns-c2-0 ng-star-inserted" type="button" aria-expanded="false"><clr-icon class="clr-treenode-caret-icon" shape="caret" dir="right"><svg version="1.1" viewBox="0 0 36 36" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" focusable="false" aria-hidden="true" role="img"><path class="clr-i-outline clr-i-outline-path-1" d="M29.52,22.52,18,10.6,6.48,22.52a1.7,1.7,0,0,0,2.45,2.36L18,15.49l9.08,9.39a1.7,1.7,0,0,0,2.45-2.36Z"></path></svg></clr-icon></button>

## Location
https://clarity.design/documentation/badges

## Priority
High

## Recommended fix
Use aria-label to provide an accessible name for this control. Provide scripting to update the  aria-label 
value according to the value of aria-expanded as seen below. 

```
<!-- when collapsed, the label should contain the word "expand" -->
<button aria-label="expand inbox" class="clr-treenode-caret ng-tns-c2-0 ng-star-inserted" type="button" aria-expanded="false">
```

```
<!-- when expanded, the label should contain the word "collapse"-->
<button aria-label="collapse inbox" class="clr-treenode-caret ng-tns-c2-0 ng-star-inserted" type="button" aria-expanded="true">
```

## LOE
Medium

coryrylan:
fixed in https://github.com/vmware/clarity/pull/3206

Issue #3172, opened 2019-03-05
The vertical ellipses button on the design page announces as "button" when read by a screen reader


chlane:
## Description  
The vertical ellipses button on the design page announces as "button" when read by a screen reader. This issue was reproduced using VoiceOver/Safari/iOS 12.1.4.  The word "button" alone does not describe the purpose of this control. This will confuse screen reader users.  

see the relevant HTML below.

`<button class="header-overflow-trigger" type="button"><span></span></button>`

This issue is related to https://github.com/vmware/clarity/issues/3155

This issue also occurs with the vertical ellipses button used to select changelogs at https://clarity.design/news. 

## Location     
https://clarity.design/icons Any page where the vertical ellipses button is used.

## Priority     
High

## Recommended fix      
Add text that describes the purpose of this button  by either using visible text, accessible offscreen text or aria-label.

/* .CSS class for accessible offscreen text /
.offscreenText {position: absolute;clip: rect(1px 1px 1px 1px); / IE6, IE7 */clip: rect(1px, 1px, 1px, 1px);clip-path: inset(50%);padding: 0;border: 0;height: 1px; width: 1px; overflow: hidden;white-space: nowrap;}

## LOE  
Low

coryrylan:
fixed in https://github.com/vmware/clarity/pull/3206

Issue #3155, opened 2019-03-02
The Hamburger menu button lacks accessible name


chlane:
## Description  
The Hamburger menu button lacks accessible name. Screen readers will announce the word "button" when reading this control. The user will not understand what the button does without more information.

## Location     
 Mobile iOS home page

## Priority     
High 

## Recommended fix      
Change to "main menu" by either using visual text, accessible offscreen text or aria-label.

/* .CSS class for accessible offscreen text  */
.offscreenText {position: absolute;clip: rect(1px 1px 1px 1px); /* IE6, IE7 */clip: rect(1px, 1px, 1px, 1px);clip-path: inset(50%);padding: 0;border: 0;height: 1px; width: 1px; overflow: hidden;white-space: nowrap;}

<button aria-label="main menu" class="header-hamburger-trigger" type="button">
<span></span>
</button>

## LOE  
Fast

coryrylan:
fixed in https://github.com/vmware/clarity/pull/3206

PR #3206, opened 2019-03-11
[WEBSITE] lighthouse audit fixes

### LINKED ISSUE #3154: Close button icons do not have an accessible name
## Description
When using VoiceOver/Safari/iOS 12.1.4 and swiping to the "X"  icon in the .top right corner of  the "we are hiring"  container, the user hears this control announced as the word "button".  Screen reader users will not know what this button does without providing an accessible name.

In the HTML for this button below, we can see that there is no text or attribute for this button to have an accessible name.

`<button _ngcontent-c0="" class="close" type="button"><clr-icon _ngcontent-c0="" shape="close"><svg version="1.1" viewBox="0 0 36 36" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" focusable="false" aria-hidden="true" role="img"><path class="clr-i-outline clr-i-outline-path-1" d="M19.41,18l8.29-8.29a1,1,0,0,0-1.41-1.41L18,16.59,9.71,8.29A1,1,0,0,0,8.29,9.71L16.59,18,8.29,26.29a1,1,0,1,0,1.41,1.41L18,19.41l8.29,8.29a1,1,0,0,0,1.41-1.41Z"></path></svg></clr-icon></button>`

## Location
Mobile iOS home page

## Priority
High

## Recommended fix
I recommend using aria-label to add an accessible name for this button as seen below.

`<button aria-label="close we are hiring" class="header-hamburger-trigger" type="button"><span></span></button>`

## LOE
fast

coryrylan:
Multiple small fixes for the documentation site related to SEO and a11y.

closes #3154 #3155 #3172 #3184


gnomeontherun:
The icons you added have a shadow, is that intentional?

coryrylan:
> The icons you added have a shadow, is that intentional?

Good catch, it wasn't intentional. I didnt notice the shadow in VSCode when looking at the images. I'll update it.

coryrylan:
Icons fixed

src/website/src/app/utils/external-link.directive.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive, ElementRef, HostBinding, PLATFORM_ID, Inject } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
+
+@Directive({
```
gnomeontherun: We should consider this for our a11y toolkit.

src/website/src/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "Clarity Design",
+  "short_name": "Clarity",
+  "theme_color": "#0065ab",
+  "background_color": "#ffffff",
+  "display": "standalone",
+  "Scope": "/",
```
gnomeontherun: Caps, is that an issue?

coryrylan: Hmm it was generated by a tool but I will double check to see what the spec is.

src/website/src/app/utils/external-link.directive.ts
coryrylan: Is this something that already exists? If so where?

PR #3211, opened 2019-03-12
[NG] Stop caching children for lazy recursive trees

### LINKED ISSUE #3209: [Tree View][Lazy Loading] Selection Change Events not getting triggered for tree view
## Describe the bug

We have used tree view with lazy loading , recursive tree and selection. Since there are large number of nodes normal tree can't be used for our use case. We depend on the (clrSelectedChange) event to update the child nodes on selection change. Everything works fine during the first load and first expand.  But once we collapse and expand the node the second time, the (clrSelectedChange) is not received thereafter for the child nodes. Since this event is not received thereafter the child selection can't be changed.

## How to reproduce

https://stackblitz.com/edit/clarity-dark-theme-v1-0-treeview-issue-events

1. Expand the nodes and check and uncheck the nodes
2. Check that selection change events are received
3. Collapse and Expand any node
4. Check that selection change event is not received on the childnode
    Events are now received only on the parent nodes change

## Expected behavior
Selection change event should be received on all the nodes

## Versions

**App**

* Angular: 7.2
* Clarity: 1.04

**Device:**

* OS: [Windows Server 2012]
* Browser [Chrome]
* Version 72.0.3626.121 

## Additional notes

This puts a limitation on the usage of tree node component selection when large number of nodes are present. The tree node without lazy loading can't be used because the the events take a long time to be propagated and there is no event which tells us that the propagation has completed. And with lazy loading the selection is not consistent after the second collapse and expand. Let us know if any work rounds possible for tree nodes with selection in large node setup.

youdz:
We used to fetch children for lazy recursive trees only once, but we closed their selection Observable when we shouldn't. Rather than avoid closing the Observables when we reuse models, I decided to stop caching children altogether, letting the app decide whether they want to cache or not and do it on their side. It's less intrusive and more flexible.

Fixes #3209.

### DIFF
--- a/src/clr-angular/data/tree-view/recursive-children.spec.ts
+++ b/src/clr-angular/data/tree-view/recursive-children.spec.ts
@@ -62,6 +62,11 @@ export default function(): void {
     beforeEach(function(this: Context) {
       this.featuresService = this.getProvider<TreeFeaturesService<TestNode>>(TreeFeaturesService);
       this.expandService = this.getProvider(Expand);
+      this.featuresService.recursion = {
+        template: this.hostComponent.template,
+        root: [TEST_ROOT],
+      };
+      this.detectChanges();
     });
 
     it('can create a ClrRecursiveForOfContext for a node', function(this: Context) {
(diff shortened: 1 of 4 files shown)


src/clr-angular/data/tree-view/recursive-children.spec.ts
```diff
@@ -72,50 +77,52 @@ export default function(): void {
...
-        template: this.hostComponent.template,
-        root: [TEST_ROOT],
-      };
       delete this.hostComponent.parent;
       this.hostComponent.children = TEST_ROOT.children;
       this.detectChanges();
       expect(this.clarityElement.textContent).toMatch(/A\s*B\s*C/);
     });
 
+    it('does not destroy the children when he parent becomes collapsed if the tree is eager', function(this: Context) {
```
Shijir: >he parent

omg 😱 👆 this is terrible :trollface:

youdz:
I’m happy to update names if you have something in mind. This was pretty rushed to be fair.

PR #3217, opened 2019-03-15
[NG] column state provider is injected

Shijir:


src/clr-angular/data/datagrid/providers/column-state.provider.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { InjectionToken } from '@angular/core';
+import { BehaviorSubject } from 'rxjs/index';
```
gnomeontherun: `from 'rxjs';`

PR #3220, opened 2019-03-18
[UI] header inline text and icon support

### LINKED ISSUE #2788: [Header] Nav-link with icon + text - how?
## Summary

There's no example of how to create icon + text item in header-actions.
There's only a single icon, or just text, or text with drop-down, but no icon + text.
So, it's not clear how to implement it best way.
* What is the change?
Add another class like **nav-icon-text**, which will provide correct layout and opacity for such item. Right now, clr-icon has _position:absolute_ if using **nav-link** class, or it misses opacity if not using **nav-link**.
* Why should it go in Clarity?
Because there are use cases for this, so it will reduce the amount of useless CSS and will help users follow UI guidelines without guessing margins/paddings for self-made replacements.
No
Should be

## Use case

_Provide some specific details about how this change would improve your application and use case._

Displaying username, for instance.

## Examples

_Are there any examples in other projects that demonstrate this feature or enhancement?_
image

## Workarounds
_Is there a workaround for this feature at the moment, even if it is less than ideal?_

Either not using nav-link or nav-icon class, and stying your item completely as you need, or using the nav-link class, but overriding icon position and content spacing.

coryrylan:
Refactor header scss, add tests, add
inline nav text + icon support.

closes #2788

Release Notes:
Header can now use text and icons together in a single nav link.


coryrylan:
Just realized there was a duplicate gemini test. I'll remove my duplicate I made then rebase.

PR #3225, opened 2019-03-20
[WEBSITE] spring logo update

### LINKED ISSUE #3223: Insert a temporary doodle graphic for Clarity landing page
## Replace Clarity hero image with spring time doodle

* Temporary image for Spring. It should be the same size as the current logo image: 693 x 520 
(Not urgent, but should be implemented before spring is over)
* Bonus - could we try making the panel on top of the logo 90% opacity? See below
Mockup:

coryrylan:
- Adds new spring logo to website homepage
- Automatically changes home page logo based on current season of northern hemisphere.
- optimizing/compressing the website SVG files.

closes #3223


src/website/src/styles/home.scss
```diff
@@ -104,7 +104,7 @@
     position: absolute;
     bottom: 0;
     left: 612px;
-    background-image: url('../assets/images/clarity_logo.svg');
+    background-image: url('../assets/images/clarity-logo-spring.svg');
```
gnomeontherun: Can we come up with a means to cycle through the logos based on season, with a fallback?

coryrylan: Yeah, I don't think that would be too difficult. It could check the current local date is between specified seasons/dates and toggle the background image based on that.

src/website/src/styles/home.scss
```diff
@@ -104,9 +104,17 @@
     position: absolute;
     bottom: 0;
     left: 612px;
-    background-image: url('../assets/images/clarity_logo.svg');
+    background-image: url('../assets/images/clarity-logo.svg');
     background-size: cover;
     background-position: center center;
+
+    &--spring {
```
gnomeontherun: Minor, but we are not trying to do BEM in our sass, though its in a few places.

hippee-lee: Way back a long time ago we were trying to use the [SUITECSS naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md)

But it wasn't always uniformly applied across the code base. And, I don't think we have documented it anywhere. I learned about it through discussions w/ @adityarb88 (👋🏼- we miss you :) )

PR #3227, opened 2019-03-21
Fix tree view flickering from async eventemitter

### LINKED ISSUE #3073: Tree View - Checkbox Tree "Flickering"
## Describe the bug

When the checkbox on the top node is clicked multiple times (quite quickly), the checkboxes of the child nodes start to "flicker".

## How to reproduce

[Stackblitz Example](https://stackblitz.com/edit/clarity-dark-theme-v1-0-treeview-issue)

1.  Click checkbox on node "Group A" twice very quickly (like a double click)
2.  Again, click checkbox on node "Group A" twice very quickly
3.  Note that the checkboxes of the child nodes start to flicker endlessly

## Expected behavior

No "flickering" should occur.

## Versions

**App**

* Angular: 7.2.0
* Node: ?
* Clarity: 1.0.4

**Device:**

* Type: HP zBook
* OS: Windows 10
* Browser Chrome
* Version 71.0.3578.98 (64-Bit)

## Additional notes

The problem seems to get worse the more components are present on the page. As it might be difficult to reproduce the issue on another environment, I'm attaching a screen capture from my machine.

[Tree View Flickering Video Capture.zip](https://github.com/vmware/clarity/files/2796344/Tree.View.Flickering.Video.Capture.zip)

gnomeontherun:
fix(tree) set tree view node event emitter to synchronous and don't output selected changes when setting input on node

A tree view with a number of nodes could be tricked into a state of uncertainty (flickering) by double clicking fast to trigger too many async event loops in the bindings. This stops the event propagation from firing in this specific case, but keeps it in the remainder where it is needed to update other nodes.

fixes #3073

### Release notes

Bugfix for tree view where it could trigger a flickering state by double clicking fast, or in any case where the tree tried to update before the tree had stabilized.

### DIFF
--- a/src/clr-angular/data/tree-view/tree-node.spec.ts
+++ b/src/clr-angular/data/tree-view/tree-node.spec.ts
@@ -5,7 +5,6 @@
  */
 
 import { Component, ViewChild } from '@angular/core';
-import { fakeAsync, tick } from '@angular/core/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { RecursiveTreeNodeModel } from './models/recursive-tree-node.model';
 
@@ -163,19 +162,22 @@ export default function(): void {
     describe('Template API', function() {
       spec(ClrTreeNode, TestComponent, ClrTreeViewModule, { imports: [NoopAnimationsModule, ClrIconModule] });
 
-      it(
-        'offers a [(clrSelected)] two-way binding',
-        fakeAsync(function(this: Context) {
-          this.hostComponent.selected = ClrSelectedState.SELECTED;
-          this.detectChanges();
-          expect(this.clarityDirective.selected).toBe(ClrSelectedState.SELECTED);
-          this.clarityDirective.selected = ClrSelectedState.INDETERMINATE;
-          // We need fakeAsync and tick because the EventEmitter is asynchronous
-          tick();
-          // I don't know why Typescript forces me to cast this
-          expect<ClrSelectedState>(this.hostComponent.selected).toBe(ClrSelectedState.INDETERMINATE);
-        })
-      );
+      it('offers a [(clrSelected)] two-way binding, but skips emitting output when setting input', function(this: Context) {
+        this.hostComponent.selected = ClrSelectedState.SELECTED;
+        this.detectChanges();
+        expect(this.clarityDirective.selected).toBe(ClrSelectedState.SELECTED);
+        this.clarityDirective.selected = ClrSelectedState.UNSELECTED;
+        this.detectChanges();
+        // This may seem unintuitive, but actually is due to the tree flickering issues.
+        // We don't want to emit the change in this case, or it can get confused.
+        // Here we are expecting that the output has not emitted.
+        // See https://github.com/vmware/clarity/issues/3073
+        expect<ClrSelectedState>(this.hostComponent.selected).toBe(ClrSelectedState.SELECTED);
+        // By setting the model here, we will emit the binding since it's not through the input
+        this.clarityDirective._model.setSelected(ClrSelectedState.INDETERMINATE, true, true);
+        this.detectChanges();
+        expect<ClrSelectedState>(this.hostComponent.selected).toBe(ClrSelectedState.INDETERMINATE);
+      });
 
       it('offers a [(clrExpanded)] two-way binding', function(this: Context) {
         this.hostComponent.expanded = true;
--- a/src/clr-angular/data/tree-view/tree-node.ts
+++ b/src/clr-angular/data/tree-view/tree-node.ts
@@ -19,6 +19,7 @@ import {
   SkipSelf,
 } from '@angular/core';
 import { Subscription } from 'rxjs';
+import { filter } from 'rxjs/operators';
 
 import { Expand } from '../../utils/expand/providers/expand';
 import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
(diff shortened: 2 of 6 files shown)


src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -90,12 +92,15 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
...
     }
-    // We propagate only if the tree is in smart mode
+    // We propagate only if the tree is in smart mode, and skip emitting the output when we set the input
+    // See https://github.com/vmware/clarity/issues/3073
+    this.skipEmitChange = true;
     this._model.setSelected(value, this.featuresService.eager, this.featuresService.eager);
+    this.skipEmitChange = false;
   }
 
   // We need an async EventEmitter or we will trigger chocolate errors like it's 2016.
```
youdz: Please remove this obsolete comment.

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -163,19 +165,34 @@ export default function(): void {
...
+      it('offers a [(clrSelected)] two-way binding', function(this: Context) {
+        // Here we are testing the child two-way binding, which should be updated by two-way always.
+        const child = this.fixture.debugElement.query(By.directive(ClrTreeNode));
+        this.hostComponent.selected = ClrSelectedState.SELECTED;
+        this.detectChanges();
+        expect(this.clarityDirective.selected).toBe(ClrSelectedState.SELECTED);
+        expect(child.componentInstance.selected).toBe(ClrSelectedState.SELECTED);
+        this.clarityDirective.selected = ClrSelectedState.UNSELECTED;
+        this.detectChanges();
+        expect(child.componentInstance.selected).toBe(ClrSelectedState.UNSELECTED);
```
youdz: Testing the child here seems to have nothing to do with the current unit test. I get that the line 
```typescript
this.clarityDirective.selected = ClrSelectedState.UNSELECTED;
```
doesn't actually trigger the output for the parent, but this test relies on the entire logic of downward propagation of selection for eager trees, which feels wrong for a pure two-way binding test.

I think the next unit test you added covers the two-way binding, so you can remove this one or change it to describe more specifically what you're testing (presumably chocolate errors?).

src/dev/src/app/tree-view/eager-recursive-tree/eager-recursive-tree.ts
```diff
@@ -41,12 +44,25 @@ export class EagerRecursiveTreeDemo {
     },
   ];
 
-  singleRootSelected: { [key: string]: ClrSelectedState } = {};
-  multiRootSelected: { [key: string]: ClrSelectedState } = {};
+  private buildDefaultSelected(rootMap, selectedMap: SelectedMap = {}) {
+    if (!Array.isArray(rootMap)) {
+      rootMap = [rootMap];
```
Shijir: In this case of `!Array.isArray(rootMap)`, `rootMap` is `nodelist`, no? Shouldn't we do `Array.from(rootMap)` instead of `[rootMap]`?

src/dev/src/app/tree-view/eager-recursive-tree/eager-recursive-tree.ts
```diff
@@ -41,12 +44,25 @@ export class EagerRecursiveTreeDemo {
     },
   ];
 
-  singleRootSelected: { [key: string]: ClrSelectedState } = {};
-  multiRootSelected: { [key: string]: ClrSelectedState } = {};
+  private buildDefaultSelected(rootMap, selectedMap: SelectedMap = {}) {
```
Shijir: Can we have a type on `rootMap` parameter?

src/clr-angular/data/tree-view/tree-node.spec.ts
gnomeontherun: Since the selection will propigate to the child, I'm testing that as the two way binding to get around the input change limitation I'm testing in the other test. If you're happy with the second test then I'll just remove this one.

src/dev/src/app/tree-view/eager-recursive-tree/eager-recursive-tree.ts
gnomeontherun: `rootMap` is either an array or an object, see the `singleRoot` and `multiRoot` properties on the class.

src/dev/src/app/tree-view/eager-recursive-tree/eager-recursive-tree.ts
gnomeontherun: Added.

src/clr-angular/data/tree-view/tree-node.ts
gnomeontherun: Done.

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -90,12 +92,15 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
...
-  @Output('clrSelectedChange') selectedChange = new EventEmitter<ClrSelectedState>(true);
+  @Output('clrSelectedChange') selectedChange = new EventEmitter<ClrSelectedState>(false);
```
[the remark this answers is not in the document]
gnomeontherun: I know, but decided to leave it to be clear that we want it false.

src/dev/src/app/tree-view/eager-recursive-tree/eager-recursive-tree.ts
Shijir: But on line 51, we have `this.buildDefaultSelected(node.children, selectedMap);` 
So we are basically putting `node.children`, which is `NodeList` in the array.

Issue #2904, opened 2018-12-03
Error on auto expanding row details in datagrid


veereshradder:
## Describe the bug

Have a set of rows with predefined values 'row.expand'(true or false) to expand rows automatically on initial render. Getting below console error while auto-expanding rows in clarity Datagrid.
`ERROR DOMException: Failed to execute 'animate' on 'Element': Partial keyframes are not supported.
at DatagridRowExpandAnimation.push../node_modules/@clr/angular/esm5/clr-angular.js.DatagridRowExpandAnimation.run`

To be more specific the error occure in following line, due to oldHeight is 'NaN' and this will be 'NaNpx' which is invalid value for animation.
`DatagridRowExpandAnimation.prototype.run = {
    ...
    this.running = this.el.nativeElement.animate({ height: [this.oldHeight + 'px', newHeight + 'px'], overflowY: ['hidden', 'hidden'], easing: 'ease-in-out' }, { duration: 200 });
    ...
    }`

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v1-0-88t8rr

1.  Bind data to datagrid which has property to expand row detail example 'row.expand (true | false)'
2.  On initial render datagrid logs an error in console
for large data set(around 50 records) browser will become unresponsive.

## Expected behavior

Allow to automitically expand rows on initial render

## Versions

**App**

* Angular: 6
* Node: 8
* Clarity: 0.13

**Device:**

* Type: Laptop
* OS: Windows
* Browser Chrome
* Version 70.0.3538.110

youdz:
The good news is that the app still works, you can expand/collapse the details fine.
The bad news is that I think it's a bug due to the weird lifecycle of our datagrid rendering, which indeed seems to have a problem with animations. To be more precise, I believe the details are properly instantiated at the correct time, but they're instantiated **detached** from the DOM, which causes the error mentioned in the bug report. It'll probably take quite a bit of investigation to fix.

veereshradder:
@Jinnie Sorry, yes doc mentioned not to use inputs on this clrIfExpanded directive but I found this way of using in some other place(SO I believe), never thought it could give other problem.

@Jinnie @youdz  thanks for your response. looking forward to seeing this fix in the future release.
for time being I managed to solve this with small workaround.. 
https://stackblitz.com/edit/clarity-light-theme-v1-0-88t8rr
please feel free if you have any inputs on this.

Cheers

kumars17:
[...]

PR #3229, opened 2019-03-21
Update to row-expand-animation

mattmutt:
helps resolve and patch defect fully from https://github.com/vmware/clarity/issues/2904

+ rapidly toggling row expansion on a delicate, underpowered processor / GPU can raise this condition
+ animation can be a costly process, since layout is occurring by the compositor within the expanded row's content
+ the `run()` routine clears the `oldHeight` property from the instance, with this patch a re-entrant call to `run` will be thoroughly guarded now against the possibility of a missing, hence undefined, `oldHeight`. Animation will not occur since the state is undefined anyways.

Proof of race condition:
expandable datagrid row race condition

Signed-off-by: Matt <blacknugget@excite.com>

### DIFF
--- a/src/clr-angular/data/datagrid/animation-hack/row-expand-animation.spec.ts
+++ b/src/clr-angular/data/datagrid/animation-hack/row-expand-animation.spec.ts
@@ -20,7 +20,7 @@ export default function(): void {
   // Commenting this out because PhantomJS is being uncooperative.
   // I lost too much time trying to get it to pass, but this should just go away anyway once the
   // new cool features of Angular 4.1 animations come in.
-  describe('DatagridRowExpandAnimation directive', function() {
+  xdescribe('DatagridRowExpandAnimation directive', function() {
     beforeEach(function() {
       // We do not use the TestContext on purpose, because we want to test this directive in isolation,
       // without all other components and directives on the same selector.
--- a/src/clr-angular/data/datagrid/animation-hack/row-expand-animation.spec.ts
+++ b/src/clr-angular/data/datagrid/animation-hack/row-expand-animation.spec.ts
@@ -3,87 +3,63 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-
-/*
- * This is a hack that we have to write for now because of bugs and limitations in Angular,
- * please do not use this as an example.
- */
-
-import { Directive, ElementRef, Renderer2 } from '@angular/core';
+import { Component } from '@angular/core';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 
 import { Expand } from '../../../utils/expand/providers/expand';
-import { DomAdapter } from '../../../utils/dom-adapter/dom-adapter';
+import { MOCK_DOM_ADAPTER_PROVIDER } from '../../../utils/dom-adapter/dom-adapter.mock';
+import { DatagridRenderOrganizer } from '../render/render-organizer';
 
-@Directive({ selector: 'clr-dg-row' })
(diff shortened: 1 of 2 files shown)


gnomeontherun:
@mattmutt Is there a reproduction of this issue to review?

mattmutt:
Can provide an internal URL to the client with credentials to observe the same scenario in our app that has a customized expandable row component.

gnomeontherun:
At a minimum, we should have a unit test to cover this case. We should also would have an isolated reproduction since there is no corresponding bug report for this issue. It is hard to review and understand if this fix is appropriate or not without that.

mattmutt:
Not the only one experiencing this. Please patch with either my PR or fix the broken logic in the DomAdapter class:

https://github.com/vmware/clarity/issues/3255

Jinnie:
Note: the animation hack class will probably soon be gone. I am currently working on refactoring of this approach. I will link this issue once I have a PR. Please be patient.

gnomeontherun:
@mattmutt Can you review the errors in Travis CI (formatting issues) and look at creating a unit test to cover this case? I'm still not certain I know exactly what you are seeing, and we can't merge a bugfix for code that doesn't have a clear reproducible scenario or unit test.

src/clr-angular/data/datagrid/animation-hack/row-expand-animation.spec.ts
```diff
@@ -45,6 +45,19 @@ export default function(): void {
       this.expand.expanded = true;
       expect(this.clarityElement.style.height).toBeTruthy();
     });
+    it(
+      'successive triggering of expand and collapse should not throw an error during `animate()`',
+      fakeAsync(() => {
+        expect(() => {
+          this.expand.expanded = false;
+          this.expand.expanded = true;
```
gnomeontherun: This doesn't really change the expanded state and cause toggling, since you're inside of a fakeAsync. I'm still not sure how to reproduce this issue clearly, but this test isn't really doing anything with animations to address your core issue. I'm still trying to get an isolated demo here.

mattmutt: slow down the CPU profile and the animations will fail if a user rapidly clicks on the row expansion icon
Sometimes toggling row expansion on a row that has a radio button, but is in the unselected state can cause the problem ( in a real app, have seen it )

The test I wrote may not even work on the pipeline, because the original Clarity lead developer stated the fact 2 years ago in the test file. It was never corrected.

If you leave my test intact and revert the small bit of code changes in the PR, what happens if you perform said test locally? share any results

gnomeontherun: This test is inside of a non-running test suite, so let's just revert it.

src/clr-angular/data/datagrid/animation-hack/row-expand-animation.ts
```diff
@@ -70,6 +70,10 @@ export class DatagridRowExpandAnimation {
   }
 
   private run() {
+    // defense against race condition when rapid, successive toggling occurs and `oldHeight` property is removed
+    if (!this.hasOwnProperty('oldHeight')) {
```
gnomeontherun: I'm wondering why we don't just move `delete this.oldHeight` inside of the `onfinish` handler instead of this guard? I'm worried about side effects. When I run this on 6x slowdown, the animations are janky and it does seem to resolve, but I'd rather delete the oldHeight in the correct place after the animation is done instead of this extra check, as I see the same result.

gnomeontherun:
I'm going to close this as it should be resolved by #3339 and that is the course of action we'd like to take.

Issue #1240, opened 2017-08-01
[NG] Collapse Animation doesn't work when using clrIfExpanded Directive


adityarb88:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior

When an Element using `clrIfExpanded` is collapsed, it should first animate the collapse and then the children should be removed from the View

### Actual behavior

When an Element using `clrIfExpanded is collapsed, animation doesnt work because the child nodes are removed from the DOM immediately.

### Reproduction of behavior

Tree View:
https://plnkr.co/edit/NWGdymT6uBaSNnW3iJmw?p=preview

### Environment details

* **Angular version:** 4.2.X

* **Clarity version:** 0.10.0

* **OS and version:** All

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Jinnie:
After discussion with the team it was decided not to fix this, as we would have to introduce asynchronous update of the model, bound to an animation lifecycle event. The lost synchronicity and increased testing complexity are not justified with the minor visual animation gain.

PR #3230, opened 2019-03-22
Treeview collapse animation fix (#1240)

Jinnie:
Node content cleared only after the collapsing animation is complete

### DIFF
--- a/src/clr-angular/data/tree-view/models/expanded-state.enum.ts
+++ b/src/clr-angular/data/tree-view/models/expanded-state.enum.ts
@@ -1,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum ClrExpandedState {
+  EXPANDED = 'expanded',
+  COLLAPSED = 'collapsed',
+}
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -1132,6 +1132,7 @@ export declare class ClrTreeNode<T> implements OnInit, OnDestroy {
     expandable: boolean | undefined;
     expanded: boolean;
     expandedChange: EventEmitter<boolean>;
+    expandedState: ClrExpandedState;
     featuresService: TreeFeaturesService<T>;
     nodeId: string;
     readonly rootAriaMultiSelectable: boolean;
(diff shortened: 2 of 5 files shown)


src/clr-angular/data/tree-view/models/expanded-state.enum.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum ClrExpandedState {
+  EXPANDED = 'expanded',
+  COLLAPSED = 'collapsed',
+}
```
coryrylan: Is this needed because `childNodesState` animation is using `expanded` and `'collapsed` as its state names?

Jinnie: That was the intention, but I guess it's not mandatory. It would be nice to have textual states we can use directly in the tests, I think.

Jinnie: @Shijir I don't think we have other option if we want to keep the child nodes visible before the animation is over. 
If the team agrees synchronicity is more important - I'm okay to drop this change entirely.

coryrylan: I double checked my local tests and I am able to write unit tests without worrying about using fakeAsync or tick as long as the test module uses the Noop Animations module. Still going to research more into this since the stepper has to use the same API.

coryrylan:
This PR I am unsure ideally what we should do. I look like it is not possible for `*clrIfExpanded` to work well with Animations unless the service is set once the animation completes. `(@animation.done)="toggleExpanded()"`. By using the animations done API it makes the update async which causes the test to need to use fakeAsync. This will continue to come up if we introduce more animations to the components. What do we value more, Animations or keeping the tests for conditional directive updates synchronous?

We could have a disclaimer in the docs that using the conditional directives may cause animations to not work. Ideally, we eventually could have a docs page about the conditional directives :) 

The stepper has a very similar scenario and I was able to have a workaround keeping the test synchronous by triggering the next step from a mock service and not from triggering a click in the UI bypassing the animation.

Jinnie:
It only happens on the collapsing animation and the animation actually triggers. The problem is that the data inside disappears too soon. Which is a rather minor inconvenience. I will leave this open till the end of this week, to give time for more opinions. Then I will be closing it in favor of simpler model updates and testing scenarios.

PR #3231, opened 2019-03-22
reduce logging for netlify builds

gnomeontherun:
We have a problem with too much logging during the build for Netlify that causes timeouts.


src/website/prerender.ts
```diff
@@ -89,20 +88,18 @@ const deploy = () => {
 
 // Writes rendered HTML to index.html, replacing the file if it already exists.
 const renderer = url => {
-  const spinner = ora('Route: ' + url.route).start();
+  console.log('Route: ' + url.route);
```
coryrylan: Is this to just give feedback while waiting on the pre-render build?

gnomeontherun: This just helps to know what routers were rendered, if they fail the alert is below. Before we used `ora` which does a spinner in the CLI to show completion, but that might have been problematic too.

PR #3240, opened 2019-03-25
[NG] refactor hideable columns

Shijir:
This is internal refactoring changes for hideable columns.


gnomeontherun:
@Shijir is this a WIP or are you ready for reviews?

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -118,4 +124,12 @@ export class DatagridHeaderRenderer implements OnDestroy {
       this.autoSet = true;
     }
   }
+
+  private setVisibility(state: DatagridColumnState) {
```
gnomeontherun: We should standardize the nomenclature around `hidden` or `visibility`. Since the feature is called hideable columns, I think hidden is more clear though technically we are setting the visibility property in the class.

src/dev/src/app/datagrid/hide-show-columns/hide-show.html
```diff
@@ -84,7 +84,7 @@ <h2>Datagrid: Hide-show columns demo</h2>
     </clr-dg-footer>
 </clr-datagrid>
 
-<h2>Hide-show columns demo with custom buttons</h2>
+<!--<h2>Hide-show columns demo with custom buttons</h2>-->
```
gnomeontherun: Oops?

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -9,11 +9,12 @@ import { BehaviorSubject, Subscription } from 'rxjs';
 import { DomAdapter } from '../../../utils/dom-adapter/dom-adapter';
 import { DatagridRenderStep } from '../enums/render-step.enum';
 import { ColumnResizerService } from '../providers/column-resizer.service';
-import { STRICT_WIDTH_CLASS } from './constants';
+import { HIDDEN_COLUMN_CLASS, STRICT_WIDTH_CLASS } from './constants';
 import { DatagridRenderOrganizer } from './render-organizer';
 import { DatagridColumnState } from '../interfaces/column-state.interface';
 import { DatagridColumnChanges } from '../enums/column-changes.enum';
 import { COLUMN_STATE, COLUMN_STATE_PROVIDER } from '../providers/column-state.provider';
+import { skip } from 'rxjs/internal/operators';
```
gnomeontherun: Remove

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -66,4 +78,12 @@ export class DatagridCellRenderer implements OnDestroy {
     }
     this.renderer.setStyle(this.el.nativeElement, 'width', state.width + 'px');
   }
+
+  private setVisibility(state: DatagridColumnState) {
```
gnomeontherun: Same here, let's keep name something like `setHidden`

src/clr-angular/data/datagrid/enums/column-changes.enum.ts
```diff
@@ -6,4 +6,10 @@
 
 export enum DatagridColumnChanges {
   WIDTH,
+  HIDEABILITY,
 }
+
+export const ALL_COLUMN_CHANGES: DatagridColumnChanges[] = [
+  DatagridColumnChanges.WIDTH,
+  DatagridColumnChanges.HIDEABILITY,
```
gnomeontherun: nit `HIDDEN`, this doesn't toggle the column's ability to be hidden or not.

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
```diff
@@ -4,9 +4,17 @@
...
  */
 import { DatagridColumnChanges } from '../enums/column-changes.enum';
+import { ClrDatagridHideableColumn } from '@clr/angular';
+import { TemplateRef } from '@angular/core';
 
 export interface DatagridColumnState {
   changes?: DatagridColumnChanges[]; // This is an array of change types to update
   width?: number; // This is the width calculated for the column
   strictWidth?: number; // This is the strict width if defined in styles/css
+  isHidden?: boolean; // This is the state of column visibility
```
gnomeontherun: Prefer to use `hidden` to track hidden state.

Then I want a separate property to track if a column can be hidden (which right now is defined by the presence of the column template ref).

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
```diff
@@ -4,9 +4,17 @@
...
+  titleTemplateRef?: TemplateRef<ClrDatagridHideableColumn>;
```
gnomeontherun: Add comment here about this property's purpose.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -3,141 +3,109 @@
...
 import { ClrDatagridColumnToggleButton } from './datagrid-column-toggle-button';
 import { ClrDatagridColumnToggleTitle } from './datagrid-column-toggle-title';
-import { DatagridHideableColumnModel } from './datagrid-hideable-column.model';
-import { ColumnToggleButtonsService } from './providers/column-toggle-buttons.service';
-import { HideableColumnService } from './providers/hideable-column.service';
+
 import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
+import { ColumnsService } from './providers/columns.service';
+import { DatagridColumnState } from './interfaces/column-state.interface';
+import { BehaviorSubject } from 'rxjs/index';
```
gnomeontherun: You've got the wrong import paths on a few rxjs imports like this one.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -3,141 +3,109 @@
...
+      <div class="switch-footer" *ngIf="buttons.length > 0">
+        <ng-content select="clr-dg-column-toggle-button"></ng-content>
+      </div>
+      <div class="switch-footer" *ngIf="buttons.length === 0">
+        <div>
+          <button
+            class="btn btn-sm btn-link p6 text-uppercase"
+            [disabled]="allColumnsVisible"
+            (click)="selectAll()"
+            type="button">Select All
```
gnomeontherun: This should use the commonStrings service, and then we wouldn't need the `clr-dg-column-toggle-button` at all.

src/clr-angular/data/datagrid/render/row-renderer.ts
```diff
@@ -3,29 +3,54 @@
...
+    // This method runs in four cases:
+    // 1. When the initial rows appear on the first page.
+    //    In this case, the method will be called in DatagridMainRenderer.
+    // 2. When columns (corresponding header/cells) get added and deleted.
+    //    In this case, the method will be called in DatagridMainRenderer. (Read the note on this case above).
+    // 3. When rows load asynchronously.
+    //    In this case, the method will be called in this class.
+    // 4. When rows load after switching pages.
+    //    In this case, the method will be called in this class (Basically, same as the case 3).
+    if (this.cells.length === this.columnsService.columns.length) {
```
gnomeontherun: Don't we have a potential issue here, where we hide a column and service would have n+1 columns in it?

gnomeontherun: It is right here, this guards that only when the headers have stabilized do we run this.

src/dev/src/app/datagrid/hide-show-columns/hide-show.html
Shijir: Oops, reverted.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
Shijir: As per our discussion, using commonStrings would introduce a breaking change. As this is internal refactoring change, I think it's better to not have any breaking changes for now. But that's being said, we could use the commonStrings service for both toggle title and button in the future.

src/clr-angular/data/datagrid/enums/column-changes.enum.ts
Shijir: done

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
Shijir: done

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
Shijir: done

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
Shijir: done

src/clr-angular/data/datagrid/render/header-renderer.ts
Shijir: done

src/clr-angular/data/datagrid/render/cell-renderer.ts
Shijir: done

src/clr-angular/data/datagrid/interfaces/column-state.interface.ts
```diff
@@ -3,10 +3,18 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import { TemplateRef } from '@angular/core';
 import { DatagridColumnChanges } from '../enums/column-changes.enum';
 
 export interface DatagridColumnState {
```
gnomeontherun: nit: rename this to `ColumnState` since our counterpart is `ColumnStateDiff`

src/clr-angular/data/datagrid/datagrid-column.spec.ts
```diff
@@ -25,6 +23,7 @@ import { StateDebouncer } from './providers/state-debouncer.provider';
 import { TableSizeService } from './providers/table-size.service';
 import { DomAdapter } from '../../utils/dom-adapter/dom-adapter';
 import { DatagridRenderOrganizer } from './render/render-organizer';
+import { ColumnsService } from './providers/columns.service';
 
 const PROVIDERS_NEEDED = [
```
gnomeontherun: Can we just reuse the same providers list for this test too?

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -40,12 +45,19 @@ export class DatagridCellRenderer implements OnDestroy {
   }
 
   private stateChanges(state: DatagridColumnState) {
+    if (this.runAllChanges) {
```
gnomeontherun: No test case for running all changes, can we add something?

src/clr-angular/data/datagrid/providers/columns.service.mock.ts
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { ColumnsService } from './columns.service';
+import { EmbeddedViewRef, Injectable, TemplateRef } from '@angular/core';
+import { columnStateFactory } from './column-state.provider';
+import { DatagridColumnChanges } from '../enums/column-changes.enum';
+import { ClrDatagridHideableColumn } from '@clr/angular';
```
gnomeontherun: import

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -3,43 +3,32 @@
...
 
 @Injectable()
-export class ColumnsService implements OnDestroy {
-  subscriptions: Subscription[] = [];
+export class ColumnsService {
   columns: BehaviorSubject<DatagridColumnState>[] = [];
 
-  constructor(private organizer: DatagridRenderOrganizer) {
-    this.subscriptions.push(
-      this.organizer.filterRenderSteps(DatagridRenderStep.CLEAR_WIDTHS).subscribe(() => this.reset())
```
gnomeontherun: I don't remember why this is being removed.

Shijir: This callback of `DatagridRenderStep.CLEAR_WIDTHS` get called with a change of `{width: 0}`. But instantly after that, the state is again overridden from the callback of `DatagridRenderStep.CLEAR_WIDTHS` in `HeaderRenderer`. So this was totally unnecessary. Even though the value gets overridden instantly, it was still calling the dom renderer.

src/clr-angular/data/datagrid/providers/columns.service.mock.ts
Shijir: Done

src/clr-angular/data/datagrid/datagrid-column.spec.ts
Shijir: Done

src/clr-angular/data/datagrid/render/cell-renderer.ts
Shijir: Done

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -76,15 +78,24 @@ export default function(): void {
...
       domAdapter = <MockDomAdapter>context.getClarityProvider(DomAdapter);
       organizer = <MockDatagridRenderOrganizer>context.getClarityProvider(DatagridRenderOrganizer);
-      stateSub = context.clarityDirective.columnState;
+      columnsService = context.getClarityProvider(ColumnsService);
+      context.clarityDirective.setColumnState(0);
+      stateSub = columnsService.columns[0];
+    });
+
+    it('sets its state in columns service at given index', function() {
+      expect(columnsService.columns.length).toBe(1);
```
gnomeontherun: This test isn't checking that the state is actually set properly, just that there is a value.

src/clr-angular/data/datagrid/render/constants.ts
```diff
@@ -9,3 +9,5 @@
 export const NO_LAYOUT_CLASS = 'datagrid-no-layout';
 export const COMPUTE_WIDTH_CLASS = 'datagrid-computing-columns-width';
 export const STRICT_WIDTH_CLASS = 'datagrid-fixed-width';
+export const HIDDEN_COLUMN_CLASS = 'datagrid-column-hidden';
```
gnomeontherun: I don't see any reason why we can't just have 1 hidden class for both columns and cells, that will make it easier later to dedupe some of the header/cell renderer logic.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: I recall talking to Eudes about this, and this was duplicating the behavior we had before. There was a reason related to ensuring the calculations are correct. @hippee-lee can you verify if this can be removed from the original refactor?

src/clr-angular/data/datagrid/enums/column-changes.enum.ts
```diff
@@ -6,4 +6,7 @@
 
 export enum DatagridColumnChanges {
   WIDTH,
+  HIDDEN,
 }
+
+export const ALL_COLUMN_CHANGES: DatagridColumnChanges[] = [DatagridColumnChanges.WIDTH, DatagridColumnChanges.HIDDEN];
```
gnomeontherun: I think we could iterate over the keys of the enum to build ALL_COLUMN_CHANGES, untested but something like

```
export const ALL_COLUMN_CHANGES: DatagridColumnChanges[] = Object.keys(DatagridColumnChanges).map(key => DatagridColumnChanges[key]);```

File: src/clr-angular/data/datagrid/datagrid-footer.spec.ts [hunk 19]
```diff
@@ -71,22 +50,22 @@ export default function(): void {
         clarityDirectiveSelection.selectionType = SelectionType.Multi;
         clarityDirectiveSelection.current.push(1);
 
-        context.clarityDirective.cdr.markForCheck();
+        //context.clarityDirective.cdr.markForCheck();
```
gnomeontherun: Can we remove these?

src/clr-angular/data/datagrid/providers/columns.service.ts
hippee-lee: I believe that duplication was there because cells and columns used to be handled independently. IIR cells always removed the width and strict width css class that overrides the flex behavior because the rendering cycle would determine that from the column. I think it had to be this way for the use case when all columns had been resized except for the last one and we had to remove the fixed-width class so it would flex. 

Now that cells and columns are managed from a single source of truth for width state it looks like this can be removed. I manually tested with the dev app and it behaved as I expected it to.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -30,11 +30,6 @@ export class DatagridHeaderRenderer implements OnDestroy {
     this.subscriptions.push(
       this.organizer.filterRenderSteps(DatagridRenderStep.CLEAR_WIDTHS).subscribe(() => this.clearWidth())
     );
-    this.subscriptions.push(
-      this.organizer
-        .filterRenderSteps(DatagridRenderStep.DETECT_STRICT_WIDTHS)
-        .subscribe(() => this.detectStrictWidth())
```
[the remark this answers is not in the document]
Shijir: Oh we actually call it in the `getColumnWidthState()` method.

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
```diff
@@ -130,12 +130,20 @@ export default function(): void {
...
 
       stateSub.next({ changes: [DatagridColumnChanges.WIDTH], width: 123, strictWidth: 0 });
       expect(context.clarityElement.style.width).toBe('123px');
-      expect(context.clarityElement.classList).not.toContain('datagrid-fixed-width');
+      expect(context.clarityElement.classList).not.toContain(STRICT_WIDTH_CLASS);
+    });
+
+    it('sets proper hidden class for hidden cell', function() {
+      stateSub.next({ changes: [DatagridColumnChanges.HIDDEN], hidden: true });
+      console.log(context.clarityElement);
```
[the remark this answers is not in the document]
Shijir: Done

src/clr-angular/data/datagrid/datagrid-footer.spec.ts
Shijir: Yes, forgot to remove these ones. Done now, thanks 👌

src/clr-angular/data/datagrid/enums/column-changes.enum.ts
Shijir: Ah yes, but we still need to filter out the string number properties from the object version. So this one will work for us:

```
export const ALL_COLUMN_CHANGES: DatagridColumnChanges[] = Object.keys(DatagridColumnChanges)
  .map(key => DatagridColumnChanges[key])
  .filter(key => key && isNaN(key));
```

src/clr-angular/data/datagrid/render/header-renderer.spec.ts
Shijir: But every property will be undefined here as this is just testing whether the header is sending its initial pristine state to ColumnService.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: Thanks resolving.

PR #3241, opened 2019-03-25
[UI] fix node height inconsistency

### LINKED ISSUE #3142: Weird TreeView items design
## Summary

_Currently, there are few thing in treeview, which are weird:_

_1. Expandable nodes can't be "active"_
_2. Regular nodes that aren't supposed to be selected has height of 30px, but selectable ones - 34px; text in selectable nodes is shifted to the right._

image

* **What is the change?** Make expandable nodes selectable for all tree view types; make selectable and non-selectable nodes have identical height and text position.
* **Why should it go in Clarity?** Because current implementation is pretty limited by things described above.
* **Does this change impact existing behaviors? If so how?** it adds ability to make node both expandable and active. Other changes are cosmetic.

## Use case

_File Browser. Folders can be expandable and selectable simultaneously. Files and folders can exist on a same level of hierarchy._

## Workarounds

For selectable expandable nodes - not sure, for cosmetic changes - css overrides.

coryrylan:
Fixes node height inconsistency as described in #3142 
Node elements should be 32px in height as described in spec #239

closes #3142 


src/clr-angular/data/tree-view/_tree-view.clarity.scss
```diff
@@ -117,7 +117,7 @@
     //Others
     background: transparent;
     border-radius: $clr-tree-border-radius 0 0 $clr-tree-border-radius;
```
gnomeontherun: If we don't have a background or border, we don't need a radius...

Issue #3248, opened 2019-03-26
Removal of 2.0 deprecations


gnomeontherun:
We need to remove the following:

* [x] Deprecated grid file, and any lingering uses of old grid
* [x] Audit component templates for old grid classes and make sure we have scss parity (matt)
* [x] Deprecated old form module (Cory)
* [x] Old form styles (Cory)
* [x] Remove examples of `clr-checkbox` in button groups (Cory)
* [x] Remove deprecated bootstrap sass (Cory)
* [x] Remove old datepicker layout option (Matt)
* [x] Unnecessary gemini tests (Matt)

coryrylan:
Have majority of the deprecated forms module and styles removed. Everything builds but I need to go back and check all three projects and gemini for any possible issues.

coryrylan:
Is the older `.toggle-switch` part of the old forms deprecation? I see that the new component was added a couple of months ago.

PR #3249, opened 2019-03-26
deprecation(grid) remove old grid fully from the project

gnomeontherun:
* Removed the deprecated grid CSS files and build chain
* Removed deprecated grid gemini tests
* Updated all dev/ks-app/website pages that used old grid still
* Removed 'old lists in cards', 'old progress bar cards' demos, no longer valid
* Added precommit hook for license headers

Part of #3248


src/dev/src/app/iconography/icon-selection.demo.html
```diff
@@ -8,7 +8,7 @@
 
     <section>
         <h5>Preview Settings</h5>
-        <div class="row clr-icon-preview-settings">
+        <div class="clr-icon-preview-settings">
```
hippee-lee: Does it still need to be `clr-row` here?

src/ks-app/src/app/containers/iconography/icon-selection.component.html
```diff
@@ -9,7 +9,7 @@ <h1>Iconography</h1>
 <div class="clr-icon-selection">
   <section>
     <h5>Preview Settings</h5>
-    <div class="row clr-icon-preview-settings">
+    <div class="clr-icon-preview-settings">
```
hippee-lee: Same question here about the removed row. Intentional?

src/dev/src/app/iconography/icon-selection.demo.html
gnomeontherun: For both of these questions, there were no column cells here and the row class is not necessary.

PR #3252, opened 2019-03-27
[NG] Focus management utilities

youdz:
Given the complexity and particular constraints of each of our components, writing a "just slap this in" focus management directive turned out to be impractical. Instead, I had to expose smaller pieces that you combine together with a bit more integration on the component side, but hopefully reducing code duplication enough to be valuable.

Here is the architecture of these new focus management utilities:
- Based on Chris' preference, we will be using `aria-activedescendant` rather than actual focus management.
- All focusable objects we are going to work with should implement the `FocusableItem` interface.
- These focusable objects have optional pointers in all directions to indicate what the next object in that direction is.
- A singleton FocusService per focus-managed component will work with this "graph" of focusable items and navigate it according to the keys pressed by the user.
- In order to help build the graph of focusable items, reusable linkers are provided. I only implemented 2 so far, but added a comment noting that we can easily add more in the future.
- A typical focusable item for our components will simply toggle the `.focus` CSS class on itself when focused. Once again, this is because we use `aria-activedescendant` so the item isn't actually focused, it just needs to look focused.

**************

The second commit uses the common focus management utilities to enable keyboard navigation of dropdowns.

The biggest hurdle here is that inside of a dropdown menu, the focusable items can be either a dropdown item or the trigger of a nested dropdown. But because we can't query deeply as that would break multiple levels of nesting, our `@ContentChildren` query only has access to the top-level children. This means we have to work with the nested dropdowns themselves, not their trigger. To get this to work, I introduce a custom FocusableItem implementation for each dropdown which "proxies" focus behavior to its trigger, and bridges keyboard navigation by pointing to its children.

Note that this relies on the fact that `@ContentChildren` doesn't have to query for components or directives, it can query directly for providers. So any component or directive declaring said provider will be part of the resulting `QueryList`.

I wish I could have spent a bit more time trying to extract more reusable pieces from the custom DropdownFocusHandler implementation, since the Combobox will share most of it expect the nesting management. But it's a bit too entangled right now and I am under a pretty hard time constraint, so I'm sorry to let @jeeyun handle the extraction for when she'll implement this for combobox.

This implementation uses a pattern you might not have faced yet with Observables: observables that execute an action when any observer subscribes to them (in this case opening/closing the dropdown). This lets us avoid adding a bunch of observables in the FocusService to try and subscribe to them from every focusable item, which would end up in nice spaghetti code. Instead, when the focus service tries to go down, we can simply intercept that with our on-subscribe pattern, open the dropdown, and then go down.

There were a lot of edge cases to test, I hope I my specs are comprehensive enough.

Here is a live demo of this implementation: http://clarity-accessible-dropdown.surge.sh/

### DIFF
--- a/src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
+++ b/src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
@@ -43,7 +43,7 @@
   text-transform: none;
 
   &:hover,
-  &.focus {
+  &.clr-focus {
     background-color: $clr-dropdown-bg-hover-color;
     color: $clr-dropdown-item-text-color;
     text-decoration: none;
@@ -73,7 +73,7 @@
     }
 
     &:active,
-    &.focus {
+    &.clr-focus {
       background: none;
       box-shadow: none;
     }
--- a/src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
+++ b/src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
@@ -43,7 +43,7 @@
   text-transform: none;
 
   &:hover,
-  &:focus {
+  &.focus {
     background-color: $clr-dropdown-bg-hover-color;
     color: $clr-dropdown-item-text-color;
     text-decoration: none;
@@ -73,7 +73,7 @@
     }
 
     &:active,
-    &:focus {
+    &.focus {
       background: none;
       box-shadow: none;
     }
--- a/src/clr-angular/utils/focus/direction.enum.ts
+++ b/src/clr-angular/utils/focus/direction.enum.ts
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum Direction {
+  UP = 'up',
+  DOWN = 'down',
+  LEFT = 'left',
+  RIGHT = 'right',
+}
--- a/src/clr-angular/utils/focus/focusable-item/linkers.ts
+++ b/src/clr-angular/utils/focus/focusable-item/linkers.ts
@@ -5,7 +5,7 @@
  */
 
 import { Observable } from 'rxjs';
-import { Direction } from '../direction.enum';
+import { ArrowKeyDirection } from '../arrow-key-direction.enum';
 import { FocusableItem } from './focusable-item';
 
 /**
@@ -14,7 +14,7 @@ import { FocusableItem } from './focusable-item';
 export function linkParent(
   items: FocusableItem[],
   parent: FocusableItem | Observable<FocusableItem>,
-  direction: Direction
+  direction: ArrowKeyDirection
 ) {
   items.forEach(item => (item[direction] = parent));
 }
--- a/src/clr-angular/utils/focus/focusable-item/linkers.ts
+++ b/src/clr-angular/utils/focus/focusable-item/linkers.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Observable } from 'rxjs';
+import { Direction } from '../direction.enum';
+import { FocusableItem } from './focusable-item';
+
+/**
+ * Links a set of focusable items to a parent along one direction
+ */
+export function linkParent(
+  items: FocusableItem[],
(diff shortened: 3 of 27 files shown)


youdz:
Yes, my bad on `wrapObservable()`, it deserved to be part of the next PR.

The idea isn't one I'm completely a fan of, but it's the best I could come up with that addressed all the weird cases our components with `*clrIfOpen` or other structural directives faced.
What it does is it lets you perform an operation when someone **subscribes** to your Observable. So if someone is asking for the next item to focus to, you can first open/expand/etc and then you emit the first child once you have it.

src/clr-angular/utils/focus/focusable-item/linkers.ts
```diff
@@ -0,0 +1,42 @@
...
+export function linkVertical(items: FocusableItem[], loop = true) {
+  items.forEach((item, index) => {
+    if (index > 0) {
+      item.up = items[index - 1];
+    }
+    if (index < items.length - 1) {
+      item.down = items[index + 1];
+    }
+  });
+  if (loop && items.length) {
```
Shijir: Shouldn't this guard another condition where `items.length > 1`? Otherwise, it would link to itself?

youdz: `if (items.length)` and `if (items.length > 1)` do the same thing, 0 is the only falsy number in Javascript.

src/clr-angular/utils/focus/direction.enum.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum Direction {
```
gnomeontherun: I would like a better name for this.

youdz:
The newly pushed commit uses the common focus management utilities to enable keyboard navigation of dropdowns.

The biggest hurdle here is that inside of a dropdown menu, the focusable items can be either a dropdown item or the trigger of a nested dropdown. But because we can't query deeply as that would break multiple levels of nesting, our `@ContentChildren` query only has access to the top-level children. This means we have to work with the nested dropdowns themselves, not their trigger. To get this to work, I introduce a custom FocusableItem implementation for each dropdown which "proxies" focus behavior to its trigger, and bridges keyboard navigation by pointing to its children.

Note that this relies on the fact that `@ContentChildren` doesn't have to query for components or directives, it can query directly for providers. So any component or directive declaring said provider will be part of the resulting `QueryList`.

I wish I could have spent a bit more time trying to extract more reusable pieces from the custom DropdownFocusHandler implementation, since the Combobox will share most of it expect the nesting management. But it's a bit too entangled right now and I am under a pretty hard time constraint, so I'm sorry to let @jeeyun handle the extraction for when she'll implement this for combobox.

This implementation uses a pattern you might not have faced yet with Observables: observables that execute an action when any observer subscribes to them (in this case opening/closing the dropdown). This lets us avoid adding a bunch of observables in the FocusService to try and subscribe to them from every focusable item, which would end up in nice spaghetti code. Instead, when the focus service tries to go down, we can simply intercept that with our on-subscribe pattern, open the dropdown, and then go down.

There were a lot of edge cases to test, I hope I my specs are comprehensive enough.

Here is a live demo of this implementation: http://clarity-accessible-dropdown.surge.sh/

I'll add this description to the opening post.

src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
```diff
@@ -43,7 +43,7 @@
   text-transform: none;
 
   &:hover,
-  &:focus {
+  &.focus {
```
jeeyun: Do you know if we are supposed to `.clr-` prefix the `focus` class? I started out without then added it in combobox.

src/clr-angular/utils/focus/direction.enum.ts
jeeyun: NavigationDirection? KeyboardNavDirection? OneDirection?

Shijir: KeyboardNavDirection 👍 
OneDirection 😂

youdz: I'm going with `ArrowKeyDirection`.

But `OneDirection` was a close second. 👍

src/clr-angular/popover/dropdown/_menu-mixins.clarity.scss
youdz: I think you're right, I'll update that to `.clr-focus`.

hippee-lee:
Closing this PR as @jeeyun has taken it over and added changes here in #3283 that are needed to get it merged into master.

Issue #1458, opened 2017-09-12
Disable checkboxes in tree view


demus:
**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x ] enhancement
```

### Expected behavior

Add the ability to disable checkboxes in tree view. Currently as far as I can tell they cannot be disabled.

vivek203:
Do we have an update on this issue?

tarquinrex:
This is really needed, please can you take a look. Thanks.

typhoon1978:
Hi all, any update about this? I think it's something really needed !!!

mathisscott:
@youdz ^

youdz:
Not part of the refactor at the moment, but I believe it should be fairly straightforward to implement right after it. Since our API is now entirely based on per-node bindings, adding a `[clrSelectable]` input to nodes is quite easy, unless I missed some side-effect.

This could be good for @Jinnie to take soon after the new tree is merged in (any day now). 👍

ps37:
@youdz I will be happy to make the contribution by adding a new input to the tree node.

youdz:
That's perfect, then I suggest you check out our [coding guidelines](https://github.com/vmware/clarity/blob/master/CODING_GUIDELINES.md).

For a well-defined feature like this that shouldn't take many lines of code, you don't have to use a topic branch. Just write the feature as a single commit (including unit tests), and submit a PR against `master`. 👍

ps37:
I have a doubt regarding the use case:
-Should disabling a parent node disable all of its child nodes?

-if so, if we disable a parent node and dynamically enable it again, should all the child nodes be enabled again without honoring the disabled input set on a child node?

For example, if `dynamicBoolean` is set to `false` initially and then later set to `true`, should the child node `A-2` node get enabled or be disabled?
```
<clr-tree-node [clrSelectable]="dynamicBoolean">
    A
    <clr-tree-node>
      A-1
    </clr-tree-node>
    <clr-tree-node [clrSelectable]="false">
      A-2
      <clr-tree-node>
        A-2.1
      </clr-tree-node>
    </clr-tree-node>
  </clr-tree-node>
```

vivek203:
If a parent is disabled, all its children should be disabled too.
If all the children of a parent are disabled, the parent should be disabled as well.

In the case of dynamically changing parent/children, the local condition should be given higher priority than the parent policy.

[...]

PR #3261, opened 2019-03-29
[NG] Disable checkboxes in tree view

ps37:
Add the ability to dynamically disable selection of tree node checkboxes.

closes the issue: https://github.com/vmware/clarity/issues/1458

Signed-off-by: Prudhvi Simhadri <prudhvi.af121@gmail.com>

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -389,6 +389,7 @@ export declare class ClrDatagridDetailHeader {
 export declare class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>> implements CustomFilter, OnDestroy {
     readonly active: boolean;
     anchor: ElementRef;
+    ariaExpanded: boolean;
     commonStrings: ClrCommonStringsService;
     customFilter: ClrDatagridFilterInterface<T> | RegisteredFilter<T, ClrDatagridFilterInterface<T>>;
     open: boolean;
--- a/src/clr-angular/data/datagrid/datagrid-filter.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-filter.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -20,6 +20,12 @@ import { ClrPopoverEventsService } from '../../utils/popover/providers/popover-e
 
 class MockRenderer {
   listen() {}
+}
+
+function cleanPopoverDOM(component) {
+  const popoverContent = document.querySelectorAll('.clr-popover-content');
+  popoverContent.forEach(content => document.body.removeChild(content));
+  component.ngOnDestroy();
 }
 
 export default function(): void {
@@ -45,9 +51,7 @@ export default function(): void {
       });
 
       afterEach(function() {
-        const popoverContent = document.querySelectorAll('.clr-popover-content');
-        popoverContent.forEach(content => document.body.removeChild(content));
-        component.ngOnDestroy();
+        cleanPopoverDOM(component);
       });
 
       it('registers to the FiltersProvider provider', function() {
@@ -132,6 +136,34 @@ export default function(): void {
           },
         ]);
         context.testComponent.filter = filter;
+      });
+
+      afterEach(function() {
+        cleanPopoverDOM(context.clarityDirective);
+      });
+
+      it('correctly associates the popover content with the aria-controls value', function() {
+        const toggle: HTMLButtonElement = context.clarityElement.querySelector('.datagrid-filter-toggle');
+        toggle.click();
+        context.detectChanges();
+        const popover: HTMLDivElement = document.querySelector('.datagrid-filter');
+        expect(toggle.getAttribute('aria-controls')).toEqual(popover.getAttribute('id'));
+      });
+
+      it('correctly updates the aria-expanded state', function() {
+        const toggle: HTMLButtonElement = context.clarityElement.querySelector('.datagrid-filter-toggle');
+        expect(toggle.getAttribute('aria-expanded')).toBe('false');
+        toggle.click();
+        context.detectChanges();
+        expect(toggle.getAttribute('aria-expanded')).toBe('true');
+      });
+
+      it('has a button with the correct common string for datagridFilterAriaLabel', function() {
+        const toggle: HTMLButtonElement = context.clarityElement.querySelector('.datagrid-filter-toggle');
+        const commonStrings: ClrCommonStringsService = context.fixture.debugElement.injector.get(
+          ClrCommonStringsService
+        );
+        expect(toggle.getAttribute('aria-label')).toBe(commonStrings.keys.datagridFilterAriaLabel);
       });
 
       it('projects content into the dropdown', function() {
--- a/src/clr-angular/data/datagrid/datagrid-filter.ts
+++ b/src/clr-angular/data/datagrid/datagrid-filter.ts
@@ -42,6 +42,9 @@ import { isPlatformBrowser } from '@angular/common';
       <button class="datagrid-filter-toggle"
               type="button"
               #anchor
+              [attr.aria-label]="commonStrings.keys.datagridFilterAriaLabel"
+              [attr.aria-expanded]="ariaExpanded"
+              [attr.aria-controls]="popoverId"
               clrPopoverAnchor
               clrPopoverOpenCloseButton
               [class.datagrid-filter-open]="open"
@@ -66,6 +69,8 @@ import { isPlatformBrowser } from '@angular/common';
 export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>>
   implements CustomFilter, OnDestroy {
   private subs: Subscription[] = [];
+  public ariaExpanded: boolean = false;
+
   constructor(
     _filters: FiltersProvider<T>,
     public commonStrings: ClrCommonStringsService,
@@ -77,6 +82,7 @@ export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDa
     this.subs.push(
       smartToggleService.openChange.subscribe(change => {
         this.open = change;
+        this.ariaExpanded = change;
       })
     );
   }
--- a/src/clr-core/common/services/common-strings.default.ts
+++ b/src/clr-core/common/services/common-strings.default.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -48,6 +48,7 @@ export const commonStringsDefault: ClrCommonStrings = {
   singleSelectionAriaLabel: 'Single selection header',
   singleActionableAriaLabel: 'Single actionable header',
   detailExpandableAriaLabel: 'Toggle more row content',
+  datagridFilterAriaLabel: 'Toggle column filter',
   // Alert
   alertCloseButtonAriaLabel: 'Close alert',
   // Date Picker
--- a/src/clr-core/common/services/common-strings.interface.ts
+++ b/src/clr-core/common/services/common-strings.interface.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -116,6 +116,11 @@ export interface ClrCommonStrings {
    * Datagrid numeric filter: max
    */
   maxValue?: string;
+  /**
+   * Datagrid filter toggle button
+   */
+  datagridFilterAriaLabel?: string;
+
   /**
    * Modal start of content
    */


ps37:
@youdz Can you please take a look at this review?

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -55,6 +83,23 @@ export abstract class TreeNodeModel<T> {
     this.setSelected(newState, true, propagate);
   }
 
+  // To bypass the default event handlers and programmatically set the checked' & 'indeterminate' DOM properties of input checkbox
+  checkboxToggleHandler(propagate: boolean, event: Event) {
+    this.toggleSelection(propagate);
+    // When a user triggers the click event on checkbox inputs, checkbox has toggled its state from checked to unchecked or vice versa.
+    // And that does not get over written by the conditional operations being bound to 'checked' & 'indeterminate' DOM
```
youdz: Technically, what happens is that changing the bound property from `false` to `false` doesn't actually change it so the `this.checked = false` assignment does literally nothing, and Angular simply does not update the binding since nothing apparently changed.

You can see it very clearly in this example: https://stackblitz.com/edit/prevent-checkbox-change?file=src%2Fapp%2Fapp.component.html
I think what you have is fine, but it has nothing to do with the link you posted so I'd rather you remove it.

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -55,6 +83,23 @@ export abstract class TreeNodeModel<T> {
...
+    // properties in the HTML template. So, they are programatically being set here.
+    // I scratched my head to find this. Thanks to https://www.wearemindscape.com/blog/2012/06/12/preventing-html-checkbox-toggle-with-javascript/
+    const inputElement = event.srcElement as HTMLInputElement;
+    inputElement.checked = false;
+    inputElement.indeterminate = false;
+    if (this.selected.value === ClrSelectedState.SELECTED) {
+      inputElement.checked = true;
+    } else if (this.selected.value === ClrSelectedState.INDETERMINATE) {
+      inputElement.indeterminate = true;
+    }
```
youdz: Nitpick: these 7 lines can be simply written as
```typescript
inputElement.checked = this.selected.value === ClrSelectedState.SELECTED;
inputElement.indeterminate = this.selected.value === ClrSelectedState.INDETERMINATE;
```

src/dev/src/app/tree-view/eager-declarative-tree/eager-declarative-tree.ts
```diff
@@ -13,4 +13,25 @@ import { Component } from '@angular/core';
...
+  isNodeA1disabled: boolean = false;
+  isNodeA2disabled: boolean = false;
+  isNodeA3disabled: boolean = false;
+
+  isNodeA23disabled: boolean = false;
+
+  toggleNode(name) {
+    if (name === 'A-1') {
+      this.isNodeA1disabled = !this.isNodeA1disabled;
+    } else if (name === 'A-2') {
```
youdz: This demo is a bit too messy for my taste. We try to have our demos somewhat similar to what we expect our consumers to be doing in real-life application, and a long series of chained `else if` with hardcoded values, and corresponding manual properties with hardcoded names is just not realistic for actual applications.

I know it doesn't change much, but just making this demo a tiny bit more dynamic would go a long way for future maintenance and debugging. Thank you. 🙂

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -30,22 +33,47 @@ export abstract class TreeNodeModel<T> {
...
 
+  setDisabled(value: boolean, propagateUp: boolean, propagateDown: boolean) {
+    if (this.disabled.value === value) {
+      return;
+    }
+    // If selection of the parent is disabled, then this node also has to be disabled
+    const disabledValue = this.parent && this.parent.disabled.value ? true : value;
+    this.disabled.next(disabledValue);
+    // Don't change the disableSelection state of the children for which the clrDisableSelection is set to true
+    if (propagateDown) {
```
youdz: I missed that one earlier, you want `if (propagateDown && this.children)`. Otherwise you get the issue you had in unit tests which is an `undefined` exception when trying to iterate over the children.

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -144,6 +160,7 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
       this._model.selected.pipe(filter(() => !this.skipEmitChange)).subscribe(value => this.selectedChange.emit(value))
     );
     this.subscriptions.push(this.expandService.expandChange.subscribe(value => this.expandedChange.emit(value)));
+    this.subscriptions.push(this._model.disabled.subscribe(value => this.disabledChange.emit(value)));
```
youdz: @gnomeontherun: Be careful with this, it has the same behavior as the selection that caused problems and that you very recently fixed. It's the async event emitter + the output that fires when you receive an input. I will let you handle it, I think your fix for selection should work fine here too.

src/clr-angular/data/tree-view/models/tree-node.model.spec.ts
```diff
@@ -16,7 +16,7 @@ class TestModel extends TreeNodeModel<string> {
     this.parent = parent;
   }
 
-  children: TestModel[];
+  children: TestModel[] = [];
```
youdz: Please remove this once you add the `if (this.children)` I mentioned in a previous comment. Our tree nodes are supposed to work fine with a `null` or `undefined` list of children.

src/clr-angular/data/tree-view/models/tree-node.model.spec.ts
```diff
@@ -132,5 +132,120 @@ export default function(): void {
...
+          expect(root.children[1].selected.value).toBe(ClrSelectedState.UNSELECTED);
+          root.toggleSelection(true);
+          expect(root.children[0].selected.value).toBe(ClrSelectedState.SELECTED);
+          expect(root.children[1].selected.value).toBe(ClrSelectedState.UNSELECTED);
+        });
+
+        describe('Indeterminate parent', function() {
+          beforeEach(function() {
+            // Unselected and disabeld child
+            expect(root.children[1].selected.value).toBe(ClrSelectedState.UNSELECTED);
```
youdz: I don't understand why you're using an `expect` in a `beforeEach`. You should only be doing setup work here.

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -30,22 +33,47 @@ export abstract class TreeNodeModel<T> {
...
       this.parent._updateSelectionFromChildren();
     }
   }
 
+  setDisabled(value: boolean, propagateUp: boolean, propagateDown: boolean) {
+    if (this.disabled.value === value) {
+      return;
+    }
+    // If selection of the parent is disabled, then this node also has to be disabled
+    const disabledValue = this.parent && this.parent.disabled.value ? true : value;
```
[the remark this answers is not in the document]
ps37: hmm... We do have the ability to disable selection on each individual tree node right. So, for a use case similar to this, "a node has 100 children but you only get to pick 3 at a time at most", they would have to disable the 97 children and leave the parent and the other 3 children enabled.
Also with this, they can do a group selection/de-selection on all the enabled nodes however sparsely they might get distributed among the children.

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -99,4 +144,21 @@ export abstract class TreeNodeModel<T> {
       this.parent._updateSelectionFromChildren();
     }
   }
+
+  private get areAllChildrenDisabled(): boolean {
+    return !this.children.find(child => !child.disabled.value);
```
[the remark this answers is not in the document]
ps37: Thank you! That code has more clarity! ;)

ps37:
@youdz I addressed your review comments. Can you please take a look?

gnomeontherun:
I'll be helping with the review since @youdz has left the team. Can you check on the issues with the build? I see some linting errors and a gemini test that is failing.

src/clr-angular/data/tree-view/models/tree-node.model.spec.ts
```diff
@@ -132,5 +132,119 @@ export default function(): void {
...
+        expect(child.disabled.value).toBe(true);
+        // Once the parent is disabled, a child cannot be enabled
+        child.children[0].setDisabled(false, false, false);
+        expect(child.children[1].disabled.value).toBe(true);
+      });
+
+      it('can propagate the disabled state to children', function() {
+        // Disabling a parent disables all the children down the tree
+        root.setDisabled(true, false, true);
+        [...root.children, ...child.children].forEach(n => expect(n.disabled.value).toBe(true));
```
gnomeontherun: I think it would be more efficient to add up children that are true and compare that against the total length of items, so we only run one `expect` check instead of N `expect`s.

src/clr-angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -55,6 +83,18 @@ export abstract class TreeNodeModel<T> {
     this.setSelected(newState, true, propagate);
   }
 
+  // To bypass the default event handlers and programmatically set the checked' & 'indeterminate' DOM properties of input checkbox
+  checkboxToggleHandler(propagate: boolean, event: Event) {
```
gnomeontherun: I don't see a need for this new method instead of just adding this logic into the existing `toggleSelection` method, can you merge them so we don't have two public methods here?

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -186,6 +209,20 @@ export default function(): void {
         this.clarityDirective.expanded = false;
         expect(this.hostComponent.expanded).toBeFalse();
       });
+
```
gnomeontherun: We need another test, similar to the test above titled `offers a [(clrSelected)] two-way binding, but skips emitting output when setting input` but testing the `clrDisabled` binding skips the emit.

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -186,6 +209,20 @@ export default function(): void {
...
+      it(
+        'offers a [(clrDisabled)] two-way binding',
+        fakeAsync(function(this: Context) {
```
gnomeontherun: This is sync, don't need to wrap in `fakeAsync`. See the EventEmitter which is set to sync.

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -186,6 +209,20 @@ export default function(): void {
...
+          this.hostComponent.disabled = true;
+          this.detectChanges();
+          expect(this.clarityDirective.disabled).toBe(true);
+          this.clarityDirective.disabled = false;
+          // We need fakeAsync and tick because the EventEmitter is asynchronous
```
gnomeontherun: nit: remove this line and 222

src/clr-angular/data/tree-view/tree-node.spec.ts
```diff
@@ -186,6 +209,20 @@ export default function(): void {
...
+          tick();
+          // I don't know why Typescript forces me to cast this
+          expect<boolean>(this.hostComponent.disabled).toBe(false);
```
gnomeontherun: Do we really need to cast here? I think a recent TS update fixed this inference issue `expect(this.hostComponent.disabled).toBe(false);`

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -77,6 +77,15 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
     return !!this.expandService.expandable || this._model.children.length > 0;
   }
 
+  @Input('clrDisabled')
+  get disabled(): boolean {
+    return this._model.disabled.value;
+  }
+  set disabled(value: boolean) {
+    this._model.isDisabledInputSetToTrue = !!value;
+    this._model.setDisabled(!!value, this.featuresService.eager, this.featuresService.eager);
```
gnomeontherun: To make this work properly and avoid a repeat of #3073, we should set a flag prior to running this line and reset it right afterwards. See commit d5ea70da1c0c7aa563ca260195f07b6b199f4efe, and update to the following.

```
set disabled(value: boolean) {
  this._model.isDisabledInputSetToTrue = !!value;
  this.skipEmitChange = true;
  this._model.setDisabled(!!value, this.featuresService.eager, this.featuresService.eager);
  this.skipEmitChange = false;
}
```

src/clr-angular/data/tree-view/tree-node.ts
gnomeontherun: Indeed, once we have the flag setup per my comment above, this line should be

```
    this.subscriptions.push(
      this._model.disabled.pipe(filter(() => !this.skipEmitChange)).subscribe(value => this.disabledChange.emit(value))
    );
```

src/dev/src/app/tree-view/eager-declarative-tree/eager-declarative-tree.html
```diff
@@ -54,6 +54,44 @@ <h4>Selection</h4>
   </clr-tree-node>
 </div>
 
+<h4>Selection with Disabled nodes</h4>
```
gnomeontherun: I would like to have a demo of this for each of the following:

* Eager declarative tree (this file)
* Eager recursive tree (missing)
* Lazy declarative tree (missing)
* Lazy recursive tree (missing)

src/dev/src/app/tree-view/eager-declarative-tree/eager-declarative-tree.html
```diff
@@ -54,6 +54,44 @@ <h4>Selection</h4>
...
+<div class="clr-example example-disabled-nodes">
+  <span>Click on the name of a node to select for toggling it's disabled state:</span>
+  <div>
+    <button class="btn" [disabled]="!selectedNodeForDisabling" [class.btn-success-outline]="isSelectedNodeDisabled"
+      [class.btn-danger-outline]="!isSelectedNodeDisabled" (click)="toggleDisableStateOfSelectedNode()">
+      {{ isSelectedNodeDisabled ? "Enable" : "Disable" }} node {{ selectedNodeForDisablingName }}
+    </button>
+  </div>
+  <clr-tree-node [clrSelected]="false" (click)="selectNodeForDisabling($event)">
```
gnomeontherun: I'd much rather this demo be simpler, where you have a button to set the disabled state on a particular node (or several buttons to show different nodes) rather than the specialized logic and display changes to make the titles clickable. I don't want our demos to override the style or behavior of a component just to demonstrate a different feature, as people sometimes see these dev examples and think its a feature. Also it would make it easier to have a demo for each of the permutations.

src/dev/src/app/tree-view/eager-declarative-tree/eager-declarative-tree.ts
```diff
@@ -1,9 +1,10 @@
+import { ClrTreeNode } from './../../../../../clr-angular/data/tree-view/tree-node';
```
gnomeontherun: import from `@clr/angular`

src/dev/src/app/tree-view/eager-declarative-tree/eager-declarative-tree.ts
```diff
@@ -1,9 +1,10 @@
...
 /*
  * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
```
gnomeontherun: Need to bump copyright year on any of these touched files to 2019.

src/dev/src/app/tree-view/tree-view.demo.scss
```diff
@@ -52,3 +52,12 @@
     }
   }
 }
+
+.example-disabled-nodes {
+  & ::ng-deep div.clr-treenode-content {
```
gnomeontherun: If we update the demo, these styles could be removed.

src/clr-angular/data/tree-view/models/tree-node.model.spec.ts
```diff
@@ -132,5 +132,119 @@ export default function(): void {
...
+      });
+
+      it('allows to set the disabled state of a node without propagating', function() {
+        child.setDisabled(true, false, false);
+        expect(child.disabled.value).toBe(true);
+        [root, ...child.children].forEach(n => expect(n.disabled.value).toBe(false));
+      });
+
+      it('emits disabled changes only when it actually changes', function() {
+        let nbChanges = 0;
```
coryrylan: nit: avoid abbreviations, `nbChanges` => `numberOfChanges` or `count`

gnomeontherun:
@ps37 Any updates on your side about taking a look at the reviews?

gnomeontherun:
Hello @ps37 do you have the ability to review the comments and address them? Thanks!

ps37:
@gnomeontherun I live in Boston but I was out in California for last few weeks because of a family emergency and I came back to work today.
I have the ability to review and address the comments. I just need sometime to catch up with work and I will address them by next weekend. Is that okay?

gnomeontherun:
Hope all is well! I just wanted to make sure we didn't lose track of things here, so take the time you need. We just are trying to keep track of status on various efforts. Please let us know when you have a chance to update and we'll review!

gnomeontherun:
Hi @ps37 just checking in, anything we can help with?

ps37:
Hi Jeremy,

Nothing as of now. I will address the comments by next week. Will ping you
if I need any help.

Thank you,
Prudhvi.

ps37:
> Hi @ps37 just checking in, anything we can help with?

@gnomeontherun I apologize for the delay. Just addressed review comments.

gnomeontherun:
@ps37 Hey there, we'd like to get this either closed or merged, its been open for many months now. Are you able to rebase onto master, squash your commits, and update the testing to be fixed? We'll have to do another review round at this stage since its been a while and we need to check that everything is in sync with latest master.

Otherwise, we might end up pulling it through the last mile here to get it resolved. Thanks!

ps37:
> @ps37 Hey there, we'd like to get this either closed or merged, its been open for many months now. Are you able to rebase onto master, squash your commits, and update the testing to be fixed? We'll have to do another review round at this stage since its been a while and we need to check that everything is in sync with latest master.
> 
> Otherwise, we might end up pulling it through the last mile here to get it resolved. Thanks!

I can squash my commits, rebase onto master and post it for review. However, I need help with the following though:
- update the testing to be fixed?: What do you mean by this?
- there is a failing visual test on set4 ✘ tree-view Eager declarative tree default [chrome]: I am not sure how to get this fixed.

gnomeontherun:
The Travis job shows two problems in testing.

1. Golden file changes so run `npm run golden:fix` to update that file for us.
2. Gemini tests, follow https://github.com/vmware/clarity/wiki/CSS-Regression-Testing (you only need to run it for `set4`)

Then rebase onto master and squash your commits and we can verify everything again looks good.

ps37:
@gnomeontherun Can you please review this change?

gnomeontherun:
It is on my list to do this week.

ps37:
@gnomeontherun, I am Just wondering if I am supposed to do anything with this review?

gnomeontherun:
I've found 3 issues by testing it out and am spending some time to look at the code to identify sources.

1) In eager declarative/recursive mode, the demo has a problem. Look at the demos and start by disabling a node. Then select all, which correctly shows the top level as indeterminate. However, clicking on the top level to unselect no longer works.
2) In lazy declarative/recursive mode, a different issue occurs from the top level item. To see it, start by selecting item `1`, which shows the top level item as checked. Then expand that node and you see the top level item remains checked even though a child node is not selected. Then expand `1.2` and you'll see the top level changes to indeterminate. Click the top node again and it goes to checked, and then can be clicked again to uncheck the whole tree. 
3) The Lazy recursive async tree with disabled nodes is using the same model as the async tree, and doesn't actually show disabling of nodes. This needs to be updated.

Now in lazy mode, Clarity has no idea what the tree model is and depends on the application to tell it what to do. The basic rule here is that if 1+ of a nodes children are selected but not all (regardless of disabled status), that node should be indeterminate. You can see this state below easily happens and is not what correctly reflects the selection state.

image

Rather it should be like this 

image

gnomeontherun:
As soon as #3851 is merged, I'm going to take a look at this to see if we can resolve the conflicts and issues to close this out.

gnomeontherun:
Putting some notes here for posterity. 

The first problem of not being unable to deselect a parent node that has disabled children is a logical trap we're in right now. Its still accounting for all disabled children's state when propagating and just reasserts the same indeterminate state. In other words, if all enabled tree nodes are selected but not disabled nodes, the root node is indeterminate (as expected) but selecting the top node triggers a call to the children to change to selected (which all of them already are) except the disabled ones, so we return to the same state again. 

Looking at a solution to check the state of children nodes before deciding if the change from indeterminate state should go to selected or unselected, and ignoring the state of disabled items during this calculation (but not during the actual state propagation).

gnomeontherun:
I made a few other small tweaks and identified that the other issue is in the demo infinite-tree, and not inside of the Tree View component itself. The demo infinite-tree model has the same basic problem above for not checking if it has any disabled children in its own state calculation. This shouldn't be an issue if the state of a checkbox is managed by the server but our demo needs to have something functional here.

gnomeontherun:
@ps37 what happened here? I've got your original commit plus a few updates to address some of the problems. I haven't solved the last issue yet, which is why this was still open.

ps37:
@gnomeontherun I apologize for the confusion. I wanted to get latest code on Clarity's master as I wanted to work on something and **I accidentally closed the merge request** by force pushing onto my forked repo's master.
I am unable to open this pull request as It seems that re-opening a pull request is bugged if the branch changes after the pull request was closed.
I reverted my master to a state having the changes from my original commit and I opened a new pull request(https://github.com/vmware/clarity/pull/4209) that is a continuation of this pull request.

PR #3283, opened 2019-04-09
[ng] Accessible dropdown

jeeyun:
This is a new PR to continue where https://github.com/vmware/clarity/pull/3252 dropped off.

I squashed the earlier commits by @youdz into one commit, and adding top of that for review purposes. We'll eventually squash and rebase before merging to master.


src/clr-angular/popover/dropdown/dropdown-menu.ts
```diff
@@ -15,17 +29,19 @@ import { POPOVER_HOST_ANCHOR } from '../common/popover-host-anchor.token';
...
     injector: Injector,
     @Optional()
     @Inject(POPOVER_HOST_ANCHOR)
-    parentHost: ElementRef,
+    parentHost: ElementRef<HTMLElement>,
     @Optional()
     @SkipSelf()
-    nested: ClrDropdownMenu
+    nested: ClrDropdownMenu,
+    focusHandler: DropdownFocusHandler
```
gnomeontherun: nit `private focusHandler: DropdownFocusHandler` so we don't have to manually assign it later.

src/clr-angular/popover/dropdown/dropdown-menu.ts
```diff
@@ -86,4 +103,19 @@ export class ClrDropdownMenu extends AbstractPopover {
...
   }
+
+  private focusHandler: DropdownFocusHandler;
+  @ContentChildren(FocusableItem) items: QueryList<FocusableItem>;
+
+  ngAfterContentInit() {
+    this.focusHandler.container = this.el.nativeElement;
+    this.items.changes.subscribe(() => this.focusHandler.addChildren(this.items.toArray()));
+    // I saw this on GitHub as a solution to avoid code duplication because of missed QueryList changes
+    this.items.notifyOnChanges();
```
gnomeontherun: Should this go inside of the changes handler on line 112?

src/clr-angular/popover/dropdown/dropdown-menu.ts
jeeyun: So if I do that, TS compiler complains:
```A 'super' call must be the first statement in the constructor when a class contains initialized properties or has parameter properties.
```
I don't want to move up the `super` call, because I think the warning should fire rather than the `super` call if parentHost is not there. Is it ok to leave this as is?

package-lock.json
```diff
@@ -7286,20 +7286,21 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
```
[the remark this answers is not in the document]
jeeyun: Will revert!

src/clr-angular/popover/dropdown/dropdown-menu.ts
jeeyun: No. Actually that line is what's causing the change event to fire.

jeeyun:
Just updated it with dox on a11y:

Screen Shot 2019-05-14 at 10 40 34 AM

We are going to focus on documenting it for our Angular components for now. If there's great demand for guidance / code example on how to make our static components accessible, we will revisit this.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * If the dropdown was opened by clicking on the trigger, we automatically move to the first item
+   */
+  moveToFirstItemWhenOpen() {
+    this.ifOpenService.openChange.subscribe(open => {
+      if (open && this.ifOpenService.originalEvent) {
+        // Even if we properly waited for ngAfterViewInit, the container still wouldn't be attached to the DOM.
+        // So setTimeout is the only way to wait for the container to be ready to move focus to first item.
+        setTimeout(() => {
+          this.focusService.moveTo(this);
```
Shijir: I see now two methods that use `setTimeout`. If possible, we should try to avoid it as much as possible. I understand why you needed them. But it seems like we could avoid both usages by placing these subscriptions in the container setter: 
```ts
set container(el: HTMLElement) {
  if(this.openChangeSubscription) {
       this.openChangeSubscription.unsubscribe();
  }
  this.openChangeSubscription = this.ifOpenService.openChange.subscribe(open => {  ...  });
  ....
}
```

src/clr-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -3,20 +3,36 @@
...
   ) {}
 
+  @Input()
+  set disabled(value: boolean | string) {
+    // Empty string attribute evaluates to false but should disable the item, so we need to add a special case for it.
+    this.focusableItem.disabled = !!value || value === '';
+  }
+
   ngAfterViewInit() {
     this.renderer.listen(this.el.nativeElement, 'click', () => this.onDropdownItemClick());
```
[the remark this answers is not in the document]
jeeyun: Good catch. I'm going to call the unlisten method that gets returned from this call in the `ngOnDestroy`.

src/clr-angular/popover/dropdown/dropdown-trigger.ts
```diff
@@ -17,16 +18,24 @@ import { ClrDropdown } from './dropdown';
     '[class.dropdown-item]': '!isRootLevelToggle',
     '[class.expandable]': '!isRootLevelToggle',
     '[class.active]': 'active',
+    '[attr.aria-haspopup]': '"menu"',
+    '[attr.aria-expanded]': 'ifOpenService.open',
```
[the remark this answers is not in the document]
jeeyun: Yup, working fine in prod mode. Maybe the host binding has different rules than the template?

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
jeeyun: This doesn't work, because the openChange fires already once before the set container is called that we want to catch and act upon. I tried to make it execute the logic that the subscription would execute in this setter and then subscribe for the future events, but still some of the other states are inconsistent at that time of initial execution and produces error. For now, I'll create a separate ticket to see if we can clean up the setTimeout logic at a later time.

jeeyun: New issue filed: https://github.com/vmware/clarity/issues/3362

src/clr-angular/popover/dropdown/dropdown-trigger.ts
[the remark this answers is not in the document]
Shijir: Most probably 😄

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
Shijir: Sounds good. Thanks, Jeeyun 👌

Issue #936, opened 2017-05-16
Clarity icons getter functions naming causing webpack/uglify to manipulate quotes around function names, causing catastrophic failure in browser


miclettej:
**Select one ...**  (check one with "x")
```
[x ] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
Browser should be able to load clarity icons index.js files without error

### Actual behavior
When using webpack to build with clarity icons, webpack is grabbing the latest version of uglify. The uglify version is incorrectly processing the icon getter functions because they are named with a hyphen (get ellipses-vertical...). This causes a fatal error in the browser, where the console shows an error "Unexpected Token" at the first instance of one of these functions in index.js.

If we go into node_modules/webpack after running npm install and then force webpack to use uglify 2.6.0 the issue is no longer reproducible.

### Reproduction of behavior
This is a build related issue

### Environment details

* **Angular version:** 4.0.0

* **Clarity version:** 0.9.0, 0.9.2, 0.9.4

* **OS and version:** 

* **Browser:** [all]

miclettej:
Dropping bug. Got a sound workaround from another team. Including the "uglifyjs-webpack-plugin":"^0.4.3"  module in our package.json and pointed webpack to use this alleviates the issue.

PR #3294, opened 2019-04-13
Improved icon typings.

lukeshiru:
- Improved icon typings to have infered keys. This can be used to check icons in compile time.
- Added VSCode setting to use local TypeScript.
- Removed settings.json from .gitignore.
- Added utility to add icons to the global ClarityIcons namespace.
- Consistent property naming for icon shapes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,7 @@ src/clr-angular/package.json
 
 # IDE - VSCode
 .vscode/*
-# !.vscode/settings.json
+!.vscode/settings.json
 # !.vscode/tasks.json
 # !.vscode/launch.json
 # !.vscode/extensions.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "typescript.tsdk": "./node_modules/typescript/lib"
+}
--- a/src/clr-icons/shapes/all-shapes.ts
+++ b/src/clr-icons/shapes/all-shapes.ts
@@ -1,43 +1,49 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { ChartShapes } from './chart-shapes';
-import { CommerceShapes } from './commerce-shapes';
-import { CoreShapes } from './core-shapes';
-import { EssentialShapes } from './essential-shapes';
-import { MediaShapes } from './media-shapes';
-import { SocialShapes } from './social-shapes';
-import { TechnologyShapes } from './technology-shapes';
-import { TextEditShapes } from './text-edit-shapes';
-import { TravelShapes } from './travel-shapes';
+import { ChartShapes, ChartShape } from './chart-shapes';
+import { CommerceShapes, CommerceShape } from './commerce-shapes';
+import { CoreShapes, CoreShape } from './core-shapes';
+import { EssentialShapes, EssentialShape } from './essential-shapes';
+import { MediaShapes, MediaShape } from './media-shapes';
+import { SocialShapes, SocialShape } from './social-shapes';
+import { TechnologyShapes, TechnologyShape } from './technology-shapes';
+import { TextEditShapes, TextEditShape } from './text-edit-shapes';
+import { TravelShapes, TravelShape } from './travel-shapes';
+import safeWindowAdd from '../utils/safe-window-add';
 
-const allShapesSets = [
-  CoreShapes,
-  CommerceShapes,
-  EssentialShapes,
-  MediaShapes,
-  SocialShapes,
-  TechnologyShapes,
-  TravelShapes,
-  ChartShapes,
-  TextEditShapes,
-];
+/**
+ * Valid shapes.
+ */
+export type AllShape =
+  | ChartShape
+  | CommerceShape
+  | CoreShape
+  | EssentialShape
+  | MediaShape
+  | SocialShape
+  | TechnologyShape
+  | TextEditShape
+  | TravelShape;
 
-const allShapes: any = {};
+/**
+ * AllShapes valid type.
+ */
+type AllShapesType = { [shape in AllShape]: string };
 
-for (const set of allShapesSets) {
-  for (const shape in set) {
-    if (set.hasOwnProperty(shape)) {
-      allShapes[shape] = set[shape];
-    }
-  }
-}
+export const AllShapes: AllShapesType = {
+  ...CoreShapes,
+  ...CommerceShapes,
+  ...EssentialShapes,
+  ...MediaShapes,
+  ...SocialShapes,
+  ...TechnologyShapes,
+  ...TravelShapes,
+  ...ChartShapes,
+  ...TextEditShapes,
+};
 
-if (typeof window !== 'undefined' && window.hasOwnProperty('ClarityIcons')) {
-  window.ClarityIcons.add(allShapes);
-}
-
-export { allShapes as AllShapes };
+safeWindowAdd(AllShapes);
--- a/src/clr-icons/shapes/chart-shapes.ts
+++ b/src/clr-icons/shapes/chart-shapes.ts
@@ -1,10 +1,10 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { descriptorConfig } from '../utils/descriptor-config';
 import { clrIconSVG } from '../utils/svg-tag-generator';
+import safeWindowAdd from '../utils/safe-window-add';
 
 /* tslint:disable:variable-name */
 export const ClrShapeAxisChart = clrIconSVG(`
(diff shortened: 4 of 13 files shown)


coryrylan:
Thank you @lukeshiru for contributing. It looks like there is an issue with the build in this PR. Taking a look at it its looks like it is related to the TypeScript package change you made. Please remove the TS package update from the PR as we manage dependency updates separately and this should allow your PR to build in Travis.

lukeshiru:
@coryrylan done! The TypeScript update to `3.4.x` should happen at some point in the near future, to make functionality like *const assertions* to work. Its a minor update, so it should go smoothly.

src/clr-icons/shapes/all-shapes.ts
```diff
@@ -24,20 +44,9 @@ const allShapesSets = [
...
-  }
-}
-
-if (typeof window !== 'undefined' && window.hasOwnProperty('ClarityIcons')) {
-  window.ClarityIcons.add(allShapes);
-}
+]
+  .map(set => Object.keys(set).map(shape => ({ shape, value: set[shape] })))
+  .reduce((shapes, shape) => [...shapes, ...shape], [])
+  .reduce((shapes, { shape, value }) => ({ ...shapes, [shape]: value }), {} as AllShapesType);
```
Shijir: Isn't it overcomplicating things here? As far as I can see, this can be simply: 
```
const allShapes: any = {
  ...CoreShapes,
  ...CommerceShapes,
  ...EssentialShapes,
  ...MediaShapes,
  ...SocialShapes,
  ...TechnologyShapes,
  ...TravelShapes,
  ...ChartShapes,
  ...TextEditShapes
};

```

Shijir:
Hi @lukeshiru, Thank you for your contribution. This PR brings nice cleanup and organization in the icons codebase. I have only one questions I posted above. And I am almost ready to approve PR.

src/clr-icons/shapes/chart-shapes.ts
```diff
@@ -222,24 +222,26 @@ export const ClrShapeBellCurve = clrIconSVG(
...
+  'heat-map': ClrShapeHeatMap,
   'line-chart': ClrShapeLineChart,
   'pie-chart': ClrShapePieChart,
-  'tick-chart': ClrShapeTickChart,
   'scatter-plot': ClrShapeScatterPlot,
-  'box-plot': ClrShapeBoxPlot,
-  'heat-map': ClrShapeHeatMap,
-  'bell-curve': ClrShapeBellCurve,
+  'tick-chart': ClrShapeTickChart,
+  /** Alias for: line-chart */ 'analytics': ClrShapeLineChart,
```
Shijir: Oops, I overlooked this one. This should point to the reference of the shape. That's why it was like this:

```
Object.defineProperty(ChartShapes, 'analytics', descriptorConfig(ChartShapes['line-chart']));
```

Otherwise, it will dump its SVG markup again in the published .js file that will dramatically increase the file size.

lukeshiru: The issue with using `Object.defineProperty` is that you loss TypeScript inferance, because it can't predict what are you doing in there. The effect of the following 3 examples is exactly the same:
```js
const foo = "foo";
const bar = "bar";

const example1 = { foo: foo, bar: bar, baz: bar };
// { foo: "foo", bar: "bar", baz: "bar" }

const example2 = { foo: foo, bar: bar };
example2.baz = bar;
// { foo: "foo", bar: "bar", baz: "bar" }

const example3 = { foo: foo, bar: bar };
Object.defineProperty(example3, 'baz', descriptorConfig(example3.bar));
// { foo: "foo", bar: "bar", baz: "bar" }
```

I even took a look at the minified code in `@clr/icons/shapes/all-shapes.min.js`, and this doesn't have any kind of performance effect.

src/clr-icons/shapes/all-shapes.ts
lukeshiru: You're right, I tried to kind of match the old logic with better techniques instead of just optimizing it 100%. Fixed in the latest push, thanks for the suggestion!

src/clr-icons/shapes/chart-shapes.ts
Shijir: > The effect of the following 3 examples is exactly the same

example 3 is different as it doesn't assign `baz` **the value of bar**, but the reference of  bar.  It's the same as using object literal getter: 
```
const example3 = {
   foo: foo, 
   bar: bar,
   get baz() {
      return this.bar;
   }
};
```
As icons have a large svg templates, we should only reference to the existing shapes in the set, instead of dumping the same markup for the aliases, which would result in larger file `.min.js` sizes.

Also, I just remember why we couldn't just use object literal `getter` at the first place and had to resort to `Object.defineProperty`. It was because UglifyJS had a bug/problem that doesn't recognize getter properties with `-` in it and creates runtime errors when users try to use them.
So I am afraid we couldn't use your changes on these aliases' part. I am open to your suggestions though.

lukeshiru: I tested the getter approach with UglifyJS and it works as expected:
```js
const foo = "foo";
const bar = "bar";
const example = {
	foo: foo,
	bar: bar,
	get "this-works"() {
		return this.bar;
	},
	get thisAlsoWorks() {
		return this.bar;
	}
};
```
Returns:
```js
const foo="foo",bar="bar",example={foo:foo,bar:bar,get"this-works"(){return this.bar},get thisAlsoWorks(){return this.bar}};```

I feel is the best one, because it complies with the minifying requirements and also with the typings for TS.

File: src/clr-icons/shapes/essential-shapes.ts [hunk 3]
```diff
@@ -902,150 +896,183 @@ export const ClrShapeDotCircle = clrIconSVG(
...
+  /** Alias for: pencil */ get edit(): string {
+    return this['pencil'];
+  },
+  /** Alias for: pinboard */ get pinned(): string {
+    return this['pinboard'];
+  },
+  /** Alias for: plus */ get add(): string {
+    return this['plus'];
+  },
+  /** Alias for: resize */ get 'resize-up'(): string {
```
Shijir: As I said before, this is syntactically nicer, and we could leverage TS by doing this way. But we have already tried this in the past. And this got us bad when Webpack/UglifyJS did aggressive minification resulting in runtime errors: 
https://github.com/vmware/clarity/issues/936
https://github.com/vmware/clarity/pull/945

So before getting these changes in, we have to make sure it works as expected with no errors on our end.

lukeshiru: ok, let me know when you fixed the bugs in your WebPack configuration and I'll rebase this PR to latest.

hippee-lee: Webpack config is the standard webpack for the Angular-CLI. Its not somehting we have access to change. 

I _think_ what @Shijir means is that your changes need to work with whats already in place for testing, building and publishing).

lukeshiru: I ran `npm run build:icons` with no issues, it generated the corresponding files (even the minified ones), is there any other script I have to run to make sure this works?

src/clr-icons/shapes/core-shapes.ts
```diff
@@ -304,54 +303,84 @@ export const ClrShapeStepForward2 = clrIconSVG(
...
+  },
+  /** Alias for: times */ get close(): string {
+    return this['times'];
+  },
+  /** Alias for: user */ get avatar(): string {
+    return this['user'];
+  },
 };
 
-Object.defineProperty(CoreShapes, 'house', descriptorConfig(CoreShapes.home));
```
gnomeontherun: Due to concerns about regressions, please revert the `Object.defineProperty` calls in all the files and remove the getters. We have to test these changes extensively by building the icons and importing them into several different types of applications (both Angular CLI, standalone) and that work will be done as part of https://github.com/vmware/clarity/issues/2599.

.gitignore
```diff
@@ -27,7 +27,7 @@ src/clr-angular/package.json
 
 # IDE - VSCode
 .vscode/*
-# !.vscode/settings.json
+!.vscode/settings.json
```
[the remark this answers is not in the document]
lukeshiru: Line 29 makes all files inside the .vscode file to be ignored, 30 makes it so settings.json is not ignored, and the IDE takes the local TypeScript installation instead of the one bundled with VSCode. This makes TypeScript related functionality consistent between all VSCode developers.

In old versions of VSCode, things like `keyof typeof ChartShapes` would be highlighted as an error, so this setting enforces that everyone is using the correct version of TypeScript. Is a common good practice with VSCode+TypeScript to set the tsdk to be the local installation.

src/clr-icons/shapes/core-shapes.ts
lukeshiru: This PR can wait until all fixes are done, as I said in a previous comment. Rolling back to Object.defineProperty not only makes the purpose of this PR meaningless (because typing will not work as expected, due to lack of type inference with Object.defineProperty on TypeScript), but also after the build steps are fixed, this will have to be "rolled back again" to getters, so is better option to just wait until the build is fixed.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "./node_modules/typescript/lib"
```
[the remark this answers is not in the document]
lukeshiru: Covered in this comment: https://github.com/vmware/clarity/pull/3294#discussion_r280605206

src/clr-icons/shapes/core-shapes.ts
Shijir: I agree with @gnomeontherun's comment. Also, I don't see this PR becoming meaningless if do reverts as he suggested. The value we were are gaining from these type enforcement on `getters` is not worth it compare to the potential breaks we might introduce to the library. The UglifyJS is one of the deeply nested dependencies so we don't have much control over it. We are especially wary of this change at this moment because we are in the process of delivering Clarity 2.0. 

We could introduce this kind of type enforcement changes when we start tackling #2599. By then, we will run extensive integration, unit, and other tests.

lukeshiru: This PR makes it so when you import any set of icons, you have auto-completition of the icons in that set, and errors if you try to use an icon that doesn't exist. TypeScript as I said before can't infer anything from `Object.defineProperty` so all alias will be marked as "errors" and will not get auto-completition.

I'll wait until you solve the issues you have with your build (I didn't find any running the build, but I take your word on that), and once it works with getters and setters (kind of a core functionality nowadays), I'll update this PR.

gnomeontherun: The build runs just fine, but consuming the build in a new project and testing it is not something we have built into our primary testing pipeline right now.

coryrylan:
Due to the work and research being worked to fix the tree shaking issue https://github.com/vmware/clarity/issues/2599 we are closing the PR until we are ready to address it. I opened an issue https://github.com/vmware/clarity/issues/3337 for us to track. When we implement the tree shaking changes that will be the optimal time to also address the typings issues. This PR is still very valuable and will give us context of what we should be aware of to properly fix typings and the tree shaking issues.

PR #3295, opened 2019-04-13
[NG] refactor if expanded directive

### LINKED ISSUE #893: [NG] Relocate the expand util (*clrIfExpanded structural directive) into utils/conditional
## Update
The `*clrIfExpanded` structural directive was extracted from the datagrid and moved to `utils/expanded`. The next step is to consolidate it with the other conditional structural directives in the `utils/conditional` folder. 

## Original
We want to extract the `*clrIfExpanded` structural directive and the `RowExpand` provider from Datagrid, so that any other components relying on an expand/collapse pattern can reuse them.

The [structural directive](https://github.com/vmware/clarity/blob/master/src/clarity-angular/data/datagrid/datagrid-if-expanded.ts) will pretty much stay the same. In anticipation of this, the selector doesn't mention "datagrid", but the class name itself will have to change.

The [provider](https://github.com/vmware/clarity/blob/master/src/clarity-angular/data/datagrid/providers/row-expand.ts), on the other hand, has be split into a more generic class, and the Datagrid will extend it to add its own custom features (like `replace`). Just make sure, after extending, that the child class provider is still registered for the parent class token.
The features that should be in the generic provider are:
- `expanded` boolean.
- `expandChange` Observable.
- `LoadingListener`

If components other than Datagrid can use proper Angular animations for expand/collapse, then the `animate` Observable should not be part of the generic provider.

coryrylan:
Move if-expanded to conditionals and make service generic.

closes #893


src/clr-angular/data/datagrid/datagrid-if-expanded.service.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ClrLoadingState } from '../../utils/loading/loading';
+
+@Injectable()
+export class DatagridIfExpandService extends IfExpandService {
+  // due to the es5 spec if the set is overridden on base class the getter must also be overridden
+  get expanded(): boolean {
+    return this._expanded;
+  }
+
+  set expanded(value: boolean) {
```
Shijir: This setter seems to duplicate the base class's setter. Can we directly subscribe to `expandChange` in the `row-expand-animation.ts` so we don't even need to create another subject for animation only?

Shijir:
LGTM. I have only a few questions that I've posted above :)

src/clr-angular/data/datagrid/animation-hack/row-expand-animation.ts
```diff
@@ -11,16 +11,16 @@
...
 export class DatagridRowExpandAnimation {
   constructor(
     private el: ElementRef,
     private domAdapter: DomAdapter,
     private renderer: Renderer2,
-    private expand: Expand
+    private expand: DatagridIfExpandService
   ) {
     if (expand && expand.animate) {
       expand.animate.subscribe(() => {
```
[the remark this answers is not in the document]
Jinnie: This hack class will soon be gone. Not sure it's worth improving it at the moment.

src/clr-angular/utils/conditional/if-expanded.service.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+import { Injectable } from '@angular/core';
+import { Observable, Subject } from 'rxjs';
+
+import { LoadingListener } from '../loading/loading-listener';
+import { ClrLoadingState } from '../loading/loading';
+
+@Injectable()
+export class IfExpandService implements LoadingListener {
+  public expandable: number = 0;
```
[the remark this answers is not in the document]
coryrylan: It looks like the ifExpanded directive uses it to track how many instances of the directive are active. Other components like the dg-row and tree view check for the value.

src/clr-angular/data/datagrid/datagrid-if-expanded.service.ts
coryrylan: Hmm I don't think so, the issue I ran into was that the animation observable needs to fire when `loadingStateChange()` is called to know if it is not loading currently so I couldn't get rid of it. Once the row if updated to use ng animations this should be able to go away altogether which is what I think @Jinnie was referring to.

Shijir: ah I see.

Issue #3279, opened 2019-04-08
clr-tab-content not taking 'height' into consideration


aleklol:
Hi all,

Since switching to Clarity 1.0 I've noticed a bug in the functionality of clr-tab-content.
For some reason, if you specify a 'height' or any style in the 'clr-tab-content', it is not considered in the resulting HTML.
This causes problems because if you use other Angular components inside the tab content, which rely on relative height (for instance height: 40%), it no longer knows relative to what this should be calculated.
For example if I have the following code:
`<div style="height: 100%;">`
`<clr-tabs>`
`		<clr-tab>`
`			<button clrTabLink>Some Tab</button>`
`			<clr-tab-content style="height: 100%;">`
`				<as-split>`
`					<as-split-area [size]="50">`
`						<div style="height: 100%; background-color: green;"></div>`
`					</as-split-area>`
`					<as-split-area [size]="50">`
`						<div style="height: 100%; background-color: orange;"></div>`
`					</as-split-area>`
`				</as-split>`
`			</clr-tab-content>`
`		</clr-tab>`
`	</clr-tabs>`
`</div>`
Only the orange part will show up, because the [size] passed to <as-split-area> is relative

aleklol:
Update:
Here is a link to stackblitz reproducing the issue:
https://stackblitz.com/edit/clarity-u95irx?file=app%2Fapp.component.html

aleklol:
Hi @Jinnie 
One more thing which could be relevant. Prior to bumping Clarity to 1.1, the version where everything worked fine was 0.13.
Cheers,
Alek

Jinnie:
I guess you closed this by mistake with your last comment. Reopening, as it is a valid issue and I'll be submitting a fix soon.

PR #3298, opened 2019-04-15
[NG] Enable tab content to scale to tabs container (#3279)

Jinnie:
The introduction of block-level section element had introduced the
need to explicitly manage the container height. Reverting it to
inline will allow direct visibility between the tab content and tabs
parent container.

### DIFF
--- a/src/clr-angular/layout/tabs/_tabs.clarity.scss
+++ b/src/clr-angular/layout/tabs/_tabs.clarity.scss
@@ -28,6 +28,10 @@ button.nav-link {
   }
 }
 
+.tab-content {
+  display: inline;
+}
+
 .tabs-vertical {
   display: flex;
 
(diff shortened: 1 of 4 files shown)


Shijir:
Hi Ivan, as the tabs component is a layout component, shouldn't we rather have `block`, `flex`, `grid` as a display property instead of inline? Secondly, it seems like our tabs component works just fine if user sets the height on the tab content.

Jinnie:
Hi @Shijir,
1.0.x is fine, the problem is with 1.1.x - https://stackblitz.com/edit/clarity-tabs-with-heigth-1-1-x?file=package.json
About if it would be better to have the tabs as block level - it probably would, but it will be a breaking change, as it will require users to update their current code. This specific change is more aimed at restoring the previous way this component worked.
If you need more information I can point you to the slack discussion we had on this topic, where both these options were compared and discussed.

Shijir:
As per our discussion, I've given it :shipit: . Please make sure it works across different browsers. Thank you :)

PR #3305, opened 2019-04-18
[NG] a11y improvements for control errors

### LINKED ISSUE #3289: A11y - clr-control-error
## Describe the bug

- Currently when an input has a validation error message is rendered by `clr-control-error` it is not read aloud to the screen reader. `clr-control-error` should probably announce to the user via `role="alert"` or `aria-live="assertive"`. (https://v1.clarity.design/input)

- The `clr-control-error` should also associate the message to the input label via `aria-describedby="label-id"`. The associated input can also have a `aria-invalid` attribute when invalid to help announce the invalid state.

- The `clr-control-error` could also have `tabindex="0"` set so the errors easily announced as the user tabs through inputs.

I tested most of this functionality and documented my findings in this blog post
https://coryrylan.com/blog/build-accessible-forms-with-angular

Demo:
https://stackblitz.com/edit/angular-b1kevk 

We should check with the a11y team on the best recommendations and make sure these are reasonable updates to make for `clr-control-error`.

coryrylan:
Minor adjustments for `clr-control-error` to make it more accessible to screen readers.

closes #3289 

### DIFF
--- a/src/clr-angular/forms/common/error.ts
+++ b/src/clr-angular/forms/common/error.ts
@@ -1,16 +1,39 @@
 /**
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Component } from '@angular/core';
+import { Component, Optional, HostBinding, Input } from '@angular/core';
+import { ControlIdService } from './providers/control-id.service';
+import { Subscription } from 'rxjs';
 
 @Component({
   selector: 'clr-control-error',
   template: `
     <ng-content></ng-content>
     `,
-  host: { '[class.clr-subtext]': 'true' },
+  host: {
+    '[class.clr-subtext]': 'true',
+    '[attr.aria-live]': '"polite"',
+  },
 })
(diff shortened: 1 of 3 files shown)


coryrylan:
@hippee-lee It would get projected in unaltered. This is a similar strategy to the auto id mapping between a label and input for the form control wrappers. If the user explicitly set their own input `id` and `describedby` then the `clr-control-error` will leave the `describedby` attr off and not interfere with the custom association created.

src/clr-angular/forms/common/error.ts
```diff
@@ -1,16 +1,39 @@
...
 
 @Component({
   selector: 'clr-control-error',
   template: `
     <ng-content></ng-content>
     `,
-  host: { '[class.clr-subtext]': 'true' },
+  host: {
+    '[class.clr-subtext]': 'true',
+    '[attr.aria-live]': '"polite"',
```
gnomeontherun: Do we want this to be "assertive"?

coryrylan: The issue with assertive currently is if the form validation is triggered multiple times in the form (quickly tab, force form validation programmatically) the assertive will cut off other assertive announcements since ideally there should only be one aria-live per form to handle all messages.

This is a stop-gap for messages to be announced until we can come up with a solution on how to group all of the messages into a single aria-live assertive in a single form.

PR #3312, opened 2019-04-22
datagrid deprecations for row selection and column toggle title/buttons

### LINKED ISSUE #2933: [Deprecation] Remove `rowSelectionMode` from Datagrid
## Summary

The `rowSelectionMode` Input should be removed from ClrDatagrid due to accessibility and maintenance issues. 

Removal of `@Input('clrDgRowSelection')` and the associated property on the `Selection` provider (rowSelectionMode).
* Why should it be removed from Clarity?
1. The clearly breaks accessibility in datagrid rows. 
2. Removes associated complexity in templates and rendering from the datagrid rendering cycle. 
Yes, it removes the ability to click to select a row. 
Yes.

## Use case

n/a

## Examples

n/a

## Workarounds

n/a

### LINKED ISSUE #3304: [Deprecation] Remove Datagrid Column Toggle Button/Title and use CommonStrings
## Summary

We have no need for custom titles and buttons with the CommonStrings pattern, so we should normalize our implementation for the column toggler.

## Use case

Simplicity.

## Examples

n/a

## Workarounds

n/a

gnomeontherun:
Deprecate the row selection and column toggler overrides. Row selection (#2933) is not accessible and column toggler (#3304) is being brought into line with the other i18n strings in Clarity.

This also adds a new i18n page in docs with how to use strings service.

closes #3304
closes #2933

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -28,7 +28,7 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
     <div class="column-switch"
          *clrPopoverOld="open; anchor: anchor; anchorPoint: anchorPoint; popoverPoint: popoverPoint">
       <div class="switch-header">
-        <ng-container *ngIf="!customToggleTitle">Show Columns</ng-container>
+        <ng-container *ngIf="!customToggleTitle">{{commonStrings.showColumns}}</ng-container>
         <ng-content select="clr-dg-column-toggle-title"></ng-content>
         <button
           class="btn btn-sm btn-link toggle-switch-close-button"
@@ -52,12 +52,13 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
       </ul>
       <div class="switch-footer">
         <ng-content select="clr-dg-column-toggle-button"></ng-content>
-        <clr-dg-column-toggle-button *ngIf="!customToggleButton">Select All</clr-dg-column-toggle-button>
+        <clr-dg-column-toggle-button *ngIf="!customToggleButton">{{commonStrings.selectAll}}</clr-dg-column-toggle-button>
       </div>
     </div>
   `,
   host: { '[class.column-switch-wrapper]': 'true', '[class.active]': 'open' },
 })
+/** @deprecated since 2.0, remove in 3.0 */
 export class ClrDatagridColumnToggle {
   /***
    * Popover init
--- a/src/website/src/app/documentation/demos/datagrid/hide-show-columns/hide-show-columns.html
+++ b/src/website/src/app/documentation/demos/datagrid/hide-show-columns/hide-show-columns.html
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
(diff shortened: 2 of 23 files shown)


src/website/src/app/documentation/demos/datagrid/hide-show-columns/hide-show-columns.html
```diff
@@ -70,3 +60,18 @@ <h4>Custom Toggle Strings</h4>
 
 <clr-code-snippet [clrCode]="example"  clrLanguage="html"></clr-code-snippet>
 
+<h4>Custom Toggle Strings</h4>
+
+<clr-alert [clrAlertType]="'danger'" [clrAlertClosable]="false">
+  <div clr-alert-item class="alert-item">
```
coryrylan: Is there any behavioral difference between `<div clr-alert-item class="alert-item">` and using `<clr-alert-item>`?

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -28,7 +28,7 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
     <div class="column-switch"
          *clrPopoverOld="open; anchor: anchor; anchorPoint: anchorPoint; popoverPoint: popoverPoint">
       <div class="switch-header">
-        <ng-container *ngIf="!customToggleTitle">Show Columns</ng-container>
+        <ng-container *ngIf="!customToggleTitle">{{commonStrings.showColumns}}</ng-container>
```
Shijir: Now they really don't need to have `*ngIf="!customToggleTitle"` right? because they don't need to project their custom content in order to provide non-english text.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -52,12 +52,13 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
       </ul>
       <div class="switch-footer">
         <ng-content select="clr-dg-column-toggle-button"></ng-content>
-        <clr-dg-column-toggle-button *ngIf="!customToggleButton">Select All</clr-dg-column-toggle-button>
+        <clr-dg-column-toggle-button *ngIf="!customToggleButton">{{commonStrings.selectAll}}</clr-dg-column-toggle-button>
```
Shijir: Same here? We no longer need to use `*ngIf` and `@ContentChild`

src/website/src/app/documentation/demos/datagrid/hide-show-columns/hide-show-columns.html
mathisscott: IIRC, `<div clr-alert-item...` was the old way. it has been deprecated. As long as it has the `alert-item` class it should still look ok. The CSS is on that classname. But `<clr-alert-item>` is the preferred way to do alerts.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
gnomeontherun: This is a non-breaking change for the deprecation, which gives the consumers a cycle to migrate to common strings patterns. It should remain unless you are voting to make the breaking change now?

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
hippee-lee: I'm just thinking outloud here but, what if a consumer needs/wants still need to override the common string, just for one datagrid in the app but not others?

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
hippee-lee: I vote for non-breaking. I'm not 💯 this is even a valid use case but, I'm still wondering if a consumer decides/needs to override the common strings just for one datagrid.

src/website/src/app/documentation/demos/datagrid/hide-show-columns/hide-show-columns.html
gnomeontherun: Yeah, I copied this from another alert and should update them.

src/website/src/app/documentation/documentation-routing.module.ts
```diff
@@ -150,6 +150,14 @@ const documentationRoutes: Routes = [
           browserTitle: 'Inputs',
         },
       },
+      {
+        path: 'internationalization',
```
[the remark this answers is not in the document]
gnomeontherun: This is part of the URL and thus SEO.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
gnomeontherun: This is not something we support anywhere with the string service, if you reuse any component with the string its always the same. If this use case comes up, we'd have to design a different solution, but for now I say its not valid.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
gnomeontherun: @Shijir We need the ngIf for backwards compat for the deprecation. 
@hippee-lee I've never heard that ask, but have heard someone want to override the string once instead of in every datagrid like this old pattern requires.

PR #3316, opened 2019-04-24
[WEBSITE] view encapsulation for documentation component styles

coryrylan:
- disable module preload for improved initial startup performance
- split component specific styles to use view encapsulation instead of being global (home, community, news, icons). Also deleted scss files that no longer used in the docs site. (will help with future updates/refactors with the website)
- CSS did not change much, mostly moving files around.


gnomeontherun:
Are we sure there are no conflicts with removing some of the sass files here?

coryrylan:
Conflicts as in the CSS was being used elsewhere? For the files that were deleted and not moved to their specific components, I searched the repo for any reference to those selectors and did not find any.  Example `.layout-get-started` and `.layout-error-page` were not found in the repo except in those sass files.

mathisscott:
> Conflicts as in the CSS was being used elsewhere? For the files that were deleted and not moved to their specific components, I searched the repo for any reference to those selectors and did not find any. Example `.layout-get-started` and `.layout-error-page` were not found in the repo except in those sass files.

Those `.layout-*` are only used on the website and only on the logical pages. So `.layout-error-page` would be used on our 404 and 500 pages. `.layout-get-started` is only used for our get-started page, etc.

coryrylan:
I double checked those pages this morning just to be sure. The error component has its own component level styles and its layout selector is `layout-error`. The selectors leftover in `error-pages.scss` does not match anywhere.

The `layout-get-started` is not used as the body selector was commented out. Commenting back in it still is not used as the selectors in the `get-started.scss` file don't match what is on the actual getting started page.

It looks like these were leftover files from a previous design that no longer is being used.

Issue #3273, opened 2019-04-04
[Tabs][Datagrid]Initial column width for datagrid column appearing as zero


winstonfernandes:
## Describe the bug
When datagrid is placed within tabs the initial width of column header is appearing as zero due to this there are alignment issues when minimum column width is provided. When the column
 is resized the widths are updated correctly
## How to reproduce

https://stackblitz.com/edit/clarity-dark-theme-v1-0-datagrid-issue

Check the width of the column during after initial rendering.

1.  Create datagrid in tabs
2. check the column width
## Expected behavior

The column width should be non-zero after initial rendering

## Versions

App

Angular: 7.2
Clarity: 1.1.2
Device:

OS: [Windows Server 2012]
Browser [Chrome]
Version 72.0.3626.121

![Width_Zero](https://user-images.githubusercontent.com/48466504/55558756-68bcbd80-570a-11e9-8f1b-7235d8a62d4f.PNG)

hippee-lee:
Hi there, 

I looked at the example (thank you for that). It looks like there were some implementation issues that were causing the misalignment. It's hard to tell what the data should be but when I clean up the template and declared the datagrid as described in the documentation for [binding properties](https://clarity.design/documentation/datagrid/binding-properties) column widths work as intended. 

I'm closing this as it doesn't seem to be an issue. If you have more questions about getting the datagrid up and running please [ask on stack overflow and tag it with `vmware-clarity`.](https://stackoverflow.com/questions/tagged/vmware-clarity). If I misunderstood what the datagrid was trying to do please let me know. 

Here is a working version of the original example: https://stackblitz.com/edit/gh-3273-datagrid-column-width-question

winstonfernandes:
The working version still shows the width as zero
![image](https://user-images.githubusercontent.com/48466504/55604330-d0b3e800-578c-11e9-925b-6283621c0feb.png)

winstonfernandes:
Proper width is shown on resizing any column. Please let me know if my understanding is wrong. There will be alignment issues if min-width is required on the column. The screenshot after resizing is attached.
![image](https://user-images.githubusercontent.com/48466504/55604448-7f582880-578d-11e9-9a3b-70ca6146ceaf.png)

hippee-lee:
There is something odd with your example code but I didn't dig deep enough to figure out exactly what. I checked with a simple datagrid that column widths are calculating as expected: https://stackblitz.com/edit/datagrid-simple-columnwidth

[...]

PR #3318, opened 2019-04-24
[NG] Fix datagrid column sizing inside *clrIfActive (#3273)

Jinnie:
Datagrid column sizing lifecycle was calculating column widths
when tab-content was not yet attached to the DOM, resulting in
width: 0px.
Partly reverted the projection back to ng-content.
Added integration test to verify this scenario.


src/clr-angular/data/datagrid/datagrid-tabs.integration.spec.ts
```diff
@@ -0,0 +1,57 @@
+/*
```
hippee-lee: This test belongs in `datagrid.spec.ts` that's where we do integration testing for datagrid where we need to do more than a pure unit test on a datagrid component. Also, I'm not a fan of adding the `*.integration.spec` filename to the project.

src/clr-angular/data/datagrid/datagrid-tabs.integration.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        </clr-tab-content>
+      </clr-tab>
+    </clr-tabs>
+    `,
+})
+class IntegrationTest {
+  items = Array(10).fill(0);
+}
+
+// Verifies issue #3273: https://github.com/vmware/clarity/issues/3273
```
hippee-lee: Please remove this its not needed here.

src/clr-angular/data/datagrid/all.spec.ts
```diff
@@ -35,6 +35,7 @@ import DatagridPlaceholderSpecs from './datagrid-placeholder.spec';
 import DatagridRowDetailSpecs from './datagrid-row-detail.spec';
 import DatagridRowSpecs from './datagrid-row.spec';
 import DatagridSpecs from './datagrid.spec';
+import DatagridTabsIntegrationSpecs from './datagrid-tabs.integration.spec';
```
hippee-lee: This test can be moved into `datagrid.spec.ts`.

hippee-lee: This test can be moved into `datagrid.spec.ts`.

hippee-lee: This test can be moved into `datagrid.spec.ts`.

src/clr-angular/data/datagrid/datagrid-tabs.integration.spec.ts
Jinnie: Moved!

src/clr-angular/data/datagrid/datagrid-tabs.integration.spec.ts
Jinnie: Removed!

src/clr-angular/data/datagrid/all.spec.ts
Jinnie: Done, done and done :)

PR #3319, opened 2019-04-24
[ng] remove deprecated dropdown

jeeyun:
We have a legacy markup for Angular version of the dropdown which used a regular div with `class='dropdown-menu`. Deprecating as it hasn't been officially supported for a while.

Please note that the css class is still valid for the static markup. We're only removing support for it in the Angular component.


gnomeontherun:
Did you check the website for old examples of this?

jeeyun:
@gnomeontherun yup. doesn't even exist anywhere in our dox. it was only present in our demo/dev app.

PR #3323, opened 2019-04-26
[UI] update to use dart sass package

coryrylan:
The latest Angular CLI 8 release removed its use of node-sass and the c++ binaries in favor of the Dart Sass implementation. This will be better for consumers as the Dart version compiles to JavaScript so users will no longer have to download or compile the c++ binaries locally.
https://github.com/angular/angular-cli/commit/ce158990202428ae1d893787bcbcbfc763e99bb4

Unfortunately there are some small differences between the two implementations which means if you use the latest CLI today and pull in our Sass for themes it will not compile in the Dart implementation.
This PR moves Clarity UI and its dependencies to use use the `sass` package.

There will be a follow up PR to update the Clarity documentation for the latest theming setup.


src/clr-angular/data/_tables.clarity.scss
```diff
@@ -81,7 +81,7 @@
     }
   }
 
-  &#{$table}-compact {
+  #{$table}-compact {
```
coryrylan: Dart Sass Error:
`Top-level selectors may not contain the parent selector "&"`

src/clr-angular/utils/_helpers.clarity.scss
```diff
@@ -13,15 +13,6 @@
   @return $remSize * $clr-font-base-size * 1px;
 }
 
-//Get a selector for all the input types specified
-@function inputs($types...) {
-  $inputs: ();
-  @each $type in $types {
-    $inputs: $inputs, unquote('input[type=' + $type + ']');
-  }
-  @return quote($inputs);
```
coryrylan: Dart Sass implementation of quote did not work with maps. I removed the `inputs` function altogether as it was only being used in the Stack View component.

src/clr-angular/emphasis/_labels.clarity.scss
```diff
@@ -5,12 +5,12 @@
 // TODO: Split badges and labels into separate files.
 
 @mixin generateLabelStyles($color: gray, $alias: 1) {
-  &.label-#{$color},
+  &.label-#{'' + $color},
```
coryrylan: Hard error with this color variable in the dart sass version.

`You probably don't mean to use the color value orange in interpolation here.
It may end up represented as orange, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "orange").
If you really want to use the color value here, use '"" + $color'.`

PR #3329, opened 2019-04-30
[NG] Lifecycle fix for tabs templating initialization (#3273)

Jinnie:
Viewchild setter initialization of ref in service, as advised by Shijir.
This overcomes latency in the initialization of the container vs the
templated content.


src/clr-angular/layout/tabs/tab.spec.ts
```diff
@@ -15,12 +15,28 @@ import { TABS_ID_PROVIDER } from './tabs-id.provider';
...
   template: `
+        <ng-container #tabContentViewContainer></ng-container>
+   `,
+})
+class TestWrapper {
+  @ViewChild('tabContentViewContainer', { read: ViewContainerRef })
+  set tabContentViewContainer(value: ViewContainerRef) {
+    this.tabsService.tabContentViewContainer = value;
+  }
+  constructor(private tabsService: TabsService) {}
```
coryrylan: Changes look good to me, only request would maybe we put a comment in the tab or test describing that this is a fix due to a lifecycle issue. It adds necessary complexity that without context of the original issue may not be clear.

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -3,35 +3,44 @@
...
+    private tabsService: TabsService
   ) {
     if (!this.tabContentId) {
       this.tabContentId = 'clr-tab-content-' + nbTabContentComponents++;
     }
   }
 
+  @ViewChild('tabContentProjectedRef')
+  private set templateRef(value: TemplateRef<ClrTabContent>) {
+    this.tabsService.tabContentViewContainer.createEmbeddedView(value);
```
Shijir: I think we are forgetting to clear the view when the tab content becomes inactive. So as a result, it will embed the same view on every time the tab content instantiates with `*clrIfActive`.

Jinnie: I can't reproduce this. The reference gets overwritten on tab change with the new template reference. Can you provide more details? Or explain how do you imagine a test for this?
I don't mind a proper cleanup, but want to be sure I understand you well.

Shijir: In tabs with `*clrIfActive`, if you leave a tab and come back to it, you would see the content being appended over and over again. So the test should check we have only one embedded tab content no matter how many times we come back to the same tab.

src/clr-angular/layout/tabs/tab.spec.ts
Jinnie: Added a few lines in tab-content, where the view is created.

src/clr-angular/layout/tabs/tab-content.ts
Jinnie: Fixed. And a test added. Good catch, Shijir!

src/clr-angular/layout/tabs/tab-content.ts
```diff
@@ -48,4 +64,11 @@ export class ClrTabContent {
   get active() {
     return this.ifActiveService.current === this.id;
   }
+
+  ngOnDestroy(): void {
+    const index = this.tabsService.tabContentViewContainer.indexOf(this.viewRef);
+    if (index > -1) {
+      this.tabsService.tabContentViewContainer.remove(index);
```
Shijir: Nice 👍

src/clr-angular/layout/tabs/tabs.spec.ts
```diff
@@ -215,6 +216,20 @@ describe('Tabs', () => {
...
+      const ifActiveService = context.getClarityProvider(IfActiveService);
+      const originalId = ifActiveService.current;
+      const initialView = tabsService.tabContentViewContainer.get(0);
+      // leave current tab
+      ifActiveService.current = -1;
+      context.detectChanges();
+      // get back to initial tab
+      ifActiveService.current = originalId;
+      context.detectChanges();
+      expect(tabsService.tabContentViewContainer.indexOf(initialView)).toBe(-1);
```
Shijir: Nit: This is testing only on a service level. But we could make it stricter by testing directly on the projected content (whether the content elements are projected only once or not 🤷‍♂ ).

Issue #552, opened 2017-03-02
[UI] Contextual dialog component request


Jinnie:
**Select one ...**  (check one with "x")
```
[ ] bug
[x] feature request
[ ] enhancement
```

### Expected behavior

Hi all,
I'd like to discuss with you a component that we need. I've already had to implement one version or another of it in two active projects, and I know about at least one fellow-project, that is using similar component and would be interested to have it in Clarity. I'm talking about a contextual dialog. I'd usually use it in conjunction with a datagrid, but it's not necessarily the only case. It has to display contextual data in a manner that does not conceal the entire datagrid, so the user is able to quickly click through the rows of the datagrid and browse items.
The contextual data and actions can be various - it's up to the user to implement. But usually it will consist of displaying data and allowing editing operations, bringing additional data from the server, performing data analysis and visualization, retrieving log entries, etc, etc...
Once the user has found a record of interest, he should be able to easily expand this dialog to the entire available workspace.

### Actual behavior

I have previously tried doing this with modals, but they don't satisfy many of the basic requirements:
- They conceal the underlying datagrid.
- They can't be easily extended to full screen, or even if it is done - they look weird.
- If the component is supposed to be embedded in another application, the popups interact very badly with the hosting page.
- Multiple modals bring general problems and are not a good experience, IMO.
I've considered using new routable views for the dialog workspace (which is large by default), but the problems with the access to the datagrid and fullscreen mode remain.

### Reproduction of behavior

The following plunker illustrates the idea:
http://plnkr.co/5ECLdt
I've intentionally not done any designing efforts on it, but sticked only to basic functionality. It's a POC, a demo.
Click on the items. Then keep clicking on them while the dialog is open. Stretch to full screen.

I am open for discussions and suggestions, and ready to bring other interested people and teams into the conversation. Many variations are possible and good to be considered, like compressing the datagrid instead of overlapping it, like allowing manual dragging of the dialog border, etc.
In case you consider the component as viable and useful, I can commit myself to the full implementation, after we've cleared all UX details and have some UI mocks, approved or delivered by you.

Thanks for the attention and best regards!

reddolan:
@Jinnie have you looked at a recent design spec on Grid Row Detail #547? It looks like you could use that pattern to fill this need you have. The detail space will have the ability to load custom views.

Jinnie:
Yes, @reddolan, I was happy to see this spec some time ago, and it will be my preferred choice for smaller amounts of data.
[...]

PR #3338, opened 2019-05-07
Initial detail pane implementation

gnomeontherun:
This is a first PR for the detail pane. Looking mostly for any major issues and API problems, most of what is left is detail work on styles and focus management.

Tests are not yet included, and this won't be the final PR.


src/dev/src/app/datagrid/detail/detail.html
```diff
@@ -0,0 +1,60 @@
...
+
+    <!-- This is here but it will not be enabled -->
+    <clr-dg-row-detail *clrIfExpanded>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin in neque in ante placerat mattis id sed quam.
+      Proin rhoncus lacus et tempor dignissim. Vivamus sem quam, pellentesque aliquet suscipit eget, pellentesque sed arcu.
+      Vivamus in dui lectus. Suspendisse cursus est ac nisl imperdiet viverra.
+    </clr-dg-row-detail>
+  </clr-dg-row>
+
+  <clr-dg-detail>
```
Shijir: Do we need this wrapper component `clr-dg-detail` at all?

src/dev/src/app/datagrid/detail/detail.html
```diff
@@ -0,0 +1,60 @@
...
+<h2>Detail Pane</h2>
+
+<style>
+clr-datagrid ::ng-deep .datagrid-detail-body {
+  padding: 0 1rem;
+}
+</style>
+
+<clr-datagrid [(clrDgSelected)]="selection">
+  <clr-dg-column [clrDgField]="'id'">User ID</clr-dg-column>
```
Shijir: Hmm should there be some kind of input this column will be used for detail pane? I am thinking of questions like what if the user reorders this column, etc?

src/clr-angular/data/datagrid/datagrid-detail.ts
```diff
@@ -0,0 +1,92 @@
...
+import { Subscription } from 'rxjs';
+import { FocusTrapDirective } from '../../utils/focus-trap/focus-trap.directive';
+
+@Component({
+  selector: 'clr-dg-detail',
+  host: {
+    '[class.datagrid-detail-pane-wrapper]': 'true',
+  },
+  template: `
+    <div clrFocusTrap class="datagrid-detail-pane">
```
Shijir: If `clr-dg-detail` is placed inside the datagrid, it will be always present there. And we shouldn't have the focus trap a component that's always present.

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -1,15 +1,38 @@
...
+import { ALL_COLUMN_CHANGES } from '../enums/column-changes.enum';
 
 @Injectable()
 export class ColumnsService {
   columns: BehaviorSubject<ColumnState>[] = [];
+  private _cache: ColumnState[];
+
+  cache() {
+    this._cache = this.columns.map(subject => {
+      const value = subject.value;
```
Shijir: I think we should have a new object like this:
```
const value = { ...subject.value };
```
Otherwise, the cached state objects still have reference to the actual column states. If they change, the cached objects will change.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -56,7 +57,7 @@ import { WrappedColumn } from './wrapped-column';
                <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
             </span>
 
-          <clr-dg-column-separator></clr-dg-column-separator>
+          <clr-dg-column-separator *ngIf="!detailService.isOpen"></clr-dg-column-separator>
```
Shijir: Ah this makes it easy. So users won't be able to change the state of a column with detail pane open.

Shijir:
I haven't reviewed the implementation in details yet. But the API looks good to me. It's clear, straightforward, and nothing surprising.

src/dev/src/app/datagrid/detail/detail.html
gnomeontherun: Good question. This somewhat implicit that the first column is going to be used as the single column during the detail pane being open. It doesn't matter if it was hidden or reordered right now, the first column is displayed. This was part of some discussion earlier, and though we could have an input to define which column to show, the argument was basically that the first column should remain the displayed column because you should put an important column first.

src/clr-angular/data/datagrid/providers/columns.service.ts
gnomeontherun: Yes, should fix up.

src/clr-angular/data/datagrid/datagrid-detail.ts
gnomeontherun: Good call, I will have to put an *ngIf on this element.

src/dev/src/app/datagrid/detail/detail.html
gnomeontherun: It is outside so I can find it to determine if detail pane is being used. Putting it inside a template, we'd have to ask people to name it something like `<ng-template #clrDetailPane>` which seems less explicit.

Shijir: Maybe, we could use `clrIfDetail` directive to figure out whether detail pane is being used or not? 
```
<ng-template clrIfDetail> <clr-dg-detail> ... </clr-dg-detail></ng-template>
```
Or we could create a structural directive 

```
<clr-dg-detail *clrIfDetailOpen> ... </clr-dg-detail>
```
We have a similar concept in D'nd with `*clrIfDragged`. `*clrIfDragged` used to instantiate whatever is used as a custom ghost during dragging, but also we use it to check whether a user gave us their own ghost to use.

gnomeontherun: If we want to capture the event, we'd have to desugar. However, most of the time it could be done as this

```
<clr-dg-detail *clrIfDetail="let detail">
    <clr-dg-detail-header>...</clr-dg-detail-header>
    <clr-dg-detail-body>...</clr-dg-detail-body>
  </clr-dg-detail>```

File: src/clr-angular/data/datagrid/_datagrid.clarity.scss [hunk 6]
```diff
@@ -915,9 +915,146 @@
...
   /*
-        The following classes are used by the renderer when performing internal operations.
-        Using the browser calculations improves rendering performance.
-    */
+   * The following are styles for Detail Pane
+   */
+  .datagrid-outer-wrapper {
+    display: flex;
+    width: 100%;
+    overflow: hidden;
```
hippee-lee: Be careful here. It looks like vertical scrolling is broken. Seems related to the re-introduction of wrappers of wrappers 😟. Setting a fixed height directly on the datagrid (smaller than the calculated height for pagination) doesn’t allow me to scroll and the footer is hidden.

hippee-lee:
This is only a partial review, sorry am trying to fit it in with piecemeal where I can. 

One general comment so far and then one specific comment wrt scrolling.
1.	This is offered as  IMO and FWIW - especially if there have already been discussion around it. I have a UX concern with allowing pagination to navigate while details are shown.  It seems confusing to me to have an items detail showing and be able to navigate to a different page of items and not see what row is showing in the detail Does this mean that the consumer would needs to include relevent column info in the detail panel to keep the user oriented?

src/clr-angular/data/datagrid/datagrid.module.ts
```diff
@@ -47,6 +47,10 @@ import { DatagridRowRenderer } from './render/row-renderer';
 import { WrappedCell } from './wrapped-cell';
 import { WrappedColumn } from './wrapped-column';
 import { WrappedRow } from './wrapped-row';
+import { ClrDatagridDetail, ClrDatagridDetailDirective } from './datagrid-detail';
```
hippee-lee: Not a nit, just a question: is ther a reason to keep component and directive together in the same file? Its the first time I've seen us organize like that.

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -1,60 +1,71 @@
...
-                <div role="columnheader" class="datagrid-column datagrid-select datagrid-fixed-column"
-                     *ngIf="selection.selectionType === SELECTION_TYPE.Single">
-                  <div class="datagrid-column-separator"></div>
-                </div>
-                <!-- header for single row action; only displayType if we have at least one actionable row in datagrid -->
-                <div role="columnheader" class="datagrid-column datagrid-row-actions datagrid-fixed-column"
-                     *ngIf="rowActionService.hasActionableRow">
-                  <div class="datagrid-column-separator"></div>
+<div class="datagrid-outer-wrapper">
+  <div class="datagrid-wrapper" [class.datagrid-detail-open]="detailService.isOpen">
```
hippee-lee: As I mentioned before, this breaks scrolling. Will sync up for ideas on how to handle this but am going to block on it for now.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -74,6 +75,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
     public globalExpandable: ExpandableRowsCount,
     public expand: DatagridIfExpandService,
     private displayMode: DisplayModeService,
+    public detailService: DetailService,
```
hippee-lee: Detail service is also injected into `ClrDatagrid` constructor, doesn't this give us multiple instances of DetailService? e.g 

1. One for ClrDatagrid
2. And one for _every_ ClrDatagridRow

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -915,9 +915,146 @@
...
+
+    &.datagrid-detail-open {
+      width: auto;
+    }
+  }
+
+  .datagrid-row-detail-open {
+    position: relative;
+
+    &:before {
```
hippee-lee: I'm seeing a line on the RHS of the inset triangle when detail is opened. This seperates the row from the detail pane. Is that correct?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -915,9 +915,146 @@
...
+   * The following are styles for Detail Pane
+   */
+  .datagrid-outer-wrapper {
+    display: flex;
+    width: 100%;
+    overflow: hidden;
+  }
+
+  .datagrid-detail-open {
+    flex: 1 0 0;
```
hippee-lee: Playing with the demo I'm seeing the column re-size when the detail pane is open. 

Using `flex: none` here seems to fix it.

src/clr-angular/data/datagrid/datagrid-column.ts
hippee-lee: +1

src/clr-angular/data/datagrid/datagrid.module.ts
gnomeontherun: I've changed this already, haven't pushed it.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
gnomeontherun: Yeah, this needs to be different.

src/clr-angular/data/datagrid/providers/detail.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  open(item: any) {
+    this.cache = item;
+    this.toggleState = true;
+    this._state.next(this.toggleState);
+  }
+
+  toggle(item: any) {
+    if (this.toggleState && this.cache !== item) {
+      this.open(item);
```
Jinnie: Here we directly jump to another element, thus we omit the firing of the `stateChange(false)`, which leads to omitting         `this.columnsService.resetToLastCache();` in `MainRenderer#toggleDetailPane()`, which I guess leads to the following problem:
- open detal
- open other detail
- close detail
- observe columns - only the 1st is present though we don't have details open.
I may be wrong this is the reason, but the problem is reproducible.

src/clr-angular/data/datagrid/datagrid-detail.ts
```diff
@@ -0,0 +1,92 @@
...
+  @HostListener('body:keyup.escape')
+  close() {
+    this.detailService.close();
+  }
+}
+
+@Directive({
+  selector: '[clrIfDetail]',
+})
+export class ClrDatagridDetailDirective implements OnInit, OnDestroy {
```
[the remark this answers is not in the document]
gnomeontherun: We need the structural directive to pass in the data for the selected row, so unless the detail pane is static content it would not work without it. 

I don't see a reuse of this conditional directive anywhere.

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -55,6 +55,15 @@
             <div class="spinner spinner-sm" *ngIf="expand.loading"></div>
           </ng-container>
         </div>
+        <div *ngIf="detailService.isEnabled"
+             class="datagrid-detail-caret datagrid-fixed-column datagrid-cell">
+          <button (click)="detailService.toggle(item)" type="button" class="datagrid-detail-caret-button" [class.is-open]="detailService.isRowOpen(item)">
+            <clr-icon shape="angle-double"
```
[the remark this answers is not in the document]
gnomeontherun: It was decided to not use the icon in the spec due to ambiguity. The arrow is more akin to the expandable row and more clear.

src/clr-angular/data/datagrid/providers/detail.service.ts
gnomeontherun: Thanks, will review this.

src/clr-angular/data/datagrid/datagrid-row.html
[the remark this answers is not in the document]
Jinnie: Have we thought about moving it to the end of the row? Now it's too far away from the action it triggers. Looks a bit lonely there to me.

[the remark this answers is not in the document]
hippee-lee: As prep for pinned columns I added a sticky column for row action buttons as well as the pinned column and cells. 

The thought I had was that these are things that should scroll and should always be visible to the user. MO, I think the detail trigger belongs grouped with the others.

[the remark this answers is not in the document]
gnomeontherun: The pattern is that all actionable items (at least that we can control) are in the left most columns. It also sits in the same space as the expandable row icon.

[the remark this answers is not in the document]
hippee-lee: Agreed. When I get back to pinned columns  all of those items/elements will all get moved into the sticky column container and rows _should_ scroll underneath them.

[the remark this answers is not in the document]
Jinnie: To me, personally, having right faced arrows on the left side of the row is confusing. They don't mean much to me. But as long as UX folks are fine with it, then I'm done commenting it.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -915,9 +915,174 @@
...
+
+    .datagrid-detail-header-title {
+      flex: 1 1 auto;
+    }
+
+    .datagrid-detail-pane-close {
+      flex: 1 1 4rem;
+      padding: 0;
+      cursor: pointer;
+
```
hippee-lee: Does this need something like:
```
    display: flex;
    justify-content: flex-end;
```

to keep the close button on the right side of its container?

src/clr-angular/data/datagrid/datagrid-detail-body.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      beforeEach(function() {
+        context = this.create(ClrDatagridDetailBody, FullTest);
+        context.detectChanges();
+      });
+
+      it('should wire up host bindings', () => {
+        expect(context.clarityElement.className).toContain('datagrid-detail-body');
+      });
+
+      it('projects content into the detail pane ', () => {
```
hippee-lee: nit: isn't this more like a Typescript API test? I see that we have it in both places throughout the library though.

src/clr-angular/data/datagrid/datagrid-detail-header.ts
```diff
@@ -0,0 +1,46 @@
...
+@Component({
+  selector: 'clr-dg-detail-header',
+  host: {
+    '[class.datagrid-detail-header]': 'true',
+  },
+  template: `
+    <div class="datagrid-detail-header-title" tabindex="0">
+      <ng-content></ng-content>
+    </div>
+    <div class="datagrid-detail-pane-close">
```
hippee-lee: I noted this in the css I think, when I run this code the close button is on the rhs of its container.

src/clr-angular/data/datagrid/datagrid-detail.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      let context: TestContext<ClrDatagridDetail, FullTest>;
+      let detailService: DetailService;
+
+      beforeEach(function() {
+        context = this.create(ClrDatagridDetail, FullTest, [DetailService]);
+        detailService = context.getClarityProvider(DetailService);
+        context.detectChanges();
+      });
+
+      it('should close and hide content', () => {
```
hippee-lee: Does this need a test that explicetly uses an ESC event? e.g:
```
  const event = new KeyboardEvent('keyup', {
    key: 'Escape',
  });
  document.body.dispatchEvent(event);
```

src/clr-angular/data/datagrid/datagrid-detail-header.ts
```diff
@@ -0,0 +1,46 @@
...
+})
+export class ClrDatagridDetailHeader {
+  constructor(
+    public detailService: DetailService,
+    public commonStrings: ClrCommonStrings,
+    @Inject(PLATFORM_ID) private platformId: Object,
+    private element: ElementRef
+  ) {}
+
+  ngAfterContentInit() {
```
Shijir: It should after the view is initialized right? With ngAfterViewInit will remove the need for `setTimeout`, I believe. Also, I think we should stick to the Angular way of accessing elements in the template so can we use template reference variable?

src/clr-angular/data/datagrid/datagrid-detail-body.spec.ts
gnomeontherun: I think the grouping of tests is difficult in some cases, and I'd rather not create multiple groups for something this simple. Either way, its the view ensuring we basically have ng-content right?

/shrug

src/clr-angular/data/datagrid/datagrid-detail.spec.ts
gnomeontherun: There is one below.

src/clr-angular/data/datagrid/datagrid-detail.ts
```diff
@@ -0,0 +1,49 @@
...
+  private listener: () => void;
+
+  // For some reason was not getting the HostListener to fire in tests...need to investigate
+  constructor(
+    public detailService: DetailService,
+    private renderer: Renderer2,
+    @Inject(PLATFORM_ID) private platformId: Object
+  ) {
+    this.listener = this.renderer.listen('document', 'keydown', event => {
+      if (event.keyCode === ESC) {
```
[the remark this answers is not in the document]
gnomeontherun: Based on my comment, HostListener wasn't working in tests but this was. I will look into it again though.

src/clr-angular/data/datagrid/datagrid-detail.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+    });
+  }
+
+  ngOnDestroy() {
+    if (this.listener) {
+      this.listener();
+      delete this.listener;
+    }
+    const button: HTMLButtonElement = document.querySelector('.datagrid-detail-caret-button.is-open');
```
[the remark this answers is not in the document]
gnomeontherun: I did this quick and dirty to get it working but yes its not ideal. It would have this problem, and I need to wire up a service to track the trigger button instead.

src/clr-angular/data/datagrid/datagrid-detail.ts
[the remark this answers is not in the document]
gnomeontherun: Also keyCode is deprecated but code is not supported on IE11.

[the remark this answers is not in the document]
Shijir: IIRC `KeyboardEvent.key` is the one that works across Browsers with minimal inconsistencies. But Angular Pseudo Elements is there for removing this kind of quirky browser supports. Also, this not working in the tests might be because of how we implemented this `createKeyboardEvent` utility method. But It's tricky to emulate keyboard events, to say the least:
https://stackoverflow.com/questions/31491549/document-dispatchevent-not-working-in-firefox
https://stackoverflow.com/questions/8942678/keyboardevent-in-chrome-keycode-is-0/12522752#12522752

src/clr-angular/data/datagrid/datagrid-detail.ts
```diff
@@ -0,0 +1,35 @@
...
+    </div>
+    `,
+})
+export class ClrDatagridDetail {
+  constructor(public detailService: DetailService, public commonStrings: ClrCommonStrings) {}
+
+  @HostListener('document:keyup', ['$event'])
+  closeCheck(event: KeyboardEvent) {
+    // @TODO refactor when IE11 is dead
+    if ((event && (event.key && event.key === 'Escape')) || (event.key && event.key === 'Esc')) {
```
Shijir: Why aren't we letting Angular Pseudo Events do this for us? Or `document:keyup.esc` or `document:keyup.escape` didn't work?

src/clr-angular/data/datagrid/datagrid-detail-header.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(
+    public detailService: DetailService,
+    public commonStrings: ClrCommonStrings,
+    @Inject(PLATFORM_ID) private platformId: Object
+  ) {}
+
+  @ViewChild('title') title;
+
+  ngAfterViewInit() {
+    // Set Timeout is required to move focus properly
```
Shijir: This comment can be removed now :)

colinreedmiller:
The row selection is doubly afforded by highlighted row and checkbox. While this in not confusing in the non-detail cases, the detail case suggests that the double affordance of highlighted row + selected checkbox may be less than ideal, particularly in the case of a single selection / detail open: 

Screen Shot 2019-07-01 at 8 29 09 AM

Addressing this is out of scope for the detail implementation, however the value of the double affordance for row selection needs to be investigated.

gnomeontherun:
@colinreedmiller I agree its potentially confusing. I've implemented the original spec behavior, but if there is a potential solution, we should address it before release.

colinreedmiller:
@gnomeontherun - the change that would make the detail view clearer would be to remove the grey highlight that indicates selected row, leaving only the selected checkbox. However, this would open the question of the rest of the selected states when master detail is not used.


Screen Shot 2019-07-01 at 10 52 52 AM

Jinnie:
This PR will also close #552.

gnomeontherun:
Closing this one and will start fresh soon.

Issue #2162, opened 2018-04-05
Implement `clr-` prefix for css classes


gnomeontherun:
We want to implement all public styles with the `clr-` prefix. There should be a shim layer to support both styles.

* [ ] Implement new selectors
* [ ] Support old selectors via shim
* [ ] Look at using ng update to change

Investigate options to either ship both selectors in the same package, or as separate files for each use case.

mathisscott:
As part of this work we also want to change old BEM variables to the new kebab case. So something like `$clr-element--thing` becomes `$clr-element-thing`

kylebradshaw:
We've come up against this issue in our attempts to leverage Clarity in our established angular project. Some of the Clarity component styles have collisions with our current codebase. 
Ex. `.btn`,  `.btn-group`, .dropdown`, `.radio`, `.checkbox` etc
The Clarity css `clr-ui.css` is not namespaced/prefixed consistently.

I asked the question here: https://stackoverflow.com/questions/59274087/best-way-to-prefix-all-clarity-style-rules-for-inclusion-into-established-projec

Looks like in `@clr/ui@latest` the `.clr*` prefix is leveraged in some spots but not others. I see that this was changed to *Someday* - what is a good stop-gap to use? Is there a recommended postcss plugin that we might be able to leverage?

mathisscott:
Hi @kylebradshaw 
After we got the 👍 to work on a foundational library of web-components, prefixing the global CSS became lower in priority. The future for us is Shadow DOM encapsulation to prevent styles clashing as you describe.

One additional problem with munging libraries together, though, is that even once (or if) we prefix all the classes, there will still be collisions as libraries like Bootstrap don't prefix their own styles.

The best path forward is to overwrite what you need to with CSS in your own project.

kylebradshaw:
Thanks for your reply. I started to investigate some `postcss` solutions that might enable me to re-write some of the compiled css after the fact and landed on this lib https://www.npmjs.com/package/postcss-prefixer which seems to give me what I'm looking for, I just need to test it as a viable solution in our own project and will let you know how that works out.

gnomeontherun:
Just FYI, a prefixer is helpful only if you can update all of the HTML at the same time.

mathisscott:
Given the direction `@clr/ui` and `@clr/core` are moving, this is a non-issue.

PR #3339, opened 2019-05-07
Remove datagrid animation hack

Jinnie:
The hack is replaced with Angular animation. The hack was used because
dynamic animations were not possible in old Angular versions.
The cleanup change for the focus trap is needed because of testing
issues with NoopAnimationsModule. Though it only affects tests, we'd
better keep the tests and make minor modification to the code, as with
the current NoopAnimationsModule implementation we can't fix the tests
other ways.


src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -45,18 +48,29 @@ let nbRow: number = 0;
...
     role: 'rowgroup',
   },
   providers: [
     DatagridIfExpandService,
     { provide: IfExpandService, useExisting: DatagridIfExpandService },
     { provide: LoadingListener, useExisting: DatagridIfExpandService },
   ],
+  animations: [
+    trigger('expandAnim', [
+      transition('void => *', []),
```
Shijir: what is this empty transition for?

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -68,6 +82,15 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
   public replaced;
 
+  private expandedState;
+
+  @HostBinding('@expandAnim')
+  get expandAnim() {
+    return { value: this.expandedState, params: { oldHeight: this.oldHeight } };
+  }
+
+  @HostBinding('style.height') activeHeight: string = 'auto';
```
Shijir: This should be included in the host's style just like you did with `'[style.overflow-y]': '"hidden"'`

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -168,6 +194,14 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
   }
 
   ngAfterViewInit() {
+    this.subscriptions.push(
+      this.expand.animate.subscribe(() => {
+        this.activeHeight = 'auto';
+        setTimeout(() => {
```
Shijir: `setTimeout` is no go for us as this will trigger change detection through the whole application. I am also wondering why you needed this.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -213,9 +247,16 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
+    this.expandedState = this.expand.expanded ? 'expanded' : 'collapsed';
   }
 
   public get _view() {
     return this.wrappedInjector.get(WrappedRow, this.vcr).rowView;
   }
+
+  private getCurrentHeight() {
+    const height = this.domAdapter.computedHeight(this.el.nativeElement) || 0;
+    return height + 'px';
```
Shijir: nit: can we just return the number without turning it into a string with the `px`. There are some cases where we need to use just a plain number value of dom properties. If we need to turn it into px, we can do so by`${this.getCurrentHeight()}px`. Once it's returning px value in a string, we have to parse it back in order to get the number version.

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: To skip animation if initialized as expanded.

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: Yes, I think it was more dynamic when I started but now it can be moved to the basic stylesheets.

Jinnie: Then again - wrong - it should be a host binding. This is only initial value.
It gets changed dynamically in onToggle()

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: done

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: As we discussed offline, there is certain lifecycle issue similar to those we observed in other cases (like the tree) when used in conjunction with the clrIfExpanded directive.
I will add a new bug for tracking a possible improvement research, once this may get pushed.
The possible deprecation scenario of the expanded rows will take some time, so as we already have this nearly done, I'd prefer to push it and remove the hack. If someone objects this, please, post a clear no-go for this whole PR, so I can close it and move forward.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: Yes, but this should still move to the host property of the component. The reason is that we rarely use HostBinding for style bindings.

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -213,9 +247,16 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
 
   ngOnInit() {
     this.wrappedInjector = new HostWrapper(WrappedRow, this.vcr);
+    this.oldHeight = this.getCurrentHeight();
```
Shijir: This is one should be moved to `ngAfterViewInit` where the view would be settled.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: I would still press on this to get removed. I think if not the main, it's one of the main improvements we can get from this PR is removal of this setTimeout. Having this setTimeout in this PR, we are still dragging some part of the old hack to this new animation solution.

As a library, Clarity shouldn't be triggering extra change detections throughout the entirety of a user's application.
Is there any way we could time the animation without triggering change detection through the entire app?

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -45,18 +48,29 @@ let nbRow: number = 0;
     '[class.datagrid-row]': 'true',
     '[class.datagrid-selected]': 'selected',
     '[attr.aria-owns]': 'id',
+    '[style.overflow-y]': '"hidden"',
```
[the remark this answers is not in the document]
Jinnie: Note no myself: this should go away, as it was moved to the clr-expandable.

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: The last suggested rework removes the need for setTimeout or detectChanges. It introduces a new intermediary component that will handle the animations. It extracts the complexities of the content-based animations and can be reused.

src/clr-angular/data/datagrid/datagrid-row.ts
Jinnie: gone in rework

src/clr-angular/utils/animations/expandable/expandable.ts
```diff
@@ -0,0 +1,61 @@
...
+  styles: [
+    `
+    :host {
+      display: block;
+      overflow: hidden;
+    }
+  `,
+  ],
+  host: {
+    class: 'expandable',
```
Shijir: The same reason as above would apply here too. So I am kinda questioning the existence of this class.

src/clr-angular/utils/animations/expandable/expandable.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  animations: [
+    trigger('expandAnimation', [
+      transition('void => *', []),
+      transition('* => *', [
```
Shijir: I still don't see why you need `transition('void => *', []),`. 
Also, this transition `'* => *'` is overriding `'void => *'` because an asterisk * or wildcard matches any animation state. So the first transition is not even being used.

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -3,10 +3,17 @@
   Clicking of that wrapper label will equate to clicking on the whole row, which triggers the checkbox to toggle.
 -->
 <label class="datagrid-row-clickable" *ngIf="selection.rowSelectionMode">
-  <ng-template [ngTemplateOutlet]="rowContent"></ng-template>
+  <clr-expandable [clrExpandTrigger]="expanded" *ngIf="globalExpandable.hasExpandableRow">
```
Shijir: Interesting. So placing these animations one level deep inside here instead of the host element helping us to avoid `setTimeout`. I think the animation might have been too eager because it didn't wait for all these templates to be inserted/outlet first. So we should probably try to placing the animation on the div that's wrapping the row content. That might solve this whole animation issue without creating an extra wrapper component.

src/clr-angular/utils/animations/expandable/expandable.ts
Jinnie: Description:
We need the initial transition to avoid animating when we initialize a component as expanded.
Reproduction:
1. Go to https://stackblitz.com/edit/clarity-light-theme-v1-0-bt6xj9?file=src/app/expandable.component.ts
2. See commented transition on line 29. Uncomment it.
3. Observe that the component is rendered without animation.
4. Comment it.
5. Observe that now animation is visible when rendering the component anew.

Expected behavior:
No animation on initialization.

Hope that helps :)

src/clr-angular/utils/animations/expandable/expandable.ts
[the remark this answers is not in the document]
Jinnie: We're not using the `clr-` prefix everywhere (see datagrid) though we are using it some places (see wizard). My point was that we advertise using class-based selectors where needed, not element selectors. Also as this is a building-block type of component, it could be used in other components or even standalone. Then it may get some styling.
I am not huge fan of doing stuff "for the unclear future cases", so I'm okay to remove it.

src/clr-angular/utils/animations/expandable/expandable.ts
```diff
@@ -0,0 +1,61 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component, ElementRef, HostBinding, HostListener, Input } from '@angular/core';
+
+import { animate, style, transition, trigger } from '@angular/animations';
+import { DomAdapter } from '../../dom-adapter/dom-adapter';
+
+@Component({
+  selector: 'clr-expandable',
```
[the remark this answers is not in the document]
Jinnie: Could be done - now it may be easily confused with the if-expandable. Will think for a better name.

src/clr-angular/data/datagrid/datagrid-row.html
Jinnie: The main problem is that there is a very specific time slot when the animation has to start, that's between rendering the content and updating the parent that wraps it.
If it's too early - the animation finishes before we have the content loaded.
If it's too late - then the wrapper gets expanded before the animation takes place and in the best case this leads to flicker, like this:
1) initial->100%, immediate on wrapper fit to content
2) back to initial->100%, as requested by animation
Step 1) may seem questionable and it can be skipped if we keep static initial height, but this will make it impossible to auto resize the content after the animation, for example with window resize, DG column resize, etc. Which we do not want.

The initial idea behind the component was to take advantage of the onChanges lifecycle hook of the component and/or try to add auto-detection of content size changes (without using the trigger), but this has so far been unsuccessful and greatly increases complexity of the solution. 
I find having a separate component for this even in the current state useful, because:
1) It separates the animation logic from the component.
2) Makes it reusable.
3) If we decide to improve on this later, we have to do it on one place only.

src/clr-angular/utils/animations/expandable/expandable.ts
```diff
@@ -0,0 +1,61 @@
...
+  template: `
+    <ng-content></ng-content>
+  `,
+  styles: [
+    `
+    :host {
+      display: block;
+      overflow: hidden;
```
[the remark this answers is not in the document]
Jinnie: Makes sense. Will be done.

src/clr-angular/utils/animations/expandable/expandable.ts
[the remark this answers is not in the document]
coryrylan: The prefixing is still a work in progress. https://github.com/vmware/clarity/issues/2162 The main issue with not prefixing is a generic selector name like `expandable` has a high probability of it getting styles globally that we didnt intend to have. Either way once we turn on shadow DOM encapsulation we will only need to prefix public selectors.

src/clr-angular/data/datagrid/datagrid-row.ts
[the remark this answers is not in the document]
Jinnie: done

src/clr-angular/utils/animations/expandable/expandable.ts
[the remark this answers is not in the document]
Jinnie: Done. Only changed the selector. The class is in animation/ folder anyway that should be enough. Changing class/file names will clutter the review for little gain.

src/clr-angular/utils/animations/expandable/expandable.ts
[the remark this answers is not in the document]
Jinnie: Removed.

Jinnie: Removed.

src/clr-angular/utils/animations/expandable/expandable.ts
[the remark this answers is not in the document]
Jinnie: Done for the `overflow`.
Can't be done for the `display: block`. It's pointless to apply height transformation on non-block contents. The attempt to do it as a transformation step and then animate and then revert is unsuccessful and leads to broken animation.
In the provided example about the tree, we animate on `div.clr-treenode-children`, which gets implicit block display being a `div`. So it is block after all, we're giving just the same level of flexibility as here.

src/clr-angular/utils/animations/expandable/expandable.ts
coryrylan: I have a similar empty transition in the stepper to prevent the animation animation on initialization.

Shijir: I see. I made the wrong assumption that the second transition would override the first one. Also, I wasn't completely sure whether it's a good idea or not to have the logic of escaping the first animation run in its metadata. We have another way of doing it, which is disabling the animation on initialization using: https://angular.io/api/animations/trigger#disabling-animations. But this is definitely not a blocker.

src/clr-angular/data/datagrid/datagrid-row.html
Shijir: If you still find the component for-animation solution the most viable one, I am willing to go forward with it. Regarding the reusability of the animations, we can have it in exportable functions just like we currently do in `utils/animations` which is what is recommended in Angular doc. But my real reluctance to go with a component for animation was that it could create just another layer of complexity related to change detection. That's why I recommended trying out the animations on the exact element that expands/collapse. Another thing we should note is that if use this current solution inside a component that queries its children components with ContentChildren/ContentChild, we would break their direct parent/child relationship by having an extra wrapper between them.

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -3,10 +3,17 @@
   Clicking of that wrapper label will equate to clicking on the whole row, which triggers the checkbox to toggle.
 -->
 <label class="datagrid-row-clickable" *ngIf="selection.rowSelectionMode">
-  <ng-template [ngTemplateOutlet]="rowContent"></ng-template>
+  <clr-expandable-animation [clrExpandTrigger]="expanded" *ngIf="globalExpandable.hasExpandableRow">
```
Shijir: I am almost certain that it's breaking when the content is lazy loaded. The animation would run before the content. So we need to hook it with animate subject somehow.

src/clr-angular/data/datagrid/datagrid-row.html
Jinnie: The animation component is part of the ViewChildren of the ClrDatagridRow component. It's not supplied by the user, but used internally in the component. Whether it's a div or a clr-expandable-animation, it's all the same - no effect on the parent/child relationship with the cells which are ContentChildren.

src/clr-angular/data/datagrid/datagrid-row.html
Jinnie: Yes, you are right, it's not hard to support this case. 
Done!

src/clr-angular/data/datagrid/datagrid-row.html
Jinnie: And second, about exporting it as a function, I don't think the animation definition on its own is some state of the art technology, the tricky parts are in supplying and evaluating the startHeight parameter.
Also, you see that this issue alone already takes about a month to set it all up properly. I think the benefit would be in having a component where the fine details are glued together, and only a minor set of inputs are left for wiring it into a new component.

brannislav:
@Jinnie is it possible to merge the fix in Clarity 1 upcoming versions?

Issue #3300, opened 2019-04-16
Recursive lazy-loaded TreeView is broken when using promises


RoyJacobs:
## Describe the bug

When opening an expandable node in a recursive lazy-loaded TreeView an error is thrown and logged in the console:

```Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'ngIf: true'. Current value: 'ngIf: false'.```

## How to reproduce

1.  Go to https://stackblitz.com/edit/clarity-light-theme-v1-0-vppqhq
2.  Click on node `b` to expand it
3.  See error in console

## Expected behavior

No error in the console.

## Versions

* Clarity: 1.1.2

## Additional notes

If I return an `Observable` from `getTreeChildren` instead of a `Promise`, it works correctly in 1.1.2 (simply remove the `.toPromise()` in `getTreeChildren`. In 1.1.1 both `Observable` and `Promise` responses work correctly.

This issue does not occur in Clarity 1.1.1. I assume this is related to either https://github.com/vmware/clarity/pull/3211 or https://github.com/vmware/clarity/pull/3227 (looking at the changelog of 1.1.2)

**Comment by RoyJacobs** [comment]
I can workaround it by not setting `[clrLazy]` to `true`, but this is a bit unfortunate since it now requires my whole tree to be loaded. For now it's okay, though.

**Comment by Jinnie** [comment]
It's because the flag for the `loading` spinner is set synchronously and then updated asynchronously in the model, which is outside of the Angular lifecycle handling. Typical EHCBIWC error.
I reproduced this with both a subscription and a promise on my environment.
The rxjs/of operator is actually running with a synchronous scheduler by default, that's why it "seems to" pass. If you use a real observable, or change the scheduler to an async one (like this:
```
  getTreeChildren = (item: Item) => {
    return of([
      { text: 'a', isExpandable: false },
      { text: 'b', isExpandable: true }
    ], asyncScheduler)
  }
```
), you will get the error with it, too.

KANekT:
in 3.0.0-next.7 i have this error

use   root$ = new BehaviorSubject<[]>([]);

`    <clr-tree [clrLazy]="true">
      <clr-tree-node
        *clrRecursiveFor="
          let file of root$ | async;
          let i = index;
          getChildren: getChildren
        "
        [clrExpandable]="file.isNode"
        [(clrSelected)]="file.selected"
      >
        {{ file.title }}
      </clr-tree-node>
[...]

PR #3344, opened 2019-05-08
[NG] Recursive lazy-loaded TreeView errors (#3300)

Jinnie:
- "Chocolate" errors when expanding async without preloading.
  It was due to the loading flag initial setting to "true".
- "Property of undefined" error when collapsing async with preloading.
  Children already removed when performing a check.


src/clr-angular/data/tree-view/recursive-for-of.ts
```diff
@@ -1,25 +1,27 @@
...
 }
 
 @Directive({ selector: '[clrRecursiveFor][clrRecursiveForOf]' })
-export class ClrRecursiveForOf<T> implements OnChanges {
+export class ClrRecursiveForOf<T> implements OnChanges, OnDestroy {
   constructor(
     private template: TemplateRef<ClrRecursiveForOfContext<T>>,
-    private featuresService: TreeFeaturesService<T>
+    private featuresService: TreeFeaturesService<T>,
+    private cdr: ChangeDetectorRef
```
hippee-lee: This seems like a `Chocolate error`(™). Did you try to fix this error following the willy wonka/oompa-loompa pattern?

Jinnie: At first it looked like a very good idea, but after I looked in detail how to do it I think it's not applicable in this case. Why?
1) The `recursive-for-of` directive may be a good candidate for the Willy Wonka half of the task, but
2) The Oompa-loompa part has to be implemented in the `recursive-tree-node-model` which, being a model, is a bit outside of the ordinary angular lifecycle. That's why I need to trigger change detection indirectly through a service anyway.
We can discuss it in more detail offline.

src/clr-angular/data/tree-view/tree-features.service.ts
```diff
@@ -15,6 +16,7 @@ export class TreeFeaturesService<T> {
     template: TemplateRef<ClrRecursiveForOfContext<T>>;
     root: RecursiveTreeNodeModel<T>[];
   };
+  childrenFetched: Subject<void> = new Subject();
```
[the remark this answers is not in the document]
Jinnie: It's not because models need to emit their children fetched events from it. As I didn't want very node to subscribe to each of it's children, I decided to extract it to a common service where one observable can serve all updates. As the final goal is to trigger change detection it doesn't really matter which children have been fetched.

src/clr-angular/data/tree-view/recursive-for-of.ts
Jinnie: Okay, we had an offline discussion and did some tests with Matt (thanks!) and here is the summary:
- willy wonka/oompa-loompa pattern is unsuitable because one of the classes involved is a model class, that knows nothing about Angular.
- Tried to substitute the Subject instance in the featuresService with first an `Injector` and second with a `ChangeDetectorRef` but in both cases ended with test nightmares, and the model is used as a model in the tests (and should be) and there are many tests where generated instances are compared with manually created sample instances. If these models include Angular classes, it's very hard to supply the proper dependencies. This is not only hard to test, but tells us that we've abused the class meant to be a model class and in general that our approach is not really correct.
So, the initial implementation seems to be the best choice of all evaluated, only we have to make sure that we'll not leak subscriptions in the case that ngOnChanges is triggered more than once (which is unlikely but not impossible).

src/clr-angular/data/tree-view/tree-features.service.ts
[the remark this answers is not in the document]
hippee-lee: I think its ok.

src/clr-angular/data/tree-view/recursive-for-of.ts
```diff
@@ -28,17 +30,30 @@ export class ClrRecursiveForOf<T> implements OnChanges {
...
   ngOnChanges() {
     let wrapped: RecursiveTreeNodeModel<T>[];
     if (Array.isArray(this.nodes)) {
-      wrapped = this.nodes.map(node => new RecursiveTreeNodeModel(node, null, this.getChildren));
+      wrapped = this.nodes.map(node => new RecursiveTreeNodeModel(node, null, this.getChildren, this.featuresService));
     } else {
-      wrapped = [new RecursiveTreeNodeModel(this.nodes, null, this.getChildren)];
+      wrapped = [new RecursiveTreeNodeModel(this.nodes, null, this.getChildren, this.featuresService)];
     }
+    this.subscriptions.push(
```
hippee-lee: Can we leave a note here that this may be a performance issue if the children inputs change a lot, or add a guard here?

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -79,7 +79,7 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
     if (typeof this.expandable !== 'undefined') {
       return this.expandable;
     }
-    return !!this.expandService.expandable || this._model.children.length > 0;
+    return !!this.expandService.expandable || (this._model.children && this._model.children.length > 0);
```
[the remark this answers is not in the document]
Jinnie: done

src/clr-angular/data/tree-view/recursive-for-of.ts
Jinnie: Guarded!

Issue #3269, opened 2019-04-02
Remove inline font from CSS


coryrylan:
We can help improve the initial render of Clarity based apps by 1 or more seconds by removing the current font set from the inline CSS and use a seperate font file. Ideally this should be done when we update to the new custom font. Things to note:

- Should take advantage of new font file formats like woff2 [more details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization)
- CSS font family should use appropriate font display such as `swap` or `fallback` [more details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization#using_font-display) 
- We should update the Angular schematic to automatically move the font files into the application assets

mathisscott:
@coryrylan 
This needs to wait until after our path on Metropolis is clear. It would be unnecessary re-work otherwise. That said, it definitely needs to be done.

Do you think this can be accomplished without breaking changes?

coryrylan:
I think this would be a breaking change because we would require them to update the `angular.json` to copy the new font files over when they deploy the application. That can be automated with a schematic for new projects. You could make one for upgrading but technically a breaking change since they are having make a minor configuration change in an existing project.

mathisscott:
So 3.0 milestone it is.

gnomeontherun:
Handled this by loading files from clarity city font package.

PR #3353, opened 2019-05-12
Metropolis font metrics `backtick` fix

mattmutt:
Regression found on Unicode character 0060 width allocation and fixed

New metrics revised and a 303-glyph set supplied as a replacement for the inline Clarity CSS. This is a complete Latin encoding.

Font packaging sidenote:
Highly recommend: removing all inlined BASE 64 chunks. Not only does this add to the overload payload size of Clarity, but it is perhaps inefficient because a text format is used to carry not one, but multiple fonts. Newer font encoding methodologies have since been released, such as WOFF2 for modern browsers. I understand Issue #3269 was recently made, hope it gets in soon. As for dealing with FOUT issue, if and when it occurs, I can recommend clean workarounds in the browser.

metroplis-quick-brown-fox
metropolis-fix-proof

Signed-off-by: Matt blacknugget@excite.com


hippee-lee:
This change seems to have had an affect on all of the visual regression suites. 

You can run run visual regression tests locally. Here is how light theme, set one could be run:
```
npm run test:visual:light set1
```

If the failures that are reported are something that should be included with the changes to the font, we need to update the reference images. Reference image updates can be done via: 

```
npm run visual:fix:light set1```

gnomeontherun:
This is only being applied to one font (regular), shouldn’t it be applied to all instances? I’m also curious why there are so many visual diff failures here, something we’ll need to analyze very closely.

gnomeontherun:
@mattmutt Do you have any follow up on this? We would need to see the fix not break the visual diff testing and also be applied to all font weights.

mattmutt:
I have not done any deeper visual difference checking on the font metrics. Can the team report back on their assessment with examples of broken/missing characters? Since there may be variations on the weights, I'll leave it to the visual designers to consider the other font permutations; didn't want to incur other changes.

hippee-lee:
Hey Matt, to see whats going on you can run `npm run test:visual:light setX` for each of the 4 sets. 
I'm sorry I am not a font expert so its difficult so correlate the changes to characters. 
The change as provided seems to have a dramatic effect on all of the visual regression sets. I would have expected them to remain unaffected by this change as I'm assuming it only related to the backtic character.

mattmutt:
@hippee-lee Here are some visual diffs. Because font rasterization engines are different, the subpixel aliasing differences will become apparent. Now I do challenge a human to be able to hand pick these subtle differences when placed side by side. Ultimately one would correctly guess that engines that transform OTF to WOFF have different mathematical libraries. For the changes here, I only modified the metric on the backtick.

visual diff
visual diff

gnomeontherun:
Our images don't use the backtick character in nearly any place, so seeing so many broken image diffs is concerning. One of the images shows more significant differences (the info alert) and I would have to inspect every image to know what else might be affected. The fact its only applied to one of the fonts doesn't really address the problem the whole way, so if people need this fix 

We've discussed this internally, and are concerned about the fallout of the changes. Also, we have plans for addressing this in a more holistic manner soon that would allow teams to opt-into fixes so we don't get caught in a situation where this could accidentally break an application.

Projects can opt into this as a workaround for the backtick issue if the differences don't impact their project, but we're not comfortable shipping this change. Thanks @mattmutt for digging into this, and we'll point people to this if they need to patch it on their side until our next stage is ready.

Issue #2006, opened 2018-02-21
Update docs for datagrid multiselect


shravansofts:
**Select one ...**  (check one with "x")
```
[x] bug
[ ] feature request
[ ] enhancement
```

### Expected behavior
https://vmware.github.io/clarity/documentation/v0.11/datagrid/selection

### Actual behavior
Datagrid multiselect not working due to some module error.

### Reproduction of behavior
https://stackblitz.com/edit/clarity-light-theme-v11-7x1m99?file=app%2Fapp.component.html

### Environment details

* **Angular version:** 4.x.x

* **Clarity version:**

* **OS and version:** 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

hippee-lee:
Hi @shravansofts - I don't seem to be able to replicate a module error with the stackblitz you provided. It runs and I don't see any errors. Can you provide more help so the issue can be replicated? 

Perhaps I am I missing something from the issue report? 

Here is a quick video with the stackblitz you provided. 
![issue-2006](https://user-images.githubusercontent.com/433692/36492727-ac2867ea-16e2-11e8-97cb-3d7aa683d0b0.gif)

youdz:
To add a little bit, I see 2 problems in your stackblitz:

1. You're binding `[(clrDgSelected)]="selected"`, but your property on the component is called `selectedRows`.
2. You are not initializing you `selectedRows` property, which is something we need. It should be: `selectedRows: any[] = [];`

Here is a working Stackblitz: https://stackblitz.com/edit/clarity-light-theme-v11-bn7kva?file=app%2Fapp.component.ts

shravansofts:
@youdz Thanks for the clarification. Documentation is not specifying this initializing aspect.
To avoid this kind of queries we need a plunker link for each component demo. I have raised a separate issue for the same [here](https://github.com/vmware/clarity/issues/2008).

I observed a typo in the documentation of selection as shown below. Please correct it.
![image](https://user-images.githubusercontent.com/12848838/36518605-303fada2-17ad-11e8-8d7c-d26a2c00173a.png)

PR #3366, opened 2019-05-17
WEB Updating docs for datagrid selection

bdryanovski:
Update example for datagrid

fixes: #2006


Jinnie:
Please make sure if this needs to be fixed in other versions too

coryrylan:
Looks like that bug is in the 0.12x version of the docs as well.

bdryanovski:
@Jinnie - I will be happy to document that part as well.  But will need some help on how to rebuild part.
@coryrylan - done

bdryanovski:
I manage to find how to update the cache;

I gonna leave this here for my future self:
```bash
$ cd v0.VERSION/
$ npm install
$ npm run deploy:ci
$ cd ..; cp -a deploy/documentation/v0.VERSION/** v0.VERSION-cache
```

I'm not quite sure why there is no cache directory for `v0.12` & `v0.13`

Jinnie:
Because .12 and .13 used to be updated rather frequently (when 1.0 and 2.0 were not around). And caching .10 and .11 was speeding the build process.
Now 1.0 and 2.0 follow absolutely different deployment process, so we can say that all 0.x are sort of "cached" inside the old process.

bdryanovski:
It seems that this is no longer the valid case so I'm closing it.

Issue #1930, opened 2018-02-02
The sort buttons do not have meaningful description


mathisscott:
URL: http://clarity-ks.surge.sh/datagrid

Reproduced Steps:
1. Go to Datagrid page
2. Check “Name” property of the sort buttons such as sort of Name button. Please refer to screen shot.

Actual Results:
The “Name” property do not have meaningful description.

Expected Results:
The “Name” property should have meaningful description. Such as “Sort of Name” for the button.

Screen shot:

![image.png](https://images.zenhubusercontent.com/582f4cffb9dc3c6220222f57/2c1cf200-d412-41b8-88cf-5ad87828665a)

NOTE: we need to know if this is a Clarity issue or implementation issue.

PR #3384, opened 2019-05-23
[NG] Adding aria-label to clr-dg-column button.

bdryanovski:
Adding accessibility `aria-label` to `clr-dg-column` when sorting is enabled
Adding key `sorfOf` into CommonString
Updating i18n documentation

fixes: #1930


src/website/src/app/documentation/demos/i18n/i18n.demo.ts
```diff
@@ -39,5 +39,6 @@ export class I18nDemo extends ClarityDocComponent {
     { key: 'rowActions', role: 'Datagrid: row actions icon alt text' },
     { key: 'pickColumns', role: 'Datagrid: show and hide columns icon alt text' },
     { key: 'showColumns', role: 'Datagrid: show columns title' },
+    { key: 'sortOf', role: 'Datagrid: sort oof columns title' },
```
Jinnie: typo: oof

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -30,6 +30,7 @@ export class ClrCommonStringsService implements ClrCommonStrings {
   rowActions = 'Available actions';
   pickColumns = 'Show or hide columns';
   showColumns = 'Show Columns';
+  sortOf = 'Sort of';
```
Jinnie: nit: not sure we need the "of". "Sort column" maybe? Any native speakers? :)

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -44,30 +45,41 @@ import { WrappedColumn } from './wrapped-column';
               <ng-content></ng-content>
           </ng-template>
 
-          <button class="datagrid-column-title" *ngIf="sortable" (click)="sort()" type="button">
-              <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
+          <button 
+            class="datagrid-column-title" 
+            [attr.aria-label]="ariaButtonLabel"
```
Jinnie: It would be nice to have a test that verifies this.

bdryanovski: In progress :)

src/clr-angular/utils/i18n/common-strings.service.ts
bdryanovski: 'Sort' to be

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -228,6 +240,10 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, Datag
     }
   }
 
+  public get ariaButtonLabel() {
```
gnomeontherun: Our pattern has really been to just make the commonStrings public (like you have) and then reference the field directly without a getter like this.

gnomeontherun:
I think there is probably a second piece to this, where we look at how to use `aria-sort` to inform the screen reader about what direction the sorting is. https://www.w3.org/WAI/PF/aria/states_and_properties#aria-sort

src/website/src/app/documentation/demos/i18n/i18n.demo.ts
```diff
@@ -39,5 +39,6 @@ export class I18nDemo extends ClarityDocComponent {
     { key: 'rowActions', role: 'Datagrid: row actions icon alt text' },
     { key: 'pickColumns', role: 'Datagrid: show and hide columns icon alt text' },
     { key: 'showColumns', role: 'Datagrid: show columns title' },
+    { key: 'sortOf', role: 'Datagrid: sort of columns title' },
```
gnomeontherun: This is out of sync.

src/clr-angular/data/datagrid/datagrid-column.ts
Jinnie: This is remainder of the 1st attempt to supply text like "Sort on Name". Not needed now.

Jinnie:
The aria-sort is already in place, but it is responsible for announcing things like: "This column has been sorted in ascending order". From the spec you referenced:
> Indicates if items in a table or grid are sorted in ascending or descending order.
It is not responsible for announcing what the button in the header does. It has to be applied to grid or table headers only, it has no notion of the possible buttons inside.
Together with Bozhidar we verified how other solutions like Material are handling it. Material does almost exactly what we do here, other frameworks didn't handle it at all. Bozhidar can say if there were any other options observed.

gnomeontherun:
Ok, got confused in my review, aria-sort is there.

bdryanovski:
The material solution for this type of issues is something like that:

```js
// https://github.com/angular/components/blob/9eeb4b5aa0473c0815a73bfe3c2ed6164d86a3b3/src/material/sort/sort-header-intl.ts

  /** ARIA label for the sorting button. */
  sortButtonLabel = (id: string) => {
    return `Change sorting for ${id}`;
  }
}

// https://github.com/angular/components/blob/9eeb4b5aa0473c0815a73bfe3c2ed6164d86a3b3/src/material/sort/sort-header.html
  <button class="mat-sort-header-button" type="button"
          [attr.disabled]="_isDisabled() || null"
          [attr.aria-label]="_intl.sortButtonLabel(id)"
          (focus)="_setIndicatorHintVisible(true)"
          (blur)="_setIndicatorHintVisible(false)">
    <ng-content></ng-content>
  </button>
```

Which is a nice and simple solution, the tricky part is that they have already the name of the column.

```js
<th mat-sort-header="name">Dessert</th>
<th mat-sort-header="calories">Calories</th>
``` 
I have hard times getting the content from the table column:
  * First, the `ng-template` is not rendered at the time when I'm trying to generate the `ariaButtonLabel` 
  * I'm not always sure that the content for `columnTitle` will be text only. This could lead to some strange results?

On the other topic, `aria-sort` only tell if the column is ASC or DESC, I think that the issue here is that the button for changing this sorting is not clear on what it will do when pressed. On my tests, I got only the button name and that's a button nothing more - I was not clear at anytime that this will change the sorting order. 

The best solution will be to be able to create a string that will say `Sorting for ${columnTitle}` - but I'm not sure how many hacks will require this.

If someone has some clever idea, don't be shy?

src/clr-angular/data/datagrid/datagrid-column.ts
bdryanovski: This was because I try to get the name of the `columnTitle` and `concat` them. If we stay with this solution it's no need for this type `set/get` function. 

Will leave this discussion open, to remind me to rework it later on

src/clr-angular/data/datagrid/datagrid-column.spec.ts
```diff
@@ -317,6 +319,13 @@ export default function(): void {
         expect(context.clarityDirective.sortOrder).toBe(ClrDatagridSortOrder.DESC);
       });
 
+      it('add aria-label to button', function() {
+        context.testComponent.comparator = new TestComparator();
+        context.detectChanges();
+        const title = context.clarityElement.querySelector('.datagrid-column-title');
+        console.log(title, title.attributes, context.testComponent, title.attributes['aria-label'].value);
```
gnomeontherun: Remove the log, and you need an `expect()` for this to be a real test.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -44,30 +45,41 @@ import { WrappedColumn } from './wrapped-column';
...
+              <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
+          </span>
 
           <clr-dg-column-separator></clr-dg-column-separator>
       </div>
   `,
   host: {
     '[class.datagrid-column]': 'true',
     '[attr.aria-sort]': 'ariaSort',
+    '[attr.tabindex]': '-1',
```
gnomeontherun: Why did you add this?

src/clr-angular/utils/i18n/common-strings.interface.ts
```diff
@@ -89,4 +89,8 @@ export abstract class ClrCommonStrings {
    * Datagrid: show columns
    */
   showColumns?: string;
+  /**
+   * Datagrid: sort of columns
+   */
+  sort?: string;
```
gnomeontherun: Can we make this `sortColumn` since Sort is pretty generic and might be used for other things, and the other datagrid ones follow this pattern.

src/clr-angular/data/datagrid/datagrid-column.ts
bdryanovski: This is from one of your comment under #1930 
`Can also set tabindex -1 on the column resize handles`

But maybe I did not get it right?

src/clr-angular/data/datagrid/datagrid-column.spec.ts
bdryanovski: I had stashed this exact change and forgot to pop it out ... that was a lame mistake from my side.. sorry for that.

src/clr-angular/data/datagrid/datagrid-column.ts
gnomeontherun: That isn't where you put it, you put it on the column container. I think we can just remove this line with no change anyways.

src/website/src/sitemap.xml
```diff
@@ -292,6 +292,10 @@
         <loc>https://clarity.design/news</loc>
         <changefreq>daily</changefreq>
     </url>
+    <url>
```
gnomeontherun: I think you'll have to revert this file, it should be fixed elsewhere.

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -30,6 +30,7 @@ export class ClrCommonStringsService implements ClrCommonStrings {
   rowActions = 'Available actions';
   pickColumns = 'Show or hide columns';
   showColumns = 'Show Columns';
+  sortColumn = 'Sort';
```
gnomeontherun: I meant to also have this be `sortColumn = 'Sort column'`;

PR #3387, opened 2019-05-23
change datagrid cell to gridcell role

gnomeontherun:
This is the basic fix, but I'm still not getting it to read the column headers when reading a cell with VoiceOver. I also cannot get it to work in _any_ `role="grid"` component (ours or found anywhere on the web), unless it is an actual table element.


clane:
Verified with Safari/VoiceOver on MacOS but unable to verify with FireFox/NVDA on Windows because https://deploy-preview-3387--vmware-clarity.netlify.com/  is down. Please advise.

chlane:
Can't verify fix on Windows with FireFox/NVDA because https://deploy-preview-3387--vmware-clarity.netlify.com won't load due to the following errors.

ReferenceError: customElements is not defined[Learn More]
index.ts:27:2
Error: Can't resolve all parameters for ApplicationModule: (?).
compiler.js:

PR #3391, opened 2019-05-27
Rebased numeric filter

amellnik:
This is targeting the same topic branch, but contains the changes from #2996 on top of recent master.


gnomeontherun:
This doesn't look right, can you try to rebase onto master and then open a PR to master? We don't want to do any more on the topic branch because its far out.

amellnik:
Ah sure -- I assumed that you wanted to avoid PRs directly to master and would fast-forward this topic branch to match it.  I'll close this one and make another.

PR #3400, opened 2019-05-30
[ng] clear focus on focused items inside dropdown container on focusout

jeeyun:
There was a small bug for the dropdown where the item that was focused retains the focus when you close the dropdown. This was happening only on dropdowns whose items weren't being destroyed (i.e. not using `*clrIfOpen`). This is a fix for that.

### DIFF
--- a/src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
+++ b/src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
@@ -16,6 +16,7 @@ import { FocusService } from '../../../utils/focus/focus.service';
 import { FocusableItem } from '../../../utils/focus/focusable-item/focusable-item';
 import { linkParent, linkVertical } from '../../../utils/focus/focusable-item/linkers';
 import { wrapObservable } from '../../../utils/focus/wrap-observable';
+import { take } from 'rxjs/operators';
 
 @Injectable()
 export class DropdownFocusHandler implements FocusableItem {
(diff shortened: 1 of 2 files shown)


src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
```diff
@@ -113,6 +113,11 @@ export class DropdownFocusHandler implements FocusableItem {
       this.renderer.setAttribute(el, 'tabindex', '-1');
       // When the user moves focus outside of the menu, we close the dropdown
       this.renderer.listen(el, 'focusout', event => {
+        // we clear out any existing focus on the items
+        this.children.subscribe(items => {
```
coryrylan: Should we be concerned about unsubscribing after all the items have blurred?

jeeyun: Hmm so I think we don't have to worry about memory leaks for services (as opposed to components)? Let me know if my assumption is incorrect.

coryrylan: Looking over the code again you are right. Since the subject is not tied to a longstanding event we should not have to worry about the subscriptions. It really depends on the Observable and its source event, ex interval or a browser event we would have to handle subscriptions, this is one of the reasons services have a `ngOnDestroy` lifecycle method.

With the subscribe in the focus out, would the subscribe get called every time more children are added via `addChildren` and not just when `focusout` fires? I guess if you added a console log there does it fire more than just the focusout event?

If it does then maybe something like this?

```typescript
this.children.pipe(take(1)).subscribe(items => 
  items.forEach(item => item.blur())
);
```

jeeyun: So I played around with it a bit. It doesn't get called additionally in any of the combinations I've tried. But happy to keep the `take(1)` in there as the subscription isn't needed after that initial action. However, I did notice that it was getting called twice in certain cases (e.g. clicking on an item inside the dropdown menu as opposed to tabbing away), since I'm listening on `focusout` which bubbles. I'm switching to `blur` event to remove the duplicate.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
```diff
@@ -16,6 +16,7 @@ import { FocusService } from '../../../utils/focus/focus.service';
 import { FocusableItem } from '../../../utils/focus/focusable-item/focusable-item';
 import { linkParent, linkVertical } from '../../../utils/focus/focusable-item/linkers';
 import { wrapObservable } from '../../../utils/focus/wrap-observable';
+import { take } from 'rxjs/internal/operators';
```
hippee-lee: Should this just be importing from `rxjs/operators` ?

jeeyun: Thanks for catching that. Updated.

PR #3405, opened 2019-06-01
chore(core): Update packages for Angular v8.0.0 support

hippee-lee:
As part of upgrading for Clarity v2.0.0 `ng update @angular/cli @angular/core` was run. Changes to config and code files were made to address failing testing and build errors. ted:

1. Angular packages were updated
2. Gemini is commented out: Angular 8 requires node 12 and a gemini dependency doesn't compile on node 12
3. Added node-sass back: for some reason sass (sass-dart) was not getting picked up and the build errors out)
4. Modified test utils to assign functions directly (Object.assign is only assigning properties after the update)
5. Updates jasmine packages
6. Updated misc package files that were behind
7. Application routes were migrated to the new lazy module form
8. Updates ViewChild (s) in website and fixes the website app

`ng update` made changes to the tsconfig files (module: esnext) that changes the way functions are imported and spied on when there is no parental object. I removed those changes but want to call them out for attention during the review since this seems to be something the cli is now defaulting to.

There is one `xit` in the wizard header action tests. It is failing and it looks like the DOM isn't fully cleaned up in between tests. Still need to investigate more here.


coryrylan:
Do we have a plan for replacement of Gemini and the time line for that? Just concerned with disabling the tests and the upcoming changes with CSS Custom properties. Are we planing on migrating to Applitools soon? Maybe we can collaborate applitools upgrade with CSS Custom properties?

src/website/src/app/documentation/demos/wizard/wizard.demo.html
```diff
@@ -701,7 +701,7 @@ <h5>The wizard</h5>
     ...
 &#125;)
 export class WizardJumpToDemo &#123;
-    @ViewChild("ohai", {static: false}) wizard: ClrWizard; // makes wizard accessible through this.wizard!!!
+    @ViewChild("ohai") wizard: ClrWizard; // makes wizard accessible through this.wizard!!!
```
coryrylan: Is the static still needed for the docs? I saw that it was causing the site build to fail because it needed to be escaped.

hippee-lee: I'm removing the docs updates for now. Rather than escaping them I'd rather port them over to the `const` pattern that we now use. 

wrt replacing Gemini - Once I get Angular8 certified and v2 release I'll update the Aplitools branch: https://github.com/vmware/clarity/tree/topic/applitools and get that ready to merge into master so that we can replace gemini with that.

golden/clr-angular.d.ts
```diff
@@ -232,7 +232,7 @@ export declare class ClrDatagrid<T = any> implements AfterContentInit, AfterView
     placeholder: ClrDatagridPlaceholder<T>;
     refresh: EventEmitter<ClrDatagridStateInterface<T>>;
     rowActionService: RowActionService;
-    /** @deprecated */ rowSelectionMode: boolean;
```
coryrylan: Any idea why the @deprecated comments were removed? I guess the `ts-api-guardian` update must of dropped them?

gnomeontherun: It appears so.

PR #3374, opened 2019-05-18
[NG] feat(smart-popover): Adds SmartPopover Utility

hippee-lee:
### Description
As described in the README for this utility directives and services are organized with the following structure:
- services: events, positioning and toggle
- enums: alignment, axis, side, viewport-violation
- directives: smart-anchor, smart-close-button, smart-open-close-button, smart-popover-content
- interfaces: popover-content-offset, smart-position, visibility-coords
- operators(functional): ClrTransform(Type), flipSides, flipAxis, nudgeContent, flipSidesAndNudgeContent, align, alignHorizontal, alignVertical & testVisibility

The datagrid popover components (ClrDgFilter, ClrDatagridActionOverflow & ClrDatagridColumnToggle) have been updated to use the new popover utility.

### TODO
1. Testing for services and directives the make up a SmartPopover complex
2. Deprecate the old popover solution (similar to the Forms strategy)

Signed-off-by: Matt Hippely <mhippely@vmware.com>

PR #3407, opened 2019-06-03
[NG] Update button labels for datagrid-column-toggle

### LINKED ISSUE #3371: Some Datagrid buttons are missing button names
Some of the buttons in the datagrid are just announced as 'button' with a screen reader, and we need to add meaningful names to them.

* [ ] The close button in the clr-dg-column-toggle component lacks a mechanism for providing an accessible name
* [ ] The toggle button inside the clr-dg-column-toggle component lacks a mechanism for providing an accessible name
* [ ] There is no way for the developer using Clarity to provide an accessible name for the datagrid-action-toggle button

bdryanovski:
Accessibility changes for datagrid-column-toggle:
  * adding `aria-controls` to toggleButton 
  * adding `aria-label` to the close button


bdryanovski:
So we had a talk with @Jinnie about the need of having this:

```js
this.menuWrapper.nativeElement.childNodes.forEach(function(node: HTMLElement) {
       node.setAttribute('aria-label', node.innerText);
});
```
So I wanted to start a discussion about it - Do we really need to add additional `aria-label` to items that from outside and could be anything that the user wants to put. I made sure to modify only `.action-item` - also, this is the only content that I could partially predict.

Jinnie:
Not sure we're approaching the right problem here.
@gnomeontherun could you give some more information on what's the actual problem and expected behavior here:
`Screen reader users will not be aware of actions available from the clr-dg-action-overflow component when it appears`
When I tested it, it announces "Available actions, button" and then after you click it you can then browse the options one by one. Do we need to announce something when the popup is open? What?

gnomeontherun:
Ok I've had a chance to review, and I apologize for the lack of detail around this particular one. The issue has two potential solutions.

1) When the trigger opens the action overflow, focus should be moved to the first item
2) Use aria-live to announce the overflow has opened

I just reviewed the smart popover #3374 and this is being refactored as part of that step. However, it doesn't appear to have any logic to move the focus inside. Since this task is really not related to the button naming problem, I'm going to remove it from the original ticket and put it with the focus management ticket. So update the PR accordingly.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -22,6 +22,7 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
       #anchor
       (click)="toggleSwitchPanel()"
       class="btn btn-sm btn-link column-toggle--action"
+      aria-controls="expandable"
```
jeeyun: I believe you should be assigning a dynamic id to the div that gets toggled by this button, and set the `aria-controls` value to that id.

src/clr-angular/data/datagrid/datagrid-action-overflow.spec.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
```
gnomeontherun: nit: I guess this no longer was changed.

src/clr-angular/data/datagrid/datagrid-column-toggle.spec.ts
```diff
@@ -82,6 +83,29 @@ export default function(): void {
         expect(context.clarityElement.querySelectorAll('.column-switch').length).toBe(0);
       });
 
+      it('#anchor should have aria-controls set', function() {
+        expect(
+          context.clarityElement.querySelector('button.column-toggle--action').attributes['aria-controls'].value
+        ).toBeDefined();
```
gnomeontherun: I think this can be defined but still empy, we should check it has a value. Honestly, this seems unnecessary as you are checking more specifically in the next test.

bdryanovski: You are right, this test doesn't prove anything :) - will remove it

bdryanovski:
Could not pass `netlify/vmware-clarity/deploy-preview Expected`

PR #3417, opened 2019-06-05
[ng] Dropdown focus backport

jeeyun:
Backporting #3400, #3450, #3488, #3501 to 1.x.

Note that there's a very old, undocumented version of dropdown markup that hasn't been supported since before v0.10 (i.e. using `<div class="dropdown-menu"></div>` in the Angular component). This backport will NOT provide keyboard navigation for that deprecated markup.


gnomeontherun:
Just to be clear, while it doesn't support the old markup (understandable), does it break its behavior?

jeeyun:
@gnomeontherun - it doesn't break its previous behavior, but preserves it. Before this PR, the old markup did not support keyboard navigation (does not respond to arrow keys, etc). After this PR, the behavior will remain the same for the old markup.

src/clr-angular/popover/dropdown/dropdown-trigger.ts
```diff
@@ -17,16 +18,24 @@ import { ClrDropdown } from './dropdown';
     '[class.dropdown-item]': '!isRootLevelToggle',
     '[class.expandable]': '!isRootLevelToggle',
     '[class.active]': 'active',
+    '[attr.aria-haspopup]': '"menu"',
+    '[attr.aria-expanded]': 'ifOpenService.open',
```
gnomeontherun: We need to apply the same fix as #3450.

jeeyun: Absolutely. I'm also merging fixes from https://github.com/vmware/clarity/pull/3440 into this PR.

PR #3421, opened 2019-06-06
[NG][WEB] Adding aria-label to datagrid-pagination

### LINKED ISSUE #3189: [Accessibility] Clarity datagrid clr-dg-pagination enhancement request to improve screen reader experience
Working to make our application WCAG compliant, I'm running into issues with the datagrid clr-dg-pagination.

Would like to see:
- Ability to set titles on the buttons so they're not just read out "button" with screen reader. Would like something like
```
<clr-dg-pagination #pagination [clrDgPageSize]="10"
    [firstPageButtonTitle]="'First Page'" 
    [prevPageButtonTitle]="'Previous Page'" 
    [nextPageButtonTitle]="'Next Page'" 
    [lastPageButtonTitle]="'Last Page'">
```

- Ability to add a title on the input text field. Something like [currentPageInputTitle]="'Current page number'"

- Add role="navigation" aria-label="Pagination Navigation" in the element wrapping the navigation (need to allow i18n of the aria-label, so would need another new input parameter)

- Setting title on the select using the text in clr-dg-page-size. Example rendered html:
`<clr-dg-page-size [clrPageSizeOptions]="[10,20,50,100]">Users per page</clr-dg-page-size>`
```
<clr-dg-page-size _ngcontent-c22="">
    Users per page
    <div class="clr-select-wrapper">
<!-- add title to select here, can use the value from component text  -->
        <select class="clr-page-size-select ng-pristine ng-valid ng-touched" title="Users per page">
            <option value="0: 10" class="ng-star-inserted">10</option>
            <option value="1: 20" class="ng-star-inserted">20</option>
            <option value="2: 50" class="ng-star-inserted">50</option>
            <option value="3: 100" class="ng-star-inserted">100</option>
        </select>
    </div>
</clr-dg-page-size>
```

Thanks!

bdryanovski:
Adding `aria-label` to the buttons located inside `datagrid-pagination` component

  * Adding `[attr.aria-label]` to buttons to display proper action 
  * Update the documentation for `CommonStringsService`
  * Adding translated examples for `i18n` demo in French
  * Adding Unittest for the new `aria-label` attributes

The correct way of changing the value of the `[attr.aria-label]` is by extending and overwriting the `CommonStringsService`  [Internationalization  Documentation](https://clarity.design/documentation/internationalization)

Close #3189

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-pagination.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-pagination.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -9,16 +9,19 @@ import { ClrDatagridPagination } from './datagrid-pagination';
 import { TestContext } from './helpers.spec';
 import { Page } from './providers/page';
 import { StateDebouncer } from './providers/state-debouncer.provider';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 export default function(): void {
   describe('ClrDatagridPagination component', function() {
     describe('Typescript API', function() {
       let pageService: Page;
       let component: ClrDatagridPagination;
+      let commonStrings: ClrCommonStringsService;
 
       beforeEach(function() {
         pageService = new Page(new StateDebouncer());
-        component = new ClrDatagridPagination(pageService);
+        commonStrings = new ClrCommonStringsService();
+        component = new ClrDatagridPagination(pageService, commonStrings);
         component.ngOnInit(); // For the subscription that will get destroyed.
       });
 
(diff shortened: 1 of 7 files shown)


src/clr-angular/data/datagrid/datagrid-pagination.spec.ts
```diff
@@ -309,6 +312,38 @@ export default function(): void {
...
+        commonStrings = new ClrCommonStringsService();
+
+        context.testComponent.size = 10;
+        context.testComponent.total = 100;
+        context.testComponent.current = 10;
+        context.detectChanges();
+      });
+
+      it('expect buttons to have the correct aria-label from ClrCommonStringsService', function() {
+        const firstPage = context.clarityElement.querySelector('.pagination-first');
```
gnomeontherun: nit - seems unnecessary to define all of these variables for one use.

bdryanovski: I had even more crazy idea - but for readability, I did not commit it.

```js
[
  ['.pagination-first', 'firstPage'],
  ['.pagination-last', 'lastPage'], 
  // ...
].forEach((button) => {
  expect(context.clarityElement.querySelector(button[0]).attributes['aria-label'].value)
  .toBe(commonStrings[button[1]])
})
```

I also don't like creating `alias/shortcut` variables - but the lines become too long and hard to fallow.
```js
// Example
expect(context.clarityElement.querySelector('.pagination-first').attributes['aria-label'].value).toBe(commonStrings.firstPage);
```

Another idea will be to just write it like that...

```js
const { querySelector } = context.clarityElement;
expect(querySelector('.pagination-first').attributes['aria-label'].value).toBe(commonStrings.firstPage);
// ...
```

Issue #3415, opened 2019-06-05
Wizard buttons should be disabled properly


gnomeontherun:
When a wizard has steps that cannot be navigated to, the button/links are still tabbable instead of being fully disabled, which means you can tab to it and select it (but nothing happens because its disabled). The disabled state is just a CSS class right now instead of using the `disabled` attribute. We should update them to use a `disabled` state to disable tab.

PR #3431, opened 2019-06-10
[NG] Add disabled attribute to Wizard > Stepnav > Button

bdryanovski:
Adding `disabled` attribute to `clr-wizard-stepnav-item` - Now when pressing `Tab` the disabled buttons are no longer selectable. 

Before: 
Screen Shot 2019-06-10 at 3 43 27 PM

After:
Screen Shot 2019-06-10 at 3 43 04 PM

There is a visible style change - Not sure if this ok or not? Right now the visual tests are disabled but they will probably fail for that change.

Close: #3415

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -225,17 +225,22 @@ export declare abstract class ClrCommonStrings {
     close?: string;
     collapse?: string;
     current?: string;
+    currentPage?: string;
     danger?: string;
     expand?: string;
+    firstPage?: string;
     hide?: string;
     info?: string;
+    lastPage?: string;
     maxValue?: string;
     minValue?: string;
     more?: string;
     next?: string;
+    nextPage?: string;
     open?: string;
     pickColumns?: string;
     previous?: string;
+    previousPage?: string;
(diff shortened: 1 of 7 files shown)


src/clr-angular/wizard/wizard-stepnav-item.ts
```diff
@@ -66,6 +66,13 @@ export class ClrWizardStepnavItem {
     return this.pageCollection.previousPageIsCompleted(this.page);
   }
 
+  public get _disabledButton(): string | null {
```
mathisscott: Do we need a new public getter here? It's just a ternary that's already referencing the component's API. Would it be better to make it `[attr.disabled]="this.isDisabled ? '' : null"`? Or is there a reason we need it to be a getter?

mathisscott: And we would want a unit test for this in either case...

bdryanovski: No reason at all - I saw the implementation from `clr-wizard-button`.

About the unit test - it's coming :)

PR #3432, opened 2019-06-10
[NG] Add disabled attribute to Wizard > Stepnav > Button

bdryanovski:
Adding `disabled` attribute to `clr-wizard-stepnav-item` - Now when pressing `Tab` the disabled buttons are no longer selectable. 

Before: 
Screen Shot 2019-06-10 at 3 43 27 PM

After:
Screen Shot 2019-06-10 at 3 43 04 PM

There is a visible style change - Not sure if this ok or not? Right now the visual tests are disabled but they will probably fail for that change.

Close: #3415 

PS. 
Sorry for closing/opening the PR - Automation scripts have gone wrong!


gnomeontherun:
@colinreedmiller Do you have any issues with making this change? This is actually desirable from a11y perspective because disabled states are more clear and if I recall @chlane says this is desirable and doesn't have to meet contrast guidelines due to being inactive.

colinreedmiller:
This change is fine - preferable in fact. Ideal would be to give feedback for the disabled state on click with a message e.g. 'Step 2 is disabled until step 1 is complete' --- but doing so is not a blocker to making the visual change.

gnomeontherun:
@bdryanovski Can you update the gemini images on this before merging? We have turned it off for master but I know this changes some. This will just help keep them in sync for later.

bdryanovski:
Sure. Done

PR #3434, opened 2019-06-10
[NG][WEB] Adding aria-label to datagrid-pagination - backport v1

bdryanovski:
Same as #3189 but backport it for V1


src/website/src/app/documentation/demos/i18n/i18n.demo.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component } from '@angular/core';
+import { ClarityDocComponent } from '../clarity-doc';
+
+@Component({
```
gnomeontherun: This is a new page that wasn't originally added into v1, so you'll have to include the whole i18n example and add it to the routing for the docs module here.

bdryanovski: This sounds good but I think that this is not in the scope of the PR - Will create a new one especially for that and introducing this change only - So will remove this from my PR right now.

PR #3436, opened 2019-06-10
fix(website) remove the need to use SSR

gnomeontherun:
This turns off the SSR and Angular universal configuration, as it appears Google Search is smart enough now not to need prerendered pages. SEO was the driver for adding it to begin with, and this fixes the build for the website by not needing SSR instead of waiting for updates to packages. This also removes compodoc, which is having fits, until we have a better strategy. Also removes node-sass because its not needed.


package.json
```diff
@@ -101,7 +101,6 @@
     "@angular/cli": "8.0.0",
     "@angular/compiler-cli": "8.0.0",
     "@angular/language-service": "8.0.0",
-    "@compodoc/compodoc": "1.1.9",
```
coryrylan: Was compodoc doc causing issues? Just curious over its removal

scripts/update-license.js
```diff
@@ -22,7 +22,7 @@ const year = new Date().getFullYear();
 
 files.forEach(file => {
   const filePath = path.join(__dirname, '../', file);
-  if (supported.indexOf(path.extname(filePath)) > -1 && fs.existsSync(filePath)) {
+  if (fs.existsSync(filePath) && supported.indexOf(path.extname(filePath)) > -1) {
```
bdryanovski:
```js
supported.indexOf(path.extname(filePath)) > -1
```
Could be 
```js
supported.includes(path.extname(filePath))
```

`Array.includes` is supported from node 6.x.x - and is more readable by skipping that messy `> -1` check.

.nvmrc
```diff
@@ -1 +1 @@
-v12.3.1
+v10.16.0
```
bdryanovski: Is this version ok for v2?

package.json
gnomeontherun: Compodoc was running into memory issues, and my thinking is if we settle on using it we'll dockerize the build tooling for it separately since its a fairly large package.

.nvmrc
gnomeontherun: This is ok, minimum is `10.9.0` according to Angular's package.json file.

scripts/update-license.js
gnomeontherun: I tend to forget nice things are available because they aren't on the web.

PR #3440, opened 2019-06-11
[ng] add aria-disabled and role separator to dropdown menu

jeeyun:
We already have functionality to skip over disabled items in dropdown when using keyboard navigation. This requires the menu item to have the `disabled` attribute, which the demo code was missing. Angular component documentation already had it so we are good there.

Adding `aria-disabled` allows the disabled item to be announced as such when using screen reader. 

Also adding `role='separator'` to the splitters in the demo / website code. This part happens on the application side so no impact on the component itself.


jeeyun:
@Jinnie thanks for catching that! this should've been caught earlier. updated code + tests.

jeeyun:
This PR was created while we were trying to change up netlify integration. The branch protection seems out of sync, so I'm going to try to open a new PR.

PR #3443, opened 2019-06-11
[NG] Reorder column views by dragging

Shijir:
This PR contains the main logic for reordering the column views. What is missing from this PR is the reordering animation of the columns, which will be added in with the next PR.


src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,122 @@
...
+      if (onReordering) {
+        this._reorderCompleted.next();
+      }
+
+      this.columnsService.requestFirstVisibleChangeCheck();
+      this.columnsService.requestLastVisibleChangeCheck();
+    }
+  }
+
+  public orderAt(index: number): number {
```
bdryanovski:
```js
function orderAt(index: number): number {
  return (this.orders || [])[index] || -1
}
```
Talking to the Duck: If `this.order` is not existing use empty array, try to find an item in this array by index, If I find one, return it to me else return `-1`

* `this.orders` must always be an array of numbers - no need to check it - if there is some way to put something that is not a number this must be handled at that place.
* `this.orders` is initialized at the start with empty array we could remove this `(this.orders || [])`

it seems that this a use case here .. so maybe an initial value may be skipped
```js
private hasDiffWith(newOrders: number[]): boolean {
    if (!this.orders) {
      return true;
    }
    ...
```

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (!this.orders) {
+      return true;
+    }
+    if (!newOrders) {
+      return false;
+    }
+    if (this.orders.length !== newOrders.length) {
+      return true;
+    }
+    return newOrders.filter((newOrder, index) => newOrder !== this.orders[index]).length > 0;
```
bdryanovski: I can't remember where I saw it but for small single typed arrays a faster solution when you want to compare them for equal order is to cast them to strings

```js
return newOrders.toString() !== this.orders.toString()
```
For a huge list, this could start to get slower, but I think is faster than looping over one array and asking to access another one and at the same time comparing them

There is no way to break at the first match - If even one is not at the correct place the result will be the same. but with the `filter` you will have to go to the end, to be absolutely sure

Another option is to use `every`

```js
return newOrders.every((newOrder, index) => newOrder !== this.orders[index]);
```

src/clr-angular/data/datagrid/providers/view-manager.service.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface ViewAccessor {
+  _view: EmbeddedViewRef<void>;
+  order?: number;
+}
+
+// ViewManagerService is a service class that contains utility methods for managing views.
+@Injectable()
+export class ViewManagerService {
+  detachAllViews(containerRef: ViewContainerRef): void {
+    for (let i = containerRef.length; i > 0; i--) {
```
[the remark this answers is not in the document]
Shijir: We are using `let i` to track the number of decrement, but not using it as an index. 
We could have resorted to `while` loop like here:
```ts
while(containerRef.length > 0) {
   containerRef.detach().detectChanges()
}
```
Is this what you are suggesting 👆 ?

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir:
```ts
function orderAt(index: number): number {
  return (this.orders || [])[index] || -1
}
```
this approach would return -1 if the `this.orders[index]` returns 0.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir: I like your second suggestion 👍 . But you meant `.some()` instead of `.every()`, right?

bdryanovski: `Array.some()` is even better. I always forget about it.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
bdryanovski: My Duck did not expect to have `0` in the list :) So I did not test for that case. But you are right.

```js
var orders = [0,2,0,4]
function foo(index) { return (orders || [])[index] || -1}
foo(2)
// => -1
```

src/clr-angular/data/datagrid/providers/view-manager.service.ts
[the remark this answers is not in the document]
bdryanovski: `while loop` or `for loop` is the same for me. Just didn't get the idea why :) but now is clear

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -19,6 +19,37 @@ export class ColumnsService {
...
+  get orderOfFirstVisible(): number {
+    return Math.min(...this.columnStates.filter(state => !state.hidden).map(state => state.order));
+  }
+
+  // First and last visible columns may change in 3 different cases:
+  // -  When columns are reordered
+  // -  When the current first/last gets hidden state
+  // -  When the column at first/last gets removed or a new column added at first/last index
+  private _isFirstVisibleChanged: Subject<void> = new Subject<void>();
+  private _isLastVisibleChanged: Subject<void> = new Subject<void>();
```
gnomeontherun: As I commented elsewhere, I'm not sure we need two different Subjects for this. I'm also not sure if this is the right place, it seems different from what the service core purpose is right now, should this go into the reorder service?

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -178,28 +184,25 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
-            this._scrollableCells.insert(cell._view);
-          });
+          this.viewManager.insertAllViews(this._scrollableCells, this.assignRawOrders(), true);
         }
       }),
       this.expand.animate.subscribe(() => {
         this.expandAnimationTrigger = !this.expandAnimationTrigger;
+      }),
+      // A subscription that listens for view reordering
+      this.columnReorderService.reorderRequested.subscribe(reorderRequest => {
```
gnomeontherun: Need to clean up subscriptions in several places.

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -18,25 +18,30 @@
...
                 <!-- header for datagrid where you can select one row only -->
                 <div role="columnheader" class="datagrid-column datagrid-select datagrid-fixed-column"
                      *ngIf="selection.selectionType === SELECTION_TYPE.Single">
+                  <div class="datagrid-column-wrapper"></div>
                   <div class="datagrid-column-separator"></div>
                 </div>
                 <!-- header for single row action; only displayType if we have at least one actionable row in datagrid -->
                 <div role="columnheader" class="datagrid-column datagrid-row-actions datagrid-fixed-column"
                      *ngIf="rowActionService.hasActionableRow">
+                  <div class="datagrid-column-wrapper"></div>
```
gnomeontherun: You can't reorder these columns can you?

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -228,43 +234,40 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
-        this.columns.forEach(column => {
-          this._projectedCalculationColumns.insert(column._view);
-        });
-        this.rows.forEach(row => {
-          this._calculationRows.insert(row._view);
-        });
-      }
-    });
+    this._subscriptions.push(
+      this.displayMode.view.subscribe(viewChange => {
```
gnomeontherun: @hippee-lee I just wanted to ping you to review this since you set up the original.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -67,14 +86,36 @@ export class ClrDatagridRowDetail<T = any> implements AfterContentInit, OnDestro
...
+        this._detailCells.move(sourceView, reorderRequest.targetIndex);
       })
     );
   }
 
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
   }
+
+  private assignRawOrders(): ClrDatagridCell[] {
```
gnomeontherun: This is duplicated, if you wanted to put it into a shared service or not.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -67,14 +86,36 @@ export class ClrDatagridRowDetail<T = any> implements AfterContentInit, OnDestro
...
     this.subscriptions.push(
       this.expand.replace.subscribe(replaceChange => {
         this.replacedRow = replaceChange;
+      }),
+      this.cells.changes.subscribe(() => {
+        this.viewManager.detachAllViews(this._detailCells);
+        this.viewManager.insertAllViews(this._detailCells, this.assignRawOrders(), true);
+      }),
+      // A subscription that listens for view reordering
+      this.columnReorderService.reorderRequested.subscribe(reorderRequest => {
```
coryrylan: Suggestion, we could break up the subscriptions into smaller methods with descriptive names to help code readability. Something roughly like:

```typescript
this.subscriptions.push(
  this.replaceRowOnReplaceChanges(),
  this.updateViewsOnCellChange(),
  this.updateColumnOrderOnReorderChange()
);
```

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -168,8 +172,10 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
 
   ngAfterContentInit() {
     this.dgCells.changes.subscribe(() => {
-      this.dgCells.forEach(cell => {
-        this._scrollableCells.insert(cell._view);
+      this.dgCells.forEach((cell, index) => {
+        if (this._scrollableCells.indexOf(cell._view) === -1) {
+          this._scrollableCells.insert(cell._view, this.columnReorderService.orderAt(index));
+        }
```
coryrylan: Missing `this.subscriptions.push()` also recommend moving logic into a descriptive method
Something like this maybe?

```typescript
this.subscriptions.push(this.updateScrollableCellsOnCellChanges());
```

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -327,12 +327,23 @@
       vertical-align: top;
       border: none;
 
-      &:last-child {
-        .datagrid-column-separator {
+      .datagrid-column-droppable.draggable-match {
```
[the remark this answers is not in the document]
Shijir: That's a good idea. But `.draggable-match` doesn't come from datagrid, this is from DnD's droppable directive. I will dedicate one commit in this PR for prefixing existing DnD css classes with `clr`.

src/clr-angular/data/datagrid/providers/view-manager.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  insertAllViews(containerRef: ViewContainerRef, viewAccessors: ViewAccessor[], ordered = false): void {
+    if (ordered) {
+      this.setInUniqOrders(viewAccessors).forEach(viewAccessor => containerRef.insert(viewAccessor._view));
+    } else {
+      viewAccessors.forEach(viewAccessor => containerRef.insert(viewAccessor._view));
+    }
+  }
+
+  private setInUniqOrders(viewAccessors: ViewAccessor[]): ViewAccessor[] {
```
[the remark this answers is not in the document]
Shijir: I will rename it `fixOrders`.

src/clr-angular/data/datagrid/providers/view-manager.service.ts
```diff
@@ -0,0 +1,40 @@
...
+// Those components are currently ClrDatagridColumn, ClrDatagridCell, and ClrDatagridRow.
+export interface ViewAccessor {
+  _view: EmbeddedViewRef<void>;
+  order?: number;
+}
+
+// ViewManagerService is a service class that contains utility methods for managing views.
+@Injectable()
+export class ViewManagerService {
+  detachAllViews(containerRef: ViewContainerRef): void {
```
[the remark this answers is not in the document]
Shijir: This is not related to view cleanup issue. This one is for having the detach/insert all view codes in a single service and use it across multiple components to reduce the code repeatition.

src/clr-angular/data/datagrid/render/cell-renderer.ts
```diff
@@ -26,6 +26,11 @@ export class DatagridCellRenderer implements OnDestroy {
     }
 
     this.runAllChanges = ALL_COLUMN_CHANGES;
+
+    // WARNING: This subscription takes place before this.clearWidth method gets called.
```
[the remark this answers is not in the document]
Shijir: I could remove it. It was basically a warning to myself. I don't think it could be useful in the future 🤔

src/clr-angular/data/datagrid/wrapped-cell.spec.ts
```diff
@@ -37,8 +37,5 @@ export default function(): void {
     it('should have a templateRef to the portal', function(this: TestContext) {
       expect(this.wrapper.templateRef).toBeDefined();
     });
-    it('projects content into the template', function(this: TestContext) {
```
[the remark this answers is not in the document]
Shijir: We were testing something that's is within Angular's scope. We don't need to access the inner property value of the EmbeddedView and test it.

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -130,4 +135,20 @@ export class DatagridHeaderRenderer implements OnDestroy {
       this.renderer.removeClass(this.el.nativeElement, HIDDEN_COLUMN_CLASS);
     }
   }
+
+  public isFirstVisible() {
```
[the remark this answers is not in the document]
Shijir: Currently working on them

src/clr-angular/data/datagrid/providers/column-reorder.mock.ts
```diff
@@ -0,0 +1,12 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+
+import { ColumnReorderService } from './column-reorder.service';
+
+@Injectable()
+export class MockColumnReorderService extends ColumnReorderService {}
```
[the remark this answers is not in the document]
Shijir: I can remove it. You're right.

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -89,6 +89,12 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
         this.setupColumns();
         this.columnsSizesStable = false;
         this.stabilizeColumns();
+      }),
+      this.columnsService.checkFirstVisible.subscribe(() => {
+        this.findFirstVisible();
```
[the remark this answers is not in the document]
Shijir: I had to check only lastVisible first initially. But it turned out to be also that I need to check firstVisible. I was also expecting there could be cases that I could check only one of them. But currently, there is no such case. So I will look into how I could merge them.

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -89,6 +89,12 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
         this.setupColumns();
         this.columnsSizesStable = false;
         this.stabilizeColumns();
+      }),
+      this.columnsService.checkFirstVisible.subscribe(() => {
```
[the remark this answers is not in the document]
Shijir: They get pushed into the array of subscriptions.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: They get pushed into the array of subscriptions too :D  Does it look like it's not being cleaned up because it's in the same `.push()`? I could separate them if you prefer.

src/clr-angular/data/datagrid/datagrid.html
Shijir: No, they are just static.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
Shijir: This one is similar to one in `ClrDatagridRow`. `ClrDatagrid` has a method that has the same name as this one but the implementation is different. So basically only `ClrDatagridRow` and `ClrDatagridRowDetail` shares this method. And this doesn't really belong to any of the services we currently have and it's only one method so I decided to leave it in its individual component files.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -491,6 +504,63 @@
...
+
+        &-drag-mode .draggable-match {
+          display: none;
+        }
+
+        &-drag-mode ~ .datagrid-column .draggable-match .datagrid-column-drop-line {
+          margin-left: auto;
+        }
+      }
+      .draggable-ghost .datagrid-column-wrapper {
```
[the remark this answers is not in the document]
Shijir: This `draggable-ghost` also comes from DnD directives as well. I am thinking about prefixing it `clr` to.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
Shijir: I like that. I will work on that.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: Thanks. will fix it.

src/clr-angular/data/datagrid/datagrid-row.ts
gnomeontherun: Given the diff doesn't show the full code block, I didn't see the push above. Up to you if you want to make it more obvious.

src/clr-angular/data/datagrid/providers/view-manager.service.ts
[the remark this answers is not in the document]
Shijir: changed to `prioritizeByOrder`

PR #3455, opened 2019-06-13
[NG] fixes modal body content keyboard scroll support

### LINKED ISSUE #3424: Cannot scroll content in a wizard page with keyboard
## Describe the bug

In the wizard, if the content of the page is larger than the viewport you cannot use a keyboard to scroll the content. This currently only works with the mouse scroller. 

## How to reproduce

Open a wizard with content, and resize the browser so the content is overflowed. Then use keyboard and try to access the content directly with keyboard only to scroll.

## Expected behavior

The container for the wizard page should have `tabindex="0"` so that it can be tabbed into, and then the user can use up/down arrow keys to scroll the content.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This might be a problem in modal content as well, please check and verify its not just wizards.

coryrylan:
Adds a directive to dynamically attach the proper tabindex on instances of `.modal-body`.

closes #3424

### DIFF
--- a/src/clr-angular/modal/modal-body.ts
+++ b/src/clr-angular/modal/modal-body.ts
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive } from '@angular/core';
+
+@Directive({
+  selector: '.modal-body',
+  host: {
+    '[attr.tabindex]': '"0"',
+  },
+})
+export class ClrModalBody {}
--- a/src/clr-angular/modal/modal-body.spec.ts
+++ b/src/clr-angular/modal/modal-body.spec.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component, ViewChild, ElementRef } from '@angular/core';
+import { TestBed, ComponentFixture } from '@angular/core/testing';
+import { ClrModalBody } from './modal-body';
+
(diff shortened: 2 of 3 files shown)


src/clr-angular/modal/modal-body.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive } from '@angular/core';
+
+@Directive({
+  selector: '.modal-body',
```
Shijir: Is it a good idea to tie this directive with a class name? I can see this directive being used almost all content container that has `overflow: auto` or `overflow: scroll`.

coryrylan: I was thinking the same thing. The reason I used a directive was because the `.modal-body` selector is added by the app side so its not in the template of the component which is the scrollable part of the modal. Really the API needs to change where modal body can be guaranteed to exist in the modal template and we have something like `<clr-modal-body>` instead of the user using `modal-body`.

```
<clr-modal [(clrModalOpen)]="closable">
  <h3 class="modal-title">Title</h3>
  <div class="modal-body">
    ...
  </div>
  <div class="modal-footer">
    ...
  </div>
</clr-modal>
```

Shijir: Gotcha. I also asked some questions on the issue this PR is tackling. It seems like this issue is prominent in modal and wizard components so we could just proceed with this current approach.

gnomeontherun: If this is something we need to use more broadly, we could make it more generic (either by adding more selectors or using a generic selector and apply the attribute anywhere its needed).

PR #3458, opened 2019-06-13
[UI] fix line height issue for custom text spacing

### LINKED ISSUE #3367: Test for custom text spacing support - Accessibility
## Describe the bug

The spec states that our application should not have any rendering issues or loss of content when these styles are applied by the user via a plugin or a11y tool. We should test Clarity to make sure the there is no clipping or overflow issues when the styles are applied. Bug example screenshots are in the spec. https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html (updated by Cory)

## Success Criterion 1.4.12 Text Spacing (Level AA)

In content implemented using markup languages that support the following text style properties, no loss of content or functionality occurs by setting all of the following and by changing no other style property:

* Line height (line spacing) to at least 1.5 times the font size;
* Spacing following paragraphs to at least 2 times the font size;
* Letter spacing (tracking) to at least 0.12 times the font size;
* Word spacing to at least 0.16 times the font size.

coryrylan:
Fixes broken links in header layout when custom text spacing is applied to meet a11y spacing requirements. Test and ran gemini locally. Deleted unnecessary scoped header link styles.

closes #3367

### DIFF
--- a/src/clr-angular/layout/main-container/_variables.header.scss
+++ b/src/clr-angular/layout/main-container/_variables.header.scss
@@ -19,6 +19,7 @@ $clr-header-nav-hover-opacity: 1 !default;
 $clr-header-outline-offset: -0.208333rem !default;
 $clr-header-clarity-icons-size: 1rem !default;
 $clr-header-nav-text-horizontal-padding: 1rem !default;
+$clr-header-nav-text-vertical-padding: 0.75rem !default;
 $clr-header-2-bg-color: #485969 !default; // Hard coded because this color is not in the color palette. Its reserved for headers only
 $clr-header-3-bg-color: $clr-color-secondary-action-1000 !default;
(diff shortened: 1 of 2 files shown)


src/clr-angular/layout/main-container/_variables.header.scss
```diff
@@ -19,6 +19,7 @@ $clr-header-nav-hover-opacity: 1 !default;
 $clr-header-outline-offset: -0.208333rem !default;
 $clr-header-clarity-icons-size: 1rem !default;
 $clr-header-nav-text-horizontal-padding: 1rem !default;
+$clr-header-nav-text-vertical-padding: 0.75rem !default;
```
mathisscott: ❤️ for following the naming convention!

PR #3459, opened 2019-06-14
[CHORE] add gemini tests back to build

coryrylan:
This reenables gemini until we move to node 12 or applitools.


.travis.yml
```diff
@@ -1,11 +1,15 @@
 language: node_js
 node_js:
-- '12'
+- '10'
```
coryrylan: Since node 12 gave us issues with Angular and Gemini do we want this at node 10 or 11?

bdryanovski: I'm with version 10.13.0 - and everything is working fine ..so 10 sound good

hippee-lee: I've been using Angular 8 with v11 and it has worked also. No preference here for the node version.

PR #3477, opened 2019-06-18
Modal close button needs an accessible name 

### LINKED ISSUE #3413: Modal close button needs an accessible name
The modal close button is simply just an X SVG, and we need to add the appropriate Close title to it like we have in other places using commonStrings.

bdryanovski:
Adding `aria-label` to the close button on top of the modal header.

Close #3413

### DIFF
--- a/src/clr-angular/modal/modal.html
+++ b/src/clr-angular/modal/modal.html
@@ -1,6 +1,6 @@
 
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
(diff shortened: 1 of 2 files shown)


src/clr-angular/modal/modal.html
```diff
@@ -22,7 +22,7 @@
 
         <div class="modal-content">
           <div class="modal-header">
-            <button type="button" class="close" *ngIf="closable" (click)="close()">
+            <button type="button" [attr.aria-label]="commonStrings.close" class="close" *ngIf="closable" (click)="close()">
               <clr-icon shape="close" [attr.title]="commonStrings.close"></clr-icon>
```
mathisscott: I don't think we need the title on the icon anymore.

bdryanovski: Done ;)

PR #3478, opened 2019-06-18
Backport: Issue 3413

bdryanovski:
Adding `aria-label` to the close button on top of the modal header.

### DIFF
--- a/src/clr-angular/modal/modal.html
+++ b/src/clr-angular/modal/modal.html
@@ -1,6 +1,6 @@
 
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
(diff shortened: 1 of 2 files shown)


src/clr-angular/modal/modal.html
```diff
@@ -22,7 +22,7 @@
 
         <div class="modal-content">
           <div class="modal-header">
-            <button type="button" class="close" *ngIf="closable" (click)="close()">
+            <button type="button" [attr.aria-label]="commonStrings.close" class="close" *ngIf="closable" (click)="close()">
               <clr-icon shape="close" [attr.title]="commonStrings.close"></clr-icon>
```
mathisscott: I don't think we need the title on the icon anymore.

bdryanovski: Done ;)

Issue #3370, opened 2019-05-17
Datagrid focus management 


gnomeontherun:
Inside of the Datagrid, there are a few interactions that the focus is not managed in the best way.

* [x] When using the page count select box, when the selection is changed the focus should move back to the start of the newly displayed content.
* [x] When using the previous/next pagination buttons, the focus should move back to the start of the newly displayed content.
* [x] When closing a filter, the focus is lost to the page when it should return to the filter toggle button.
* [x] When using the select all button, the focus is lost to the page when it should remain on the select all button.
* [x] Screen reader users will not be aware of actions available from the clr-dg-action-overflow component when it appears because focus remains on the trigger but should move into the menu.

gnomeontherun:
@Jinnie I added one more task above from #3371 since it is part of focus management.

Jinnie:
> When using the select all button, the focus is lost to the page when it should remain on the select all button.

Is this about the hideable columns `select all`? It gets disabled after all options are selected. We can't give focus to a disabled form item. @gnomeontherun, please provide more details about this sub-task.

Jinnie:
Update about "Select All" button topic - after an offline discussion it was decided that focus should go to the header and a success message should be announced.

Jinnie:
Short update about the Hideable / Select All subtask:
* [ ] Focus lost on column checkbox toggle
* [ ] Nothing announced after dialog opened

PR #3481, opened 2019-06-19
Datagrid focus management (#3370)

Jinnie:
Done (4/5)
- When using the page count select box, when the selection is changed the focus should move back to the start of the newly displayed content.
- When using the previous/next pagination buttons, the focus should move back to the start of the newly displayed content.
- When closing a filter, the focus is lost to the page when it should return to the filter toggle button.
- Screen reader users will not be aware of actions available from the clr-dg-action-overflow component when it appears because focus remains on the trigger but should move into the menu.

Todo (1/5)
- When using the select all button, the focus is lost to the page when it should remain on the select all button.

Submitting 4/5 for early discussion on the clr-dg-action-overflow focus approach, which was inspired by Cory's work on the date-picker focus management. (Separate commits/per feature available)

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-action-overflow.ts
+++ b/src/clr-angular/data/datagrid/datagrid-action-overflow.ts
@@ -1,36 +1,64 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, EventEmitter, Input, OnDestroy, Output } from '@angular/core';
+import {
+  Component,
+  EventEmitter,
+  Input,
+  OnDestroy,
+  Output,
+  ElementRef,
+  Directive,
+  NgZone,
+  Inject,
+  PLATFORM_ID,
+} from '@angular/core';
 
 import { Point } from '../../popover/common/popover';
 
 import { RowActionService } from './providers/row-action-service';
 import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
+import { isPlatformBrowser } from '@angular/common';
+import { filter, first, last, take, takeLast, debounceTime } from 'rxjs/operators';
+import { Subscription } from 'rxjs';
+
+let clrDgActionId = 0;
 
 @Component({
   selector: 'clr-dg-action-overflow',
   template: `
-        <button (click)="toggle($event)" type="button" class="datagrid-action-toggle" #anchor>
+        <button (click)="toggle($event)" type="button" class="datagrid-action-toggle" #anchor role="button" 
+                    aria-haspopup="true" [attr.aria-controls]="popoverId" [attr.aria-expanded]="open">
             <clr-icon shape="ellipsis-vertical" [attr.title]="commonStrings.rowActions"></clr-icon>
         </button>
         <ng-template [(clrPopoverOld)]="open" [clrPopoverOldAnchor]="anchor" [clrPopoverOldAnchorPoint]="anchorPoint"
                      [clrPopoverOldPopoverPoint]="popoverPoint">
-            <div #menu class="datagrid-action-overflow" (clrOutsideClick)="close($event)" [clrStrict]="true">
+            <div [attr.class]="overflowClassName" (clrOutsideClick)="close($event)" [clrStrict]="true" 
+                    role="menu" [attr.id]="popoverId" [attr.aria-hidden]="!open">
                 <ng-content></ng-content>
             </div>
         </ng-template>
     `,
 })
 export class ClrDatagridActionOverflow implements OnDestroy {
+  public overflowClassName = 'datagrid-action-overflow';
   public anchorPoint: Point = Point.RIGHT_CENTER;
   public popoverPoint: Point = Point.LEFT_CENTER;
 
-  constructor(private rowActionService: RowActionService, public commonStrings: ClrCommonStrings) {
+  constructor(
+    private rowActionService: RowActionService,
+    public commonStrings: ClrCommonStrings,
+    private ref: ElementRef,
+    @Inject(PLATFORM_ID) private platformId: object,
+    private zone: NgZone
+  ) {
     this.rowActionService.register();
+    this.popoverId = 'clr-action-menu' + clrDgActionId++;
   }
+
+  popoverId: string;
 
   ngOnDestroy() {
     this.rowActionService.unregister();
@@ -50,6 +78,21 @@ export class ClrDatagridActionOverflow implements OnDestroy {
     if (boolOpen !== this._open) {
       this._open = boolOpen;
       this.openChanged.emit(boolOpen);
+      if (boolOpen && isPlatformBrowser(this.platformId)) {
+        this.zone.runOutsideAngular(() => {
+          // It emits 2+ onStable events in a synchronous sequence. We need to wait for the last one.
+          // Some of them are probably only in dev mode, but some are caused by collapsing the previous overflow window.
+          this.zone.onStable
+            .asObservable()
+            .pipe(debounceTime(0), take(1))
+            .subscribe(() => {
+              const firstButton = this.ref.nativeElement.querySelector(`.${this.overflowClassName} button`);
+              if (firstButton) {
+                firstButton.focus();
+              }
+            });
+        });
+      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/clr-angular/data/datagrid/datagrid-filter.ts
```diff
@@ -71,6 +75,9 @@ export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDa
     if (boolOpen !== this._open) {
       this._open = boolOpen;
       this.openChanged.emit(boolOpen);
+      if (!boolOpen) {
+        this.anchor.nativeElement.focus();
```
gnomeontherun: It is unlikely but possible this could run in SSR, so we should wrap it in a platform check.

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -50,6 +76,21 @@ export class ClrDatagridActionOverflow implements OnDestroy {
     if (boolOpen !== this._open) {
       this._open = boolOpen;
       this.openChanged.emit(boolOpen);
+      if (boolOpen && isPlatformBrowser(this.platformId)) {
+        this.zone.runOutsideAngular(() => {
+          // It emits 2+ onStable events in a synchronous sequence. We need to wait for the last one.
```
gnomeontherun: Have we tested this running on prod mode?

Jinnie: Thanks for the suggestion. Tested it - works fine.

src/clr-angular/data/datagrid/datagrid-filter.ts
Jinnie: Done!

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -1,37 +1,63 @@
...
 export class ClrDatagridActionOverflow implements OnDestroy {
+  public overflowClassName = 'datagrid-action-overflow';
   public anchorPoint: Point = Point.RIGHT_CENTER;
   public popoverPoint: Point = Point.LEFT_CENTER;
 
-  constructor(private rowActionService: RowActionService, public commonStrings: ClrCommonStrings) {
+  constructor(
+    private rowActionService: RowActionService,
+    public commonStrings: ClrCommonStrings,
+    private ref: ElementRef,
```
Shijir: Would you mind renaming it as `elRef` or `elementRef`? we usually have prefix for element refs just to avoid confusions from ViewRef, ComponentRef, etc...

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -50,6 +76,21 @@ export class ClrDatagridActionOverflow implements OnDestroy {
     if (boolOpen !== this._open) {
       this._open = boolOpen;
       this.openChanged.emit(boolOpen);
+      if (boolOpen && isPlatformBrowser(this.platformId)) {
+        this.zone.runOutsideAngular(() => {
+          // It emits 2+ onStable events in a synchronous sequence. We need to wait for the last one.
+          // Some of them are probably only in dev mode, but some are caused by collapsing the previous overflow window.
+          this.zone.onStable
```
Shijir: This might seem like :trollface: request. But can't we just use setTimeout here? Because it will run outside of angular zone so it won't trigger any change detection, and no other method is dependent on that call, but only Element.focus().

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -50,6 +76,21 @@ export class ClrDatagridActionOverflow implements OnDestroy {
...
+            .asObservable()
+            .pipe(debounceTime(0), take(1))
+            .subscribe(() => {
+              const firstButton = this.ref.nativeElement.querySelector(`.${this.overflowClassName} button`);
```
Shijir: I think we should be a bit stricter to select the buttons here ( `button.action-item` ) since we recommend buttons with `action-item` class?

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Jinnie: :) I don't consider it :trollface: 
I was also first thinking about `setTimeout`, but `zone.onStable` looks like a bit more specific and controlled lifecycle handler that still does the job. While `setTimeout` is a heavyweight weapon. I don't mind using it, but I would like to hear more opinions before I switch to it.
`Why` would we prefer `setTimeout` over `zone.onStable`?

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Jinnie: In a previous PR, on another issue, I was required to remove the specific implementation (button, div, whatever) from a selector and stick to using the class-name, as behavior identifying but not bound to implementation. I started to agree with this.
Why you don't think that a class name is enough?

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Jinnie: Done!

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
gnomeontherun: Ultimately, we don't want to be linked to Zones because long term they might get removed from Angular (Ivy makes this potentially a thing), and as you can see the API for this is...cumbersome.

Shijir: The main drawback I saw here was it currently fires twice and we are picking 2nd one to subscribe. After introducing changes somewhere, what if it starts emits more than that, and the 2nd becomes not what we want to subscribe to any longer? On the contrast, `setTimeout` is called outside of the zone, so we don't need to worry about triggering change detection. Secondly, we know it will run only once and after all microtasks so there's more assurance there. After all, it's also a pretty common pattern to call setTimeout outside of the zone. Plus, I think (it might be subjective 😛 ) it's is concise and non-complex readability-wise.

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Shijir: Because users can project any elements they want there. In our document, we recommend projecting buttons with a class .action-item, but we don't forbid them from using their completely customized actionable items. If they have their own checkboxes and buttons, our current implementation could set the focus on the wrong element. So in order to prevent conflicts with app's implementation there, we should set this default focus behavior only on what we recommend.

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Jinnie: Just for the record, we're taking the last of the synchronous series (that's what the debounceTime(0) stands for), second, third, fifth, no matter.
I agree to the statement that we want to keep away from zone intricacies when their future is uncertain. So I am leaving the "runOutsideAngular" but removing the confusing "onStable".
Thanks for the comments!

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
Jinnie: If they put buttons without the .action-item class, our focus will be the last problem to worry about:
Screen Shot 2019-06-26 at 2 26 34 PM

I added the class to the selector, though I don't think that this will be the exact reason for someone to decide to follow our recommendations.

src/clr-angular/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -50,6 +75,18 @@ export class ClrDatagridActionOverflow implements OnDestroy {
     if (boolOpen !== this._open) {
       this._open = boolOpen;
       this.openChanged.emit(boolOpen);
+      if (boolOpen && isPlatformBrowser(this.platformId)) {
+        this.zone.runOutsideAngular(() => {
+          setTimeout(() => {
+            const firstButton = this.elementRef.nativeElement.querySelector(
+              `.${this.overflowClassName} button.action-item`
```
Shijir: Now we are targeting specific buttons in the component so we don't `.${this.overflowClassName}` part and have the class on the wrapper div element directly.

Jinnie: Yeah, it will be easier to read and maintain. I'll add it to the next PR, that will handle the "select all" issue.

PR #3485, opened 2019-06-20
[NG] improve modal a11y for screen readers

### LINKED ISSUE #3414: Modal focus management
The modal can be improved with better focus management to help keyboard users to access content more logically.

* [ ] When opening a modal, focus should always move to the modal title (also with wizards)
* [ ] When closing a modal, focus should always return to the previous location

This should already be happening with our focus trap, but it was reported to have some problems with wizards so perhaps there is a problem only in how wizards are using it. Need to verify the behavior fully in modals and wizards.

coryrylan:
MacOS Safari + Voice Over do not properly support aria/role attributes for modals. This PR introduces screen reader only text that allows a screen reader only user to know when they are entering and leaving a modal.

closes #3414

### DIFF
--- a/src/clr-angular/modal/modal.html
+++ b/src/clr-angular/modal/modal.html
@@ -15,7 +15,7 @@
          role="dialog"
          [attr.aria-hidden]="!_open"
          [attr.aria-labelledby]="modalId">
-
+      <div class="clr-sr-only">{{commonStrings.modalContentStart}}</div>
       <div class="modal-content-wrapper">
         <!-- only used in wizards -->
         <ng-content select=".modal-nav"></ng-content>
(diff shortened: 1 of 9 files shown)


src/clr-angular/modal/modal.html
```diff
@@ -15,7 +15,7 @@
          role="dialog"
          [attr.aria-hidden]="!_open"
          [attr.aria-labelledby]="modalId">
-
+      <div class="clr-screen-reader-only">{{commonStrings.modalContentStart}}</div>
```
mathisscott: Using `*-sr-only` is fairly standard in place of the longer `*-screen-reader-only`.

Are you open to changing this to `clr-sr-only`?

coryrylan: Can do.

PR #3488, opened 2019-06-20
[ng] more a11y for dropdown menu + fix bug on arrow up/down key on trigger

jeeyun:
This PR is identical to https://github.com/vmware/clarity/pull/3440, which was stuck with netlify branch protection that was out of sync.


RacsoRF:
xd

PR #3489, opened 2019-06-20
[WEBSITE] fix color page

Shijir:
- extracted color data files into separate files
- improved organization in colors demo
- fixed copy functionality
- fixed copy icon on color palettes


src/website/src/app/documentation/demos/color/clipboard-copy.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(@Inject(PLATFORM_ID) private platformId: Object) {}
+
+  private createTextareaEl() {
+    this.textareaEl = document.createElement('textarea');
+
+    // make it off screen
+    this.textareaEl.setAttribute('readonly', '');
+    this.textareaEl.style.position = 'absolute';
+    this.textareaEl.style.left = '-9999px';
```
mathisscott: Is there a better way to force this offscreen? This approach seems heavy handed. What if you just put it at 101% left? Just enough to force it off the screen?

Shijir: ooh I was thinking to use our existing `off-screen` class on that. Thanks for catching it 👍

src/website/src/styles/components.scss
```diff
@@ -677,3 +677,18 @@ _:-ms-fullscreen,
     border: 1px solid #cccccc;
   }
 }
+
+// Utility class for an textarea element that's temporarily
+// created and used for copying content into clipboard
+.offscreen-clipboard-textarea {
```
gnomeontherun: Is this really any different than the off-screen class we already have?

src/website/src/app/documentation/demos/color/color.demo.ts
```diff
@@ -18,131 +21,6 @@ import { ClarityDocComponent } from '../clarity-doc';
 export class ColorDemo extends ClarityDocComponent {
   constructor() {
     super('color');
+    ClarityIcons.add({ copy: EssentialShapes.copy });
```
gnomeontherun: Wait, why do we have to do this? Aren't icons added globally at the root?

src/website/src/styles/components.scss
Shijir: Yeah, that's what I was intending to use. But in that class, we have these two lines which prevents selection
```
overflow: hidden !important;
visibility: hidden !important;
```

src/website/src/app/documentation/demos/color/color.demo.ts
Shijir: no we have `import '@clr/icons';` in main.ts, which will provide only core shapes

PR #3493, opened 2019-06-21
[NG] Create new component clrSpinner

### LINKED ISSUE #3423: Loading events are not announced to the user
## Describe the bug

For cases where there is an async loading behavior, wizard page changes and datagrid expandable rows for example, screen readers will not know that the icon has changed from the expandable option to a loading indicator. 

## How to reproduce

- Using a screen reader, try a wizard with async behavior
- Tab to the Next button and activate it
- Notice that the screen reader does not announce any status messages about the loading of the page. But sighted users see an infinite spinner animation letting them know the page is still loading.

## Expected behavior

Use an empty and polite live region with aria-busy set to false like as seen below.

<p aria-live="polite" aria-busy="false"></p>

Then after the Next button is activated, it should be populated with the loading message as seen below.

<p aria-live="polite" aria-busy="true">
Loading ...
</p>

This should be repeated for long wait times. 

## Recommendation

Use an empty and polite live region with aria-busy set to false like as seen below.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

WCAG 2.1 Guideline: 
4.1.3 Status Messages
(Level AA)
In content implemented using markup languages, status messages can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

bdryanovski:
* create new component clrSpinner
* Update componentlist.json
* Update component for ks-app
* Add new translation `loading`
* Adding async Wizard demo for ks-app
* Add small documentation for clrSpinner
* `datagrid.html` use now `clrSpinner` 

ClrSpinner must: 
  - Add `aria-live` & `aria-busy`
  - Attach `.spinner` class automaticaly

### TODO:
  - Backport it to V1

Close #3423


src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Directive } from '@angular/core';
+
+export const SPINNER_BASE_CLASS = 'spinner';
+
+@Directive({
+  selector: '[clrSpinner]',
```
coryrylan: A possible solution @Shijir mentioned is instead of us adding a new public API we could have the directive match on the existing CSS selector. `selector: '.spinner'`. This would fix anyone already using `.spinner` in existing applications without having to introduce a new public API.

coryrylan: A possible solution @Shijir mentioned is instead of us adding a new public API we could have the directive match on the existing CSS selector. `selector: '.spinner'`. This would fix anyone already using `.spinner` in existing applications without having to introduce a new public API.

coryrylan: I think there are a few options we could take with the spinner API
1. We do nothing and document that the app developer should add the `aria-live` and `aria-busy` when appropriate
2. We use a directive to automatically add the aria labels on all spinners with maybe an input to disable aria label announcements
3. We use a optional directive to add aria labels for convenience but not changing the current default behavior
4. We create a entirely new public API clr-spinner that encapsulates all of the above and determine what are sensible defaults

Questions for the API is do we optimize for the common use case and it may be the wrong default in certain uses? Do we trust the app developer to use the appropriate aria attributes on the application side if we do nothing?

bdryanovski: I'm not sure for the idea of using the CSS selector. This in a way is limiting the developer free choice to use CSS or component in his product. I think that this has to stay optional, also if this is auto-attach to every selector `.spinner` in feature all later changes on this component will be breaking change if for some reason we need to add something new to them.

For the API part: 
 * I think that we must add a documentation section on when to add aria-live/busy - there some cases that we won't be able to handle just by adding a new component.
 * We could create service to announce some best practices in development mode - for example `@Directive({ selector: 'spinner' }) class Spinner { constructor() { /* check to see if element have aria* attributes if not throw console.log */ }` 
 * Maybe a Service to handle all `aria-live` messages won't be a bad idea 

I think that the developer will be nice to know how to improve their code with `aria*` tags. A documentation page will come in handy. Also, think that this could be a great addition to the "design system" guide.

If some of the components could handle this - this ok - still we have to let the developer choose there classes and components, give them the freedom to create more complex components.

gnomeontherun: @bdryanovski can you setup a time for us to talk about this? Sounds like we need to think through this a bit more.

bdryanovski: Sure

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,21 @@
...
+  selector: 'clr-spinner',
+  template: `<ng-content></ng-content>`,
+  host: {
+    class: SPINNER_BASE_CLASS,
+    '[attr.aria-live]': 'liveRegion',
+    '[attr.aria-busy]': 'true',
+  },
+})
+export class ClrSpinner {
+  @Input('liveRegion') liveRegion: string = 'polite';
```
coryrylan: Instead of `string` type you can use a union type of `@Input('liveRegion') liveRegion: 'polite' | 'assertive' = 'polite';

bdryanovski: I like that :)

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -53,7 +53,7 @@
 </div>
 <ng-content select="clr-dg-footer"></ng-content>
 <div class="datagrid-spinner" *ngIf="loading">
-    <div class="spinner spinner-md">Loading...</div>
+    <clr-spinner class="spinner-md">Loading...</clr-spinner>
```
coryrylan: Similar to the progress bar suggestion, let's take the opportunity to encapsulate the classes in the template and provide attribute style hooks.

`<clr-spinner medium inverse inline></clr-spinner>`

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -53,7 +53,7 @@
 </div>
 <ng-content select="clr-dg-footer"></ng-content>
 <div class="datagrid-spinner" *ngIf="loading">
-    <div class="spinner spinner-md">Loading...</div>
+    <clr-spinner medium>Loading...</clr-spinner>
```
gnomeontherun: I think we don't want to have `...` as it gets read out `period period period` or something. Should probably use ellipsis instead `&hellip;`

src/clr-angular/progress/spinner/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export * from './spinner.module';
```
gnomeontherun: We should be exporting the ClrSpinner component too.

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Component, Input } from '@angular/core';
+
+export const SPINNER_BASE_CLASS = 'spinner';
+
+export const SPINNER_INVERSE = 'spinner-inverse';
+export const SPINNER_INLINE = 'spinner-inline';
+
+export const SPINNER_SMALL_SIZE = 'spinner-sm';
+export const SPINNER_MEDIUM_SIZE = 'spinner-md';
+export const SPINNER_LARGE_SIZE = 'spinner-lg';
```
gnomeontherun: I don't think we should export these, I know its used in the tests but its not really saving us much effort IMO. Also it will bleed into the public API.

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,71 @@
...
+  @Input('inline') inline: string;
+  @Input('inverse') inverse: string;
+
+  // Size
+  @Input('small') small: string;
+  @Input('medium') medium: string;
+  /* No need to handle large - default value */
+
+  // Aria Live
+  @Input('assertive') assertive: string;
```
gnomeontherun: All of these inputs should be booleans.

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,71 @@
...
+  @Input('off') off: string;
+  /* No need to handle polite - default value */
+
+  get setAriaLive() {
+    if (this.assertive === '') {
```
gnomeontherun: If you make them booleans than you can just do this to coerce all values into boolean and tests against them. I don't think this will work if someone wants to bind `[assertive]="null"` which Angular supports, because it won't equal an empty string.

`if (!!this.assertive) {`

I'm not sure how much work we want to do to save people from doing things like `assertive="false"` because this is not how attribute booleans work http://w3c.github.io/html/infrastructure.html#boolean-attribute.

src/ks-app/src/app/containers/wizard/wizards.component.html
```diff
@@ -6,115 +6,181 @@
...
+      <clr-wizard-page
+        clrWizardPagePreventDefault="true"
+        (clrWizardPageOnCommit)="onCommit()"
+        (clrWizardPageOnCancel)="doCancel()"
+      >
+        <ng-template clrPageTitle>Form with async validation</ng-template>
+        <!-- mandatory -->
+
+        <clr-spinner *ngIf="loadingFlag">
+          Loading...
```
gnomeontherun: Same thing here with ellipsis, in fact just put something more specific like `Validating your submission`.

src/ks-app/src/app/containers/wizard/wizards.component.html
```diff
@@ -6,115 +6,181 @@
 
 <h1>Wizards</h1>
 <div class="clr-row">
-    <div class="clr-col-12 clr-col-lg-6">
-        <h4>Form validation</h4>
-        <button class="btn btn-primary" (click)="formOpen = !formOpen">Open Wizard</button>
-        <clr-wizard #formWizard
```
gnomeontherun: Just to help with diffs, can you revert these formatting changes?

src/website/src/app/documentation/demos/spinners/spinner-component.html
```diff
@@ -0,0 +1,39 @@
...
+
+<h5>Modify</h5>
+<p>
+    Add modifier like <code class="clr-code">inline</code> or <code class="clr-code">inverse</code>. If you need to make 
+    change on the size you could add <code class="clr-code">small</code> or <code class="clr-code">medium</code>, the 
+    default size is auto set to <code class="clr-code">large</code>.
+</p>
+
+<div class="clr-example squeeze">
+    <clr-spinner inline>Downloading...</clr-spinner>
```
gnomeontherun: Ellipsis.

Also, these should use an ngIf and a button to turn them on, otherwise they'll announce first thing when someone opens the page. This is probably good to mention in the docs here. The real value is the live region, so we need to make it clear _why_ this is used and to always put ngIf on it.

src/website/src/app/documentation/demos/spinners/spinner-component.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component } from '@angular/core';
+
+const EXAMPLE = `
+<clr-spinner>
```
gnomeontherun: Ellipsis again, and put ngIfs on sample code.

src/website/src/app/documentation/demos/spinners/spinner.demo.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 ul {
-  font-weight: 600;
+  // font-weight: 600;
```
gnomeontherun: Intentional?

src/website/src/app/documentation/demos/wizard/wizard-async-completion.demo.html
```diff
@@ -46,9 +46,9 @@
             </clr-alert-item>
         </clr-alert>
 
-        <div class="spinner" *ngIf="loadingFlag">
+        <clr-spinner *ngIf="loadingFlag">
             Loading...
```
gnomeontherun: Ellipsis

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -0,0 +1,71 @@
...
+    '[attr.aria-busy]': 'true',
+  },
+})
+export class ClrSpinner {
+  // Style
+  @Input('inline') inline: string;
+  @Input('inverse') inverse: string;
+
+  // Size
+  @Input('small') small: string;
```
coryrylan: nit: you don't need the inner string in the decorator if the property name is the same,
`@Input() small: boolean`

src/clr-angular/progress/spinner/index.ts
```diff
@@ -0,0 +1,8 @@
...
+export { ClrSpinner } from './spinner';
```
gnomeontherun: export * from './spinner';

src/clr-angular/utils/component/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export { isAttributeSet } from './is-attribute-set';
```
gnomeontherun: We just want to use `export * from './is-attribute-set';` format for these index files.

src/clr-angular/utils/component/is-attribute-set.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    expect(isAttributeSet('')).toBeTruthy();
+    expect(isAttributeSet(true)).toBeTruthy();
+    expect(isAttributeSet('true')).toBeTruthy();
+    expect(isAttributeSet(2)).toBeTruthy();
+    expect(isAttributeSet({ a: 'foo' })).toBeTruthy();
+  });
+
+  it('should return boolean false', () => {
+    expect(isAttributeSet('false')).toBeFalsy();
+    expect(isAttributeSet('random string ')).toBeFalsy();
```
gnomeontherun: This should be true.

src/website/src/app/documentation/demos/spinners/spinner-component.html
```diff
@@ -0,0 +1,43 @@
...
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h3 id="guidelines">Spinner Component</h3>
+<p>
+    Clarity provide component <code class="clr-code">clrSpinner</code>. 
+</p>
+
+<div class="clr-example squeeze">
+    <clr-spinner>Loading data ...</clr-spinner>
```
gnomeontherun: Still `...` here. Should clean them up everywhere.

src/website/src/app/documentation/demos/spinners/spinner-component.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+import { Component } from '@angular/core';
+
+const EXAMPLE = `
+<clr-spinner *ngIf="fetchingUserInformation">Loading data </clr-spinner>
+`;
+
+const EXAMPLE1 = `
+<clr-spinner *ngIf="downloadingFile" inline>Downloading </clr-spinner>
+<span>
```
gnomeontherun: nit: doesn't need to be on separate lines.

src/clr-angular/utils/component/is-attribute-set.ts
```diff
@@ -0,0 +1,25 @@
...
+    return value === '' || value === 'true';
+  }
+  if (typeof value === 'number') {
+    // Must not be 0
+    return value >= 1;
+  }
+  if (typeof value === 'object') {
+    return Object.keys(value).length > 0 && value.constructor === Object;
+  }
+  // Boolean value will be read as it is, everything else is false
```
gnomeontherun: This is actually backwards from HTML attribute semantics. Any value that is provided is technically `true`. Even `inverse="false"` is technically true. We can just do a basic check on truthyness of `value` with the one except of an empty string which evaluates to `false`.

I think we should adhere to technical semantics and this basically becomes the following.

```
if (value === '' || value) { 
  // true 
} else { 
  // false 
};
```

src/clr-angular/utils/component/is-attribute-set.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export function isAttributeSet(value: string | boolean | number | object): boolean {
```
gnomeontherun: In this case, it would only be a `string | boolean`.

bdryanovski: My idea for that function was to be able to have this type of API
```
<component danger> </component>

<component [danger]="displayDanger"></component>
```

I think that the spinner or progress bar will benefit when it comes to changing the styles on the go.

The issue with that was that there is no clean way to use the:

```js
<component danger></component> 
```
For the reason that Angular will return 
```js
this.danger // '' (empty string)
```

You have to make a few checks before you are sure that the value will be `Truthy`
```js
// => value = ''          // true
// => value = 'true'      // true
// => value = 'anything'  // true
// => value = 'false'     // false
```
so value must pass all this checks so we could have the result below.

```js
// All of them must be true
<component danger></component>
<component danger="true"></component>
<component danger="any-string-value-must-be-true"></component>
<component [danger]="shouldDisplayError"></component>

// ALl of them must be false.
<component></component>
<component danger="false"></component>
<component [danger]="iWillBeAlwaysFalse"></component>
```
  
The checks for the `Object | Number` - I Over Engineered it.  

And I think to be more clear I think that the name of the function must be something like that

```js
isBooleanAttributeSet()
```

I don't think that there is a need to check for something else than this

```js
  // for null just return false no need to check anything
  if (value === null) {
    return false;
  }
  if (typeof value === 'string') {
    // Empty string is valid, 'true' as string is also valid
    return value === '' || value.length > 0 && value !== 'false' ;
  }
  // Boolean value will be read as it is, everything else is false
  return typeof value === 'boolean' ? value : false;
```

I'm not sure that even `null` value is possible - but will keep it just to be sure that there will be no corner case. ( I think that `null` is impossible to be returned as a value from Angular.)

src/clr-angular/utils/component/is-attribute-set.ts
bdryanovski: I'm addressing this in my later comment so will close that one.

src/clr-angular/utils/component/is-boolean-attribute-set.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('isAttributeSet', () => {
+  it('should return boolean true', () => {
+    expect(isBooleanAttributeSet('')).toBeTruthy();
+    expect(isBooleanAttributeSet(true)).toBeTruthy();
+    expect(isBooleanAttributeSet('true')).toBeTruthy();
+    expect(isBooleanAttributeSet('random string ')).toBeTruthy();
+  });
+
+  it('should return boolean false', () => {
+    expect(isBooleanAttributeSet('false')).toBeFalsy();
```
gnomeontherun: @coryrylan Do we want this? This is a special case that isn't really to spec.

coryrylan: This should be truthy not falsy if the string is `'false'`, we don't want to diverge from the HTML attribute spec as this will become more important with custom elements in the future. Example here: https://stackblitz.com/edit/angular-shr5nb

src/clr-angular/utils/component/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export * from './is-attribute-set';
```
gnomeontherun: Don't need to export this file. Can delete this whole file.

src/website/src/app/documentation/demos/spinners/spinner.demo.html
```diff
@@ -116,6 +116,8 @@ <h4 id="spinners-versus-progress-bars">Spinners Versus Progress Bars</h4>
                 bar</a> that serves the same use cases as a spinner. Using a spinner or an indeterminate
                 progress bar is a matter of spacing, visual consistency, and the object the user selected
                 to begin the process.</p>
+
+            <clr-spinner-component></clr-spinner-component>
```
gnomeontherun: Let's comment this out for now with a TODO talking about website release.

src/website/src/app/documentation/demos/spinners/spinner-component.html
```diff
@@ -0,0 +1,43 @@
...
+    <clr-spinner>Loading data</clr-spinner>
+</div>
+<clr-code-snippet [clrCode]="example"></clr-code-snippet>
+
+<h5>Modify</h5>
+<p>
+    Add modifier like <code class="clr-code">inline</code> or <code class="clr-code">inverse</code>. If you need to make 
+    change on the size you could add <code class="clr-code">small</code> or <code class="clr-code">medium</code>, the 
+    default size is auto set to <code class="clr-code">large</code>.
+</p>
```
gnomeontherun: Add a table of the modifiers, like in Stepper.

src/website/src/app/documentation/demos/spinners/spinner-component.html
```diff
@@ -0,0 +1,43 @@
+<!--
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h3 id="guidelines">Spinner Component</h3>
+<p>
+    Clarity provide component <code class="clr-code">clrSpinner</code>.
```
gnomeontherun: Clarity provides a component to help with displaying a spinner for loading purposes. This component also helps with accessibility needs to announce loading tasks to a screen reader user.

src/website/src/app/documentation/demos/spinners/spinner-component.html
```diff
@@ -0,0 +1,43 @@
...
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+
+<h3 id="guidelines">Spinner Component</h3>
+<p>
+    Clarity provide component <code class="clr-code">clrSpinner</code>. 
+</p>
+
+<div class="clr-example squeeze">
```
gnomeontherun: Have these behind an ngIf so they don't fire on load, and make a button to toggle loading for each demo.

PR #3495, opened 2019-06-24
[BASE] Kicking off base with first utility file

mathisscott:
• This commit sets up a library where common/shared services can live
• Currently this area is not being pulled in by the clr-angular or any other project but it will be in the future
• The risk on merging this is low because it is private to Clarity and currently low-key
• I have tested this locally in ks-app and dev app and it works fine as it is
• I also tested this in ks-app with AOT and it also worked fine

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ IE11
✔︎ Edge

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -317,6 +317,36 @@
           }
         }
       }
+    },
+    "clr-base": {
+      "projectType": "library",
+      "root": "src/clr-base",
+      "sourceRoot": "src/clr-base/src",
+      "prefix": "clrx",
+      "architect": {
+        "build": {
+          "builder": "@angular-devkit/build-ng-packagr:build",
+          "options": {
+            "tsConfig": "src/clr-base/tsconfig.base.json",
+            "project": "src/clr-base/ng-package.json"
+          }
+        },
+        "test": {
+          "builder": "@angular-devkit/build-angular:karma",
+          "options": {
+            "main": "src/clr-base/src/test.ts",
+            "tsConfig": "src/clr-base/tsconfig.spec.json",
+            "karmaConfig": "src/clr-base/karma.conf.js"
+          }
+        },
+        "lint": {
+          "builder": "@angular-devkit/build-angular:tslint",
+          "options": {
+            "tsConfig": ["src/clr-base/tsconfig.base.json", "src/clr-base/tsconfig.spec.json"],
+            "exclude": ["**/node_modules/**"]
+          }
+        }
+      }
     }
   },
   "schematics": {
--- a/src/clr-base/tsconfig.base.json
+++ b/src/clr-base/tsconfig.base.json
@@ -1,0 +1,25 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "./",
+    "declaration": true,
+    "module": "es2015",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "../../dist/clr-base",
+    // Only necessary because @types/uglify-js can't find types for source-map
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "es5",
+    "lib": ["es2017", "dom"],
+    "paths": {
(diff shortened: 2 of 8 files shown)


angular.json
```diff
@@ -317,6 +317,36 @@
...
+        },
+        "test": {
+          "builder": "@angular-devkit/build-angular:karma",
+          "options": {
+            "main": "src/clr-base/src/test.ts",
+            "tsConfig": "src/clr-base/tsconfig.spec.json",
+            "karmaConfig": "src/clr-base/karma.conf.js"
+          }
+        },
+        "lint": {
```
gnomeontherun: I think this is unnecessary, since our linter runs from clr-angular runs the whole project.

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "paths": {
       "@clr/angular": ["src/clr-angular"],
       "@clr/icons": ["src/clr-icons"],
-      "@clr/icons/shapes/*": ["src/clr-icons/shapes", "src/clr-icons/shapes/*"]
+      "@clr/icons/shapes/*": ["src/clr-icons/shapes", "src/clr-icons/shapes/*"],
+      "clr-base": ["src/clr-base", "src/clr-base/*"]
```
gnomeontherun: I think we want this to be `"@clr/base": ["src/clr-base"]`

src/clr-base/tsconfig.base.json
```diff
@@ -0,0 +1,25 @@
...
+    "strict": true,
+    "target": "es5",
+    "lib": ["es2017", "dom"],
+    "paths": {
+      "@clr/angular": ["../clr-angular"],
+      "@clr/icons": ["../clr-icons"],
+      "@clr/icons/shapes/*": ["../clr-icons/shapes", "../clr-icons/shapes/*"]
+    }
+  },
+  "include": ["custom_typings/**/*.ts", "src/**/*.ts"],
```
gnomeontherun: Not sure what custom_typings directory is about, can we just do this

`"include": ["src/**/*.ts"],`

src/clr-base/tsconfig.base.json
```diff
@@ -0,0 +1,25 @@
...
+  "compilerOptions": {
+    "baseUrl": "./",
+    "declaration": true,
+    "module": "es2015",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "../../dist/clr-base",
+    // Only necessary because @types/uglify-js can't find types for source-map
```
Shijir: I believe JSON files don't allow comments. That's why github is highlighting this as syntax error 😬

angular.json
```diff
@@ -317,6 +317,36 @@
           }
         }
       }
+    },
+    "clr-base": {
+      "projectType": "library",
+      "root": "src/clr-base",
+      "sourceRoot": "src/clr-base/src",
+      "prefix": "clrx",
```
Shijir: is `clrx` correct here 🤔 ?

mathisscott: It's basically `tbd` at the moment. `clrx` is a placeholder until we land on a prefix in the near future.

src/clr-base/tsconfig.base.json
mathisscott: I'll update that. Thanks!

src/clr-base/tsconfig.base.json
mathisscott: I was planning on using a custom typings directory. That's over-optimization at this point tho. I'll remove it. Thanks!

src/clr-base/tsconfig.base.json
gnomeontherun: It isn't allowed normally, but there are some JSON parsers that allow it and consequently this is one case it can be there.

Shijir: EDIT: Jeremy is correct here. Even though Github is highlighting comments here, Typescript does allow comments in JSON format files: https://github.com/microsoft/TypeScript/issues/4987

PR #3496, opened 2019-06-24
[NG] wizard a11y

### LINKED ISSUE #3399: Wizard focus management
The wizard can benefit from a number of focus management improvements to ensure as they navigate around the wizard.

* [x] Opening a wizard should move focus into the wizard (may be just a modal fix)
* [x] Closing a wizard should move focus back to page (may be just a modal fix)
* [x] Should not be able to focus on wizard container, place tabindex=-1 (may be just a modal fix)
* [x] Remove the `tablist` role and treat wizard nav links as regular hyperlinks
* [x] Using the nav links should move focus to the wizard page title on change
* [x] Using next and previous buttons should move focus to the wizard page title
* [x] When there are conditional steps, focus needs to be managed so it doesn't get lost on page changes

coryrylan:
- Refactor TSDoc comments to use latest standard
- Refactor Wizard Component Subscriptions
- Removes Tab Roles (reducing verbosity with screen reader)
- Auto focuses wizard step heading when navigating between steps
- Fix bug where next step event did not fire when clicking next button in wizard

closes #3399


src/clr-angular/wizard/_wizard.clarity.scss
```diff
@@ -48,6 +48,9 @@
       display: flex;
       flex-direction: row;
       width: 100%;
+      // align focus style across modal title
```
coryrylan: This was a weird rendering bug where the line height was not high enough the focus would render incorrectly on the title element.

src/clr-angular/wizard/wizard.ts
```diff
@@ -716,23 +431,60 @@ export class ClrWizard implements OnInit, OnDestroy, AfterContentInit, DoCheck {
...
-  public reset() {
+  public reset(): void {
     this.pageCollection.reset();
     this.onReset.next();
   }
+
+  private listenForNextPageChanges(): Subscription {
+    return this.navService.movedToNextPage.subscribe(() => {
+      this.onMoveNext.emit();
+      this.wizardTitle.nativeElement.focus();
```
coryrylan: TODO: Add missing browser check

coryrylan:
Rebasing and cleaning up commits

PR #3506, opened 2019-06-25
[NG] adding labels to  Datagrid single & expandable column

bdryanovski:
Datagrid (first) row for expandable and single selection don't have a human-readable label.

I'm not sure that the labels are the best one - do we need to add some specific label like `Expandable` and `Single selection` or we could go with `Select` and `Expand` 

### TODO
Backport it for `v1`

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid.html
+++ b/src/clr-angular/data/datagrid/datagrid.html
@@ -26,17 +26,17 @@
                 </div>
                 <!-- header for datagrid where you can select one row only -->
                 <div role="columnheader" class="datagrid-column datagrid-select datagrid-fixed-column"
-                     *ngIf="selection.selectionType === SELECTION_TYPE.Single">
+                     *ngIf="selection.selectionType === SELECTION_TYPE.Single" [attr.aria-label]="clrDgSingleSelectionAriaLabel">
                   <div class="datagrid-column-separator"></div>
                 </div>
                 <!-- header for single row action; only displayType if we have at least one actionable row in datagrid -->
                 <div role="columnheader" class="datagrid-column datagrid-row-actions datagrid-fixed-column"
-                     *ngIf="rowActionService.hasActionableRow">
+                     *ngIf="rowActionService.hasActionableRow" [attr.aria-label]="clrDgSingleActionableAriaLabel">
                   <div class="datagrid-column-separator"></div>
                 </div>
                 <!-- header for carets; only displayType if we have at least one expandable row in datagrid -->
                 <div role="columnheader" class="datagrid-column datagrid-expandable-caret datagrid-fixed-column"
-                     *ngIf="expandableRows.hasExpandableRow">
+                     *ngIf="expandableRows.hasExpandableRow" [attr.aria-label]="clrDetailExpandableAriaLabel">
                   <div class="datagrid-column-separator"></div>
                 </div>
                 <ng-container #projectedDisplayColumns></ng-container>
--- a/src/clr-angular/data/datagrid/datagrid.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -28,11 +28,16 @@ import { Sort } from './providers/sort';
 import { DatagridRenderOrganizer } from './render/render-organizer';
 import { SelectionType } from './enums/selection-type';
 import { HIDDEN_COLUMN_CLASS } from './render/constants';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 @Component({
   template: `
-    <clr-datagrid *ngIf="!destroy"
-                  [(clrDgSelected)]="selected" [clrDgLoading]="loading" (clrDgRefresh)="refresh($event)">
+    <clr-datagrid 
+      *ngIf="!destroy"
+      [(clrDgSelected)]="selected" 
+      [clrDgLoading]="loading" 
+      (clrDgRefresh)="refresh($event)"
+      >
         <clr-dg-column>
             First
             <clr-dg-filter *ngIf="filter" [clrDgFilter]="testFilter"></clr-dg-filter>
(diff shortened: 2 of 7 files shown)


src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -31,12 +31,12 @@
                 </div>
                 <!-- header for single row action; only displayType if we have at least one actionable row in datagrid -->
                 <div role="columnheader" class="datagrid-column datagrid-row-actions datagrid-fixed-column"
-                     *ngIf="rowActionService.hasActionableRow">
+                     *ngIf="rowActionService.hasActionableRow" [attr.aria-label]="commonStrings.select">
```
Shijir: Could you add a corresponding test for this change? Also, is it even a right pattern to place `aria-label` on a non-interactive element such as this `div` element here?

bdryanovski: There is no existing text that could be used for labeling so that's why I added `aria-label`. 

From the spec of `aria-label`, I don't see any issue adding it to any type of element.

@chlane 👷 


About the tests, give me a sec;)

chlane:
Yes aria-label can used on any element that needs an accessible name.

src/clr-angular/data/datagrid/datagrid.html
```diff
@@ -26,17 +26,17 @@
                 </div>
                 <!-- header for datagrid where you can select one row only -->
                 <div role="columnheader" class="datagrid-column datagrid-select datagrid-fixed-column"
-                     *ngIf="selection.selectionType === SELECTION_TYPE.Single">
+                     *ngIf="selection.selectionType === SELECTION_TYPE.Single" [attr.aria-label]="commonStrings.select">
```
gnomeontherun: My concern here is that the text is not descriptive enough, just saying 'Expand' is confusing especially when there is no actual action that can be taken. I think we need new strings for these cases to explain better what is happening.

bdryanovski: I also have the same concern and wanted to add some new `commonString` but was not sure what to name it. Maybe "Expand row" or "Show additional content"

PS.
We are talking about `Expand` not `Select` right?!

gnomeontherun: Both cases really. We probably also should allow apps to provide a custom string per row. I'm doing this in the detail pane like so.

https://github.com/vmware/clarity/pull/3338/files#diff-8b30d14d89ae1c6bc80dd53548d75f7dR157

gnomeontherun: This is still a request to support.

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -143,6 +143,33 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+  get clrDetailSingleSelection(): string {
+    return this._clrDetailSingleSelection || this.commonStrings.selectRow;
+  }
+  set clrDetailSingleSelection(value: string) {
+    this._clrDetailSingleSelection = value;
+  }
+
+  private _clrDetailActionableSelection: string = '';
+  @Input('clrDetailActionableSelection')
+  get clrDetailActionableSelection(): string {
```
gnomeontherun: ```@Input() clrDgActionOverflowAriaLabel = this.commonStrings.datagridActionOverflow;```

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -143,6 +143,33 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+    return this._clrDetailActionableSelection || this.commonStrings.selectRow;
+  }
+  set clrDetailActionableSelection(value: string) {
+    this._clrDetailActionableSelection = value;
+  }
+
+  private _clrDetailExpandableRow: string = '';
+  @Input('clrDetailExpandableRow')
```
gnomeontherun: ```@Input() clrDgExpandableAriaLabel = this.commonStrings.datagridExpandableOverflow;```

This would be used for both expandable rows and master/detail.

gnomeontherun:
You need to revert the package-lock changes.

src/website/src/app/documentation/demos/i18n/i18n.demo.ts
```diff
@@ -55,5 +55,8 @@ export class I18nDemo extends ClarityDocComponent {
     },
     { key: 'allColumnsSelected', role: 'Datagrid: screen reader only confirmation that all columns were selected' },
     { key: 'loading', role: 'Display loading text (Default: Loading)' },
+    { key: 'singleSelectionAriaLabel', role: 'Datagrid: AriaLabel for single selection in header' },
+    { key: 'singleActionableAriaLabel', role: 'Datagrid: AriaLabel for Actionable selection in header' },
```
gnomeontherun: nit: `Datagrid: aria label for header single selection header column` and `Datagrid: aria label for row action header column`

src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -46,6 +46,9 @@ export class ClrCommonStringsService implements ClrCommonStrings {
   signpostToggle = 'Signpost Toggle';
   signpostClose = 'Signpost Close';
   loading = 'Loading';
+  singleSelectionAriaLabel = 'Single selection header';
+  singleActionableAriaLabel = 'Single actionable header';
+  detailExpandableAriaLabel = 'Expandable Row';
```
gnomeontherun: This is probably not the best, it would be better to day 'Toggle more row content', since that is what the button does.

src/clr-angular/data/datagrid/datagrid.spec.ts
```diff
@@ -778,6 +815,16 @@ export default function(): void {
       });
 
       describe('View', function() {
+        /*
+         * For some reason this test is breaking all other tests. 
+         * Not sure why - need to investigate more 
+         * 
+        it('should have aria-label with value `Select`', function() {
```
gnomeontherun: I'm not overly worried about this test, but it would better to do `xit` instead of commenting it out.

bdryanovski: I know, but even when it `xit` i get this error

```
HeadlessChrome 75.0.3770 (Mac OS X 10.14.5) ERROR
  {
    "message": "An error was thrown in afterAll\nUncaught TypeError: Cannot read property 'nativeElement' of undefined thrown",
    "str": "An error was thrown in afterAll\nUncaught TypeError: Cannot read property 'nativeElement' of undefined thrown"
  }

```

Something is very wrong with these lines. I try even writing them again just to be sure that there is no strange symbol.

gnomeontherun:
Can you get rid of package lock changes?

PR #3512, opened 2019-06-25
[WEBSITE] fixes resource links

### LINKED ISSUE #3451: Release notes bug: Figma Design Resources For Clarity lead to 404
## Describe the bug

The figma design resources buttons in the version 2.0.0 release notes link to `https://clarity.design/null`

## How to reproduce

https://clarity.design/news/2.0.0

### LINKED ISSUE #3491: Update Icon Lib 1.0.0 > 2.0.0
## Describe the bug
The Icon Sketch library on the website documentation is still linked with the old 1.0.0 version
https://github.com/vmware/clarity-assets/raw/master/sketch/icons/clarity-library-icons-1.0.0.sketch

## How to reproduce
1.  Go to 'https://clarity.design/documentation/get-started'
2.  Scroll down to 'Using Clarity Design Resources', download Icon lib.

## Expected
https://github.com/vmware/clarity-assets/raw/master/sketch/icons/clarity-library-icons-2.0.0.sketch

## Versions
current website version

Shijir:
- updated url to sketch 2.0.0 (closes #3491)
- fixed figma urls in 2.0.0 release notes (closes #3451)

### DIFF
--- a/src/website/src/releases/2.x/2.0.0.html
+++ b/src/website/src/releases/2.x/2.0.0.html
@@ -51,13 +51,13 @@
             code in use check out the documentation <a routerLink="/documentation/stepper">here</a>.</p>
         <h6 new-component>Figma Design Resources For Clarity (beta)</h6>
         <p>Jumpstart your Clarity design with one of the Figma libraries available here:</p>
-        <a [href]="lightFigma" target="_blank"
+        <a href="https://www.figma.com/file/nRpy4E8bYy9hiDc8iORxyp/Clarity-UI-Library-light-2.0.0?node-id=262%3A7" target="_blank"
            class="btn btn-primary btn-sm" rel="noopener">UI Library - Light ({{sketchVersion}})
         </a>
-        <a [href]="darkFigma" target="_blank"
+        <a href="https://www.figma.com/file/b0FLd4CGw6JIXZ8kwXKaE6/Clarity-UI-Library-dark-2.0.0?node-id=88%3A0" target="_blank"
            class="btn btn-primary btn-sm" rel="noopener">UI Library - Dark ({{sketchVersion}})
         </a>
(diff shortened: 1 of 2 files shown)


src/website/src/releases/2.x/2.0.0.html
```diff
@@ -51,13 +51,13 @@ <h6 new-component>New Stepper Component (<a href="https://github.com/vmware/clar
             code in use check out the documentation <a routerLink="/documentation/stepper">here</a>.</p>
         <h6 new-component>Figma Design Resources For Clarity (beta)</h6>
         <p>Jumpstart your Clarity design with one of the Figma libraries available here:</p>
-        <a [href]="lightFigma" target="_blank"
+        <a href="https://www.figma.com/file/nRpy4E8bYy9hiDc8iORxyp/Clarity-UI-Library-light-2.0.0?node-id=262%3A7" target="_blank"
```
hippee-lee: As I mentioned in the slack channel. I would prefer to keep this bound to `lightFigma` and import the url string from `environment.prod.ts` into the `news.component.ts` file like it is in the `get-started.component.ts` file.

Shijir: But `environment.prod.ts` should only contain dev/prod environment-related data though. That's why I suggested a dedicated file for resource links which would contain this category of data.

Another problem is that `environment.prod.ts` is global so if we update the figma links properties in that file that will take effect in all versions of release notes unless we keep creating new properties for all versions of figma links in those files which we probably wouldn't want that.

gnomeontherun: Another issue is that we should be showing the figma links in the release notes header box, sketch files are listed there now.

I'm ok with hard coding these examples since its part of the release notes and it would lock the version though.

PR #3524, opened 2019-06-26
[ng] fix(datepicker): Add HostListener to focus on the datepicker toggle button

### LINKED ISSUE #3468: Focus is lost when closing date picker with ESC
## Describe the bug

The focus is lost when closing the date picker with the ESC key.

## How to reproduce

- Open a date picker demo on documentation https://clarity.design/documentation/datepicker
- Press ESC while picker is open
- Focus is not clear

## Expected behavior

Focus should return to the button that is used to activate the date picker on close.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

hippee-lee:
This is a backport of the fix for v1 into the v1 codebase. 

I had to modify the fix because of the dual templates that support both old and new forms in v1. 
I don't really like this fix but it avoids modifying other parts of v1 (AbstractPopover also has an escape listener and that seems to interfere with the HostListener in v1). Also ran into an issue where I couldn't `@ViewChild` the buttons because they are in templates. 

- closes #3468

### DIFF
--- a/src/clr-angular/forms/datepicker/date-container.spec.ts
+++ b/src/clr-angular/forms/datepicker/date-container.spec.ts
@@ -115,6 +115,20 @@ export default function() {
         context.detectChanges();
       });
 
+      it('returns focus to the toggle button when closed with the escape key', () => {
+        const actionButton: HTMLButtonElement = context.clarityElement.querySelector('.clr-input-group-icon-action');
+        const actionButtonSpy = spyOn(actionButton, 'focus');
+        expect(actionButtonSpy.calls.count()).toBe(0);
+        actionButton.click();
+        context.detectChanges();
+        const event = new KeyboardEvent('keyup', {
+          key: 'Escape',
+        });
+        document.body.dispatchEvent(event);
+        context.detectChanges();
+        expect(actionButtonSpy.calls.count()).toBe(1);
+      });
+
       it('applies the clr-form-control class', () => {
         expect(context.clarityElement.className).toContain('clr-form-control');
       });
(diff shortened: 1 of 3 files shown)


src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -115,6 +115,20 @@ export default function() {
...
+        actionButton.click();
+        context.detectChanges();
+        const event = new KeyboardEvent('keyup', {
+          key: 'Escape',
+        });
+        document.body.dispatchEvent(event);
+        context.detectChanges();
+        expect(actionButtonSpy.calls.count()).toBe(1);
+      });
+
```
jeeyun: Shouldn't the test assert that the active element is back on the toggle button after pressing esc?

hippee-lee: I'm testing the the toggle button focus method was called after an escape event. It seemed the best way test that the toggle button had received focus.  

Is the suggestion the I should also add
```expect(document.activeElement).isEqual(actionButton);```?

jeeyun: Ah, I overlooked the part that spies on the focus method. Thank you for clarifying. 

Yes, my suggestion would have been to  add something like `expect(document.activeElement).isEqual(actionButton);`. But I understand that this is a backport; I wouldn't block this PR on it.

PR #3532, opened 2019-06-28
[NG] fix(signpost): Adds a11y to the signpost

### LINKED ISSUE #3523: Screen readers don't read signpost content when opened
## Describe the bug

Screen readers don't read signpost content when opened nor do they know what the trigger buttons will do.

## How to reproduce

- Using a screen reader activate the Signpost link under Signpost Demos
- Observe that you hear nothing

## Expected behavior

The signpost content should be read and should indicate the trigger behavior.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

Do the following for the signpost trigger button.

1) Provide an accessible name that describes what the button does for the user using aria-label. Of course its purpose is for toggling the visibility of hidden content but we probably want something user friendly. 

2) Manage aria-expanded on the signpost-trigger button below.

Do the following for the signpost close button.

1) Ensure that when the signpost close button is activated, aria-expanded is set to false on the signpost-trigger button
2) Focus should return to the signpost-trigger button after the signpost is closed by the signpost close button. 

Finally, use aria-controls on both the signpost trigger button and signpost close button to create a reference between the signpost content and these two controls that toggle its readability.

hippee-lee:
- Adds proper aria attributes to signpost trigger and signpost content components
- adds service for uniq id's (on content and used in aria-controls
- closes #3523


src/clr-angular/popover/signpost/_signposts.clarity.scss
```diff
@@ -340,20 +340,25 @@
...
-    flex-direction: column;
+    //display: flex;
+    //flex-direction: column;
     z-index: map-get($clr-layers, tooltips);
+    position: relative;
   }
 
   .signpost-content-body {
-    flex: 1 1 auto;
+    //flex: 1 1 auto;
```
coryrylan: nit: delete the unused commented code

src/clr-angular/popover/signpost/signpost.spec.ts
```diff
@@ -97,6 +103,50 @@ export default function(): void {
...
       });
     });
+
+    describe('aria-control values', () => {
+      spec(ClrSignpost, TestDefaultSignpost, ClrSignpostModule);
+
+      function checkAriaControlsId(id: string, element: HTMLElement) {
+        const triggerControlsValue = element.querySelector('.signpost-action').getAttribute('aria-controls');
+        const closeControlsValue = element
+          .querySelector('.signpost-content > .signpost-wrap >' + ' .signpost-content-header > .signpost-action')
```
coryrylan: Could this querySelector be simplified? Seems like it doesn't need to append two separate strings.

src/clr-angular/popover/signpost/_signposts.clarity.scss
hippee-lee: I removed this and two others

src/clr-angular/popover/signpost/signpost.spec.ts
hippee-lee: I simplified this.

hippee-lee:
While addressing feedback I updated the reference images. Due to the layout changes to make the projected content accessible there were small visual changes to the dev app content and focus behavior.

src/clr-angular/popover/signpost/providers/signpost-id.service.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { Subject } from 'rxjs';
+
+@Injectable()
+export class SignpostIdService {
+  private _id: Subject<string> = new Subject<string>();
+
+  setId(id: string) {
+    this._id.next(id);
```
mathisscott: Why is this `setId()` and not `set id():`?

src/clr-angular/popover/signpost/signpost-trigger.ts
```diff
@@ -19,16 +30,30 @@ import { IfOpenService } from '../../utils/conditional/if-open.service';
  */
 export class ClrSignpostTrigger implements OnDestroy {
   private subscriptions: Subscription[] = [];
+  public ariaExpanded;
```
mathisscott: Does `ariaExpanded` have a type?

It looks like it's a boolean...

src/clr-angular/popover/signpost/signpost.spec.ts
```diff
@@ -97,6 +103,50 @@ export default function(): void {
...
+      spec(ClrSignpost, TestDefaultSignpost, ClrSignpostModule);
+
+      function checkAriaControlsId(id: string, element: HTMLElement) {
+        const triggerControlsValue = element.querySelector('.signpost-action').getAttribute('aria-controls');
+        const closeControlsValue = element
+          .querySelector('.signpost-content .signpost-action')
+          .getAttribute('aria-controls');
+        const contentId = element.querySelector('.signpost-content').getAttribute('id');
+
+        expect(id).toBe(contentId);
```
mathisscott: This is great but is it clear where this breaks when it breaks?

If not, we may want to add the optional string on the expect...

`expect(id).toBe(whatWeAreTesting, 'string that is passed that describes what we are testing and why it might be broken')`

src/clr-angular/popover/signpost/providers/signpost-id.service.ts
hippee-lee: Getters and setters in typescript need to have the same type. I'm using a getter for the Observable to have cleaner syntax when subscribing and that means I can't have  a setter with a boolean arg.

src/clr-angular/popover/signpost/signpost-trigger.ts
hippee-lee: I thought it would get inferred by the `tsc` but will add it for Clarity.

src/clr-angular/popover/signpost/signpost.spec.ts
hippee-lee: I will add the string to the expect for more descriptive failures.

PR #3533, opened 2019-06-28
[NG]  Init clrProgressBar Component

### LINKED ISSUE #3516: Progress bar updates are not announced to screen reader users
## Describe the bug

If a progress bar indicates updates that are meaningful, they are not announced.

## How to reproduce

https://clarity.design/documentation/progress

- Using a screen reader, activate the first 'Show' button on the page and observe that you do not hear any information about the progress bar updates that are occurring visually

## Expected behavior

Use a live region to announce progress updates as seen at
http://chris.homepages.org/clane/portfolio/ariaProgressBar/ariaProgressBar.html

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This is very similar to #3423, if not a duplicate.

bdryanovski:
Base on the same component design as `clrSpinner`. `clrProgressbar` must provide `aria-*` attributes to the component.

```html
<clr-progress-bar clrValue="45"></clr-progress-bar>
```
The component by default will handle `<number>%` format. If there is a need of some other suffix the developer will have to handle it by itself - passing `displayval` 

There is an option to change the `aria-live` same as `clrSpinner` - `liveRegion`

Attaching any of the standard HTML classes must provide the same result, like `labeled`, `progress-fade`, etc ...

Close #3516


src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -0,0 +1,48 @@
...
+  host: {
+    '[attr.class]': 'elementClass',
+  },
+})
+export class ClrProgressBar {
+  @Input('class') classes: string;
+  @Input('max') max: number = 100;
+  @Input('value') value: number = 0;
+  @Input('displayval') displayval: string;
+  @Input('liveRegion') liveRegion: string = 'polite';
```
coryrylan: The Progress bar will need to support taking in an `id` input to set on the inner `id` of the progress bar for a11y. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

You can set the inner id of the progress via an input and then delete the outer wrapper id to prevent duplicate ids with a HostBinding. (There might be a better way to do this but this is what I came up with) https://coryrylan.com/blog/prevent-attribute-reflection-in-angular

src/website/src/app/documentation/demos/progress-bars/progress-bar-component.demo.html
```diff
@@ -0,0 +1,40 @@
...
+    <br/>
+
+    <clr-code-snippet [clrCode]="example1"></clr-code-snippet>
+    <clr-progress-bar value="65" class="labeled"></clr-progress-bar>
+
+    <br/>
+    
+    <clr-code-snippet [clrCode]="example2"></clr-code-snippet>
+    <div class="progress-block">
+        <label>Text on the left</label>
```
coryrylan: You can use this example to associate a `label` to the progress component via `id` and `for`.

src/clr-angular/progress/progress-bars/progress-bar.ts
bdryanovski: Done ;)

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component, Input, HostBinding } from '@angular/core';
+
+@Component({
+  selector: 'clr-progress-bar',
+  template: `
+  <ng-container>
```
gnomeontherun: Container?

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -0,0 +1,62 @@
...
+    <progress [id]="id" [attr.max]="max" [attr.value]="value" [attr.data-displayval]="displayValue"></progress>
+    <span *ngIf="displayAriaLive()" [attr.aria-live]="liveRegion">{{ displayValue }}</span>
+  </ng-container>
+  `,
+  host: {
+    '[attr.class]': 'elementClass',
```
gnomeontherun: I'm not a fan of this pattern, because at first I thought you might be resetting any classes. I would rather add the targeted specific classes as necessary instead of trying to manage the class list ourselves. 

In fact, I would prefer to have the variations provided as attributes instead.

```
<clr-progress-bar danger fade></clr-progress-bar>
```

Then internally we build the necessary classes. That really internalizes the classes and structure, and I think aligns to future work.

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -0,0 +1,62 @@
...
+})
+export class ClrProgressBar {
+  /**
+   * Handle component ID
+   */
+  private _ID = '';
+  @HostBinding('attr.id') externalId: string = '';
+  @Input('id')
+  set progressId(value: string) {
+    this._ID = value;
```
gnomeontherun: Why don't we just autogenerate a unique id here like elsewhere?

src/clr-angular/progress/progress-bars/progress-bar.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        declarations: [TestComponent],
+        imports: [ClrProgressBarModule],
+      });
+
+      fixture = TestBed.createComponent(TestComponent);
+      fixture.detectChanges();
+
+      clrProgressBar = fixture.debugElement.query(By.directive(ClrProgressBar)).nativeElement;
+    });
+    it('expect to project content', () => {
```
gnomeontherun: There could be more tests.

1) Test the default displayValue, and overriden value
2) Check the ID is set and linked
3) Test all inputs get set properly

src/clr-angular/progress/progress-bars/progress-bar.ts
bdryanovski: It's React habit. Did not even cross my mind that I don't need it.

gnomeontherun:
I'm going to review this only after we get the other one nailed down, since it has some duplicate patterns.

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -0,0 +1,116 @@
...
+      return 'assertive';
+    }
+    if (isBooleanAttributeSet(this.off)) {
+      return 'off';
+    }
+    return 'polite';
+  }
+
+  constructor(@Inject(UNIQUE_ID) progressBarId: string) {
+    this.id = progressBarId;
```
gnomeontherun: I don't understand the purpose of this, a random ID doesn't give the application a way to know what ID will be assigned, and there is no use of this except to set the ID. Am I missing something or can we get away without setting a unique ID, and let apps set their own ID if they wish?

src/website/src/app/documentation/demos/progress-bars/progress-bar-component.demo.html
```diff
@@ -0,0 +1,61 @@
+<!--
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
```
gnomeontherun: There is no demo for the dev app.

src/website/src/app/documentation/demos/progress-bars/progress-bars.demo.html
```diff
@@ -189,5 +189,10 @@ <h4 id="progress-bar-versus-spinner">Progress Bar Versus Spinner</h4>
                 a matter of available space and activation point. In many cases, spinners are best
                 because they occupy less space.</p>
         </div>
+
+        <div id="design-guidelines">
```
gnomeontherun: We want to comment this out for now until next release.

src/clr-angular/progress/progress-bars/progress-bar.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  template: `<clr-progress-bar 
+    clrLabeled 
+    clrFade 
+    clrLoop
+    clrSuccess
+    clrDanger
+    clrFlash
+    clrFlashDanger
+
+    class="random"
```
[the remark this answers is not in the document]
bdryanovski: I never thought that space could be funny :)

It's more of a visual thing to separate custom API from the native one.

src/clr-angular/progress/progress-bars/progress-bar.ts
bdryanovski: https://github.com/vmware/clarity/pull/3533/files/35efa317d0609fde3951ceb33b2d77d44e9710d5#r299520849

Did you mean something like `${progressBarId}progressbar`

gnomeontherun: That commit doesn't show up, but I mean why do we need to set a default ID? If the user sets one, we pass it along otherwise can it just be empty?

PR #3537, opened 2019-07-01
Datagrid focus management (part 2 - show/hide columns)

### LINKED ISSUE #3370: Datagrid focus management

Jinnie:
- fixed focus lost after "select all" button clicked
- fixed no notification on popup opened
- fixed focus lost on column checkbox toggled
- minor selector improvement suggested by Shijir in part 1

closes #3370

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
@@ -3,10 +3,10 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component } from '@angular/core';
+import { Component, EventEmitter, Output } from '@angular/core';
 import { ColumnsService } from './providers/columns.service';
 import { ColumnState } from './interfaces/column-state.interface';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Observable, Subject } from 'rxjs';
 import { DatagridColumnChanges } from './enums/column-changes.enum';
 
 @Component({
@@ -24,6 +24,13 @@ import { DatagridColumnChanges } from './enums/column-changes.enum';
 export class ClrDatagridColumnToggleButton {
   constructor(private columnsService: ColumnsService) {}
 
+  private allSelected: Subject<boolean> = new EventEmitter();
+
+  @Output('clrAllSelected')
+  get clrAllSelected(): Observable<boolean> {
+    return this.allSelected.asObservable();
+  }
+
   private hideableColumns(): BehaviorSubject<ColumnState>[] {
     return this.columnsService.columns.filter(column => column.value.hideable);
   }
@@ -39,5 +46,6 @@ export class ClrDatagridColumnToggleButton {
         changes: [DatagridColumnChanges.HIDDEN],
       })
     );
+    this.allSelected.next(true);
   }
 }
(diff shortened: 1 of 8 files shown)


src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -41,6 +41,8 @@ export class ClrCommonStringsService implements ClrCommonStrings {
   maxValue = 'Max value';
   modalContentStart = 'Beginning of Modal Content';
   modalContentEnd = 'End of Modal Content';
+  showColumnsMenuDescription = 'Show or hide columns menu';
```
bdryanovski: Can you add them here as well 

```
src/website/src/app/documentation/demos/i18n/i18n.demo.ts
```

Jinnie: Thanks! Done.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -106,5 +117,22 @@ export class ClrDatagridColumnToggle {
 
   toggleSwitchPanel() {
     this.open = !this.open;
+    if (this.open && isPlatformBrowser(this.platformId)) {
+      this.zone.runOutsideAngular(() => {
```
gnomeontherun: Not necessarily something we should do in this PR, but seems like we'll be doing this a number of times. It might be good to extract this into a service so we do it the same way each time with the same checks.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -76,11 +81,17 @@ export class ClrDatagridColumnToggle {
   customToggleTitle: ClrDatagridColumnToggleTitle;
   @ContentChild(ClrDatagridColumnToggleButton, { static: false })
   customToggleButton: ClrDatagridColumnToggleButton;
+  @ViewChild('menuDescription', { read: ElementRef, static: false })
+  menuDescriptionElement: ElementRef<HTMLElement>;
```
gnomeontherun: These can be private.

src/clr-angular/data/datagrid/datagrid-column-toggle-button.ts
```diff
@@ -39,5 +46,6 @@ export class ClrDatagridColumnToggleButton {
         changes: [DatagridColumnChanges.HIDDEN],
       })
     );
+    this.allSelected.next(true);
```
gnomeontherun: We're missing the case where you select them all individually, I'm not sure if we want to announce at this point too or not?

Jinnie: It was about getting response. If the user selects one by one, he gets "option x, checked" feedback. With the Select All it was just silence.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
Jinnie: Not a bad idea. Let's adopt the "Rule or three" principle :) When we do it for the 3rd time, we extract it into a helper class or service.

src/clr-angular/data/datagrid/datagrid-column-toggle.ts
Jinnie: Done

PR #3539, opened 2019-07-01
Topic/vol15icons

mathisscott:
[ICONS] Adding vol.15 icons

• Includes more text-edit icons (Text Editor)
• Includes all kubernetes icons (Technology)
• Includes a "beta" icon (Essentials)
• Includes a storage adapter icon (Technology)
• Verified locally by building website and also verified full build

Tested in:
✔︎ Chrome


src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -1162,6 +1162,36 @@ export const ClrShapeHostGroup = clrIconSVG(
...
+);
+export const ClrShapeKubernetesNamespace = clrIconSVG(
+  `<path d="M27,4.18a1,1,0,1,0-1,1.73l6,3.47V26.62l-6,3.47a1,1,0,0,0-.37,1.36,1,1,0,0,0,1.37.37l7-4.05V8.23Z" class="clr-i-outline clr-i-outline-path-1"/>
+  <path d="M9.68,29.9,4,26.62V9.38L9.68,6.1a1,1,0,1,0-1-1.73L2,8.23V27.77l6.68,3.86a1,1,0,0,0,1.37-.37A1,1,0,0,0,9.68,29.9Z" class="clr-i-outline clr-i-outline-path-2"/>
+  <path d="M10,12V24a2,2,0,0,0,2,2H24a2,2,0,0,0,2-2V12a2,2,0,0,0-2-2H12A2,2,0,0,0,10,12Zm7,0v5H12V12Zm-5,7h5v5H12Zm7,5V19h5v5Zm5-7H19V12h5Z" class="clr-i-outline clr-i-outline-path-3"/>`
+);
+export const ClrShapeKubernetesNode = clrIconSVG(
+  `<path d="M18,30.66,7,24.33V11.67L18,5.34l11,6.33V24.33ZM9,23.18l9,5.17,9-5.17V12.82L18,7.65,9,12.82Z" class="clr-i-outline clr-i-outline-path-1"/>`
+);
+export const ClrShapeKubernetesPod = clrIconSVG(
```
Shijir: These `Kubernetes` icons seem to be too specific to Kubernetes product and not really necessary for other teams or our general audience/users. Why couldn't we ask them to add these icons on their part?

gnomeontherun: Also haven't we avoided brand icons in the past? Is this something we can ship?

mathisscott: @lil-kim will have more info on the whys behind this addition.

re: what i know. My understanding is that third-parties need these icons for integrations with vmware products around k8s. In order for them to use these, they need to be open source. My understanding may be wrong. But @lil-kim can clear that up, if so.

re: brand icons. Yes, we typically avoid them. But k8s is open source. And these icons represent clusters/pods and are not representative of k8s brand. They could easily ship without the k8s name but the name was retained, I'm supposing, to distinguish them from other icons in the Tech set.

lil-kim: @joshjohnson72 requested these icons be available in the library, but he had no preference whether it was public or internal only.

mathisscott: would people feel better about these additions if the word "Kubernetes" was removed?

gnomeontherun: I don't have a problem with adding logos personally, if legally we are able. It just seems like the odd duck out if we start here. I should have raised this earlier. Perhaps it would also make sense to eventually have a set specifically for logos?

mathisscott: That's the hiccup, though. These _aren't_ logos at all. See [https://kubernetes.io/](https://kubernetes.io/).

I can remove the word "Kubernetes" from them to minimize confusion. The only conflict is "cluster". And I can name that "pod-cluster" or something to make it not conflict. I'd also update "cluster-group" to "pod-cluster-group" for sanity's sake.

Shijir: I think removing `Kubernetes` will help. Possible changes I am visualizing is here:
1. `kubernetes-node` -> `node`
2. `kubernetes-cluster` -> `node-cluster` ( or simply `nodes` like we have `user` and `users`?)
3. `kubernetes-cluster-group` -> `node-cluster-group`, `node-groups`, ...

4. `kubernetes-namespace` and `kubernetes-pod` - these ones are hard. Could removing `Kubernetes ` solve the issue? Or can we find even better/generic names for these icons?

mathisscott: `namespace` and `pod` work fine on their own, i think. i can make these changes if they are agreeable.

joshjohnson72: So `node`, `node-cluster`, `node-cluster-group`, `namespace`, & `pod`?  Those seem good to me.

mathisscott: I'll update those in the AM.

src/clr-icons/shapes/technology-shapes.ts
```diff
@@ -1250,6 +1280,12 @@ export const TechnologyShapes: any = {
   inductor: ClrShapeInductor,
   resistor: ClrShapeResistor,
   'host-group': ClrShapeHostGroup,
+  'storage-adapter': ClrShapeStorageAdapter,
+  'node-cluster-group': ClrShapeNodeClusterGroup,
```
Shijir: After some thinking, I think it's better to go with even shorter and more generic and more common names. I think it's better to make the following updates: `node-cluster-group` would become `node-groups` and `node-cluster` would be just `nodes`.

src/clr-icons/shapes/text-edit-shapes.ts
```diff
@@ -96,6 +137,13 @@ export const TextEditShapes: any = {
   'align-middle': ClrShapeAlignMiddle,
   'align-top': ClrShapeAlignTop,
   language: ClrShapeLanguage,
+  'foreground-color': ClrShapeForegroundColor,
```
Shijir: This one seems to go more commonly as `font-color` or `text-color`.

src/clr-icons/shapes/text-edit-shapes.ts
```diff
@@ -96,6 +137,13 @@ export const TextEditShapes: any = {
...
+  'highlight-color': ClrShapeHighlightColor,
```
Shijir: This one goes more commonly as `highlighter`.

mathisscott:
I will merge after the 2.0.2 release

PR #3555, opened 2019-07-02
Icons in buttons need accessible names

### LINKED ISSUE #3530: Icons in buttons need accessible names
## Describe the bug

The documentation and demos do not properly setup buttons with icons so a screen reader can announce the button.

## How to reproduce

https://clarity.design/documentation/button-group

- Using a screen reader find the icons examples
- Read each icon button and observe that you do not hear any names for these controls

## Expected behavior

Buttons should have accessible names. Also documentation should tell people they cannot just put in icons without accessible names.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This applies to documentation for both buttons and button groups, worth searching all of the docs for any other icon buttons missing these details.

bdryanovski:
* Update website button group examples by adding `title` to button groups that have only icons.
* Adding a reminder for Developers to add a title to icons when the button doesn't have human-readable text
* Use only `Core Icons` inside the Buttons & Buttons Group examples. No need to load other icons sets.

* Make `clrIcon` shape attribute case-insensitive - `HOME` and `home` must equal the same icon.

Close #3530 

TODO: 
Backport it to V1

### DIFF
--- a/src/clr-icons/clr-icons-element.ts
+++ b/src/clr-icons/clr-icons-element.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -57,6 +57,10 @@ ClarityIconElement.prototype._setIconSize = function(size: string) {
   }
 };
 
+ClarityIconElement.prototype._normalizeShape = function(value: string): string {
+  return value.split(/\s/)[0].toLowerCase();
+};
+
 ClarityIconElement.prototype.connectedCallback = function() {
   // One thing to note here is that the attributeChangedCallback method is called for every attribute first
   // before this connectedCallback method called only once when the custom element is inserted into the DOM.
@@ -82,7 +86,7 @@ ClarityIconElement.prototype.connectedCallback = function() {
   // This means even if the shape is not found, the injected shape will have the user-given size.
 
   if (this.hasAttribute('shape')) {
-    const shapeAttrValue = this.getAttribute('shape').split(/\s/)[0];
+    const shapeAttrValue = this._normalizeShape(this.getAttribute('shape'));
 
     this._shapeTemplateSubscription = ShapeTemplateObserver.instance.subscribeTo(
       shapeAttrValue,
@@ -135,7 +139,7 @@ ClarityIconElement.prototype.attributeChangedCallback = function(
   // This means even if the shape is not found, the injected shape will have the user-given size.
 
   if (attributeName === 'shape') {
-    this.currentShapeAttrVal = newValue.split(/\s/)[0];
+    this.currentShapeAttrVal = this._normalizeShape(newValue);
 
     // transfer change handler callback to new shape name
     if (this._shapeTemplateSubscription) {
--- a/src/website/src/app/documentation/demos/button-group/button-group.demo.html
+++ b/src/website/src/app/documentation/demos/button-group/button-group.demo.html
@@ -185,16 +185,13 @@
                             <div class="clrweb-DoxMedia-img">
                                 <div class="btn-group btn-primary btn-icon">
                                     <button class="btn">
-                                        <clr-icon shape="plus"></clr-icon>
-                                        <span class="clr-icon-title">Add</span>
+                                        <clr-icon shape="check" title="check"></clr-icon>
                                     </button>
                                     <button class="btn">
-                                        <clr-icon shape="favorite"></clr-icon>
-                                        <span class="clr-icon-title">Favorite</span>
+                                        <clr-icon shape="home" title="home"></clr-icon>
                                     </button>
                                     <button class="btn">
-                                        <clr-icon shape="download"></clr-icon>
-                                        <span class="clr-icon-title">Download</span>
+                                        <clr-icon shape="user" title="user"></clr-icon>
                                     </button>
                                 </div>
                             </div>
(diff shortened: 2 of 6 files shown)


bdryanovski:
I'm hijacking one of @hippee-lee `clrIcons` test for the case-insensitivity change. I don't think that I need one more test just for that.

src/website/src/app/documentation/demos/button-group/button-group.demo.ts
```diff
@@ -1,11 +1,28 @@
...
 
+const ACCESSIBILITY_EXAMPLE = `
+<div class="btn-group btn-primary btn-icon">
+    <button class="btn">
+        <clr-icon shape="check" title="Check"></clr-icon>
+        <span class="clr-icon-title">Check</span>
+    </button>
+    <button class="btn">
+        <clr-icon shape="home" title="home"></clr-icon>
+        <span class="clr-icon-title">Home</span>
```
coryrylan: Same note with possible screen reader reading text twice.

src/clr-icons/clr-icons-element.ts
```diff
@@ -135,7 +137,7 @@ ClarityIconElement.prototype.attributeChangedCallback = function(
   // This means even if the shape is not found, the injected shape will have the user-given size.
 
   if (attributeName === 'shape') {
-    this.currentShapeAttrVal = newValue.split(/\s/)[0];
+    this.currentShapeAttrVal = newValue.split(/\s/)[0].toLowerCase();
```
mathisscott: Is this complex enough or being used enough to abstract it into a function?

That way you could just test the function in isolation and not have two different unit tests testing the same thing in `clr-icons.spec.ts`.

bdryanovski: I manage to test it with one test only. 

And moved it into function `_normalizeShape` so feature changes will be mostly oneliners ..

src/website/src/app/documentation/demos/button-group/button-group.demo.html
```diff
@@ -185,16 +185,16 @@ <h5>Width & Spacing</h5>
                             <div class="clrweb-DoxMedia-img">
                                 <div class="btn-group btn-primary btn-icon">
                                     <button class="btn">
-                                        <clr-icon shape="plus"></clr-icon>
-                                        <span class="clr-icon-title">Add</span>
+                                        <clr-icon shape="check" title="check"></clr-icon>
```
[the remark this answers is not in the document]
bdryanovski: Not sure if I'm doing it wrong (the test).. 

```html
 <div class="btn-group btn-primary btn-icon">
       <button class="btn">
              <clr-icon shape="check" title="Check"></clr-icon>
              <span class="clr-icon-title">Check</span>
      </button>
      ....
```

Kapture 2019-07-05 at 15 24 38

When removed `title="Check"` the reading was only "button" - Not sure why the screen reader ignore the `<span class="clr-icon-title">Check</span>`


This is working fine..
```html
 <button class="btn">
     User Profile
     <clr-icon shape="user"></clr-icon>
</button>


 <button class="btn">
     <clr-icon shape="user"></clr-icon>
     User Profile
</button>
```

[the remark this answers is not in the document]
coryrylan: I think this is because the `clr-icon-title` uses display none to hide the text and the screen reader won't pick it up because of this. https://github.com/vmware/clarity/blob/master/src/clr-angular/button/button-group/_button-group.clarity.scss#L128 I don't think we want to use the `.clr-icon-title` anymore but either use the `title` on the icon element or `aria-label` on the button as described here https://clarity.design/icons/accessibility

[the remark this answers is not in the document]
coryrylan: The reason I think the `clr-icon-title` exists in those examples was it was to trying to make it accessible but it was before we knew to properly use the `aria-label` or the `title` input on the icon.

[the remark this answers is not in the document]
coryrylan: It looks like in the docs the `clr-icon-title` is only used for the overflow group 

Screen Shot 2019-07-05 at 8 52 38 AM

src/clr-icons/clr-icons-element.ts
```diff
@@ -57,6 +57,10 @@ ClarityIconElement.prototype._setIconSize = function(size: string) {
   }
 };
 
+ClarityIconElement.prototype._normalizeShape = function(value: string): string {
+  return value.split(/\s/)[0].toLowerCase();
+};
```
mathisscott: It doesn't sit well with me that people can define shape as `home`, `HOME`, `HoMe`, or `HomE improvement` and all get the same icon. That's weird and not something you would expect from an API. I can't, for example, expect to use `@Component`, `@component`, and  `@ComPoneNT` in Angular.

But I acknowledge that we were already doctoring the shape string and reluctantly approve. As long as this is something that is only there to guard users against screwing up and not a feature we are promoting.

gnomeontherun: Well, semantically HTML tags and attributes are not case sensitive, but values are. This is a departure from that, but I think most browsers handle this except in cases like `placeholder=""` where it is taken as case sensitive.

src/website/src/app/documentation/demos/button-group/button-group.demo.html
```diff
@@ -303,6 +297,15 @@ <h6 class="clrweb-DoxMedia-do-dont">Don't</h6>
                 </div>
             </div>
 
+            <h3>Accessibility</h3>
+            <p>
+                If your icon button is has no text, we recommend adding the <code class="clr-code">title=""</code> attribute to
```
Jinnie: Nit: ...~is~ has...

PR #3559, opened 2019-07-02
[NG] stepper a11y states

### LINKED ISSUE #3529: Stepper content states not announcing to screen readers
## Describe the bug

When a stepper is in an error or success state, the state is only represented with color/icon and not announced (such as "completed"). Also, there are cases where the panel cannot be closed (step 1 which is not completed), and should be marked as such.

## How to reproduce

https://clarity.design/documentation/stepper

- Using as screen reader, go to the stepper example
- Trigger an error message by using the Tab key to focus the 'First Name' field and then press the Tab key again to move focus to the next control in the Tab order.
- Observe that an error icon has appeared within the "clr-accordion-header-button" button 
- Tab to the "clr-accordion-header-button" button and observe that you do not hear any information about the application being in an error state when the read this button
- Same for success

- Navigate to a stepper with a screen reader and try to close a step that is not closable, get confusing messaging about expandable state

## Expected behavior

- Create a textual error/success message to be used with aria-describedby to programmatically associate the error/success message with the "clr-accordion-header-button" button.
- Set `disabled` state on clr-accordion-header-button when state cannot be toggled (even if open).

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

coryrylan:
- Describe current state with button, helpful for screen reader users who tab back to a prior step.
- Disable button when locked on step (error or inactive)

closes #3529

### DIFF
--- a/src/clr-angular/accordion/accordion-panel.html
+++ b/src/clr-angular/accordion/accordion-panel.html
@@ -1,5 +1,5 @@
 <ng-container *ngIf="panel | async; let panel">
-  <div *ngIf="panel.status !== AccordionStatus.Inactive" aria-live="assertive" class="clr-sr-only">
+  <div *ngIf="panel.status !== AccordionStatus.Inactive" [id]="getAccordionStatusId(panel.templateId)" aria-live="assertive" class="clr-sr-only">
     <ng-container *ngIf="panel.status === AccordionStatus.Error">{{commonStrings.danger}}</ng-container>
     <ng-container *ngIf="panel.status === AccordionStatus.Complete">{{commonStrings.success}}</ng-container>
   </div>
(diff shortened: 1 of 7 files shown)


src/clr-angular/accordion/accordion-panel.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="panel | async; let panel">
-  <div *ngIf="panel.status !== AccordionStatus.Inactive" aria-live="assertive" class="clr-sr-only">
+  <div *ngIf="panel.status !== AccordionStatus.Inactive" [id]="'clr-accordion-header-status-' + panel.templateId" aria-live="assertive" class="clr-sr-only">
```
mathisscott: this already has the approvals so I won't ding it. 😊 

but usually when i find myself doing things like concatenating the same string in multiple places, i'll make it a function or a method so i don't, say, change `clr-accordion-header-status` to something different in one place and forget about it in the other...

coryrylan: Fixed up

PR #3579, opened 2019-07-03
Backport * Datagrid focus management (part 2 - show/hide columns)

Jinnie:
Note! Some parts rewritten due to big codebase difference.
Will appreciate a review!!!
(cherry picked from commit 5de0dd6d456434a743593f80b9336eeeef5639f1)
- fixed focus lost after "select all" button clicked
- fixed no notification on popup opened
- fixed focus lost on column checkbox toggled

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-toggle.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-toggle.ts
@@ -3,7 +3,19 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, ContentChild, ContentChildren, OnDestroy, OnInit, QueryList } from '@angular/core';
+import {
+  Component,
+  ContentChild,
+  ContentChildren,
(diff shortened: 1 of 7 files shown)


src/clr-angular/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -131,6 +152,9 @@ export class ClrDatagridColumnToggle implements OnInit, OnDestroy {
     this.hideableColumnService.showHiddenColumns();
     this.allColumnsVisible = this.hideableColumnService.checkForAllColumnsVisible;
     this.columnToggleButtons.selectAllDisabled = this.allColumnsVisible;
+    if (this.allSelectedElement) {
```
Jinnie: Honestly, I don't like this, as it's here only to make some tests pass, that are directly creating `new ClrDatagridColumnToggle(....... )` and not initializing the component as such.
But I also don't thinks it's worth prioritizing the update of these tests at the moment, just for the backported issue.

Issue #2807, opened 2018-11-02
Login page design enhancements


gnomeontherun:
## Summary

The login screen has some design concerns, and needs some attention to detail.

* The vertical rhythm is does not flow well
* There are some small details, padding/margins/spacing that need perfection
* It doesn't use form labels properly

## Use case

Apps that need logins with excellent login screens.

## Examples

google.com

## Workarounds

Just manual CSS overrides.

mathisscott:
This will also require reaching out to VMware product teams to gauge impact on products currently using Clarity.

gkhadK:
@gnomeontherun I understand this bug is for login page improvements. But just wanted to check if we are tracking the placeholder label issues all together here or in a separate bug? Along with login page, other Impacted components are - Datalist input fields, forms. text areas, inputs etc. Please advise.

gnomeontherun:
There are some examples on the website today that don't have labels, but those are documentation issues and not problems with Clarity itself. We have other issues for missing labels in documentation that probably cover all of the cases.

gnomeontherun:
The feature request here has been captured into our list and we’re going to take it into consideration as we develop Clarity Core capabilities. In an effort to clean up our backlog and focus our attention, I’m going to close this as captured in our feature requests. Please follow our development and releases to see when we release relevant components to make this possible. Future feature requests can be made in our [GitHub Discussions](https://github.com/vmware/clarity/discussions).

PR #3584, opened 2019-07-03
fix(login) the login screen was missing labels for form fields

### LINKED ISSUE #3522: Login form doesn't use labels
## Describe the bug

Login form fields do not have labels, and screen readers will not understand their roles.

## How to reproduce

https://clarity.design/documentation/login

- Using a screen reader, read any of the fields in the 'Login Page' example and observe that they are not labelled.

## Expected behavior

Use <label> elements with the for attribute to provide visible textual labels for all input fields.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This changes the login layout and should have UX input before implementation.

gnomeontherun:
Also updated to use alerts, since those will announce when errors are triggered.

closes #3522

### DIFF
--- a/src/dev/src/app/login/login.demo.html
+++ b/src/dev/src/app/login/login.demo.html
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
@@ -14,15 +14,18 @@
       </section>
       <form class="login-group">
         <clr-select-container>
+          <label class="clr-sr-only">User Role</label>
           <select clrSelect name="type" [(ngModel)]="form.type">
             <option value="local">Local Users</option>
             <option value="admin">Administrator</option>
           </select>
         </clr-select-container>
         <clr-input-container>
+          <label class="clr-sr-only">Username</label>
           <input type="text" name="username" clrInput placeholder="Username" [(ngModel)]="form.username"/>
         </clr-input-container>
         <clr-password-container>
+          <label class="clr-sr-only">Password</label>
           <input type="password" name="password" clrPassword placeholder="Password" [(ngModel)]="form.password"/>
         </clr-password-container>
         <clr-checkbox-wrapper>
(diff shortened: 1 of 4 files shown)


src/dev/src/app/login/login.demo.html
```diff
@@ -12,28 +12,32 @@ <h3 class="welcome">Welcome to</h3>
...
         <clr-select-container>
+          <label>User Role</label>
           <select clrSelect name="type" [(ngModel)]="form.type">
             <option value="local">Local Users</option>
             <option value="admin">Administrator</option>
           </select>
         </clr-select-container>
         <clr-input-container>
-          <input type="text" name="username" clrInput placeholder="Username" [(ngModel)]="form.username"/>
+          <label>Username</label>
```
mathisscott: Can we make these labels screenreader only?

Adding the labels is a breaking change for the visual design. The labels and placeholders are now redundant (visually).

If we can present the labels as `sr-only` items, then I think we can move forward.

If not, I think we need to involve visual design and likely Jehad as well because this is a visual breaking change that will affect all of our applications.

And we are planning to drop it in a minor and backport it...

gnomeontherun: Well...since this is fully on the app side to implement we can let them decide can't we? Its a breaking visual change but its opt in.

mathisscott: I don't think it will be approached as "opt in" though. Applications are going to go forward with this new look (including duplicate text in labels and placeholders).

If this is our recommendation, I think it needs a pass through UX. I don't think we can push this out as-is.

colinreedmiller:
The addition of visual labels will be an improvement to the UX of the login and it will be consistent with forms elsewhere. Some investigation into the appropriate use of hint text will need to be done. When this design enhancement to the login is complete I will need to share out with product teams - office hours may be a good forum for this.

chlane:
Just noting that by not providing visible labels in this case we may have to explain why Clarity differs from the Lightening Design System. 

See the following text from https://lightningdesignsystem.com/accessibility/guidelines/web-design/

"Give every form field a visible label.
Placeholder text isn’t enough—it can deeply confuse users when it disappears.

The only exceptions to this rule (which still require a visually hidden label for assistive technology users):

Search fields, where the magnifying glass icon serves as a visible label
Publishers’ rich text area, where surrounding elements provide context"


Regarding the statement "Placeholder text isn’t enough—it can deeply confuse users when it disappears".  This issue is more severe for screen magnification users who make think the visible label has moved out of their viewport and may try to look for it to no avail.

gnomeontherun:
We've already discussed this change is what we can make currently. @colinreedmiller is aware of this and is reviewing the login page. We don't have to explain why we differ from a different design system, we just need to explain ourselves against WCAG guidelines. We want to make an improvement here, this was the minimal we could commit to in v1/v2. See #2807

PR #3599, opened 2019-07-08
[WEB] Bug/3505

### LINKED ISSUE #3505: Datagrid: sorting, filter at website
## Describe the bug
On the clarity website documentation, the datagrid  sorting and filtering does not work properly

## How to reproduce

Go to 'https://clarity.design/documentation/datagrid/full'

**Sorting**
1. Click on column 'Pokemon'
2. **Result**: column not sorted alphabetically

**Sorting**
1. Click on column 'User ID'
2. **Result**: column not sortable, on click nothing happens. 
Mouse pointer changes to pointer > therefore user expects it is clickable.
Either mouse pointer should stay 'default' or column should be sortable.

**Filter**
1. Click on filter-icon in column 'Creation date'
2. Enter '19'
3. **Result**: rows not matching the filter-pattern are displayed. Eg. "Feb 24, 2018"

## Versions

**App**
* Clarity: 2.0.1

**Device:**
* Type: HP Notebook
* OS: Win10
* Browser: Chrome 75

bdryanovski:
* Fix issue with sorting by Pokemon name column.
* User ID was not sortable but the 'hand' cursor was used when hovering the column. Remove 'hand' cursor when the column is not clickable.
* Filter for Date column is working as expected. 

The reason for showing columns that at first don't include the number `19` or any other number - is that the rows are created with `new Date()` 

```js
const row = new Date();
row.toString()
// => "Mon Jul 08 2018 19:40:56 GMT+0300 (Eastern European Summer Time)"

// We display 
"Jul 8 2018"

// When filtering 
row.toString().includes('19') 
// => true
```

So keep in mind what type of date have to be filtered and what is the exact value of the row.

Close #3505

### DIFF
--- a/src/website/src/app/documentation/demos/datagrid/filtering/filtering.html
+++ b/src/website/src/app/documentation/demos/datagrid/filtering/filtering.html
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
@@ -80,6 +80,18 @@
 
 <clr-code-snippet [clrCode]="examples.colorFilterTS" clrLanguage="typescript"></clr-code-snippet>
 <clr-code-snippet [clrCode]="examples.colorFilterHTML" clrLanguage="html"></clr-code-snippet>
+
+<p>
+    By default, filtering searches the original model value for matches. In cases where you format the text 
+    for display (such as using a pipe), you may want to create a custom filter to handle searching the
+    formatted text. Otherwise, the results you see may not be filtered in the way you expect.
+</p>
+
+<clr-code-snippet [clrCode]="examples.filterSearchResults" clrLanguage="html"></clr-code-snippet>
+<p>
+    In the example above you will need to go with custom filters that will take into account 
+    the data that the user sees is the same that he is searching into.
+</p>
 
 <h4>Preset Column Filters</h4>
 
--- a/src/website/src/app/documentation/demos/datagrid/utils/pokemon-comparator.ts
+++ b/src/website/src/app/documentation/demos/datagrid/utils/pokemon-comparator.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
--- a/src/clr-angular/utils/_helpers.clarity.scss
+++ b/src/clr-angular/utils/_helpers.clarity.scss
@@ -28,7 +28,10 @@
   border-radius: 0;
   box-shadow: none;
   background: none;
-  cursor: pointer;
+
+  @at-root button#{&} {
+    cursor: pointer;
+  }
 }
 
 //Dropdowns & Tooltips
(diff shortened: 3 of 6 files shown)


gnomeontherun:
To address this filter by date, maybe we should just convert the date to a formatted string without time and such. This is purely a demo problem it seems.

bdryanovski:
I'm not sure that we could generate `Date` object without time in it ;)
Possible solutions could be:

* Remove the filter on date columns and add the demo on another column
* Write documentation about it.
* Write string representation of `Date` and don't use `| date` to display it. So no hidden content is present.

gnomeontherun:
Sorting a date with a string filter is a problem anyways, the date object is not the same as what is projected into the field. This is a bad pattern to demonstrate. Perhaps we just remove the sorting/filtering for the date column.

Also, do we really need a comparator filter here for pokemon? Why is this necessary vs a built-in string filter?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -422,6 +422,8 @@
         flex: 1 1 auto;
         align-self: flex-start;
         display: flex;
+        // Keep default cursor - change only if needed
+        cursor: default;
```
Shijir: @bdryanovski I think this isn't fixing the culprit of the problem, but it's trying to suppress/override the symptom. For some reason, we have `pointer` style for some Datagrid buttons which of course should have `pointer` cursor by default. So there was no need to have the `pointer` style in the first place. I tracked where this `pointer` style was coming from, and I found that it's in the `clr-no-styles-button()` mixin in `_helpers.clarity.scss`. The solution is quite possibly removing it from there. But this pointer style was introduced by this [PR](https://github.com/vmware/clarity/pull/3133) which tried to solve another problem. So this issue needs broader investigation.

src/website/src/app/documentation/demos/datagrid/utils/pokemon-comparator.ts
```diff
@@ -8,6 +8,14 @@ import { User } from '../inventory/user';
 
 export class PokemonComparator implements ClrDatagridComparatorInterface<User> {
   compare(a: User, b: User) {
-    return a.pokemon.number - b.pokemon.number;
+    const aPokemon = a.pokemon.name.toLowerCase();
```
Shijir: Please take a look at the expandable datagrid demo. Sorting is just working fine. The issue was that we simply forgot to indicate that we are using the number property of Pokemon object to sort them like in the custom sorting demo: https://clarity.design/documentation/datagrid/custom-sorting.

So introducing this change will break other demos that sort items using the number property of Pokemon object

src/clr-angular/data/datagrid/_datagrid.clarity.scss
coryrylan: Agree this is not quite the right fix. The button reset we probably still want to have a pointer but since we use the reset on non button elements as well its causing the above issue. A possible fix is to scope the pointer to only apply to button elements on the reset.

```scss
@mixin clr-no-styles-button {
  margin: 0;
  padding: 0;
  border: none;
  border-radius: 0;
  box-shadow: none;
  background: none;
  
  @at-root button#{&} {
    cursor: pointer;
  }
}
```

This will allow the reset to be used on any element but apply the proper cursor on buttons.

src/website/src/app/documentation/demos/datagrid/filtering/filtering.html
```diff
@@ -81,6 +81,17 @@ <h2>Custom filters</h2>
 <clr-code-snippet [clrCode]="examples.colorFilterTS" clrLanguage="typescript"></clr-code-snippet>
 <clr-code-snippet [clrCode]="examples.colorFilterHTML" clrLanguage="html"></clr-code-snippet>
 
+<p>
+    When you apply the filter to columns that use some visual transformation like <code class="clr-code">Pipes</code> or manually apply information by concatenating strings,
```
gnomeontherun: `By default, filtering searches the original model value for matches. In cases where you format the text for display (such as using a pipe), you may want to create a custom filter to handle searching the formatted text. Otherwise, the results you see may not be filtered in the way you expect.`

PR #3606, opened 2019-07-09
[UI] Update line-height for clr-control-label

### LINKED ISSUE #3422: Labels on forms are misaligned vertically
## Describe the bug

Not sure if this is only on the clarity doc site or not, but you see it there. The labels on the new forms are not aligned properly vertically.

## How to reproduce

https://clarity.design/documentation/forms#guidelines

## Expected behavior

Be aligned

## Versions

1.x and 2.x

**Device:**

Chrome

bdryanovski:
Label and Container were not aligned with little

Before
59040465-b5c12b00-8844-11e9-9f2e-8c67e6d1c3ea

Now
Screen Shot 2019-07-09 at 4 14 01 PM

Close #3422

TODO:
Backport v1


bdryanovski:
I may need some time to fix the broken tests :)

bdryanovski:
I had to fix some unittest that depended on the height of some fields c8eb90e46dc17f775d8a4cbaef1d118971a9d512

The thing that is bothering me is that some of the heights were strange numbers like 27, 56 and so on.. Not sure why?! Also to fix some of them I had to disable the `verticalRhythm` check - is this ok?

src/clr-angular/forms/styles/_containers.clarity.scss
```diff
@@ -28,7 +28,7 @@
     color: $clr-forms-label-color;
     font-size: $clr-forms-label-font-size;
     font-weight: $clr-forms-label-font-weight;
-    line-height: $clr-forms-baseline * 3;
+    line-height: $clr-forms-baseline * 4;
```
gnomeontherun: The fix here is different, because making a larger line-height causes issues when the label is longer than its space and wraps. We want to have 18px line height for those cases, so the fix I believe would be putting a `margin-top: 0.125rem` on the label only when it is in horizontal or compact mode. This is actually fixed in compact mode further down.

src/clr-angular/forms/styles/containers.spec.ts
```diff
@@ -277,7 +277,7 @@ describe('Form layouts', () => {
     const computed = height(selector);
     expect(Math.round(computed)).toEqual(Math.round(value));
     if (verticalRhythm) {
-      expect(value % 6).toEqual(0);
+      expect(value % base).toEqual(0);
```
gnomeontherun: After making the change above, many of these tests should pass again without changes.

PR #3612, opened 2019-07-11
fix: reverting app-level orange alerts to intended color

### LINKED ISSUE #3604: Documentation: App level alert> bg color of warning alert is wrong
## App level alert - bg color of warning alert is wrong.

The BG color used to be a rust color which was chosen because it met contrast requirements with white text. It's now a much lighter color with white text and fails contrast a11y. It should be reverted back to the rust color: #C25400

1.  Go to https://clarity.design/documentation/alerts

## Expected behavior

Revert the color back to:
#C25400
It should look like this:

mathisscott:
• During the SASS variable cleanup, I made updates to the orange color palette that broke the a11y contrast of app-level alerts
• This change fixes that regression

Tested in:
✔︎ Chrome
✔︎ Firefox

Fixes #3604


mathisscott:
@Jinnie 
Digging back through the changes, it appears the change of the orange app-alert happened as a result of aligning our orange color palette with the orange color palette on our website. I recall a discussion where we highlighted that the new orange was good because it was more distinct from the red app-alert. But clearly that change needs more thought.

In summary, yes. We agree. 😊 But it requires more design and we want to solve for an a11y issue with this PR.

Issue #3553, opened 2019-07-02
Dropdown close example needs more content for accessibility


gnomeontherun:
## Describe the bug

The documentation example for the Dropdown Close example needs an accessible name for the menu button

## How to reproduce

https://clarity.design/documentation/dropdowns

- View dropdown page
- The close button (the icon example at the bottom) needs an accessible name.

## Expected behavior

It has an accessible name.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

_Add any other notes about the problem here._

Issue #3552, opened 2019-07-02
Screen readers are announcing an incorrect menu item count on dropdown


gnomeontherun:
## Describe the bug

Screen readers are announcing an incorrect menu item count on dropdown. This may be caused by some items missing role="menuitem".

## How to reproduce

https://stackblitz.com/github/clane/clarityTest2/tree/master/clarity-test-app

- Using Mac OS X 10_14_4/VoiceOver/Safari Version 12.1.1 (14607.2.6.1.1)
- Activate the Dropdown Demos link under the Dropdown heading
- Activate the Dropdown 1 link under the 'Positioning Demo' heading
- Observe that VoiceOver announces that the menu has 6 items, but there are only 4 visible.

## Expected behavior

There should be proper semantics.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

Follow the example at
http://whatsock.com/tsg/Coding%20Arena/ARIA%20Menus/Vertical%20(Internal%20Content)/demo.htm

Issue #3551, opened 2019-07-02
Dropdown headers are not read by screen readers


gnomeontherun:
## Describe the bug

The Dropdown headers are not read by screen reader users when they open the dropdown. 

This is because they are implemented as ARIA Menus, and as such they should not include headers. ARIA menus should only contain child elements with the following roles. 

- menuitem
- menuitemcheckbox
- menuitemradio

See https://www.w3.org/TR/wai-aria-practices-1.1/#menu.

## How to reproduce

https://stackblitz.com/github/clane/clarityTest2/tree/master/clarity-test-app

- Activate the Dropdown Demos link under the Dropdown heading
- Activate the Dropdown 1 link under the 'Positioning Demo' heading
- Observe that the header is not announced
- Use the up and down arrow keys to navigate through the contents of the menu and notice that you cannot read the heading this way.

## Expected behavior

Implement correct aria and semantics.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

If we have to support Windows 10/JAWS 2019/Edge, I recommend to remove the header from the ARIA menu. If not, I recommend using aria-describedby to use the header as an accessible description for each related menuitem as seen in the "aria-describedby example" example under the 'Dropdown Examples' heading. While this will not cause the screen reader to read the heading when using the arrow keys, users will hear the heading as a description of each menu item as they read them using the arrow keys.

PR #3619, opened 2019-07-12
fix: dropdown item count by screen reader + docs update

jeeyun:

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Screen reader (VoiceOver; possibly others too) is incorrectly announcing the total number of items in dropdown, as it counts every child element (including ones without the `role='menuitem'`). Also the dropdown headers are not read by screen reader.

Issue Number: #3551, #3552, #3553

## What is the new behavior?
Screen reader correctly only counts the items in dropdown. Docs are updated to show how to make the dropdown items announce the dropdown header.

* [ ] Yes
* [x] No

### DIFF
--- a/src/website/src/app/documentation/demos/dropdown/dropdown.demo.html
+++ b/src/website/src/app/documentation/demos/dropdown/dropdown.demo.html
@@ -137,7 +137,7 @@
               </div>
             </div>
           </div>
-x
+
           <h3>Placement</h3>
           <div class="clr-row">
             <div class="clr-col-12 clr-col-lg-6">
@@ -173,7 +173,7 @@ x
                     right of the first nested menu, creating a zig-zag pattern.
                   </p>
                 </div>
-              </div>f
+              </div>
             </div>
           </div>
         </div>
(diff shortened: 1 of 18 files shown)


src/website/src/app/documentation/demos/dropdown/dropdown.demo.html
```diff
@@ -186,7 +186,8 @@ <h2>Accessibility</h2>
             <li>The element with the <code class="clr-code">button</code> role should also have an <code class="clr-code">aria-expanded</code>
               attribute which is set to true when the dropdown menu is open
             </li>
-            <li>The items should have the role <code class="clr-code">menuitem</code></li>
+            <li>The items should be inside an element with the role <code class="clr-code">menu</code>, and have the role <code class="clr-code">menuitem</code></li>
+            <li>If the menu uses headings, the heading should be used as an accessible description for each related item</li>
```
Jinnie: Nit: Not sure if we should add a line about separators. I wouldn't like our documentation to be a comprehensive A11Y guide, but if it will be a common case it may be worth it.

jeeyun:
> Seems to announce pretty correctly now, but I noticed an issue - the open dropdown with arrow keys seems to be broken - it scrolls the viewport and immediately closes the dropdown.

Unfortunately this happens under a specific case (I think it's having a parent container with position relative and it actually needs to scroll when you arrow up/down).  😞  It is scheduled to be corrected with smart popover and not something introduced by this PR.

coryrylan:
I recreated the same issue @Jinnie ran into as well but on the prod version of the site so I think it is an issue that has been around and not related directly to this PR. I opened a separate issue for this bug https://github.com/vmware/clarity/issues/3622 to track and fix.

coryrylan:
```
I recreated the same issue @Jinnie ran into as well but on the prod version of the site so I think it is an issue that has been around and not related directly to this PR. I opened a separate issue for this bug #3622 to track and fix.
```

Edit: added note about smart popovers to issue

src/website/src/app/documentation/demos/dropdown/dropdown.demo.html
jeeyun: I agree with you that we need to figure out what's the right level of detail for our dox. I can add it for now with an understanding that it may be removed or expanded in a future review of dox.

PR #3626, opened 2019-07-16
fix(stepper): remove interstitial content

### LINKED ISSUE #3560: Stepper has an illogical sequence for the expanded button
## Describe the bug

The interstitial content in the 'clr-accordion-header' used in the Stepper component presents an illogical reading sequence for screen reader users. When screen reader users hear a button announce an expanded or collapsed state they will think that the button controls the readability of upcoming content. 

Discovering content related to the "clr-accordion-header-button" button immediately after hearing that it is in a collapsed state may be confusing. While this is a low severity issue when there is a small amount of interstitial content, the severity will increase in relation to the amount of interstitial content.

## How to reproduce

https://clarity.design/documentation/stepper

- Using a screen reader go to Stepper example
- Read one of the collapsed Stepper header buttons
- Observe that you read the value of the description immediately after reading the collapsed button. 

## Expected behavior

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

The issue is that the content that is controlled by the "clr-accordion-header-button" toggle button should be the next sibling of the toggle button. Having content that is not controlled by the toggle button as the next sibling will be confusing for screen reader users because it will seem like the toggle button does not work since it does not control the interstitial content.

coryrylan:
# PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

It can be confusing to have content between the step button and the panel it controls when using a screen reader.

Issue Number:
closes #3560

## What is the new behavior?

Move description as part of the button content to prevent interstitial content between button and panel for screen readers. This is not a breaking change for the angular component but does limit user to not being able to put other interactive content in step heading button (not recommended prior).

This PR also fixes a layout issue in accordion/stepper when the title could not expand if no description was given. Gemini tests updated to check for expanded title use case.

* [ ] Yes
* [x] No


src/clr-angular/accordion/_accordion.clarity.scss
```diff
@@ -64,34 +59,36 @@
...
     display: inline-block;
+
+    @media (min-width: $clr-accordion-responsive-breakpoint) {
+      min-width: 7.2rem;
+    }
+  }
+
+  .clr-accordion-header-has-description {
+    .clr-accordion-title {
+      max-width: 11rem;
```
gnomeontherun: What is the reasoning for this width? Can we let it dynamically flow instead?

coryrylan: The design spec for the stepper/accordion had a max width set on both title and description and called for it to text wrap. The original max width was on the button element which wrapped only the title but now the button spans the entire width so I had to move it to the inner title class.

src/clr-angular/accordion/accordion-panel.html
```diff
@@ -5,18 +5,17 @@
...
         class="clr-accordion-header-button"
         (click)="togglePanel()"
-        (focus)="focusHeader = true"
-        (blur)="focusHeader = false"
         [id]="getAccordionHeaderId(panel.templateId)"
         [disabled]="panel.disabled"
         [attr.aria-controls]="getAccordionContentId(panel.templateId)"
         [attr.aria-expanded]="panel.open"
         [attr.aria-describedby]="getAccordionStatusId(panel.templateId)"
+        [class.clr-accordion-header-has-description]="description.children.length != 0"
```
Shijir: Regarding `description.children.length`,  we should avoid this DOM property traversal. Two reasons for this:
1. if the description gets removed for some reasons, the class binding wouldn't be updated as we haven't queried children DOM in Angular way.
2. As for server-side rendering, we should avoid traversing nested DOM properties as much as possible.

We could accomplish it more Angular way if we use `ContentChild` on `ClrAccordionTitle` or/and `ClrAccordionTitle`.

src/clr-angular/accordion/accordion-panel.html
```diff
@@ -25,8 +24,10 @@
           <clr-icon shape="check-circle" class="clr-accordion-complete-icon"></clr-icon>
         </span>
         <ng-content select="clr-accordion-title, clr-step-title"></ng-content>
+        <div #description>
```
Shijir: Block-level elements are not allowed inside a button element. But if resort to `ContentChild` on `ClrAccordionTitle`/`ClrAccordionDescription` as I mentioned above, we could avoid this extra wrapper element altogether.

src/clr-angular/accordion/accordion-panel.html
```diff
@@ -15,7 +15,7 @@
         [attr.aria-controls]="getAccordionContentId(panel.templateId)"
         [attr.aria-expanded]="panel.open"
         [attr.aria-describedby]="getAccordionStatusId(panel.templateId)"
-        [class.clr-accordion-header-has-description]="description.children.length != 0"
+        [class.clr-accordion-header-has-description]="(accordionDescription.changes | async)?.length || accordionDescription.length"
```
coryrylan: Unfortunately changes does not emit initial value. This gets changes + initial without creating chocolate errors

src/clr-angular/accordion/accordion-panel.html
coryrylan: Used `ContentChildren` to get updates about changes as using just `ContentChild` did not update if the description was dynamically added/remove.

PR #3629, opened 2019-07-17
feat(clrSpinner): adding new component clrSpinner - Backport to v1

bdryanovski:
Backporting clrSpinner to v1 - I just will leave it until the end of the weak if someone has some comments before merging it into v1

## PR Checklist

Please check if your PR fulfills the following requirements:

* [x] Tests for the changes have been added (for bug fixes / features)
* [x] Docs have been added / updated (for bug fixes / features)
* [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

* [ ] Bugfix
* [x] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [x] Documentation content changes
* [ ] clarity.design website / infrastructure changes
* [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #3423 #3493 

## Does this PR introduce a breaking change?

* [ ] Yes
* [x] No

PR #3631, opened 2019-07-17
Fix/clr spinner

bdryanovski:
Update the API of the clrSpinner to use prefix before every input

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/progress/spinner/spinner.ts
+++ b/src/clr-angular/progress/spinner/spinner.ts
@@ -3,16 +3,8 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, Input } from '@angular/core';
+import { Component, Input, HostBinding } from '@angular/core';
 import { isBooleanAttributeSet } from '../../utils/component/is-boolean-attribute-set';
-
-const SPINNER_BASE_CLASS = 'spinner';
-
-const SPINNER_INVERSE = 'spinner-inverse';
-const SPINNER_INLINE = 'spinner-inline';
-
-const SPINNER_SMALL_SIZE = 'spinner-sm';
-const SPINNER_MEDIUM_SIZE = 'spinner-md';
 
 @Component({
   selector: 'clr-spinner',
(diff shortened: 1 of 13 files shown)


Shijir:
I've requested some changes on https://github.com/vmware/clarity/pull/3629. The same changes could be added in this PR as well.

bdryanovski:
done ;)

src/website/src/app/documentation/demos/spinners/spinner.demo.html
```diff
@@ -117,8 +117,8 @@ <h4 id="spinners-versus-progress-bars">Spinners Versus Progress Bars</h4>
                 progress bar is a matter of spacing, visual consistency, and the object the user selected
                 to begin the process.</p>
 <!-- 
-<clr-spinner-component></clr-spinner-component>
 -->
+<clr-spinner-component></clr-spinner-component>
```
gnomeontherun: Comment this back out?

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -50,6 +49,10 @@ export class ClrSpinner {
     return 'polite';
   }
 
+  // Get class values before changing it
+  @Input('class') userClasses: string = '';
```
Shijir: Unfortunately, we cannot do this because users could define classes through bindings like this: `[class.user-class]` and `[ngClass]="..."`, which won't be registered here. We shouldn't really try to manipulate classes in bulk here. As I mentioned in the other PR, we need to use HostBinding on each of the properties. That's safe Angular way.

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -3,41 +3,80 @@
...
+  set clrSmall(value: boolean | string) {
+    this._small = isBooleanAttributeSet(value);
+  }
+
+  private _medium: boolean;
+  @HostBinding('class.spinner-md')
+  get mediumClass() {
+    // You could have only one set
+    if (this._small) {
+      return false;
```
Shijir: So the behavior is when `clrSmall` is on the spinner, it overrides the `clrMedium`? if that's the case, you need to make `this._medium` to `false` when `this._small` becomes `true`.

Shijir: And please add tests for this behavior.

PR #3632, opened 2019-07-17
fix(datepicker): adds aria-labels for buttons

### LINKED ISSUE #3467: Date picker buttons need names
## Describe the bug

The date picker switcher icons (inside the popover to nagivate forwards/backwards) and all buttons need accessible names.

## How to reproduce

See the demos at https://clarity.design/documentation/datepicker

## Expected behavior

All date picker buttons should read out a meaningful label. This includes the the cells for the days of the month, month of the year, or year selection panes.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This is probably more additions to commonStrings.

hippee-lee:
- adds proper labels for all datepicker buttons
- adds live region for calendar view that updates month/year values for screen readers
- adds live region to year view that updates the decade range for screen readers
- closes #3467

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Date picker needs the correct aria-labels on all of its buttons. It also needs two aria-live regions that will update screen readers in the daypicker view **and** the yearpicker view. 

Issue Number: #3467

## What is the new behavior?
All datepicker buttons have an aria-label related to what the button action does. There are two aria-live regions that change depending on user interactions with the datepicker. 

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/forms/datepicker/date-container.ts
+++ b/src/clr-angular/forms/datepicker/date-container.ts
@@ -33,7 +33,14 @@ import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service
         <div class="clr-input-wrapper">
           <div class="clr-input-group" [class.clr-focus]="focus">
             <ng-content select="[clrDate]"></ng-content>
-            <button #actionButton type="button" class="clr-input-group-icon-action" (click)="toggleDatepicker($event)" *ngIf="isEnabled" [attr.title]="commonStrings.keys.open" [disabled]="control?.disabled">
+            <button #actionButton 
+                    type="button" 
+                    class="clr-input-group-icon-action"
+                    [attr.title]="commonStrings.keys.datepickerToggle"
+                    [attr.aria-label]="commonStrings.keys.datepickerToggle"
+                    [disabled]="control?.disabled"
+                    (click)="toggleDatepicker($event)"
+                    *ngIf="isEnabled">
               <clr-icon shape="calendar"></clr-icon>
             </button>
             <clr-datepicker-view-manager *clrIfOpen clrFocusTrap></clr-datepicker-view-manager>
@@ -56,6 +63,7 @@ import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service
     DateNavigationService,
     DatepickerEnabledService,
     DateFormControlService,
+    ClrCommonStringsService,
   ],
   host: {
     '[class.clr-form-control-disabled]': 'control?.disabled',
--- a/src/clr-angular/forms/datepicker/daypicker.html
+++ b/src/clr-angular/forms/datepicker/daypicker.html
@@ -1,21 +1,45 @@
 <div class="calendar-header">
+    <div aria-live="polite" class="clr-sr-only">
+      {{ ariaLiveMonth }}.  {{updateAriaLiveYear}}.
+    </div>
     <div class="calendar-pickers">
-        <button class="calendar-btn monthpicker-trigger" type="button" (click)="changeToMonthView()">
-            {{calendarMonth}}
+        <button
+                class="calendar-btn monthpicker-trigger"
+                type="button" (click)="changeToMonthView()"
+                [attr.aria-label]="monthAttrString"
+                [attr.title]="monthAttrString">
+                {{calendarMonth}}
         </button>
-        <button class="calendar-btn yearpicker-trigger" type="button" (click)="changeToYearView()">
+        <button
+                class="calendar-btn yearpicker-trigger"
+                type="button"
+                (click)="changeToYearView()"
+                [attr.aria-label]="yearAttrString"
+                [attr.title]="yearAttrString">
             {{calendarYear}}
         </button>
     </div>
     <div class="calendar-switchers">
-        <button class="calendar-btn switcher" type="button" (click)="previousMonth()">
-            <clr-icon shape="angle" dir="left" [attr.title]="commonStrings.keys.previous"></clr-icon>
+        <button
+            class="calendar-btn switcher"
+            type="button"
+            (click)="previousMonth()"
+            [attr.aria-label]="commonStrings.keys.datepickerPreviousMonth">
+            <clr-icon shape="angle" dir="left" [attr.title]="commonStrings.keys.datepickerPreviousMonth"></clr-icon>
         </button>
-        <button class="calendar-btn switcher" type="button" (click)="currentMonth()">
-            <clr-icon shape="event" [attr.title]="commonStrings.keys.current"></clr-icon>
+        <button
+            class="calendar-btn switcher"
+            type="button"
+            (click)="currentMonth()"
+            [attr.aria-label]="commonStrings.keys.datepickerCurrentMonth">
+            <clr-icon shape="event" [attr.title]="commonStrings.keys.datepickerCurrentMonth"></clr-icon>
         </button>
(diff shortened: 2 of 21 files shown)


hippee-lee:
@chlane - I published the dev app with a modified datepicker here: http://clr-datepicker-a11y.surge.sh

Can you review it based on our conversations and leave comments/feedback or give this PR a ✅if its behaving the way you want.

chlane:
Just tested with voiceover and finding that the button that triggers the calendar and the buttons at top of datepicker lack accessible names. The first two buttons at the top left are announced as 'NAN' and 'Undefined'. Also, the live regions announcements are only working when the buttons a triggered with Ctrl+Option+Space. Can they be made to work with the Enter key as well?

chlane:
LIve region updates are working when Enter is pressed on the various next and previous buttons

src/clr-angular/forms/datepicker/daypicker.html
```diff
@@ -1,21 +1,46 @@
 <div class="calendar-header">
+    <div aria-live="polite" class="clr-sr-only">
+        The current month is {{ calendarMonth }}.
```
gnomeontherun: This cannot be hard coded. We need to make this a translatable string. The problem is, this has a variable unlike any other string today.

src/clr-angular/forms/datepicker/yearpicker.ts
```diff
@@ -16,29 +16,44 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
...
-                [attr.tabindex]="getTabIndex(year)"
-                [class.is-selected]="year === calendarYear"
-                (click)="changeYear(year)">
-                {{year}}
-            </button>
-        </div>
-    `,
+    <div class="year-switchers">
+      <div aria-live="polite" class="clr-sr-only">
+        The current decade is {{decadeRange}}.
```
gnomeontherun: This again cannot be hard coded :(

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -33,7 +33,14 @@ import { ClrCommonStrings } from '../../utils/i18n/common-strings.interface';
         <div class="clr-input-wrapper">
           <div class="clr-input-group" [class.clr-focus]="focus">
             <ng-content select="[clrDate]"></ng-content>
-            <button #actionButton type="button" class="clr-input-group-icon-action" (click)="toggleDatepicker($event)" *ngIf="isEnabled" [attr.title]="commonStrings.open" [disabled]="control?.disabled">
+            <button #actionButton 
+                    type="button"
+                    class="clr-input-group-icon-action"
+                    [attr.title]="commonStrings.openDatepicker"
+                    [attr.aria-label]="commonStrings.openDatepicker"
```
bdryanovski: Why are we setting both `aria.title` and `attr.aria-label`? Are they not the same when it comes to screen reader?

hippee-lee: Title is needed for tooltips when mouse hovers the button. aria-label is needed to ensure screen readers (that don't read title) will read it.  

@chlane - did I miss anything?

chlane:
No that’s great thanks Matt

gnomeontherun:
> @gnomeontherun - I think I addressed the hard coding problem with a combination of a commonString phrase and leveraging the browsers i18n functionality built into the datepicker already. Do you see any issues with the change?

Yes, sorry since I'm not in I haven't been able to follow up on this fast enough.

There are a few potential word order problems. First some languages might not organize the statement with the month at the end (I see an example in Hindi where the month is not the last word). Second, RTL languages (which we don't really support) wouldn't work. Third, depending on how the statement grammar works, the noun may undergo declension. I don't see us solving all of these right now, especially the second and third point, but the first we could handle. 

I'm also ok with backlogging a ticket to fix this later (since there might be other places where these problems arise as we're adding far more strings than we ever hoped to do). Our language solution is meant to be small and quick, but with the scale we're deploying strings this is increasingly difficult to do well.

src/clr-angular/forms/datepicker/daypicker.html
hippee-lee: I've updated to use a single string with a CommonStrings service that can replace parts of the string.

src/clr-angular/utils/i18n/strings.service.ts
```diff
@@ -0,0 +1,14 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class StringsService {
+  static replaceCommonString(source: string, key: string, value: string | number): string {
+    return source.replace(key, value.toString());
```
jeeyun: I think @gnomeontherun was already touching on this but I think the new `ClrCommonStringsService`'s `parse` method can be used instead:

https://github.com/vmware/clarity/blob/d293c8f56f2853aeed0d6b0cf770d91a5534c2a7/src/clr-angular/utils/i18n/common-strings.service.ts#L34-L43

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -33,7 +33,14 @@ import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service
         <div class="clr-input-wrapper">
           <div class="clr-input-group" [class.clr-focus]="focus">
             <ng-content select="[clrDate]"></ng-content>
-            <button #actionButton type="button" class="clr-input-group-icon-action" (click)="toggleDatepicker($event)" *ngIf="isEnabled" [attr.title]="commonStrings.keys.open" [disabled]="control?.disabled">
+            <button #actionButton 
+                    type="button"
```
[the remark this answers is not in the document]
gnomeontherun: We do need it, because without it the button becomes a submit button. Its strange behavior but thats HTML spec for ya.

src/clr-angular/forms/datepicker/daypicker.html
```diff
@@ -1,21 +1,45 @@
 <div class="calendar-header">
+    <div aria-live="polite" class="clr-sr-only">
+      {{ ariaLiveMonth }}.  {{updateAriaLiveYear}}.
+    </div>
     <div class="calendar-pickers">
-        <button class="calendar-btn monthpicker-trigger" type="button" (click)="changeToMonthView()">
-            {{calendarMonth}}
+        <button
+                class="calendar-btn monthpicker-trigger"
+                type="button" (click)="changeToMonthView()"
```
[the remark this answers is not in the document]
hippee-lee: Resolving based on the previous comment about making sure this isn't a submit button.

src/clr-angular/utils/i18n/strings.service.ts
hippee-lee: Yeah, this was the  band-aid I used before his changes went in. I'm removing this service and will push the update soon.

src/clr-angular/forms/datepicker/date-container.ts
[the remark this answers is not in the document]
jeeyun: Wow, I didn't realize that. Good to know. 👍

Issue #3528, opened 2019-06-27
Stepper focus management


gnomeontherun:
## Describe the bug

There are a few places where focus is not properly managed in the stepper.

* [ ] When using the next button in each step, the focus should move to the next step title

* [x] When there are errors, the next button retains focus instead of the form fields itself (Will be resolved from https://github.com/vmware/clarity/issues/3556)

## How to reproduce

https://clarity.design/documentation/stepper

- View the stepper and navigate using next/back buttons
- Focus is lost or remains on the button and not new content

## Expected behavior

The focus should move to newly selected panel.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

_Add any other notes about the problem here._

coryrylan:
> When there are errors, the next button retains focus instead of the form fields itself (this is really a forms problem but might require work here)

@chlane What is the expected behavior here?

chlane:
The expected behavior is that screen reader users know about the errors somehow.  My recommendation for do this is in https://github.com/vmware/clarity/issues/3556 which is specific to forms.

coryrylan:
Closed in https://github.com/vmware/clarity/pull/3649

PR #3649, opened 2019-07-26
fix: focus management for stepper

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

When navigating between steps using the next button the focus was not set on the next step heading button.

Issue Number: N/A
https://github.com/vmware/clarity/issues/3528

## What is the new behavior?
When the user clicks next on a valid step the next step is automatically focused.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/accordion/accordion-panel.html
+++ b/src/clr-angular/accordion/accordion-panel.html
@@ -11,11 +11,13 @@
         class="clr-accordion-header-button"
         (click)="togglePanel()"
         [id]="getAccordionHeaderId(panel.templateId)"
-        [disabled]="panel.disabled"
+        [disabled]="isAccordion && panel.disabled"
+        [attr.aria-disabled]="!isAccordion && panel.disabled"
         [attr.aria-controls]="getAccordionContentId(panel.templateId)"
         [attr.aria-expanded]="panel.open"
         [attr.aria-describedby]="getAccordionStatusId(panel.templateId)"
         [class.clr-accordion-header-has-description]="(accordionDescription.changes | async)?.length || accordionDescription.length"
+        #headerButton
       >
         <span class="clr-accordion-status">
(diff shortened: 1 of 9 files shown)


coryrylan:
WIP: need to update the aria announcements to remove the success state as it is no longer needed with the focus management.

src/clr-angular/accordion/accordion-panel.html
```diff
@@ -11,12 +10,17 @@
         class="clr-accordion-header-button"
         (click)="togglePanel()"
         [id]="getAccordionHeaderId(panel.templateId)"
-        [disabled]="panel.disabled"
+        [disabled]="isAccordion && panel.disabled"
+        [attr.aria-disabled]="!isAccordion && panel.disabled"
```
jeeyun: Can you tell me why accordion uses `disabled` attribute and stepper uses `aria-disabled`? Also doesn't presence of `disabled` attribute make it disabled by default even if its value is set to `false`?

coryrylan: The reason the accordion uses `disabled` is we want the button truly disabled, cant be focused or clicked. When using `disabled` you can't trigger a focus so on the stepper we use `aria-disabled` so we can auto focus to the next step and read both the step title and let the screen user know the button is disabled.

When `panel.disabled` is `false` that will set the `[disabled]` JS property on the button which will reflect that state into the HTML and remove the `disabled` attribute. If I had used `[attr.disabled]="panel.disabled"` then I set the attribute skipping the property completely and `disabled="false"` would be added to the button disabling it incorrectly.

jeeyun: Thanks for the detailed explanation. It all makes sense now. 👍

PR #3660, opened 2019-08-01
refactor(i18n): replace original provider with new implementation

gnomeontherun:
BREAKING CHANGE: The previous i18n service would get removed from a production build in Angular since it wasn't referenced. To address this, there is now a service that Clarity provides itself and you can pass a custom, new localized set of strings into the service instead of trying to use the provider tree in Angular. This will be helpful to future clr-base.

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [x] Yes
* [ ] No

The problem is that unless an application explicitly defines the provider today for ClrCommonStrings, production builds will mark it as unused and tree shake it out. This is because the provider service is actually an abstract class, so it requires an implementation to remain in the build.

This change is outlined in the documentation, but the migration path is as follows.

1. Remove the provider from your `AppModule` for `ClrCommonStrings`.

2. Change your class to an object with the same key names. 

```
export const MyTranslation: ClrCommonStrings = {
  open: 'open',
  ...other keys
};
```

3. Inject the `ClrCommonStringsService` into your `AppComponent`, and pass in the localized strings.

```
import { ClrCommonStringsService } from '@clr/angular';
import { MyTranslation } from './translations/my';

@Component({...})
export class AppComponent {
  constructor(commonStrings: ClrCommonStringsService) {
    commonStrings.localize(MyTranslation);
  }
}
```


src/clr-angular/utils/i18n/common-strings.service.ts
```diff
@@ -3,65 +3,30 @@
...
-  return defaults;
-}
 
-export const COMMON_STRINGS_PROVIDER: InjectableProvider = {
-  useFactory: commonStringsFactory,
-  // We have a circular dependency for now, we can address it later once these
-  // tree-shakeable providers have proper documentation.
-  deps: [[new Optional(), new SkipSelf(), forwardRef(() => ClrCommonStrings)]],
-};
+  get(key: keyof ClrCommonStrings, tokens: { [key: string]: string } = {}) {
```
coryrylan: I think this is too dynamic. If I make a typo in the key its not caught by angular's language service or typescript in the build which leads to a silent error. Something like this will get a better dev experience and type safety in the templates.

```typescript
export class ClrCommonStringsService {
  private _strings = commonStringsDefault;

  get strings(): Readonly<ClrCommonStrings> {
    return this._strings;
  }

  localize(overrides: ClrCommonStrings) {
    this._strings = { ...this.strings, ...overrides };
  }
...
}
```

Screen Shot 2019-08-01 at 4 17 14 PM

If you update the strings to be public and readonly then they can be read but only set through the localize method. The `get` would still be used but would really just be a find replace method now since the value is passed in.

Screen Shot 2019-08-01 at 4 21 44 PM

gnomeontherun: I pushed changes for this

commitlint.config.js
```diff
@@ -1,3 +1,45 @@
...
+  'card',
+  'checkbox',
+  'datagrid',
+  'date-picker',
+  'dropdown',
+  'form',
+  'grid',
+  'header',
+  'i18n',
+  'input',
```
mathisscott: We could use a scope for general stuff. Looking at this list, I don't know what I would put the CSS work under. It doesn't really apply to build and touches about 18 other scopes. Also anything specific to base or any refactor that spans across more than one component. It would be nice to have a scope for those...

gnomeontherun: If something bridges over many aspects, I don't think you'd put in a scope. The CSS work is a good example. If it isn't isolated to one scope then no scope. Thoughts?

coryrylan: Right, I think scopes are meant to be optional and just help identify for specific issues. If not specific then no scope is needed. example `refactor: update scss vars to use css custom props`

mathisscott: But the rule on line 43. Does it always expect a scope? Or is that just saying that if there is a scope it always has to be in this enum?

gnomeontherun: It will only validate a scope against this list when it is provided, commits with no scopes are allowed.

Issue #3622, opened 2019-07-15
Dropdown flashes open when using arrow keys


coryrylan:
## Describe the bug

When a dropdown button target is focused the dropdown does not open when using the arrow keys but instead flashes open and quickly closes.

## How to reproduce

You can recreate this by looking at the dropdown angular component demo in the docs.
Will be fixed with smart popovers PR.
https://clarity.design/documentation/dropdowns

1.  Focus the dropdown button
2. Use the down arrow key and note how it flashes open but then quickly closes.

## Expected behavior

The dropdown should open and remain open without scrolling the page.

## Versions

* Angular: 8.x
* Clarity: 2.x

**Device:**

* Type: Macbook
* OS: MacOS
* Browser Chrome
* Version 47

AlexMarcuBytex:
While i was investigating this issue, i found that this bug reproduces just on nested dropdowns. If we re looking for a quick fix, we can replace that nested dropdown with a normal one, considering that isn't needed a nested one in that specific demo. I think this bug occurs because of the "clrIfOpen" directive. If i dont use it, the bug doesnt reproduce either the dropdown is nested or not.
Should this bug be fixed the way i presented above? 
Also, assign this to me for more investigations if needed.

gnomeontherun:
@Jinnie Can you take a quick look?

Jinnie:
The `clrIfOpen` is a supported and even recommended scenario.
I think that it does not _cause_ a problem, but mostly _reveals_ an underlying one.
At some point, the `keydown` event must be preventDefault-ed, so we don't allow the whole view to scroll. Probably some component does that when the menu is initialized, but it's not done when only the `dropdown-trigger` is initialized.
If you follow the code, it registers itself to a DropdownFocusHandler service where the actual handlers for arrow-up/down are registered. We have two options:
- either add preventDefault to the handlers
- or try to reuse the /src/clr-angular/utils/focus/key-focus/util.ts#preventArrowKeyScroll` helper if possible.

bbogdanov:
@Jinnie The `preventDefault` will help with the arrow keys, but what bothers me is that the popover gets closed when the parent component is scrolled. Is this an expected behavior?

PR #3661, opened 2019-08-01
fix(ng): make disabled dropdown item focusable

jeeyun:
This PR makes the disabled dropdown item focusable. We are following this guideline and good practices: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls.

This PR also fixes keyboard navigation to support ESC key correctly. 

## PR Checklist

* [x] If applicable, have a visual design approval

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3661

## What is the new behavior?

This PR makes the disabled dropdown item focusable. Now disabled item can receive focus when navigating through keyboard or using screen reader.

Previously our global ESC key listener closed all of our popovers. To assist non-sighted or keyboard-only users, we are fixing the keyboard navigation to close only the active menu (or submenu) on each ESC key press.

* [x] Yes
* [ ] No

The accessibility issues addressed here are partially fixed from Clarity side but the application developer is partially responsible to make their application fully accessible. This is because dropdown uses content projection and Clarity has no access or control over those parts.

In particular the following should be noted:
- We discourage you from using button for dropdown item UNLESS the item opens a submenu
- Correct ARIA- attributes should be added as shown in the updated documentation
- We will be deprecating `[disabled]` input on dropdown items as disabled items need to be focusable. Application should use `[clrDisabled]` input instead and make sure the item is not activated on click if disabled.


src/clr-angular/popover/common/abstract-popover.ts
```diff
@@ -114,11 +114,13 @@ export abstract class AbstractPopover implements AfterViewChecked, OnDestroy {
...
-        this.ifOpenService.open = false;
-      }
-    });
+    if (!this.popoverOptions.ignoreGlobalESCListener) {
+      this.documentESCListener = this.renderer.listen('document', 'keydown', event => {
+        if (event && event.keyCode === ESC) {
+          this.ifOpenService.open = false;
+        }
+      });
+    }
```
jeeyun: Essentially, the current version of popover listens for ESC key on document and closes them. This needs to be overridden for dropdown menu as the ESC key should only close the current popover and move the cursor to its parent.

src/clr-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -31,7 +31,7 @@ export class ClrDropdownItem implements AfterViewInit {
 
   private unlisten;
 
-  @Input()
+  @Input('clrDisabled')
```
jeeyun: turns out we can't use the `disabled` attribute anymore for dropdown item since IE won't allow focus on any element that has this. I'm renaming to `clrDisabled` here so it'll still auto attach the right class name and `aria-disabled` value.

src/clr-angular/utils/_reboot.clarity.scss
```diff
@@ -97,9 +97,9 @@
   // might still respond to pointer events.
   //
   // Credit: https://github.com/suitcss/base
-  [tabindex='-1']:focus {
-    outline: none !important;
-  }
+  //[tabindex='-1']:focus {
+  //  outline: none !important;
+  //}
```
jeeyun: I couldn't override this reboot's global setting by trying to change it in menu mixin. I think we would like to allow outline in tabindex=-1 cases globally anyways because those are the ones we'd programmatically manipulate to focus. Let me know if anyone thinks differently.

src/clr-angular/utils/focus/focus.service.ts
```diff
@@ -45,31 +45,29 @@ export class FocusService {
...
         const nextObs = isObservable(next) ? next : of(next);
         nextObs.subscribe(item => {
-          if (item.disabled) {
-            return this.move(direction, item);
-          } else {
-            this.moveTo(item);
-            return true;
-          }
+          this.moveTo(item);
+          return true;
```
jeeyun: Disabled items now need to be focusable so we are removing that conditional check to see if it's not disabled before moving focus.

src/clr-angular/utils/_reboot.clarity.scss
jeeyun: Thanks for the comment. As this reboot was taken from bootstrap originally, I thought all the commented out parts were left intentional with comments above them to show that it was done intentionally (you can see examples of doing that throughout the file). Do you think there's value in doing that? I can either add a comment above it (I don't yet have it) or remove the commented out lines as you suggest.

src/clr-angular/utils/focus/focusable-item/basic-focusable-item.service.ts
```diff
@@ -24,10 +24,16 @@ export class BasicFocusableItem implements FocusableItem {
   disabled = false;
 
   focus() {
-    this.renderer.addClass(this.el.nativeElement, 'clr-focus');
+    this.renderer.setAttribute(this.el.nativeElement, 'tabindex', '0');
```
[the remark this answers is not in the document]
jeeyun: You are right. Making the changes.

src/clr-angular/popover/common/abstract-popover.ts
```diff
@@ -114,11 +114,13 @@ export abstract class AbstractPopover implements AfterViewChecked, OnDestroy {
...
 
   private attachESCListener(): void {
-    this.documentESCListener = this.renderer.listen('document', 'keydown', event => {
-      if (event && event.keyCode === ESC) {
-        this.ifOpenService.open = false;
-      }
-    });
+    if (!this.popoverOptions.ignoreGlobalESCListener) {
+      this.documentESCListener = this.renderer.listen('document', 'keydown', event => {
+        if (event && event.keyCode === ESC) {
```
Shijir: I believe `KeyboardEvent.keyCode` is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Recommended alternative is: `KeyboardEvent.key`

jeeyun: Updated!

Jinnie: Good to know, thanks Shijir. I also switched to key (from code) in my directive, though for different reasons.

jeeyun:
Hi @Shijir the issue you mention re:scroll is  an existing issue. There is a separate ticket for it: https://github.com/vmware/clarity/issues/3622.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
```diff
@@ -109,11 +99,26 @@ export class DropdownFocusHandler implements FocusableItem {
...
+
+    // whether root container or not, tab key should always toggle (i.e. close) the container
+    this.renderer.listen(el, 'keydown.tab', event => this.ifOpenService.toggleWithEvent(event));
+
+    if (this.parent) {
+      // if it's a nested container, pressing esc has the same effect as pressing left key, which closes the current
+      // popup and moves up to its parent. Here, we preventDefault and stop propagation so that the parent container
+      // doesn't receive the esc keydown
+      this.renderer.listen(el, 'keydown.esc', event => {
+        this.focusService.move(ArrowKeyDirection.LEFT, event);
```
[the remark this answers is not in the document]
jeeyun: I think you are right. Removing preventDefault.

As for your other comment, I think this is ok to stop propagation because the dropdown is a complex ARIA-widget for which we are defining and providing expected keyboard navigation behavior.

src/clr-angular/popover/dropdown/dropdown-menu.ts
```diff
@@ -57,6 +57,7 @@ export class ClrDropdownMenu extends AbstractPopover implements AfterContentInit
       this.popoverPoint = Point.LEFT_TOP;
     }
     this.popoverOptions.allowMultipleOpen = true;
+    this.popoverOptions.ignoreGlobalESCListener = true;
```
hippee-lee: So drop down bypasses the ESC listener in AbstractPopover and implements its own event listeners. This makes sense to me. 👍🏽

src/clr-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -41,6 +43,20 @@ export class ClrDropdownItem implements AfterViewInit {
     return this.focusableItem.disabled;
   }
 
+  /*
+   * @deprecated: the presence of this attribute makes it not-focusable in IE11. Use [clrDisabled] input instead.
```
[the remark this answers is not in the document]
jeeyun: Yes, so I'm not deprecating yet but marking them for deprecation. Perhaps I should make that more evident in my comment. Will update.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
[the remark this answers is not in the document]
Shijir: The case where it could be problematic is that when a user opens our dropdown menu and something like dialog, modal, or toast pops up asynchronously that normally gets closed by global `esc` event would be interfered with this stopPropagation.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
```diff
@@ -92,11 +83,10 @@ export class DropdownFocusHandler implements FocusableItem {
   set trigger(el: HTMLElement) {
     this._trigger = el;
     this.renderer.setAttribute(el, 'id', this.id);
+
     if (this.parent) {
-      // The root trigger needs to be in the tab flow, but nested ones are removed like any other dropdown item.
-      this.renderer.setAttribute(el, 'tabindex', '-1');
+      this.renderer.listen(el, 'keydown.arrowright', event => this.ifOpenService.toggleWithEvent(event));
```
Shijir: We need to remove these listeners to prevent DOM leakages.

jeeyun:
Leaving a note on usage of `stopPropagation` in this PR:

Concern was raised on using `stopPropagation` on the `ESC` key while navigating inside the dropdown component. The specific use case for why this could be problematic is as follows:

1. User is navigating through a dropdown that has nested menus and is currently focused on a submenu item
2. An async operation introduces a dom element that requires attention (modal, toast, alert, etc)
3. Said element normally responds to ESC key and closes it
4. Because the dropdown element had focus, it captured and stopped propagating the event and will therefore not receive the ESC key event

The above use case assumes that the disruptive dom element introduced doesn't move the focus to itself when it appears. This pattern not accessible (especially for screen reader users) and therefore not recommended, and is not the behavior of Clarity components. 

After discussion with team, I'm making the decision to not address the use case addressed here for now. A possible solution involves introducing additional complexity at the cost of maintenance. This is not to say that we may not change this in the future. Just documenting current decision.

src/clr-angular/popover/dropdown/providers/dropdown-focus-handler.service.ts
jeeyun: Added unlisteners to this service and `focus-service.ts` class. This class's onDestroy method will call both.

PR #3663, opened 2019-08-02
feat: added custom properties test page to dev app

mathisscott:
• created a test page where we can prove custom properties work as expected in IE and Edge
• created a test web component to prove that custom properties are piercing the shadowDOM as expected in IE and Edge
• proof-of-architecture for both custom properties and lit element
• updated dev-app with custom webpack config to transpile lit element to es5 (it's only delivered in es6)
• added a few npm scripts to help with validating clr-base builds and testing in IE
• added package.json to test WC so that it could be loaded independently
• updated dev app files to render web components

Tested in:
✔︎ Chrome
✔︎ Firefox
✔︎ IE11
✔︎ Edge

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [x] Build related changes
* [ ] CI related changes

I added a custom properties page to the dev app.

Issue Number: N/A

* [ ] Yes
* [x] No

## Other information

Like everything else with base ATM, these are largely internal-facing changes.


src/clr-base/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "main": "index.js",
+    "types": "index.d.ts",
+    "repository": {
+        "type": "git",
+        "url": "ssh://git@git.eng.vmware.com/clarity.git"
+    },
+    "peerDependencies": {
+        "@webcomponents/shadycss": "1.9.1",
+        "@webcomponents/webcomponentsjs": "2.2.10",
+        "lit-element": "2.2.0"
```
hippee-lee: question: will `@clr/base` need to keep up/release major versions with lit-element versions, in a similar way that `@clr/angular` matches `@angular/*`?

src/clr-base/src/lib/utils/helpers/exists.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+describe('Functional Helper: ', () => {
+  describe('exists() ', () => {
+    beforeEach(function() {
+      (window as any).test = {
+        levelOne: {
+          levelTwo: {
+            levelThree: {
+              notTrue: false,
+              notStrung: '',
```
hippee-lee: `notStrung`? intentional or a typo for notString?

src/dev/src/app/custom-props/custom-props.demo.html
```diff
@@ -0,0 +1,23 @@
...
+  <p>The button below should change the color of the header above (outside the ShadowDOM) and the custom element at the bottom of the page (inside the ShadowDOM).</p>
+
+  <p><button type="button" class="btn btn-outline" (click)="cycleThemes()">Cycle Theme</button></p>
+</section>
+
+<hr style="margin: 1rem 0">
+
+<section>
+  <clr-wc-element title="Secret Message?">
+    Try Elm. It's fun!
```
hippee-lee: 🤣🏆

tsconfig.json
```diff
@@ -16,7 +16,8 @@
       "@clr/angular": ["src/clr-angular"],
       "@clr/icons": ["src/clr-icons"],
       "@clr/icons/shapes/*": ["src/clr-icons/shapes", "src/clr-icons/shapes/*"],
-      "@clr/base": ["src/clr-base", "src/clr-base/*"]
+      "@clr/base": ["src/clr-base"],
+      "@clr/base/*": ["src/clr-base", "src/clr-base/*"]
```
hippee-lee: Should there be two of these?

src/clr-base/package.json
coryrylan: `@clr-base` does not have to keep in sync with lit-element. We define what version we want to use. You can have multiple versions in the same app, it gets bundled into the module and webpack can dedupe versions at the application level. Ideally though best practices we should stay on latest the same way we do with Angular.

tsconfig.json
coryrylan: Yes, this allows TS to map to the root `@clr/base` path as well as defined deep paths.
This is the default the CLI also generates. If we move `clr-angular` to follow the CLI practice of pointing to dist at dev time it would be similar.

src/dev/src/app/custom-props/custom-props.demo.module.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { CommonModule } from '@angular/common';
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { ClarityModule } from '@clr/angular';
+
+import { CustomPropsDemo } from './custom-props.demo';
+import { ROUTING } from './custom-props.routing';
+
+import '@clr/base/src/lib/components/clr-wc-element/clr-wc-element';
```
coryrylan: I think this import path can be simplified based on the path map setup to

`import '@clr/base/clr-wc-element';`

If not something might not be quite right on the path mapping. Small nit we can fix this later as well once we have more components.

src/dev/src/app/custom-props/custom-props.demo.ts
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component } from '@angular/core';
+import { runCssVarsPolyfill } from '@clr/base/src/lib/utils/css-vars/css-vars';
```
coryrylan: I think this is the same as above, if this doesn't work it's because it's probably missing from the `index.ts` and `public_api.ts` export. Again minor nit we can fix later.

`import { runCssVarsPolyfill } from '@clr/base';`

src/dev/src/main.ts
```diff
@@ -11,6 +12,8 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
 
+import { runCssVarsPolyfill } from '@clr/base/src';
```
coryrylan: Path nit, does `import { runCssVarsPolyfill } from '@clr/base';` not work? If not we can look into the path mapping stuff in a later pr.

src/clr-base/src/lib/utils/helpers/exists.spec.ts
mathisscott: Having fun with unit tests means writing unit tests is fun! It's intentional.

custom-webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+const webpack = require('webpack');
+const pkg = require('./package.json');
+const path = require('path');
+
+module.exports = config => {
+  /**
+   * We need to add a rule to compile the lit-element module (which is
+   * shipped in es6 format) to es5
```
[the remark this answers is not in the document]
mathisscott: Will do.

src/clr-base/src/lib/components/clr-wc-element/clr-wc-element.ts
```diff
@@ -0,0 +1,70 @@
...
+    `;
+  }
+
+  toggle() {
+    this.visible = !this.visible;
+    this.dispatchEvent(new CustomEvent('visibleChange', { detail: this.visible }));
+  }
+}
+
+customElements.define(TAG_NAME, ClrWcElement);
```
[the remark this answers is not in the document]
mathisscott: The first reason is legibility. The second was our discussions about dynamic prefixing. I'll remove this for now because it feels over-optimized and won't scale when we add more components.

PR #3666, opened 2019-08-05
fix: clr-base import paths / build

coryrylan:
Aligns clr-base import paths to mimic final library build output. This moves the library code up a directory to a flat structure for the path mapping.

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

Current paths do not match the output for final build.

Issue Number: N/A

## What is the new behavior?
The paths and directories match the final build output for publishing.

* [ ] Yes
* [x] No


src/dev/src/app/custom-props/custom-props.demo.module.ts
```diff
@@ -12,7 +12,7 @@ import { ClarityModule } from '@clr/angular';
 import { CustomPropsDemo } from './custom-props.demo';
 import { ROUTING } from './custom-props.routing';
 
-import '@clr/base/components/clr-wc-element';
+import '@clr/base/clr-wc-element';
```
mathisscott: You don't want the components to live in their own folder like helpers and utils?

coryrylan: Updated to use `/components` directory and path

src/clr-base/utils/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "index.ts"
+    }
```
coryrylan: Utils has its own entry file now. This will be needed for importing shared code between components when added in the future. Examples:
 
https://github.com/ng-packagr/ng-packagr/pull/685
https://github.com/coryrylan/lithium/tree/master/projects/lithium

src/clr-base/index.ts
```diff
@@ -4,4 +4,5 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-export * from './lib/utils/css-vars/css-vars';
+export * from './utils';
+export * from './components/clr-wc-element/clr-wc-element';
```
coryrylan: Main entry point exports all entry points (utils + components). This will allow users to import just individual components or all components.

```typescript
// import single component
import '@clr/base/components/clr-wc-element';

// import all components
import '@clr/base';
```

src/clr-base/utils/css-vars/css-vars.spec.ts
```diff
@@ -4,26 +4,25 @@
...
 
 describe('Utilities - CssVarsPolyfill ', () => {
   describe('runCssVarsPolyfill() ', () => {
     it('calls cssVars() with default config', () => {
-      const setPropertyTypeSpy = jasmine.createSpy('setPropertyTypeSpy');
-      spyOnProperty(ponyfill, 'cssVars').and.returnValue(setPropertyTypeSpy);
+      spyOn(ponyfill, 'cssVars');
       runCssVarsPolyfill();
-      expect(setPropertyTypeSpy).toHaveBeenCalledWith(cssVarsDefaultConfig);
+      expect(ponyfill.cssVars).toHaveBeenCalledWith(cssVarsDefaultConfig);
```
coryrylan: Not sure why but after converting to TS the `spyOnProperty` would no longer work.

angular.json
```diff
@@ -253,8 +253,8 @@
     "clr-base": {
       "projectType": "library",
       "root": "src/clr-base",
-      "sourceRoot": "src/clr-base/src",
-      "prefix": "clrx",
+      "sourceRoot": "src/clr-base",
+      "prefix": "clr-wc",
```
mathisscott: ❤️

src/clr-base/index.ts
mathisscott: 🏆

src/clr-base/utils/css-vars/css-vars.spec.ts
mathisscott: Probably because it's being treated as a module. the SpyOnProperty was needed because of weirdness with jasmine when a file exported a function.

src/clr-base/utils/package.json
mathisscott: 👍

src/clr-base/index.ts
Shijir: QQ: don't we need to support the following? Because one WC could have multiple files that could be exported as a module?
```
export * from './components/clr-wc-element';
```

coryrylan:
I'll have another small fixup to add to this, pointing the dev app to the build output the cli produces `/dist/clr-base` does not work properly. I have some more work to do here.

src/clr-base/package.json
```diff
@@ -3,20 +3,23 @@
     "version": "2.0.2",
     "description": "Custom Elements for Clarity",
     "homepage": "https://clarity.design",
+    "sideEffects": [
+      "./**/*clr-wc-element.js"
```
coryrylan: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

When we start to add additional components we can add them like this:
https://github.com/coryrylan/lithium/blob/master/projects/lithium/package.json#L4

This will make sure random utilities is not pulled into the bundles. Example the ie ponyfills. The side effects will allow the bundler to only include everything in the given file if imported. We only want side effects in component/element imports so the element will auto register.

PR #3669, opened 2019-08-05
fix(tooltip): adds a11y to tooltip

hippee-lee:

## PR Checklist

## PR Type

This change improves the way screen readers handle tooltips in supproted browsers.
    
1. Adds a tooltip id provider to utilize aria-described by on triggers that associate with content id attributes
2. Adds proper rols to tooltip triggers (button) and content (tooltip)
3. Adds an aria-label to triggers (handles screen reader support for IE11
4. Updates clr-icons to handle setting the button role for this use case
5. Tooltip specs are updated to use the all.spec.ts pattern and the new spec utilities (got rid of some TODO's in spec files)

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [x] clarity.design website / infrastructure changes

Screen readers currently do not read descriptions of the tooltip triggers. 

Issue Number: 3525

## What is the new behavior?
Supported screen readers read and describe the tooltip triggers and associated content in supported browsers. 

* [ ] Yes
* [x] No


src/clr-angular/popover/tooltip/tooltip-trigger.ts
```diff
@@ -1,14 +1,34 @@
...
 
-@Directive({ selector: '[clrTooltipTrigger]', host: { '[attr.tabindex]': '0', '[class.tooltip-trigger]': 'true' } })
+@Directive({
+  selector: '[clrTooltipTrigger]',
+  host: {
+    tabindex: '0',
+    '[class.tooltip-trigger]': 'true',
+    '[attr.aria-describedby]': 'ariaDescribedBy',
+    '[attr.aria-label]': 'tooltipLabel',
+    role: 'button',
```
Shijir: Isn't `role: 'tooltip'` a better fit for this?

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
```diff
@@ -0,0 +1,22 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { Subject } from 'rxjs';
+
+@Injectable()
+export class TooltipIdService {
```
Shijir: Do we really need a service for this tooltip id?

src/clr-angular/popover/tooltip/tooltip-trigger.ts
hippee-lee: Not for the trigger. tooltip needs to be on the content.

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
hippee-lee: Yes, we need to connect triggers (aria-describedby) with the id of tooltip content.

src/clr-angular/popover/tooltip/tooltip-trigger.ts
Shijir: Ah I see.

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
Shijir: Sorry. I meant to ask why we need an observable for the tooltip id in this service? Can't we have `@Inject(UNIQUE_ID) public tooltipId: string` in the `clr-tooltip` component? So that both tooltip trigger and content can access the same uniq id?

hippee-lee: If the consumer is using `*clrIfOpen` the content `id` will change when the content is destroyed and re-created.

Shijir: Yes, but that's true because we are injecting the unique id provider in the tooltip-content. If we inject the unique ide provider in the tooltip component itself which contains both trigger and content, there's no need to keep creating new ids, no?

hippee-lee: Correct. I followed the same pattern that I used in signposts. I may have had a reason for doing it that way in there but don't recall offhand.

If you feel this should change, I'm happy to re-work it and update the testing.

Shijir: I see. I think that could be better because we could get rid of this service. That means we don't have to subscribe to id observable, and the code would be a bit simpler.

hippee-lee: We still have to inject the id into the child components though. 
Wouldn't that add similar code to inject the parent, get the id and set it in the trigger/content?

I like having the id in a service b/c it doesn't seem like ID belongs on the ClrTooltip. The ClrTooltip doesn't use it in the templates or in its code.

Shijir: You only need to provide it in the tooltip component and only inject it into the tooltip content component and trigger directive. The id doesn't have to change every time content pane opens. As such, we don't need observable there. This could be just simply a string id provided via UNIQUE_ID_PROVIDER. So we don't really need another service to maintain and test.

src/clr-angular/popover/tooltip/tooltip-content.spec.ts
```diff
@@ -1,67 +1,71 @@
...
-  // FIXME: this doesn't even run yet, we don't have an all.spec for tooltips.
   describe('TooltipContent component', function() {
-    let context: TestContext<ClrTooltipContent, SimpleTest>;
+    spec(ClrTooltipContent, SimpleTest, ClrTooltipModule, { providers: [IfOpenService, UNIQUE_ID_PROVIDER] });
 
-    beforeEach(function() {
-      context = this.create(ClrTooltipContent, SimpleTest, [IfOpenService]);
-      context.getClarityProvider(IfOpenService).open = true;
-      context.detectChanges();
+    beforeEach(function(this: TooltipContext) {
```
gnomeontherun: Need a test for setting the `id` and `role` values?

src/clr-angular/popover/tooltip/tooltip-trigger.ts
```diff
@@ -1,14 +1,32 @@
...
+    '[class.tooltip-trigger]': 'true',
+    '[attr.aria-describedby]': 'tooltipId',
+    '[attr.aria-label]': 'tooltipLabel',
+    role: 'button',
+  },
+})
 export class ClrTooltipTrigger {
-  constructor(private ifOpenService: IfOpenService) {}
+  public ariaDescribedBy;
+  private subs: Subscription[] = [];
```
gnomeontherun: I see no subscriptions here? Remove this and the onDestroy

src/clr-angular/popover/tooltip/tooltip.spec.ts
```diff
@@ -35,3 +20,27 @@ describe('Tooltip component', function() {
...
+export default function(): void {
+  describe('Tooltip component', function() {
+    let context: TestContext<ClrTooltip, SimpleTest>;
+
+    addHelpers();
+    beforeEach(function() {
+      context = this.create(ClrTooltip, SimpleTest);
+    });
+
+    describe('TypeScript API', () => {
```
gnomeontherun: What about a test case where I set an `id` manually, check does it get pulled through

src/clr-angular/popover/tooltip/tooltip-content.ts
```diff
@@ -19,22 +20,25 @@ const SIZES: string[] = ['xs', 'sm', 'md', 'lg'];
     `,
   host: {
     '[class.tooltip-content]': 'true',
-    // I'm giving up on animation, they did not work before and will not work now.
-    // Too many conflicts with Clarity UI.
     '[style.opacity]': '1',
+    role: 'tooltip',
+    '[id]': 'tooltipId',
```
gnomeontherun: Issue here is that it will clobber someone's custom ID. We should still allow them to be set by the dev.

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
hippee-lee: I moved it to `tooltip.ts`

src/clr-angular/popover/tooltip/tooltip-content.ts
hippee-lee: It seems like we still need a tooltip ID service here. In the custom case it's needed to communicate that there is a custom ID on the content and communicates that to the trigger.  Does that sound right?

src/clr-angular/popover/tooltip/tooltip-trigger.ts
hippee-lee: I had to add the subscription back into the trigger. The component will keep this.

src/clr-angular/popover/tooltip/tooltip-content.spec.ts
hippee-lee: Added input for id. Will add tests that cover use cases.

src/clr-angular/popover/tooltip/tooltip.spec.ts
hippee-lee: Added an input for id that is used to associate content id <---> trigger aria-describedby
If a custom string is passed thats what is used. If the id value is set on the input with {undefined|null|''} an empty string is used, and if no id is set at all clarity handles setting a unique id for the tooltip complex.

src/clr-angular/popover/tooltip/tooltip-content.ts
jeeyun: Could consider using the existing `AriaService` for this. I think it isn't used widely within our library but I do think it's a good fit given that we have to establish relationship between two elements in a few places in our components.

src/clr-angular/popover/tooltip/tooltip-trigger.spec.ts
```diff
@@ -1,43 +1,66 @@
...
+        expect(this.clarityElement.getAttribute('tabindex')).toEqual('0');
+      });
+    });
   });
 }
 
 @Component({
   template: `
-        <span clrTooltipTrigger>
+        <span clrTooltipTrigger [clrTooltipLabel]="label">
```
jeeyun: Why is it necessary to create/use this `[clrTooltipLabel]` as opposed to having them add the aria-label directly like below?
```
<span clrTooltipTrigger aria-label="[some custom label]">
...
</span>
```

src/clr-angular/popover/tooltip/tooltip-content.ts
hippee-lee: I looked into `AriaService` but ultimately stayed with the unique id provider. I left a comment should we see benefit in centralizing the id for establishing relationships in the future.

src/clr-angular/popover/tooltip/tooltip-trigger.spec.ts
hippee-lee: Good catch. I removed this input.

src/clr-icons/clr-icons-element.ts
```diff
@@ -67,10 +71,19 @@ ClarityIconElement.prototype.connectedCallback = function() {
   // So we could check whether the attribute values really changed or not.
   // If not, we don't need to execute the same codes again.
 
-  // We want to hide the custom element from screen readers but allow the svg/img content to still be read inline
-  // Adding role=none allows the screen reader to skip the custom element as if it were a div or span.
-  // https://www.scottohara.me/blog/2018/05/05/hidden-vs-none.html
-  this.setAttribute('role', 'none');
+  if (this.hasAttribute('role')) {
```
Shijir: Wait, why do we need this? If the user places this button role on the icon, that should be it. Why do we set it again on the icon again here?

hippee-lee: adding the role of 'none' hides this from the screen readers.
Even with role='button on the clr-icon element screen readers will ignore this.

Shijir: You have clrTooltipTrigger that will add a role value to the icon here as you defined in the host. I don't understand what you are trying to do here. If it has to have the 'button' role, the directive will place that, no need to change anything inside the clr-icon implementation.  

When or why should it have the role 'none'?

hippee-lee: This overrides anything that the directive does. Adding role to a clr-icon will always get whatever this code says. In my investigation I found that this is the reason screen readers were ignoring it.

Do you have suggestions for informing screen readers with a role of button and keeping the role of none here?

hippee-lee: Role of none was added here to address other a11y issues: https://github.com/vmware/clarity/pull/3448/commits/4d3d1a2aa5707b426d998c3de780986698bffa32

Shijir: That's by default, it has 'none' role. Here, the tooltip trigger will override that `none` role with `button`. That's it. You don't have to do anything in `clr-icon` element.

hippee-lee: No, I tried. It always gets a role of none no matter what the directive code does.

Shijir: Whatever role the directive assigns to the icon, it will stay on `clr-icon` https://stackblitz.com/edit/clarity-light-theme-v2-4y5ayd

src/clr-icons/clr-icons-element.ts
```diff
@@ -57,6 +57,10 @@ ClarityIconElement.prototype._setIconSize = function(size: string) {
   }
 };
 
+ClarityIconElement.prototype._setRoleAttr = function(role: string) {
```
Shijir: we don't need this one now too. It's not called anywhere now

hippee-lee: Right, I forgot about that one. WIll remove.

src/clr-icons/clr-icons-element.ts
hippee-lee: Resolved this by switching the directive to attribute binding instead of setting the property directly. That eliminates the need for this change. Will remove.

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { Subject } from 'rxjs';
```
Shijir: 😱 breaching the DRY principle :trollface: 

nit: we could import them from the same `rxjs`?

src/clr-angular/popover/tooltip/tooltip-content.ts
```diff
@@ -19,25 +21,32 @@ const SIZES: string[] = ['xs', 'sm', 'md', 'lg'];
     `,
   host: {
     '[class.tooltip-content]': 'true',
-    // I'm giving up on animation, they did not work before and will not work now.
-    // Too many conflicts with Clarity UI.
     '[style.opacity]': '1',
+    role: 'tooltip',
```
Shijir: nit: do you mind changing it to`'[attr.role]': '"tooltip"'` as well?

src/clr-angular/popover/tooltip/tooltip-trigger.ts
```diff
@@ -3,30 +3,28 @@
...
   private subs: Subscription[] = [];
-  constructor(@Inject(UNIQUE_ID) public tooltipId: string, private ifOpenService: IfOpenService) {}
-
-  // This must be supplied by consumers
-  // This must be unique (among tooltips)
-  // This must be human readable e.g 'Tooltip 1' or 'Email input field tooltip`
-  @Input('clrTooltipLabel') tooltipLabel;
+  constructor(private ifOpenService: IfOpenService, private tooltipIdService: TooltipIdService) {
+    // The aria-described by comes from the id of content. It
+    this.subs.push(this.tooltipIdService.id.subscribe(idChange => (this.ariaDescribedBy = idChange)));
```
jeeyun: Just for clarity could you change `idChange` to `tooltipId`?

src/clr-angular/popover/tooltip/providers/tooltip-id.service.ts
hippee-lee: Sure thanks for pointing it out, I'll combine them into one. IntelliJ handles this and keeps them hidden in collapsed comments.

src/clr-angular/popover/tooltip/tooltip-content.ts
hippee-lee: Yes. Will change.

src/clr-angular/popover/tooltip/tooltip-trigger.ts
hippee-lee: Yes. Will change.

Issue #3497, opened 2019-06-24
Alerts in Wizards are read immediately when not urgent


gnomeontherun:
## Describe the bug

When a wizard has an alert in the page, it is read immediately and it is a simple status alert.

## How to reproduce

https://stackblitz.com/github/clane/clarityTest2/tree/master/clarity-test-app

- Using a screen reader, activate the WIZARD + ASYNC FORM VALIDATION button under the Wizards - Wizard with Async Form Validation example
- Notice that the next announcement from the screen reader is "This 
wiki article
 might help you answer the question."

## Expected behavior

The alert should not be read unless it is urgent.

Recommendation: 
Use polite live regions for status updates and ensure they are empty when the page first loads. Then dynamically populate them with messages as needed.

## Additional notes

This might be something we need to address with documentation as well for alerts, as they can be placed anywhere with this same type of problem. This is really not a Wizard issue, but Alerts not always being needed to read out to screen readers.

hippee-lee:
Modify ClrAlert to default to polite live regions and provide an input to set the live region to urgent.

bdryanovski:
The issue here is that the `clrAlert` use `role="alert"` 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

This automatically set the `aria-live="assertive"` so when trying to change it with `aria-live="polite"` it has no effect at all.

So I got two solutions on the issue here.

* Remove `role="alert"` and handle it with `aria-live="setAriaState"` and input to change its value. 
* Add input to add or remove `role="alert"` if needed. Basicly trigger between `aria-live="off" and `aria-live="assertive"` - there is no option for "polite".

Also when using `role="alert"`.

> If an element already has role="alert" and is initially hidden using CSS, making it visible will cause the alert to fire as if it was just added to the page. This means that an existing alert can be "reused" multiple times.  

When the Alert is `danger` or `warning` could be useful to have the `role="alert"` but this still depends on the use case.

I'm more for removing the `role="alert"` and going with `aria-live` - let the developer handle this in whatever manner the application required it.

Open for discussion 🤓

bdryanovski:
Version A:

```html
// Wait it's turn
<clr-alert ... clrPolite></clr-alert>

// Mimic role="alert" and interept the voiceover
<clr-alert ... clrAssertive></clr-alert>

// Don't do anything.
<clr-alert ... clrOff></clr-alert> 
```

Version B:

```html
[...]

PR #3676, opened 2019-08-07
fix: alerts in wizzard are read immediately when not urgent

bdryanovski:
When a wizard has an alert in the page, it is read immediately and it is a simple status alert.

`clrAlert` has `role="alert" this automatically set the `aria-live="assertive"` so when trying to change it with `aria-live="polite"` it has no effect at all.

## PR Checklist

## PR Type

* [x] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3497

## What is the new behavior?

`role="alert"` was replaced with `aria-live` for more control on how the event has to be fired.

```html
// Wait your turn
<clr-alert ... clrPolite></clr-alert>

// Same as role="alert" 
<clr-alert ... clrAssertive></clr-alert>

// Don't do anything.
<clr-alert ... clrOff></clr-alert> 
```

* [ ] Yes
* [x] No

## Other information

Close: #3497

### DIFF
--- a/src/clr-angular/emphasis/alert/alert.spec.ts
+++ b/src/clr-angular/emphasis/alert/alert.spec.ts
@@ -19,7 +19,11 @@ const CLOSE_ARIA_LABEL = 'Close Test Alert';
             [clrAlertClosable]="isClosable"
             [(clrAlertClosed)]="closed"
             [clrAlertAppLevel]="isAppLevel"
-            [clrCloseButtonAriaLabel]="closeAriaLabel">
+            [clrCloseButtonAriaLabel]="closeAriaLabel"
+            [clrOff]="clrOff"
+            [clrAssertive]="clrAssertive"
+            [clrPolite]="clrPolite"
+            >
             <div class="alert-item">
                 <span class="alert-text">
                 {{alertMsg}}
@@ -38,6 +42,11 @@ class TestComponent {
   closed: boolean = false;
   isAppLevel: boolean = false;
   closeAriaLabel: string = CLOSE_ARIA_LABEL;
+
+  // AriaLive
+  clrOff: boolean = false;
+  clrAssertive: boolean = false;
+  clrPolite: boolean = false;
 
   alertMsg: string = 'This is an alert!';
 }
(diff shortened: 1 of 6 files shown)


src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -42,6 +43,28 @@ export class ClrAlert {
...
+   * Polite is default if non is passed.
+   *
+   * In the case of setting all of them to true. Assertive will be used.
+   *
+   */
+  get setAriaLive(): string {
+    if (isBooleanAttributeSet(this.assertive)) {
+      return 'assertive';
+    }
+    if (isBooleanAttributeSet(this.off)) {
```
mathisscott: Right now, this appears to be the only tested conditional path in `setAriaLive`. We need more tests.

bdryanovski: You are right :) on it...

src/clr-angular/emphasis/alert/alert.spec.ts
```diff
@@ -129,12 +136,25 @@ export default function(): void {
...
       expect(myAlert.getAttribute('aria-live')).toBe('off');
     });
 
+    it('should have an aria-live value of assertive when apply clrAssertive', () => {
+      const myAlert: HTMLElement = compiled.querySelector('.alert');
+      fixture.componentInstance.clrAssertive = true;
+      fixture.detectChanges();
+      expect(myAlert.getAttribute('aria-live')).toBe('assertive');
+    });
+
```
mathisscott: One last bit... we need to test the cascade (Assertive, Off, Polite). Can we add tests that set all three to true and validate that `aria-live` is `assertive` and also a test that sets off and polite to true and tests that `aria-live` is `off`?

One day these tests will save your bacon when somebody comes in here and wrecks your inputs with an enum or decides to make the logic into a one-liner to prove their haxxor skillz.

bdryanovski: LOL

bdryanovski: BTW, I had one but did not commit it - thinking this is too much ... My initial test was based on the priority testing them .. ;)

src/clr-angular/emphasis/alert/alert.spec.ts
```diff
@@ -121,21 +130,46 @@ export default function(): void {
...
+    it('should have an aria-live value of assertive when apply clrAssertive', () => {
+      const myAlert: HTMLElement = compiled.querySelector('.alert');
+      fixture.componentInstance.clrAssertive = true;
+      fixture.detectChanges();
+      expect(myAlert.getAttribute('aria-live')).toBe('assertive');
+    });
+
+    it('should fallow the aria-live priority when all of them are set', () => {
```
jeeyun: typo: fallow -> follow

bdryanovski:
This pattern is used in previews component like clrSpinner, clrPogressBar. I could not find the discussion right now. But the short story is this will help when Web Components are ready. 

So this is the reason for that. Some of the components could get a little cleaner but the idea is to have attribute-based API as much as close to the native attribute that you could get with Web Components.

It behaves the same as `<input disabled=disabled />` 

https://github.com/vmware/clarity/pull/3493
https://github.com/vmware/clarity/pull/3533

jeeyun:
Thank you for the explanation @bdryanovski. 👍I missed the discussions on that, so will sync up offline on those decisions.

PR #3682, opened 2019-08-08
fix: improve screen reader behaviors on stack-view and modal

### LINKED ISSUE #3566: Stack view edit mode accessibility issues
## Describe the bug

The edit mode of stack view has several defects 

- The form fields are not labeled (should probably use new forms inputs)
- The stack-block-changed triangle needs a unique textual equivalent (when you've edited a field) Screen reader users will not hear what these shapes are meant to convey when they read them. It is important that part of the textual equivalent is unique, e.g., 'record 1 has changed' and 'record 2 has changed'.
- Close button doesn't have an accessible name
- Screen readers don't hear anything when edit dialog appears

## How to reproduce

- Using a screen reader, go to the 'Stack View Edit' example
- Activate the Edit Button
- Observe you don't hear anything about the now opened edit dialog.
- Read any of the input fields and try to determine the purpose of any field.
- Edit a field
- Read any content or control visually associated with the stack-block-changed triangle. Notice that you do not hear what his shape is meant to convey. You would expect to hear something like 'One of Jon Doe's records has changed'. 
- Close button in edit mode is not accessible (may be solved by modal refactorings)

## Expected behavior

- Use aria-label to provide a unique accessible name for each input field.
- Use aria-label to provide a unique textual equivalent for the stack-block-changed triangle. The textual equivalent should reflect the context of each stack.
- While a stack is collapsed, Provide an 'id' attribute with the stack-block-changed triangle to be used with aria-describedby on the adjacent stack buttons. This will allow the stack-block-changed triangle(s) to provide the intended information to screen reader users.
- Use aria-label to provide and accessible name such as 'close stack edit mode dialog'
- Move focus to the modal title on open

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

This is a combination of several issues related to edit mode, some of which may be resolved in the modal itself already.

gnomeontherun:
This addresses screen reader issues with the stack-view, where labels
and inputs need to be linked with describedby, adds screenreader text
for indicating if a block has been changed (currently only a CSS
triangle shape), and moves the focus into the modal title for editing.

closes #3566

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

* The stack view doesn't link the labels and input in edit mode unless the applications do it on their side.
* The stack view has a visual treatment for when a block has changed but it was not announced to screen readers.
* The focus remains on the trigger element for a modal open action. 

Issue Number: #3566 

## What is the new behavior?

* This will automatically link the label and input with an id/aria-describedby. 
* This adds offscreen text before the label to indicate the changed state.
* Focus is moved to the modal title when modal is opened (which is inside of the trap). 

* [ ] Yes
* [x] No

## Other information

There is a change where the modal title has the focus outline on open since it becomes focused. This is an expected side effect, but visually we may get backlash for this visual behavior. Need to validate this behavior fully.


src/clr-angular/modal/modal.ts
```diff
@@ -67,17 +72,10 @@ export class ClrModal implements OnChanges, OnDestroy {
...
   ) {}
 
-  get sizeClass(): string {
-    if (this.size) {
-      return 'modal-' + this.size;
-    } else {
-      return '';
-    }
-  }
-
```
jeeyun: was this something we haven't been using for a while and just getting cleaned up here?

mathisscott: yeah. we apparently moved away from interpolating that class string to do a `[class.modal-*]` boolean binding in modal.html

src/clr-angular/data/stack-view/stack-content-input.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Directive({
+  selector: 'clr-stack-content > input[type="text"]',
+  host: {
+    '[class.clr-input]': 'true',
+    '[attr.aria-describedby]': 'stackBlockService?.id',
+  },
+})
+export class ClrStackContentInput {
+  constructor(@Optional() public stackBlockService: StackBlockService) {}
```
mathisscott: Is there a path whereby `stackBlockService` would not be defined?

src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -84,16 +89,19 @@ export class ClrStackBlock implements OnInit {
...
     @Optional()
     private parent: ClrStackBlock,
+    public stackBlockService: StackBlockService,
+    @Inject(UNIQUE_ID) uniqueId: string,
     public commonStrings: ClrCommonStringsService
   ) {
     if (parent) {
       parent.addChild();
     }
+    this.stackBlockService.id = uniqueId;
```
Shijir: Shouldn't we use the same concept as we used in our tooltip component? First, we need to check user-defined id before setting out own unique id?

gnomeontherun: This sets the unique ID, and if the user sets one the input setter overrides it.

Shijir: Do we have the input setter for id in this component? We are setting the id only here in the constructor?

gnomeontherun: Look at the template and notice that the id is put into the `dt` element which is internalized. Therefore an ID provided on the stack block should not be moved into the child element.

src/clr-angular/data/stack-view/stack-content-input.ts
gnomeontherun: The problem was that I forgot Angular doesn't support child combinator selectors, so `clr-stack-block > input[type="text"]` was just turning into `input[type="text"]` and matching too many elements. I fixed this by making the directive selector unique but apps will need to add this directive to their inputs to get the behaviors to work properly.

src/clr-angular/data/stack-view/stack-block.ts
Shijir: Actually, I don't think we don't need to concern about user-defined ids. As the `dt` element is hidden in the stackblock's template.

src/clr-angular/data/stack-view/stack-block.service.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class StackBlockService {
+  public id;
```
Shijir: This is related to what I mentioned earlier. Users won't be able to set their own ids, as the element we set the unique id on is wrapped inside our component template. So If users don't need to set their own ids, we don't need this service the id.

As the stackblock component injects the unique id provider, all its children can inject and access the autogenerated id.

gnomeontherun: Good catch, I had bigger plans originally but this service doesn't need to exist.

src/clr-angular/modal/modal.ts
```diff
@@ -121,6 +119,8 @@ export class ClrModal implements OnChanges, OnDestroy {
   fadeDone(e: AnimationEvent) {
     if (e.toState === 'void') {
       this._openChanged.emit(false);
+    } else if (e.toState === 'false' && isPlatformBrowser(this.platformId) && this.modalTitle) {
```
Shijir: Hmm. So this is now dependent on fadeDone animation callback. But looks like we already had `openChanged` dependent on fadeDone though. I think we could make use of *TODO* comment here to remind us to re-visit this later and decouple these from animation callback.

gnomeontherun: Added a comment, though I don't recall having a discussion about not wanting to avoid animation callbacks so if we want to tackle that refactoring we should backlog a story about that.

PR #3694, opened 2019-08-12
docs: updating documentation language

### LINKED ISSUE #3412: Grid needs documentation about a11y limitation of using flex-order
Flexbox ordering is not acknowledged by screen readers, so we should make a note in our documentation that DOM order is the preferred way to ensure screen readers get content in the right order. There may be uses for the column ordering that wouldn't break accessibility, but we need to properly document this for consumers.

clane:
1. remove text supporting use of CSS order property, close #3412
2. fixed toggle docs language to reflect the need for visual labelling

Signed-off-by: Chris Lane <chlane@vmware.com>

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Documentation is incorrect regarding CSS order property usage and visual labelling for toggle buttons

Issue Number: #3412

## What is the new behavior?
NA

* [ ] Yes
* [x] No

### DIFF
--- a/src/website/src/app/documentation/demos/grid/grid.demo.html
+++ b/src/website/src/app/documentation/demos/grid/grid.demo.html
@@ -219,6 +219,14 @@
 
             <h3 id="column-ordering">Column Ordering</h3>
 
+            <clr-alert [clrAlertType]="'danger'" [clrAlertClosable]="false">
+                <clr-alert-item>
+                      <span class="alert-text">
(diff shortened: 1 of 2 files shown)


gnomeontherun:
Technically we aren't removing the flex-order option, we're just removing the documentation for it. I'd rather we put in a deprecation notice and explain why its not a good choice to use it, so people may learn about it better.

PR #3706, opened 2019-08-16
fix(dropdown): handle error when there are no items in dropdown to focus

### LINKED ISSUE #3610: Latest 1.1.5 clarity causes exception when focusing our dropdowns
## Describe the bug

We have extended the clr-dropdown in our component and with latest Clarity 1.1.5 we are getting exceptions when focusing the control. It was working with 1.1.4

## How to reproduce

Focus the control. Exception seen:

`
core.js:15724 ERROR TypeError: Cannot read property 'disabled' of undefined
    at SafeSubscriber._next (clr-angular.js:17470)
    at SafeSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.SafeSubscriber.__tryOrUnsub (Subscriber.js:194)
    at SafeSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.SafeSubscriber.next (Subscriber.js:132)
    at Subscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber._next (Subscriber.js:76)
    at Subscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (Subscriber.js:53)
    at MapSubscriber.push../node_modules/rxjs/_esm5/internal/operators/map.js.MapSubscriber._next (map.js:41)
    at MapSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (Subscriber.js:53)
    at ReplaySubject.push../node_modules/rxjs/_esm5/internal/Subject.js.Subject.next (Subject.js:47)
    at ReplaySubject.push../node_modules/rxjs/_esm5/internal/ReplaySubject.js.ReplaySubject.nextInfiniteTimeWindow (ReplaySubject.js:39)
    at DropdownFocusHandler.push../node_modules/@clr/angular/fesm5/clr-angular.js.DropdownFocusHandler.addChildren (clr-angular.js:17833)
`

## Expected behavior

No exception and working :)

## Versions

**App**

* Angular: [e.g. 7]
* Clarity: [e.g. 1.1.5]

**Device:**

* Browser [chrome 75]

## Additional notes

It's possible we are not using the dropdown properly, or this is a regression. We just updated Clarity with no other changes.

bdryanovski:
Fix issue with when there are no items into the dropdown the handle focus will throw an error when try to `moveTo` item. Right now the fix will just ignore this event and nothing will happen.

**Note**: We maybe need to handle this type of empty dropdowns with some message or animation that describe that this dropdown is empty and not stuck on the render stage.

This issue will close #3610 the last part with disabled that is not working. I had a talk with @jeeyun, we don't allow anymore to have disabled button into dropdowns based on the a11y spec. This is related to focusing on elements and so on.

The developer has to use `[clrDisabled]` if need to apply the `disabled="true"`
 
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3610 

* [ ] Yes
* [x] No

## Other information

Check to see if this error is present at v2 - if yes, cherry-pick it

### DIFF
--- a/src/clr-angular/utils/focus/focus.service.ts
+++ b/src/clr-angular/utils/focus/focus.service.ts
@@ -27,17 +27,15 @@ export class FocusService {
   listenToArrowKeys(el: HTMLElement) {
     // The following listeners return false when there was an action to take for the key pressed,
     // in order to prevent the default behavior of that key.
+    this._unlistenFuncs.push(this.renderer.listen(el, 'keydown.arrowup', event => !this.move(ArrowKeyDirection.UP)));
     this._unlistenFuncs.push(
-      this.renderer.listen(el, 'keydown.arrowup', event => !this.move(ArrowKeyDirection.UP, event))
+      this.renderer.listen(el, 'keydown.arrowdown', event => !this.move(ArrowKeyDirection.DOWN))
     );
     this._unlistenFuncs.push(
-      this.renderer.listen(el, 'keydown.arrowdown', event => !this.move(ArrowKeyDirection.DOWN, event))
+      this.renderer.listen(el, 'keydown.arrowleft', event => !this.move(ArrowKeyDirection.LEFT))
     );
     this._unlistenFuncs.push(
-      this.renderer.listen(el, 'keydown.arrowleft', event => !this.move(ArrowKeyDirection.LEFT, event))
-    );
-    this._unlistenFuncs.push(
-      this.renderer.listen(el, 'keydown.arrowright', event => !this.move(ArrowKeyDirection.RIGHT, event))
+      this.renderer.listen(el, 'keydown.arrowright', event => !this.move(ArrowKeyDirection.RIGHT))
     );
   }
 
@@ -51,6 +49,15 @@ export class FocusService {
   }
 
   moveTo(item: FocusableItem) {
+    /**
+     * Make sure that item is not undefined,
+     * This is safety net in the case that someone sometime decide to
+     * call this method without having FocusableItem.
+     */
+    if (item === undefined) {
+      return;
+    }
+
     if (this.current) {
       this.current.blur();
     }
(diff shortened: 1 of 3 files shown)


src/clr-angular/utils/focus/focus.service.ts
```diff
@@ -72,8 +81,7 @@ export class FocusService {
         // If performance ever matters for this, we can refactor using additional private methods.
         const nextObs = isObservable(next) ? next : of(next);
         nextObs.subscribe(item => {
-          this.moveTo(item);
-          return true;
+          return this.moveTo(item);
```
hippee-lee: Not a blocker, mabye a nit because it wasn't already here:

Does it make sense to also add the boolean return type here for clarity? Similar to how it was added to moveTo?

src/clr-angular/utils/focus/focus.service.ts
```diff
@@ -50,12 +50,21 @@ export class FocusService {
     this._unlistenFuncs.push(this.renderer.listen(el, 'keydown.enter', () => !this.activateCurrent()));
   }
 
-  moveTo(item: FocusableItem) {
+  moveTo(item: FocusableItem): boolean {
+    /**
+     * Make sure that item is not undefined
+     */
+    if (item === undefined) {
+      return false;
```
gnomeontherun: I'd like @jeeyun to take a look, but I'd think we'd prefer not to have `moveTo` return a value here. We just want to silently not move focus if there is nothing to focus on. Then you would just `return` and I think leave the rest the same.

jeeyun: I agree with you @bdryanovski that our current setup is pretty brittle. Thank you for adding the guard to make sure that the item is defined inside the `moveTo` function.

Even though I suggested earlier that we not introduce the boolean return value, I can see the value in it now. It's up to you if you want to make it return boolean and use that inside the `move` function as you did earlier in this PR. Sorry, I'm not trying to make you go back/forth! 😛 After reading your thought I'm agreeing that returning the boolean value 1. doesn't introduce breaking changes, 2. may add value in its usage. We cannot yet make the function private as you suggested because it's being used in one place outside of the class (in DropdownFocusHandler class).

On a side note, some of the keyboard navigation logic in dropdown seem a bit brittle in general. I'd like for us to review this in the long run to see if we'll keep the current setup or refactor to use the focus directive that we have written since then.

bdryanovski: I think that you are looking on some older version, cause at my last commit `move()` return only undefined.

And yes, I forgot to submit my comments, I saw it at the demo .. ;)

PR #3712, opened 2019-08-19
Applitools integration

gnomeontherun:
## PR Checklist

* [x] If applicable, have a visual design approval

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [x] CI related changes

Issue Number: N/A

## What is the new behavior?

This adds Applitools to our testing setup, and has 3 tests (Toggles, Checkboxes, Colors). It also runs on IE11, Firefox, Chrome, and Edge. It also supports light and dark.

* [ ] Yes
* [x] No

## Other information

This also should setup the GitHub integration where it will block a PR if there is a failure in the visual testing.


cypress/support/commands.js
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
```
hippee-lee: Everything is commented out in this file. Do we really want to add it if its not being used yet?

gnomeontherun: This is a file generated by cypress, I'm ok with removing it since we probably don't need it.

bdryanovski:
I think it will be good to add some kind of small intro documentation into the development guideline. 

* What you need to do to run the visual tests
* Short explanation on the newly added package.json commands
* What to do when Error X ...

gnomeontherun:
There is a test with an intentional change that needs to be reviewed before it can be approved.

gnomeontherun:
The reason for so many gemini test changes is simply the change in how the dev app main-container styles are changed. Prior to the change made in the dev app, gemini nor applitools could scroll and snapshot, but now it can. This means previously if the content was less than the browser size, the screenshot would often snap a much larger page anyways, but if the content was larger than the browser size, it would not snapshot the overflow content. Now it will crop or expand to the content of the page and snapshot.

cypress/integration/buttons.js
```diff
@@ -0,0 +1,58 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { checkEyes, setup } from '../util';
+
+const test = 'Buttons';
+
+// TODO: Enable these tests again after validation
+export function ButtonSpecs() {
+  xdescribe(test, () => {
```
hippee-lee: Should this suite be skipped?

cypress/integration/spec.js
```diff
@@ -0,0 +1,22 @@
...
+import { CheckboxesSpec } from './checkboxes';
+import { ColorSpec } from './color';
+import { ListsSpec } from './lists';
+import { TogglesSpec } from './toggles';
+
+// Organized this way to make one batch for all of the tests in Applitools,
+// otherwise it treats each file as a different batch and makes it hard to
+// see a single run as one unit.
+describe(`Clarity - ${Cypress.env('CLARITY_THEME')}`, () => {
+  // ButtonSpecs();
```
hippee-lee: This is commented out but xdescribed in the suite. Not sure what it should be. You may have covered this in dev share and I missed it.

scripts/load-vars.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+curl -g https://us-central1-clarity-design-system.cloudfunctions.net/applitools -o .env
```
hippee-lee: I know there isn't much we can do but 🙁

cypress/integration/buttons.js
gnomeontherun: Yes, I wanted to leave the test suite but not run it for now.

cypress/integration/spec.js
gnomeontherun: I just commented it out as well to be certain.

scripts/load-vars.sh
gnomeontherun: Its only meant to run on CI and this is the easiest script to write without writing a longer node script.

gnomeontherun:
> Still lgtm - one thought on the travis.yml. We used to be limited to 5 concurrent jobs but now its off in the settings so adding a 6th job isn't a big deal. If we want to pare down on jobs we could combine the gemini suites into 3 if we had to.

My idea is we'd shuffle a few around to get it down to 5 jobs again, but didn't want to do that yet. Was hopeful we could also remove some gemini tests too.

Issue #3577, opened 2019-07-02
Vertical nav focus management


gnomeontherun:
## Describe the bug

Vertical nav links do not navigate screen reader users to new content when the link is clicked and the page isn't a full page reload (Angular routing).

## How to reproduce

https://clarity.design/documentation/vertical-nav

- Using a screen reader, go to https://clarity.design/documentation/vertical-nav
- Activate of the links in the Basic Structure example
- Note that you do not hear any new content

## Expected behavior

We should create an example that moves focus to the new content when routing. Focus should move into the content.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

This might be a requirement for apps since we can't always know what the right choice is in this case.

Shijir:
closed by #3725

PR #3725, opened 2019-08-21
feat: add clrFocusOnViewInit for Vertical Nav focus management

Shijir:
## PR Checklist

## PR Type

This PR introduces `clrFocusOnViewInit`. This directive moves the focus to the element it's placed on when it's rendered in the DOM. And the first consumer component of this directive is the vertical navigation component.

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Screen readers don't register when new page content appears. As a result, screen reader users are not able to know what happens when they click on the router link and loads new content.

Issue Number: #3577 

## What is the new behavior?

This directive moves the focus to the element it's placed on when it's rendered in the DOM. So if now app developers use the directive together with the vertical nav, the screen readers able to read what kind of content is loading when the router changes.

* [x] Yes 
* [ ] No


src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
```diff
@@ -0,0 +1,56 @@
...
+})
+export class ClrFocusOnViewInit implements AfterViewInit {
+  constructor(private el: ElementRef, @Inject(PLATFORM_ID) private platformId: Object) {}
+
+  // negative tabindex as we shouldn't have this in the tab order
+  private _tabIndex: number = -1;
+
+  // set to true as we will intend to set the tab index by default
+  // if the element the directive is placed on has tabindex already, we will use that existing one.
+  private tabIndexSet: boolean = true;
```
hippee-lee: This will always be set. Does it sound more natural to call it `hasDefaultTabIndex` or `hasCustomTabIndex`?

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.tabIndexSet = false;
+    } else {
+      this._tabIndex = -1;
+    }
+  }
+
+  @HostListener('focusout')
+  onFocusout() {
+    // Once the focusout is registered, we should remove the tabindex we set.
+    // As it has already served its purpose during view initialization.
```
hippee-lee: I'm not sure if I understand this correctly. Should the element have tabindex='-1' after initialization and all of the lifecycles have run or does this remove it when the focusout event is executed? 

When I run the dev app I'm not seeing tabindex on the focused element after navigating back and forth (http://localhost:4200/vertical-nav/accessibility/beatles) but the test for adding -1 passes. I do see it when I refresh the page though so I'm not sure I understand how the directive should behave.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  it('sets focus on the directive element on view intialization', () => {
+    component.buttonElRef.nativeElement.focus();
+    expect(document.activeElement).toBe(component.buttonElRef.nativeElement);
+    component.displayNoExistingTabindex = true;
+    fixture.detectChanges();
+    expect(document.activeElement).toBe(component.focusOnItElRef.nativeElement);
+  });
+
+  it('removes set tabindex attribute after focus shifts another element', () => {
```
hippee-lee: Maybe change the name/description? This seems to be testing the default tabindex behavior, after focus shifts to anouther element.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    component.displayNoExistingTabindex = true;
+    fixture.detectChanges();
+    expect(component.focusOnItElRef.nativeElement.getAttribute('tabindex')).toBe('-1');
+    component.buttonElRef.nativeElement.focus();
+    fixture.detectChanges();
+    expect(component.focusOnItElRef.nativeElement.getAttribute('tabindex')).toBeNull();
+  });
+
+  it('keeps existing attribute after focus shifts another element', () => {
```
hippee-lee: Maybe change the name/description? This seems to be testing the custom tabindex behavior, after focus shifts to anouther element.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
Shijir: I like `hasDefaultTabIndex`. I will use it. Thanks for the suggestion.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
Shijir: The goal of the directive is to set the focus on its host element once the directive's view gets initialized. And once the user moves the focus to somewhere by clicking with her mouse or pressing the tab key, tabindex='-1' should be removed because our goal is not to make the element focusable, but to facilitate screen reader to catch new rendered content/title/header. 

As for not finding the tabindex, I think you are clicking with your mouse to inspect elements in the web inspection tool, the focus will be lost from the directive element so the `focusout` is being fired on the directive resulting in the removal of the tabindex.

hippee-lee: I see, yeah I can't seem to replicate the loss of tabindex so it must have been removing it when I inspected. 

The final visual difference I noticed here is that every element this is placed on seems to get a focus ring indicating it is focused, if this is only for screen readers should we handle mitigating the visual change or is it intended to have the focus ring.

clane:
Regarding the focus outline that occurs when the new directive is used, I agree with Shijir in that we should leave it in place. It will benefit sighted keyboard-only users and the consumer can remove the outline if they desire to.

clane:
Regarding the first sentence of the accessibility documentation, this directive is not for informing screen reader users of the purpose of each routing link. The directive allows consumers to choose the element they would like assistive technologies to read first when a new new view is loaded.

Issue #1018, opened 2017-06-06
Disable the row selection according to property of a record in Clarity datagrid.


rangagv:
Clarity datagrid having a multi selection, should have the ability to disable row selection according to some fields of the record. We have a use case where, all the objects are listed in the datagrid. User can select some records in the grids and perform some action on the selected item. There are some objects on which user does not have permission to perform the action. In this case, user should not have capability to select those particular rows. 

**Select one ...**  (check one with "x")
```
[ ] bug
[ ] feature request
[x] enhancement
```

### Expected behavior
If clarity datagrids have the capability to enable/disable the row selection according to a particular filed of object, user will have proper indication that he is able to perform some action on those objects.

### Actual behavior
Now there is no way, that programmatically we can disable the check box or radio box selection of the clarity datagrid row. Should have some thing like "clrDgSelected" for enable/disable the row.

### Reproduction of behavior

* Include a link to the reproduction scenario you created by forking one of the Clarity Plunker Templates:

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:0.9

* **OS and version:Mac

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

adityarb88:
@rangagv Thanks for the feedback. I have tagged our UX team on this issue. They will take a look and respond as soon as possible.

jaffoneh:
@rangagv quick question: when you say selection, are you using checkboxes for selection?

rangagv:
@jaffoneh Yes, we are using the check box selections.

juanmendes:
Just to clarify, what we are looking for is a `clrDgSelectable` attribute on rows. I believe this should work for both checkbox or regular selection.

rangagv:
clrDgSelectable works for selection/unselection, but I requested to enable/disable of selection for rows.

juanmendes:
@rangagv I think I see what you mean.

Let me try it: you are saying that `clrDgSelectable` sounds like a row cannot be selected (but could be deselected)? However,  what you want is the ability to set a property that means a row cannot change its selection state from the UI, whether it's selected or unselected, it will remain that way.

If that's what you are saying, I do agree, and the name I proposed is not very accurate. Maybe `clrChangeSelectionStateEnabled` is a better name, meaning the selection state cannot be changed through the UI.

[...]

PR #3749, opened 2019-08-27
feat(datagrid): disable single or multi rows from selection

bdryanovski:
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Right now there is no way to disable a row from Datagrid and prevent interaction with the row.

Issue Number: #1018 

## What is the new behavior?

Added new input to `ClrDatagridRow` called `cldDgUnselectable` that except `boolean`. When set to true it will prevent the click event from firing and won't let the user change the state of the row. Also will mark the checkbox or radio button as disabled and change the cursor to mark the row as locked.

This is valid for both Single and Multi selectable rows. 

#### Single selection
* Normal
* Disabled and can't be changed
* Disabled and marked as selected, also can't be changed
Screen Shot 2019-08-27 at 5 09 32 PM

#### Multi selection
* Disabled and marked as selected
* Disabled and can't be changed
* Normal
Screen Shot 2019-08-27 at 5 08 27 PM

* [ ] Yes
* [x] No

## Other information
Close: #1018 

Todo: need to be backported to V1

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-row.html
+++ b/src/clr-angular/data/datagrid/datagrid-row.html
@@ -32,18 +32,27 @@
     <div class="datagrid-row-scrollable" [ngClass]="{'is-replaced': replaced && expanded}">
       <div class="datagrid-scrolling-cells">
         <div *ngIf="selection.selectionType === SELECTION_TYPE.Multi"
-             class="datagrid-select datagrid-fixed-column datagrid-cell" role="gridcell">
+             class="datagrid-select datagrid-fixed-column datagrid-cell" 
+             [ngClass]="{ 'clr-form-control-disabled': !clrDgSelectable }"
+             role="gridcell">
+             
           <input clrCheckbox type="checkbox" [ngModel]="selected" (ngModelChange)="toggle($event)" [id]="checkboxId"
+                 [attr.disabled]="clrDgSelectable ? null : true"
+                 [attr.aria-disabled]="clrDgSelectable ? null : true"
                  [attr.aria-label]="commonStrings.keys.select">
         </div>
         <div *ngIf="selection.selectionType === SELECTION_TYPE.Single"
-             class="datagrid-select datagrid-fixed-column datagrid-cell" role="gridcell">
+             class="datagrid-select datagrid-fixed-column datagrid-cell" role="gridcell"
+             [ngClass]="{ 'clr-form-control-disabled': !clrDgSelectable }"
+             >
             <!-- TODO: it would be better if in addition to the generic "Select" label, we could add aria-labelledby
             to label the radio by the first cell in the row (typically an id or name).
             It's pretty easy to label it with the whole row since we already have an id for it, but in most
             cases the row is far too long to serve as a label, the screenreader reads every single cell content. -->
             <input type="radio" clrRadio [id]="radioId" [name]="selection.id + '-radio'" [value]="item"
                    [(ngModel)]="selection.currentSingle" [checked]="selection.currentSingle === item"
+                   [attr.disabled]="clrDgSelectable ? null : true"
+                   [attr.aria-disabled]="clrDgSelectable ? null : true"
                    [attr.aria-label]="commonStrings.keys.select">
         </div>
         <div *ngIf="rowActionService.hasActionableRow"
--- a/src/clr-angular/data/datagrid/datagrid-row.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-row.spec.ts
@@ -87,6 +87,42 @@ export default function(): void {
       });
     });
 
+    describe('Conditional selection', function() {
+      let context: TestContext<ClrDatagridRow, SelectableRow>;
+      let selectionProvider: Selection;
+
+      let radio: HTMLElement;
+      let checkbox: HTMLElement;
+
+      beforeEach(function() {
+        context = this.create(ClrDatagridRow, SelectableRow, DATAGRID_SPEC_PROVIDERS);
+        selectionProvider = TestBed.get(Selection);
+        TestBed.get(Items).all = [{ id: 1 }, { id: 2 }];
+      });
+
+      it('should toggle when clrDgSelectable is false for type  SelectionType.Multi', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        checkbox = context.clarityElement.querySelector("input[type='checkbox']");
+
+        expect(checkbox.getAttribute('disabled')).toBe('true');
+        expect(checkbox.getAttribute('aria-disabled')).toBe('true');
+
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(true);
+      });
+
+      it('should be able you toggle the state of selection when clrDgSelectable is false', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.clarityDirective.toggle(true);
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(false);
+      });
+    });
+
     describe('Selection', function() {
       // Until we can properly type "this"
       let context: TestContext<ClrDatagridRow, FullTest>;
@@ -422,6 +458,14 @@ export default function(): void {
     `,
 })
 class ProjectionTest {}
+
+@Component({
+  template: `<clr-dg-row [clrDgSelectable]="clrDgSelectable" [clrDgItem]="item">None</clr-dg-row>`,
+})
+class SelectableRow {
+  clrDgSelectable = true;
+  item: Item = { id: 42 };
+}
 
 @Component({ template: `<clr-dg-row [clrDgItem]="item" [(clrDgSelected)]="selected">Hello world</clr-dg-row>` })
 class FullTest {
--- a/src/clr-angular/data/datagrid/datagrid-row.ts
+++ b/src/clr-angular/data/datagrid/datagrid-row.ts
@@ -131,6 +131,16 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit 
     }
   }
 
+  // By default every item is selectable
+  @Input('clrDgSelectable')
+  public set clrDgSelectable(value: boolean) {
+    this.selection.lockItem(this.item, value === false);
+  }
+
+  public get clrDgSelectable() {
+    return !this.selection.isLocked(this.item);
+  }
+
   @Output('clrDgSelectedChange') selectedChanged = new EventEmitter<boolean>(false);
 
   public toggle(selected = !this.selected) {
--- a/src/clr-angular/data/datagrid/providers/selection.spec.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -639,6 +639,120 @@ export default function(): void {
             expect(selectionInstance.isSelected(itemsA[2])).toBe(false);
           })
         );
+      });
+    });
+
+    fdescribe('conditional selection', function() {
+      let selectionInstance: Selection<any>;
+      let sortInstance: Sort<any>;
+      let pageInstance: Page;
+      let filtersInstance: FiltersProvider<any>;
+      let itemsInstance: Items<any>;
+
+      beforeEach(function() {
+        const stateDebouncer = new StateDebouncer();
+        pageInstance = new Page(stateDebouncer);
+        filtersInstance = new FiltersProvider(pageInstance, stateDebouncer);
+        sortInstance = new Sort(stateDebouncer);
+        itemsInstance = new Items(filtersInstance, sortInstance, pageInstance);
+        itemsInstance.smartenUp();
+        itemsInstance.all = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+
+        selectionInstance = new Selection(itemsInstance, filtersInstance);
+      });
+
+      afterEach(function() {
+        selectionInstance.destroy();
+        itemsInstance.destroy();
+      });
+
+      it("can't select item if it's locked when selection type is Multi", function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        // lock a row
+        selectionInstance.lockItem(4, true);
+        // make sure it's locked
+        expect(selectionInstance.isLocked(4)).toBe(true);
+        // select all
+        selectionInstance.toggleAll();
+        // locked row is not selected
+        expect(selectionInstance.isSelected(4)).toBe(false);
+
+        // unselect all
+        selectionInstance.toggleAll();
+        // locked row is not selectable
+        expect(selectionInstance.current.length).toEqual(0);
+        expect(selectionInstance.isSelected(4)).toBe(false);
+      });
+
+      /*
+       * single selection is done with radio button so there is no logic to prevent
+       * it from the provider - this must be handle at the component level
+       */
+      it("should not block selecting item if it's locked when selection type is Single", function() {
+        selectionInstance.selectionType = SelectionType.Single;
+        // lock a row
+        selectionInstance.lockItem(4, true);
+        // make sure it's locked
+        expect(selectionInstance.isLocked(4)).toBe(true);
+        selectionInstance.currentSingle = 4;
+        expect(selectionInstance.isSelected(4)).toBe(true);
+      });
+
+      it('should let you select locked item when using setSelected', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+
+        selectionInstance.lockItem(4, true);
+        selectionInstance.setSelected(4, true);
+        expect(selectionInstance.isSelected(4)).toBe(true);
+      });
+
+      it('should return true when all posible items are selected excluding locked', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        selectionInstance.lockItem(4, true);
+        selectionInstance.toggleAll();
+        expect(selectionInstance.isAllSelected()).toBe(true);
+      });
+
+      /**
+       * Clearing selection must not change the lockedRef.
+       */
+      it('should not reset lockedRefs when clearing selection', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        selectionInstance.lockItem(4, true);
+        selectionInstance.clearSelection();
+        expect(selectionInstance.isLocked(4)).toBe(true);
+      });
+
+      it('should remove locked item when is no longer part from the items list and type is Single', function() {
+        selectionInstance.selectionType = SelectionType.Single;
+        selectionInstance.lockItem(4, true);
+        itemsInstance.all = [5, 6, 7];
+        expect(selectionInstance.isLocked(4)).toBe(false);
+      });
+
+      it('should remove locked item when is no longer part from the items list and type is Multi', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        selectionInstance.lockItem(6, true);
+        itemsInstance.all = [5, 7, 8];
+        expect(selectionInstance.isLocked(6)).toBe(false);
+      });
+
+      it('should use trackBy to find already locked items when the list is replaced (integer)', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        selectionInstance.lockItem(2, true);
+        // Create new list of items
+        itemsInstance.all = [2, 5, 6, 7];
+        expect(selectionInstance.isLocked(2)).toBe(true);
+      });
+
+      it('should use trackBy to find already locked items when the list is replaced (object)', function() {
+        itemsInstance.trackBy = (index, { id }) => id;
+        selectionInstance.selectionType = SelectionType.Multi;
+        itemsInstance.all = [{ id: 1 }, { id: 2 }];
+        selectionInstance.lockItem({ id: 2 }, true);
+        // Create new list of items
+        itemsInstance.all = [{ id: 1 }, { id: 2 }];
+        expect(selectionInstance.isLocked({ id: 2 })).toBe(true);
       });
     });
   });
--- a/src/clr-angular/data/datagrid/providers/selection.ts
+++ b/src/clr-angular/data/datagrid/providers/selection.ts
@@ -17,6 +17,7 @@ export class Selection<T = any> {
   public id: string;
   private prevSelectionRefs: T[] = []; // Refs of selected items
   private prevSingleSelectionRef: T; // Ref of single selected item
+  private lockedRefs: T[] = []; // Ref of licked items
 
   constructor(private _items: Items<T>, private _filters: FiltersProvider<T>) {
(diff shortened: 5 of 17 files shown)


src/website/src/app/documentation/demos/datagrid/single-selection/single-selection.html
```diff
@@ -115,3 +115,30 @@ <h2>Selection (Single)</h2>
...
+<p>
+    Mark a row with <code class="clr-code">clrDgUnselectable</code>, this way the state of the row could not be changed by 
+    user interactions. This property will work with checkbox and radio button rows.
+</p>
+<clr-code-snippet [clrCode]="unselectableRow" clrLanguage="html"></clr-code-snippet>
+
+<button class="btn btn-primary" (click)="lockRow()">Lock Rows</button>
+<button class="btn btn-primary" (click)="unlockRow()">Unlock Rows</button>
+
+<clr-datagrid [(clrDgSingleSelected)]="rowSelected" [clrDgRowSelection]="true">
```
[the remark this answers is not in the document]
bdryanovski: Ok, will go and check other templates that we are using and will remove them as well

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -32,18 +32,25 @@
...
+             class="datagrid-select datagrid-fixed-column datagrid-cell" role="gridcell"
+             [ngClass]="{ 'clr-form-control-disabled': unselectable }"
+             >
             <!-- TODO: it would be better if in addition to the generic "Select" label, we could add aria-labelledby
             to label the radio by the first cell in the row (typically an id or name).
             It's pretty easy to label it with the whole row since we already have an id for it, but in most
             cases the row is far too long to serve as a label, the screenreader reads every single cell content. -->
             <input type="radio" clrRadio [id]="radioId" [name]="selection.id + '-radio'" [value]="item"
                    [(ngModel)]="selection.currentSingle" [checked]="selection.currentSingle === item"
+                   [attr.disabled]="unselectable ? true : null"
```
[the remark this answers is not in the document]
bdryanovski: Ok

bdryanovski:
You don't have to be able to check them from the header toggle - so this is not ok ;)

src/website/src/app/documentation/demos/datagrid/selection/selection.ts
```diff
@@ -56,13 +63,41 @@ export class DatagridSelectionDemo {
...
     inventory.reset();
     this.users = inventory.all;
+
+    this.lockedUsers = [...inventory.all];
+  }
+
+  unlockRows() {
+    this.lockedUsers = this.lockedUsers.map(row => {
+      row.locked = undefined;
+      return row;
```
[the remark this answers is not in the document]
bdryanovski: I already fix that .. need just to push it ;)

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -315,36 +316,71 @@ export class Selection<T = any> {
...
+    const nbDisplayed = displayedItems.length;
     if (nbDisplayed < 1) {
       return false;
     }
     const temp: T[] = displayedItems.filter(item => this.current.indexOf(item) > -1);
     return temp.length === displayedItems.length;
   }
 
+  private canItBeLocked() {
+    return [SelectionType.Multi, SelectionType.Single].indexOf(this._selectionType) > -1;
```
gnomeontherun:
```
return this._selectionType !== SelectionType.None
```

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -129,9 +130,29 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
     }
   }
 
+  /** By default every item is selectable */
+  @Input('clrDgSelectable')
+  public set clrDgSelectable(value: boolean) {
+    if ([SelectionType.Single, SelectionType.Multi].indexOf(this.selection.selectionType) > -1) {
+      this.selection.lockItem(this.item, !parseBooleanAttribute(value));
```
gnomeontherun: Since we are only handling Angular inputs here, we don't need `parseBooleanAttribute()`, just set it to the value provided.

Also the selection type check is unnecessarily complex, just `this.selection.selectionType !== SelectionType.None` would suffice here.

```
if (this.selection.selectionType !== SelectionType.None) {
  this.selection.lockItem(this.item, value);
}
```

src/clr-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -131,6 +139,13 @@ export default function(): void {
         expect(selectionInstance.isAllSelected()).toBe(true);
       });
 
+      it('should return true when all posible items are selected excluding locked', function() {
+        selectionInstance.selectionType = SelectionType.Multi;
+        selectionInstance.lockItem(4, true);
+        selectionInstance.toggleAll();
+        expect(selectionInstance.isAllSelected()).toBe(true);
```
gnomeontherun: We should have a test for the inverse of `toggleAll()` to test that it works to unselect, also a test if the locked items are checked too.

src/clr-angular/utils/component/parse-boolean-attribute.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export function parseBooleanAttribute(value: string | boolean | number): boolean {
```
gnomeontherun: We don't need this function, plus it duplicates what is in the other boolean attribute helper function.

src/dev/src/app/datagrid/disabled-rows/disabled-rows.ts
```diff
@@ -0,0 +1,79 @@
...
+  selectedRows: User[] = [];
+
+  constructor(inventory: Inventory) {
+    inventory.size = 10;
+    inventory.reset();
+    this.users = inventory.all;
+    this.usersMulti = inventory.all;
+  }
+
+  unlockRow() {
```
gnomeontherun: nit: these names are not clear with which is single/multi

src/dev/src/app/datagrid/disabled-rows/disabled-rows.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.users = this.users.map(row => {
+      row.locked = undefined;
```
gnomeontherun: It seems strange to set to 'undefined' instead of to `false`. You could also `delete row.locked;` if you want to remove it.

src/dev/src/app/datagrid/selection-row-mode/selection-row-mode.html
```diff
@@ -35,7 +35,7 @@ <h2>Selection (Row mode)</h2>
 
 <h3>Single selection</h3>
 
-<clr-datagrid [(clrDgSingleSelected)]="singleSelected" [clrDgRowSelection]="true">
```
gnomeontherun: We should leave these in place, revert this file. This feature hasn't been removed so we need to have one place to test it.

src/website/src/app/documentation/demos/datagrid/selection/selection.html
```diff
@@ -102,3 +102,30 @@ <h2>Selection</h2>
     If you need to listen to when the selection changes, you can use Angular's <a href="https://angular.io/guide/template-syntax">two way binding</a> to listen to the <code class="clr-code">(clrDgSelectedChange)</code> event:
 </p>
 <clr-code-snippet [clrCode]="selectionChangeEventExample" clrLanguage="html"></clr-code-snippet>
+
+<p>
+    Mark a row with <code class="clr-code">clrDgSelectable</code>, this way the state of the row could not be changed by 
+    user interactions. This property will work with checkbox and radio button rows.
```
gnomeontherun: > This property works only when using single or multi-selection modes.

src/website/src/app/documentation/demos/datagrid/selection/selection.ts
```diff
@@ -27,7 +27,7 @@ const SINGLE_ROW_EXAMPLE = `
 `;
 
 const ROW_SELECTION_EXAMPLE = `
-<clr-datagrid [(clrDgSelected)]="rowSelected" [clrDgRowSelection]="true">
```
gnomeontherun: Put this back, because its still in the documentation.

src/website/src/app/documentation/demos/datagrid/single-selection/single-selection.ts
```diff
@@ -21,7 +21,7 @@ Selected user: <span class="username" *ngIf="selectedUser">{{selectedUser.name}}
 `;
 
 const ROW_SELECTION_EXAMPLE = `
-<clr-datagrid [(clrDgSingleSelected)]="selectedUser" [clrDgRowSelection]="true">
```
gnomeontherun: Put this back.

src/website/src/app/documentation/demos/datagrid/selection/selection.html
```diff
@@ -71,7 +71,7 @@ <h2>Selection</h2>
     For row selection mode add <code class="clr-code">[(clrDgRowSelection)]</code> and
     set it to <code class="clr-code">true</code>.
 </p>
-<clr-datagrid [(clrDgSelected)]="rowSelected" [clrDgRowSelection]="true">
```
gnomeontherun: Out of scope ask: can you put the same deprecation warning above this demo as you see https://deploy-preview-3749--vmware-clarity.netlify.com/documentation/datagrid/selection-single?

src/website/src/app/documentation/demos/datagrid/single-selection/single-selection.ts
```diff
@@ -50,13 +58,41 @@ export class DatagridSelectionSingleDemo {
...
+    // reset selections
+    this.rowSelected = undefined;
+
+    this.lockedUsers = this.lockedUsers.map((user, index) => {
+      /** lock some rows */
+      if ([2, 3, 5, 10].includes(index)) {
+        user.locked = true;
+      }
+      /** also mark them as selected */
+      if ([5].includes(index)) {
```
gnomeontherun: Remove this, it is confusing to me trying to use the demo for the selection to change on me.

src/website/src/app/documentation/demos/datagrid/selection/selection.ts
```diff
@@ -56,13 +63,41 @@ export class DatagridSelectionDemo {
...
+    // reset selections
+    this.selected = [];
+
+    this.lockedUsers = this.lockedUsers.map((user, index) => {
+      /** lock some rows */
+      if ([2, 3, 5, 9].includes(index)) {
+        user.locked = true;
+      }
+      /** also mark them as selected */
+      if ([2, 5].includes(index)) {
```
gnomeontherun: Remove this selected state setting, the demo is confusing to lock and change selection at once.

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -315,36 +316,71 @@ export class Selection<T = any> {
...
+  }
+
+  /**
+   * Lock and unlock item
+   */
+  public lockItem(item: T, lock: boolean) {
+    if (this.canItBeLocked()) {
+      if (lock === true) {
+        /* Add to lockedRef */
+        this.lockedRefs.push(item);
```
gnomeontherun: If we are storing a ref, we need to first pass it through the trackBy function so we can keep track of the rows. If someone has a dynamic table where they change the list of items, this list of refs will be stale because we'll have new rows.

1) Look above at how trackBy is used to calculate a reference, and also make sure you have a test.
2) When the whole list of items is reset (see above in _allChanges subscription), you'll want to clear the old `lockedRefs` list and let it get rebuilt from the new items, so we don't have a memory leak. The array could continue to grow in side with stale refs otherwise.
3) Anytime you are looking up an item against the `lockedRefs`, you'll have to recompute the ref from the item like in the `isLocked` method below.

src/dev/src/app/datagrid/disabled-rows/disabled-rows.ts
```diff
@@ -0,0 +1,79 @@
...
+  unlockRows() {
+    this.usersMulti = this.usersMulti.map(row => {
+      row.locked = undefined;
+      return row;
+    });
+  }
+
+  lockRow() {
+    // reset selections
+    this.selected = undefined;
```
coryrylan: For clarity I would keep boolean values strictly true/false unless it is being used in a true tri state where the is a unset state.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -87,6 +87,51 @@ export default function(): void {
...
+        TestBed.get(Items).all = [{ id: 1 }, { id: 2 }];
+      });
+
+      it('should not toggle when clrDgSelectable is false for type SelectionType.Single', function() {
+        selectionProvider.selectionType = SelectionType.Single;
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+
+        expect(context.clarityElement.querySelector("input[type='radio']").getAttribute('disabled')).toBe('true');
+        expect(context.clarityElement.querySelector("input[type='radio']").getAttribute('aria-disabled')).toBe('true');
```
coryrylan: nit: you could put the radio query in the before each to reduce a bit of duplication
`radio = context.clarityElement.querySelector("input[type='radio']"`

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -315,36 +316,71 @@ export class Selection<T = any> {
     if (this._selectionType !== SelectionType.Multi || !this._items.displayed) {
       return false;
     }
-    const displayedItems: T[] = this._items.displayed;
-    const nbDisplayed = this._items.displayed.length;
+    /** make sure to exclude the locked items from the list */
```
coryrylan: nit: Check through the comments in the PR. We should probably not use the `/**` notation for comments as they are for TSDoc and would show up in generated docs. We are planning on providing this in the future so we should only use that notation if we want it to show up in the docs.

src/website/src/app/documentation/demos/datagrid/single-selection/single-selection.html
```diff
@@ -115,3 +115,30 @@ <h2>Selection (Single)</h2>
     If you need to listen to when the selection changes, you can use Angular's <a href="https://angular.io/guide/template-syntax">two way binding</a> to listen to the <code class="clr-code">(clrDgSingleSelectedChange)</code> event:
 </p>
 <clr-code-snippet [clrCode]="selectionChangeEventExample" clrLanguage="html"></clr-code-snippet>
+
+<p>
+    Mark a row with <code class="clr-code">clrDgSelectable</code>, this way the state of the row could not be changed by
```
gnomeontherun: > In order to conditionally disable selection on a row, use the `clrDgSelectable` input to disable selection state changes. This has to be done on each row you wish to disable, and works with single and multi selection.

src/website/src/app/documentation/demos/datagrid/selection/selection.html
gnomeontherun: Copy my other message up here, it is not currently clear why you would use it.

src/dev/src/app/datagrid/conditional-selection/conditional-selection.ts
```diff
@@ -0,0 +1,71 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component } from '@angular/core';
+
+import { Inventory } from '../inventory/inventory';
+import { User } from '../inventory/user';
+
+@Component({
+  selector: 'clr-datagrid-conditional-selection-demo',
```
gnomeontherun: nit: not used since its routable

src/clr-angular/data/datagrid/providers/selection.ts
gnomeontherun: I still don't see the trackBy here, please review this again.

bdryanovski:
Ignore the `dco-required` it will be fixed when I rebase the code

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -144,6 +165,7 @@ export class Selection<T = any> {
     this.prevSelectionRefs = [];
     this._currentSingle = null;
     this.prevSingleSelectionRef = null;
+    this.lockedRefs = [];
```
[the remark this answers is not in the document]
bdryanovski: You are right!

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -315,31 +353,69 @@ export class Selection<T = any> {
     if (this._selectionType !== SelectionType.Multi || !this._items.displayed) {
       return false;
     }
-    const displayedItems: T[] = this._items.displayed;
-    const nbDisplayed = this._items.displayed.length;
+    /** make sure to exclude the locked items from the list */
+    const displayedItems: T[] = this._items.displayed.filter(item => !(this.lockedRefs.indexOf(item) > -1));
```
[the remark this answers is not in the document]
bdryanovski: I miss that..

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -315,31 +353,69 @@ export class Selection<T = any> {
...
     if (nbDisplayed < 1) {
       return false;
     }
     const temp: T[] = displayedItems.filter(item => this.current.indexOf(item) > -1);
     return temp.length === displayedItems.length;
   }
 
+  private canItBeLocked() {
+    // We depend on the trackBy and all so there are part of the requirment of is item could be locked
+    return this._selectionType !== SelectionType.None && this._items.all && this._items.trackBy;
```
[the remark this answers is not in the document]
bdryanovski: The reason for the other two `&&` is that in some cases the `all` and `trackBy` are not defined. And there is dependency when you want to check the `ref`. It could be the only case that is reproducible into the tests, I did not found any example into the dev app to fail.

With the latest changes, we could go with `this._selectionType !== SelectionType.None`.

src/clr-angular/data/datagrid/providers/selection.spec.ts
```diff
@@ -641,6 +641,104 @@ export default function(): void {
         );
       });
     });
+
+    describe('conditional selection', function() {
```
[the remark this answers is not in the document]
bdryanovski: The `trackBy` function is the same as the one used for selection and filtering so it's cover above. 

The test inside this `describe` block all use and depend on `isLocked` & `lockItem` that are the main entry points when it comes to checking or managing locked items.

I added one more test below to test when replacing the list of items and an item that when passed will match `lockedRef` stay locked. ( when the list of the datagrid is replaced but contain mostly the same item and they have to stay locked )

I'm not sure what test to write to check is `trackBy` miss-used or not working as expected?  If you have something in mind or I'm missing something obvious?

bdryanovski:
Could you provide me with some example of:

>If I want to use [(clrDgSelected)] for setting selection in multi-select mode, locking breaks this behavior. It works fine until I lock something.

src/clr-angular/data/datagrid/providers/selection.spec.ts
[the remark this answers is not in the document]
gnomeontherun: The problem with that test is that numbers always have reference equality, `2 === 2` is always true. What you need to do is setup a test like this.

```
      it('should use trackBy to find already locked items when the list is replaced', function() {
       itemsInstance.all = [{id: 1}, {id: 2}];
        selectionInstance.selectionType = SelectionType.Multi;
        selectionInstance.lockItem({id: 2}, true);
        // Create new list of items
        itemsInstance.all = [{id: 1}, {id: 2}];
        expect(selectionInstance.isLocked({id: 2})).toBe(true);
      });
```

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -156,6 +178,8 @@ export class Selection<T = any> {
       return;
     }
     this._selectionType = value;
+    // when type change so lockedRefs also have to be reset
+    this.lockedRefs = [];
```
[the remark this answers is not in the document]
gnomeontherun: Can you remove this then?

src/clr-angular/data/datagrid/providers/selection.ts
```diff
@@ -198,6 +222,10 @@ export class Selection<T = any> {
     if (value === this._currentSingle) {
       return;
     }
+    // locked items can't be selected
```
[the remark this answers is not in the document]
gnomeontherun: This is still here, we need to remove this and the next comment below.

We basically only want to prevent users from selecting/unselecting a locked item. However, an application might want to programmatically change the selected state of a locked row (I can't think of many great scenarios but I still believe we should allow it).

src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -87,6 +87,54 @@ export default function(): void {
...
+        expect(context.clarityDirective.selected).toBe(false);
+      });
+
+      it('should not be able you toggle the state of selection when clrDgSelectable is false', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.clarityDirective.toggle(true);
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(true);
```
gnomeontherun: Looking at this again, this is one of the cases we want to allow. Users cannot call the `toggle()` method, they can only select the button (which is disabled). So we want to be able to allow the API to toggle a locked row but not a button click. That means line 149 of datagrid-row.ts also needs to be removed. This puts the API in agreement, where they can set the selected state manually through the input binding or use `toggle()`.

src/clr-angular/data/datagrid/providers/selection.ts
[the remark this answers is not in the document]
bdryanovski: Not sure why this is visible, cause on my file it's not. it's something wrong with the diffs here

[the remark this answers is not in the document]
bdryanovski: Screen Shot 2019-09-26 at 10 10 58 AM

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -129,9 +129,27 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
+  }
+
   @Output('clrDgSelectedChange') selectedChanged = new EventEmitter<boolean>(false);
 
   public toggle(selected = !this.selected) {
+    /**
+     * When selectable is false we won't let the user change
+     * the state until set to undefined or true
+     */
+    if (this.clrDgSelectable === false) {
```
gnomeontherun: This is what we needed to remove.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
bdryanovski: Sorry for the late reply, had to test it before making any changes. It seems that it works as planned even without the toggle() return.

src/dev/src/app/datagrid/conditional-selection/conditional-selection.ts
bdryanovski: all other examples have one so didn't want to break the tradition :)

hamza1100:
Hello @bdryanovski .
Could you please share an example using this **clrDgUnselectable** attribute? I've tried this but it's not working for me.

bdryanovski:
@hamza1100 you could read and find an example [here](https://clarity.design/documentation/datagrid/selection)

PR #3756, opened 2019-08-29
fix(datagrid): datagrid current page not rendered correctly in IE 11

### LINKED ISSUE #3747: Datagrid current page not rendered correctly in IE 11
## Describe the bug

In version 1.2.x the current page in the pagination footer is rendered correctly:

Datagrid paging v1 2 x

However, in version 2.x the current page in the pagination footer is not rendered correctly:

Datagrid paging v2 0

## Expected behavior

The current page number should render correctly.

## Versions

**App**

* Angular: 8.2.3
* Node: 12.9
* Clarity: 2.1.2

**Device:**

* Type: Dell Latitude
* OS: Windows 7 Pro
* Browser IE 11 
* Version 11.0.9600.19403

## Additional notes

AlexMarcuBytex:
closes #3747 

Signed-off-by: Marcu Alexandru alexandru.marcu@bytex.ro

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3747 

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -697,6 +697,7 @@
       border: $clr-default-borderwidth solid $clr-datagrid-pagination-input-border-color;
       border-radius: $clr-global-borderradius;
       line-height: 1rem;
+      font-size: 0.45833rem;
       min-width: 1rem;
       text-align: center;
       transition: none !important;
--- a/src/clr-angular/data/datagrid/datagrid-pagination.ts
+++ b/src/clr-angular/data/datagrid/datagrid-pagination.ts
@@ -47,17 +47,17 @@ import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service
         >
         <clr-icon shape="angle left"></clr-icon>
       </button>
-      <input 
-        #currentPageInput 
-        type="text" 
-        class="pagination-current" 
-        [size]="page.last.toString().length" 
-        [value]="page.current"
-        (keydown.enter)="updateCurrentPage($event)" 
-        (blur)="updateCurrentPage($event)"
-        [attr.aria-label]="commonStrings.keys.currentPage"
-        />
-        &nbsp;/&nbsp;<span [attr.aria-label]="commonStrings.keys.totalPages">{{page.last}}</span>
+        <input 
+          #currentPageInput 
+          type="text" 
+          class="pagination-current clr-input" 
+          [size]="page.last.toString().length" 
+          [value]="page.current"
+          (keydown.enter)="updateCurrentPage($event)" 
+          (blur)="updateCurrentPage($event)"
+          [attr.aria-label]="commonStrings.keys.currentPage"
+          />
+          &nbsp;/&nbsp;<span [attr.aria-label]="commonStrings.keys.totalPages">{{page.last}}</span>
       <button 
         type="button"
         class="pagination-next" 


src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -709,6 +709,13 @@
       }
     }
 
+    //fixed with this mixin because the IE11 and IE10 mixins didnt work
+    @include fixForIE9AndUp() {
+      .pagination-current {
+        line-height: 0.5rem;
```
hippee-lee: lets use line-height: normal here instead. That will fix the issue in IE11.

src/clr-angular/utils/_helpers.clarity.scss
```diff
@@ -63,6 +63,12 @@ $opp-directions: (
   }
 }
 
+@mixin fixForIE9AndUp() {
```
hippee-lee: I would rather not call this fixForIE9AndUp because Clarity only supports IE11 and up. Yes, I know there is one below with the name fixForIE10AndUp but that is from when we used to support IE10 and it should probably be removed.

What do you think about fixForIE11 - since this is only for IE11.

src/clr-angular/utils/_helpers.clarity.scss
```diff
@@ -63,6 +63,12 @@ $opp-directions: (
...
+  @media screen and (min-width: 0\0) and (min-resolution: +72dpi) {
```
hippee-lee: How sure are you that this only targets IE11? Can you comment/link to an explanation of the hack?

src/clr-angular/utils/_helpers.clarity.scss
AlexMarcuBytex: i tried to use the fixForIE11 but it didnt work. also, i see that that function its not used anywhere. maybe we should find another hack that works only for IE11. what do you think?

src/clr-angular/utils/_helpers.clarity.scss
AlexMarcuBytex: This is targeting more IE versions ( 9, 10 and 11). Many sources doesnt explain properly how these hacks work.
Meanwhile, i found another hack which seems to have a plausible explanation
``@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {  
   /* specific styles go here */  
}``
This works for 10 and 11 only. It seems like IE10 and 11 (only) support the -ms-high-contrast media query. If the actual IE11 targeting mixin doesnt work, what do you think about changing it with another one?

src/clr-angular/data/datagrid/_datagrid.clarity.scss
AlexMarcuBytex: changed. works fine.

gnomeontherun:
We don't want to be putting in browser hacks, and I just tried a different tact. In the pagination HTML, I just added `clr-input` to the pagination input and I believe the styling is correct. Can you look into that instead as an option?

src/clr-angular/utils/_helpers.clarity.scss
hippee-lee: I think this is the right approach. What happened that prevented `fixForIE11` from working? Was there an error you saw? Could it have anything to do with the numbers at the end?

src/clr-angular/utils/_helpers.clarity.scss
hippee-lee: This 
```
@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) { /* specific styles go here */ }
```
was the hack I also saw for targeting IE11. Lets go with that and figure out why the mixin isn't working for you. Can you provide any feedback or compile errors related to the the mixin name?

hippee-lee: Lets try the recommendation below from @gnomeontherun before we proceed with the hacky solution.

src/clr-angular/utils/_helpers.clarity.scss
hippee-lee: Lets try the recommendation below from @gnomeontherun before we proceed with the hacky solution.

src/clr-angular/data/datagrid/datagrid-pagination.ts
```diff
@@ -47,17 +47,17 @@ import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service
...
-        [value]="page.current"
-        (keydown.enter)="updateCurrentPage($event)" 
-        (blur)="updateCurrentPage($event)"
-        [attr.aria-label]="commonStrings.keys.currentPage"
-        />
-        &nbsp;/&nbsp;<span [attr.aria-label]="commonStrings.keys.totalPages">{{page.last}}</span>
+        <input 
+          #currentPageInput 
+          type="text" 
+          class="pagination-current clr-input"
```
mathisscott: You are going to need to add `font-size: 0.45833rem` to `.pagination-current` in `_datagrid.clarity.scss` or the text inside the pagination field is going to look like Andre the Giant.

PR #3760, opened 2019-08-29
fix(a11y): add bumpers to the datepicker

### LINKED ISSUE #3466: Screen readers can read beyond the content of the date picker popovers
## Describe the bug

Screen readers can read beyond the content of the date picker popovers.

## How to reproduce

- Open any date picker example using a screen reader.
- Attempt to move the reading cursor beyond the date picker.

## Expected behavior

Follow the Accessible Calendar Picker example at http://whatsock.com/bootstrap. It is using role="application" which prevents screen readers from capturing the arrow key events. The arrow keys are then programmed for cell to cell navigation. Also should use `aria-modal` to indicate to screen readers about modality.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

We have a different solution and pattern for modals here https://github.com/vmware/clarity/issues/3420, if the expected implementation above isn't working we should follow this pattern instead.

Jinnie:
closes #3466

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

No indication of beginning and end of modal when using screen reader.
Weekdays headings (SMTWTFS) are only read as a letter.
Issue Number: #3466

## What is the new behavior?
Begin/End bumpers added, only visible to screen readers.
Weekdays are read with their full names.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/forms/datepicker/providers/locale-helper.service.ts
+++ b/src/clr-angular/forms/datepicker/providers/locale-helper.service.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -14,6 +14,7 @@ import {
   TranslationWidth,
 } from '@angular/common';
 import { Inject, Injectable, LOCALE_ID } from '@angular/core';
+import { ClrDayOfWeek } from '../interfaces/day-of-week.interface';
 
 /**
  * This service extracts the Angular CLDR data needed by the datepicker.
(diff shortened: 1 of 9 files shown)


src/clr-angular/forms/datepicker/providers/locale-helper.service.ts
```diff
@@ -15,6 +15,8 @@ import {
 } from '@angular/common';
 import { Inject, Injectable, LOCALE_ID } from '@angular/core';
 
+export type DayOfWeekType = { day: string; narrow: string };
```
coryrylan: Just to be more consistent with the codebase and objects could we make this an `interface` instead of a type unless there is a particular reason to use `type`?

Jinnie: I first intended to only use it internally in the component, but that was too short-sighted of an expectation. Moving to an interface, thanks.

Jinnie: done

Issue #3767, opened 2019-08-30
Signpost trigger is stealing focus by default


Shijir:
## Describe the bug

The signpost trigger is stealing focus when it's initialized.

## How to reproduce

https://stackblitz.com/edit/signpost-trigger-stealing-focus

I think the culprit is in `clrIfOpen` because it looks like the directive has the same name input
```
@Input('clrIfOpen')
  public set open(value: boolean) {
    this.ifOpenService.open = value;
  }
```
So just by using this directive, we would set the value to the `open` property of `ifOpenService`.

## Expected behavior

The signpost trigger shouldn't set focus on itself when it's initialized.
It should get the focus only after its popover content gets closed.

## Versions

**App**

* Angular: [8.x.x]
* Node: [e.g. 8.10.0]
* Clarity: [2.x.x]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

Shijir:
Another issue I noticed is that when there are multiple signpost triggers, and I click on one of them after clicking on another, the focus is going back to the inactive signpost trigger.

![Screen Shot 2019-08-30 at 7 34 51 AM](https://user-images.githubusercontent.com/3958008/64029078-b8a48280-caf8-11e9-95cf-1c8a2a79efa7.png)

Shijir:
Closed by #3774

Issue #3762, opened 2019-08-29
Auto focus causes page scroll for ClrAlert and ClrSignpost components


envyN:
## Describe the bug

If we open https://clarity.design/documentation/signposts we are automatically scrolled down to the last signpost trigger on the page.
This is not expected

## How to reproduce

Visit 
https://clarity.design/documentation/signposts 
or
https://clarity.design/documentation/alerts

## Expected behavior

Signpost trigger shouldnt be focused automatically(at least not if its in closed state)

## Versions

**App**

* Clarity: [1.2.x] and [2.x]

## Additional notes

This behavior is breaking apps if they use libraries like ngx-infinite-scroll since clarity scrolls it down to the last signpost which triggers the infinite scroll fetch, and the app is stuck in a continuous fetch and focus loop.

Shijir:
closed by #3787 and #3774

PR #3774, opened 2019-08-31
fix(signpost): fix signpost focus management

Shijir:
- fix focus stealing on initialization 
- sets focus back on trigger properly when it's right

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

- it was setting focus on the trigger on initialization so it moves the scrolling unexpectedly.
- not allowing focus on another interactive element when a user clicks on that element while the signpost is open
- not moving focus to the content so screenreaders cannot read what is rendered in the signpost content

Issue Number: #3767 , #3762

## What is the new behavior?

- doesn't set focus on trigger initially
- sets focus back on trigger properly in appropriate cases
- using clrFocusOnViewInit on signpost content so the screenreader reads the signpost content when it opens

* [ ] Yes
* [x] No


src/clr-angular/popover/signpost/signpost-trigger.ts
```diff
@@ -32,26 +33,37 @@ export class ClrSignpostTrigger implements OnDestroy {
...
+    private injector: Injector,
     @Inject(PLATFORM_ID) private platformId: Object
   ) {
+    this.document = this.injector.get(DOCUMENT);
+  }
+
+  ngOnInit() {
+    this.signpostFocusManager.triggerEl = this.el.nativeElement;
     this.subscriptions.push(
       this.ifOpenService.openChange.subscribe((isOpen: boolean) => {
```
bdryanovski: I think this could be simplified a little.

```js
this.ifOpenService.openChange.subscribe((value: boolean) => {
  this.ariaExpanded = value;
  if (this.isOpen === undefined && typeof value === 'boolean') {
    this.isOpen = value;
    return;
  }

  this.isOpen = value;
  if (!this.isOpen && this.document.activeElement === this.document.body && isPlatformBrowser(this.platformId)) {
    this.singpostFocusManager.focusTrigger();
  }
}
```

* `this.isOpen` and `isOpen` look a lot alike especially in long lines and are hard to read.
*  `else` statement could get skipped.

mathisscott: Why are we setting `this.isOpen` to the same thing on each conditional branch? That doesn't make any sense.

Shijir: @bdryanovski I've simplified by using moving most of it into a new method. Also, added some comments explaining that we need to focus only if the signpost goes from a previously open state to closed state. Also, I think made the code reflect that more clearly: 

```
if (!this.isOpen && prevIsOpen) {
 this.focusOnClose();
}
```

@mathisscott we are setting `this.isOpen` to track values emitted by `this.ifOpenService.openChange`. And we need to set the focus only if we detect that the state is changed from open to closed.

src/clr-angular/popover/signpost/signpost-trigger.ts
```diff
@@ -32,26 +33,37 @@ export class ClrSignpostTrigger implements OnDestroy {
...
-          this.renderer.addClass(this.el.nativeElement, 'active');
+        this.ariaExpanded = isOpen;
+        // on initial value setting, we shouldn't move focus at all.
+        if (typeof this.isOpen === 'undefined' && typeof isOpen === 'boolean') {
+          this.isOpen = isOpen;
         } else {
-          this.renderer.removeClass(this.el.nativeElement, 'active');
-          if (isPlatformBrowser(this.platformId)) {
-            this.el.nativeElement.focus();
+          this.isOpen = isOpen;
```
mathisscott: Yes, but...

You are setting `this.isOpen` to the same value whether the conditional is true or false.

What is the point of the conditional?

Shijir: Before assigning new `isOpen` emitted value that to `this.isOpen`, we need to save the previous value somewhere (in this, it's saved to the variable `prevIsOpen`)
```
const prevIsOpen = this.isOpen;
this.isOpen = isOpen; 
```
so we can check if previous isOpen is false and new isOpen is true.

Also, `this.isOpen` is public property used for binding.

mathisscott: Mkay.

But that is not what this code is doing.

Here is the logic in this function...

```
if (x < 0) {
  y = 0;
} else {
  y = 0;
}

I don't see any logic here testing if the previous `this.isOpen` is true and the new one is `false`. I see logic that tests if the current value is undefined and the new one is a boolean.

And then setting the current value to the new value in BOTH cases.

File: src/clr-angular/popover/signpost/providers/signpost-focus-manager.service.ts [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class SignpostFocusManager {
+  private _triggerEl: HTMLElement;
+
+  set triggerEl(value: HTMLElement) {
+    this._triggerEl = value;
```
gnomeontherun: nit: I don't see value in this setter, why not just have a public property?

src/clr-angular/popover/signpost/signpost.spec.ts
```diff
@@ -57,6 +57,95 @@ export default function(): void {
...
+        let signpostContent: HTMLElement = this.hostElement.querySelector('.signpost-content');
+        expect(signpostContent).not.toBeNull();
+
+        // dynamic click doesn't set the focus so here manually focusing first
+        this.hostComponent.outsideClickBtn.nativeElement.focus();
+        this.hostComponent.outsideClickBtn.nativeElement.click();
+        this.detectChanges();
+
+        signpostContent = this.hostElement.querySelector('.signpost-content');
+        expect(signpostContent).toBeNull();
```
gnomeontherun: nit: don't see why we need a variable here when its only used once in both cases. This appears to be true in several tests, skip the local variable so it isn't confused.

src/clr-angular/popover/signpost/signpost-trigger.ts
Shijir: We first check if `this.isOpen` is `undefined`. If it's we will just set a new value to it. If it's no longer `undefined`, not only we still assign a new value, but also we call `this.signpostFocusManager.focusTrigger()`.  

But this part of the code has already been updated. I think the update makes it a bit clearer.

src/clr-angular/popover/signpost/signpost-trigger.ts
```diff
@@ -30,33 +31,52 @@ import { SignpostIdService } from './providers/signpost-id.service';
...
+        }
       }),
       this.signpostIdService.id.subscribe(idChange => (this.ariaControl = idChange))
     );
   }
 
+  private focusOnClose() {
+    // we have to set the focus back on the trigger only if the focus is reset back to the body element
+    // if the focus is on another element, we are not allowed to move that focus back to this trigger again.
+    if (!this.isOpen && this.document.activeElement === this.document.body && isPlatformBrowser(this.platformId)) {
```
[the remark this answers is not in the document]
Shijir: it's there. Or are we talking about the same thing?

src/clr-angular/popover/signpost/signpost-trigger.ts
```diff
@@ -30,33 +31,52 @@ import { SignpostIdService } from './providers/signpost-id.service';
...
+  public isOpen: boolean;
 
   constructor(
     private ifOpenService: IfOpenService,
-    private renderer: Renderer2,
     private el: ElementRef,
     public commonStrings: ClrCommonStringsService,
     private signpostIdService: SignpostIdService,
+    private signpostFocusManager: SignpostFocusManager,
+    private injector: Injector,
```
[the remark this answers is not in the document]
Shijir: I actually tried it. But for some reason, it wasn't giving me the document reference. I will check it again.

src/clr-angular/popover/signpost/signpost-content.ts
```diff
@@ -134,4 +141,10 @@ export class ClrSignpostContent extends AbstractPopover {
     this.popoverOptions.offsetY = setPosition.offsetY;
     this.popoverOptions.offsetX = setPosition.offsetX;
   }
+
+  ngOnDestroy() {
+    if (this.el.nativeElement.contains(this.document.activeElement) && isPlatformBrowser(this.platformId)) {
```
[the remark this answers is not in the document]
Shijir: I am checking it here

src/clr-angular/popover/signpost/providers/signpost-focus-manager.service.ts
Shijir: It's not passing el from one class to another so it doesn't need to be public. This `_triggerEl` is only for internal use of the service.

src/clr-angular/popover/signpost/signpost-trigger.ts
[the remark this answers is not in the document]
gnomeontherun: Put the platform check as the first conditional check in the if statement.

src/clr-angular/popover/signpost/signpost.spec.ts
Shijir: it's because the reference doesn't get updated when the dom gets changes. We need to retrieve it again to get the updated result for the selector.

src/clr-angular/popover/signpost/providers/signpost-focus-manager.service.ts
gnomeontherun: The setter is public (and when it compiles down its public), and even if you aren't ever getting it from elsewhere its overhead to privatize it IMO. Its a 'nit'.

src/clr-angular/popover/signpost/signpost.spec.ts
gnomeontherun: I know, my suggestion was to just query the selector twice in the expect instead of putting it in a variable first. That was confusing to me at first.

src/clr-angular/popover/signpost/signpost-content.ts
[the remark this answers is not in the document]
gnomeontherun: Put it before we try to access nativeElement, or this will still fail on SSR.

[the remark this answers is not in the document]
Shijir: True. Will update.

src/clr-angular/popover/signpost/signpost-trigger.ts
mathisscott: Your update makes it clearer. The previous code wasn't doing what you describe. But the new code appears to be. Thanks.

src/clr-angular/popover/signpost/providers/signpost-focus-manager.service.ts
Shijir: No that's why it's private and has setter because it makes it obvious that the element shouldn't be read from outside.

src/clr-angular/popover/signpost/signpost-content.ts
```diff
@@ -67,8 +71,11 @@ export class ClrSignpostContent extends AbstractPopover {
     this.position = 'right-middle';
     this.closeOnOutsideClick = true;
     this.signpostIdService.setId(signpostContentId);
+    this.document = this.injector.get(DOCUMENT);
```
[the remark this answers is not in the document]
Shijir: Yes, I will update this one too.

src/clr-angular/popover/signpost/signpost.spec.ts
Shijir: Done

PR #3662, opened 2019-08-02
feat: add aria-label for alert close button

bdryanovski:
Add `aria-label` that the developer could overwrite if needed to provide additional information on the close button text.

## PR Checklist

Please check if your PR fulfills the following requirements:

* [x] Tests for the changes have been added (for bug fixes / features)
* [x] Docs have been added / updated (for bug fixes / features)
* [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

* [ ] Bugfix
* [x] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [x] Documentation content changes
* [ ] clarity.design website / infrastructure changes
* [ ] Other... Please describe:

## What is the current behavior?
Right now there is no way to change the alert close button text - also we use generic `Close` text that could be misleading.

Issue Number: #3489 

## Does this PR introduce a breaking change?

* [ ] Yes
* [x] No

## Other information

Close: #3498 

**NEED TO BE BACKPORTED TO V1**

PR #3781, opened 2019-09-03
feat(dropdown) - add translatable aria-labels to clrDropdownTrigger

### LINKED ISSUE #3722: feat(dropdown): add translatable aria-labels to clrDropdownTrigger
## Summary

Implement translatable aria-labels to dropdowns.

* What is the change?

Add translatable aria-labels to clrDropdownTrigger button and change it based on the dropdown current state (open or close).

* Why should it go in Clarity?
Right now all dropdown doesn't have aria-label that is working out of the box and developer have to do it manually. Also, have to track the state of the dropdown by writing additional code.

No

## Use case
Right now adding aria-label to `clrDropdownTrigger` is done like this:

```html
<!-- Before open the dropdown -->
<button 
  class="nav-text" 
  clrDropdownTrigger 
  aria-label="Open dropdown"
>...</button>

<!-- After you open the dropdown you have to change it-->
<button 
  class="nav-text" 
  clrDropdownTrigger 
  aria-label="Close dropdown"
>...</button>
```
Quick workarround on that could be just to use the word `Toggle dropdown`. But this could not be ideal in all cases.

By adding `Input` we could:

* Simplify the implementation
* Automate it and write less code
* Get aria-label for all dropdowns out of the box

```html
<button 
  class="nav-text" 
  clrDropdownTrigger 
  clrAriaLabelDropdownOpen="open something something" 
  clrAriaLabelDropdownClose="close something something..."
>
```

```typescript
// dropdown-trigger.ts
host: {
  ...
 [attr.aria-label]: 'setAriaLabel',
  ...
}

get setAriaLabel() {
  return this.active ? this.clrAriaLabelDropdownOpen : this.clrAriaLabelDropdownClose;
}

...
@Input('clrAriaLabelDropdownOpen') clrAriaLabelDropdownOpen: string = commonStrings.keys.ariaLabelDropdownOpen;
@Input('clrAriaLabelDropdownClose') clrAriaLabelDropdownClose: string = commonStrings.keys.ariaLabelDropdownClose;
```

And the default for all dropdown will be
```typescript
// common-strings.default.ts
export const commonStringsDefault: ClrCommonStrings = {
  ...
  ariaLabelDropdownOpen: 'open dropdown',
  ariaLabelDropdownClose: 'close dropdown',
}
```

## Workarounds

* Add it by hand

AlexMarcuBytex:
closes #3722

Signed-off-by: Marcu Alexandru <alexandru.marcu@bytex.ro>

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3722 

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
+++ b/src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
@@ -12,15 +12,19 @@ import { spec, TestContext } from '../../utils/testing/helpers.spec';
 import { ClrDropdownTrigger } from './dropdown-trigger';
 import { ClrDropdown } from './dropdown';
 import { DROPDOWN_FOCUS_HANDLER_PROVIDER, DropdownFocusHandler } from './providers/dropdown-focus-handler.service';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 @Component({
   template: `
-    <button clrDropdownTrigger>Hello world</button>
+    <button clrDropdownTrigger [clrAriaLabelDropdownOpen]="openLabel" [clrAriaLabelDropdownClose]="closeLabel">Hello world</button>
   `,
   // These services are declared here because they need the renderer
   providers: [FocusService, DROPDOWN_FOCUS_HANDLER_PROVIDER],
 })
-class SimpleTest {}
+class SimpleTest {
+  openLabel: string;
+  closeLabel: string;
+}
 
 export default function(): void {
   describe('DropdownTrigger directive', function() {
@@ -48,6 +52,27 @@ export default function(): void {
 
     it('declares itself to the DropdownFocusHandler', function(this: Context) {
       expect(this.getClarityProvider(DropdownFocusHandler).trigger).toBe(this.clarityElement);
+    });
+
+    it('aria label should receive the correct value', function(this: Context) {
+      const ifOpenService = this.getProvider(IfOpenService);
+      const commonStringsService = this.getProvider(ClrCommonStringsService);
+
+      ifOpenService.open = false;
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe(commonStringsService.keys.dropdownOpenAriaLabel);
+
+      this.hostComponent.openLabel = 'New value for open';
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('New value for open');
+
+      ifOpenService.open = true;
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe(commonStringsService.keys.dropdownCloseAriaLabel);
+
+      this.hostComponent.closeLabel = 'New value for close';
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('New value for close');
     });
   });
 }
--- a/src/clr-angular/popover/dropdown/dropdown-trigger.ts
+++ b/src/clr-angular/popover/dropdown/dropdown-trigger.ts
@@ -3,12 +3,13 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Directive, ElementRef, HostListener } from '@angular/core';
+import { Directive, ElementRef, HostListener, Input } from '@angular/core';
 
 import { IfOpenService } from '../../utils/conditional/if-open.service';
 
 import { ClrDropdown } from './dropdown';
 import { DropdownFocusHandler } from './providers/dropdown-focus-handler.service';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 @Directive({
   // We support both selectors for legacy reasons
@@ -20,6 +21,7 @@ import { DropdownFocusHandler } from './providers/dropdown-focus-handler.service
     '[class.active]': 'active',
     '[attr.aria-haspopup]': '"menu"',
     '[attr.aria-expanded]': 'active',
+    '[attr.aria-label]': 'ariaLabel',
   },
 })
 export class ClrDropdownTrigger {
@@ -29,7 +31,8 @@ export class ClrDropdownTrigger {
     dropdown: ClrDropdown,
     private ifOpenService: IfOpenService,
     el: ElementRef<HTMLElement>,
-    focusHandler: DropdownFocusHandler
+    focusHandler: DropdownFocusHandler,
+    private commonStringsService: ClrCommonStringsService
   ) {
     // if the containing dropdown has a parent, then this is not the root level one
     if (dropdown.parent) {
@@ -42,8 +45,24 @@ export class ClrDropdownTrigger {
     return this.ifOpenService.open;
   }
 
+  get ariaLabel() {
+    return this.active ? this.clrAriaLabelDropdownClose : this.clrAriaLabelDropdownOpen;
+  }
+
   @HostListener('click', ['$event'])
   onDropdownTriggerClick(event: any): void {
     this.ifOpenService.toggleWithEvent(event);
+  }
+
+  @Input('clrAriaLabelDropdownOpen') clrAriaLabelDropdownOpen: string;
+  @Input('clrAriaLabelDropdownClose') clrAriaLabelDropdownClose: string;
+
+  ngOnInit() {
+    if (!this.clrAriaLabelDropdownOpen) {
+      this.clrAriaLabelDropdownOpen = this.commonStringsService.keys.dropdownOpenAriaLabel;
+    }
+    if (!this.clrAriaLabelDropdownClose) {
+      this.clrAriaLabelDropdownClose = this.commonStringsService.keys.dropdownCloseAriaLabel;
+    }
   }
 }
--- a/src/clr-angular/utils/i18n/common-strings.default.ts
+++ b/src/clr-angular/utils/i18n/common-strings.default.ts
@@ -47,6 +47,9 @@ export const commonStringsDefault: ClrCommonStrings = {
   detailExpandableAriaLabel: 'Toggle more row content',
   // Alert
   alertCloseButtonAriaLabel: 'Close alert',
+  //Dropdown
+  dropdownOpenAriaLabel: 'Open dropdown',
+  dropdownCloseAriaLabel: 'Close dropdown',
   // Date Picker
   datepickerToggle: 'Toggle datepicker',
   datepickerPreviousMonth: 'Previous month',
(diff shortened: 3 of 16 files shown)


src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
```diff
@@ -49,5 +49,23 @@ export default function(): void {
     it('declares itself to the DropdownFocusHandler', function(this: Context) {
       expect(this.getClarityProvider(DropdownFocusHandler).trigger).toBe(this.clarityElement);
     });
+
```
bdryanovski: It will be good to test that the component will get its default aria-label value, not only that is `undefined`. 

This is also true `'' !== undefined`

src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
```diff
@@ -49,5 +49,23 @@ export default function(): void {
...
+    it('default value should be added when inputs are not receiving data', function(this: Context) {
+      expect(this.clarityDirective.clrAriaLabelDropdownOpen).not.toBeUndefined();
+      expect(this.clarityDirective.clrAriaLabelDropdownClose).not.toBeUndefined();
+    });
+
+    it('value received through inputs should be set to aria-label', function(this: Context) {
+      const ifOpenService = this.getProvider(IfOpenService);
+      ifOpenService.open = false;
+      this.clarityDirective.clrAriaLabelDropdownOpen = 'New value for open';
+      this.clarityElement.setAttribute('aria-label', this.clarityDirective.setAriaLabel);
```
bdryanovski:
```
      this.clarityElement.setAttribute('aria-label', this.clarityDirective.setAriaLabel);
```
This is the thing that you want to test, but you do it instead of the component. Better will be to just set the `Input`  and let the component do the job, by setting the correct value for `aria-label` and pass whatever flows it needs to do. For example, `setAriaLabel` could have additional logic that may or may not change the value.

```
@Component({
  template: `
    <button clrDropdownTrigger>Hello world</button>
  `,
  // These services are declared here because they need the renderer
  providers: [FocusService, DROPDOWN_FOCUS_HANDLER_PROVIDER],
})
class SimpleTest {}
```

Create or update the above template to include `clrAriaLabelDropdownClose` and `clrAriaLabelDropdownOpen` as inputs and change them to see if they apply the correct behavior. There are a lot of examples out there.

AlexMarcuBytex: i ll keep the template as it is because i need it without inputs for the first test. i ll just set the new values of the inputs and use detectchanges() for the second.
```const ifOpenService = this.getProvider(IfOpenService);
      ifOpenService.open = false;
      this.clarityDirective.clrAriaLabelDropdownOpen = 'New value for open';
      this.detectChanges();
      expect(this.clarityElement.getAttribute('aria-label')).toBe('New value for open');
```
How s this way? It is ok?

bdryanovski: By default, the component should set the default value. And you could change attributes this way.

```js
@Component({
  template: `
    <button clrDropdownTrigger [clrAriaLabelDropdownOpen]="openLabel" [clrAriaLabelDropdownClose]="closeLabel" >Hello world</button>
  `,
  // These services are declared here because they need the renderer
  providers: [FocusService, DROPDOWN_FOCUS_HANDLER_PROVIDER],
})
class SimpleTest {
  openLabel: string
  closeLabel: string
}
```

Inside the test you could update and change the value:

```js
expect(this.clarityElement.getAttribute('aria-label')).toBe('default text when no value is passed');
this.testComponent.openLabel = "some text";
this.detectChanges();
expect(this.clarityElement.getAttribute('aria-label')).toBe('some text');
```

src/clr-angular/utils/i18n/common-strings.default.ts
```diff
@@ -47,6 +47,9 @@ export const commonStringsDefault: ClrCommonStrings = {
   detailExpandableAriaLabel: 'Toggle more row content',
   // Alert
   alertCloseButtonAriaLabel: 'Close alert',
+  //Dropdown
+  ariaLabelDropdownOpen: 'Open dropdown',
```
gnomeontherun: We want to keep same naming conventions, so let's make these `dropdownOpenAriaLabel` and `dropdownCloseAriaLabel`. This naming will change across the whole PR so I'm just making this one comment for that.

src/clr-angular/popover/dropdown/dropdown-trigger.ts
```diff
@@ -42,8 +45,24 @@ export class ClrDropdownTrigger {
...
+  get setAriaLabel() {
+    return this.active ? this.clrAriaLabelDropdownClose : this.clrAriaLabelDropdownOpen;
+  }
+
   @HostListener('click', ['$event'])
   onDropdownTriggerClick(event: any): void {
     this.ifOpenService.toggleWithEvent(event);
   }
+
+  @Input('clrAriaLabelDropdownOpen') clrAriaLabelDropdownOpen: string;
```
gnomeontherun: What we can do is the same as for the Alert close button PR: https://github.com/vmware/clarity/pull/3662/files#diff-c844262cb1b2d10864f118f627115edcR31

`@Input() clrDropdownOpenAriaLabel: string = this.commonStrings.keys.dropdownOpenAriaLabel;` and you don't have to have the `ngOnInit` here. Anything locally provided will override the defaults then.

src/clr-angular/popover/dropdown/dropdown-trigger.ts
```diff
@@ -42,8 +45,24 @@ export class ClrDropdownTrigger {
     return this.ifOpenService.open;
   }
 
+  get setAriaLabel() {
```
gnomeontherun: nit, naming is a little strange to me for a property, I'd like this to just be `ariaLabel` since this method is not setting anything directly.

src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
```diff
@@ -49,5 +52,25 @@ export default function(): void {
...
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('Open dropdown');
+
+      this.hostComponent.openLabel = 'New value for open';
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('New value for open');
+
+      ifOpenService.open = true;
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('Close dropdown');
```
gnomeontherun: Should grab the common strings service and test against the default value instead of hard coding it here inside of toBe, same for next expect.

src/clr-angular/popover/dropdown/dropdown-trigger.ts
```diff
@@ -3,12 +3,13 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Directive, ElementRef, HostListener } from '@angular/core';
+import { Directive, ElementRef, HostListener, Input, SimpleChanges } from '@angular/core';
```
gnomeontherun: Don't need `SimpleChanges`

src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
```diff
@@ -49,5 +52,25 @@ export default function(): void {
...
+
+      this.hostComponent.closeLabel = 'New value for close';
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('aria-label')).toBe('New value for close');
+    });
```
gnomeontherun: Can we add a controller test that tests just the `ariaLabel` getter controller property that it returns the value based on if it is open or not?

src/clr-angular/popover/dropdown/dropdown-trigger.ts
AlexMarcuBytex: I changed the way i set the default value because the tests were failing. If you have the case when you pass a string variable that is undefined to the input ( like in the test ), the default value wont be set. 
I see this assigning ```@Input() clrDropdownOpenAriaLabel: string = this.commonStrings.keys.dropdownOpenAriaLabel``` 
only works if you dont use the input in html at all. But if you want to set the aria label in the case i told you above, the solution i found is using onInit.

src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
AlexMarcuBytex:
```   
      const ifOpenService = this.getProvider(IfOpenService);
      ifOpenService.open = false;
      this.detectChanges();
      expect(this.clarityDirective.ariaLabel).toBe(this.clarityDirective.clrAriaLabelDropdownOpen);
```
Like this but for both cases?

src/clr-angular/popover/dropdown/dropdown-trigger.ts
bdryanovski: I'm not sure that understand the issue here when it comes to the tests and testing the inputs you need two simple tests.

* test to see if the component/directive will set the default value when you don't pass anything to it. ( `[clrDropdownOpenAriaLabel]="undefined"` could be seen as not passed at all )
* test to see that when passing `clrDropdownOpenAriaLabel` it will change the `aria-label` at the dom. 

In both cases, checking to see that the `this.clrDropdownOpenAriaLabel`, is not enough. We need to be sure that the correct value is set at the DOM.

This way you gonna check the following things:

* Input exists and is working
* `ariaLabel` is setting the correct string
* host element is getting the correct `aria-label` value.

AlexMarcuBytex: Thanks for the tips! The problem is that `[clrDropdownOpenAriaLabel]="undefined"`  isnt the same thing as not passing a value. If you pass any reference to the input, that will overrides the default set. So the directive will never set a default value if there s a null or undefined reference. One solution to catch that case is to use ngOnInit. 
What i wanted to say is that the tests for the default value will always crash in this case.

AlexMarcuBytex: Tell me if i get it correctly. Should I consider that there wont be inputs that are undefined?  And also make tests for inputs to be expected as defined?

src/clr-angular/popover/dropdown/dropdown-trigger.spec.ts
gnomeontherun: Yes

gnomeontherun:
See https://github.com/vmware/clarity/issues/3722#issuecomment-528098193.

Closing this as not a feature we want to build.

PR #3787, opened 2019-09-04
fix: focus stealing in modal by using clrFocusOnViewInit

Shijir:
- This PR extends clrFocusOnViewInit by making it disablable

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

- the modal heading steals focus when it opens. Some of the demos we have use open modals so they steal focus
- the modal heading shouldn't be keyboard focusable so it should have `tabindex = "-1"` instead of `tabindex = "0"`.
- Users cannot disable `clrFocusOnViewInit`.

Issue Number: #3762 

## What is the new behavior?

- the modal heading doesn't steal focus in alerts demo now as we are intercepting the default `FOCUS_ON_VIEW_INIT` service and overrides the value to `false` in the demo component now.

- Users can now disable `clrFocusOnViewInit` directive now two ways: 
  1.  `[clrFocusOnViewInit]="false" // default value is true`.
  2.  Via intercepting injection as mentioned above.

* [ ] Yes
* [x] No


src/clr-angular/index.ts
```diff
@@ -21,5 +21,7 @@ export * from './accordion/stepper/index';
 export * from './progress/spinner/index';
 export * from './progress/progress-bars/index';
 
+export { FOCUS_ON_VIEW_INIT } from './utils/focus/focus-on-view-init/focus-on-view-init.provider';
```
gnomeontherun: Can we just export this through the utils directory instead of explicitly here? We want to avoid this.

src/clr-angular/modal/modal.ts
```diff
@@ -9,22 +9,18 @@ import {
...
-  ElementRef,
-  PLATFORM_ID,
 } from '@angular/core';
-
 import { FocusTrapDirective } from '../utils/focus-trap/focus-trap.directive';
-import { ScrollingService } from '../utils/scrolling/scrolling-service';
 import { ClrCommonStringsService } from '../utils/i18n/common-strings.service';
 import { UNIQUE_ID, UNIQUE_ID_PROVIDER } from '../utils/id-generator/id-generator.service';
-import { isPlatformBrowser } from '@angular/common';
+import { ScrollingService } from '../utils/scrolling/scrolling-service';
```
gnomeontherun: Unused?

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.module.ts
```diff
@@ -6,11 +6,14 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 
+import { FOCUS_ON_VIEW_INIT_PROVIDER } from './focus-on-view-init.provider';
+
 import { FOCUS_ON_VIEW_INIT_DIRECTIVES } from './index';
```
gnomeontherun: We don't want to define anything in the index file, only reexport. That was one thing that caused some problems in the past, can you move those into here?

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
```diff
@@ -43,21 +46,34 @@ export class ClrFocusOnViewInit implements AfterViewInit {
...
-    if (isPlatformBrowser(this.platformId)) {
-      if (this.document.activeElement === this.el.nativeElement) {
-        return;
-      }
-      if (this.el.nativeElement) {
+    this.focus();
+  }
+
+  private focus() {
+    if (!isPlatformBrowser(this.platformId)) {
```
gnomeontherun: nit: we can just condense into one `if` block

src/clr-angular/index.ts
Shijir: Yeah, I think we can do that.

src/clr-angular/modal/modal.ts
Shijir: This is still in use. It's just the line position that's been changed.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.module.ts
Shijir: We can do that. But several of our components and utility directives are using this pattern.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
Shijir: I think these multiple levels of if statements would convey checks in different categories and priorities more clearly: 1. Check platform 2. Check enabled 3. Check focused.

Also, I was avoiding a long if statement.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.module.ts
gnomeontherun: Anything that we publicly export into public_api does not have that pattern, there are some old ones in utilities we don't export that weren't refactored and they are ok.

Shijir: Ah I see. I made these changes looking at the focus-trap directive. So that makes sense.

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.spec.ts
```diff
@@ -19,67 +21,137 @@ import { ClrFocusOnViewInitModule } from './focus-on-view-init.module';
...
+    });
 
-  it('sets focus on the directive element on view intialization', () => {
-    component.buttonElRef.nativeElement.focus();
-    expect(document.activeElement).toBe(component.buttonElRef.nativeElement);
-    component.displayNoExistingTabindex = true;
-    fixture.detectChanges();
-    expect(document.activeElement).toBe(component.focusOnItElRef.nativeElement);
-  });
+    it('sets focus on the directive element on view intialization', () => {
```
jeeyun: typo: initialization

src/clr-angular/utils/focus/focus-on-view-init/focus-on-view-init.ts
```diff
@@ -25,10 +26,17 @@ export class ClrFocusOnViewInit implements AfterViewInit {
   constructor(
     private el: ElementRef,
     @Inject(PLATFORM_ID) private platformId: Object,
-    private injector: Injector,
+    @Inject(FOCUS_ON_VIEW_INIT) private focusOnViewInit: boolean,
+    @Inject(DOCUMENT) document: any,
```
gnomeontherun: We can't just make the `private` here instead of assigning it below?

gnomeontherun:
I think the `disabled` attribute request makes sense, but also is beyond scope here and something that apps can avoid anyways by not mixing these things.

Shijir:
Yeah, I think it's better to extend the directive with such an ability to recognize `disabled` attribute in a separate PR.

Issue #3483, opened 2019-06-19
Datagird numeric filters don't work & have undefined placeholders


amellnik:
## Describe the bug

Datagrid numeric filters don't have the right placeholders:

From the stackblitz linked below:
![Screen Shot 2019-06-19 at 1 02 17 PM](https://user-images.githubusercontent.com/7132414/59796502-8376eb00-9292-11e9-8ceb-533e55090411.png)

Seen in the wild: 
![Screen Shot 2019-06-19 at 12 51 54 PM](https://user-images.githubusercontent.com/7132414/59796506-85d94500-9292-11e9-9279-99d3c465c48e.png)

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v2-nm7acf

1.  Open the filter on the `wins` column

## Expected behavior

Inputting values on the numeric filters should result in the filters being applied.

The placeholders should say `Min value` and `Max value`.  

## Versions

**App**

* Angular: 8
* Node:  12
* Clarity: 2.0.1

**Device:**

* Type: Mac
* OS: MacOS
* Browser Chrome
* Version 74

## Additional notes

At first I thought this was just my project since the placeholders do show up correctly in the docs: https://clarity.design/documentation/datagrid/binding-properties

![Screen Shot 2019-06-19 at 12 52 53 PM](https://user-images.githubusercontent.com/7132414/59796596-b8833d80-9292-11e9-98e1-221611881c60.png)

However, this also shows up in a fresh stackblitz so maybe something else is going on.  This was probably introduced by me with https://github.com/vmware/clarity/pull/3397 but I'm not sure how.  

Edit:  I realized that the filters don't even work.  I think something changed during the angular 8 migration / clarity 2.0 release that I missed which broke these.  In particular, the filters work fine on the documentation website so it looks like there's either some sort of dependency difference or something changed since the website was deployed.  

Edit2: The filters work fine on master when running `npm start` or `npm website:start`, but building the packages and using them in a sample project (like the stackblitz above) causes them not to work, so it's something in the build/release step.

bdryanovski:
Affected versions are `2.0.0`, `2.0.1`, `2.0.2`, `2.1.0`.

It seems that `commonStrings` is an empty object. If my memory serves me well, this was a known issue that is resolved with the change of how we provide the translation and the new service.

https://github.com/vmware/clarity/pull/3660
https://clarity.design/documentation/internationalization

This is resolve in all version after and including `2.1.1`

[...]

Issue #3486, opened 2019-06-20
[Datagrid] The 2nd custom string filter doesn't work


amellnik:
## Describe the bug

Adding custom string filters to columns in a Datagrid works for every column except for the 2nd.  

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v2-cwbulg?file=src%2Fapp%2Fapp.component.html

1.  Try adding filter values to any of the four columns.  The filters work properly for every column except the 2nd.  

## Expected behavior

All columns work.  

## Versions

**App**

* Angular: 8
* Node: 12
* Clarity: 2

**Device:**

* Type: Mac
* OS: MacOS
* Browser: Chrome (also Firefox)
* Version 47

## Additional notes

This might be related to https://github.com/vmware/clarity/issues/3282

hippee-lee:
Question posted on Stack overflow that might be related: 
https://stackoverflow.com/questions/57066356/clarity-datagrid-custom-filtering-isnt-working-with-more-than-one-filter
Stackblitz: https://stackblitz.com/edit/clarity-light-theme-v2-dwqmgr

spitfjre:
I can confirm this problem in all versions since 2.0.0, that only the first custom filter appears in the `state.filters` object. For all others `state.filters` is `undefined`.

kathygit:
Hi, I wonder if there is a date for it get fixed. This issue prevents us from use Clr 2.0. Thank you!

kathygit:
I dug into the code ... Here is the reason:
When we use clr-dg-string-filter and clr-dg-numeric-filter, we don't project new content, but alter the accepts() to define our own filter criteria. 
In this case _this.customFilter = false

```
var ClrDatagridColumn = /** @class */ (function (_super) {
    __extends(ClrDatagridColumn, _super);
    function ClrDatagridColumn(_sort, filters, vcr, commonStrings) {
        var _this = _super.call(this, filters) || this;
       .....
           /**
         * A custom filter for this column that can be provided in the projected content
         */
        **_this.customFilter = false;**
```

kathygit:
I am not sure what the fix should be, maybe in DatagridFilterRegistrar ?
We shall add a check to search the registrar by filter name ? 

```
DatagridFilterRegistrar.prototype.deleteFilter = function () {
        if (this.registered) {
            this.registered.unregister();
            delete this.registered;
        }
    };
```

kathygit:
or, just remove this block
```
[...]

PR #3825, opened 2019-09-16
fix(datagrid): numeric filters are not working 

bdryanovski:
Numeric filter was not working when you only use binding input

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Changing the order of `@Input` fix the issue with `colType` beeing undefined at the time when `numeric` filter is selected.

Issue Number: #3483

* [ ] Yes
* [x] No

## Other information
Close: #3483

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.ts
@@ -47,7 +47,7 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
                   *ngIf="field && !customFilter && (colType=='string')"
                   [clrDgStringFilter]="registered"
                   [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
-          
+
           <clr-dg-numeric-filter
                   *ngIf="field && !customFilter && (colType=='number')"
                   [clrDgNumericFilter]="registered"
@@ -57,11 +57,11 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
               <ng-content></ng-content>
           </ng-template>
 
-          <button 
-            class="datagrid-column-title" 
+          <button
+            class="datagrid-column-title"
             [attr.aria-label]="commonStrings.keys.sortColumn"
-            *ngIf="sortable" 
-            (click)="sort()" 
+            *ngIf="sortable"
+            (click)="sort()"
             type="button">
               <ng-container  *ngTemplateOutlet="columnTitle"></ng-container>
               <clr-icon
@@ -129,6 +129,15 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
     return this._field;
   }
 
+  /*
+  * What type is this column?  This defaults to STRING, but can also be
+  * set to NUMBER.  Unsupported types default to STRING. Users can set it
+  * via the [clrDgColType] input by setting it to 'string' or 'number'.
+  */
+
+  // TODO: We might want to make this an enum in the future
+  @Input('clrDgColType') colType: 'string' | 'number' = 'string';
+
   @Input('clrDgField')
   public set field(field: string) {
     if (typeof field === 'string') {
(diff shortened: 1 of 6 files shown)


bdryanovski:
Not sure how to test it. 

The issue is that they have to be in a specific order so they will be defined. Right now it won't depend that is set or not it will work as it should work. It will get the default value of `'string'` if it's not set. And the internal assignment inside ` @Input('clrDgColType')` will go before `@Input('clrDgField')` that require to have the value of `this.colType` for it's internal logic.

If you have some idea of a test that could be meaningful not just "a test"  

I see that there are no tests covering the `clrDgColType` input so I could try to cover it. But still think that they won't catch this type of bug in the feature

Shijir:
@bdryanovski The reason why we have the bug in the first place is that it could be either we had incomplete testing or wrong testing, right? And I looked it up where we are testing this numeric filter. And I've just found out that we are not testing it at all.

bdryanovski:
@Shijir Added a few tests

Shijir:
I suspect that when the original PR submitter(I think it was an external contributor) introduced this PR, he forgot to include the feature testing in `all.spec.ts` file. Look up where `datagrid-numeric-filter.spec.ts` is imported. It's not imported in anywhere. The feature wasn't tested at all. We need to import the tests from that file and run them.

bdryanovski:
It seems that this is the case
```
      DatagridNumericFilter component
        ✖ receives an input for the filter value
        ✖ wires the RegisteredFilter correctly
        ✖ receives an input for the filter logic
        ✖ registers a filter
        ✖ registers itself as a CustomFilter provider
        ✖ displays numeric inputs when open
        ✖ focuses on the input when the filter opens
        ✖ offers two way binding on the filtered state
      DatagridNumericFilterImpl
        ✔ becomes active when either high or low limits are set
        ✖ filters numbers that are above the high limit
        ✖ filters numbers that are below the low limit
        ✖ only allows values within the range when both filters are set
        ✔ exposes state
        ✖ compares filters
        with DatagridPropertyNumericFilter
          ✔ exposes state
          ✖ compares filters
```
I have to inspect them and see they are not working

amellnik:
@Shijir -- This one is on me.  My PR was in review when Clarity migrated to Angular 8 and it looks like I screwed up the rebase.  Sorry about that!
@bdryanovski Thanks!

gnomeontherun:
@bdryanovski can you provide more detail about the problem you found, and how you fixed it before we merge this?

bdryanovski:
Sure,

The code before my change

```typescript
 template: `
          <clr-dg-string-filter
                  *ngIf="field && !customFilter && (colType=='string')"
                  [(clrFilterValue)]="filterValue"></clr-dg-string-filter>
          
          <clr-dg-numeric-filter
                  *ngIf="field && !customFilter && (colType=='number')"
                  [clrDgNumericFilter]="registered"
                  [(clrFilterValue)]="filterValue"></clr-dg-numeric-filter>
`
 // ...
 @Input('clrDgField')
  public set field(field: string) {
    if (typeof field === 'string') {
      this._field = field;
      if (!this.customFilter) {
        if (this.colType === 'number') {
          this.setFilter(new DatagridNumericFilterImpl(new DatagridPropertyNumericFilter(field)));
        } else {
          this.setFilter(new DatagridStringFilterImpl(new DatagridPropertyStringFilter(field)));
        }
      }
      if (!this._sortBy) {
        this._sortBy = new DatagridPropertyComparator(field);
      }
    }
  }

// ...

@Input('clrDgColType') colType: 'string' | 'number' = 'string';

```
When we check the value of `this.colType` inside the `field` setter the value of `colType` is always `undefined` and we always set the `DatagridPropertyStringFilter` but in the same time, the template is rendered last so we show the number filter.

```
1. @Input('clrDgField')
2. @Input('clrDgColType')
3. render()
```

So in the end, we have a visible filter, but is not register and so the changes are not applied to the list.

Shijir:
Does this PR also fix https://github.com/vmware/clarity/issues/3486?

kathygit:
This issue will not fix #3486, they r 2 different issues.

bdryanovski:
@Shijir no, this is only fixing the issue that you could not enable numeric filters.

#3486 - is something related to how the filters are registered. When you remove the `[clDgField]` the filters start to work but I think that this is not the solution that we searching for.

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -129,6 +129,15 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
     return this._field;
   }
 
+  /*
```
gnomeontherun: I don't see any actual changes in this file, can you revert it?

bdryanovski: This is the most important change in the pull-request moving the `clrDgColType` above `clrDgField`

bdryanovski: Also alignment of the comment block - but that is not that important

PR #3830, opened 2019-09-17
fix: form a11y

### LINKED ISSUE #3556: Screen readers may miss inline errors on form controls
## Describe the bug

Screen reader users may miss the inline errors. While the inline errors are readable using the reading cursor, screen reader users do not hear a status update telling them that the application is in an error state. Therefore, they will not know to look for the inline error messages. Furthermore, since the inline errors are not programmatically associated with their respective fields, screen reader users will not hear the errors when they read the fields as expected while the form is an an error state.

## How to reproduce

- Use screen reader
- Activate the Form Demos link
- Press the Tab key to focus the text field in the 'Helper and validation messages' example
- Press the Tab key again or press Shift + Tab
- Observe that you do not hear the following - 'Error message that appears after focus is lost and control is invalid'

Using VoiceOver's reading cursor

- Refresh the page
- Read the text field in the 'Helper and validation messages' example using Ctrl + Shift + left or right arrow key
- Move the reading cursor away from the text field using Ctrl + Shift + left or right arrow key
- Observe that you do not hear the following - 'Error message that appears after focus is lost and control is invalid'

## Expected behavior

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes
- [x] Use aria-describedby (https://www.w3.org/TR/wai-aria-1.1/#aria-describedby) to enable screen readers to announce error specific information when reading the fields in error by programmatically associating the errors with their respective fields. 
- [ ] Use an assertive live region to announce a status message for the entire form like "There are errors with the following fields: field 1, field 2 ..."
- [ ] If there are multiple errors, focus should be moved to first field in error

hippee-lee:
fixes #3556

fix: label test for form labels array

## PR Checklist

## PR Type

The change addresses two a11y issues with forms:
1) The form is submitted and there is an error focus should be moved to the first control that is in error state.
2) When the form is submitted there should be an aria live element that has a summary of the controls in error state.

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

No aria live and no focus when a submitted form has errors. 

Issue Number: 3556

## What is the new behavior?
When a submitted form has errors the first control wiht an error gets focus and the aria live region updates with a list of controls that have errors. 

* [ ] Yes
* [x] No


src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +49,52 @@ export class ClrForm {
...
+
+  ngOnInit() {
+    // ClrForms need to have aria live element for screen readers.
+    this.ariaLiveElement = document.createElement('div');
+    this.ariaLiveElement.setAttribute('aria-live', 'polite');
+    this.ariaLiveElement.setAttribute('role', 'status');
+    this.ariaLiveElement.setAttribute('aria-atomic', 'true');
+    this.ariaLiveElement.setAttribute('aria-relevent', 'text');
+    this.ariaLiveElement.classList.add('clr-sr-only');
+    this.renderer.insertBefore(this.el.nativeElement, this.ariaLiveElement, this.el.nativeElement.firstChild);
```
Shijir: All the above DOM methods are native DOM methods, why is this one alone using Angular renderer2 here?

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +49,52 @@ export class ClrForm {
   markAsTouched() {
     this.markControlService.markAsTouched();
   }
+
+  @HostListener('submit')
+  onFormSubmit() {
+    this.markAsTouched();
+    // https://github.com/angular/angular/issues/12463
+    if (isPlatformBrowser(this.platformId)) {
+      const invalidControl = this.el.nativeElement.querySelector('.ng-invalid');
```
Shijir: Why are we using native querying here? Also, we shouldn't rely on Angular's internal class, `.ng-invalid`.
I see that we have a pretty nice structure in our `clrForm`. So can we use Angular querying instead:

```
@ContentChildren(ClrInputContainer) inputContainers: QueryList<ClrInputContainer>;


// and then somewhere in a method

const invalidInputControllers = inputContainers.filter(inputContainer => inputContainer.invalid);

invalidInputControllers[0].setFocusOnInput(); // we can easily add this method in the input container component
```

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +49,52 @@ export class ClrForm {
...
+  }
+
+  private resetAriaLive() {
+    this.ariaLiveElement.innerHTML = '';
+  }
+
+  private updateAriaLive(): void {
+    let updatedLiveText = this.commonStrings.keys.formErrorSummary;
+    this.resetAriaLive();
+    const invalidControls = Array.prototype.slice.call(this.el.nativeElement.querySelectorAll('.ng-invalid'));
```
Shijir: The same issue here. This method is using native querying also relying on Angular's internal class. Also, I am really trying to understand what's going on here in this method. Is it trying to get all error messages and append them? Can we use:

```
@ContentChildren(ClrControlError, {descendants: true, read: ElementRef }) errors: QueryList<ElementRef>;

// then use Angular query
this.errors.map((erEl) => erEl.nativeElement.textContent) // this will give you all error messages in array.


```

src/clr-angular/forms/common/form.ts
hippee-lee: I didn't realize that the renderer could work on elements that aren't in the DOM already I'll switch to use the renderer.

src/clr-angular/forms/common/form.ts
hippee-lee: I think the issue with using content children is there would need to be one for each control type (select, radio, input etc). Moving to @ContentChildren introduces some things forms would have to do accounting for:
1. Which controls are invalid - the filter would work but ... there would have to be an @ContentChildren for every form control type (excluding combobox which doesn't have a container yet)
2. How would I know which control is the first invalid control in the Form if they are all separated into different @ContentChildren? .

src/clr-angular/forms/common/form.ts
hippee-lee: For every invalid control I need the projected content of the label (not the error message) so I can append it to the aria live region. The aria live region is supposed to read out a summary but not all of the details for each error.

src/clr-angular/forms/common/form.ts
Shijir: I was suggesting to `@ContentChildren` in `ClrForm` directive. The code snippet I am showing would be used in `ClrForm`. Once you filter only invalid input containers and have them in array, you could do `invalidInputControllers[0]`, which is the first invalid input container. 

And do we even need to concern about the ones without a container? Those ones wouldn't have `clr-control-error`, right?

If you are concerned with inputs not nested in a container, we could directly query `clrInput`, and filter the invalid ones with `descendants: true` flag in `ContentChildren`

src/clr-angular/forms/common/form.ts
hippee-lee: I left comments but they may only make sense to me. 
I'll try to explain again. 

1. I create an array of all invalid controls on the form (invalidControls)
2. then I make an array of only their id's (errorIDs)
3. Then I figure out which text to append from the values stored in the `this.labels`. 
4. Once I have an array or only the label text I can append it to a string that uses the formErrorSummary common string and update the aria live region. 

If forms switches to @ContentChildren this would change. But that depends on the other behavior that has to happen here which is putting focus on the first control that has an error state.

src/clr-angular/forms/common/form.ts
hippee-lee: I don't have a concern with inputs not nested in an clr-input-container - I don't think this behavior should be supported for controls that are not in a container But I also have to account for the other control containers (clr-checkbox-container, clr-toggle-container, clr-date-container, clr-password-container, clr-radio-container, clr-select-container, and clr-textarea-container).  Unless I'm mistaken, it is not possible to have one ContentChildren with multiple Types so forms.ts would end up with multiple @ContentChildren. 

I'm ok with that but would have to modify the logic to find a different way to figure out which control is first so it could be focused. I'm also ok with adding the focus method to the containers but they would all need the change.

Shijir: ok I see what you are saying now. Sorry, I assumed we are `clrInput` in all control container components.

Shijir: Ok I have an idea on to avoid using both querySelector and `.ng-invalid`. Let's discuss that on Slack.

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,74 @@ export class ClrForm {
...
+    this.ariaLiveText = this.commonStrings.keys.formErrorSummary;
+    this.resetAriaLive();
+
+    this.ariaLiveElement.textContent =
+      this.commonStrings.keys.formErrorSummary +
+      // Only use labels that have a 'for' attribute
+      this.labels
+        .filter(label => label.forAttr !== undefined)
+        // Only use labels that have a control id that matches the for attribute
+        .filter(label => this.invalidControls.filter(control => control.id === label.forAttr))
```
Shijir: > .filter(label => this.invalidControls.filter(control => control.id === label.forAttr))

Is this filter really working? Because the nested `.filter` will always return a new array. Even when there is no match found, it still returns a new array that is empty, which is still truthy.

src/clr-angular/forms/common/label.ts
```diff
@@ -27,6 +42,12 @@ export class ClrLabel implements OnInit, OnDestroy {
   private subscriptions: Subscription[] = [];
   private enableGrid = true;
 
+  get labelText(): string {
+    if (isPlatformBrowser(this.platformId)) {
+      return this.el.nativeElement.textContent;
```
Shijir: We don't really need to check platform here. Even if `.textContent` itself turns out to be `undefined`, there is no harm as we don't try to access and use deeper nested properties from `.textContent`.

src/clr-angular/forms/common/label.spec.ts
```diff
@@ -14,6 +14,16 @@ import { ClrLabel } from './label';
 import { ControlIdService } from './providers/control-id.service';
 import { Layouts, LayoutService } from './providers/layout.service';
 import { NgControlService } from './providers/ng-control.service';
+import { MarkControlService } from './providers/mark-control.service';
+import { ClrForm } from './form';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
+
+export class MockElementRef extends ElementRef {
```
Shijir: Is this mock class being used?

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,74 @@ export class ClrForm {
...
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-atomic', 'true');
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-relevent', 'text');
+      this.renderer.addClass(this.ariaLiveElement, 'clr-sr-only');
+      this.renderer.insertBefore(this.el.nativeElement, this.ariaLiveElement, this.el.nativeElement.firstChild);
+    }
+  }
+
+  private removeAriaLiveElement() {
+    if (isPlatformBrowser(this.platformId)) {
+      this.renderer.removeChild(this.el.nativeElement, this.ariaLiveElement);
```
Shijir: For creating and removing an element, I've learned that it's better to use Native DOM API. It's due to the weird DOM leakage issue in Angular I filed a while ago: https://github.com/angular/angular/issues/26954. So we started using Native `.removeChild` in FocusTrapDirective. In your case, it would be just:

```
this.el.nativeElement.removeChild(this.ariaLiveElement)
```

Also, just in case, we need to wipe out the property value we used for the element: `delete this.ariaLiveElement`;

hippee-lee: If I read the issue correctly it is only an issue when testing. The memory leak will not occur in production or dev serve environment. Right?  

Anyway, I'll go back to removing it with the native DOM method like I had to begin with and add the delete.

src/clr-angular/forms/common/label.spec.ts
hippee-lee: no, will remove.

src/clr-angular/forms/common/label.ts
hippee-lee: ok. Will remove.

src/clr-angular/forms/common/form.ts
Shijir: Yeah, it's only testing related issue. You can still use the attribute/class manipulation Renderer2 APIs, but only creation/deletion of elements need to be implemented with Native DOM APIs.

src/clr-angular/forms/common/form.ts
hippee-lee: The array is initialized to empty so I think that should be ok. 

You are right its not working. I'll have to come up with a different way. The simple example we started with doesn't really apply any more because the @ContentChildren has all of the labels not just the ones with for attributes and the invalidIds array has objects so checking `indexFor` won't work here. 

I'll push an update soon.

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,71 @@ export class ClrForm {
...
+    }
+
+    this.ariaLiveText = this.commonStrings.keys.formErrorSummary;
+    this.resetAriaLive();
+
+    this.ariaLiveElement.textContent =
+      this.commonStrings.keys.formErrorSummary +
+      this.labels
+        .filter(label => label.forAttr !== undefined)
+        .filter(label => this.invalidControls.find(control => label.forAttr === control.id))
```
Shijir: We don't really need to use filtering twice for the same `label.forAttr`. As far as I see, the first filtering is unnecessary. If it's `undefined`, this check `label.forAttr === control.id` would weed out it just as well.

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,71 @@ export class ClrForm {
...
+  private resetAriaLive() {
+    this.ariaLiveElement.innerHTML = '';
+  }
+
+  private updateAriaLive(): void {
+    if (this.invalidControls.length === 0) {
+      return;
+    }
+
+    this.ariaLiveText = this.commonStrings.keys.formErrorSummary;
```
Shijir: Is this `this.ariaLiveText` being used anywhere?

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,71 @@ export class ClrForm {
...
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-atomic', 'true');
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-relevent', 'text');
+      this.renderer.addClass(this.ariaLiveElement, 'clr-sr-only');
+      this.renderer.insertBefore(this.el.nativeElement, this.ariaLiveElement, this.el.nativeElement.firstChild);
+    }
+  }
+
+  private removeAriaLiveElement() {
+    if (isPlatformBrowser(this.platformId)) {
+      this.el.nativeElement.removeChild(this.ariaLiveElement);
```
Shijir: oh we need to check if `this.ariaLiveElement` is defined besides checking platform. Also, `resetAriaLive` method needs the same check before accessing the `.innerHTML` property.

src/clr-angular/forms/common/form.ts
hippee-lee: nope its left over from the other iterations. Will remove.

src/clr-angular/forms/common/form.ts
hippee-lee: I'll add it in the HostListener method.

src/clr-angular/forms/common/form.ts
hippee-lee: will remove.

src/clr-angular/forms/common/form.ts
Shijir: Wait why in HostListener? This method is being called in `ngOnDestroy`

hippee-lee: resetAriaLive is different from removeAriaLive.

It needs reset to only update the innerText so it works with screen readers. 
It needs removeAriaLive to do the cleanup wrt to removing the child from the node and deleting the ariaLiveElement. removeAriaLiveElement is only called from ngOnDestroy.

Shijir: Oh I thought you are adding it only in the HostListener. But I see you've also added that check here in this method. So looks good 👌

src/clr-angular/forms/common/form.ts
```diff
@@ -17,8 +30,19 @@ import { MarkControlService } from './providers/mark-control.service';
     '[class.clr-form-compact]': 'layoutService.isCompact()',
   },
 })
-export class ClrForm {
-  constructor(public layoutService: LayoutService, private markControlService: MarkControlService) {}
+export class ClrForm implements OnInit {
+  private ariaLiveElement: HTMLDivElement;
+  private invalidControls = [];
+  ariaLiveText = '';
```
Shijir: need to remove this one as we don't use it anymore

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,69 @@ export class ClrForm {
...
+  }
+
+  private updateAriaLive(): void {
+    if (this.invalidControls.length === 0) {
+      return;
+    }
+
+    this.resetAriaLive();
+
+    this.ariaLiveElement.textContent =
```
Shijir: Nit: we are using `.textContent` here to update the text content in this method, but in `resetAriaLive()` we are using `.innerHTML`. Can we stick to either `.innerHTML` or `.textContent` for consistency?

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,69 @@ export class ClrForm {
...
+  private resetAriaLive() {
+    this.ariaLiveElement.innerHTML = '';
+  }
+
+  private updateAriaLive(): void {
+    if (this.invalidControls.length === 0) {
+      return;
+    }
+
+    this.resetAriaLive();
```
Shijir: We don't need to call reset here because, on the next line, we would completely override the content.

src/clr-angular/forms/common/form.ts
```diff
@@ -17,8 +30,19 @@ import { MarkControlService } from './providers/mark-control.service';
     '[class.clr-form-compact]': 'layoutService.isCompact()',
   },
 })
-export class ClrForm {
-  constructor(public layoutService: LayoutService, private markControlService: MarkControlService) {}
+export class ClrForm implements OnInit {
```
Shijir: nit: we are using `OnDestroy` as well

src/clr-angular/forms/common/form.ts
hippee-lee: Yes, I'll stick with `textContent`

src/clr-angular/forms/common/form.ts
hippee-lee: ok. I will remove it.

src/clr-angular/forms/common/form.ts
hippee-lee: 👍🏽 - I'll add it to the ClrForm class implements.

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,69 @@ export class ClrForm {
...
+        .map(label => label.labelText)
+        .join(' ');
+  }
+
+  private createAriaLiveElement() {
+    if (isPlatformBrowser(this.platformId)) {
+      // ClrForms need to have aria live element for screen readers.
+      this.ariaLiveElement = document.createElement('div');
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-live', 'polite');
+      this.renderer.setAttribute(this.ariaLiveElement, 'aria-live', 'polite');
```
Shijir: This line is duplicated

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,60 @@ export class ClrForm {
...
+  ngOnInit() {
+    this.createAriaLiveElement();
+  }
+
+  private updateAriaLive(): void {
+    if (this.invalidControls.length === 0) {
+      return;
+    }
+
+    this.ariaLiveElement.textContent =
```
gnomeontherun: I find this awkward. This could be a loooong list. Was this the requirement or optional?

I think it would be more practical to state something like `There are # of invalid fields, you're now on the first error` since we're moving focus.

src/clr-angular/forms/common/label.ts
```diff
@@ -27,6 +27,10 @@ export class ClrLabel implements OnInit, OnDestroy {
   private subscriptions: Subscription[] = [];
   private enableGrid = true;
 
+  get labelText(): string {
+    return this.el.nativeElement.textContent;
```
gnomeontherun: Need a platform check here.

src/dev/src/app/forms/a11y/a11y.ts
```diff
@@ -0,0 +1,37 @@
...
+    container: '',
+    required: '',
+  };
+
+  submitReactive() {
+    console.log('FormsA11yReactive');
+  }
+
+  submitTemplate() {
+    console.log('FormsA11yReactive');
```
gnomeontherun: Typo

src/website/src/app/documentation/demos/checkboxes/checkboxes.demo.html
```diff
@@ -156,31 +173,37 @@ <h4>Indeterminate checkboxes</h4>
...
+            the label like you see below (the CSS selectors are specific about this ordering). You'll want to tie the
+            input and label together by providing an id for the label to connect. Ensure you add the
+            <code class="clr-code">clr-checkbox</code> class to the checkbox button as well.</p>
 
         <p></p>
 
-        <div class="clr-checkbox-wrapper">
-            <input type="checkbox" id="checkbox0" name="checkbox-basic" value="option1" class="clr-checkbox">
-            <label for="checkbox0">My choice</label>
+<!--        <div class="clr-checkbox-wrapper">-->
```
gnomeontherun: ?

src/clr-angular/forms/common/form.ts
hippee-lee: This was a request during discussion with @clane. Chris, are you ok with the propsed text change from above?

src/clr-angular/forms/common/label.ts
hippee-lee: It was there, @shijir requested that it be removed in his feedback here: https://github.com/vmware/clarity/pull/3830#discussion_r327350898

It's easy to put back but would like agreement from everyone before I change it again. @shijir, any objections?

Shijir: We don't really need to inject another provider that wouldn't offer any value in this case. If we were accessing method beyond `this.el.nativeElement.textContent` like `this.el.nativeElement.textContent.doSomething()`, then sure let's check the platform. But we are only interested in `.textContent` property. If it's `undefined`, then let it be. It just won't be used. If it holds a text value, then we will use that. But what we could improve is this:
```
get labelText(): string {
    return this.el.nativeElement && this.el.nativeElement.textContent;
```

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,60 @@ export class ClrForm {
...
+      this.commonStrings.keys.formErrorSummary +
+      this.labels
+        .filter(label => this.invalidControls.find(control => label.forAttr === control.id))
+        .map(label => label.labelText)
+        .join(' ');
+  }
+
+  private createAriaLiveElement() {
```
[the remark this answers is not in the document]
hippee-lee: I'll put a comment here for now.

src/dev/src/app/forms/a11y/a11y.ts
hippee-lee: Will update to Template.

src/website/src/app/documentation/demos/checkboxes/checkboxes.demo.html
hippee-lee: Will remove.

src/clr-angular/forms/common/label.ts
gnomeontherun: I'm ok with what Shijir proposes, just don't want to have SSR problems.

src/clr-angular/forms/common/form.spec.ts
```diff
@@ -66,5 +93,31 @@ export default function(): void {
...
+      const ariaLiveText = `${commonStrings.keys.formErrorSummary}${
+        fixture.componentInstance.testLabel.nativeElement.innerText
+      }`;
+      expect(ariaLiveElement.innerText).toBe(ariaLiveText);
+    });
+
+    it('responds when the form is submitted', function() {
+      fixture.detectChanges();
+      spyOn(directive.componentInstance.form, 'onFormSubmit');
+      // spyOn(fixture.componentInstance.submitBtn.nativeElement, 'focus');
```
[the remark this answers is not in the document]
hippee-lee: Will remove

src/clr-angular/forms/common/form.ts
chlane: I think  "There are # of fields in error" is good enough. The focus movement will cause the label for the first field to read by the screen reader.  The live region should use aria-live="polite" so that that all announcements are heard and the content currently being read from the screen reader's cursor position is also still announced.

src/clr-angular/forms/common/form.ts
```diff
@@ -28,4 +52,60 @@ export class ClrForm {
   markAsTouched() {
     this.markControlService.markAsTouched();
   }
+
+  @ContentChildren(ClrLabel, { descendants: true })
+  labels: QueryList<ClrLabel>;
+
+  @HostListener('submit')
+  onFormSubmit() {
```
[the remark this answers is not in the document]
hippee-lee: I updated the hostlistener to call markAsTouched() and added optional args and a guard in t markAsTouched that allow consumers to prevent the aria live from being updated.

src/clr-angular/forms/common/label.ts
hippee-lee: Done.

chlane:
Partially fixed. Tested on https://stackblitz.com/edit/github-ofgjza.

According to https://confluence.eng.vmware.com/display/A11Y/Screen+Reader+Error+Messages, we don't want the error message itself to be a live region. So please remove the aria-live property from the element below.

<clr-control-error _ngcontent-esu-c1="" class="clr-subtext" aria-live="polite" id="clr-form-control-3-error">Error message that appears after focus is lost and control is invalid</clr-control-error>

PR #3837, opened 2019-09-20
build: improve base build, linting and TSDocs

coryrylan:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [x] CI related changes

## What is the new behavior?
 This PR updates the following
- Updated TSLint and TSconfig to latest Angular CLI updates and additional auto fixes (clr-base only)
- Updated TS config to use recommended strict compiler checks (clr-base only)
- Updated `pacakge.json` to be organized by project type 
- Added basic TSDoc site prototype in base dev app
- Added `clr-ui` project to help start transition css out of `clr-angular` when ready.
- Added golden file check for `clr-base` API

* [ ] Yes
* [x] No


bdryanovski:
I think that we could skip the `:start` suffix.

```
npm run website
npm run base
...
``` 
The same as `npm run start`

.prettierignore
```diff
@@ -8,4 +8,5 @@ golden/
 src/website/src/releases/final-template/auto-generated-routes.ts
 src/website/src/sitemap.xml
 src/clr-base/utils/css-vars/css-vars-ponyfill-fork.ts
-src/clr-base/**/*.css.ts
\ No newline at end of file
+src/clr-base/**/*.css.ts
```
gnomeontherun: I'm worried about the number of files that we're generating in our src directory, can't we be putting them into `dist` or somewhere else?

src/clr-ui/variables/_variables.scss
```diff
@@ -0,0 +1,3 @@
+// TODO: these files need to be moved out of clr-angular and into a clr-ui directory
+@import './../../clr-angular/color/variables.color';
```
mathisscott: I like that this migration has started. Ultimately, wouldn't we want the actual files to live in `clr-ui` and for `clr-angular` and others to import them as is being done here?

What's our strategy there?

coryrylan: Yeah thats the plan, this was just so the project was set up and building for the clr-dev app. I didn't want to move anything until the CSS custom property branch was merged.

.prettierignore
coryrylan: TypeScript won't allow the imports to be outside the root source directory, I could put them in a dist inside the base src `/src/clr-base/dist/styles/` then in the component you would import

`import { styles } from './../dist/styles/button.element.css';`

instead of 

`import { styles } from './button.element.css';`

The project will only generate one file per component. Everything else will live in clr-ui.

coryrylan: I'll work on this some more see if I can get some path mapping to help with this.

Issue #3603, opened 2019-07-08
Tree View lacks typeahead functionality: Best Practice


gnomeontherun:
## Describe the bug

Tree Views lack Type-ahead functionality. Type-ahead is recommended for all trees, especially for trees with more than 7 root nodes. 

## How to reproduce

https://stackblitz.com/github/clane/clarityTest2/tree/master/clarity-test-app

- Using the keyboard alone try to navigate any of the tree examples by typing the first character seen in any treenode.
- Observe that this is not possible

## Expected behavior

Implement Type-ahead functionality as described below.

- Type a character: focus moves to the next node with a name that starts with the typed character.
- Type multiple characters in rapid succession: focus moves to the next node with a name that starts with the string of characters typed.

See
https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

_Add any other notes about the problem here._

hippee-lee:
Is this search only for nodes in the DOM? Does it matter if they are hidden or visible?

shravansofts:
This is an important feature.
Why this is marked as LowPriority? :(

gnomeontherun:
It is low because its a recommendation but not a requirement, and it is also quite difficult to implement because of the nature of lazy loaded tree views not knowing the whole tree. We don't have a model of the data, but only know about the structure once it is rendered, so traversing to search for values is also a performance drain that we have to consider.

chlane:
The WCAG verbiage says "Type-ahead is recommended for all trees, especially for trees with more than 7 root nodes:".  It looks like the use of the word "recommended" lowers its priority

gnomeontherun:
Hi there, this is an automated message.

In an effort to dedupe and track accessibility reports consistently, this issue is being closed and tracked internally with our Accessibility team. It will still be triaged and addressed as part of ongoing work.

PR #3851, opened 2019-09-25
feat: a11y and focus management for tree-view

### LINKED ISSUE #3569: Keyboard interaction for Tree View
## Describe the bug

The Tree View does not support the correct keyboard interactions. Currently , tree nodes can be focused with the Tab key and activated with the Enter key. But the main behaviors expected for trees are as follows.

1) The entire tree structure should be focusable as a whole when using the Tab key
2) When the tree has focus, the user should be able to use the arrow keys to navigate the nodes in the tree in any direction depending on the complexity of the tree.
3) See the rest of the expected keyboard interactions at https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView below.

## How to reproduce

- Go to the Tree View - Basic and try to use the arrow keys to navigate the tree nodes and find that you cannot.

## Expected behavior

See the Accessible Tree example at http://whatsock.com/bootstrap/

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

### LINKED ISSUE #3571: Tree View selected state not announced
## Describe the bug

The selected state of treenode is not announced by screen readers. Currently, when a treenode is selected, a background color of #d9e4ea is applied. But blind screen users cannot see this background color and do not hear any feedback telling the a given treenode has been selected.

## How to reproduce

- Using a screen reader open the Tree View - Icons demo
- "Highlight" one of the tree nodes by using the Tab key to focus it and then pressing either the Enter or the Spacebar keys
- Notice that the screen reader does not announce any information regarding the selected state of the treenode

## Expected behavior

Use aria-selected to convey the selected state of a treenode as described at https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

### LINKED ISSUE #3573: Checkboxes in Tree Views need keyboard interaction design
## Describe the bug

Currently, our trees do not support arrow key navigation. Once they do, actually including checkboxes in tree items will force us to create an overly complex keyboard interaction to operate them. Trees should gain focus as a whole and then arrow keys should be used to navigate through the tree items. If there are checkboxes inside of the tree item, we will have to design a keyboard interaction to break out of the tree navigation context and navigate within the tree item itself. There is some guidance about doing this for grid cells that we can possible adapt. See "Editing and Navigating Inside a Cell" at https://www.w3.org/TR/wai-aria-practices-1.1/#gridNav_inside.

Another solution, could be to simply show icons of checkboxes while using a standard Tree View implementation(https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView). aria-selected can be used to logically convey the selected state while the icon of the checked checkbox conveys it visually. aria-describedby can be used to communicate the indeterminate state. But we don't have to do that either, since the indeterminate state is visual only(https://css-tricks.com/indeterminate-checkboxes/) but it could be beneficial to do so. 

## How to reproduce

TBD

## Expected behavior

TBD

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

Shijir:

## PR Checklist

## PR Type

* [x] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

The Tree View does not support the correct keyboard interactions. 

Issue Number: 

closes #3569
closes #3571
closes #3573

## What is the new behavior?

Adding keyboard focus management that complies with: https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-22

* [x] Yes*
* [ ] No

Previously, `clr-tree` was optional. But now we are making it mandatory. 

## Other information

- **Type-ahead** feature is not implemented as that was not in the scope of this PR. We have a separate tracking issue for the feature: https://github.com/vmware/clarity/issues/3603
- After discussing with the team, I've made it so that when the tree receives focus, the first visible node will be focused. This doesn't follow the authoring guideline, which says *the first visible selected node* should get focused. The reason why we didn't follow the authoring guideline on this specific behavior is that we believe that this behavior will cause the abrupt page scrolling if the first selected node if it exists in a deeply nested tree, resulting in poor user experience.

### TreeView a11y examples:
- [w3c](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html)
- [Open Ajax](http://www.oaa-accessibility.org/examplep/treeview2/)
- [whatsock example](http://whatsock.com/tsg/Coding%20Arena/ARIA%20Menus/Variation%20ARIA%20Tree%20With%20Right%20Click/demo.htm)


Shijir:
Preview: https://deploy-preview-3851--vmware-clarity.netlify.com/documentation/tree-view

src/clr-angular/data/tree-view/tree-focus-manager.service.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  get focusRequest(): Observable<string> {
+    return this._focusRequest.asObservable();
+  }
+
+  get focusChange(): Observable<string> {
+    return this._focusChange.asObservable();
+  }
+
+  private findParentChildren(model: TreeNodeModel<T>): TreeNodeModel<T>[] {
```
jeeyun: Is this essentially finding a node's siblings? Maybe `findSiblings` is a better name?

Shijir: I was debating with myself on this too. Yeah, `findSiblings` is nice, but I ended up choosing `findParentChildren` because it's not only requesting its siblings but also itself.

src/clr-angular/data/tree-view/tree-focus-manager.service.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (model.expanded && model.children.length > 0) {
+      const children = model.children;
+      const lastChild = children[children.length - 1];
+      return this.findLastVisibleInNode(lastChild);
+    } else {
+      return model;
+    }
+  }
+
+  private findNextInParentChildren(model: TreeNodeModel<T>): TreeNodeModel<T> {
```
jeeyun: Similar comment on naming here.. something like `findNextFocusableNode` is probably clearer. The logic looks good. 👍

src/clr-angular/data/tree-view/tree-node.ts
```diff
@@ -40,26 +48,36 @@ import { TREE_FEATURES_PROVIDER, TreeFeaturesService } from './tree-features.ser
...
+      ]),
+      transition('* => void', [
+        style({ display: 'block', 'overflow-y': 'hidden' }),
+        animate(200, style({ height: 0 })),
+      ]),
     ]),
   ],
-  host: { '[class.clr-tree-node]': 'true' },
+  host: {
+    '[attr.role]': '"treeitem"',
```
jeeyun: If we have the role of `treeitem` on the `contentContainer`, I don't think the host should have it also..? Ultimately, the element that receives focus should have both the role and all the `aria-` attributes assigned to it.

src/clr-angular/data/tree-view/tree.ts
```diff
@@ -14,15 +26,65 @@ import { TREE_FEATURES_PROVIDER, TreeFeaturesService } from './tree-features.ser
...
 
   @Input('clrLazy')
   set lazy(value: boolean) {
     this.featuresService.eager = !value;
   }
+
+  tabindex = 0;
+
+  get isMultiSelectable() {
+    return this.featuresService.selectable && this.rootNodes.length > 0;
```
jeeyun: `this.featuresService.selectable` was blowing up for me in certain cases (maybe lazy loaded recursive?). Could you make sure that this works in all cases?

src/clr-angular/layout/main-container/_layout.clarity.scss
```diff
@@ -7,6 +7,7 @@
     display: flex;
     flex-direction: column;
     height: 100vh;
+    min-height: 100vh;
```
jeeyun: Was this broken due to applitools changes? Will this go away after rebase on master?

Shijir: ah yeah 😂 I forgot to remove this one. Thanks for catching this.

src/clr-angular/data/tree-view/tree-node.ts
Shijir: Good question. This was a fix to the Safari + VO. 
I had this previously without `role="treeitem"` on the `clr-tree-node` host element as follows:
```
<clr-tree-node>
  <div class="clr-tree-node-content-container" role="treeitem">...</div>
  <div class="clr-treenode-children" role="group">...</div>
<clr-tree-node>
```
However, the Safari +VO combination wasn't treating the `group` and `treeitem` as parts of the same `treeitem`. Then, after reading some articles, I tried to solve with `aria-owns` attribute like this:
```
<clr-tree-node>
  <div class="clr-tree-node-content-container" role="treeitem" aria-owns="clr-uniq-id-00">...</div>
  <div class="clr-treenode-children" role="group" id="clr-uniq-id-00">...</div>
<clr-tree-node>
```
Still the same result. The Safari + VO didn't tree them as one peace. Finally, I tried this:

```
<clr-tree-node role="treeitem">
  <div class="clr-tree-node-content-container" role="treeitem">...</div>
  <div class="clr-treenode-children" role="group">...</div>
<clr-tree-node>
```

Now, The Safari + VO started working the same as it worked in the spec guideline examples.

src/clr-angular/data/tree-view/tree.ts
Shijir: Sure. Will check them.

gnomeontherun:
I've been reviewing, and right now I propose this go only into v3. It has too many changing behaviors and is a breaking change. If we get into a pickle we can consider a backport then.

src/clr-angular/data/tree-view/tree-node.html
```diff
@@ -4,36 +4,39 @@
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
 
-<div class="clr-tree-node-content-container">
+<div #contentContainer role="treeitem" 
+  class="clr-tree-node-content-container"
+  [attr.aria-expanded]="isExpandable() ? expanded : null"
+  [attr.aria-selected]="ariaSelected"
+  [attr.tabindex]="contentContainerTabindex" 
+  (keydown)="onKeyDown($event)" (focus)="broadcastFocusOnContainer()">
```
gnomeontherun: nit: put `focus` on new line

src/clr-angular/data/tree-view/tree-node.html
```diff
@@ -4,36 +4,39 @@
...
   </div>
-  <div class="clr-treenode-content" [id]="nodeId">
+  <div class="clr-treenode-content">
     <ng-content></ng-content>
   </div>
 </div>
-<div class="clr-treenode-children"
-     [@childNodesState]="expandService.expanded ? 'expanded' : 'collapsed'"
-     [attr.role]="isExpandable() ? 'group' : null">
+<div class="clr-treenode-children" [@toggleChildrenAnim]="expandService.expanded ? 'expanded' : 'collapsed'" [attr.role]="isExpandable() ? 'group' : null">
```
gnomeontherun: nit: put attributes on separate lines like other elements.

src/clr-angular/data/tree-view/tree.ts
```diff
@@ -14,15 +26,65 @@ import { TREE_FEATURES_PROVIDER, TreeFeaturesService } from './tree-features.ser
...
+    '[attr.tabindex]': 'tabindex',
+    '[attr.role]': '"tree"',
+    '[attr.aria-multiselectable]': 'isMultiSelectable',
+  },
 })
-export class ClrTree<T> {
-  // This component can also be used just to declare providers once for trees with multiple root nodes.
+export class ClrTree<T> implements AfterContentInit, OnDestroy {
+  constructor(
+    public featuresService: TreeFeaturesService<T>,
```
gnomeontherun: We talked about this, since its only going into master this breaking change of requiring `clr-tree` is acceptable.

chlane:
We also decided not to include other behaviors from the spec like the type-ahead functionality for trees with more than 7 nodes

gnomeontherun:
nit: use `feat(tree-view)` to scope the commit

src/clr-angular/data/tree-view/tree.ts
Shijir: I've tested it in all browsers, and I haven't experienced any issues here.

src/clr-angular/data/tree-view/tree.ts
```diff
@@ -14,15 +26,65 @@ import { TREE_FEATURES_PROVIDER, TreeFeaturesService } from './tree-features.ser
...
+    public focusManager: TreeFocusManagerService<T>,
```
hippee-lee: nit: `focusManagerService`? to keep the name aligned with the featureService.

PR #3876, opened 2019-10-04
fix: popover adapter

Jinnie:
simple wrapper for smart popovers

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Smart popovers can now be used with the old-style string positions. This wrapper makes sure all a11y is set correctly OOTB. Could be used for quick transition of Signposts Popovers Dropdowns, etc. to the new popovers, with minor or not any API changes.

* [ ] Yes
* [x] No


src/clr-angular/utils/popover/adapter/operators.ts
```diff
@@ -0,0 +1,123 @@
...
+export function transposeStringPosition(position: string): ClrPopoverPosition {
+  return SIMPLE_TO_SMART_POSITIONS[position];
+}
+
+export function validPosition(position: string): boolean {
+  return position && !!SIMPLE_TO_SMART_POSITIONS[position];
+}
+
+export function positionsEqual(pos1: ClrPopoverPosition, pos2: ClrPopoverPosition): boolean {
+  if (pos1.anchor !== pos2.anchor) {
```
Jinnie: As Bozhidar suggested, this could go in one if. Looking at it here again, I agree.

Jinnie:
This is a quick early PR to unblock Bozhidar and Matt. 
This component and tools are not yet used anywhere in the project.
When we start using them, the next should be finished too:
[ ] Better inline documentation
[ ] Tests

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: It could just keep in mind my unit-test :)

src/clr-angular/utils/popover/adapter/operators.ts
```diff
@@ -0,0 +1,124 @@
...
+    anchor: ClrAlignment.CENTER,
+    content: ClrAlignment.END,
+  },
+};
+
+export function posiblePositionsAsStrings() {
+  return Object.keys(SIMPLE_TO_SMART_POSITIONS);
+}
+
+export function defaultPosition() {
```
hippee-lee: I don't think this name will work. Default position for dropdowns is `bottom-left`. 
It seems like the component should be responsible for setting the default string. Not here.

src/clr-angular/utils/popover/adapter/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export * from './popover-adapter';
```
hippee-lee: I don't think this should be part of the public API. It is an internal only component. 

Having it exported from the index means it will get picked up when the golden file is built.

src/clr-angular/utils/popover/adapter/operators.ts
```diff
@@ -0,0 +1,124 @@
...
+  return SIMPLE_TO_SMART_POSITIONS['bottom-right'];
+}
+export function transposePositionString(position: ClrPopoverPosition): string {
```
hippee-lee: I'm not sure why this is a separate file instead of part of the position-operators.ts. These functions all operate on ClrPositions (or the string representation of a position).

src/clr-angular/utils/popover/providers/popover-position.service.ts
```diff
@@ -5,32 +5,42 @@
...
   }
 
+  get smartPosition(): ClrPopoverPosition {
+    return this._smartPosition;
+  }
+
   constructor(private eventService: ClrPopoverEventsService, @Inject(PLATFORM_ID) public platformId: Object) {}
 
   public alignContent(content: HTMLElement): ClrPopoverContentOffset {
+    // debugger;
```
hippee-lee: Comment?

src/clr-angular/utils/popover/adapter/popover-adapter.module.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ClrIconModule } from '../../../icon/icon.module';
+import { ClrCommonPopoverModule } from '../../../popover/common/popover.module';
+import { ClrFocusOnViewInitModule } from '../../focus/focus-on-view-init/focus-on-view-init.module';
+import { ClrPopoverModuleNext } from '../popover.module';
+
+import { ClrPopoverAdapter } from './popover-adapter';
+
+export const CLR_SIMPLE_POPOVER_DIRECTIVES: Type<any>[] = [ClrPopoverAdapter];
+
+@NgModule({
```
hippee-lee: Do we really need a module for this? Since its an internal component shouldn't the consuming module import the PopoverAdapter?

src/clr-angular/utils/popover/adapter/popover-adapter.module.ts
```diff
@@ -0,0 +1,24 @@
...
+  imports: [CommonModule, ClrCommonPopoverModule, ClrIconModule, ClrFocusOnViewInitModule, ClrPopoverModuleNext],
+  declarations: [CLR_SIMPLE_POPOVER_DIRECTIVES],
+  exports: [CLR_SIMPLE_POPOVER_DIRECTIVES],
+})
+export class ClrSimplePopoverModule {}
```
hippee-lee: nit: if we do keep this I prefer to have it named like the component ClrPopoverAdapterModule

src/clr-angular/utils/popover/providers/popover-position.service.ts
```diff
@@ -5,32 +5,42 @@
...
 import { ClrPopoverContentOffset } from '../interfaces/popover-content-offset.interface';
 import { ClrViewportViolation } from '../enums/viewport-violation.enum';
 import { align, flipSidesAndNudgeContent, flipSides, nudgeContent, testVisibility } from '../position-operators';
 import { ClrAxis } from '../enums/axis.enum';
+import { Observable } from 'rxjs';
 
 @Injectable()
 export class ClrPopoverPositionService {
+  private _smartPositionReady: EventEmitter<ClrPopoverPosition> = new EventEmitter();
+  smartPositionReady: Observable<ClrPopoverPosition> = this._smartPositionReady.asObservable();
```
hippee-lee: Can we call this just smartPosition and drop the ready?

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: Ok but let us return `null` when we could not find the position from 
 * `transposeStringPosition`
 * `transposePositionString`

And everyone will have to handle the default state by themself.

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: they could go there this is not a problem, 
But I could argue that the naming of the method is quite clear for example:

```js
transposePositionString()
transposeStringPosition()
```
They look too similar and sound similar but are getting different arguments.

Could go with something like that:
```js
transposePositionObject()
transposePositionString()
```
Or even use the word `find` cause that is what we do, going over a list of elements to find a match.

```js
findPositionObject('bottom-left');
findPositionString(ClrPopoverPosition);
```

src/clr-angular/utils/popover/adapter/operators.ts
Jinnie: `requestedPosition` ?

Bozhidar, you have point... now one returns null, the other undefined. Returning null should be the preferred one.

src/clr-angular/utils/popover/adapter/popover-adapter.module.ts
Jinnie: I've missed this during refactoring. Thanks.

src/clr-angular/utils/popover/adapter/popover-adapter.module.ts
Jinnie: Yes, probably we could import the adapter component directly. Let's try it this way. The less boilerplate code, the better.

src/clr-angular/utils/popover/providers/popover-position.service.ts
Jinnie: Okay, we talked with Bozhidar, that we drop the current smartPosition property and only leave the Observable, but it will have to become BehaviorSubject then.

src/clr-angular/utils/popover/providers/popover-position.service.ts
Jinnie: Oops!
Please don't uncomment, only delete :)

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: That's ok as well :) just didn't like the `PositionString`, `StringPosition` I got the idea just was a bit cryptic of don't know the code.

src/clr-angular/utils/popover/providers/popover-position.service.ts
bdryanovski: I went with delete :)

src/clr-angular/utils/popover/position-operators.spec.ts
```diff
@@ -16,10 +16,57 @@ import {
...
+      posiblePositionsAsStrings().forEach(position => {
+        expect(position).toBe(transposePositionToString(SIMPLE_TO_SMART_POSITIONS[position]));
+      });
+
+      // Not a valid position
+      expect(null).toBe(transposePositionToString({ axis: -5, side: -5, anchor: -5, content: -5 }));
+    });
+
+    it('should check that the position is valid', () => {
+      expect(validPosition('bottom-right')).toBe(true);
```
Jinnie: Maybe we should try all possible positions :(

src/clr-angular/utils/popover/position-operators.spec.ts
```diff
@@ -16,10 +16,57 @@ import {
...
+      expect(validPosition('NOT_VALID_NAME')).toBe(false);
+    });
+
+    it("should compare two position and return false if they don't match", () => {
```
Jinnie: It would be nice to try with randomly ordered properties. The current implementation should be okay, but some future optimized comparison function might fail.

src/clr-angular/utils/popover/providers/popover-position.service.ts
```diff
@@ -5,42 +5,40 @@
...
 import { align, flipSidesAndNudgeContent, flipSides, nudgeContent, testVisibility } from '../position-operators';
 import { ClrAxis } from '../enums/axis.enum';
-import { Observable } from 'rxjs';
+import { Observable, BehaviorSubject } from 'rxjs';
 
 @Injectable()
 export class ClrPopoverPositionService {
-  private _smartPositionReady: EventEmitter<ClrPopoverPosition> = new EventEmitter();
-  smartPositionReady: Observable<ClrPopoverPosition> = this._smartPositionReady.asObservable();
+  private _smartPositionEmitter: BehaviorSubject<ClrPopoverPosition> = new BehaviorSubject(null);
```
Jinnie: nit: "Emitter" is a bit misleading now.

src/clr-angular/utils/popover/position-operators.spec.ts
bdryanovski: No easy way to check them all without write expectations as a list of keys that we expect to have.

```js
const positionList = Object.keys(SIMPLE_TO_SMART_POSITIONS)
```
`validPosition` just test the key is there. Even could argue that the name is misleading, we don't test is this position a really valid position we just test that the key is there.

```js
Object.keys(SIMPLE_TO_SMART_POSITIONS).forEach((position) => {
  expect(validPosition(position)).toBe(true)
});
```

The problem for me is that this test will always pass :) 

```js
const positionList = ['bottom-left', 'bottom-right' ... ];
```

If we track all keys by hand we always will have to update the list and the test will never fail if we add a new one. It will fail only if we remove one.

```js
expect(positionList.length).toBe(Object.keys(SIMPLE_TO_SMART_POSITIONS).length);
```

This could solve it in a way .. but still, a lot of manual support will be needed to keep this working ...

src/clr-angular/utils/popover/providers/popover-position.service.ts
bdryanovski: That is right! ;) Will fix it

src/clr-angular/utils/popover/adapter/popover-adapter.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.subs.push(
+        this.toggleService.openChange.subscribe(change => {
+          this.ifOpen.open = change;
+        })
+      );
+    }
+    this.subs.push(
+      this.positionService.smartPosition.subscribe(finalPosition => {
+        const resolvedPosition = transposePositionToString(finalPosition);
+        this._smartPositionReady.next(resolvedPosition);
```
bdryanovski: this could be merged into one line ...

src/clr-angular/utils/popover/adapter/operators.ts
hippee-lee: i like `find`. Its a little clearer whats going on vs `transpose`.

src/clr-angular/utils/popover/position-operators.spec.ts
hippee-lee: I'm in agreement with Ivan here even though its a lot of manual support. If we are going to provide the adapter to support the existing/legacy positions, imo we should test that 

1) nothing new is added and 
2) everything currently supported still works

This way in a year when someone removes one  or adds a new one (unlikely but you never know) we are aware that something has changed.

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: Are we all agreeing on `findPositionObject` and `findPositionString` ?

src/clr-angular/utils/popover/position-operators.spec.ts
bdryanovski: So we are going with copy the SIMPLE_TO_SMART_POSITIONS into the test and expect to get the same `key => object` as part of the test. 

```js
const TEST_VERSION_OF_SIMPLE_TO_SMART_POSITIONS = { ... }

Object.keys(TEST_VERSION_OF_SIMPLE_TO_SMART_POSITIONS).forEach((positionKey) => {
  expect(SIMPLE_TO_SMART_POSITIONS[positionKey]).toEqual(TEST_VERSION_OF_SIMPLE_TO_SMART_POSITIONS[positionKey]);
})
```
Or could go really crazy 

```js
expect(JSON.stringify(TEST_VERSION_OF_SIMPLE_TO_SMART_POSITIONS)).toBe(JSON.stringify(SIMPLE_TO_SMART_POSITIONS))
```

I just wanted to be clear that I got the idea right!

hippee-lee: Don't copy, import it directly and use it to test with. 

It should have length of ... 
it should have 'top-left' key && top-left should return ClrPosition { ... } 
etc. 

This way we are directly testing the thing used in the operators and it will be more likely that we catch changes if they happen.

src/clr-angular/utils/popover/adapter/operators.ts
bdryanovski: Done now both methods are returning `null` as standard response when no position is found

src/clr-angular/utils/popover/adapter/index.ts
bdryanovski: There was no need for the file after all

.favorites.json
```diff
@@ -0,0 +1,26 @@
+[
```
Jinnie: Don't think this file should be submitted

src/clr-angular/utils/popover/position-operators.spec.ts
```diff
@@ -16,10 +16,140 @@ import {
...
+    side: ClrSide.BEFORE,
+    anchor: ClrAlignment.CENTER,
+    content: ClrAlignment.END,
+  },
+};
+
+export function ClrPositionHelperSpec(): void {
+  describe('ClrPosition helper functions', () => {
+    it('should return array of posible positions', () => {
+      expect(Object.keys(TEST_VERSION_OF_SIMPLE_TO_SMART_POSITIONS).length).toBe(12);
```
Jinnie: Why do we test the helper data that we defined in this same spec file?

.favorites.json
bdryanovski: will revert it .. in my next commit I had it on gitignore but on the other branch

src/clr-angular/utils/popover/position-operators.spec.ts
bdryanovski: You are right here, will update it

Jinnie:
Closing this as we decided that the 3 cases have some requirements specifics that complicate the use of an abstraction too much and we're dropping the popover adapter.

Issue #2897, opened 2018-11-29
[forms] Generic way to override column widths


technbuzz:
## Summary

I wish there was a way to configure on top level of how the column classes are decided for **label** and actual **input element**. 
With the new forms we don't need to link label and input using for and id attributes respectively. I wish we could have some configs to make our life much easier.

## Use case

Already using **clr-input-container** does so much for me. I should not be overriding and adding custom classes to **label** and as well as **input element**

## Examples

config:{ defaultLabelColumns: 3, defaultFormContianerColumns: 9}

gnomeontherun:
Thanks for the request, this was something I had in mind as a potential enhancement but wanted to hear if people requested it.

I don't have a specific API in mind yet, but I was thinking it would be similar to how `<form clrForm clrLayout="horizontal" clrLayoutColumns="3,9">` then it would set for the whole form. We don't have a global service to track this setting at the application level, so that would be a larger enhancement and I think this would be a good first step. Thoughts?

dorthrithil:
I was just about to request the same, as horizontal forms in small modals don't look good (line breaks in labels because of not enough space, but 50% empty space on the right side). This feature will help to get around this problem.

mojoLyon:
Hi.
Anymore information on this enhancement ?

dorthrithil:
It's currently not assigned to any milestone and not mentioned in the [work in progress](https://github.com/vmware/clarity/wiki/Work-in-progress) file. So this will probably take some more time.

Here's a code snipped I use as a workaround. This would change the column widths in forms from `2 / 10` to `3 / 9` if you add the class `override-col-widths-3-9` to your form tag.

```css
form.override-col-widths-3-9 {

  .clr-row {

    .clr-col-md-2 {

      @media (min-width: 768px) {
        flex: 0 0 25%;
        max-width: 25%;
      }

    }

    .clr-col-md-10 {

      @media (min-width: 768px) {
        flex: 0 0 75%;
        max-width: 75%;
      }

    }

  }

}
```

dorthrithil:
@gnomeontherun I think `clrLayoutColumns="3,9"` is fine.  At least for the application I'm building I don't see the need for a global service. We have 50% forms in modals/wizards which need the overridden column widths and 50% "full width" forms which fare well with the default setting. So even with a global service I would have to override the column widths for the same number of forms as without. 
[...]

PR #3883, opened 2019-10-07
feat(form): generic way to override column widths

andreiconstantinbytex:
#2897 

Signed-off-by: Andrei Constantin <andrei.constantin@bytex.ro>

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

The horizontal and compact form layouts only allow a default label col size of 2.

Issue Number: #2897 

## What is the new behavior?
The horizontal and compact form layouts allows setting any label col size within range 1 through 12.

* [ ] Yes
* [x] No


src/clr-angular/forms/common/label.ts
```diff
@@ -45,7 +45,7 @@ export class ClrLabel implements OnInit, OnDestroy {
       this.el.nativeElement.className.indexOf('clr-col') < 0
     ) {
       this.renderer.addClass(this.el.nativeElement, 'clr-col-12');
-      this.renderer.addClass(this.el.nativeElement, 'clr-col-md-2');
+      this.renderer.addClass(this.el.nativeElement, `clr-col-md-${this.layoutService.labelSize}`);
```
gnomeontherun: We need a test to verify this configured value gets projected.

src/clr-angular/forms/common/layout.ts
```diff
@@ -12,6 +12,7 @@ import { Layouts, LayoutService } from './providers/layout.service';
 })
 export class ClrLayout implements OnInit {
   @Input('clrLayout') layout: Layouts;
+  @Input('clrLabelSize') labelSize: number;
```
gnomeontherun: We can just set a default value here and then the logic below can always assume a value. Doing the following is probably better so we can better handle dynamic inputs.

```
private _labelSize: number = 2;
@Input('clrLabelSize') 
set labelSize(size: number) {
  if (this.layoutService.labelSizeIsValid(size) {
    this._labelSize = size;
  }
}
get labelSize(): number {
  return this._labelSize;
}
```

src/clr-angular/forms/common/layout.ts
```diff
@@ -20,5 +21,9 @@ export class ClrLayout implements OnInit {
     if (this.layout && this.layoutService.isValid(this.layout)) {
       this.layoutService.layout = this.layout;
     }
+    // Only set the label width if it is a valid option
+    if (this.labelSize && !this.layoutService.isVertical() && this.layoutService.labelSizeIsValid(this.labelSize)) {
```
gnomeontherun: This can be removed if you use the above getter/setter.

src/clr-angular/forms/common/layout.spec.ts
```diff
@@ -10,39 +10,213 @@ import { By } from '@angular/platform-browser';
...
+      const TestClass = customTestComponentFactory(<Layouts>'invalid');
+      TestBed.configureTestingModule({ declarations: [ClrLayout, TestClass] });
+      const fixture = TestBed.createComponent(TestClass);
       const service = fixture.debugElement.injector.get(LayoutService);
       fixture.detectChanges();
+
       expect(service.layout).toEqual(Layouts.HORIZONTAL);
     });
+
+    it('horizontal layout label should have a default size of 2', () => {
```
gnomeontherun: `it('label should have a default size of 2'), () => {`

Just test that the label is 2 by default.

src/clr-angular/forms/common/providers/control-class.service.spec.ts
```diff
@@ -55,7 +55,7 @@ export default function(): void {
 
     it('should return any additional classes passed by the control', function() {
       service.className = 'test-class';
-      expect(service.controlClass(false, false, 'extra-class')).toBe('test-class extra-class');
+      expect(service.controlClass(false, false, 2, 'extra-class')).toBe('test-class extra-class');
```
gnomeontherun: Please add a test that confirms that if you pass a different labelSize that it renders the right class.

src/clr-angular/forms/common/utils/constants.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+/**
+ * This is the maximum number of column cells of the layout.
+ */
+export const COLUMN_SIZE: number = 12;
```
gnomeontherun: I think we can just store this as a `readonly` property of the layout service to avoid this extra const and file.

src/clr-angular/forms/password/password-container.ts
```diff
@@ -118,7 +118,11 @@ export class ClrPasswordContainer implements DynamicWrapper, OnDestroy {
   }
 
   controlClass() {
-    return this.controlClassService.controlClass(this.invalid, this.addGrid());
+    return this.controlClassService.controlClass(
+      this.invalid,
+      this.addGrid(),
+      this.layoutService ? this.layoutService.labelSize : null
```
gnomeontherun: We should be able to just pass `this.layoutService.labelSize` to all of these calls to `controlClass()` to simplify.

src/website/src/app/documentation/demos/forms/ng/label-size.html
```diff
@@ -0,0 +1,8 @@
+<form clrForm clrLayout="horizontal" clrLabelSize="1">
```
gnomeontherun: It would be nicer to use `clrLabelSize="4"` so the difference is more obvious.

src/website/src/app/documentation/demos/forms/forms.demo.html
```diff
@@ -190,6 +190,12 @@ <h4>Layout options</h4>
 
             <clr-code-snippet [clrCode]="ngLayouts"></clr-code-snippet>
 
+            <h4>Layout with grid</h4>
```
gnomeontherun: Let's put this block right above `Overriding column widths` section.

src/clr-angular/forms/common/layout.ts
andreiconstantinbytex: Should it still assign it to the service only if the layout is non-vertical?

src/clr-angular/forms/common/layout.ts
andreiconstantinbytex: This was my initial approach but I decided to keep it consistent with the layout logic. Thanks!

src/clr-angular/forms/common/utils/constants.ts
andreiconstantinbytex: This would imply injecting the layout service into the control class service in order to calculate the appropiate col size for any given container

src/clr-angular/forms/password/password-container.ts
andreiconstantinbytex: the reason I chose this logic is because the tests are failing when, for instance, a password container is used outside a clrForm

src/clr-angular/forms/common/layout.ts
gnomeontherun: Ok I missed this the first time, but this stuff needs to go onto the ClrForm and not ClrLayout. 

Also let's put the value directly on the service regardless of what layout, because it only gets used for horizontal/vertical anyways so storing it is fine.

src/clr-angular/forms/common/layout.ts
gnomeontherun: This should just be removed. We set it on the service always, we just don't always need it for vertical.

src/clr-angular/forms/password/password-container.ts
gnomeontherun: Ahh yeah good call.

src/clr-angular/forms/common/utils/constants.ts
gnomeontherun: I don't mind injecting the service into another service, we do this a lot and seems easier than standalone `const` values.

src/clr-angular/forms/common/layout.ts
andreiconstantinbytex: removed this one and added the validation on the setter. So in some cases labelSize can be set for vertical layouts?

src/clr-angular/forms/common/layout.ts
andreiconstantinbytex: horizontal/vertical? The form's css grid options only allow horizontal/compact

src/clr-angular/forms/common/label.spec.ts
```diff
@@ -113,6 +113,22 @@ export default function(): void {
       expect(label.classList.contains('clr-col-12')).toBeTrue();
     });
 
+    fit('adds the grid classes for non-vertical non-default layouts', function() {
```
gnomeontherun: `fit`

src/clr-angular/forms/common/form.ts
```diff
@@ -35,8 +36,17 @@ export class ClrForm implements OnInit, OnDestroy {
...
+    this.layoutService.labelSize = size;
+  }
+
+  get labelSize(): number {
+    return this.layoutService.labelSize;
+  }
+
   constructor(
-    public layoutService: LayoutService,
+    private layoutService: LayoutService,
```
gnomeontherun: This needs to be public, it is used in the host bindings.

src/website/src/app/documentation/demos/forms/ng/label-size.html
```diff
@@ -0,0 +1,8 @@
+<form clrForm clrLayout="horizontal" clrLabelSize="4">
+    <span class="clr-sr-only">Tabbing through form inputs will trigger validation messages to be read aloud.</span>
```
gnomeontherun: This can be removed.

src/clr-angular/forms/common/providers/control-class.service.ts
```diff
@@ -1,22 +1,25 @@
...
 
-  controlClass(invalid = false, grid = false, additional = '') {
+  constructor(@Optional() private layoutService: LayoutService) {}
+
+  controlClass(invalid = false, grid = false, labelSize = 2, additional = '') {
     const controlClasses = [this.className, additional];
     if (invalid) {
       controlClasses.push('clr-error');
     }
     if (grid && this.className.indexOf('clr-col') === -1) {
```
gnomeontherun: We need to check if layoutService is defined in this `if` block before we use it.

src/clr-angular/forms/common/form.ts
```diff
@@ -35,6 +36,15 @@ export class ClrForm implements OnInit, OnDestroy {
   private ariaLiveElement: HTMLDivElement;
   private invalidControls = [];
 
+  @Input('clrLabelSize')
+  set labelSize(size: number) {
+    this.layoutService.labelSize = size;
+  }
+
+  get labelSize(): number {
```
gnomeontherun: Do we ever need to get this value? I think the service is always being used.

PR #3895, opened 2019-10-09
feat: datalist component

### LINKED ISSUE #3813: Datalist component
## Summary

Implement an HTML `datalist` element as an Angular control, and make sure it fits well within our forms layouts and styling.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist 

* What is the change? This is a new component but of an existing HTML element that is widely supported
* Why should it go in Clarity? It provides a limited combobox like behavior and might be ideal in many scenarios
* Does this change impact existing behaviors? If so how? No its new
* If this change introduces a new behavior, is this behavior accessible? It should be the same as other form controls

## Use case

We've seen people using dropdowns to create a combobox like functionality, but a datalist is often sufficient. 

## Examples

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

## Workarounds

_Is there a workaround for this feature at the moment, even if it is less than ideal?_

AlexMarcuBytex:
- datalist component added
- dev page added

closes #3813

Signed-off-by: Marcu Alexandru Stefanel <alexandru.marcu@bytex.ro>

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: 3813

* [ ] Yes
* [x] No


src/clr-angular/forms/datalist/providers/datalist-controller.service.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class DatalistControllerService {
+  private _dataItems: string[] = [];
```
bdryanovski: Is there any reason for having setter/getter for `dataItems` can we just go with 
```ts
public dataItems: string[] = [];
```

src/clr-angular/forms/datalist/datalist-container.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  @ContentChild(ClrLabel, { static: false })
+  label: ClrLabel;
+
+  controlClass() {
+    return this.controlClassService.controlClass(this.invalid, this.addGrid());
+  }
+
+  addGrid() {
+    if (this.layoutService && !this.layoutService.isVertical()) {
```
bdryanovski: nit: it could be shorter

```ts
return this.layoutService && !this.layoutService.isVertical();
```

AlexMarcuBytex: you re right

src/clr-angular/forms/datalist/providers/datalist-controller.service.ts
AlexMarcuBytex: for now, i guess i can keep it without getter/setter. also, i have to mention here that i use this simple service as an alternative for communication between datalist container and datalist directive. my first option was to use content child in container to receive this data input, but this approach creates a circular dependency.

src/clr-angular/forms/datalist/datalist-container.ts
```diff
@@ -0,0 +1,107 @@
...
+@Component({
+  selector: 'clr-datalist-container',
+  template: `
+    <ng-content select="label"></ng-content>
+    <label *ngIf="!label && addGrid()"></label>
+    <div class="clr-control-container" [ngClass]="controlClass()">
+      <div class="clr-input-wrapper">
+        <div class="clr-input-group" [class.clr-focus]="focus">
+          <ng-content select="[clrDatalist]"></ng-content>
+          <datalist id="data-list">
```
Jinnie: I believe we will need unique id here. Either generate one or take it from the user-defined input[list] attribute.

Jinnie:
Also I don't see an index.ts file. It should be added to the forms/index.ts too, if we want it to be available in `@clr/angular` imports.

src/clr-angular/forms/datalist/datalist-container.ts
AlexMarcuBytex: added an auto generated way.

src/clr-angular/forms/datalist/datalist-container.ts
```diff
@@ -0,0 +1,112 @@
...
+  listId: string;
+
+  constructor(
+    private datalistControllerService: DatalistControllerService,
+    private controlClassService: ControlClassService,
+    private layoutService: LayoutService,
+    private ifErrorService: IfErrorService,
+    private focusService: FocusService,
+    private ngControlService: NgControlService,
+    @Inject(UNIQUE_ID) public id: string
```
Jinnie: We don't need this to be public, it's a temporary unique id, which is better to stay encapsulated. `listId` is the real ID that we want to expose as part of the API

src/clr-angular/forms/datalist/datalist-container.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.subscriptions.push(
+      this.focusService.focusChange.subscribe(state => {
+        this.focus = state;
+      })
+    );
+    this.subscriptions.push(
+      this.ngControlService.controlChanges.subscribe(control => {
+        this.control = control;
+      })
+    );
```
coryrylan: I believe you can clean this up a bit by combining the subscriptions into the subscription push call.

```typescript
  this.subscriptions.push(
    this.ifErrorService.statusChanges.subscribe(invalid => this.invalid = invalid),
    this.focusService.focusChange.subscribe(state => this.focus = state),
    this.ngControlService.controlChanges.subscribe(control => this.control = control)
  );
```

src/dev/src/app/datalist/datalist.demo.html
```diff
@@ -0,0 +1,35 @@
...
+
+<h2>Horizontal <button class="btn btn-sm btn-primary" (click)="disabled = !disabled">Toggle Disabled</button></h2>
+<form clrForm clrLayout="horizontal">
+  <clr-datalist-container>
+    <input clrDatalist [clrData]="data" [(ngModel)]="vertical.one" placeholder="No label, wrapper" name="Option2"/>
+  </clr-datalist-container>
+
+  <clr-datalist-container>
+    <label>Full example</label>
+    <input clrDatalist [clrData]="data" [(ngModel)]="vertical.two" placeholder="Full example" required name="Option3" minlength="8" />
```
coryrylan: Instead of passing in a `Input` of `clrData` we can keep the datalist element ouside of the component and allow the consumer to declarativley create the list. We typically will try to keep things declarative if possible since it gives the consumer more flexibility.


```html
<clr-datalist-container>
  <label>Full example</label>
  <input clrDatalist [(ngModel)]="vertical.two" placeholder="Full example" required name="Option3" minlength="8"  />
  <datalist>
    <option *ngFor="let item of items" [value]="item"></option>
  </datalist>
</clr-datalist-container>
```

AlexMarcuBytex: This can be done. My first thought was that having the datalist into container component would be easier to use on the consumer's side and i would have more control over how the datalist and input are linked through a preset id. My question now is whether we will have to let the consumer set the ids himself or how should we proceed?

coryrylan: You can follow a similar pattern to the other inputs which auto preset the id if a custom id is not present.

src/clr-angular/forms/datalist/datalist-container.ts
```diff
@@ -0,0 +1,78 @@
...
+import { ClrLabel } from '../common/label';
+
+@Component({
+  selector: 'clr-datalist-container',
+  template: `
+    <ng-content select="label"></ng-content>
+    <label *ngIf="!label && addGrid()"></label>
+    <div class="clr-control-container" [ngClass]="controlClass()">
+      <div class="clr-input-wrapper">
+        <div class="clr-input-group" [class.clr-focus]="focus">
```
mathisscott: Just a question on my part here. Is there any benefit to using `[class.clr-focus]` over an `[ngClass]` object here?

I don't know which one is better or if it even matters in this case.

gnomeontherun: `[class.clr-focus]` is a shorthand so you don't have to create a function as well, yield the same result but this is more declarative in the template itself.

This PR is superceded by #4121 and will be closed by it.

gnomeontherun:
Closing in favor of #4121

PR #3897, opened 2019-10-09
chore: prevent fdescribe, fit, debugger from commiting

bdryanovski:
For the past week, I manage to push several times `fdescribe`, `debugger` and so on.

This is pre-commit hook that will prevent you from committing this type of content. It could be extended later on. 

There are NPM packages Git Hook packages that could do the same thing, but right now I don't think that there is a big need for that. 

In the rare case that you will need to commit change that include `fdescribe` or some of the other words. This could be done with: 

```bash
$ git commit -m "fix: skip few tests from running " -s -S --no-verify
```

Screen Shot 2019-10-09 at 8 40 16 PM

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "format": "prettier --list-different './**/*.{js,json,md,ts,scss}'",
     "format:fix": "pretty-quick --staged",
     "format:file": "prettier --write",
+    "git:check-content": "node ./scripts/commit-content-check.js",
     "gemini": "npm-run-all core:sass 'gemini:light -- {*}' 'gemini:dark -- {*}' --",
     "gemini:light": "./scripts/docker-cdc.js -t",
     "gemini:dark": "./scripts/docker-cdc.js -t -c dark",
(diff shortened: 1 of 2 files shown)


coryrylan:
TSLint will prevent the `debugger` keyword and there is also TSLint rules for Jasmine https://github.com/kaiza/tslint-jasmine-rules While these wont catch at time of commit they would catch at build time and prevent it from being merged.

gnomeontherun:
Yes, we shouldn't duplicate what TSLint does by writing more scripts. I've also avoided adding linting to commit hooks because it can be a bit frustrating at times when you're doing a lot of exploratory work.

bdryanovski:
I'm ok with this checks, but don't see the reason that this has to be committed and tested on the remote side when it could be prevented and save time waiting for the build to fail. Right now nothing is preventing me to do it and no CI is telling me that I'm stupid and forgot the enable all tests or most of the time save the file.

Also, I'm thinking that a check for `<<<<<<<`, `>>>>>>>` and `=======`. So non umerged files could be commited.

gnomeontherun:
I'm going to close this as something we don't want right now. You might be able to commit merge conflicts, but it won't build so will get flagged on Travis. If you want to run the checks locally you certainly can.

Issue #3737, opened 2019-08-23
[WEB] Custom filters example incorrect and ambiguous


Jinnie:
## Describe the bug

An example on the custom filters website page incorrectly adds clrDgItem property to a clr-dg-cell, while it should be using clr-dg-column instead.

## How to reproduce

1.  Go to https://clarity.design/documentation/datagrid/custom-filtering
2.  Scroll down to the example which description starts with `By default, filtering searches the original model value for matches`
3.  Verify that the example contains (at 2 places):
> <clr-dg-cell [clrDgField]="...

## Expected behavior

> <clr-dg-column [clrDgField]="

## Additional notes

The whole example is a bit ambiguous as it is not clearly stated that it illustrates problematic behavior with **default** filtering.
We have a case where it was understood like example of how it needs to be done.
I suggest that we find a way to explicitly and visibly identify that this is not a positive, but a negative example.

amellnik:
I can take a swing at this -- I was going to make a PR to propose other changes to the DataGrid docs (like how the usage notes appear at the bottom of every section).

gnomeontherun:
Awesome, I'll assign you!

Jinnie:
Verified this has been fixed in both websites

PR #3903, opened 2019-10-13
Tweak datagrid docs

amellnik:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

A code snippet in the custom filters section of the datagrid docs contains a typo.  Usage notes for Datagrids appear in top level document component.

Issue Number: #3737 (and also makes other changes)

## What is the new behavior?

Fixes the typo and moves the usage notes out to a child component

* [ ] Yes
* [x] No

### DIFF
--- a/src/website/src/app/documentation/demos/datagrid/datagrid.demo.module.ts
+++ b/src/website/src/app/documentation/demos/datagrid/datagrid.demo.module.ts
@@ -11,6 +11,7 @@ import { Routes, RouterModule } from '@angular/router';
 import { ClarityModule } from '@clr/angular';
 
 import { DatagridBasicStructureDemo } from './basic-structure/basic-structure';
+import { DatagridUsageDemo } from './usage/usage';
 import { DatagridBatchActionDemo } from './batch-action/batch-action';
 import { DatagridBindingPropertiesDemo } from './binding-properties/binding-properties';
 import { DatagridCustomRenderingDemo } from './custom-rendering/custom-rendering';
@@ -179,6 +180,13 @@ const routes: Routes = [
           demoName: 'Full Demo',
         },
       },
+      {
+        path: 'usage',
+        component: DatagridUsageDemo,
+        data: {
+          demoName: 'Usage',
+        },
+      },
     ],
   },
 ];
@@ -187,6 +195,7 @@ const routes: Routes = [
   imports: [CommonModule, FormsModule, ClarityModule, DocWrapperModule, RouterModule.forChild(routes), UtilsModule],
   declarations: [
     DatagridBasicStructureDemo,
+    DatagridUsageDemo,
     DatagridBatchActionDemo,
     DatagridBindingPropertiesDemo,
     DatagridCompactDemo,
--- a/src/website/src/app/documentation/demos/datagrid/filtering/examples.ts
+++ b/src/website/src/app/documentation/demos/datagrid/filtering/examples.ts
@@ -118,16 +118,23 @@ class ColorFilter implements ClrDatagridFilterInterface<User> {
 </clr-dg-column>
 `,
   filterSearchResults: `
-<-- 
-  Will search into user.creation that in this case is 'Date' object
-  and not the result of the pipe 'Jan 6, 2018' 
--->
-<clr-dg-cell [clrDgField]="'creation'">{{ user.creation | date }}</clr-dg-cell>
+<clr-datagrid>
+    <-- 
+      This will search into user.creation (a Date object) and not the result
+      of the pipe (the string 'Jan 6, 2018').
+    -->
+    <clr-dg-column [clrDgField]="'creation'">Creation Date</clr-dg-column>
 
-<-- 
-  Will search into user.name and will not include user.id, 
-  searching for user.id will not return any result.
--->
-<clr-dg-cell [clrDgField]="'name'">{{ user.id }} : {{ user.name }}</clr-dg-cell>
+    <-- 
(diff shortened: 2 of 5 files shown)


amellnik:
I'm not sure what's going on with the applitools test, but it says that it's related to toggles and I can't see the details -- probably unrelated?

bdryanovski:
We have some issues with Applitools and we are working on it

src/website/src/app/documentation/demos/datagrid/datagrid.demo.module.ts
```diff
@@ -53,6 +54,13 @@ const routes: Routes = [
           demoName: 'Basic Structure',
         },
       },
+      {
+        path: 'usage',
```
Jinnie: About extraction of the usage section - I like it, but I'd love to hear from other team members, too. And maybe I'd put it at the end of the list, not after Basic Structure.

src/website/src/app/documentation/demos/datagrid/filtering/examples.ts
```diff
@@ -122,12 +122,12 @@ class ColorFilter implements ClrDatagridFilterInterface<User> {
   Will search into user.creation that in this case is 'Date' object
   and not the result of the pipe 'Jan 6, 2018' 
 -->
-<clr-dg-cell [clrDgField]="'creation'">{{ user.creation | date }}</clr-dg-cell>
+<clr-dg-column [clrDgField]="'creation'">{{ user.creation | date }}</clr-dg-column>
```
Jinnie: Here the problem is more complicated. It's that this whole code example is NOT directly related to how you use the component, but illustrates a situation where custom filters may need to be used - as when you want to filter by date, but it will "try" to filter based on the js date object, instead of the representation (as shown in the comment above). But people quickly scan the examples and try to copy and use this snippet which is wrong on many levels. I'd either remove the `[clrDgField]="'creation'"` and leave the `clr-dg-cell`, or add 1st the column then the cell below it. This still will not help much with perceiving the snipped of "example of the example", but at least people will not be copying the wrong code.
With your change the example will still be wrong, as it will be trying to put user.creation field in the column heading :)

src/website/src/app/documentation/demos/datagrid/datagrid.demo.module.ts
amellnik: Agreed -- I think the end of the list (or possibly as the first item) would be better than the current location.

src/website/src/app/documentation/demos/datagrid/filtering/examples.ts
amellnik: Whoops -- I went through this too quickly.  I'll fix this later today.

Jinnie: It only proves that what we have there is really unclear :D

package-lock.json
```diff
@@ -6145,7 +6145,7 @@
     },
     "callsites": {
       "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
```
gnomeontherun: Please revert lock file changes.

src/website/src/app/documentation/demos/datagrid/datagrid.demo.module.ts
amellnik: Now showing this at the end.

src/website/src/app/documentation/demos/datagrid/filtering/examples.ts
```diff
@@ -118,16 +118,23 @@ class ColorFilter implements ClrDatagridFilterInterface<User> {
...
 `,
   filterSearchResults: `
-<-- 
-  Will search into user.creation that in this case is 'Date' object
-  and not the result of the pipe 'Jan 6, 2018' 
--->
-<clr-dg-cell [clrDgField]="'creation'">{{ user.creation | date }}</clr-dg-cell>
+<clr-datagrid>
+    <-- 
+      This will search into user.creation (a Date object) and not the result
```
Jinnie: The comment here refers to "the result of the pipe", which you already removed.
I suggest adding back the 
"<clr-dg-cell ~[clrDgField]="'creation'"~>{{ user.creation | date }}</clr-dg-cell>"
after the `clr-dg-column` definition. It's okay to have both lines, as long as they are correct.

amellnik: I made this code snippet a bit longer so that it includes two column definitions (with the comments) as well as the row definitions below (which includes the date pipe).  Does that make sense or is the row part too far away?

amellnik: @Jinnie Just following up -- does that make sense?  I could also move the comments above the datagrid definition if you think that would make it easier to read.

gnomeontherun:
@Jinnie you have a block on this PR, can you review?

coryrylan:
@amellnik thanks for the contribution! It looks like there is a commit that is not properly signed. Could you go through the five commits and check to see if they are all have a commit signature as described by the vmwclabot above? `Signed-off-by: John Doe <john.doe@email.org>`

Everything looks good, once signed off I can merge your PR into master 👍

amellnik:
@coryrylan -- Am I missing something?  I think all the commits should have the DCO and the bot seems happy.

image

gnomeontherun:
@amellnik Can we squash this into one commit? The DCO looks right.

Issue #3932, opened 2019-10-22
clrDroppable does not work with fullTemplateTypeCheck


derkoe:
## Describe the bug

When using `clrDroppable` in an Angular application with `"fullTemplateTypeCheck"` activated in `"angularCompilerOptions"` then the application does not compile with:

    Directive ClrDroppable, Property 'isDraggableMatch' is private and only accessible within class 'ClrDroppable<T>'

## How to reproduce

- Create a new Angular app: `ng new my-app` (current Angular CLI sets `"fullTemplateTypeCheck": true` in `angularCompilerOptions` in `tsconfig.json`).
- Add Clarity `ng add @clr/angular`
- Add the following to `app.component.html`:
  ```
  <div clrDraggable>Drag this</div>
  <div clrDroppable>Drop area</div>
  ```
-  Run `ng build --prod`

## Expected behavior

The application should build with `"fullTemplateTypeCheck": true`.

## Versions

**App**

* Angular: 8.1.3
* Node: v10.16.3
* Clarity: 2.3.0

**Device:**

* Type: [e.g. MacBook]
* OS: Win10
* Browser: all
* Version: all

## Additional notes

Already created a fix in https://github.com/porscheinformatik/clarity/tree/fix/full-template-type-check

gnomeontherun:
This was closed by #3933.

PR #3933, opened 2019-10-22
fix: templates for private/unused variables

derkoe:
Fix templates for private/unused variables.

- This enables to set `"fullTemplateTypeCheck": true` in `angularCompilerOptions` (current Angular CLI sets this).
- Enable `fullTemplateTypeCheck` in dev application

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Angular prod build fails when `"fullTemplateTypeCheck": true` is set in `angularCompilerOptions` in `tsconfig.json`.

Issue Number: #3932

## What is the new behavior?

Angular build works :-)

* [ ] Yes
* [x] No

## Other information

None

### DIFF
--- a/src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
+++ b/src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
@@ -34,7 +34,7 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
     this.droppableEl = this.el.nativeElement;
   }
 
-  private isDraggableMatch: boolean = false;
+  isDraggableMatch: boolean = false;
   private _isDraggableOver: boolean = false;
 
   set isDraggableOver(value: boolean) {
--- a/src/dev/tsconfig.dev.json
+++ b/src/dev/tsconfig.dev.json
@@ -6,6 +6,5 @@
     "moduleResolution": "node",
     "target": "es5"
   },
-
   "exclude": ["test.ts", "**/*.spec.ts"]
 }
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -707,6 +707,7 @@ export declare class ClrDroppable<T> implements OnInit, OnDestroy {
     dropEmitter: EventEmitter<ClrDragEvent<T>>;
     dropTolerance: number | string | ClrDropToleranceInterface;
     group: string | string[];
+    isDraggableMatch: boolean;
     isDraggableOver: boolean;
     constructor(el: ElementRef, eventBus: DragAndDropEventBusService<T>, domAdapter: DomAdapter, renderer: Renderer2);
     ngOnDestroy(): void;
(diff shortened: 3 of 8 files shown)


derkoe:
Forgot the actual fix - latest commit should be alright.

src/dev/tsconfig.dev.json
```diff
@@ -6,6 +6,9 @@
     "moduleResolution": "node",
     "target": "es5"
   },
-
+  "angularCompilerOptions": {
+    "fullTemplateTypeCheck": true,
+    "strictInjectionParameters": true
+  },
```
coryrylan: This looks good. As a reminder for myself it would be good to see if we can add this to the website docs as well.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
```diff
@@ -34,7 +34,7 @@ export class ClrDroppable<T> implements OnInit, OnDestroy {
     this.droppableEl = this.el.nativeElement;
   }
 
-  private isDraggableMatch: boolean = false;
+  isDraggableMatch: boolean = false;
```
coryrylan: Do host bindings have to be public with the compiler check?

src/dev/tsconfig.dev.json
derkoe: I think this doesn't need any mention. The current Angular CLI add this to tsconfig.json when generating a new project.

src/clr-angular/utils/drag-and-drop/droppable/droppable.ts
derkoe: Yes, the app under `src/dev` (and any other application with `fullTemplateTypeCheck`) does not compile with private bindings. So if you follow the instructions in the docs and use `clrDroppable` your app won't compile. See issue #3932

src/dev/tsconfig.dev.json
```diff
@@ -6,6 +6,9 @@
     "moduleResolution": "node",
     "target": "es5"
   },
-
+  "angularCompilerOptions": {
```
gnomeontherun: My ask is to move this to the root `tsconfig.json` instead so its the default project wide.

gnomeontherun:
@derkoe Just checking, if we make that little update I think we can merge it thanks!

src/dev/tsconfig.dev.json
derkoe: I can try but it was not my intend to change the build settings for the whole project - just to fix #3932

josfh2005:
Hi @gnomeontherun, is there any chance this fix will be applied to v2?

hippee-lee:
Hi @josfh2005 - v2 is out of support as of January 1, 2021. You can find the Clarity support policy for versions [here](https://clarity.design/get-started/support/#support-policy).

Issue #3822, opened 2019-09-13
Date Picker Live Regions are not working with NVDA/Firefox


clane:
## Describe the bug

Date Picker Live Regions are not working with NVDA/Firefox.

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

1.  Using NVDA/Firefox, go to https://v2.clarity.design/datepicker
2.  Activate one of the date pickers
3.  Activate any of the calendar-switcher buttons like the one below:

```
<div class="calendar-switchers">
<button class="calendar-btn switcher" type="button" aria-label="Previous month"><clr-icon dir="left" shape="angle" role="none" title="Previous month"><svg version="1.1" viewBox="0 0 36 36" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" focusable="false" role="img" aria-labelledby="_clr_icon_13"><path class="clr-i-outline clr-i-outline-path-1" d="M29.52,22.52,18,10.6,6.48,22.52a1.7,1.7,0,0,0,2.45,2.36L18,15.49l9.08,9.39a1.7,1.7,0,0,0,2.45-2.36Z"></path></svg><span class="is-off-screen" id="_clr_icon_13">Previous month</span></clr-icon></button>

```

4. Observe that you do not hear the contents of the datepickers live region being announced. The Live Region is below for reference.

```
<div class="calendar-header">
<div aria-live="polite" class="clr-sr-only"> The current month is Sep. The current year is 2019. </div>
```

## Expected behavior

NVDA users should hear live region updates

## Versions

**App**

* Angular: NA
* Node: NA
* Clarity: 2.2.1

**Device:**

* Type: any
* OS: Windows 10
* Browser: Chrome
* Version: 76

## Additional notes

See the following example that has a live region that works with NVDA/Firefox and the Clarity Datepicker that does not.
https://stackblitz.com/github/clane/ClarityExamples/tree/master/clr-examples

Also see my findings comparing dynamic population of live regions versus dynamically adding them as a whole at the following URL.

PR #3934, opened 2019-10-22
feat: Aria live service to handle SR across browsers

bdryanovski:
New service to handle the ARIA-LIVE attribute across browsers. For the moment the service and the API are private for internal use only!

Open for discussion and changes.

## PR Checklist

* [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

* [x] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3822

## What is the new behavior?

Creating a service that will create a new DOM element inside the document body. Set the element to have `aria-live` attribute and project content inside so that the SR will catch the change and announce it. 

* [ ] Yes
* [x] No

## Other information

Close: #3822

Тhe service will create new DOM for every component that requires aria-live support. I manage to test it with about 200 new dom elements and did not find any issues on the performance part. On destroy, for every component, the DOM element must be removed.

Every aria-live container will and must have uniq #id


src/clr-angular/forms/common/error.ts
```diff
@@ -18,6 +20,17 @@ import { ControlIdService } from './providers/control-id.service';
...
-  constructor(@Optional() public controlIdService: ControlIdService) {}
+export class ClrControlError implements AfterViewInit {
+  constructor(
+    @Optional() public controlIdService: ControlIdService,
+    private ariaLiveService: AriaLiveService,
+    private el: ElementRef
+  ) {}
+
+  ngAfterViewInit() {
+    const textNode = this.el.nativeElement.childNodes[0];
```
coryrylan: This will need a browser platform check `if (isPlatformBrowser(this.platformId)) {}`

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    return ariaLiveElement;
+  }
+
+  public announce(message: string): Promise<void>;
+  public announce(message: string, politeness?: AriaLivePoliteness): Promise<void>;
+  public announce(message: string, ...args: any[]): Promise<void> {
+    let politeness = 'polite';
+    if (args[0] && typeof args[0] === 'string') {
+      politeness = args[0];
+    }
```
coryrylan: This might be able to be cleaned up by using a default param and making `AriaLivePoliteness` an Enum instead of a Type.

```typescript
announce(message = AriaLivePoliteness.off): Promise<void> { }
```

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { DOCUMENT } from '@angular/common';
+import { Inject, Injectable, NgZone, OnDestroy } from '@angular/core';
+import { uniqueIdFactory } from '../id-generator/id-generator.service';
+
+export type AriaLivePoliteness = 'off' | 'polite' | 'assertive';
```
coryrylan: A Enum could make this a little cleaner so the referencing/consuming code does not have to introduce additional string usage.

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -3,17 +3,27 @@
...
-export class ClrProgressBar {
+export class ClrProgressBar implements AfterViewInit {
+  constructor(private el: ElementRef, private ariaLiveService: AriaLiveService) {}
+
+  ngAfterViewInit() {
+    const textNode = this.el.nativeElement.childNodes[0];
+    if (textNode) {
+      this.ariaLiveService.announce(textNode.nodeValue, this.ariaLive);
+    }
+  }
```
coryrylan: This will need a browser platform check `if (isPlatformBrowser(this.platformId)) {}`

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -16,7 +18,17 @@ import { isBooleanAttributeSet } from '../../utils/component/is-boolean-attribut
...
+export class ClrSpinner implements AfterViewInit {
+  constructor(private el: ElementRef, private ariaLiveService: AriaLiveService) {}
+
+  ngAfterViewInit() {
+    const textNode = this.el.nativeElement.childNodes[0];
+    if (textNode) {
+      // Tested: when the string is different, Chrome is reading it
+      this.ariaLiveService.announce(textNode.nodeValue, this.setAriaLive);
+    }
+  }
```
coryrylan: This will need a browser platform check `if (isPlatformBrowser(this.platformId)) {}`

src/dev/src/app/accessibility/accessibility.demo.routing.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AccessibilityDemo } from './accessibility.demo';
+import { AriaLiveServiceDemo } from './aria-live-service.demo';
+
+const ROUTES: Routes = [
+  {
+    path: '',
+    component: AccessibilityDemo,
+    children: [
+      { path: '', redirectTo: 'array', pathMatch: 'full' },
+      { path: 'aria-live', component: AriaLiveServiceDemo },
```
coryrylan: Why redirect? Could just navigate to `aria-live`?

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,91 @@
...
+
+@Injectable(/*{ providedIn: 'root' }*/)
```
coryrylan: Why is `providedIn` commented out?

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: I like the idea 👍

src/dev/src/app/accessibility/accessibility.demo.routing.ts
bdryanovski: I had the idea of adding another page, but yes for now is better to skip the redirect

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: For testing .. will remove it

gnomeontherun:
Adding @coryrylan to do a peer review.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  private _TICK: number = 100;
+  private _ariaLiveElement: HTMLElement;
+  private _document: Document;
+  private _previousTimeout?: number;
+
+  private _id: string = `clr-aria-live-element-${uniqueIdFactory()}`;
+
+  constructor(private _ngZone: NgZone, @Inject(DOCUMENT) _document: any) {
+    this._document = _document;
+    this._ariaLiveElement = this._createLiveElement();
```
Shijir: We need to check the platform here or in the method, `_createLiveElement`.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private _createLiveElement(): HTMLElement {
```
Shijir: Can we rename it to more descriptive name? My suggestion is to `createLiveAriaContainer` What kind of element is it trying to create?

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (this._ariaLiveElement) {
+      this._ariaLiveElement.textContent = '';
+    }
+  }
+
+  public ngOnDestroy() {
+    clearTimeout(this._previousTimeout);
+
+    if (this._ariaLiveElement && this._ariaLiveElement.parentNode) {
+      this._ariaLiveElement.parentNode.removeChild(this._ariaLiveElement);
```
Shijir: You appended the child to the body element(on line: 43) so we know it's in the body element. Just do: 
`this.document.removeChild(this._ariaLiveElement);`

Also, we need to check the platform here.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      return new Promise(resolve => {
+        clearTimeout(this._previousTimeout);
+        /**
+         * Cause I didn't want to use `window.setTimeout` so TypeScript will stop throwing
+         * this error `Type 'Timer' is not assignable to type 'number'` , I had to use `<any>` before
+         * setTimeout()
+         *
+         * More on the topic here: https://github.com/TypeStrong/atom-typescript/issues/1053
+         */
+        this._previousTimeout = <any>setTimeout(() => {
```
Shijir: I've encountered this problem before. I think we are better off just having `any` type on the `_previousTimeout` property.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+@Injectable()
+export class AriaLiveService implements OnDestroy {
+  private _TICK: number = 100;
+  private _ariaLiveElement: HTMLElement;
+  private _document: Document;
+  private _previousTimeout?: number;
+
+  private _id: string = `clr-aria-live-element-${uniqueIdFactory()}`;
```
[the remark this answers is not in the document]
bdryanovski: I'm ok with that I also don't like the idea of having the prefix but our code is full of it - So I'm trying to keep it consistence.

```bash
$ ag 'private _' | wc -l
     323
```
If we agree that this will be changed in the feature I'm ok with that :)

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: sounds better

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    this._ariaLiveElement.setAttribute('aria-live', politeness);
+
+    // This 100ms timeout is necessary for some browser + screen-reader combinations:
+    // - Both JAWS and NVDA over IE11 will not announce anything without a non-zero timeout.
+    // - With Chrome and IE11 with NVDA or JAWS, a repeated (identical) message won't be read a
+    //   second time without clearing and then using a non-zero delay.
+    // (using JAWS 17 at time of this writing).
+    return this._ngZone.runOutsideAngular(() => {
+      return new Promise(resolve => {
```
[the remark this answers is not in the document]
bdryanovski: My use here is this 
```ts
ariaLiveService.announce('Some update text')
  .then(() => { this.triggerSomeMethodd });
```

You could know exactly when the event is pushed and could take some action after it.

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: I prefer to keep it like that. I think that `_previousTimeout` is the property that I want to know its type. `setTimeout` is just buggy and there is a mismatch between javascript and nodejs types here. The other solution, if I don't want to annotate them with strange types, is to use the `window.setTimeout` but will have to wrap them inside another platform check.

So my point here is, prefer to know my custom property type than to know the type of native method that won't change the next 100 years ;)

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: Yep this could be optimized

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: Better to be tested here that inside the `_createLiveElement`

src/clr-angular/utils/a11y/aria-live.service.ts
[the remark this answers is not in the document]
coryrylan: Agree it should not have prefix. Some of the code is not consistent on the part but we are trying to improve. Remember also for getters and setters the private backing fields typically are prefixed with `_` which is valid.

src/clr-angular/utils/a11y/aria-live.service.ts
Shijir: > <any>setTimeout(() => {

you are type-casting whatever the type of setTimeout to `<any>`. So there is no point in assigning `number` type to the property. If we want to be more strict on the type, 
we should do:
```
private previousTimeout: ReturnType<typeof setTimeout>;
```

src/clr-angular/utils/a11y/aria-live.service.ts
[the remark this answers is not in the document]
Shijir: But we don't have a use case for it currently though. I don't see any place where we are using this promise. We shouldn't just add something when we don't know for sure it would be useful or not in the future. 

And if we really need to listen to this event, why Promise...? Why not Observable?

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,113 @@
...
+  public clear() {
+    if (this.ariaLiveElement) {
+      this.ariaLiveElement.textContent = '';
+    }
+  }
+
+  public ngOnDestroy() {
+    clearTimeout(this.previousTimeout);
+
+    if (this.ariaLiveElement && this.ariaLiveElement.parentNode) {
```
Shijir: We are still missing platform check here.

src/clr-angular/utils/a11y/aria-live.service.ts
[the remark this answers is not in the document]
Shijir: We only try to use underscore prefixes for properties that have getter/setters or public properties that should be only used internally. In all other cases, underscore prefix is not recommended.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,113 @@
...
+import { uniqueIdFactory } from '../id-generator/id-generator.service';
+
+export enum AriaLivePoliteness {
+  off = 'off',
+  polite = 'polite',
+  assertive = 'assertive',
+}
+
+/**
+ * Time in miliseconds before inserting the content into the contianer
```
Shijir: nit: some spelling errors: miliseconds, contianer

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: We don't need the check for `this.ariaLiveElement.parentNode` any more and I will remove it.

For the platform check, don't see a reason to add another `&&` or `if` for that. There is a check inside the `constructor` that will prevent the creation of the `this.ariaLiveElement` and that is the only place where this variable is assigned. So the check for `undefined` will cover it.

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: Ok, let's do it like that :)

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  constructor(
+    private _ngZone: NgZone,
+    @Inject(DOCUMENT) _document: any,
+    @Inject(PLATFORM_ID) private platformId: Object
+  ) {
+    this.document = _document;
+  }
+
+  private createLiveAriaContainer(): HTMLElement {
```
Shijir: We have two methods that have almost the identical names, `createLiveContainer` 
 and `createLiveAriaContainer`, why?

Also, this method should handle inside itself whether it should create live aria container element or just exit if there is an already existing aria-live element.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    return ariaLiveElement;
+  }
+
+  public announce(message: string | HTMLElement, politeness: AriaLivePoliteness = AriaLivePoliteness.polite): void {
+    // Create aria-live container when needed.
+    if (this.createLiveContainer() === false) {
+      // when there is not ariaLiveElement we could not apply text to nothing so
+      // that's why we gonna break here.
+      return;
+    }
```
Shijir: We could simply call `createLiveAriaContainer` here without having it to return `boolean` value. 

As mentioned above, `createLiveAriaContainer` should handle inside itself whether it should create live aria container element or just exit if there is an already existing aria-live element.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    // handle DOM or string input
+    message = typeof message !== 'string' && isPlatformBrowser(this.platformId) ? message.textContent : message;
+
+    // when there is no message do NOTHING!
+    if (!message) {
+      return;
+    }
+
+    this.clear();
+    clearTimeout(this.previousTimeout);
```
Shijir: why do we have multiple `clearTimeout(this.previousTimeout);` here. It's used here, and there is another one at line:81. Although one is inside ` this._ngZone.runOutsideAngular`, both of them will run.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // handle DOM or string input
+    message = typeof message !== 'string' && isPlatformBrowser(this.platformId) ? message.textContent : message;
+
+    // when there is no message do NOTHING!
+    if (!message) {
+      return;
+    }
+
+    this.clear();
```
Shijir: I really don't see why we need to clean up the content beforehand. Assistive technologies will read the entire content change. On top of that, we have added the following line:
```
ariaLiveElement.setAttribute('aria-atomic', 'true');
```
So whatever we add to the aria-live element, assistive technologies treat changes as one whole thing.
>Assistive technologies will present the entire changed region as a whole, including the author-defined label if one exists.
https://www.digitala11y.com/aria-atomic-properties/

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    ariaLiveElement.setAttribute('aria-atomic', 'true');
+    ariaLiveElement.setAttribute('aria-live', AriaLivePoliteness.polite);
+
+    this.document.body.appendChild(ariaLiveElement);
+
+    return ariaLiveElement;
+  }
+
+  public announce(message: string | HTMLElement, politeness: AriaLivePoliteness = AriaLivePoliteness.polite): void {
```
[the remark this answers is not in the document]
coryrylan: Is there a use case where we don't want to make an announcement? Example multiple spinner components on a page. So example instead of the component passing in the `AriaLivePoliteness .off` state it would just contain the logic to not call the `announce()` method? Wouldn't the enum still need to exist for the component so the user can disable the announcement?

Seems like if you keep the enum then the component just passes it down to the service and the service can check for off and make sure it doesn't make any unnecessary DOM manipulations. I might be misunderstanding the use case for `off`

https://github.com/vmware/clarity/pull/3934/files/ef63c2334b3a3b3be820fcfc05644073ffdf71c3#diff-4b1f682e065c8811fe25073f1cffb482R120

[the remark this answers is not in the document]
coryrylan: Is there a use case where we don't want to make an announcement? Example multiple spinner components on a page. So example instead of the component passing in the `AriaLivePoliteness .off` state it would just contain the logic to not call the `announce()` method? Wouldn't the enum still need to exist for the component so the user can disable the announcement?

Seems like if you keep the enum then the component just passes it down to the service and the service can check for off and make sure it doesn't make any unnecessary DOM manipulations. I might be misunderstanding the use case for `off`

https://github.com/vmware/clarity/pull/3934/files/ef63c2334b3a3b3be820fcfc05644073ffdf71c3#diff-4b1f682e065c8811fe25073f1cffb482R120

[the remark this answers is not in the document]
Shijir: If consumer components don't want to announce anything, it simply wouldn't need to call the method from the aria-live service, right? Calling the announce method with `AriaLivePoliteness.off` itself doesn't make sense. Is it telling assistive technologies to announce the message or become silent? With `aria-life=off`, assistive technologies ignore all content change in the future.

On the side note, I haven't looked carefully into how the service is being used in consumer components. But I don't think what we have in the spinner component is correct. I will explain why... I think we need to discuss that it on Slack.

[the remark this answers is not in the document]
bdryanovski: I prefer to keep the `enum` for selecting and controlling the `politeness` of the aria. Something that I could be improved is:

```ts
  public announce(message: string | HTMLElement, politeness: AriaLivePoliteness = AriaLivePoliteness.polite): void {
    // Create aria-live container when needed.
    if (!this.ariaLiveElement && isPlatformBrowser(this.platformId)) {
      this.ariaLiveElement = this.createContainer();
    }

    // handle DOM or string input
    message = typeof message !== 'string' && isPlatformBrowser(this.platformId) ? message.textContent : message;

    // when there is no message do NOTHING!
    if (!message) {
      return;
    }

    this.ariaLiveElement.setAttribute('aria-live', politeness);
     
    if (politeness === AriaLivePoliteness.off) { return; }
   
    // ...
```
And not create new `setTimeout` and end the function there. The case with the clr-spinner is a valid one. And `off` could be used for "don't disturb" functionality. 

Going with `boolean` value to manage the state of the element is a bit hard to understand for me. I need to manage `polite`, `assertive` and `off` with `boolean` value that is `TRUE or FALSE` that is pretty ugly and may make people confused for example

```ts
ariaLiveService.announce('echo', true);
ariaLiveService.announce('echo', false);
ariaLiveService.announce('echo');
```
At first look, it looks like will announce it, the second will not, the last one .. not sure what will happen? When is `polite` and when `assertive` how to set it `off`? Just by not calling `announce` is not the answer - for example, I could have Progressbar that I don't want to announce at all but have to explain this to `clrProgressBar` and I don't have direct access to `aria-live` attribute. Maybe the progress bar is not the best example, cause it will create aria-live when there is some value to display, just empty progress bar with the value of `0` won't even create an aria-live element - here we use the last chance to start the announcing it. But still, have control in the middle of the progress to set it to off and stop it from announcing (mute it) and later on resume it without re-creating the DOM just by changing the argument - this is more efficient than removing and creating DOM.

```ts
ariaLiveService.announce('echo', 'polite');
ariaLiveService.announce('echo', 'assertive');
ariaLiveService.announce('echo', 'off');
```
Without reading the documentation I could tell what is going on. 
I went with the ENUM type but we could go with string if we really want to keep it simple. Right now we use enum that has a value of string so it's not more efficient than a simple `string` so there is no real benefit of it. But from the development point of view, it is more clear what is going on. People work better with strings than 1,0,null.

Also, don't see a huge optimization of the code by passing `boolean` and then using `if`s to handle it (branching of the code is more expensive than string values).
Our internal use of the aria-live is almost always `polite` and almost 80% of the time we don't provide an interface so it could be changed from outside. clrProgressBar, clrSpinner both components provide an option to change the `aria-live` by setting one of the attributes design do this `clrOff` `clrPolite`.

When going back on them I think that the naming is bit cryptic `clrOff` for what? but at the same time, `clrAriaLiveOff` is too long. Still, when we are providing an interface to native attribute don't see why we have to limit the values that people could set.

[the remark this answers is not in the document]
coryrylan: I think the Enum describing the state is fine. I probably would of made the same enum as well. As for the announce method I see two options:

Use a boolean to flag whether it should toggle `assertive` or not
`ariaLiveService.announce('my message', true);`

Or use a Enum to describe the announcement type 
`ariaLiveService.announce('my message', AriaLivePoliteness.polite);`

If you really wanted to enforce that `off` to not be an option in the service but keep the enum readability then just removing it from the Enum would accomplish the same as using a boolean. I am fine with the enum API since it is just as readable and does offer a bit of minor flexibility if new announcements were needed. The components however should not call the service at all if the AriaLivePoliteness of that component is `off` as its just unnecessary. 

@bdryanovski has a good point on the component API being a bit cryptic and eventually could be realigned better in a later PR with a deprecation. Maybe the components could of had a single input using the enum values instead of two. 

Overall this API should be internal so I don't feel that it is worth holding up since both are reasonably readable.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Time in milliseconds before inserting the content into the container
+ */
+export const ARIA_LIVE_TICK: number = 100;
+
+/**
+ * This service handle `aria-live` accessibility attribute. The issue is that you need
+ * to have the DOM Element with attribute `aria-live` before you could insert content
+ * and SR pick the change and announce it.
```
coryrylan: Thanks for the TSDoc comments!

src/clr-angular/utils/a11y/aria-live.service.ts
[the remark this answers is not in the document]
bdryanovski: OK, I will keep the Enum as the second param on the `announce` method. But added a check to prevent creating or updating the container when it is off. This way NO ACTION will be done, in the same, the API will stay the same.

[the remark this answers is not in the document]
Shijir: The following block of code is needed in almost all consumer components just to pass the correct argument in the method. 
```
get ariaLive(): AriaLivePoliteness {
    if (isBooleanAttributeSet(this.assertive)) {
      return 'assertive';
      return AriaLivePoliteness.assertive;
    }
    if (isBooleanAttributeSet(this.off)) {
      return 'off';
      return AriaLivePoliteness.off;
    }
    return 'polite';
    return AriaLivePoliteness.polite;
  }
```
The is the main drawback I am seeing in the current pattern. If you had the argument with boolean, you could actually just call it like this:

```
ariaLiveService.announce('my message')
```
I like what Cory suggested with using a boolean to flag whether it should toggle assertive or not.

```
// it will announce politely by default
ariaLiveService.announce('my message'); 

// it will still announce but assertively now
ariaLiveService.announce('my message, true'); 
```

But even with a boolean parameter, it's not perfect. Maybe, the culprit of the issue is that we are trying to manipulate both aria-live mode as well as the text content of the element in the same method.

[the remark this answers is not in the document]
Shijir: Also, I just realized we let users control the aria-live mode through following input controls:
```
 @Input('clrPolite') polite: boolean = true;
 @Input('clrAssertive') assertive: boolean;
 @Input('clrOff') off: boolean;
```

Here, not only clrOff being unclear to users what it would do, the entire API will be confusing to users. I wonder why this is not a single input for this, ie: `@Input('clrAriaLive')`
In this case, we could have used that enum type like this: `@Input('clrAriaLive') ariaLive: AriaLivePoliteness`.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,176 @@
...
+      this.previousTimeout = setTimeout(() => {
+        this.ariaLiveElement.textContent = <string>message;
+      }, ARIA_LIVE_TICK);
+    });
+  }
+
+  /**
+   * Remove the content inside the AriaLive Container
+   */
+  public clear() {
```
Shijir: I don't see this method being used anymore. Shouldn't we remove it?

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -0,0 +1,176 @@
...
+  ) {
+    this.document = _document;
+  }
+
+  private id: string = `clr-aria-live-element-${uniqueIdFactory()}`;
+  /**
+   * get access to the internal HTML `id` that gonna be used for the AriaLive container.
+   * @return ID of the DOM Element as string.
+   */
+  public getId(): string {
```
Shijir: Why is this still a method? This method doesn't do anything. It should be just a getter property. 
```
get id(): string {
   return this._id;
}
```

src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -3,28 +3,52 @@
...
+export class ClrAlert implements AfterViewInit {
   constructor(
     public iconService: AlertIconAndTypesService,
     public cdr: ChangeDetectorRef,
     @Optional() public multiAlertService: MultiAlertService,
-    public commonStrings: ClrCommonStringsService
+    public commonStrings: ClrCommonStringsService,
+    private ariaLiveService: AriaLiveService,
+    @Inject(PLATFORM_ID) private platformId: Object,
+    private el: ElementRef
```
Shijir: Shouldn't we add this to the alert-item component instead? By using the service here, we are going to announce only the content of whatever alert-item is projected

src/clr-angular/forms/common/form.ts
```diff
@@ -78,45 +75,16 @@ export class ClrForm implements OnInit, OnDestroy {
...
-      this.el.nativeElement.removeChild(this.ariaLiveElement);
-      delete this.ariaLiveElement;
-    }
-  }
-
-  ngOnDestroy() {
-    this.removeAriaLiveElement();
+    this.ariaLiveService.announce(
+      this.commonStrings.parse(this.commonStrings.keys.formErrorSummary, { ERROR_NUMBER: errorList.length.toString() }),
+      AriaLivePoliteness.polite
```
Shijir: no need to pass AriaLivePoliteness.polite here

src/clr-angular/progress/spinner/spinner.spec.ts
```diff
@@ -51,34 +53,48 @@ describe('ClrSpinner component', () => {
       });
     });
 
-    it('expect to project content', () => {
+    it('should to project content', () => {
```
Shijir: nit: you have multiple `should to` in this file.

src/clr-angular/progress/progress-bars/progress-bar.ts
```diff
@@ -3,17 +3,27 @@
...
-export class ClrProgressBar {
+export class ClrProgressBar implements AfterViewChecked {
+  constructor(private ariaLiveService: AriaLiveService) {}
+
+  ngAfterViewChecked() {
+    if (this.displayAriaLive()) {
+      this.ariaLiveService.announce(this.displayValue, this.ariaLive);
+    }
+  }
+
```
coryrylan: So I misunderstood how `ngAfterViewChecked` worked. I mistakenly thought it was only called on the component and its direct children but it can be called during any change detection cycle. 

Screen Shot 2019-11-14 at 11 51 37 AM

As you can see here as I click an unrelated button CD is called and the component is writing to the DOM several times. The fix I found to work locally is to remove the `ngAfterViewChecked` and instead turn the `value` property into a getter/setter combo so we can call the announcement only when the value changes. This results in write more consistent with this:

Screen Shot 2019-11-14 at 12 19 06 PM

```typescript
  private _value = 0;
  @Input('clrValue')
  get value() {
    return this._value;
  }

  set value(value) {
    this._value = value;
    if (this.displayAriaLive()) {
      this.ariaLiveService.announce(this.displayValue, this.ariaLive);
    }
  }
```

src/clr-angular/progress/spinner/spinner.ts
```diff
@@ -3,20 +3,27 @@
...
     '[attr.aria-busy]': 'true',
   },
 })
-export class ClrSpinner {
+export class ClrSpinner implements AfterViewChecked {
+  constructor(private el: ElementRef, private ariaLiveService: AriaLiveService) {}
+
+  ngAfterViewChecked() {
+    this.ariaLiveService.announce(this.el.nativeElement, this.ariaLive);
+  }
```
coryrylan: The spinner has the same change detection issue as the progress bar but can simply be updated to use `ngAfterViewInit` instead of `ngAfterViewChecked`

src/clr-angular/utils/a11y/aria-live.service.ts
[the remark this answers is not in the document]
Shijir: Exiting on `off` is an improvement if we are going with the enum. 

> If you need to pass the third option or fourth you will have to use what? undefined and null - that will go messy really fast.

Why do you want to pass more stuff there? That's actually why things could go wrong. Methods should do only one thing and do it well. That's why I suggested that the culprit of the issue lies in trying to manipulate both aria-live mode as well as the text content of the element in the same method. Assistive technologies announce only in two different modes: `polite` or `assistive`. It's binary. There is no `off` announcement; it just mutes the future announcements. 

That being said, in the future, we could refactor and improve this method like this:

```
ariaLiveSerivce.setMode(mode: AriaLiveMode); 

ariaLiveSerivce.announce(message: string); 
```

This will allow us to set the aria-live mode of the service independently from calling the announce method, which should only change the text content. But we accept aria-live mode in separate inputs ( `@Input('clrPolite')`, `@Input('clrAssertive')`, `@Input('clrOff')` ) with boolean so we won't get much benefit from the refactoring above at this point.

src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -3,28 +3,52 @@
...
   ) {}
 
+  ngAfterViewInit() {
+    if (isPlatformBrowser(this.platformId)) {
+      const textNode = this.el.nativeElement.querySelector('.alert-text').childNodes[0];
```
Shijir: No need for `..childNodes[0]`. Just select `this.el.nativeElement.querySelector('.alert-text')` and pass it to the announce method.

src/clr-angular/emphasis/alert/alert.ts
Shijir: I was thinking of the case of `clr-alerts` with multiple `clr-alert`.  Currently, it will announce all of them at initialization no matter they are displayed or not.

src/clr-angular/progress/progress-bars/progress-bar.ts
bdryanovski: Oh damn! Thanks for catching it @coryrylan. I will go and fix them right now!

src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -3,28 +3,52 @@
...
-export class ClrAlert {
+export class ClrAlert implements AfterViewInit {
   constructor(
     public iconService: AlertIconAndTypesService,
     public cdr: ChangeDetectorRef,
     @Optional() public multiAlertService: MultiAlertService,
-    public commonStrings: ClrCommonStringsService
+    public commonStrings: ClrCommonStringsService,
+    private ariaLiveService: AriaLiveService,
+    @Inject(PLATFORM_ID) private platformId: Object,
```
Shijir: Do we need this platformId anymore? There could be things we longer need in the imported modules too.

chlane:
The service works with Safari/VoiceOver and Firefox/NVDA but if used in rapid succession some updates will not be announced. It would be great if the consumer could be responsible for setting up the live region's overall container and a service like this one existed to populate the region. That would reduce our responsibility and empower our consumers to offer users more control of the screen reader user experience via the various options available to the aria-live, aria-atomic and aria-relevant properties

PR #4000, opened 2019-11-08
fix: datagrid clrDgFilterOpenChange was not fired

### LINKED ISSUE #3804: ClrDatagridFilter openChange event not being fired
## Describe the bug

The ClrDatagridFilter openChange event is not being fired.  The event name has been changed from openChanged (in v2.1.2) to openChange (in v2.2.0).

## How to reproduce

In version 2.1.2, I subscribed to the openChanged event and received events.  However, in version 2.2.0, I now am subscribed to the openChange event, but I am not receiving events.

I have provided a stackblitz.io example: https://angular-denw5k.stackblitz.io

## Expected behavior

I should receive openChange events.

## Versions

**App**

* Angular: 8.0.0
* Node: 10.14
* Clarity: 2.2.0

**Device:**

* Type: MacBook Pro
* OS: 10.14.6
* Browser Chrome
* Version Version 76.0.3809.132

bdryanovski:
`clrDgFilterOpenChange` was not called when `open` is toggle between `true|false`

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3804 

* [ ] Yes
* [x] No

## Other information

This change has to be added to `v2` branch as well. The original bug is from there.

Close #3804

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-filter.ts
+++ b/src/clr-angular/data/datagrid/datagrid-filter.ts
@@ -45,7 +45,7 @@ import { isPlatformBrowser } from '@angular/common';
               #anchor
               clrPopoverAnchor
               clrPopoverOpenCloseButton
-              [class.datagrid-filter-open]="open" 
+              [class.datagrid-filter-open]="open"
               [class.datagrid-filtered]="active">
           <clr-icon [attr.shape]="active ? 'filter-grid-circle': 'filter-grid'" class="is-solid"></clr-icon>
       </button>
@@ -63,7 +63,6 @@ import { isPlatformBrowser } from '@angular/common';
           <ng-content></ng-content>
       </div>
   `,
-  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>>
   implements CustomFilter, OnDestroy {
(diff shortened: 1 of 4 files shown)


src/clr-angular/data/datagrid/datagrid-filter.ts
```diff
@@ -63,7 +63,6 @@ import { isPlatformBrowser } from '@angular/common';
           <ng-content></ng-content>
       </div>
   `,
-  changeDetection: ChangeDetectionStrategy.OnPush,
```
gnomeontherun: This does seem like the root of the problem, what did changing the `set open` logic below change?

hippee-lee: Yes, I added this last year at the time of the first refactor for pinned columns. 
This was before the datagrid state refactor so maybe the issue I had is no longer relevant. 

Second with @gnomeontherun's question re: changed logic for the `open` getters and setters.
It seems to make more sense to me to keep it as a wrapper for `this.smartToggleService` rather than having an intermediary property that needs to be maintained.

bdryanovski: The main change is indeed is how `open` is tracked. The issue that I found is that the previews check:

```ts
public set open(open: boolean) {
  const boolOpen = !!open;
  if (boolOpen !== this.open) {
```
`boolOpen` and `this.open` is always the same. And the reason is that `this.open` return `return this.smartToggleService.open;` that is already changed when `set open` is called. 

For example, when the change is required we are tracking the `open` state by these lines:

```ts
    this.subs.push(
      smartToggleService.openChange.subscribe(change => {
        this.open = change;
      })
    );
```

and when it comes to executing the `if` we are passing for example `true` and `this.open` is already true cause it's set before even the subscribe method is invoked. So we never go in this `if` block, no event is emitted and so on.

gnomeontherun: Gotcha, yes that makes sense as the problem to have the smartPopover.open value set first, then it emits the same value (so `if (boolOpen !== this.open) {` will always be false because they will equal.

By tracking the state internally we get around this problem. This is something we're going to see in other places, so we need to keep this in mind, perhaps we can come up with a better default solution later.

Shijir: We should make sure that we remove `.markForChanges()` in this component now if are removing OnPush.

PR #4020, opened 2019-11-14
[NG] Reorder datagrid columns

Shijir:
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Currently, our Datagrid columns are not reorderable.

Issue Number: N/A

## What is the new behavior?

This PR enables reorderable columns in Datagrid.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -960,7 +960,7 @@
         display: none;
       }
 
-      .in-drag & {
+      .in-drag &:not(.datagrid-fixed-column) {
         position: relative;
         .draggable-match {
           display: flex;
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -22,7 +22,7 @@
     '.datagrid-header',
     '.datagrid-body',
     '.datagrid-row',
-    '.datagrid-column',
+    '.datagrid-column-wrapper',
     '.datagrid-cell'
   );
 
@@ -72,7 +72,7 @@
     }
 
     &.datagrid-column {
-      padding: $clr-table-topcellpadding $clr-table-cellpadding $clr-table-bottomcellpadding;
+      padding: 0;
     }
   }
 
@@ -552,12 +552,6 @@
         flex: 0 0 auto;
       }
     }
-
-    .datagrid-column:last-child {
-      .datagrid-column-separator {
-        display: none;
-      }
-    }
   }
 
   .datagrid-row-flex {
@@ -665,6 +659,23 @@
       flex: 1 1 auto;
       vertical-align: top;
       border: none;
+
+      .datagrid-column-droppable.draggable-match {
+        position: absolute;
+        top: 0;
+        left: 0;
+        width: 100%;
+        height: 100%;
+        display: none;
+        .datagrid-column-drop-line {
+          display: none;
+        }
+      }
+
+      .datagrid-column-wrapper {
+        flex: 1 1 auto;
+        width: 100%;
+      }
 
       clr-dg-filter,
       clr-dg-string-filter,
@@ -865,19 +876,19 @@
       .datagrid-column-separator {
         $dg-padding-less-border: $clr_baselineRem_0_5 - $clr_baselineRem_1px;
         position: relative;
-        left: $clr_baselineRem_0_5;
+        left: 0;
         flex: 0 0 auto;
         @include css-var(width, clr-global-borderwidth, $clr-global-borderwidth, $clr-use-custom-properties);
         order: 100;
         margin-left: auto;
         height: 100%;
 
-        &::after {
+        &::before {
           content: '';
           position: absolute;
-          height: calc(100% + #{$clr-datagrid-column-separator-expandby} - #{$clr-global-borderwidth});
+          height: calc(100% - #{$clr-datagrid-column-separator-expandby});
           width: $clr_baselineRem_1px;
-          top: calc(-0.5 * #{$clr-datagrid-column-separator-expandby} + #{$clr_baselineRem_1px});
+          top: $clr-datagrid-column-separator-expandby * 0.5;
           left: 0;
           @include css-var(
             background-color,
@@ -891,14 +902,14 @@
           position: absolute;
           width: $clr_baselineRem_0_5 + $clr_baselineRem_1px;
           right: -1 * $clr_baselineRem_0_25;
-          top: -1 * $clr_baselineRem_0_25;
+          top: 0;
           cursor: col-resize;
-          height: calc(100% + #{$clr_baselineRem_0_5} - #{$clr-global-borderwidth});
+          height: 100%;
           z-index: map-get($clr-layers, datagrid-header);
         }
         .datagrid-column-resize-tracker {
           position: absolute;
-          top: -1 * $clr_baselineRem_0_5;
+          top: 0;
           display: none;
           width: $clr_baselineRem_1px;
           height: 0;
@@ -912,6 +923,7 @@
           );
           transform: translateX(0px);
           cursor: col-resize;
+          z-index: map-get($clr-layers, draggable-ghost);
         }
 
         .exceeded-max {
@@ -934,6 +946,74 @@
       &.datagrid-row-actions {
         max-width: $clr-datagrid-fixed-column-size;
         min-width: $clr-datagrid-fixed-column-size;
+      }
+
+      &.datagrid-last-visible-column .datagrid-column-separator {
+        display: none;
+      }
+
+      &.datagrid-column-reorder-drop .datagrid-column-separator {
+        display: none;
+      }
+
+      &.datagrid-column-reorder-shift .datagrid-column-separator {
+        display: none;
+      }
+
+      .in-drag & {
+        position: relative;
+        .draggable-match {
+          display: flex;
+          border: none;
+          &.draggable-over {
+            .datagrid-column-drop-line {
+              position: relative;
+              display: block;
+              left: -1px;
+              width: 0;
+              height: 0;
+              margin-left: 0;
+              &::after {
+                content: '';
+                position: absolute;
+                display: block;
+                top: 0;
+                left: 0;
+                width: 2px;
+                height: 100vh;
+                background-color: $clr-action-purple-dark;
+                z-index: map-get($clr-layers, column-switch);
+              }
+            }
+          }
+        }
+        &.datagrid-first-visible-column .draggable-over {
+          .datagrid-column-drop-line {
+            left: 0;
+          }
+        }
+        &.datagrid-last-visible-column .draggable-over {
+          .datagrid-column-drop-line {
+            left: -2px;
+          }
+        }
+
+        &.datagrid-column-drag-mode .draggable-match {
+          display: none;
+        }
+
+        &.datagrid-column-drag-mode ~ .datagrid-column .draggable-match .datagrid-column-drop-line {
+          margin-left: auto;
+        }
+      }
+      .draggable-ghost .datagrid-column-wrapper {
+        @include clr-datagrid-column-ghost();
+      }
+
+      &.datagrid-column-reorder-drop {
+        .datagrid-column-wrapper {
+          @include clr-datagrid-column-ghost();
+        }
       }
     }
 
--- a/src/clr-angular/data/datagrid/datagrid-column-separator.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-separator.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -9,7 +9,7 @@ import { Component, Inject, Renderer2 } from '@angular/core';
 import { ClrDragEvent } from '../../utils/drag-and-drop/drag-event';
 import { ColumnResizerService } from './providers/column-resizer.service';
 import { TableSizeService } from './providers/table-size.service';
-import { UNIQUE_ID_PROVIDER, UNIQUE_ID } from '../../utils/id-generator/id-generator.service';
+import { UNIQUE_ID, UNIQUE_ID_PROVIDER } from '../../utils/id-generator/id-generator.service';
 
 @Component({
   selector: 'clr-dg-column-separator',
@@ -59,12 +59,11 @@ export class ClrDatagridColumnSeparator {
     this.renderer.setStyle(this.document.body, 'cursor', 'auto');
   }
 
+  private withinMaxResizeRange: boolean;
   private redFlagTracker(resizeTrackerEl: HTMLElement) {
-    let isWithinMaxResizeRange: boolean;
-    // @TODO(JEREMY) Review this, it will always be true because above is always null
-    if (isWithinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
-      isWithinMaxResizeRange = this.columnResizerService.isWithinMaxResizeRange;
-      if (!isWithinMaxResizeRange) {
+    if (this.withinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
+      this.withinMaxResizeRange = this.columnResizerService.isWithinMaxResizeRange;
+      if (!this.withinMaxResizeRange) {
         this.renderer.addClass(resizeTrackerEl, 'exceeded-max');
       } else {
         this.renderer.removeClass(resizeTrackerEl, 'exceeded-max');
--- a/src/clr-angular/data/datagrid/datagrid-column.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
--- a/src/clr-angular/data/datagrid/datagrid-column.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.ts
@@ -184,11 +184,17 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
 
   private listenForReorderAnimRequest() {
     return this.columnReorderService.reorderAnimRequested.subscribe(reorderAnimRequest => {
-      if (this._order === reorderAnimRequest.targetOrder) {
+      if (this._order === reorderAnimRequest.dropColumnOrder) {
         this.triggerDropAnimation(reorderAnimRequest);
-      } else if (reorderAnimRequest.sourceOrder <= this._order && this._order < reorderAnimRequest.targetOrder) {
+      } else if (
+        reorderAnimRequest.dragColumnOrder <= this._order &&
+        this._order < reorderAnimRequest.dropColumnOrder
+      ) {
         this.triggerShiftAnimation(reorderAnimRequest, 1);
-      } else if (reorderAnimRequest.sourceOrder >= this._order && this._order > reorderAnimRequest.targetOrder) {
+      } else if (
+        reorderAnimRequest.dragColumnOrder >= this._order &&
+        this._order > reorderAnimRequest.dropColumnOrder
+      ) {
         this.triggerShiftAnimation(reorderAnimRequest, -1);
       }
     });
--- a/src/clr-angular/data/datagrid/datagrid-column.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column.ts
@@ -3,7 +3,10 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+import { animate, style, transition, trigger } from '@angular/animations';
 import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ContentChild,
   EventEmitter,
@@ -13,34 +16,46 @@ import {
   OnInit,
   Output,
   ViewContainerRef,
-  ChangeDetectionStrategy,
-  ChangeDetectorRef,
+  ViewRef,
 } from '@angular/core';
 import { Subscription } from 'rxjs';
-
+import { DomAdapter } from 'src/clr-angular/utils/dom-adapter/dom-adapter';
+import { ClrDragEvent } from '../../utils/drag-and-drop/drag-event';
 import { HostWrapper } from '../../utils/host-wrapping/host-wrapper';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
+import { ClrPopoverEventsService } from '../../utils/popover/providers/popover-events.service';
+import { ClrPopoverPositionService } from '../../utils/popover/providers/popover-position.service';
+import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
 import { DatagridPropertyComparator } from './built-in/comparators/datagrid-property-comparator';
-import { DatagridPropertyStringFilter } from './built-in/filters/datagrid-property-string-filter';
-import { DatagridPropertyNumericFilter } from './built-in/filters/datagrid-property-numeric-filter';
-import { DatagridStringFilterImpl } from './built-in/filters/datagrid-string-filter-impl';
 import { DatagridNumericFilterImpl } from './built-in/filters/datagrid-numeric-filter-impl';
+import { DatagridPropertyNumericFilter } from './built-in/filters/datagrid-property-numeric-filter';
+import { DatagridPropertyStringFilter } from './built-in/filters/datagrid-property-string-filter';
+import { DatagridStringFilterImpl } from './built-in/filters/datagrid-string-filter-impl';
 import { ClrDatagridSortOrder } from './enums/sort-order.enum';
 import { ClrDatagridComparatorInterface } from './interfaces/comparator.interface';
+import { ClrDatagridFilterInterface } from './interfaces/filter.interface';
+import { ColumnReorderService, ReorderAnimRequest } from './providers/column-reorder.service';
 import { CustomFilter } from './providers/custom-filter';
 import { FiltersProvider } from './providers/filters';
 import { Sort } from './providers/sort';
+import { ViewAccessor } from './providers/view-manager.service';
 import { DatagridFilterRegistrar } from './utils/datagrid-filter-registrar';
-import { ClrDatagridFilterInterface } from './interfaces/filter.interface';
 import { WrappedColumn } from './wrapped-column';
-import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
-import { ClrPopoverPositionService } from '../../utils/popover/providers/popover-position.service';
-import { ClrPopoverEventsService } from '../../utils/popover/providers/popover-events.service';
-import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
-import { DetailService } from './providers/detail.service';
+import { DROP_ANIM_STATE, SHIFT_ANIM_STATE } from './enums/column-reorder-animation.enum';
 
 @Component({
   selector: 'clr-dg-column',
   template: `
+    <div class="datagrid-column-wrapper"
+         [clrDraggable]="_view"
+         [clrGroup]="columnsGroupId"
+         (clrDragStart)="inDragMode = true;"
+         (clrDragMove)="_markForCheck = true;"
+         (clrDragEnd)="inDragMode = false;"
+         (@reorderDropAnimation.done)="resetDropAnimation()"
+         [@reorderDropAnimation]="dropAnimationTrigger"
+         (@reorderShiftAnimation.done)="resetShiftAnimation()"
+         [@reorderShiftAnimation]="shiftAnimationTrigger">
       <div class="datagrid-column-flex">
           <!-- I'm really not happy with that select since it's not very scalable -->
           <ng-content select="clr-dg-filter, clr-dg-string-filter, clr-dg-numeric-filter"></ng-content>
@@ -75,34 +90,64 @@ import { DetailService } from './providers/detail.service';
           <span class="datagrid-column-title" *ngIf="!sortable">
               <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
           </span>
-
-          <clr-dg-column-separator *ngIf="showSeparator"></clr-dg-column-separator>
+      </div>
+      </div>
+      <clr-dg-column-separator></clr-dg-column-separator>
+      <div class="datagrid-column-droppable" 
+      clrDroppable 
+      [clrGroup]="columnsGroupId"
+      (clrDragStart)="_markForCheck = true; closeFilterWhenDragStart()"
+      (clrDrop)="requestReorder($event);">
+        <div class="datagrid-column-drop-line"></div>
       </div>
     `,
   providers: [ClrPopoverPositionService, ClrPopoverEventsService, ClrPopoverToggleService],
   host: {
     '[class.datagrid-column]': 'true',
+    '[class.datagrid-column-drag-mode]': 'inDragMode',
+    '[class.datagrid-column-reorder-drop]': 'inReorderDrop',
+    '[class.datagrid-column-reorder-shift]': 'inReorderShift',
     '[attr.aria-sort]': 'ariaSort',
     role: 'columnheader',
   },
+  animations: [
+    trigger('reorderDropAnimation', [
+      transition(`${DROP_ANIM_STATE.INACTIVE} => ${DROP_ANIM_STATE.ACTIVE}`, [
+        style({
+          position: 'fixed',
+          width: '{{width}}px',
+          height: '{{height}}px',
+          top: '{{fromTop}}px',
+          left: '{{fromLeft}}px',
+        }),
+        animate('0.2s 0.1s ease-in-out', style({ top: '{{toTop}}px', left: '{{toLeft}}px' })),
+      ]),
+    ]),
+    trigger('reorderShiftAnimation', [
+      transition(`${SHIFT_ANIM_STATE.INACTIVE} => ${SHIFT_ANIM_STATE.ACTIVE}`, [
+        style({ position: 'relative', left: '{{fromLeft}}px' }),
+        animate('0.2s ease-in-out', style({ left: '0px' })),
+      ]),
+    ]),
+  ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>>
-  implements OnDestroy, OnInit {
+  implements ViewAccessor, OnDestroy, OnInit {
   constructor(
     private _sort: Sort<T>,
     filters: FiltersProvider<T>,
     private vcr: ViewContainerRef,
-    private detailService: DetailService,
     private changeDetectorRef: ChangeDetectorRef,
-    public commonStrings: ClrCommonStringsService
+    public commonStrings: ClrCommonStringsService,
+    private columnReorderService: ColumnReorderService,
+    private filterPopoverToggleService: ClrPopoverToggleService,
+    private domAdapter: DomAdapter
   ) {
     super(filters);
     this.subscriptions.push(this.listenForSortingChanges());
-    this.subscriptions.push(this.listenForDetailPaneChanges());
+    this.subscriptions.push(this.listenForReorderAnimRequest());
   }
-
-  public showSeparator = true;
 
   /**
    * Subscription to the sort service changes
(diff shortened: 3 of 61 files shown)


src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  private _reorderAnimRequested: Subject<ReorderAnimRequest> = new Subject<ReorderAnimRequest>();
+
+  containerRef: ViewContainerRef;
+
+  get columnsGroupId(): string {
+    return this._columnsGroupId;
+  }
+
+  get ordersChange(): Observable<boolean> {
+    return this._ordersChange.asObservable();
```
hippee-lee: Should there be a test to make sure this emits at the end of `updateOrders`?

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private updateColumnStatesWithNewOrders() {
+    this.orders.forEach((order, index) => {
+      this.columnsService.emitStateChangeAt(index, { changes: [DatagridColumnChanges.ORDER], order: order });
+    });
+
+    // The first and last visible columns might
+    // have been changed after orders were changed
+    this.haveFirstAndLastVisibleColumnsChanged();
```
hippee-lee: It seems simpler to me to just put these calls directly in here rather than wrapping them in another private method so that its obvious this function triggers anyone subscribed. 
My preference is the first since its clearer to me that these private subscriptions need to be fired from this method.

```
this._isFirstVisibleChanged.next();
this._isLastVisibleChanged.next();
```
or if its better to call the method the way the renderer and hidable columns do:
```
this.requestFirstVisibleChangeCheck();
this.requestLastVisibleChangeCheck();
```

src/clr-angular/data/datagrid/providers/column-reorder.service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+    it('provides the view container', function() {
+      expect(columnReorderService.containerRef).not.toBeUndefined();
+    });
+
+    it('emits reorder request data when reorderViews method gets called', function() {
+      expect(testComponent.reorderRequest).toBeUndefined();
+      columnReorderService.reorderViews(testComponent.view1, testComponent.view2, null);
+      expect(testComponent.reorderRequest).not.toBeUndefined();
```
hippee-lee: Should this also test that the correct ReorderRequest object was emitted?

src/clr-angular/data/datagrid/providers/column-reorder.service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    it('requests first and last visible columns checks after updating orders', function() {
+      expect(testComponent.firstVisibleChecked).toBeFalsy();
+      expect(testComponent.lastVisibleChecked).toBeFalsy();
+      columnReorderService.updateOrders([1, 0, 2]);
+      expect(testComponent.firstVisibleChecked).toBeTruthy();
+      expect(testComponent.lastVisibleChecked).toBeTruthy();
+    });
+
+    it('returns new order at index', function() {
```
hippee-lee: nit: should this be phrased as `orderAt` ?

src/clr-angular/data/datagrid/providers/column-reorder.service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+      columnReorderService.orders = [1, 0, 2];
+      expect(columnReorderService.orderAt(0)).toBe(1);
+      expect(columnReorderService.orderAt(1)).toBe(0);
+      expect(columnReorderService.orderAt(2)).toBe(2);
+      expect(columnReorderService.orderAt(5)).toBe(-1);
+    });
+
+    describe('should emit proper reorder request data on moving:', () => {
```
hippee-lee: nit: can we keep to either arrow functions or a standard function declarations?

some describes/its use:
```typescript
describe('my suite', function() {...})
```
other describes/its use:
```typescript
describe('should emit proper reorder request data on moving:', () => {...}
```

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    return this._reorderRequested.asObservable();
+  }
+
+  get reorderAnimRequested(): Observable<ReorderAnimRequest> {
+    return this._reorderAnimRequested.asObservable();
+  }
+
+  // The following method is called by column ClrDatagridColumn when one column is dropped on another.
+  reorderViews(sourceView: ViewRef, targetView: ViewRef, dropEvent: ClrDragEvent<ViewRef>): void {
+    const sourceOrder = this.containerRef.indexOf(sourceView);
```
hippee-lee: Question: what do you think about changing the variable names to something more descriptive? e.g - dragColumn and dropColumn? I know under the hood its manipulating views but reading it as sourceOrder/targetOrder its less clear to me whats happening here.

hippee-lee:
I'm going to submit the review in stages. This covers some general comments for the style changes and the column-reorder.service.

src/clr-angular/data/datagrid/utils/view-manager-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+// ViewManagerUtils is a utility class that contains methods for managing views.
+@Injectable()
+export class ViewManagerUtils {
+  public static detachAllViews(containerRef: ViewContainerRef): void {
+    for (let i = containerRef.length; i > 0; i--) {
+      containerRef.detach();
+    }
+  }
+
+  public static reorderViews(containerRef: ViewContainerRef, viewAccessors: ViewAccessor[]) {
```
hippee-lee: I see this called in two places, row-detail and datagrid-row. However rowdetail calls it in `ngContentInit` and datagrid-row calls after `ngAfterViewInit`. based on the comment near usage, shouldnt they both be called in `ngAfterViewInit`?

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -241,59 +236,85 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+        this.renderer.removeClass(this.el.nativeElement, CALCULATE_MODE_CLASS);
+        this.displayViewsInMode(this._projectedDisplayColumns, this._displayedRows);
+      } else {
+        this.renderer.addClass(this.el.nativeElement, CALCULATE_MODE_CLASS);
+        this.displayViewsInMode(this._projectedCalculationColumns, this._calculationRows);
+      }
+    });
+  }
+
+  private displayViewsInMode(columnsViewContainer: ViewContainerRef, rowsViewContainer: ViewContainerRef) {
```
hippee-lee: nit: naming is a bit confusing to me. 
suggestions: insertViewsForMode, inserViewsForDisplay, insertViewsForDisplayMode. 

Not a blocker tho.

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -241,59 +236,85 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+    return this.selection.change.subscribe(s => {
+      if (this.selection.selectionType === SelectionType.Single) {
+        this.singleSelectedChanged.emit(<T>s);
+      } else if (this.selection.selectionType === SelectionType.Multi) {
+        this.selectedChanged.emit(<T[]>s);
+      }
+    });
+  }
+
+  private displayRowViewsOnRowsChange(): Subscription {
```
hippee-lee: nit: naming - it doesn't seem to describe what it is doing.
suggestions: insertRowViewsOnRowChange || insertRowViewsAfterRowChange

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -260,4 +242,58 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
   public get _view() {
     return this.wrappedInjector.get(WrappedRow, this.vcr).rowView;
   }
+
+  private resetViewsOnDisplayModeChange(): Subscription {
```
hippee-lee: nit: naming seems to reflect resetting back to some base state. it seems like its just inserting views after display mode changes.
suggestions: insertViewsAfterDisplayModeChange, insertViewsAfterModeChange

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -49,14 +64,57 @@ export class ClrDatagridRowDetail<T = any> implements AfterContentInit, OnDestro
...
+    return this.expand.replace.subscribe(replaceChange => {
+      this.replacedRow = replaceChange;
+    });
   }
 
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
   }
+
+  private updateViewsOnOrdersChange(): Subscription {
```
hippee-lee: nit: naming
suggestion: updateViewsAfterOrderChange

src/clr-angular/data/datagrid/utils/view-manager-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+  ): void {
+    if (prioritized) {
+      this.prioritizeByOrder(viewAccessors).forEach(viewAccessor => containerRef.insert(viewAccessor._view));
+    } else {
+      viewAccessors.forEach(viewAccessor => containerRef.insert(viewAccessor._view));
+    }
+  }
+
+  private static prioritizeByOrder(viewAccessors: ViewAccessor[]): ViewAccessor[] {
+    // As an example, this method would turn the following array:
```
hippee-lee: Thank you for the comments. very helpful.

src/clr-angular/data/datagrid/utils/view-manager-utils.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      testComponent = fixture.componentInstance;
+    });
+
+    afterEach(() => {
+      fixture.destroy();
+    });
+
+    it('inserts all views', function() {
+      expect(testComponent.vcr.length).toBe(0);
+      // Using Array.prototype.slice() to not modify the original array
```
hippee-lee: this is why toEqual vs toBe, right?

src/clr-angular/data/datagrid/datagrid.ts
```diff
@@ -241,59 +236,85 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
   }
 
+  private focusDgTableOnPageChange() {
+    return this.page.change.subscribe(() => {
+      this.datagridTable.nativeElement.focus();
+    });
+  }
+
+  private emitOnStateProviderChange(): Subscription {
+    return this.stateProvider.change.subscribe(state => this.refresh.emit(state));
```
bdryanovski: nit: when the function argument is matching the subscribe we could skip the function call
```ts
return this.stateProvider.change.subscribe(this.refresh.emit);
```

src/clr-angular/utils/drag-and-drop/all.spec.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
```
bdryanovski: nit: we could skip this change :)

src/clr-angular/data/datagrid/utils/view-manager-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+    // [ {view: _view, order: 1}, {view: _view, order: 2}, {view: _view, order: 0} ] into
+    // [ {view: _view, order: 0}, {view: _view, order: 1}, {view: _view, order: 2} ].
+    return viewAccessors
+      .sort((viewAccessor1: ViewAccessor, viewAccessor2: ViewAccessor) => {
+        // The most straightforward case where both columns have different order values
+        // so we can directly sort them
+        if (viewAccessor1.order > viewAccessor2.order) {
```
bdryanovski: nit:

```ts
if (viewAccessor1.order > viewAccessor2.order) {
  return 1;
} 
if (viewAccessor1.order < viewAccessor2.order) {
  return -1;
}
```
no need of the `else`

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -118,6 +124,14 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
     this.rows.forEach(row => row.setColumnState());
   }
 
+  private findFirstVisible() {
+    this.headers.forEach(header => header.isFirstVisible());
```
bdryanovski: nit: for both function thing that `Array.map` is a better option.

src/clr-angular/data/datagrid/render/main-renderer.ts
```diff
@@ -92,6 +92,12 @@ export class DatagridMainRenderer<T = any> implements AfterContentInit, AfterVie
         this.setupColumns();
         this.columnsSizesStable = false;
         this.stabilizeColumns();
+      }),
+      this.columnsService.checkFirstVisible.subscribe(() => {
```
bdryanovski: nit: 
```ts
this.columnsService.checkFirstVisible.subscribe(this.findFirstVisible);
this.columnsService.checkLastVisible.subscribe(this.findLastVisible);
```

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -49,14 +64,57 @@ export class ClrDatagridRowDetail<T = any> implements AfterContentInit, OnDestro
...
+        return;
+      }
+
+      if (this.columnReorderService.orders) {
+        if (this._detailCells.length === 0) {
+          // we just need to insert the views initially when the view container is empty.
+          ViewManagerUtils.insertAllViews(this._detailCells, this.assignRawOrders(), true);
+        } else {
+          this.cells.forEach((cell, index) => {
+            if (this._detailCells.indexOf(cell._view) === -1) {
```
bdryanovski: nit: `Array.includes` is an option here

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -22,7 +22,7 @@
     '.datagrid-header',
     '.datagrid-body',
     '.datagrid-row',
-    '.datagrid-column',
+    '.datagrid-column-wrapper',
```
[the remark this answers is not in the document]
Shijir: yeah we needed to a wrapper for columns so that we can place clrDraggable directive on.

src/clr-angular/data/datagrid/_datagrid.clarity.scss
```diff
@@ -865,19 +876,19 @@
...
 
-        &::after {
+        &::before {
           content: '';
           position: absolute;
-          height: calc(100% + #{$clr-datagrid-column-separator-expandby} - #{$clr-global-borderwidth});
+          height: calc(100% - #{$clr-datagrid-column-separator-expandby});
           width: $clr_baselineRem_1px;
-          top: calc(-0.5 * #{$clr-datagrid-column-separator-expandby} + #{$clr_baselineRem_1px});
+          top: $clr-datagrid-column-separator-expandby * 0.5;
```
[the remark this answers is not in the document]
Shijir: Since we have a new wrapper that now contains the resize handle, we needed to calculate some of these values differently.

src/clr-angular/utils/_mixins.clarity.scss
```diff
@@ -116,4 +116,26 @@ $clr-outline-spread: 2px;
   @if $str != null {
     @return #{'--' + $str};
   }
-}
\ No newline at end of file
+}
+@mixin clr-datagrid-column-ghost() {
+  position: relative;
+  background-color: $clr-thead-bgcolor;
+  box-shadow: 0 3px 4px rgba(116, 116, 116, 0.3);
```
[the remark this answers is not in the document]
Shijir: Good question. I need to figure out that question.

src/clr-angular/data/datagrid/interfaces/view-accessor.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { EmbeddedViewRef } from '@angular/core';
+
+// This is the interface that will be implemented by any component that accesses its view (EmbeddedViewRef).
+// Those components are currently ClrDatagridColumn, ClrDatagridCell, and ClrDatagridRow.
+export interface ViewAccessor {
+  _view: EmbeddedViewRef<void>;
```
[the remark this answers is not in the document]
Shijir: This is a case where we would want to keep the _view public as well as underscore prefix as we access it from outside. And at the same time, we should also let users know that it's for only internal uses and denote that this is not a public API.

src/clr-angular/data/datagrid/utils/view-manager-utils.ts
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Injectable, ViewContainerRef } from '@angular/core';
+import { ViewAccessor } from '../interfaces/view-accessor.interface';
+
+// ViewManagerUtils is a utility class that contains methods for managing views.
+@Injectable()
```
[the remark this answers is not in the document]
Shijir: Good catch. It was a service before, but I have changed it to just a utility class.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir: I don't think that the test will add any value in this case because we are testing it indirectly. If actually we don't emit updated orders at the end, almost all current tests will fail.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir: I agree. I prefer the latter choice because we could keep the subjects private in the service.

src/clr-angular/data/datagrid/providers/column-reorder.service.spec.ts
Shijir: I think it should stay as `order at` because it's a new allocated order at that particular index.

src/clr-angular/data/datagrid/providers/column-reorder.service.spec.ts
Shijir: definitely

src/clr-angular/data/datagrid/utils/view-manager-utils.ts
Shijir: Actually, we shouldn't wait unnecessarily until ngAfterViewInit in both cases.  The reason is that cell views would be already accessible in the ngAfterContentInit.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir: I have done following changes: 

sourceView -> dragColumnView
sourceOrder -> dragColumnOrder

targetView -> dropColumnView
targetOrder -> dropColumnOrder

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -260,4 +242,58 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
+      }
+    });
+  }
+
+  private resetViewsOnCellsChange(): Subscription {
+    // This method is to handle a case when cells change without headers changing.
+    // For example, one cell could be replaced by another in place.
+    return this.cells.changes.subscribe(() => {
+      this.cells.forEach((cell, index) => {
+        if (this._scrollableCells.indexOf(cell._view) === -1) {
```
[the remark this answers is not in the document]
Shijir: `this._scrollableCells` is not array. It's `ViewContainerRef`

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -260,4 +242,58 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
+      // we have to make sure there are views inserted in the view container
+      // before calling reorderViews method from viewManager service
+      if (byReordering && this._scrollableCells.length > 0) {
+        ViewManagerUtils.reorderViews(this._scrollableCells, this.assignRawOrders());
+        this.updateCellOrder();
+      }
+    });
+  }
+
+  private updateCellOrder(): void {
```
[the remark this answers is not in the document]
Shijir: Side effects of what? `this.cells` is not an array, it's `QueryList`. Once you map it it turns into a new array so you cannot assign an array back to QueryList. So we shouldn't do that.

src/clr-angular/data/datagrid/render/main-renderer.ts
Shijir: Why is map better? I am not trying to get a new array out of these headers. I just need to iterate through and *mutate* the items inside.

src/clr-angular/data/datagrid/datagrid-row-detail.ts
Shijir: Same as above, `_detailCells` is not array, but ViewContainerRef.

src/clr-angular/data/datagrid/render/main-renderer.ts
Shijir: We shouldn't do that. By referencing a function without invoking it, we would lose the correct context of this. That's one of the common red flags: https://github.com/microsoft/TypeScript/wiki/'this'-in-TypeScript

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -59,12 +59,11 @@ export class ClrDatagridColumnSeparator {
...
 
+  private withinMaxResizeRange: boolean;
   private redFlagTracker(resizeTrackerEl: HTMLElement) {
-    let isWithinMaxResizeRange: boolean;
-    // @TODO(JEREMY) Review this, it will always be true because above is always null
-    if (isWithinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
-      isWithinMaxResizeRange = this.columnResizerService.isWithinMaxResizeRange;
-      if (!isWithinMaxResizeRange) {
+    if (this.withinMaxResizeRange !== this.columnResizerService.isWithinMaxResizeRange) {
+      this.withinMaxResizeRange = this.columnResizerService.isWithinMaxResizeRange;
```
hippee-lee: Q: Could we eliminate the private `withinMaxResizeRange` and just tuse the value on the resize service directly?

Jinnie:
nit2: While dragging when some popovers are open, the popovers don't get closed until mouse-up. I guess this may be a popover issue, but it is surfacing now. Additionally, the dragged ghost gets obscured by the popover. We may have cases when popovers will not be closed by outside click, so this is valid case. I think the ghost should have a higher z-index than all else on screen, as it is the focus of the active user interaction.
Screen Shot 2019-12-12 at 7 35 51 PM

Jinnie:
nit: It would be nice if we could support the Horizontal scroll case:
Screen Shot 2019-12-12 at 7 43 08 PM
There are some invisible and unreachable columns on the right here. It's not a disaster, as it can be done by drag, scroll, drag again. And I guess the dev cost for supporting it with a single drag may be pretty high.

Jinnie:
nit?: Large data. If we move a very large column, next to shorter columns, the active drop position is very narrow. I think we're evaluating based on the middle of the ghost. This can lead to long drag before reorder is active, followed by a very short drag space before the indicator goes away. It may be better if we evaluate based on the border closer to the movement direction? Just play a bit with this case in the Smart Column Sizing datagrid sub-page on the dev-app
Screen Shot 2019-12-12 at 7 51 03 PM

Also I think that the indicator shouldn't go away when you drag past the last column. With large column dragged, at least, it feels awkward.

Jinnie:
And just a funny case that I suggest to **not** try to fix:
- hide several columns
- start dragging
- tab to the show/hide columns menu (still keep the mouse pressed) and Show All
- you can't drop on the newly exposed columns

🤣

Jinnie:
Nit: reordering column with open signpost leaves the popover hanging over the old position. Note - this will be fixed when signposts start using smart popovers.
Screen Shot 2019-12-12 at 8 15 53 PM

hippee-lee:
> nit2: While dragging when some popovers are open, the popovers don't get closed until mouse-up. I guess this may be a popover issue, but it is surfacing now. Additionally, the dragged ghost gets obscured by the popover. We may have cases when popovers will not be closed by outside click, so this is valid case. I think the ghost should have a higher z-index than all else on screen, as it is the focus of the active user interaction.
> Screen Shot 2019-12-12 at 7 35 51 PM

I think we should modify the popover code to close before dragging starts. This eliminates most of these issues and also afaik - datagrid popovers should be closed when user clicks outside of the popover.

Shijir:
@Jinnie thanks for pointing these issues.

I am not really convinced that we should close these action popovers if users opened them in the datagrid's body area. Closing popovers in the header on dragStart makes sense as that extra popover throws off droppable area calculation. Also, from UX standpoint, dragging a column while its filter popover was hanging open wasn't a good experience. But as far as popovers in the datagrid body, I see them as independent entities from column headers and don't interfere with each other.

As for the horizontal scrolling, we will ship it without horizontal scrolling for now and track that as a known issue. The implementation of that feature will go into the DnD utility library.

For long columns with large data, we cannot evaluate based on the border because given that the ghost is very large and it could cover multiple droppable columns leading them to be activated at the same time. In column dragging, only one droppable column should be activated at a time.

Jinnie:
Hi Shijir, thanks for quick reply:
- Popovers get closed anyway, but on mouse-up. Which is after drag/drop. Maybe we should discuss with @hippee-lee if this event needs to be changed.
- Nevertheless, I think the z-index of the dragged ghost should still be higher than popovers.
- The in-body popovers issue is rather small, IMO. 1 - it shall be resolved when sighposts start using smart popovers, and 2 - it's not really column-drag-pr issue, so it shouldn't stop this PR.
- Horizontal scrolling - I agree - it's a nit, and it has workaround. Just mentioning it for the record.
- We can discuss the long-data case a bit more. I can show what I saw and explain better what I meant. It also can be improved iteratively, without being a breaking change.

hippee-lee:
> * Popovers get closed anyway, but on mouse-up. Which is after drag/drop. Maybe we should discuss with @hippee-lee if this event needs to be changed.
> * Nevertheless, I think the z-index of the dragged ghost should still be higher than popovers.


Yes, I think the popovers and drag-n-drop events should be the same. I would expect the popover to close here before dragging starts. I agree w/ @jinnie the z-index still should be higher for the use case when the popover component does not close on click.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
Shijir: We shouldn't eliminate that property. If we eliminate that, we will try to set the same class to the DOM element on every drag, and drag event fires a lot. So it's there to prevent adding the same class if the DOM element has that class already.

Shijir:
@Jinnie @hippee-lee we actually have the following z-index values in our `_layers.clarity.scss`:
```
smart-popover: 5000,
draggable-ghost: 2147483647// maximum possible value for z-index across all browsers
```
So the `draggable-ghost` has actually the max possible integer for its z-index (this is supposedly the maximum possible value for z-index across modern browsers). In our case, the reason why the popover still appears over the draggable ghost is that the draggable ghost is nested inside the header which has the z-index of 500 while the smart popover has the z-index of 5000. So no matter how big number we assign to the ghost's z-index, it still appears below the popover which has.  

A possible solution to this issue could be closing smart popovers on outside `mousedown` events so that it would actually close before drag starts.

hippee-lee:
> A possible solution to this issue could be closing smart popovers on outside `mousedown` events so that it would actually close before drag starts.

Agreed. For the datagrid, I think any popover should be closed before dragging starts. This allows us to avoid the issue of dragging columns underneath popover content.

Jinnie:
This totally makes sense.
Thanks for the clarification about the header z-index. I had not yet come across this z-index nesting issue, it's an interesting case, good to know.
If we couldn't close the popovers, then we might need to try temporarily moving the ghost to the body, but for the time being it's an overkill.

src/clr-angular/data/datagrid/render/main-renderer.ts
bdryanovski: ignore it .. sorry for that ..

src/clr-angular/data/datagrid/render/header-renderer.ts
```diff
@@ -130,4 +135,28 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
+      this.setFlexibleWhenAllStrict();
+    } else {
+      this.renderer.removeClass(this.el.nativeElement, LAST_VISIBLE_COLUMN_CLASS);
+    }
+  }
+
+  private setFlexibleWhenAllStrict() {
+    const flexibleWidths = this.columnsService.columnStates.filter(state => !state.hidden && state.strictWidth === 0);
+    if (flexibleWidths.length === 0) {
+      this.columnsService.emitStateChange(this.columnState, { changes: [DatagridColumnChanges.WIDTH], strictWidth: 0 });
```
gnomeontherun: Test missing to check if this gets called.

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -41,6 +41,37 @@ export class ColumnsService {
     return this.columnStates.filter(state => state.hideable).length > 0;
   }
 
+  get orderOfLastVisible(): number {
+    return Math.max(...this.columnStates.filter(state => !state.hidden).map(state => state.order));
```
gnomeontherun: I like that you kept these calls separate for first/last just in case we have to diverge more in later work.

src/clr-angular/data/datagrid/providers/columns.service.ts
```diff
@@ -41,6 +41,37 @@ export class ColumnsService {
...
+  get orderOfFirstVisible(): number {
+    return Math.min(...this.columnStates.filter(state => !state.hidden).map(state => state.order));
+  }
+
+  // First and last visible columns may change in 3 different cases:
+  // -  When columns are reordered
+  // -  When the current first/last gets hidden state
+  // -  When the column at first/last gets removed or a new column added at first/last index
+  private _isFirstVisibleChanged: Subject<void> = new Subject<void>();
+  private _isLastVisibleChanged: Subject<void> = new Subject<void>();
```
gnomeontherun: nit: I would rather the first/last properties were kept together instead.

```
private _isFirstVisibleChanged: Subject<void> = new Subject<void>();
get checkFirstVisible(): Observable<void> {
    return this._isFirstVisibleChanged.asObservable();
}
isFirstVisibleChanged() {
    this._isFirstVisibleChanged.next();
}
// Then last visible
```

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,109 @@
...
+export type ReorderAnimRequest = ReorderRequest & ClrDragEvent<ViewRef>;
+
+@Injectable()
+export class ColumnReorderService {
+  // the common group id that will be shared across
+  // Datagrids all reorder draggable and droppables
+  private _columnsGroupId: string;
+
+  constructor(private columnsService: ColumnsService) {
+    this._columnsGroupId = 'dg-column-group-' + nbColumnsGroup++;
```
gnomeontherun: Can't we store this on the ColumnsService instead of here? It might be useful in other places to have knowledge of a unique ID for a datagrid.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  orders: number[];
+
+  private _ordersChange: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
```
gnomeontherun: nit: again would prefer to have the related properties declared together

src/clr-angular/data/datagrid/datagrid.ts
gnomeontherun: `reattachRows`?

src/clr-angular/data/datagrid/datagrid-row-detail.ts
```diff
@@ -49,14 +64,57 @@ export class ClrDatagridRowDetail<T = any> implements AfterContentInit, OnDestro
...
+        this.updateCellOrder();
+      }
+    });
+  }
+
+  private updateCellOrder(): void {
+    this.cells.forEach(cell => (cell.order = this._detailCells.indexOf(cell._view)));
+  }
+
+  private assignRawOrders(): ClrDatagridCell[] {
```
gnomeontherun: Some of this is duplicated, can we share is somehow?

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -75,34 +90,67 @@ import { DetailService } from './providers/detail.service';
           <span class="datagrid-column-title" *ngIf="!sortable">
               <ng-container *ngTemplateOutlet="columnTitle"></ng-container>
           </span>
-
-          <clr-dg-column-separator *ngIf="showSeparator"></clr-dg-column-separator>
+      </div>
+      </div>
+      <clr-dg-column-separator></clr-dg-column-separator>
+      <div class="datagrid-column-droppable" 
+      clrDroppable
```
gnomeontherun: nit: formatting around here is off

src/clr-angular/data/datagrid/datagrid-column.ts
```diff
@@ -75,34 +90,67 @@ import { DetailService } from './providers/detail.service';
...
   ) {
     super(filters);
     this.subscriptions.push(this.listenForSortingChanges());
-    this.subscriptions.push(this.listenForDetailPaneChanges());
+    this.subscriptions.push(this.listenForReorderAnimRequest());
   }
 
-  public showSeparator = true;
+  // TODO: remove once OnPush change detection strategy is removed
+  _markForCheck = false;
```
gnomeontherun: I know some things were removed, is it still a problem?

Shijir: This is no longer a problem. It will be removed

src/clr-angular/data/datagrid/datagrid-row-detail.ts
Shijir: Yes, this method is in ClrDatagrid, ClrDatagridRow, ClrDatagridRowDetail. Only ClrDatagridRow and ClrDatagridRowDetail share the exact same implementation. So the implementation is not exactly the same across these classes. So that's also why I didn't turn it into a utility method or something that's shared.

src/clr-angular/data/datagrid/datagrid-column.ts
Shijir: It should be removed when I rebase the branch as we have already got rid of OnPush in master.

src/clr-angular/data/datagrid/providers/columns.service.ts
Shijir: Isn't it easier to look up properties when the same types are kept together? For me, at least, it's easier and faster to find properties when the same types are kept together.

src/clr-angular/data/datagrid/providers/column-reorder.service.ts
Shijir: I think it's better to keep this id here. There might be other Datagrid components that might make use of DnD in the future (there is actually already one which column separators). So I would like to keep this id here as it's specifically used to group reorderable parts of the datagrid columns.

gnomeontherun:
This is no longer blocked by v9 update work.

TekSiDoT:
5 months have passed since this PR is no longer blocked by Angular 9, could we get an update the expected work remaining?

gnomeontherun:
We are blocked by Angular still, @Shijir has a PR open to resolve the last outstanding issue. https://github.com/angular/angular/pull/37167

mathisscott:
@Shijir : Do we know the status on the angular fix?

gnomeontherun:
@Shijir do we want to just close this PR and let you create a new PR when its on the latest master? This is pretty stale and can help us clean up our backlog for the moment.

Shijir:
This PR has been idle and blocked by https://github.com/angular/angular/issues/37120 for a while now. Until this underlying issue is fixed, I am closing this PR now. In the meantime, we will also explore other implementational methods that could work with the current Angular state.

aavelyn:
Hi https://github.com/angular/angular/issues/37120  the issue seems to have been closed at angular. So could this PR be reopened/reexamined? :)

hippee-lee:
@aavelyn - the datagrid (and Angular components in general) is moving into a bugfix / maintenance mode. You can read more [here](https://clarity.design/get-started/support/) for the support strategy and [here](https://clarity.design/get-started/roadmap/) for the current roadmap. 

What this means is that we won't be investing development into new features for the Angular components. Requested features are planned for the Core web component version as they are developed. Due to the differences between Angualrt and web components it also means that some features (especially features that require knowing and managing the `state` of a component require the application to handle. I believe that reorder columns will be one of those.  

So at this time we do not have plans to reopen this issue.

brannislav:
@hippee-lee can you post here a link to the column reorder request for the Core web component?

hippee-lee:
AFAIK, there is no request in github. But, the core grid use [HTML5 Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) for both rows and columns. Consumers will have access to the emitted events when users make changes. 

The grid is getting close to PR status but until it is release I don't want to write to much about the API or create a demo in case that changes. Keep an eye out for the PR in the next few weeks.

PR #4044, opened 2019-11-23
feat(badge): core badge, feat(tag): core tag

jeeyun:

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/docs/core/cwc-tag.md
+++ b/docs/core/cwc-tag.md
@@ -1,0 +1,17 @@
+# cwc-tag
+
+Web component tags.
+
+## Attributes
+
+| Attribute | Type     | Description                                      |
+|-----------|----------|--------------------------------------------------|
+| `alias`   | `string` | Sets the color of the tag (and badge if present) from a predefined list of aliases <br/> (`1`, `2`, `3`, `4`, `5`) |
+| `color`   | `string` | Sets the color of the tag (and badge if present) from a predefined list of choices <br/> (`gray`, `purple`, `blue`, `orange`, `light-blue`) |
+| `status`  | `string` | Sets the color of the tag (and badge if present) from a predefined list of statuses <br/> (`info`, `success`, `warning`, `danger`) |
+
+## Slots
+
+| Name      | Description                     |
+|-----------|---------------------------------|
+| `default` | Content slot for inside the tag |
--- a/src/clr-core/badge/badge.element.ts
+++ b/src/clr-core/badge/badge.element.ts
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { registerElementSafely } from '@clr/core/common';
+import { html, LitElement } from 'lit-element';
+import { styles } from './badge.element.css';
+
+/**
+ * Web component badges.
+ *
+ * @noInheritDoc
+ * @element cwc-badge
+ * @slot default - Content slot for inside the badge
+ * @attr {String} color - Sets the color of the badge from a predefined list of choices <br/> (`gray`, `purple`, `blue`, `orange`, `light-blue`)
+ * @attr {String} alias - Sets the color of the badge from a predefined list of aliases <br/> (`1`, `2`, `3`, `4`, `5`)
+ * @attr {String} status - Sets the color of the badge from a predefined list of statuses <br/> (`info`, `success`, `warning`, `danger`)
+ */
+// @dynamic
+export class CwcBadge extends LitElement {
+  render() {
+    return html`<slot></slot>`;
+  }
+
+  static get styles() {
+    return styles;
+  }
+}
+
+registerElementSafely('cwc-badge', CwcBadge);
--- a/src/clr-core/badge/package.json
+++ b/src/clr-core/badge/package.json
@@ -1,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "index.ts"
+    }
+  }
+}
--- a/src/clr-core/tag/package.json
+++ b/src/clr-core/tag/package.json
@@ -1,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "index.ts"
+    }
+  }
+}
--- a/src/clr-core/tag/tag.element.spec.ts
+++ b/src/clr-core/tag/tag.element.spec.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { CwcTag } from '@clr/core/tag';
+import '@clr/core/tag';
+import { componentIsStable, createTestElement, removeTestElement, waitForComponent } from '@clr/core/test/utils';
+
+describe('tag element', () => {
+  let testElement: HTMLElement;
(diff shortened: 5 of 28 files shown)


src/clr-core/tag/tag.element.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    await waitForComponent('cwc-tag');
+    component = testElement.querySelector<CwcTag>('cwc-tag');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  it('should create the component', async () => {
```
mathisscott: Should we have any tests for the clickable tag? Or do the tests on the base button cover those situations?

src/dev-core/src/app/tag/tag.demo.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Component } from '@angular/core';
+import '@clr/core/badge';
+import '@clr/core/tag';
+
+@Component({
+  selector: 'app-badge-demo',
+  templateUrl: './tag.demo.html',
+})
+export class TagDemoComponent {
+  open = false;
```
mathisscott: What does the `open` property do?

src/clr-core/badge/badge.element.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Web component badges.
+ *
+ * @noInheritDoc
+ * @element cwc-badge
+ * @slot default - Content slot for inside the badge
+ * @attr {color} - Sets the color of the badge from a predefined list of choices <br/> (`gray`, `purple`, `blue`, `orange`, `light-blue`)
+ * @attr {alias} - Sets the color of the badge from a predefined list of aliases <br/> (`1`, `2`, `3`, `4`, `5`)
+ * @attr {status} - Sets the color of the badge from a predefined list of statuses <br/> (`info`, `success`, `warning`, `danger`)
```
coryrylan: The syntax here is not quite right, `@attr {String} color - Sets the color...` The value within `{ }` is the type of the attribute which is almost always String. This will regenerate the missing attr in the generated docs markdown files.

src/clr-core/badge/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "badge.element.ts"
```
coryrylan: Can we have this point to `index.ts`? Also in the package.json we need to add a new side effect entrypoint. `"./**/clr-core-badge.js"`

https://github.com/vmware/clarity/blob/master/src/clr-core/package.json#L11

src/clr-core/tag/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "tag.element.ts"
```
coryrylan: Can we have this point to `index.ts`? Also in the package.json we need to add a new side effect entrypoint. `"./**/clr-core-tag.js"`

https://github.com/vmware/clarity/blob/master/src/clr-core/package.json#L11

src/clr-core/tag/tag.element.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Web component tags.
+ *
+ * @noInheritDoc
+ * @element cwc-tag
+ * @slot default - Content slot for inside the tag
+ * @attr {color} - Sets the color of the tag (and badge if present) from a predefined list of choices <br/> (`gray`, `purple`, `blue`, `orange`, `light-blue`)
+ * @attr {alias} - Sets the color of the tag (and badge if present) from a predefined list of aliases <br/> (`1`, `2`, `3`, `4`, `5`)
+ * @attr {status} - Sets the color of the tag (and badge if present) from a predefined list of statuses <br/> (`info`, `success`, `warning`, `danger`)
```
coryrylan: Same as above with syntax of the attrs `@attr {String} alias - Sets the color`

src/dev-core/src/app/tag/tag.demo.html
```diff
@@ -0,0 +1,36 @@
...
+<cwc-tag readonly color="gray">Austin</cwc-tag>
+<cwc-tag readonly color="purple">New York</cwc-tag>
+<cwc-tag readonly color="blue">Palo Alto</cwc-tag>
+<cwc-tag readonly color="orange">San Francisco</cwc-tag>
+<cwc-tag readonly color="light-blue">Seattle</cwc-tag>
+<cwc-tag readonly alias="1">Austin</cwc-tag>
+<cwc-tag readonly alias="2">New York</cwc-tag>
+<cwc-tag readonly alias="3">Palo Alto</cwc-tag>
+<cwc-tag readonly alias="4">San Francisco</cwc-tag>
+<cwc-tag readonly alias="5">Seattle</cwc-tag>
```
coryrylan: This is more of an API design question, is there a reason to have two different APIs for color? Seems like we could simplify the Web Component API from the existing angular/ui API by providing just one and drop `alias`. Same with badges. I may be missing some historical context why alias existed in the previous APIs.

docs/core/cwc-tag.md
```diff
@@ -0,0 +1,9 @@
+# cwc-tag
```
coryrylan: Can you add two links to the tag and badge components to this file?

https://github.com/vmware/clarity/blob/master/docs/core/README.md

src/clr-core/tag/tag.element.spec.ts
jeeyun: I think Cory and I settled on having the button test cover all the base button functionalities for now. If there are specific behaviors unique to the component (that's not directly inherited from base button), I'm open to adding those.

src/dev-core/src/app/tag/tag.demo.ts
jeeyun: What is it good for? Absolutely nothing.

Removing. :)

docs/core/cwc-tag.md
jeeyun: Yup will do

src/dev-core/src/app/tag/tag.demo.html
jeeyun: It's a good question. I don't know the historical context for the alias or whether applications use this. If no good reason, I'm in agreement that we could remove this for core. Angular/UI might need to go through proper deprecation cycle for removal.. Maybe @mathisscott can shed some light on this?

src/clr-core/badge/badge.element.ts
jeeyun: Oh you are right. I'm not sure what happened there. Fixing.

src/clr-core/badge/package.json
jeeyun: Yes and will change the tag's `package.json` as well.

jeeyun:
> Not sure how but the API files in `src/dev-core/api` that are generated are not supposed to be source controlled and are listed in gitignore.
> 
> Could you try removing them from the PR and see if they get added again if you run the full build? If git wants to add them back again then there might be a bug with the gitignore rule or build.

I think it's because an earlier version had a different path for the api docs (`src/dev-core/src/api`) and I had locally generated them and they didn't follow the current path pattern for gitignore (`src/dev-core/api`) so didn't get filtered out. Committing changes to delete them. I think it should be fine going forward.

src/dev-core/src/app/tag/tag.demo.html
coryrylan: I would vote to drop the `alias` api for core but not going to block the PR on it. We can make sure to have a decision before 3.0

PR #4058, opened 2019-11-27
fix(button): core button a11y fixes

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

## What is the new behavior?
- trigger form submits, render hidden button in light DOM
- enable anchor/link support
- emulate native keyboard behavior
- add demos to core dev app using native HTML form and anchor

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/common/base/button.base.ts
+++ b/src/clr-core/common/base/button.base.ts
@@ -4,8 +4,12 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { html, LitElement, property } from 'lit-element';
+import { html, LitElement, property, query } from 'lit-element';
 import { ifDefined } from 'lit-html/directives/if-defined';
+
+import { querySlot } from '../decorators/query-slot';
+import { KeyCodes } from './../enums/key-codes';
+import { stopEvent } from './../utils/events';
 
 // @dynamic
 export class CwcBaseButton extends LitElement {
@@ -33,6 +37,8 @@ export class CwcBaseButton extends LitElement {
     }
   }
 
+  @querySlot('a') private anchor: HTMLAnchorElement;
+
   get disabled() {
     return this._disabled;
   }
@@ -52,16 +58,20 @@ export class CwcBaseButton extends LitElement {
         ></button>`;
   }
 
+  @query('button') private templateButton: HTMLButtonElement;
+  private hiddenButton: HTMLButtonElement;
+
   protected render() {
     return html`
-        <slot></slot>
-        ${this.hiddenButtonTemplate}
+      <slot></slot>
+      ${this.hiddenButtonTemplate}
     `;
   }
 
   protected firstUpdated(props: Map<string, any>) {
     super.firstUpdated(props);
     this.updateButtonAttributes();
+    this.setupNativeButtonBehavior();
   }
 
   protected updated(props: Map<string, any>) {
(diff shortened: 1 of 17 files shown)


src/clr-core/common/utils/enums.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum KeyCodes {
+  LeftArrow = 'ArrowUp',
```
bdryanovski: is this a typo?

src/clr-core/common/utils/enums.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum KeyCodes {
+  LeftArrow = 'ArrowLeft',
```
bdryanovski: It seems that all other `KeyCodes` follow the pattern of `Arrow<Direction>` and only the `LeftArrow` have this reverse pattern `<Direction>Arrow` - is it not better to have the same pattern for all directions?

I did not find a place where we are using yet, so this will be a local change only.

coryrylan: Haha good find, that's what I get for to much copy pasting! 🤣 Will fix up. I have a couple other fixes as well. I'll wait on force pushing anything else.

bdryanovski: I think that we reach perfection for this file. :)

src/clr-core/common/utils/enums.ts
```diff
@@ -0,0 +1,22 @@
...
+  ArrowUp = 'ArrowUp',
+  ArrowRight = 'ArrowRight',
+  ArrowDown = 'ArrowDown',
+  Backspace = 'Backspace',
+  Tab = 'TAB',
+  Enter = 'Enter',
+  Escape = 'Escape',
+  Space = 'Space',
+}
+
```
jeeyun: We have this file: https://github.com/vmware/clarity/blob/master/src/clr-angular/utils/key-codes/key-codes.ts

Wondering if there's opportunities to consolidate or reuse? I also think there was minor differences in KeyCodes and/or Key depending on the browser. I was trying to find a specific example of checking for that in our code base, but maybe this can be discovered as we go along. I'll post a link if I find it.

coryrylan: Completely forgot about the existing ones, thanks! We also have https://github.com/vmware/clarity/blob/master/src/clr-angular/utils/focus/key-codes.enum.ts 
I'll update to reuse the existing ones. I remember the browser difference. In IE the keycodes are different so I'll need to update to check for that.

src/dev-core/src/app/button/button.demo.ts
```diff
@@ -13,10 +14,22 @@ import '@clr/core/button';
...
   mySubmitBtnState = ClrLoadingState.DEFAULT;
 
+  constructor(private formBuilder: FormBuilder) {
+    this.form = this.formBuilder.group({
+      name: ['hello world'],
+    });
+  }
+
+  submit() {
+    console.log('Form Submit:', this.form.value);
```
jeeyun: prob want to remove this console.log even though it's demo code? I think the alert below should be sufficient for testing / sanity check.

src/clr-core/styles/variables/_variables.global.scss
```diff
@@ -44,7 +44,7 @@ $clr_baselineRem_10: convertBaselineToBase20(240); // 240px
 $clr_baselineRem_15: convertBaselineToBase20(360); // 360px
 
 // Pixel-based utilities
-$clr_baselineRem_-1px: -convertBaselineToBase20(1);
+$clr_baselineRem_-1px: -1px;
```
mathisscott: We can't do this. It would break browser scaling. If you need a -1px measurement, your best bet is to use...

`$clr_baselineRem_-1px: -1 * $clr_baselineRem_1px`

...or you could use `-1 * $clr_baselineRem_1px` in your code to prevent us from starting a precedent of creating variables for negative values. That could potentially double the number of variables in our SASS...

coryrylan: Ended up not needing this so reverted the change.

src/dev-core/src/app/button/button.demo.ts
coryrylan: removed

src/clr-core/common/utils/enums.ts
coryrylan: Cleaned up the enum version of the duplicate in clr-core and clr-angular

src/clr-core/common/base/button.base.ts
```diff
@@ -72,10 +83,55 @@ export class CwcBaseButton extends LitElement {
...
+   * This allows us to trigger native submit events within a form element.
+   */
+  private setHiddenButton() {
+    if (!this.readonly) {
+      this.hiddenButton = this.appendChild(this.templateButton);
+    }
+  }
+
+  private setEventListeners() {
+    if (!this.readonly) {
```
mathisscott: if i toggle the `readonly` value from, say, `false` to `true`, does it go back and remove these event listeners? or do the event listeners still stick around and fire?

coryrylan: oh good catch! It should check on each event if its still in readonly or not. If I remember correctly the events will also be cleaned up when the component is destroyed automatically.

coryrylan: Simplified logic to check for updates

PR #4068, opened 2019-12-04
feat: prepare smart popover for dynamic content

Jinnie:

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

IfOpenService and ClrPopoverToggleService do nearly the same. The latter is used by the smart popovers and we want to go toward the new developments and services and avoid parallel visibility control code branches that may be in conflict with each other.

Issue Number: N/A

## What is the new behavior?
No IfOpenService anymore. Using ClrPopoverToggleService instead.

* [ ] Yes
* [x] No

## Other information

The interesting part is in popover-content.ts Most of the rest is service replacement. I could have avoided much of the changes by Proxying the ClrPopoverToggleService from a dummy IfOpenService, but I think this would lead to more unmaintainable code on the long run.

Some parts of code and tests in the Combobox module have been commented, as currently this is dead code and new changes will come with the new implementation of the Combobox.


src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -68,6 +70,18 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
...
+        this.shouldRealign = true;
+        // Avoid flickering on initialization, caused by the asynchronous nature of the
+        // check-collector pattern.
+        this.renderer.setStyle(this.view.rootNodes[0], 'opacity', '0');
+      }),
+      // Here we collect subsequent synchronously received content-check events and only take action
+      // at the end of the cycle. See bellow for details on the check-collector pattern.
+      this.checkCollector.pipe(debounceTime(0)).subscribe(() => {
+        this.doAlign();
+        this.renderer.setStyle(this.view.rootNodes[0], 'opacity', '1');
```
hippee-lee: Should doAlign be responsible for setting it visible when its ready?

src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -68,6 +70,18 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
         } else {
           this.removeContent();
         }
+      }),
```
hippee-lee: I didn't run any code so I'm not sure if its possible, but, what about combining all three of these observables into a single stream that we can check balues with?

combineLatest(this.smartOpenService.openChange, this.smartPositionService.shouldRealign, this.checkCollector).subscribe([openChange, shouldRealign, checkCollector] => {
  // combine logic for adding content and aligning it here.
})

docs/CODING_GUIDELINES_ANGULAR.md
```diff
@@ -137,7 +137,7 @@ Providers that are exported for consumers should be prefixed with `Clr` and suff
 
 ```
 ClrSomeExternalService
-ClrIfOpenService
+ClrIfActiveService
```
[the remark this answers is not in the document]
Jinnie: It's part of the documentation, it lists example services. One of these examples is the ClrIfOpenService. As this service gets removed, I substituted it with another service that exists.

src/clr-angular/utils/popover/popover-content.ts
Jinnie: It could, but:
- They are synchronous, so it's functionally the same.
- I preferred to keep "doAlign" focused on alignment stuff only; and this event handler as a controller method, that instructs: align, please; after this - return visibility.
- I also think it's better for maintenance, as it's close in code to the opacity: 0 change and it's easier to follow the relationship between the two.

src/clr-angular/utils/popover/popover-content.ts
Jinnie: Not really. 
- The CheckCollector is triggered by the ShouldRealign indirectly.
- They don't depend only on OpenChange, but can be triggered manually when manual reposition is needed.
Even if we manage to do it, they are three functionally different observables. I don't see what we could gain in terms of maintainability if we try to minimize the code.

src/clr-angular/forms/combobox/option.spec.ts
```diff
@@ -20,7 +20,11 @@ import { OptionSelectionService } from './providers/option-selection.service';
             Test
         </clr-option>
     `,
-  providers: [IfOpenService, { provide: POPOVER_HOST_ANCHOR, useExisting: ElementRef }, OptionSelectionService],
+  providers: [
+    ClrPopoverToggleService,
+    { provide: POPOVER_HOST_ANCHOR, useExisting: ElementRef },
```
hippee-lee: Do we still need this? The intention with the `util/popover` directives and services is to remove all of the legacy popover code. I think this should be removed with this refactor.

src/clr-angular/forms/combobox/options.spec.ts
```diff
@@ -28,19 +28,19 @@ class TestComponent {}
...
-  providers: [IfOpenService],
+  providers: [ClrPopoverToggleService],
 })
 class TestComponentWithError {}
 
 export default function(): void {
   describe('Select Options Menu Component', function() {
     let context: TestContext<ClrOptions, TestComponent>;
-    let ifOpenService: IfOpenService;
+    let ifOpenService: ClrPopoverToggleService;
```
hippee-lee: nit: I would like to keep the naming consistent. Can this be updated to `toggleService` or `popoverToggleService`?

src/clr-angular/forms/combobox/options.ts
```diff
@@ -41,12 +41,13 @@ export class ClrOptions extends AbstractPopover implements OnDestroy {
   }
 
   private initializeSubscriptions(): void {
-    this.sub = this.ifOpenService.ignoredElementChange.subscribe((el: ElementRef) => {
-      if (el) {
-        this.ignoredElement = el.nativeElement;
-      }
-      this.sub.unsubscribe();
-    });
+    // todo: fix on combobox merge
```
hippee-lee: nit: related to the feedback above about why not adding the ignoredElement now.

src/clr-angular/popover/dropdown/dropdown.spec.ts
```diff
@@ -186,7 +186,7 @@ export default function(): void {
     });
 
     it("doesn't close before custom click events have triggered", function() {
-      const ifOpenService = fixture.debugElement.query(By.directive(ClrDropdown)).injector.get(IfOpenService);
+      const ifOpenService = fixture.debugElement.query(By.directive(ClrDropdown)).injector.get(ClrPopoverToggleService);
```
hippee-lee: nit: naming consistency

src/clr-angular/forms/combobox/all.spec.ts
```diff
@@ -19,7 +19,8 @@ import ComboboxOptionIntegrationSpecs from './combobox-with-option.integration.s
 import ComboboxOptionsMenuIntegrationSpecs from './combobox-with-options-menu.integration.spec';
 import ComboboxSpecs from './combobox.spec';
 
-describe('Combobox component', function() {
+// disable until combobox is merged
+xdescribe('Combobox component', function() {
```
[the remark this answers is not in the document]
Jinnie: Because they refer the old (pre-Jeeyun, or early Jeeyun) code, which does not even surface anywhere. They _are_ getting changed in the new combobox implementation. I _could_ make these tests work for the old code (which, I remind, is not visible to no one, it's just some WIP), but I'll have to overwrite it anyway.
I am commenting them to skip this intermediate and unnecessary step. This will also make my life easier when merging later, because I'll have to only merge commented old code and my new changes. In the other case I'll have to be merging the fixed old code and my new code.

src/clr-angular/forms/combobox/combobox.ts
```diff
@@ -83,12 +83,13 @@ export class ClrCombobox<T> implements AfterContentInit, OnDestroy {
 
   private registerPopoverIgnoredInput() {
     if (this.input) {
-      this.ifOpenService.registerIgnoredElement(this.input);
+      // todo: intentionally commented; resolve while merging the Combobox
+      //this.toggleService.registerIgnoredElement(this.input);
```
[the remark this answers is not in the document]
Jinnie: Same as above. I already have changes ready in my combobox branch for this, which I stashed so I can separate the popover only changes from the combobox changes. 
If I fix them now, it will be fix of the tests of some dead code. And I'll have to merge 2 changes over base instead of 1 change over base.

src/clr-angular/forms/combobox/option.spec.ts
Jinnie: Similar, though not same, as the combobox.spec.ts stuff. 
Similar - it's also about code I'll very soon refactor with the new combobox code.
Not same - unlike combobox.spec.ts here it was easier for me to fix the "dead code" vs commenting it. It's too small.
I mean, in this commit, I'm only removing the IfOpenService and substituting it with ClrPopoverToggleService. But I am not yet pushing the new smart implementation.
So, you can see mixture of util/popover and popover/whatever code. Although combobox is still undead here, it does exist. I need to make the build pass. Here it was simple to get it fixed. In combobox.spec it was hard, so I just commented it to gain momentum. The value there is minimal compared to the gain. Here - it's equal - so I did it.

src/clr-angular/forms/combobox/options.spec.ts
Jinnie: Thanks! Just missed it.

src/clr-angular/forms/combobox/options.ts
Jinnie: Same as above - I prefer to fix it in real context, with the real services, real tests. This is undead code atm :)

src/clr-angular/popover/dropdown/dropdown.spec.ts
Jinnie: Thanks!

src/clr-angular/popover/tooltip/tooltip-trigger.spec.ts
```diff
@@ -28,24 +28,24 @@ class SimpleTest {
...
 export default function(): void {
   describe('TooltipTrigger component', function(this: TooltipContext) {
     spec(ClrTooltipTrigger, SimpleTest, ClrTooltipModule, {
-      providers: [IfOpenService, UNIQUE_ID_PROVIDER, TooltipIdService],
+      providers: [ClrPopoverToggleService, UNIQUE_ID_PROVIDER, TooltipIdService],
     });
 
     beforeEach(function() {
-      this.ifOpenService = this.getClarityProvider(IfOpenService);
+      this.ifOpenService = this.getClarityProvider(ClrPopoverToggleService);
```
gnomeontherun: Let's update the name here too to `toggleService`

src/clr-angular/utils/popover/providers/popover-position.service.ts
```diff
@@ -21,6 +22,13 @@ export class ClrPopoverPositionService {
   private contentOffsets: ClrPopoverContentOffset;
   private _position: ClrPopoverPosition;
 
+  private _shouldRealign: Subject<void> = new Subject();
+  shouldRealign: Observable<void> = this._shouldRealign.asObservable();
+
+  realign() {
```
gnomeontherun: Missing tests

src/clr-angular/utils/popover/providers/popover-toggle.service.ts
```diff
@@ -47,6 +47,10 @@ export class ClrPopoverToggleService {
     return this._open;
   }
 
+  get originalEvent(): Event {
```
gnomeontherun: Is this used?

src/clr-angular/forms/combobox/combobox.ts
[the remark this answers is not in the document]
gnomeontherun: When we have concerns like this, I think it helps to make a common comment anywhere you are commenting something out for a single purpose, then you can search for all of them easily. So update the comments where you've commented out something temporarily.

```
// @TODO COMBOBOX some additional comment
```

src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -98,25 +111,48 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
     // Create the view container
     this.view = this.container.createEmbeddedView(this.template);
     this.smartEventsService.contentRef = this.view.rootNodes[0]; // So we know where/what to set close focus on
-    // Position the content and add a click listener
     this.renderer.addClass(this.view.rootNodes[0], 'clr-popover-content');
+    // Reset to the begining of the document to be available for sizing/positioning calculations.
+    // If we add new content to the bottom it triggers changes in the layout that may lead to false anchor
+    // coordinates values.
+    this.renderer.setStyle(this.view.rootNodes[0], 'top', '0px');
```
gnomeontherun: Why are we changing to the top/left positioning from what we had before?

src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -98,25 +111,48 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
...
-      this.renderer.setStyle(this.view.rootNodes[0], 'top', `${positionCoords.yOffset}px`);
-      this.renderer.setStyle(this.view.rootNodes[0], 'left', `${positionCoords.xOffset}px`);
-      this.hasPositionCoords = true;
+    if (this.smartOpenService.open && this.view && this.shouldRealign) {
+      // Channel content-check event through the check-collector
+      this.checkCollector.emit();
+    }
+  }
+
+  private doAlign() {
```
gnomeontherun: nit: rename this method to `realignContent` or the like.

src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -98,25 +111,48 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
...
+    if (!this.view) {
```
gnomeontherun: Would also check if `shouldRealign` is true before going further just in case.

src/clr-angular/forms/combobox/all.spec.ts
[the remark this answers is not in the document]
Jinnie: Updated this one too to Jeremy's proposed comment format

src/clr-angular/forms/combobox/options.ts
Jinnie: Updated to Jeremy's proposed comment style

src/clr-angular/utils/popover/providers/popover-toggle.service.ts
Jinnie: Yes, in `abstract-popover` and `dropdown-focus-handler.service`

src/clr-angular/utils/popover/popover-content.ts
Jinnie: In order to make the calculations we need to have the element attached to the DOM (sounds unnecessary, but Matt can confirm that in practice we need it). It used to be added to the bottom of the document and after the calculations are done, it gets relocated to the proper position on the body. This works in the static case, but in the dynamic case it changes the overall page size and this affects negatively the smart position relocation logic.
So, as described in the comment above it, I am initializing it at absolute 0/0 instead, which detaches it from the inline DOM structure but still keeps it in the DOM. This way the reposition coordinates calculation is okay.
The actual negative effect was that the repositioned popover was moved to the top with as many pixels as its height was.

src/clr-angular/utils/popover/popover-content.ts
Jinnie: changed to `alignContent` as it is also what we run on init.

src/clr-angular/utils/popover/popover-content.ts
Jinnie: As I mentioned above, this method is used in 2 different cases:
- manual realignment through the `shouldRealign` path
- initialization - then `shouldRealign` will be `false`. 
But maybe it makes sense to move the `shouldRealign = false` line outside this method, so it's only called during the `shouldRealign` path. Same as we do with the opacity flag.
Will change it as described.

src/clr-angular/utils/popover/popover-content.ts
gnomeontherun: 👍  Thanks!

src/clr-angular/utils/popover/popover-content.ts
gnomeontherun: Thanks makes more sense.

PR #4069, opened 2019-12-04
fix(date-picker): prevent calling for focus, when date-picker is closed

### LINKED ISSUE #4060: Datepicker steals focus when hitting escape
## Describe the bug

When using a date picker it always listens for the escape key, even when closed. 

When hitting the escape key when the date picker is closed it will focus on the little calendar icon. If the date-picker is out of view the page will scroll to the picker.

## How to reproduce

1. https://v2.clarity.design/datepicker
2. Hit escape
3. The focus will shift to the first available date picker

## Expected behavior

Nothing should happen when the picker is closed.

## Versions

Reproducible on Clarity 2.3.2 and Angular 8.2.14

bdryanovski:
Prevent from calling focus on date picker when the calendar is not open (depend on ifOpenService)

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4060 

## What is the new behavior?

track ifOpenService state and prevent requesting for focus.

* [ ] Yes
* [x] No

## Other information

Close #4060

### DIFF
--- a/src/clr-angular/forms/datepicker/date-container.ts
+++ b/src/clr-angular/forms/datepicker/date-container.ts
@@ -8,7 +8,7 @@ import {
   OnDestroy,
   Optional,
   ContentChild,
-  HostListener,
+  AfterViewInit,
   ViewChild,
   ElementRef,
   Input,
@@ -81,7 +81,7 @@ import { PopoverPosition } from '../../popover/common/popover-positions';
     '[class.clr-row]': 'addGrid()',
   },
 })
-export class ClrDateContainer implements DynamicWrapper, OnDestroy {
+export class ClrDateContainer implements DynamicWrapper, OnDestroy, AfterViewInit {
   _dynamic: boolean = false;
   invalid = false;
   focus = false;
@@ -111,13 +111,6 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy {
     private ngControlService: NgControlService
   ) {
     this.subscriptions.push(
-      this._toggleService.openChange.subscribe(open => {
-        if (open) {
-          this.initializeCalendar();
-        }
-      })
-    );
-    this.subscriptions.push(
       this.focusService.focusChange.subscribe(state => {
         this.focus = state;
       })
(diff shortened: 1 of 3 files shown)


src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -131,7 +133,9 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy {
 
   @HostListener('body:keyup.escape')
   close(): void {
-    this.toggleButton.nativeElement.focus();
+    if (this.isOpen === true) {
```
hippee-lee: Could this be reduced to just:
```
if (this.isOpen) { ... } 
```

bdryanovski: Sure, 
I have this mental note in my head to write sometimes like that to simplify my head compiler :)

will push fix tomorrow morning

martinbrom:
You also need to set the local isOpen variable back to false, otherwise this happens.

Clarity-GoogleChrome2019-12-0512

martinbrom:
Still the same issue after your last change. It only works when Escape key is used to close the date picker modal. If you close it by clicking outside of it, the scrolling problem occurs.

bdryanovski:
We gonna discuss this in our daily meeting to see what we could do about it.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -129,9 +133,13 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy {
     );
   }
 
-  @HostListener('body:keyup.escape')
-  close(): void {
-    this.toggleButton.nativeElement.focus();
+  @HostListener('keydown', ['$event'])
```
hippee-lee: Lets get rid of this listener and just react to the ifOpenService.openChange subscription. The subscribe may need to move to afterViewInit but it could look like this:

```
this._ifOpenService.openChange.subscription(open => {
  if(open) {
    this.initializeCalendar();
  } else {
    this.toggleButton.nativeElement.focus()
  }
})
```

This way we can get rid of the host listener and still react to open/close changes as needed because the host listener wasn't actually closing the datepicker conteiner.

bdryanovski: We could remove but there is one side effect - it's when you click outside the calendar we gonna get back the focus to the calendar icon. So if this is ok... it's a lot simple to just re-use the `if-else` block.

The above change also will remove the need for zones.

I gonna push PR with that change and we don't like it we could always revert the last commit

bdryanovski:
I have a failing test - will try to fix it before monday

PR #4087, opened 2019-12-04
core: fix badge and tag to re-use same custom properties for its variations

jeeyun:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Instead of using specific variable names like 

Issue Number: N/A

## What is the new behavior?

Instead of using specialized css variables like `--clr-badge-gray-bg-color`, opt to reuse the generic variable name `--clr-badge-bg-color`. This allows an override for the variation like:
```
:host([attr='value']) {
  --clr-badge-bg-color: <custom background color>;
  --clr-badge-color: <custom color>;
}
```

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/badge/_properties.scss
+++ b/src/clr-core/badge/_properties.scss
@@ -4,10 +4,12 @@
 
 @include exports('badge.properties') {
   @if $clr-use-custom-properties == true {
-    :root {
-      --clr-badge-font-color-light: var(--clr-color-neutral-0);
-      --clr-badge-font-color-dark: var(--clr-color-neutral-1000);
+    :root,
+    :host {
+      --clr-badge-background-color: var(--clr-color-neutral-600);
+      --clr-badge-color: var(--clr-color-on-neutral-600);
 
+      // START: deprecated for removal in 4.0
       --clr-badge-info-bg-color: var(--clr-color-action-600);
       --clr-badge-info-color: var(--clr-color-neutral-0);
       --clr-badge-success-bg-color: var(--clr-color-success-700);
@@ -27,6 +29,7 @@
       --clr-badge-orange-color: var(--clr-badge-font-color-dark);
       --clr-badge-light-blue-bg-color: var(--clr-color-action-300);
       --clr-badge-light-blue-color: var(--clr-badge-font-color-dark);
+      // END: deprecated for removal in 4.0
     }
   }
 }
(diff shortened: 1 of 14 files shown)


src/clr-core/badge/_properties.scss
```diff
@@ -4,10 +4,15 @@
...
 @include exports('badge.properties') {
   @if $clr-use-custom-properties == true {
-    :root {
+    :root,
+    :host {
       --clr-badge-font-color-light: var(--clr-color-neutral-0);
       --clr-badge-font-color-dark: var(--clr-color-neutral-1000);
 
+      --clr-badge-bg-color: var(--clr-color-neutral-600);
+      --clr-badge-color: var(--clr-badge-font-color-light);
```
coryrylan: Looks good! Only request I have is if you don't mind updating the TSDoc @cssprop list in the badge and tag component to show just the properties that they use now

jeeyun: Updated and rebased!

src/clr-core/tag/tag.element.scss
```diff
@@ -105,16 +59,78 @@
...
+:host([color='gray']) {
+  --clr-tag-border-color: var(--clr-color-neutral-600);
 
-@include generateStatusTagStyles(info);
-@include generateStatusTagStyles(success);
-@include generateStatusTagStyles(warning);
-@include generateStatusTagStyles(danger);
+  ::slotted(cwc-badge) {
+    --clr-badge-background-color: var(--clr-color-neutral-600);
+    --clr-badge-color: var(--clr-badge-font-color-light);
```
mathisscott: This just occurred to me but you _could_ instead use `--clr-color-on-neutral-600` instead of specifying `--clr-badge-font-color-light|dark`.

This would give people one place to update their colors to achieve accessible color contrast and you could get rid of all instances of `--clr-badge-font-color-light|dark`.

jeeyun: I'm not sure if I follow. Are you suggesting a variable for the foreground / text color (i.e. `--clr-color-on-neutral-600` is the text color for background `--clr-color-neutral-600`? 

Doesn't that mean I'd have to create many more variables for each of the badge background colors that we are using here (and actually for all of our color variables potentially)?

mathisscott: You don’t have to make any variables. They all already exist. Look at the color properties in clr-angular. 

All the colors have an accessible “on” companion property

jeeyun: Wow, we should definitely be using these. Will update the PR.

Issue #2718, opened 2018-10-10
Date Picker (clrDate directive): Introduce clear value button


ivvatov:
Current behavior: 
Currently, in order to remove a selected date, one has to delete the input value manually.

Desired behavior: 
What could be more user-friendly is either (1) a "clear" button in the date popup (clr-datepicker-view-manager) or (2) an "x" button next to the date input ( like in https://material-ui.com/demos/pickers/ ).

The current behavior can be observed at the date picker clarity page (https://vmware.github.io/clarity/documentation/v0.13/datepicker as of the time of writing).

AbhiAgarwal192:
Added data-picker icon. OnClick event for the same is yet to be added. Will keep you posted.

colinreedmiller:
Should this be merged with #963 - Keyboard input for date picker resembles form inputs since forms redesign. There is a design for a clear function in the 'Forms - input groups' design spec.

![forms-imput_groups.png](https://images.zenhubusercontent.com/5bc78fe0ba9b001064ad7062/a051f965-f5f9-430f-9caf-645e58085c5d)

andreiconstantinbytex:
I would like to be assigned to this issue.

gnomeontherun:
This is really just a duplicate of #963, closing in favor of it.

PR #4092, opened 2019-12-05
feat(date-picker): add clear button to reset the date-picker value

bdryanovski:
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Right now there is no way to clear the value from the date-picker unless using a mouse or keyboard to delete character by character.

This PR introduces a way to use clear btn do this for you. 

* The button has to be visible only when there is a value into the date-picker.

Issue Number: N/A

* [ ] Yes
* [x] No

## Other information

Close: #2718


bdryanovski:
There is this change right now, that the icon is changed from `times-circle` to `times` only - The original icon from the design proposal is not part of the `Core` icons so until we decide to change it - I gonna leave it like that.

Other option will be to copy and import the icon manually as part of this component?

mathisscott:
> Other option will be to copy and import the icon manually as part of this component?

That's not a viable option until we have the `@clr/core` icons available.

Why was `times-circle` chosen in the first place? The additional stroke around the 'X' seems to be visual noise to me and I don't know what an encircled 'X' communicates that a normal 'X' would not...

bdryanovski:
It's part of the original story and the design mockups. 

68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3562633738666530626139623030313036346164373036322f61303531663936352d663566392d343330662d396361662d363435653538303835633564


I also, find the simple `X` good enough on indicating what will be the end result.

If everyone is fine with that, let's keep the `times` icon that is already part of the `Core Icons` and if in feature release we need to add the circle we could just patch it. Or the best case will be to use it from `@clr/core/icons`

gnomeontherun:
I have a much simpler implementation to show you. There is too much manual handling, we just need to have a simple button that uses the `NgControl.reset()` method to clear the input.

bdryanovski:
It looked a lot more simple but it's not working in the cases that we don't use the `NgControl`. I tested it inside the dev app and have a 50% success rate. For example on the first demo, that relay on `ngModel` it was `false`. There are few ways to send manage the internal state of the date-picker and this just don't cover all of them.  

I didn't get why we need to have this `clrClearEnabled` - is there any need to enable/disable the clear button or even providing the option to do it?

gnomeontherun:
I have the `clrButtonEnabled` option to allow people to turn it on or off. I would want this to be applied to all inputs if we make it on by default. I haven't dug far enough into your implementation, but having to use ChangeDetectorRef is a concern that I don't think we should have to have for this.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -142,6 +175,15 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy, AfterViewIni
     );
   }
 
+  private _clearEnabled = false;
+  @Input()
```
hippee-lee: Is there a spec for this that makes the clear button optional? Shouldn't it always appear if there is a value? 

It seems like an unneeded burden for consumers to have to add the input.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -122,12 +137,30 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy, AfterViewIni
...
+
   ngOnInit() {
     this.subscriptions.push(
       this.ifErrorService.statusChanges.subscribe(invalid => {
         this.invalid = invalid;
       })
     );
+
+    /** handle when to show are hide the clear button */
+    this.subscriptions.push(
```
hippee-lee: Why not use an @ContentChild for the clrDate element and subscribe to value changes there? It might eliminate the need for the service changes.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -122,12 +137,30 @@ export class ClrDateContainer implements DynamicWrapper, OnDestroy, AfterViewIni
...
+      this.dateIOService.dateInputValue.subscribe((value: string) => {
+        this.inputValue = value;
+        // on init the subscribe method will be called twise once with `null` and
+        // one with the actial value and this will trigger angular error for
+        // ExpressionChangedAfterItHasBeenCheckedError
+        this.cd.detectChanges();
```
hippee-lee: I'm still reviewing but this looks like a no go to me.

gnomeontherun:
I've updated my demo example of what I would be expecting to address the non-angular form case. 

https://github.com/gnomeontherun/clarity/commit/56eb233beb924726a73142cbda3ee4e5fda3a5e0

src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -203,6 +218,12 @@ export default function() {
         controlClassService.className = 'clr-col-2';
         expect(context.clarityDirective.controlClass()).not.toContain('clr-col-md-10');
       });
+
+      it('clearDateValue should call method from DateIOService to request clear', () => {
```
gnomeontherun: `DateFormControlService` not `DateIOService`

src/clr-angular/forms/datepicker/date-input.ts
```diff
@@ -255,7 +274,25 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
+        }
+        const input = this.dateIOService.getDateValueFromDateString(value);
+        if ((this.previousDateChange === null || this.previousDateChange === undefined) && input === null) {
+          // when the new date will result into null and previous value is null
+          // no need to update
+          return;
+        }
+
+        // toString() let me compare Date .. milliseconds are a thing!
+        if ((input && input.toString()) !== (this.previousDateChange && this.previousDateChange.toString())) {
```
gnomeontherun: Might be better to use the new `toComparisonString()` method so you get a `YYYYMMDD` string without the time values.

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -43,6 +43,18 @@ import { PopoverPosition } from '../../popover/common/popover-positions';
...
             <ng-content select="[clrDate]"></ng-content>
+            <button
+              type="button"
+              class="clr-input-group-icon-action clr-datepicker-close-btn"
+              [attr.title]="commonStrings.keys.datepickerClear"
+              [attr.aria-label]="commonStrings.keys.datepickerClear"
+              [disabled]="control?.disabled"
+              (click)="clearDateValue()"
+              *ngIf="dateValue">
+                <clr-icon shape="times"></clr-icon>
```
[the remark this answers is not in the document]
bdryanovski: `times-circle` is not part of Core shapes - so to not create breaking change or manually importing the icon I went with `times` only.

src/clr-angular/forms/datepicker/providers/date-form-control.service.ts
```diff
@@ -9,6 +9,8 @@ import { Observable, Subject } from 'rxjs';
 
 @Injectable()
 export class DateFormControlService {
+  value;
```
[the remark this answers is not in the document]
bdryanovski: Sure

src/clr-angular/forms/datepicker/date-container.ts
```diff
@@ -43,6 +43,18 @@ import { PopoverPosition } from '../../popover/common/popover-positions';
         <div class="clr-input-wrapper">
           <div class="clr-input-group" [class.clr-focus]="focus">
             <ng-content select="[clrDate]"></ng-content>
+            <button
+              type="button"
+              class="clr-input-group-icon-action clr-datepicker-close-btn"
+              [attr.title]="commonStrings.keys.datepickerClear"
+              [attr.aria-label]="commonStrings.keys.datepickerClear"
+              [disabled]="control?.disabled"
```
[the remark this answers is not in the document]
bdryanovski: I gonna merge the bugfix today so will address it in the next commit for date-picker clear/button.

src/clr-angular/forms/datepicker/date-container.spec.ts
bdryanovski: Typo!

src/clr-angular/forms/datepicker/date-input.ts
bdryanovski: Sounds, better will do!

bdryanovski: After testing it seems that the `input` could be:

* Date
* String
* empty string

So it's getting pretty complex to use `new DayModel` and converting `input` into YYYY-MM-DD.

So `toString()` is doing this a lot faster and simple  - so prefer to keep it like it is.

gnomeontherun: I think there might be edge cases, where dates have timezone conflicts. There was another bug with this problem recently, so I would prefer to make `input` be a DayModel and compare the same way. Might be good to hoist the comparison logic to one place since its being duplicated.

src/clr-angular/forms/datepicker/date-container.ts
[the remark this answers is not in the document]
gnomeontherun: We can move it to Core and then alias it in the other group, we've done it before.

src/clr-angular/forms/datepicker/date-container.spec.ts
```diff
@@ -28,13 +28,15 @@ import { DatepickerEnabledService } from './providers/datepicker-enabled.service
 import { MockDatepickerEnabledService } from './providers/datepicker-enabled.service.mock';
 import { LocaleHelperService } from './providers/locale-helper.service';
 import { ClrCommonFormsModule } from '../common/common.module';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 export default function() {
-  describe('Date Container Component', () => {
+  fdescribe('Date Container Component', () => {
```
gnomeontherun: `fdescribe`

src/clr-angular/forms/datepicker/providers/date-form-control.service.ts
```diff
@@ -49,4 +48,9 @@ export class DateFormControlService {
   public clear() {
     this.dataClearEvent.next();
   }
+
+  // friendly wrapper
+  setDisabled(state: boolean) {
```
gnomeontherun: Why?

bdryanovski: xaxa - it's gone now, probably some old cache or something. Merged again with master and now is showing it correctly.

src/clr-angular/forms/datepicker/providers/date-form-control.service.ts
```diff
@@ -30,6 +31,24 @@ export class DateFormControlService {
     this._dirtyChange.next();
   }
 
+  /**
+   * Handle Date value across components
+   */
+  private dataClearEvent: Subject<void> = new Subject();
```
hippee-lee: Perhaps I am missing something here. Is there a reason this service shouldn't use a private value to store the current date value for the datepicker (on this service) and use the typical `private property/Subject/Observable` pattern? e.g

```
private currentDate: Date | string;
private _dateValue: Subject<Date|string> = new Subject<Date|string>();
get dateValue: Observable<Date|string>() { return this._dateValue.asObservable(); }
```

Then this service would also centralize the `setter for the date value` and `clear` functions for currentDate which would allow listeners to reactively get the updates/changes to the dateValue.

bdryanovski:
@gnomeontherun thanks for the label.

For now, I want to put this on hold until there is a more viable solution on communication between components that don't generate a chocolate error.

```ts
  ngAfterViewInit() {
    // I don't know why I have to do this but after using the new HostWrapping Module I have to delay the processing
    // of the initial Input set by the user to here. If I do not 2 issues occur:
    // 1. The Input setter is called before ngOnInit. ngOnInit initializes the services without which the setter fails.
    // 2. The Renderer doesn't work before ngAfterViewInit (It used to before the new HostWrapping Module for some reason).
    // I need the renderer to set the value property on the input to make sure that if the user has supplied a Date
    // input object, we reflect it with the right date on the input field using the IO service. I am not sure if
    // these are major issues or not but just noting them down here.
    this.processInitialInputs();
  }
```

This fix is triggering another update of the `date-value` - that is triggering another update cycle inside the `date-container` in the middle of the previous. 

Stoping `processInitialInputs` will fix the issue with the double update cycle - but at the same time won't update the visual state of the component (all the time). 

Until there is a good solution on how to remove `processInitialInputs` there is no other way to handle the update change in the parent without the use of `ChangeDetectorRef`.

PR #4130, opened 2019-12-18
feat: make smart repositioning configurable in the popovers

Jinnie:

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

We can't switch off automatic repositioning of the smart popovers

Issue Number: N/A

## What is the new behavior?

We can use "smartPositioning: false" parameter on the structural directive of popover-content to disable smart positioning.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/utils/popover/popover-content.ts
+++ b/src/clr-angular/utils/popover/popover-content.ts
@@ -31,6 +31,7 @@ import { debounceTime } from 'rxjs/operators';
 export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
   private view: EmbeddedViewRef<void>;
   private subscriptions: Subscription[] = [];
+  private _smartPositioning: boolean = true;
 
   @Input('clrPopoverContent')
   public set open(value: boolean) {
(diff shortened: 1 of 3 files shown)


src/clr-angular/utils/popover/popover-content.ts
```diff
@@ -52,6 +53,11 @@ export class ClrPopoverContent implements AfterContentChecked, OnDestroy {
     this.smartEventsService.scrollToClose = !!scrollToClose;
   }
 
+  @Input('clrPopoverContentSmartPositioning')
```
bdryanovski: I don't see the need for this `setter`;

```ts
@Input('clrPopoverContentSmartPositioning') smartPositioning: boolean;
```
It will do the same job. This also will remove the need for `this. _smartPositioning`. 
I'm left with the idea that we are going away from adding `_` as the prefix to indicate private variables and use them only in the cases when we are creating `setter/getter` with the same name.
```ts
private _open: boolean;
set open(v: boolean) { /* some code transformation */ }
get open(): boolean { /* some code transformation */ }
```

Jinnie: I also had this thought at some point, but intentionally decided against it.
With this syntax it remains fixed only as input parameter. If I expose it as something that the user may read through the typescript API, then for completeness I'd better also expose smartPositioningChange observable, or maybe even an Output. For now I find this to be obsolete, as this variable can't change internally in the component and I prefer to keep the possible interactions only to the intended cases, with one-way input binding.
Hope this makes sense?

gnomeontherun:
@Jinnie are we going to close this based on our discussion? If so please do or advise what you intend to do with this PR.

PR #4142, opened 2019-12-24
feat(stack-view): a11y aria-level, aria-posinset and aria-setsize

### LINKED ISSUE #3568: Lazy loaded stack views are missing position information for screen readers
## Describe the bug

Lazy loading stack views should provide position in set information for each stack item in view at any given time. 

## How to reproduce

https://clarity.design/documentation/stack-view

- Using a screen reader find the 'Lazy Loading of Children' example and read one of the 'stack-block-label' buttons. Observe that you don't hear any information regarding the data presented in 'stack-block-label' 

## Expected behavior

Use aria-level, aria-setsize, and aria-posinset as on each 'stack-block-label' button. 

For information on using aria-level, aria-setsize, and aria-posinset see https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

bdryanovski:
Adding aria-level, aria-posinset and aria-setsize attributes to stack-view
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3568

* [ ] Yes
* [x] No

## Other information
Close #3568

### DIFF
--- a/src/clr-angular/data/stack-view/stack-block.spec.ts
+++ b/src/clr-angular/data/stack-view/stack-block.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -14,15 +14,18 @@ import { ClrStackViewModule } from './stack-view.module';
 
 @Component({
   template: `
-        <clr-stack-block>
-            <clr-stack-label>Label</clr-stack-label>
-            <clr-stack-content>Content</clr-stack-content>
-        </clr-stack-block>
-   `,
+    <clr-stack-block [clrStackViewLevel]="ariaLevel" [clrStackViewSetsize]="ariaSetsize" [clrStackViewPosinset]="ariaPosinset">
+      <clr-stack-label>Label</clr-stack-label>
+      <clr-stack-content>Content</clr-stack-content>
+    </clr-stack-block>
+  `,
 })
 class BasicBlock {
   @ViewChild(ClrStackBlock, { static: false })
   blockInstance: ClrStackBlock;
+  ariaLevel: number;
+  ariaSetsize: number;
+  ariaPosinset: number;
 }
 
 @Component({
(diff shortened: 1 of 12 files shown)


src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -41,7 +48,9 @@ import { UNIQUE_ID, UNIQUE_ID_PROVIDER } from '../../utils/id-generator/id-gener
   // Custom elements are inline by default
   styles: [
     `
-        :host { display: block; }
+      :host {
```
hippee-lee: Should this be in a scss file? I don't think we use the styles array by convention.

src/clr-angular/data/stack-view/stack-block.ts
```diff
@@ -80,11 +89,26 @@ export class ClrStackBlock implements OnInit {
     }
   }
 
+  /**
+   * Depth of the stack view starting from 1 for first level
+   */
+  @Input('clrStackViewLevel') ariaLevel: number;
```
hippee-lee: From what I see here and in the demos, this implementation means consumers will have to manage these values and updated them as the model changes. I'm not clear from reading #3568 if thats how this should behave. 

imo - The stackblock component should handle these values internally so that the consumer doesn't need to worry about managing these values.

src/clr-angular/data/stack-view/stack-block.ts
gnomeontherun: This was already there, its just reformatted.

gnomeontherun:
@hippee-lee I talked with @bdryanovski about this initially, and this is really only for the lazy case where we don't know the values defined here. There isn't much chance that we can compute these automatically, and only if know we have the full node displayed. 

The basic approach of having the app provide these values is what I wanted to see in this PR, as it will enable any lazy tree to have these values, and then we could consider how to automate it if there is a need. However, apps know these values, we do not in, lazy mode.

src/clr-angular/data/stack-view/stack-block.ts
hippee-lee: per discussion below this isn't relevent.

gnomeontherun:
@bdryanovski Can you update this to work with the new DOM structure? Otherwise I think its about ready to go.

bdryanovski:
Oh, I didn't say that there is a conflict.



... fixing

src/clr-angular/data/stack-view/stack-block.spec.ts
```diff
@@ -83,6 +86,30 @@ export default function(): void {
       return bFixture.componentInstance.blockInstance;
     }
 
+    describe('Accessibility', () => {
+      let dd: HTMLElement;
```
gnomeontherun: nit: its no longer a `dd` so maybe just use `let label: HTMLElement;` is better.

bdryanovski: I will fix it with the rebase.

chlane:
The current usage of  a11y aria-level, aria-posinset and aria-setsize is not working with Safari/VoiceOver and Firefox/NVDA because these properties are not being used in combination with roles that support them. For example, aria-posinset is supported on element with role="option" within an ARIA listbox implementation. Another example would be aria-level being used on an element with role="heading" so that element would become a semantic heading. I recommend removing aria-posinset and aria-setsize. aria-level can be used similar to how the standard HTML headings are used in https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html. Be sure to choose heading levels that fit appropriately into the document outline.

chlane:
This was my idea originally but it doesn't seem to be working, probably because aria-level, aria-posinset and aria-setsize aren't being used something like an ARIA a tree. I recommend removing aria-posinset and aria-setsize. aria-level can be used similar to how the standard HTML headings are used in https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html. Be sure to choose heading levels that fit appropriately into the document outline.

PR #4143, opened 2019-12-25
var $clr_baselineRem_-1px

KANekT:
var $clr_baselineRem_-1px not working.

in css it -convertBaselineToBase20(1) must be -0.05rem.

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/styles/variables/_variables.global.scss
+++ b/src/clr-core/styles/variables/_variables.global.scss
@@ -44,7 +44,7 @@ $clr_baselineRem_10: convertBaselineToBase20(240); // 240px
 $clr_baselineRem_15: convertBaselineToBase20(360); // 360px
 
 // Pixel-based utilities
-$clr_baselineRem_-1px: -convertBaselineToBase20(1);
+$clr_baselineRem_-1px: -1 *convertBaselineToBase20(1);
 $clr_baselineRem_0_5px: convertBaselineToBase20(0.5); // used by cwc-tag to depress the tag .5px when user clicks
 $clr_baselineRem_1px: convertBaselineToBase20(1);
 $clr_baselineRem_2px: convertBaselineToBase20(2);


gnomeontherun:
This seems to have unexpected side effects in visual testing, we need to look more closely.

mathisscott:
The recommendation is to not do `$clr_baselineRem_-*px` variables. This could conceivably double the number of rem-sizing variables we have in our SASS.

The recommendation is to multiply whatever negative rem-sizing you need by `-1`.

So in your SASS you would put...

```
whatever: -1 * $clr_baselineRem_1px
```

src/clr-core/styles/variables/_variables.global.scss
```diff
@@ -44,7 +44,7 @@ $clr_baselineRem_10: convertBaselineToBase20(240); // 240px
 $clr_baselineRem_15: convertBaselineToBase20(360); // 360px
 
 // Pixel-based utilities
-$clr_baselineRem_-1px: -convertBaselineToBase20(1);
+$clr_baselineRem_-1px: -1 *convertBaselineToBase20(1);
```
mathisscott: The recommendation is to not do `$clr_baselineRem_-*px` variables. This could conceivably double the number of rem-sizing variables we have in our SASS.

The recommendation is to multiply whatever negative rem-sizing you need by `-1`.

So in your SASS you would put...

```
whatever: -1 * $clr_baselineRem_1px
```

gnomeontherun:
This breaks some things for IE11 so there is an issue with the way that it is generating CSS perhaps with the spacing issue. However, we do use this variable in one place only, and it should really be removed. Therefore we're going to close this and suggest you just apply the negative on the app side instead of relying on this variable which may be removed.

KANekT:
Thanks

PR #4151, opened 2019-12-30
feat: update the limit of header-max-length from 72 to 100 characters

bdryanovski:
I'm getting super often this error:

```
✖   header must not be longer than 72 characters, current length is 74 [header-max-length]
```
Visible in the case when I use one of the long `scopes` like `date-picker`, `progress-bar` and so on.

So in the case of: 
```
feat(date-picker):<space><commit-message>
```
 I'm losing 19 characters from the commit message just for tagging. And including one input or output name, there are no characters left to do a meaningful sentence. And the messages become a bit cryptic just to get inside the limitation. 

I don't suggest to go and try to fill the how 100 character limit, but having some additional characters could be handy.

Also, there is similar open PR for this change inside the [commitlint](https://github.com/conventional-changelog/commitlint/issues/303)

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -45,5 +45,6 @@ module.exports = {
   extends: ['@commitlint/config-conventional'],
   rules: {
     'scope-enum': [2, 'always', scopes],
+    'header-max-length': [2, 'always', 100],
   },
 };


bdryanovski:
Currently, we are losing a lot of the message for formating and scoping. So we are left with a little as none space to write something that explains the commit. Also, this is not forcing anyone to go to a full 100 characters. Just giving you more space to write a short message without thinking about how to get that one character more. 

In general, the good practice is to go with max 50 characters for the length of the commit message. 72 is the other number that people for some reason like to limit themselves. Both numbers come just from a statistical analyze on git Kernel log. The limits of 72 are `line-limit` - design to be used into automated mailing systems - so the message is easily visible on old screens that don't have a lot of space in the X-axis. But if we want to promote and use the scope to add context, in the worse case we are having to subtract 21 (`feat:(vertical-nav): `) characters from the 50/72 limit - that is letting us with 29/51 characters left.

A better thing that we have to do is using more and more the scope option of the commit message. It's easier to get the quick context of the message just by reading `feat/fix/docs(<component-name>): `

```bash
# Not clear where?
fix: focus trap no longer steals focus when going outside

# Better
fix(datagrid): focus trap no longer steals focus when going outside
```
On the example above is clear how much additional space is added just to provide the context. 

I don't think that someone will **abuse** this limit and try to write all of his commits to the fullest extent of the rule, but don't see how 72 as a number is making things better. This limit is too conservative to forcing people of writing good messages. 

I'm betting that there will be no commit message by the end fo 2020 more than 85 ch.

PS.
Providing Github ID inside the commit message could easily help to find the original issue and related conversations. Especially helpful when generating release notes.

gnomeontherun:
We need to have a dev chat about this before we proceed. I'm not opposed to longer, but we need to be clear as a team about how we want to write messages and generate notes from it.

bdryanovski:
```bash
⧗   input: fix(date-picker): let date-container to be disabled without clrForm wrapper

Signed-off-by: Bozhidar Dryanovski <bozhidar.dryanovski@gmail.com>
✖   header must not be longer than 72 characters, current length is 75 [header-max-length]
```

PR #4154, opened 2019-12-31
core: tree shaking and performance build

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

In certain situations tree shaking icons was not functioning properly. Sometime depending on the
consuming application build the icons or components would be dropped or tree shaken out unintentionally.

Issue Number: N/A

## What is the new behavior?
- Cleans up the output build of the core package to only ship pure es2015 code. This gives us the optimal build tooling support.

- Update Clarity City usage docs with Core. Remove the global font CSS as its no longer needed and we can directly import from the new Clarity City package 1.1.0

- Fix icon tree shaking. Rename the `/icon` directory to `icon-shapes` directory to make all code related to icons side effect free. Create a new `/icon` directory that only exports a side effect to register the icon element. This ensures users can import the icon and only the icons needed within their application.

  ```typescript
  import '@clr/core/icon';
  import { ClarityIcons, userIcon, menuIcon } from '@clr/core/icon-shapes';
 
  ClarityIcons.addIcons(userIcon, menuIcon);
  ```

- Cleans up what is exported publicly. Components are no longer exported from the root entry point `import { ... } '@clr/core';`
and must be imported from their specific entry points. `import { CwcButton } from '@clr/core/button';`.
This ensures the side effect of registering the component is isolated to only its own entry
point.

- The PR also includes a small Webpack build that will consume the final Core output by importing a single component and single icon. Once completed the bundle generated will be checked in the build/CI process to ensure Webpack has properly tree shaken out unused components and icons

- Fixes issue where icon component was setting the `dir` property as a default in the constructor. This is an issue because `dir` is a existing HTML attr and causes the attr to update within the constructor. This will cause an error in some browsers as it violates the custom elements spec. https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance 
https://stackoverflow.com/questions/43836886/failed-to-construct-customelement-error-when-javascript-file-is-placed-in-head

- This PR removes the golden file check for Core. This was not checking the entire API as we only export a few things from the root entry point. I will open a new PR that will add something more robust that will check the entire public API surface for core.

* [ ] Yes
* [x] No


package.json
```diff
@@ -83,24 +82,22 @@
...
-    "core:build": "npm-run-all core:lint core:sass:components core:build:ngpackagr core:build:package core:sass core:api:editor core:api core:golden core:build:package:format",
+    "core:test:bundle": "webpack --config src/clr-core/test-bundles/webpack.config.js && bundlesize --config src/clr-core/test-bundles/bundlesize.json",
+    "core:build": "npm-run-all core:lint core:sass:components core:build:ngpackagr core:build:package core:sass core:api:editor core:api core:build:package:format",
+    "core:build:watch": "npm run core:sass && ng build clr-core --watch",
     "core:build:package": "cpy ./src/clr-core/README.md ./dist/clr-core/; replace '@VERSION' $npm_package_version ./dist/clr-core/package.json",
     "core:build:package:format": "node ./src/clr-core/package.js",
     "core:build:ngpackagr": "ng build clr-core",
-    "core:golden": "ts-api-guardian --out dist/golden/clr-core.d.ts dist/clr-core/index.d.ts",
-    "core:golden:fix": "npm run core:sass && npm run core:build:ngpackagr && ts-api-guardian --out golden/clr-core.d.ts dist/clr-core/index.d.ts",
-    "core:golden:test": "diff -u golden/clr-core.d.ts dist/golden/clr-core.d.ts",
```
mathisscott: We're getting rid of the golden file for core?

src/clr-core/badge/badge.element.ts
```diff
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { registerElementSafely } from '@clr/core/common';
+import { baseStyles, registerElementSafely } from '@clr/core/common';
```
mathisscott: 👍

package.json
coryrylan: It wasnt checking the entire API properly so we will have to come up with a new strategy for Core. I have some ideas but will attempt it in a different PR.

PR #4155, opened 2020-01-02
chore(build): move website build to circle not netlify

gnomeontherun:
Netlify has started to limit the number of build minutes available to us. This moves the build of the website to CircleCI and deploys it in essentially the same way, but saves us from using build minutes on Netlify. It actually builds and deploys the dev and website apps as separate deployments.

The main drawback here is that we lose the GitHub integration that posts comments and URLs into the PR, but we could do that manually if we setup a script.

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [x] CI related changes
* [x] clarity.design website / infrastructure changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,37 @@ jobs:
             - /home/circleci/.cache/Cypress
           key: v1-visual{{checksum "package.json"}}
       - run: sh ./scripts/circleci/visual.sh
+  website:
+    docker:
+      - image: circleci/node:dubnium
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-website{{checksum "package.json"}}
+      - run: npm install
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-website{{checksum "package.json"}}
+      - run: npm run website:render
+      - run: npm run website:deploy
+  dev:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -34,6 +34,37 @@ jobs:
...
+          keys:
+            - v1-dev{{checksum "package.json"}}
+      - run: npm install
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dev{{checksum "package.json"}}
+      - run: npm run angular:dev:build
+      - run: npm run angular:dev:deploy
+
```
coryrylan: Is it faster to run the dev app and website deploy as separate jobs in parallel or append to the existing build job and deploy from there? Since the main job has already built the libs and dev apps it would just need to then build the website and deploy.

gnomeontherun: I thought about it quickly but wanted to keep it separate for now since this is somewhat urgent to get around the build limit. 

We could adopt this now and then talk about how to organize the jobs most efficiently?

coryrylan: Sounds good!

PR #4163, opened 2020-01-06
fix(a11y): datagrid expandable row accessibility limitations

### LINKED ISSUE #3375: Document Datagrid expandable row accessibility limitations
The current expandable row pattern should be documented with the accessibility limitations, and our recommendation should focus on the detail pane pattern (#2005) when available. 

Here are the steps to improve the accessibility of this control.

1. Use aria-describedby to give users the following message "Screen reader table commands may not work for viewing expanded content, please use your screen reader's browse mode to read the content exposed by this button". See https://www.w3.org/TR/wai-aria-1.1/#aria-describedby for more information.
2. Use aria-controls to associate the button with the expanded row. See https://www.w3.org/TR/wai-aria-1.1/#aria-controls for more information.
3. Provide textual boundaries i.e, "beginning of expanded content" and "beginning of expanded content" for screen readers users to understand what part of the grid they are reading.

bdryanovski:
Provide helper text to be used for describing the expandable content. And aria control attribute to point what part of the visible aria is controlled by Expand/Collapse button.

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3375 

* [ ] Yes
* [x] No

## Other information
Close #3375

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row-detail.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row-detail.ts
@@ -12,6 +12,7 @@ import { RowActionService } from './providers/row-action-service';
 import { Selection } from './providers/selection';
 import { SelectionType } from './enums/selection-type';
 import { DatagridIfExpandService } from './datagrid-if-expanded.service';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 /**
  * Generic bland container serving various purposes for Datagrid.
@@ -19,11 +20,18 @@ import { DatagridIfExpandService } from './datagrid-if-expanded.service';
  */
 @Component({
   selector: 'clr-dg-row-detail',
-  template: ` <ng-content></ng-content> `,
+  template: `
+    <div class="clr-sr-only">
+      {{ beginningOfExpandableContentAriaText }} {{ commonStrings.keys.dategridExpandableRowsHelperText }}
+    </div>
+    <ng-content></ng-content>
+    <div class="clr-sr-only">{{ endOfExpandableContentAriaText }}</div>
+  `,
   host: {
     '[class.datagrid-row-flex]': 'true',
     '[class.datagrid-row-detail]': 'true',
     '[class.datagrid-container]': 'cells.length === 0',
+    '[attr.id]': 'expand.expandableId',
   },
 })
 export class ClrDatagridRowDetail implements AfterContentInit, OnDestroy {
@@ -34,7 +42,8 @@ export class ClrDatagridRowDetail implements AfterContentInit, OnDestroy {
     public selection: Selection,
     public rowActionService: RowActionService,
     public expand: DatagridIfExpandService,
-    public expandableRows: ExpandableRowsCount
+    public expandableRows: ExpandableRowsCount,
+    public commonStrings: ClrCommonStringsService
   ) {}
 
   @ContentChildren(ClrDatagridCell) cells: QueryList<ClrDatagridCell>;
@@ -56,5 +65,21 @@ export class ClrDatagridRowDetail implements AfterContentInit, OnDestroy {
 
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
+  }
+
+  @Input('clrRowDetailBeginningAriaText') _beginningOfExpandableContentAriaText: string;
+  public get beginningOfExpandableContentAriaText() {
+    return (
+      this._beginningOfExpandableContentAriaText ||
+      `${this.commonStrings.keys.dategridExpandableBeginningOf} ${this.commonStrings.keys.dategridExpandableRowContent}`
+    );
+  }
+
+  @Input('clrRowDetailEndAriaText') _endOfExpandableContentAriaText: string;
+  public get endOfExpandableContentAriaText() {
+    return (
+      this._endOfExpandableContentAriaText ||
+      `${this.commonStrings.keys.dategridExpandableEndOf} ${this.commonStrings.keys.dategridExpandableRowContent}`
+    );
   }
 }
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
@@ -59,6 +59,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit 
   public id: string;
   public radioId: string;
   public checkboxId: string;
+  public expandableId: string;
 
   /* reference to the enum so that template can access */
   public SELECTION_TYPE = SelectionType;
@@ -90,6 +91,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit 
     this.id = 'clr-dg-row' + nbRow;
     this.radioId = 'clr-dg-row-rd' + nbRow;
     this.checkboxId = 'clr-dg-row-cb' + nbRow;
+    this.expandableId = expand.expandableId;
 
     this.subscriptions.push(
       combineLatest(this.expand.replace, this.expand.expandChange).subscribe(
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -297,6 +297,10 @@ export interface ClrCommonStrings {
     currentPage: string;
     danger: string;
     datagridFilterAriaLabel?: string;
+    dategridExpandableBeginningOf?: string;
+    dategridExpandableEndOf?: string;
+    dategridExpandableRowContent?: string;
+    dategridExpandableRowsHelperText?: string;
     datepickerCurrentDecade: string;
     datepickerCurrentMonth: string;
     datepickerNextDecade: string;
@@ -661,6 +665,7 @@ export declare class ClrDatagridRow<T = any> implements AfterContentInit, AfterV
     expand: DatagridIfExpandService;
     expandAnimation: ClrExpandableAnimation;
     expandAnimationTrigger: boolean;
+    expandableId: string;
     get expanded(): boolean;
     set expanded(value: boolean);
     expandedChange: EventEmitter<boolean>;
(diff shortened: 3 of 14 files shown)


src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -21,7 +21,10 @@
 -->
 <ng-template #detail>
   <div [id]="expandableId">
+    <div class="clr-sr-only">{{ commonStrings.keys.dategridExpandableRowsHelperText }}</div>
```
gnomeontherun: I think this should be inside of the `clr-dg-row-detail` component instead of in its host.

gnomeontherun:
@bdryanovski can we get this fixed up soon? Need to check with @chlane on this before we merge.

chlane:
We should put the textual boundaries inside the clr-dg-row-detail as seen below. The boundary text should also provide context as well as seen below.

<clr-dg-row-detail _ngcontent-clarity-c232="" class="datagrid-row-flex datagrid-row-detail datagrid-container ng-star-inserted" style="">

<div _ngcontent-clarity-c167="" class="clr-sr-only ng-tns-c167-15">beginning of detail content for record 1234</div>

 Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin in neque in ante placerat mattis id sed quam. Proin rhoncus lacus et tempor dignissim. Vivamus sem quam, pellentesque aliquet suscipit eget, pellentesque sed arcu. Vivamus in dui lectus. Suspendisse cursus est ac nisl imperdiet viverra. Aenean sagittis nibh lacus, in eleifend urna ultrices et. Mauris porttitor nisi nec velit pharetra porttitor. Vestibulum vulputate sollicitudin dolor ut tincidunt. Phasellus vitae blandit felis. Nullam posuere ipsum tincidunt velit pellentesque rhoncus. Morbi faucibus ut ipsum at malesuada. Nam vestibulum felis sit amet metus finibus hendrerit. Fusce faucibus odio eget ex vulputate rhoncus. Fusce nec aliquam leo, at suscipit diam. 

<div _ngcontent-clarity-c167="" class="clr-sr-only ng-tns-c167-15">End of detail content for record 1234</div>
</clr-dg-row-detail>

bdryanovski:
I'm not sure that we could get this `1234` so we could create this type of the sentence `beginning of detail content for record 1234`. This could be the responsibility of the developer to provide the best text to suit the case.

I cleaned and refactor a bit the code so we could provide two `@Input`s for providing custom text.

```ts
@Input('ariaTextBeginningOfExpandableContent') ariaTextBeginningOfExpandableContent:string;
public get beginningOfExpandableContent() {
    return this.ariaTextBeginningOfExpandableContent || `${this.commonStrings.keys.beginningOf} ${this.commonStrings.keys.dategridExpandableRowContent}`;
  }

// same as endOfExpandableContent
```

gnomeontherun:
@chlane this is testable here, can you verify this behavior? https://5e7092ad6ca308c201210c1e--vmware-clarity.netlify.com

I'm only worried about having to write custom 'help text' like we have, since it is at the beginning at each of the expandable rows the user would hear it in every row. Do we _really_ need this message, and if so can it go at the top of the Datagrid instead so its only read once?

gnomeontherun:
@Shijir do you mind doing a sanity check on this one?

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -258,4 +260,20 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
   public get _view() {
     return this.wrappedInjector.get(WrappedRow, this.vcr).rowView;
   }
+
+  @Input('ariaTextBeginningOfExpandableContent') ariaTextBeginningOfExpandableContent: string;
```
Shijir, line 264: Why do we need another input property here? Users can update these default strings through CommonStrings, right?

src/clr-core/common/services/common-strings.default.ts
```diff
@@ -71,4 +71,9 @@ export const commonStringsDefault: ClrCommonStrings = {
   //Vertical Nav
   verticalNavToggle: 'Toggle vertical navigation',
   verticalNavGroupToggle: 'Toggle vertical navigation group',
+  // Datagrid expandable rows
+  beginningOf: 'Beginning of',
```
Shijir, line 75: `beginningOf` and `endOf` of ...? They need to be more descriptive.

src/clr-angular/data/datagrid/datagrid-row.html
```diff
@@ -20,7 +20,11 @@
     depending on whether the details replace the row or not.
 -->
 <ng-template #detail>
-  <ng-content select="clr-dg-row-detail"></ng-content>
+  <div [id]="expandableId">
```
Shijir, line 23: Why is this id not on the `clr-dg-row-detail`? We are creating an additional wrapper div just to add this id. Plus, we should try to isolate these expanded row changes within its own component.

src/clr-angular/data/datagrid/datagrid-row.ts
bdryanovski: You could have multiple datagrids with deferent purposes.

Datagrid A
  - Begining of User details content
  - Content
  - End of User details content

Datagrid B
 - Begining of invoice
 - Content
 - End of invoice

src/clr-core/common/services/common-strings.default.ts
bdryanovski: Yep, you are right about that. Will update them

src/clr-angular/data/datagrid/datagrid-row.html
bdryanovski: It's a bit of unwanted hack to make the `aria-control` work. The id is for `[attr.aria-controls]="expandableId"` few lines below. 

`aria-controls` require to have ID to witch to link itself. So this is the only way I found to link them both. If `clr-dg-row-detail` create it own ID, it will have to pass it to the parent. If we create ID from the parent we will have to find a way to pass it below. Or we will have to create service to track this. All the solutions are a bit overkill for something that is not even supported on more than one SR. 

If we decide to drop the `aria-control` that I still think is not well supported and is adding unneeded changes we could drop the `<div id="expandableId"></div>` and the additional attribute to the button below.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: I get that there could be multiple datagrids. But I don't see why they don't would want to customize these default strings each one of them? If they need to do that, they could interject the common string service at their component level.

src/clr-angular/data/datagrid/datagrid-row.html
Shijir: I understand you need `aria-control` as well as an ID for that. What I am saying is that we are creating additional wrappers. And really this ID belongs to the row detail. What you need is uniq id provider that is shared between datagrid row and its row detail components.

src/clr-angular/data/datagrid/datagrid-row.ts
gnomeontherun: This is part of the a11y ask, that the messages are too generic and may need to be tailored to each row or datagrid. This is where we're really hitting the limits with what we can do with a11y.

bdryanovski: Having a new instance of the `ClrCommonStrings` won't let you change the value row-by-row also if you want to do it datagrid-by-datagrid you will need to create a wrapper for them. 

This way you could have better control.

src/clr-angular/data/datagrid/datagrid-row.html
bdryanovski: I don't see a reason to have a NEW provider just to handle this case. `aria-control` is not a well-supported attribute and it's not worth the additional complexity of a provider. 

And when it comes to providing and accessing the ID is all done in one place. It's clearly visible when the id is pointing.

src/clr-angular/data/datagrid/datagrid-row.ts
Shijir: Why do you expect users to want to have custom messages in each row or datagrid? Having different messages on each row would defeat its original purpose. Rows should have consistent header and footer messages to be helpful. You can have a wrapper component or simply place a directive that injects their own custom common strings. 

But besides that, we need to remember we are adding this fix to "deprecated" `clr-dg-row-detail`. So we should really try to keep its changes within itself instead adding a new API property to `clr-dg-row`. We are just expanding the area of the potential breaking change if we ever decides to delete the row detail.

src/clr-angular/data/datagrid/datagrid-row.html
Shijir: But you are complicating the template structure by adding more wrappers just to place that one attribute on them. And that's not even added in `clr-dg-row-detail`'s own template. Keep these changes where they belong to.

I don't see how injecting unique id provider would introduce more complexity. You could also just add a unique id property to some service that both `clr-dg-row` and `clr-dg-row-detail` share.

bdryanovski: Can you provide some code examples on how this has to be done? I don't see a common service between both components?

The changes are located inside the template for `clr-dg-row-details` so when and if this is removed they will go away as well.

But if you could provide a better solution to that I'm ok with it.

src/clr-angular/data/datagrid/datagrid-row.ts
bdryanovski: As mention above this is an a11y request

tests/tslint/dist/noBarrelImportsRule.js
```diff
@@ -0,0 +1,85 @@
+"use strict";
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
```
gnomeontherun, line 4: Still need to remove this file.

bdryanovski: yap, I got this from the last merge with the master branch, but will address it on my next changes if there any

gnomeontherun:
@chlane does this work for a11y needs?

clane:
I've always recommended using based on its documented purpose which is to identify the element (or elements) whose contents or presence are controlled by the current element (https://www.w3.org/TR/wai-aria-1.1/#aria-controls).

But the ARIA spec does not provide expected behaviors from assistive technologies which means that we really don't know what to look for when testing and I've never heard a screen reader announce a relationship due to it as of yet. 

I recommend to use it for the following reasons:

1) It is a valid ARIA 1.1  property (not deprecated)
2) It has potential to be useful for interactions like trees that control the display adjacent content in the future
3) It is used in ARIA 1.1 examples like https://www.w3.org/TR/wai-aria-practices-1.1/examples/carousel/carousel-1.html

For now we would just test for its presence and make sure the id references are valid.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
```diff
@@ -264,4 +266,20 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
+    );
+  }
+
+  @Input('ariaTextEndOfExpandableContent') ariaTextEndOfExpandableContent: string;
+  public get endOfExpandableContent() {
+    return (
+      this.ariaTextEndOfExpandableContent ||
+      `${this.commonStrings.keys.dategridExpandableEndOf} ${this.commonStrings.keys.dategridExpandableRowContent}`
+    );
+  }
```
Shijir, line 284: I still don't understand why these input properties are being added to `clr-dg-row` instead of `clr-dg-row-detail`?

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row-detail.ts
```diff
@@ -57,4 +64,20 @@ export class ClrDatagridRowDetail implements AfterContentInit, OnDestroy {
   ngOnDestroy() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
   }
+
+  @Input('clrRowDetailBeginningAriaText') ariaTextBeginningOfExpandableContent: string;
+  public get beginningOfExpandableContent() {
+    return (
+      this.ariaTextBeginningOfExpandableContent ||
```
Shijir, line 71: nit: these common string property names should be aligned with their corresponding input property names too.

gnomeontherun:
Let's get this closed out.

PR #4170, opened 2020-01-07
fix: core icon performance and a11y

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

### 1. Improved icon demo performance
  - Renders demo page (416 icons) around ~630-750ms compared to 1.1s - 1.2s on docs icon demo page built in prod mode on 2018 macbook pro.
  - Remove use of getters and replace with form value changes observable. This prevents recalculation for each icon during each change detection cycle.
  - Load icons immediately and create index object inline to prevent multiple renders/change detection cycles when angular component is created.
  - Use CSS to show and hide icons to prevent large amount of components being created/destroyed when filtering/searching
  - Remove table layout to reduce number of DOM nodes improving render times

### 2. Improved icon element render performance
  - Single cwc-icon uses ~0.64ms of JS execution time vs current clr-icon at ~1.5ms (2018 macbook pro)
  - remove unnecessary id reflection, ids for a11y labeling are internal and unique so syncing was not needed for a11y
  - only write to svg and create SR label if title is set
  - default svg template to have `role="img"` instead of writing on creation
  - remove unnecessary `<div>` wrapper to reduce DOM elements

### 3. Fix a11y behavior with screen readers.
  - Icons default decorative images unless otherwise set with `title`. When set with title a SR only message will be rendered with title value. Icon should be described by the context of usage such as text within a button or a `aria-label` on a button. Only if icon is represented as stand alone content is the `title` needed. https://clarity.design/icons/accessibility 

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/icon-shapes/icon.element.spec.ts
+++ b/src/clr-core/icon-shapes/icon.element.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -14,14 +14,6 @@ const testIcon = renderIcon('test');
 describe('icon element', () => {
   let testElement: HTMLElement;
   let component: CwcIcon;
-  let idcomponent: CwcIcon;
-  let updaterObject: CwcIcon;
-  let updaterValue: string;
-
-  function updateIconSizeMock(obj: CwcIcon, val: string) {
-    updaterObject = obj;
-    updaterValue = val;
-  }
 
   beforeAll(() => {
     ClarityIcons.add({ testing: testIcon });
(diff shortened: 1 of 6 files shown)


src/clr-core/icon-shapes/icon.element.spec.ts
```diff
@@ -103,43 +90,17 @@ describe('icon element', () => {
...
-    });
 
-    it('should update id if it is changed', async () => {
-      const newId = 'kthxbye';
-      await componentIsStable(idcomponent);
-      idcomponent.setAttribute('id', newId);
-      await componentIsStable(idcomponent);
-      expect(idcomponent.shadowRoot.innerHTML.includes('id="aria-' + newId)).toBe(true);
-      expect(idcomponent.shadowRoot.innerHTML.includes('aria-labelledby="aria-' + newId)).toBe(true);
+      expect(component.shadowRoot.innerHTML.includes('id="aria-cwc-icon-')).toBe(true);
```
coryrylan: TODO: add id to check to be more thorough

src/dev-core/src/app/icon-sets/icon-sets.demo.html
```diff
@@ -1,133 +1,119 @@
...
+                      <input type="checkbox" formControlName="isInverse">
+                      Inverse
+                  </label>
+              </td>
+          </tr>
+      </tbody>
+  </table>
+  <p>
+      <label><b>Decoration:</b></label>
+      <label><input type="radio" value="none" formControlName="decoration"> None</label>
```
coryrylan: Most of these changes were prettier reformat, the main change was the removal of the table in the list

src/dev-core/src/app/icon-sets/icon-sets.demo.html
```diff
@@ -1,133 +1,119 @@
...
-        margin: 0;
-    }
-    
-    .dc-icon-box table td {
-        text-align: center;
-    }
+  .dc-icon-set cwc-icon {
+      width: 1.8rem;
+      height: 1.8rem;
+      margin-bottom: 0.4rem;
```
mathisscott: Using the `div`s and `p`s the way they are being used makes the space between the icon and the label blow out. Plz remove this `margin-bottom` declaration so that the layout remains consistent.

PR #4172, opened 2020-01-08
fix(a11y): service ClrCommonStringsService initialize twice at run-time

### LINKED ISSUE #4018: Date Picker button title not reflecting internationalization string override
## Describe the bug

Date Picker button title not reflecting internationalization string override.
Refer stackblitz link for reproduction
image

## How to reproduce
https://stackblitz.com/edit/clarity-light-theme-v2-gfvji8

## Expected behavior

It should work as documented in the guide

## Versions

**App**

* Angular: 8.2.x
* Clarity: 2.3.1

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser chrome

bdryanovski:
There are two ClrCommonStringsService instances, but you could overwrite only one, so there is two active translation that leads to different localization strings across the app. 

Close #4018

Target is V2 branch, also need to be merged into master/v3/v1

### DIFF
--- a/src/clr-angular/forms/datepicker/date-container.ts
+++ b/src/clr-angular/forms/datepicker/date-container.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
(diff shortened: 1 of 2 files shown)


bdryanovski:
Closing in favor of PR with master/v3 target branch - less PR will be needed for backporting

Issue #3608, opened 2019-07-10
Add filter property to clr-tree-node


ArturGudiev:
## Summary

Add possibility to filter tree nodes in clarity recursive tree. In simple case I would use *ngIf  in wrapped ng-container directive to filter some tree nodes but in this case it's impossible because nothing is shown.

* What is the change?
Add property to clr-tree-node to filter nodes. 

* Why should it go in Clarity?
Now it is impossible to filter tree nodes in recursive trees in templates. 

* Does this change impact existing behaviors? If so how?
No, it doesn't. 

* If this change introduces a new behavior, is this behavior accessible?
Yes. 

## Use case
I have a tree. I would like to filter it by condition in template without extra fields and methods in component. 

## Examples
see https://stackoverflow.com/questions/56971793/use-ngif-to-filter-tree-nodes-in-clarity-recursive-tree

## Workarounds

Create another field in template (e.g. displayedNodes) and filter existing nodes there.

Jinnie:
Related (though no fully equivalent) to https://github.com/vmware/clarity/issues/3603

gnomeontherun:
The feature request here has been captured into our list and we’re going to take it into consideration as we develop Clarity Core capabilities. In an effort to clean up our backlog and focus our attention, I’m going to close this as captured in our feature requests. Please follow our development and releases to see when we release relevant components to make this possible. Future feature requests can be made in our [GitHub Discussions](https://github.com/vmware/clarity/discussions).

Issue #1821, opened 2017-12-29
Tree View Drag and Drop (design spec)


gracesnoh:
### Description
Items in tree view can be rearranged using drag and drop

![dox-basic](https://user-images.githubusercontent.com/8237984/34426768-d1126e6c-ebee-11e7-9122-47e82cd8bd10.gif)

### Usage
To reorder and rearrange nodes in a tree view, resulting in a change in arrangement or nesting

### Behavior
When a tree node is dragged, a “copy” of the tree node will be dragged with the cursor. The original node will be greyed out, indicating its previous position.
![node-copy](https://user-images.githubusercontent.com/8237984/34426457-8d29de68-ebeb-11e7-9cfd-0256cb16cae4.png)

A tree node is draggable from anywhere inside of its container. If the tree node can be expanded/collapsed, selected or checked, those elements remain clickable but are also draggable.
![clickable](https://user-images.githubusercontent.com/8237984/34426543-a2432d3a-ebec-11e7-9e65-bd16721875f3.png)

There are two types of drop targets: position and parent targets. When the midpoint of the dragged node enters the bottom or top half of a node, a position drop target will be triggered in between. However, when the midpoint enters the middle 18px of a parent node, a parent drop target will be triggered.

![drop-targets](https://user-images.githubusercontent.com/8237984/34426591-0bce1d00-ebed-11e7-8088-1b49736105b0.png)

1.   **Position targets** are lines with a circle end that indicate the new position of the node. The
     result is a change in the order of nodes.
2.  **Parent targets** are outlined with a border that indicate the new parent of the node.
     The result is a node is nested in another node.

![drop-targets-2](https://user-images.githubusercontent.com/8237984/34426618-3f088dfe-ebed-11e7-8fd6-c5044b18bca9.png)

### Nesting nodes
When a tree node is dropped inside of a parent node, the node is automatically the last child node inside of the parent node.

![dox-nested-auto](https://user-images.githubusercontent.com/8237984/34426732-899d0e20-ebee-11e7-906e-b2eac1db915b.gif)

When users hover a dragged node over a closed parent node for at least 2 seconds, the parent node will open automatically to reveal its child nodes. This allows the user to specify the position of the dragged node among the child nodes.

![dox-nested-custom](https://user-images.githubusercontent.com/8237984/34426786-01a9f194-ebef-11e7-9d2d-2ad48d02d795.gif)

[...]

PR #4204, opened 2020-01-14
Topic/tree view dnd 1821

derkoe:
This PR shows a simple example implementation of the design spec in #1821

There are still some issues:
* `clr-tree-node` with `*ngIf` is not working. This is needed when you only want to make some nodes draggable (e.g. only root nodes or only leaves). See #3608 and https://stackoverflow.com/questions/56971793/use-ngif-to-filter-tree-nodes-in-clarity-recursive-tree
* Drop areas of nested nodes do not work properly - it is not clear where the item is dropped.

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #1821

* [ ] Yes
* [x] No

### DIFF
--- a/src/dev/src/app/tree-view/drag-and-drop-tree/drag-and-drop-tree.html
+++ b/src/dev/src/app/tree-view/drag-and-drop-tree/drag-and-drop-tree.html
@@ -1,0 +1,13 @@
+<h4>Drag and Drop (recursive)</h4>
+<div class="clr-example">
+  <clr-tree>
+    <ng-container *clrRecursiveFor="let item of tree; getChildren: getChildren">
+      <div clrDroppable (clrDrop)="onDrop($event.dragDataTransfer, item)"></div>
+      <clr-tree-node [clrDraggable]="item" [(clrExpanded)]="item.expanded">
+        <clr-icon [attr.shape]="getShape(item)"></clr-icon>
+        {{item.name}}
+      </clr-tree-node>
+    </ng-container>
+    <div clrDroppable tabIndex="-1" (clrDrop)="onDrop($event.dragDataTransfer)"></div>
+  </clr-tree>
+</div>
--- a/src/clr-angular/data/tree-view/_tree-view.clarity.scss
+++ b/src/clr-angular/data/tree-view/_tree-view.clarity.scss
@@ -5,6 +5,46 @@
 @import './properties.tree-view';
 
 @include exports('tree-view.clarity') {
+  clr-tree .droppable {
+    position: absolute;
+    display: none;
+    height: 18px;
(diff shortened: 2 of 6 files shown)


bdryanovski:
Hey @derkoe, thanks for the contribution but .. could you fork from v2 branch or master depending on the target version - right now there 153 commits and it's hard to see what exactly you are fixing. 

Thanks!

derkoe:
@bdryanovski actually there is only 1 commit in this PR - the others are only to catch the branch up with master. We did the PR against this branch because @gnomeontherun suggested that in #1821

> @d-m-s Alright, so here is what we can do to get started.
> I have created a feature branch for us to work on. https://github.com/vmware/clarity/tree/topic/tree-view-dnd-1821 We should use this to submit smaller PRs against so we can build it up.

gnomeontherun:
To help, I just pushed an update to the tree view branch that makes it aligned with master. @derkoe can you rebase your commit onto the updated branch so we get a single commit here? Thanks!

src/dev/src/app/tree-view/drag-and-drop-tree/drag-and-drop-tree.html
```diff
@@ -0,0 +1,13 @@
+<h4>Drag and Drop (recursive)</h4>
```
Shijir: Hi @derkoe, thank you for submitting this PR. This looks like a possible workaround to apply DnD to enable reorderable nodes in a recursive tree.

Unfortunately, I don't think we could include this PR as a feature in Clarity. First of all, we cannot recommend this API as this is adding extra div elements to place `clrDroppable`.  Also, looking at the code, it doesn't look like it followed our coding guidelines as specified in [here](https://github.com/vmware/clarity/blob/master/docs/CODING_GUIDELINES_ANGULAR.md). On top of that, as you mentioned yourself, this doesn't work reliably in some instances. To make sure its reliability and robustness, we need to write tests for it.

Shijir:
Recordable tree nodes is a prominent feature that would require a significant effort. As such, we will need continual communication and overseeing from our accessibility experts and UX designers on its implementation. Finding a scalable and intuitive API for it alone (ideally, that works for both declarative and recursive trees) is a challenge and needs research on it. As regards to this specific PR, I don't think this solution is robust and scalable enough to cover most use cases of recordable tree nodes. We have to be careful about what we recommend as a feature or API because once they are in, they will be really hard to take them out unless we are willing to introduce more breaking changes to remove them.

Shijir:
Closing this PR for now based on the reasonings I mentioned above.

PR #4207, opened 2020-01-15
fix: core ts build

coryrylan:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [x] CI related changes

## What is the new behavior?
This PR changes core to be compiled from angular CLI to use tsc directly. This is due to the CLI breaking the output of the build when making certain angular specific optimizations.

Using tsc directly the test app now uses the same output that we publish to NPM making testing more consistent. Due to the change the watch for core is slightly slower by 1-2s and the full build for core is faster from 50s to 20s. Core now only relies on the CLI to run the unit tests. In a future PR we can separate this if needed.

Core now uses [TypeScript Project references](https://www.typescriptlang.org/docs/handbook/project-references.html). The Angular CLI automated this under the hood with its entrypoint system. This is set up by adding a tsconfig to each component to treat it as its own independent sub project. This ensures dependencies are not crossed in a way that would break treeshaking and keeps each component as its own isolated entrypoint module. With project references the core ts build is locally cached which can bring the full core ts build from 20s down to about 1s.

* [ ] Yes
* [x] No


package.json
```diff
@@ -4,14 +4,14 @@
...
   "scripts": {
     "ng": "ng",
-    "clean": "del ./dist",
+    "clean": "del dist",
     "start": "npm run angular:dev:start",
-    "build": "npm-run-all build:ci website:build build:clean",
-    "build:ci": "npm-run-all --parallel format lint clean && npm-run-all build:libs build:apps test",
-    "build:libs": "npm-run-all ui:build icons:build core:build angular:build schematics:build angular:golden",
+    "build": "npm-run-all build:ci website:build",
+    "build:ci": "npm-run-all clean && npm-run-all --parallel format lint && npm-run-all build:libs build:apps build:clean test",
```
hippee-lee: Is the reason for the multiple `npm-run-all`'s that these commands need to run sequentially? 
There is a way to specify that with npm-run-all
`npm-run-all -s` || `run-s` (shorthand)

coryrylan: When we get to `build:libs build:apps build:clean` they have to run in sequentially since the apps depend on the final build output of the libraries. The tests could be ran in parallel with the previous command but did not have any performance benefit since `build:libs build:apps` internally are each building their libs and apps in parallel as well.
I'll clean up the script with the shortcuts

coryrylan: Edit: going to clean up the script in my next PR so this can get merged to fix the other issue

PR #4214, opened 2020-01-17
feat(vertical-nav): add aria-current to active link

### LINKED ISSUE #3576: Vertical nav active state not available to screen readers
## Describe the bug

A white background for the active link is used to convey information that is not communicated to screen reader users. 

## How to reproduce

https://v1.clarity.design/vertical-nav/nav-groups/credit

- Using a screen reader, go to https://v1.clarity.design/vertical-nav/nav-groups/credit
- Read any of vertical nav items that has a background color of #fff
- Notice that the screen reader does not announce the meaning of the #fff background

## Expected behavior

Dermine what is meant by the #fff and communicate it textually. aria-elected and aria-current are potential properties that could be used. If these don't match the intent of the background color I recommend using aria-label to communicate the meaning of the #fff background.

If the intent of the this background image is to convey the active state of the link I recommend using the active CSS psuedo-class seen below.

a:active { }

Review the definition of the active pseudo-class from https://www.w3.org/TR/CSS2/selector.html#dynamic-pseudo-classes below and decide if we are trying to communicate the active state of the link with the #fff background color.

The :active pseudo-class applies while an element is being activated by the user. For example, between the times the user presses the mouse button and releases it.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Version [e.g. 22]

## Additional notes

bdryanovski:
Provide vertical navigation with `arai-current="page"` when there is `routerLinkActive` set.

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [] Documentation content changes

Issue Number: #3576 

* [ ] Yes
* [ ] No

## Other information

Close #3576

### DIFF
--- a/src/website/src/app/documentation/demos/vertical-nav/vertical-nav.demo.html
+++ b/src/website/src/app/documentation/demos/vertical-nav/vertical-nav.demo.html
@@ -55,7 +55,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6 class="clrweb-DoxMedia-do-dont">Don't</h6>
                             <p>
-                                Use vertical navigation with a header and subnav. Having too many places 
+                                Use vertical navigation with a header and subnav. Having too many places
                                 for navigation is confusing for users.
                             </p>
                         </div>
@@ -66,8 +66,8 @@
             <h3>Layout</h3>
 
             <p>
-                Vertical navigation has a few layout options including dividers and section headers. Active 
-                links are displayed with a white background. Touch targets are larger for easier interaction 
+                Vertical navigation has a few layout options including dividers and section headers. Active
+                links are displayed with a white background. Touch targets are larger for easier interaction
                 and span the entire width of the navigation, similar to the space of the active link indicator.
             </p>
 
@@ -98,7 +98,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6>Dividers</h6>
                             <div>
-                                Used to separate logical clusters of navigation items. Best used when you 
+                                Used to separate logical clusters of navigation items. Best used when you
                                 have two or more links in a cluster.
                             </div>
                         </div>
@@ -115,7 +115,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6>Dividers & Headers</h6>
                             <div>
-                                Used to separate logical clusters of navigation items. Best used when you 
+                                Used to separate logical clusters of navigation items. Best used when you
                                 have two or more links in a cluster.
                             </div>
                         </div>
@@ -126,8 +126,8 @@
             <h3>Icons</h3>
 
             <p>
-                Icons should be placed to the left of the link label. Active links will display a blue icon. 
-                Make sure your icons are distinctive and easily recognizable. Touch targets include the 
+                Icons should be placed to the left of the link label. Active links will display a blue icon.
+                Make sure your icons are distinctive and easily recognizable. Touch targets include the
                 entire width of the navigation including the icon.
             </p>
 
@@ -168,9 +168,9 @@
             <h3>Hierarchy</h3>
 
             <p>
-                Hierarchy is used to show parent-child relationship between links. If a child link is active 
-                and the parent item is collapsed, the parent will display as active. When expanded, the active 
-                indicator will display on the child link. Clicking on text, carets or icons will expand and 
+                Hierarchy is used to show parent-child relationship between links. If a child link is active
+                and the parent item is collapsed, the parent will display as active. When expanded, the active
+                indicator will display on the child link. Clicking on text, carets or icons will expand and
                 collapse the parent item.
             </p>
 
@@ -185,7 +185,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6>Basic</h6>
                             <div>
-                                Top level hierarchy items are semibold font weight and have a caret to the 
+                                Top level hierarchy items are semibold font weight and have a caret to the
                                 right. Child links are normal font weight.
                             </div>
                         </div>
@@ -202,7 +202,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6>Icons</h6>
                             <div>
-                                Icons are only applied to the top level navigation items. They are not applied 
+                                Icons are only applied to the top level navigation items. They are not applied
                                 to child links.
                             </div>
                         </div>
@@ -220,7 +220,7 @@
                         <div class="clrweb-DoxMedia-text">
                             <h6>Mixed</h6>
                             <div>
-                                Top level hierarchy items without child links are semibold. Clicking on one 
+                                Top level hierarchy items without child links are semibold. Clicking on one
                                 will navigate a user to a page.
                             </div>
                         </div>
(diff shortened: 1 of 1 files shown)


src/clr-angular/layout/vertical-nav/vertical-nav-group.ts
```diff
@@ -170,4 +182,17 @@ export class ClrVerticalNavGroup implements AfterContentInit, OnDestroy {
     this._subscriptions.forEach((sub: Subscription) => sub.unsubscribe());
     this._navGroupRegistrationService.unregisterNavGroup();
   }
+
+  /** get all router-links so we could find the active one */
+  @ContentChildren(RouterLinkActive) routerLinks;
+
+  ngAfterContentChecked(): void {
+    if (this.routerLinks.length) {
```
gnomeontherun: This needs a check if `this.routerLinks` is defined first.

bdryanovski:
Will remove the`@angular/router` dependency and will update the documentation on how to deal with it by hand

src/clr-angular/layout/vertical-nav/vertical-nav-group.ts
```diff
@@ -170,4 +182,17 @@ export class ClrVerticalNavGroup implements AfterContentInit, OnDestroy {
...
+      this.routerLinks.forEach(({ isActive, element }) => {
```
Shijir: This is not the right way to add `aria-current="page"` to the active router link. What we need to here is to inject `RouterLinkActive` into `ClrVerticalNavLink` and access its `.isActive` property, e.g:

```ts
@Component({
  selector: '[clrVerticalNavLink]',
  template: `
        ...
    `,
  host: { '[attr.aria-current]': 'routerLinkActive.isActive? "page": null' },
})
export class ClrVerticalNavLink {
  constructor(..., public routerLinkActive: RouterLinkActive) {}

}

Comment (bdryanovski) [reply src/clr-angular/layout/vertical-nav/vertical-nav-group.ts]: @Shijir  your solution is looking a lot nicer, but I had a talk with @gnomeontherun and decide that this may require all other approaches on the issue.  And then we could start thinking about how to provide automatically. Right now we are covering only one specific way and we are not sure is everyone using `@angular/router`. So in a way, this is a bit limiting solution. 

Also, I got this idea for adding just a directive to handle this for us.

```html
<a clrAddAriaCurrent>...</a>
```

And based on internal rules clrAddAriaCurrent will add or remove `aria-current` for example

  *   There is `RouterLinkActive` and it is set to `active`
  *   Have class `active` 
  *   Other rules

So, for now, I will just add documentation on the issue and how to handle it by hand.

bdryanovski:
When started writing the documentation and going over the examples and our internal guideline on how to use it. I found that I may still make it automatic without adding additional dependency like the router.

I notice that we use `ClrVerticalNavLink` to wrap the content inside the `nav-text` and `nav-link` class. So I could just check for `active` class on the `HTMLElement`. This is the same class that the router will add, also the same class that we use to change the highlight into the navigation. 

So I think that this will cover all the component cases much better than depending on the router component.

kalpeshhpatel:
I think this should be handled by consumer applications and not the library.

This is how we are doing for our application:

```
    <a clrVerticalNavLink     
        routerLink="./about"
        routerLinkActive="active"
        #rla="routerLinkActive"
        [attr.aria-current]="rla.isActive ? 'page' : null"
      </a>
```

Let me know if this should be another way around.

Thanks

gnomeontherun:
@kalpeshhpatel  We wanted to help where we can, but we don't assume the Angular Router is present. Your solution is good but ultimately we can help automate this because its an extra layer of complexity for apps to manage.

bdryanovski:
The latest update on the PR. 

My initial idea was to use directive to add/remove `aria-current="page"`. But the issue is that `.active` class at most of the time is added dynamically, from the Router or from external code. So we could not use the `a.active,[routerLinkActive ]` (simplify) selector for the directive and have to monitor the DOM for changes. So this is not an option for the moment. 

Right now I'm piggybacking on the update lifecycle for `ClrVerticalNavLink` and check for the class Because of the `.active` class is the one we based our style and is the one that the router will use we could be sure that it will work 99% of the time. 

We are not providing a static code for this case, so no documentation is needed.

src/clr-angular/layout/vertical-nav/vertical-nav-link.ts
```diff
@@ -1,28 +1,35 @@
...
 
   @HostListener('click')
   public expandParentNavGroup(): void {
     if (this._navGroupService) {
       this._navGroupService.expand();
     }
   }
+
+  get setAriaCurrent() {
+    return this.el.nativeElement.classList.contains('active') ? 'page' : null;
```
gnomeontherun: Two issues, this still runs with every change detection cycle and it isn't guarded with a platform check. If we can't find a better option, then we should just change this to document that applications should add the `aria-current` value directly.

Shijir: It won't actually run every change detection cycle. We are accessing DOM api so it's idle, and it won't respond even if the `active` class is toggled. 

@bdryanovski another small issue is in the naming of this getter property. You have `getter` but the name is `setAriaCurrent` which usually suggests a method from API perspective :D. We could rename this property to `isActive`.

bdryanovski: @Shijir you are right, the name and the implementation changed over the time, so the name need to be changed as well. I will changed.

@gnomeontherun you are right for the platform check, will add it. When it comes to implementing it from outside it will be more heavy on the performance if you have to do DOM selection or going over it with reference. Class list is updated with the related component so this is done no matter what. 

It could be done in the same way with @HostBinding but decide to use the host syntax just to be a lot more clear and compact. 

One issue with that is that .nav-link and .active are added dynamically and angular is having hard time tracking them, without making DOM queries, so I find this to be the most effective way to do it. 

We could go with documentation only, but I don’t think that this is adding a lot of performance penalty.

gnomeontherun: It is running every change detection cycle. Nothing about the binding is linked to the `active` class so Angular will check each time. When you click a button it fires or add an interval and see that it fires the binding on every interval. 

Host bindings are always checked, so if their values have to be looked up, that lookup happens with each change detection. The downstream change (setting of aria-current) only changes if the lookup returns a different value. Regardless if a change is found, the DOM is still hit for _every_ vertical nav link and _every_ change detection cycle. 

I hadn't thought through what a better option would be, but this is clearly a blocker as is. Based on this, we should abandon handling it internally and document how applications should do this. The other concern is that vertical-nav _may_ be used in ways that aren't pages so the aria-current implementation might not make sense if the application is using it like a tabbed interface. They shouldn't do this probably, but it is a concern. Either way, we've tried several things and they all suffer from performance drains that we can't accept, so let's go to documentation and close this out.

bdryanovski: If this is the case, will revert the changes for the ClrVericalNavLink and will just update the documentation.

chlane:
IMO the "nav" in vertical-nav implies this component will always be used in routing scenarios where aria-current="page" is the best choice. But aria-current is flexible and can also represent current steps, locations, dates and times. It sounds like we have to make the consumer add this attribute which will work as long as we provide documentation similar to our approach with the dropdown component.

Issue #3562, opened 2019-07-02
Form validation errors are not read by some screen readers


gnomeontherun:
## Describe the bug

Error messages are not announced by some screen reader/browser combinations. This is because the current implementation does not dynamically write content to an existing live region but instead adds the live region and the message it contains to the DOM all at once. 

## How to reproduce

https://clarity.design/documentation/stepper

- Using Firefox/NVDA or IE11/NVDA or EDGE/JAWS, activate the 'Stepper Demos' link under the 'Stepper' heading
- Trigger an error message by using the Tab key to focus the 'First Name' field and then press the Tab key again to move focus to the next control in the Tab order.
- Observe that an error message appears below the 'First Name' field and after it does, the screen reader does not announce it.

## Expected behavior

The error should be announced.

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

The aria-live property indicates that an element will be updated. So it should be in the DOM when the page loads and subsequently updated.

When the form is visible on the page, the live region should be on the page as well and empty as seen below.

<clr-control-error aria-live="polite">
</clr-control-error>

Then, as errors occur, the live region should be populated as seen below.

<clr-control-error aria-live="polite">
Error: The first name field does not have a valid entry
</clr-control-error>

See https://www.w3.org/TR/wai-aria-1.1/#aria-live

gnomeontherun:
@coryrylan Can we close this as done then?

coryrylan:
I don't think so, I think this might be still an issue and needs to be verified.

clane:
Updated recommendation - use an ARIA Live Region to inform the user the application is in an error state and use aria-describedby to programmatically associate the inline error messages with their respective fields. Than focus should be set to to the first field in error.  See the example below.

```
<div aria-live="polite">There are errors ...</div>

<label for="nameField">name</name>
<input id="name" aria-describedby="inlineError1"/>
<div id="inlineError1">The name field should not contain integers</div>
```

To recap, the live region is used to perform a status update that tells the user the form is in an error state. Then focus is set to the first field in error. Since aria-describedby is used to associate the errors with their respective input fields, screen readers should announce the errors when the fields gain focus.

bdryanovski:
[...]

PR #4221, opened 2020-01-21
fix(a11y): add aria-describedby for clrInput

bdryanovski:
Add `aria-describedby` to `clrInput` in the context of `clr-input-container` to provide additional description when there is a error text.

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

## Other information

Close: #3562

### DIFF
--- a/src/clr-angular/forms/input/input.spec.ts
+++ b/src/clr-angular/forms/input/input.spec.ts
@@ -1,12 +1,12 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import { Component } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 
-import { TemplateDrivenSpec, ControlStandaloneSpec, ReactiveSpec } from '../tests/control.spec';
+import { TemplateDrivenSpec, ControlStandaloneSpec, ReactiveSpec, AccessibilitySpec } from '../tests/control.spec';
 import { ClrInputContainer } from './input-container';
 import { ClrInput } from './input';
 
@@ -26,6 +26,17 @@ class TemplateDrivenTest {}
 
 @Component({
   template: `
+    <clr-input-container>
+       <input clrInput name="model" [aria-describedby]="describedby" class="test-class" [(ngModel)]="model" />
+    <clr-input-container>
+    `,
+})
+class InputContainerTest {
+  describedby: string;
+}
+
+@Component({
+  template: `
     <div [formGroup]="example">
        <input clrInput name="model" class="test-class" formControlName="model" />
     </div>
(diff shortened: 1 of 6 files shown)


src/clr-angular/forms/input/input.spec.ts
```diff
@@ -24,6 +24,17 @@ class StandaloneUseTest {}
...
 
+@Component({
+  template: `
+    <clr-input-container>
+       <input clrInput name="model" [aria-describedby]="describedby" class="test-class" [(ngModel)]="model" />
+    <clr-input-container>
+    `,
+})
+class InputContainerTest {
+  describedby: string;
```
gnomeontherun: What are you testing here with an empty string?

src/clr-angular/forms/input/input.ts
```diff
@@ -21,8 +28,24 @@ export class ClrInput extends WrappedFormControl<ClrInputContainer> {
...
+    el: ElementRef,
+    @Optional() public controlIdService: ControlIdService
   ) {
     super(vcr, ClrInputContainer, injector, control, renderer, el);
   }
+
+  /**
+   * Provide a way to overwrite aria-describedby property.
+   */
+  @Input('aria-describedby') userDescribedById: string;
```
gnomeontherun: This is going to cause problems with the logic in wrapped-control.

Shijir:
@bdryanovski haven't we solved #3562 with the AriaLive service? If the solution provided with the AriaLive service wasn't sufficient, what particular issue is this one is solving? Have we checked with @chlane on this solution?

bdryanovski:
It seems that my PR is no longer valid - the `aria-describedby` is applied as expected. The only issue that I find here is that AriaLiveService is not working with `clr-input-error` when is inside the `clr-input-wrapper` but this is separate bug #4222  

So I'm closing this one.

clane:
@Shijir I recommended using both a live region and aria-describedby together for error handling https://github.com/vmware/clarity/issues/3562. I recommend using aria-invalid as well. See my recommendations for error handling below.

After considering the sequential nature of screen reader usage and the potential for a high volume of error messages in our applications, we recommend the following.

Use a polite live region or role="status" to announce a status message that lists the fields in error e.g. "There are errors with the following fields: Memory, New Hard Disk, NEW SCSI controllere" for more information see https://www.w3.org/TR/wai-aria-1.1/#aria-live. Do not move focus to this element.
Use aria-label="error:" and role="img" with the error icons to make these icons be announced as "error:" to blind screen reader users.
Provide visible textual error messages following the error icon e.g. "memory allocation is too low".
Use aria-describedby (https://www.w3.org/TR/wai-aria-1.1/#aria-describedby) to programmatically associate the error icons and error messages with their respective fields. This will allow blind screen reader users to hear the name of a given input field, the word "error" and the respective error message together whenever they read a given input field.
Use aria-invalid (https://www.w3.org/TR/wai-aria-1.1/#aria-invalid) when the value entered by the user invalid or out-of-range. When the user attempts to submit data involving a field for which aria-required is true, authors MAY use the aria-invalid attribute to signal there is an error. However, if the user has not attempted to submit the form, authors SHOULD NOT set the aria-invalid attribute on required widgets simply because the user has not yet entered data.
If there is a single error, focus should be moved to that field
If there are multiple errors, focus should be moved to first field in error
If any controls need to be disabled, do so both visually and programmatically i.e,. use the disabled attributed
Provide a checkbox to allow blind screen reader users to disable the status messages if the application becomes too verbose
Provide configuration options for the status messages allowing blind screen reader users the ability to control the verbosity of the status message in cases where it may be too verbose

PR #4241, opened 2020-01-29
feat(alert): core alerts

jeeyun:

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/alert/alert-actions.element.scss
+++ b/src/clr-core/alert/alert-actions.element.scss
@@ -1,0 +1,8 @@
+@import './../styles/mixins/utils';
+
+:host {
+  flex: 0 0 auto;
+  white-space: nowrap;
+
+  // TODO: add style override for dropdown component inside alert-actions after cwc-dropdown is implemented
+}
--- a/src/clr-core/alert/alert-actions.element.spec.ts
+++ b/src/clr-core/alert/alert-actions.element.spec.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { CwcAlertActions } from '@clr/core/alert';
+import '@clr/core/alert';
+import { componentIsStable, createTestElement, removeTestElement, waitForComponent } from '@clr/core/test/utils';
+
+describe('alert-actions element', () => {
+  let testElement: HTMLElement;
+  let component: CwcAlertActions;
+  const placeholderText: string = 'Alert Text Placeholder';
+
+  beforeEach(async () => {
+    testElement = createTestElement();
+    testElement.innerHTML = `<cwc-alert-actions>${placeholderText}</cwc-alert-actions>`;
+
+    await waitForComponent('cwc-alert-actions');
+    component = testElement.querySelector<CwcAlertActions>('cwc-alert-actions');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  it('should create the component', async () => {
+    await componentIsStable(component);
+    expect(component.innerText).toBe(placeholderText);
+  });
+
+  it('should have a slot attribute of value `actions`', async () => {
+    await componentIsStable(component);
+    expect(component.hasAttribute('slot')).toBe(true);
+    expect(component.getAttribute('slot')).toEqual('actions');
+  });
+});
--- a/src/clr-core/alert/alert.base.ts
+++ b/src/clr-core/alert/alert.base.ts
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import '@clr/core/button';
+import { CwcButton } from '@clr/core/button';
+import {
+  CommonStringsService,
+  event,
+  EventEmitter,
+  property,
+  querySlotAll,
+  returnOrFallthrough,
+} from '@clr/core/common';
+import {
+  checkCircleIcon,
+  ClarityIcons,
+  exclamationCircleIcon,
+  exclamationTriangleIcon,
+  infoCircleIcon,
+  timesIcon,
+} from '@clr/core/icon-shapes';
+import { html, LitElement } from 'lit-element';
+
+const iconMap = {
+  info: {
+    shape: ClarityIcons.getIconNameFromShape(infoCircleIcon),
+    title: CommonStringsService.keys.info,
+  },
+  success: {
+    shape: ClarityIcons.getIconNameFromShape(checkCircleIcon),
+    title: CommonStringsService.keys.success,
+  },
+  warning: {
+    shape: ClarityIcons.getIconNameFromShape(exclamationTriangleIcon),
+    title: CommonStringsService.keys.warning,
+  },
+  danger: {
+    shape: ClarityIcons.getIconNameFromShape(exclamationCircleIcon),
+    title: CommonStringsService.keys.danger,
+  },
+};
+
+/**
+ * Base class for alerts. Contains properties and functions common to all alerts.
+ */
+// @dynamic
+export class CwcBaseAlert extends LitElement {
+  @event() private closedChange: EventEmitter<boolean>;
+
+  /** If false, the alert will not render the close button.  */
+  @property({ type: Boolean })
+  closable = true;
+
+  /** Sets the color of the alert from a predefined list of statuses */
+  @property({ type: String })
+  status: 'info' | 'success' | 'warning' | 'danger';
+
+  /** Sets the icon shape to be used to override the default icon. The application must import the icon shape. */
+  @property({ type: String })
+  iconShape = '';
+
+  /** Sets the title attribute for the icon. This may need to be set especially if the component is using custom icon shape.  */
+  @property({ type: String })
+  iconTitle = '';
+
+  @querySlotAll('cwc-button') private buttons: NodeListOf<CwcButton>;
+
+  get alertIconShape() {
+    /* 
+     * if the component's icon-shape attribute is set, that value is used.
+     * otherwise, we check for status attribute and set an icon shape that matches.
+     * if neither is supplied we default to info-circle icon.
+     */
+    return returnOrFallthrough(
+      [[this.iconShape, () => this.iconShape], [this.status, () => iconMap[this.status].shape]],
+      () => infoCircleIcon[0]
+    );
+  }
+
+  get alertIconTitle() {
+    /* 
+     * if the component's icon-title attribute is set, that value is used.
+     * otherwise, we check for status attribute and set an icon shape that matches.
+     * if neither is supplied we default to string for "Info".
+     */
+    return returnOrFallthrough(
+      [
+        [this.iconTitle, () => this.iconTitle],
+        [this.status, () => (iconMap[this.status] ? iconMap[this.status].title : '')],
+      ],
+      () => CommonStringsService.keys.info
+    );
+  }
+
+  updateButtons() {
+    // buttons inside alert (usually app-level) have the style of small inverse buttons
+    this.buttons.forEach(button => {
+      button.status = 'inverse';
+      button.size = 'sm';
+    });
+  }
+
+  render() {
+    return html`
+      <div class="alert-wrapper">
+        <div class="alert-item">
+          <div class="alert-icon-wrapper">
+            <cwc-icon class="alert-icon" shape="${this.alertIconShape}" title="${this.alertIconTitle}"></cwc-icon>
+          </div>
+          <slot></slot>
+          <slot name="actions" @slotchange="${this.updateButtons()}"></slot>
+        </div>
+      </div>
+      ${
+        this.closable
+          ? html`<button @click="${() => this.closeAlert()}" aria-label="${
+              CommonStringsService.keys.alertCloseButtonAriaLabel
+            }"
+          class="close" action="outline" icon><cwc-icon shape="times"></cwc-icon></button>`
+          : html``
+      }
+    `;
+  }
+
+  closeAlert() {
+    this.closedChange.emit(true);
+  }
+}
+
+ClarityIcons.addIcons(checkCircleIcon, infoCircleIcon, exclamationCircleIcon, exclamationTriangleIcon, timesIcon);
--- a/src/clr-core/alert/alert.element.ts
+++ b/src/clr-core/alert/alert.element.ts
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles, property, registerElementSafely } from '@clr/core/common';
+import {
+  checkCircleIcon,
+  ClarityIcons,
+  exclamationCircleIcon,
+  exclamationTriangleIcon,
+  infoCircleIcon,
+  timesIcon,
+} from '@clr/core/icon-shapes';
+import { CwcBaseAlert } from './alert.base';
+import { styles } from './alert.element.css';
+
+/**
+ * Web component alerts.
+ *
+ * ```typescript
+ * import '@clr/core/alert';
+ * ```
+ *
+ * ```html
+ * <cwc-alert>
+ *    <cwc-alert-content>This is an alert.</cwc-alert-content>
+ * </cwc-alert>
+ * ```
+ *
+ * @noInheritDoc
+ * @beta 3.0
+ * @element cwc-alert
+ * @slot default - Content slot for inside the alert. Usually will contain at least a <cwc-alert-content> component.
+ * @cssprop --color
+ * @cssprop --background
+ * @cssprop --border-radius
+ * @cssprop --border-color
+ * @cssprop --icon-color
+ * @cssprop --close-icon-color
+ * @cssprop --close-icon-color-hover
+ */
+// @dynamic
+export class CwcAlert extends CwcBaseAlert {
+  /** Sets the overall height and width of the alert and icon based on value */
+  @property({ type: String })
+  size: 'default' | 'sm';
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+}
+
+registerElementSafely('cwc-alert', CwcAlert);
+
+ClarityIcons.addIcons(checkCircleIcon, infoCircleIcon, exclamationCircleIcon, exclamationTriangleIcon, timesIcon);
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'cwc-alert': CwcAlert;
+  }
+}
--- a/src/clr-core/alert/app-alert.element.ts
+++ b/src/clr-core/alert/app-alert.element.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles, property, registerElementSafely } from '@clr/core/common';
+import {
+  checkCircleIcon,
+  ClarityIcons,
+  exclamationCircleIcon,
+  exclamationTriangleIcon,
+  infoCircleIcon,
+  timesIcon,
+} from '@clr/core/icon-shapes';
+import { CwcBaseAlert } from './alert.base';
+import { styles } from './app-alert.element.css';
+/**
+ * Web component alerts.
+ *
+ * ```typescript
+ * import '@clr/core/alert';
+ * ```
+ *
+ * ```html
+ * <cwc-app-alert>
+ *    <cwc-alert-content>This is an alert.</cwc-alert-content>
+ * </cwc-app-alert>
+ * ```
+ *
+ * @noInheritDoc
+ * @beta 3.0
+ * @element cwc-app-alert
+ * @slot default - Content slot for inside the alert. Usually will contain at least a <cwc-alert-content> component.
+ * @cssprop --color
+ * @cssprop --background
+ * @cssprop --icon-color
+ * @cssprop --close-icon-color
+ * @cssprop --close-icon-color-hover
+ */
+// @dynamic
+export class CwcAppAlert extends CwcBaseAlert {
+  /** Sets the color of the alert from a predefined list of statuses */
+  @property({ type: String })
+  status: 'info' | 'warning' | 'danger';
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+}
+
+registerElementSafely('cwc-app-alert', CwcAppAlert);
+
+ClarityIcons.addIcons(checkCircleIcon, infoCircleIcon, exclamationCircleIcon, exclamationTriangleIcon, timesIcon);
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'cwc-app-alert': CwcAppAlert;
+  }
+}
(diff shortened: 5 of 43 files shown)


src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+   */
+  protected createContainer(): HTMLElement {
+    const ariaLiveElement = this.document.createElement('div');
+
+    ariaLiveElement.setAttribute('id', this.id);
+    // Use clarity screen reader class to hide the dom element
+    // and fix the scrollbar shake
+    ariaLiveElement.classList.add('clr-sr-only');
+
+    ariaLiveElement.setAttribute('aria-atomic', 'true');
```
mathisscott: Seems like we have an opportunity to introduce a bulk attribute setter. something along the lines of...

```
setAttributes:: Element, Array(Array(attribute_name, value, optional_if_fn)) -> void
```

src/dev-core/src/app/alert/alert.demo.html
```diff
@@ -0,0 +1,77 @@
...
+<cwc-alert status="warning">
+    <cwc-alert-text>
+        Bar
+    </cwc-alert-text>
+    <cwc-alert-actions></cwc-alert-actions>
+</cwc-alert>
+
+<cwc-alert status="success">
+    <cwc-alert-text>
+        Bar
```
mathisscott: nit: The third one should be `Baz`

src/dev-core/src/app/alert/alert.demo.html
```diff
@@ -0,0 +1,77 @@
...
+<cwc-alert>
+    <cwc-alert-text>
+        Bar
+    </cwc-alert-text>
+    <cwc-alert-actions></cwc-alert-actions>
+</cwc-alert>
+
+<cwc-alert>
+    <cwc-alert-text>
+        Bar
```
mathisscott: nit: Baz again 😛

src/dev-core/src/app/app.component.html
```diff
@@ -5,16 +5,18 @@
 <section>
   <nav>
     <a routerLink="/">Home</a>
+    <a routerLink="/i18n">i18n</a>
+    <a routerLink="/typography">Typography</a>
+    <a routerLink="/lazy-load">Lazy Load</a>
+    <a routerLink="/layout">Layout &amp; Grid</a>
+    <a routerLink="/dropdown">Test Dropdown</a>
```
mathisscott: Eeek. I had these in alphabetical order...

Could you put them back?

src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+  constructor(document: Document) {
+    this.document = document;
+  }
+
+  /**
+   * Append text content inside the AriaLive Container. This method will check if the
+   * DOM Element is existing if not it will create one for us and the will apply the text.
+   *
+   * ```typescript
+   * this.ariaLiveService.announce(this.el.nativeElement);
```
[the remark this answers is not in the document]
bdryanovski: The service is design to handle its own cases, and create new DOM Element for every component. So one or more component could broadcast there messages independently and let the SR decide the order. Also different component could have different politeness type. That was the initial idea of going with multiple instances. 

Going static is an option as well, there are not a lot going on inside the constructors. And the idea for having it this way for the angular side was to get the document injected and using the ngOnDestroy auto call when component is destroyed so we only care for creating the message and let the “nature selection” do its thing. 

For using it as static methods there have to be made some changes on how new ids are generated for the creation of aria live containers and how we keep the reference for the required element. Also to be sure that there will be no left behind DOM Elements, this could quickly create a lot of unused elements if there is a for example progress or any type of component that create message quickly.

src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Create AriaLive DOM element as a last child of the document.
+   * After the element is created, we gonna apply Clarity class to hide it from
+   * the screen and set the `aria-live` politness.
+   *
+   * `clr-sr-only` is the CSS class that is used to hide the element from the screen.
+   *
+   * @remark
+   * Calling this method multiple times will create multiple DOM Elements, that
+   * won't be tracked and will be GC after the service is destroyed.
```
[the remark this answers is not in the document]
bdryanovski: This method was a bit simplified when it comes to checking in favor of be only callable from inside other internal methods that will prevent it from been called when the context is not browser. For that reason this method is working directly with DOM method. And its more or less extracting it into function for simplicity and for the rare case of been called from other method. 

Now when I look at it one or two ifs won’t make a big difference.

In its current version if called it will create another and another element with the same ID. But we won’t cleanup all of them inside the cleanup method. The comment above is more of explaining what could be the side effect. 

There could be change to remove the previous element before creating new one.

src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Time in milliseconds before inserting the content into the container
+ */
+export const ARIA_LIVE_TICK: number = 100;
+export class CwcAriaLiveService {
+  protected previousTimeout: ReturnType<typeof setTimeout>;
+  protected ariaLiveElement: HTMLElement | null;
+  document: Document;
+
+  readonly id = `clr-aria-live-element-${this._uniqueId}`;
```
bdryanovski: I’m looking at this from my phone .. but I don’t see from where the `_uniqueId ` Is coming ? Is applyMixing the one that is creating it? Just asking ?

mathisscott: The uniqueId is added via a typescript mixin. We have several of them in core that we can reuse across different components. See ln 124 of this file for where the mixin is applied.

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
mathisscott: From looking at the code, it appears we are only serendipitously cleaning up after the service anyway. As every use of `announce()` creates a new DOM element with the same id as the previous one tied to the service.

Having multiple elements on the same page with the same id isn't supposed to happen. At the very least we need to make the cleanup more resilient. Looking over the code, I'm not convinced that it isn't leaving stragglers behind. I'd at least want to see a test that (a) uses `announce` on the service multiple times and (b) verifies that we don't have any stragglers in the DOM after cleanup. 

Based on what I'm seeing, I don't have high confidence that test would pass.

src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (!message || politeness === AriaLivePoliteness.off) {
+      return;
+    }
+
+    if (typeof message !== 'string' && this.isBrowser() && message.textContent) {
+      message = message.textContent;
+    }
+
+    // when there is no message do NOTHING!
+    if (!message) {
```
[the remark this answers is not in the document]
mathisscott: I see.

It's likely that the flow of this code could be improved. My first instinct was to remove the check on ln 53. Reviewing this again, I can see that would be bad.

But I could also see a dev that is unfamiliar with this code doing that and even imagine it passing review.

So that tells me something about the code itself.

I think we could simplify this guard and make it clearer that this isn't an unnecessary line of code.

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
mathisscott: I'd very much be in favor of writing the few lines of code it would take to make `cleanup` remove all the DOM elements and not just the last one used.

src/clr-core/alert/alert-actions.element.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+import { componentIsStable, createTestElement, removeTestElement, waitForComponent } from '@clr/core/test/utils';
+
+describe('alert-actions element', () => {
+  let testElement: HTMLElement;
+  let component: CwcAlertActions;
+  const placeholderText: string = 'Alert Text Placeholder';
+
+  beforeEach(async () => {
+    testElement = createTestElement();
+    testElement.innerHTML = `<cwc-alert-actions>${placeholderText}</cwc-alert-actions>`;
```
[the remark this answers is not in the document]
coryrylan: I think it something that has been turned on for the GIthub vmware org account.

src/clr-core/common/services/common-strings.interface.ts
```diff
@@ -1,200 +1,200 @@
...
+  open: string;
   /**
    * Close button
    */
-  close?: string;
+  close: string;
   /**
    * Show button
    */
-  show?: string;
```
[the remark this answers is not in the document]
coryrylan: I'd guess strict null check rule and also they technically always have a english default. I think we had a similar change for the icon registry.

src/clr-core/alert/alert-actions.element.scss
```diff
@@ -0,0 +1,34 @@
+@import './../styles/mixins/utils';
+@import './../styles/mixins/buttons';
+@import './variables';
+@import './properties';
+
+:host {
+  flex: 0 0 auto;
+  white-space: nowrap;
+
+  // .dropdown:last-child {
```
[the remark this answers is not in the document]
jeeyun: These are things that need to be added after core dropdown component is implemented. I'll put a comment instead to remind us when that time comes.

src/clr-core/common/services/common-strings.interface.ts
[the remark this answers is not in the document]
jeeyun: Yup. As long as we keep the strict null check rule (which I understand we'll turn off for NG9 integration and turn back on), I have to make them non-optional.

src/clr-core/common/services/common-strings.service.ts
```diff
@@ -20,7 +20,7 @@ export class CommonStringsServiceInternal {
   /**
    * Allows you to pass in new overrides for localization
    */
-  localize(overrides: ClrCommonStrings) {
+  localize(overrides: Partial<ClrCommonStrings>) {
```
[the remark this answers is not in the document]
jeeyun: You got it. 👍

src/dev-core/src/app/app.component.html
jeeyun: Yes. Noticed this and fixed it.

src/clr-core/common/utils/aria-live.service.ts
bdryanovski: Ok, so if this is really uniq and we make new instance based on component creation, this service will work as expected. But there is room for improvements. 

But whatever we change here it will be nice to make the changes into the angular service or go with a single service that is working on both packages.

The idea for this service is hard to get into without first knowing how some of the SR are working and there corner cases. Or at least we have to document it better.

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
bdryanovski: The method has to create new DOM element only if there is no existing element already. And the ID has to be different for every instance. So in use there will be only one element per component and it will reuse it for the next message. There must be no duplicate IDs. 

At least for the angular side, if we destroy the component we gonna remove the DOM element. This is valid only for the component that is initialize the service. If for some reason, the component fail and throw error or something prevent the code for finishing the ngOnDestroy this may not remove and create orphans.

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
bdryanovski: i totally agree with that, the comments need improvements. Not sure how could he simplify the code .. maybe better naming could give a hint what is going on

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
bdryanovski: i prefer to be sure that there no double element with the same id. 

There is no case at the moment where the same instance will have need of having more than one aria live dom element ref.

If this is the case there will be no need of searching or removing more than one element.

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
mathisscott: But...

...if that is true then `announce()` only adds text to the live element when there is no live element.

So where is the reuse?

Can I not call `announce()` more than once on the same aria-live element?

Because the code says that is the case...

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  get alertIconTitle() {
+    /* 
+     * if the component's icon-title attribute is set, that value is used.
+     * otherwise, we check for status attribute and set an icon shape that matches.
+     * if neither is supplied we default to string for "Info".
+     */
+    if (this.iconTitle) {
```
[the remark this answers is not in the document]
jeeyun: Writing a reusable function as per suggested!

src/clr-core/alert/_properties.scss
```diff
@@ -0,0 +1,154 @@
...
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@include exports('alert.properties') {
+  @if $clr-use-custom-properties == true {
+    :root,
+    :host {
+      // Usage: ./_alert.clarity.scss
+      // Used for 'toggle' button for dropdowns on the right side of an alert
+      --clr-alert-action-color: var(--clr-color-neutral-700);
+      --clr-alert-action-active-color: var(--clr-color-secondary-action-900);
```
coryrylan: The refactor that is on master was three parts, I am happy to pair up on this if you would like. The other components are 95% updated I am working on the last little bit.

1. Keep existing css props in clr-ui/clr-angular and no longer use them directly in core
2. Update to naming pattern removing the clr prefix and preventing global collisions
    `--[optional template element]-[css property]-[optional element state]: value;`
3. Anything using a global `--clr` var has a sass var fallback to keep component working when no global styles are present. (Storybook now will catch this for us since it loads components in isolation)

So for alerts I think it would look something like this

```scss
// alert.element.scss
:host {
  --color: var(--clr-color-neutral-700, #{$clr-color-neutral-700});
  --background: var(--clr-color-action-50, #{$clr-color-action-50});
  --border-radius: var(--clr-global-borderradius, #{$clr-global-borderradius});
  --border-color: var(--clr-color-action-400, #{$clr-color-action-400});
  --icon-color: var(--clr-color-action-600, #{$clr-color-action-600});
  --close-icon-color: var(--clr-color-neutral-700, #{$clr-color-neutral-700});
  --close-icon-color-hover: var(--clr-color-neutral-700, #{$clr-color-neutral-700});
}
```

Then altered states can follow the same pattern

```scss
:host([status='success']) {
  --background: var(--clr-color-success-50, #{$clr-color-success-50});
  --border-color: var(--clr-color-success-400, #{$clr-color-success-400});
  --icon-color: var(--clr-color-success-700, #{$clr-color-success-700});
}
```

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,158 @@
...
+ *    <cwc-alert-text>Lorem ipsum dolor sit amet</cwc-alert-text>
+ *    <cwc-alert-actions>
+ *       <cwc-button>Fix</cwc-button>
+ *    </cwc-alert-actions>
+ * </cwc-alert>
+ * ```
+ *
+ * @noInheritDoc
+ * @beta 3.0
+ * @element cwc-alert
```
coryrylan: Once the sass is updated can we add the @cssprop decorators here for the auto docs?

src/clr-core/common/utils/aria-live.service.ts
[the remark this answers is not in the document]
jeeyun: I'll be reverting the aria-live service changes in this PR. We are looking to de-couple aria-live service from components themselves because a lot of the logic is dependent on application context which the component library has no visibility or access to.

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,181 @@
...
+  /** Sets the overall height and width of the alert and icon based on value */
+  @property({ type: String })
+  size: 'default' | 'sm';
+
+  /** Sets the icon shape to be used to override the default icon. The application must import the icon shape. */
+  @property({ type: String, attribute: 'icon-shape' })
+  iconShape = '';
+
+  /** Sets the title attribute for the icon. This may need to be set especially if the component is using custom icon shape.  */
+  @property({ type: String, attribute: 'icon-title' })
```
mathisscott: Doesn't @coryrylan 's `property` do the kebab-case for us?

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,181 @@
...
+  @property({ type: Boolean })
+  closable = true;
+
+  @querySlotAll('cwc-button') private buttons: NodeListOf<CwcButton>;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    // buttons inside alert (usually app-level) have the style of small inverse buttons
+    this.buttons.forEach(button => {
```
mathisscott: Will this work if buttons are dynamically hidden or shown with something like an ngIf?

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,181 @@
...
+      button.status = 'inverse';
+      button.size = 'sm';
+    });
+  }
+
+  get alertIconShape() {
+    return returnOrFallthrough(
```
mathisscott: 🏆

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,181 @@
...
+     * if neither is supplied we default to info-circle icon.
+     */
+  }
+
+  get alertIconTitle() {
+    /* 
+     * if the component's icon-title attribute is set, that value is used.
+     * otherwise, we check for status attribute and set an icon shape that matches.
+     * if neither is supplied we default to string for "Info".
+     */
```
mathisscott: I'm curious. Does the logic in these functions impact performance? We had to pull logic out of getters in the icons because they were slowing the browser down. Typically, we have said to put logic in setters. I'm not sure if that applies here or not.

src/clr-core/common/utils/aria-live.service.spec.ts
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { AriaLivePoliteness } from '@clr/core/common';
```
mathisscott: Is this staying in?

src/clr-core/common/utils/aria-live.service.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import { AriaLivePoliteness } from '../enums/aria-live-politeness';
+import { applyMixins } from '../mixins/apply-mixins';
+import { UniqueId } from '../mixins/unique-id';
+
+/**
+ * Time in milliseconds before inserting the content into the container
+ */
+export const ARIA_LIVE_TICK: number = 100;
+export class CwcAriaLiveService {
```
mathisscott: Is this staying in?

src/clr-core/alert/alert-actions.element.scss
```diff
@@ -0,0 +1,12 @@
+@import './../styles/mixins/utils';
+
+:host {
+  flex: 0 0 auto;
+  white-space: nowrap;
+
+  // TODO: add style override for dropdown component inside alert-actions after cwc-dropdown is implemented
+
+  ::slotted(cwc-button) {
+    margin: 0;
```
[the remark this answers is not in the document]
coryrylan: Button margins have been removed so this is not needed

src/clr-core/alert/alert.element.ts
coryrylan: Good point. You can listen to the slot change event to set the buttons

```typescript

html`<slot name="actions" @slotchange="${this.updateButtons()}"></slot>`

updateButtons() {
  this.buttons.forEach(button => {
    button.status = 'inverse';
    button.size = 'sm';
  });
}
```

src/clr-core/common/utils/aria-live.service.spec.ts
jeeyun: Nope deleting soon but since my PR's Angular version now extends from it, didn't want to  delete yet because it'll break all the Angular components that extend it. Will be deleted when I revert all the aria-live changes in this PR.

src/clr-core/common/utils/aria-live.service.ts
jeeyun: Nope deleting soon but since my PR's Angular version now extends from it, didn't want to  delete yet because it'll break all the Angular components that extend it. Will be deleted when I revert all the aria-live changes in this PR.

src/clr-core/alert/alert.element.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  render() {
+    return html`
+      <div class="alert-wrapper">
+        <div class="alert-item">
+          <div class="alert-icon-wrapper">
+            <cwc-icon class="alert-icon" shape="${this.alertIconShape}" title="${this.alertIconTitle}"></cwc-icon>
+          </div>
+          <slot name="text"></slot>
```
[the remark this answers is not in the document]
jeeyun: Great point. updating.

src/clr-core/alert/alert-text.element.ts
```diff
@@ -0,0 +1,61 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { applyMixins, baseStyles, CssHelpers, registerElementSafely } from '@clr/core/common';
+import { html, LitElement } from 'lit-element';
+import { styles } from './alert-text.element.css';
+
+class AlertMixinClass extends LitElement {}
+
+applyMixins(AlertMixinClass, [CssHelpers]);
```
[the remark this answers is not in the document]
jeeyun: I think it's old/dead code. Thanks for catching that. Removing.

mathisscott:
> With the outgoing change for `AriaLiveService` ongoing public, we may have to remove some of the changes here.
> 
> But I like that the service is extending the core service. So both packages could use identical service.
> 
> Also, `AriaLivePoliteness` I think that this won't be needed inside our component any longer.

I think the near-term plan is to remove AriaLiveService from Core...

coryrylan:
Yes the near term the service will be removed from this PR and Core. There is some additional work that needs to be done for core to better support public facing utilities that we will tackle at a later point.

src/clr-core/alert/alert.base.ts
```diff
@@ -0,0 +1,148 @@
...
+ *    <cwc-alert-content>This is an alert.</cwc-alert-content>
+ * </cwc-alert>
+ * ```
+ *
+ * @noInheritDoc
+ * @beta 3.0
+ * @element cwc-alert
+ * @slot default - Content slot for inside the alert. Usually will contain at least a <cwc-alert-content> component.
+ */
+// @dynamic
```
coryrylan: nit: probably should remove the tsdoc block from the base class since the child alert class will have its own definition.

src/clr-core/alert/app-alert.element.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * ```html
+ * <cwc-app-alert>
+ *    <cwc-alert-content>This is an alert.</cwc-alert-content>
+ * </cwc-app-alert>
+ * ```
+ *
+ * @noInheritDoc
+ * @beta 3.0
+ * @element cwc-alert
```
coryrylan: `cwc-app-alert`

jmls:
is there any documentation on this new feature ?

mathisscott:
Presently our [getting started documentation is on github](https://github.com/vmware/clarity/tree/master/src/clr-core).

The current Clarity website is not architected to handle multiple frameworks/implementations of our components. We are working to improve that for 4.0.

I believe we will have more documentation on these components available when 3.0 is finalized and released (so another couple of weeks).

Issue #4062, opened 2019-12-02
core: align component theme APIs


coryrylan:
To support flexible theming for core components we need to check that we follow the following architecture with CSS Custom Props. This also reduces the number of stateful variables we have to create. If a new `status` is added we do not have to introduce two new variables to be able to configure the new status type.

```scss
:host {
  --clr-button-background: #{$clr-global-blue-100};
  --clr-button-color: #{$clr-global-blue-300};
  background: var(--clr-button-background};
  color: var(--clr-button-color);
}

:host[status="danger"] {
  --clr-button-background: #{$clr-global-red-100};
  --clr-button-color: #{$clr-global-red-300};
}
```

This architecture allows custom classes/states to be defined by the consuming application. This encourages apps stay consistent and use custom states instead of modifying the entire look and feel of Clarity globally.

```scss
// Public API
cwc-button {
  --clr-button-background: green;
  --clr-button-color: darkgreen;
}

cwc-button.product-button {
  --clr-button-background: purple;
  --clr-button-color: darkpurple;
}
```

```html
<cwc-buitton class="product-button">my purple product specific brand button</cwc-button>
```

- [x] alerts
- [x] icon
- [x] badge
- [x] tag
- [x] button
- [x] document pattern in contribution guidelines

coryrylan:
Closing in favor of continuing work https://github.com/vmware/clarity/issues/4263

Issue #4061, opened 2019-11-30
Core: remove global style references


coryrylan:
Clean up global style references and any style APIs that would cause breaking changes after 3.0
- [x] remove all component margins
- [x] remove global styles for 3.0 (typography/layout)
- [x] remove use of global css props internally with core components
- [x] remove global prefix convention on host scoped styles

## Margins
Ensure all core components have no specified `:host` margins. Margins on a host element make assumptions of the layout that is external to their responsibility. Using a design token system designers and developers should be able to layout components consistently and with specified intent and constraint. The design token values are still a wip but we want to remove margins before 3.0 to prevent future visual breaking changes.

Core will need to handle design tokens and variables slightly different than clr-ui/angular due to shadow dom and component composition but we should align the space values for the token to stay consistent.

Related issues/context:
https://github.com/vmware/clarity/issues/181
https://github.com/vmware/clarity/issues/4116
https://github.com/vmware/clarity/issues/3303

## Layout & Typography
Both the layout/grid utilities and typography also suffer from inconsistent layout margins (not typographic whitespace) and spacing. For 3.0 we will omit these from the core global styles until they can be addressed. Also defining non global explicit utils will allow possible migration strategies with scoped utils.

## Global CSS Properties
Core Components should not reference global CSS custom properties as this creates a secondary way to theme components and creates twice the API surface area. Our clr-ui dark theme follows a similar pattern and only changes the local component scoped variables leaving global values unchanged. For core components to share values such as spacing and color they internally should use private sass variables and not global CSS Custom Properties.

## Global Prefixing
Public API surface for component styling should not use the global prefixing convention. This is necessary for micro frontends where a component may need to be aliased due to app embedding. This prevents version collisions and dependencies form having to share the same global values.

```css
cwc-badge {
  --color: red;
  --background: blue;
}
```

Teams can prefix their own core components to share and embed with other products without the risk of a host app breaking their product specific styles.

```html
<style>
  cwc-badge {
   --color: green;
  }
</style>
<cwc-badge></cwc-badge> <!-- only host application badges are changed -->

<product-cwc-badge></product-cwc-badge> <!-- embedded product styles are preserved -->
```

In very rare cases a naming collision can occur with a local host var and global theming var breaking on the component ([see example](https://codepen.io/coryrylan/pen/vYEMxgo)).

```scss
// global scope
:root {
  --primary-color: blue;
}

// component scope (:host)
button {
  // naming colision, due to global and local names
[...]

Issue #4025, opened 2019-11-15
Deploy Core docs and link from website


gnomeontherun:
We need to get the Core documentation up and properly referenced in the website.

* [ ] Deploy core documentation 
* [ ] Add a link from main website to core docs 
* [ ] Update content on getting started/homepage related to Angular/Core

coryrylan:
Closed by https://github.com/vmware/clarity/pull/4314

PR #4243, opened 2020-01-29
core: API docs and style API

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [x] Feature
* [x] Build related changes
* [x] CI related changes

## What is the new behavior?

This PR adds the following changes to Core
- Updated Contribution Guidelines
- Storybook dev environment and API docs (deploy will be a separate PR) #4025 
- Refactor component style states #4062 
- Remove global style references in Core #4061 

### [Storybook and style API demo](https://test-deploy.firebaseapp.com/)

This PR primarily has two parts. First add storybook as our way to document the Core component APIs as well as provide a dev environment to work in. (Removing dev-core will be a separate PR).

The second part is updating the style APIs of the core components to allow the most flexibility with a minimal API surface. This includes ensuring components do not rely on global styles. This prevents possible future style conflicts between embedded components and allows new teams to onboard Clarity components faster without having to change large parts of their global styles. Details are listed in #4061 and would like feedback for 3.0.

This PR is blocked on merging the following are merged
https://github.com/vmware/clarity/pull/4241
https://github.com/vmware/clarity/pull/4245

* [ ] Yes
* [x] No


package.json
```diff
@@ -100,10 +102,7 @@
     "core:build:sass:components": "sass-render --q --suffix '.css.ts' -t ./src/clr-core/sass-template.js './src/clr-core/**/*element.scss'",
     "core:build:sass:global": "sass --source-map --precision=8 ./src/clr-core/styles/global.scss ./dist/clr-core/global.css",
     "core:build:sass:global:min": "csso -i ./dist/clr-core/global.css -o ./dist/clr-core/global.min.css -s file --no-restructure",
-    "core:build:api": "npm-run-all -p core:build:api:md core:build:api:editor --",
-    "core:build:api:md": "wca analyze 'src/clr-core/**/*.ts' --format=md --outDir=docs/core",
-    "core:build:api:json": "wca analyze 'src/clr-core/**/*ts' --format=json --outDir=src/dev-core/api",
-    "core:build:api:editor": "wca analyze 'src/clr-core/**/*.ts' --format=json --outFile dist/clr-core/custom-elements.json"
```
coryrylan: Storybook API docs only needs the custom-elements.json to generate all docs.

package.json
```diff
@@ -219,8 +227,6 @@
     "tslint": "5.17.0",
     "tslint-config-prettier": "1.18.0",
     "tslint-defocus": "2.0.6",
-    "typedoc": "0.15.0",
-    "typedoc-plugin-no-inherit": "1.1.10",
```
coryrylan: Typedoc is no longer needed as a direct dependency. May need later for clr-angular.

package.json
```diff
@@ -235,14 +241,8 @@
...
-      "pre-commit": "npm run format:fix && npm run license:fix && lint-staged",
+      "pre-commit": "npm run format:fix && npm run license:fix",
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
-  },
-  "lint-staged": {
-    "src/clr-core/**/*.{js,ts}": [
-      "npm run core:build:api",
-      "git add"
-    ]
```
coryrylan: Lint staged is no longer needed since we no longer need to generate markdown files for core API docs

src/clr-angular/popover/common/abstract-popover.spec.ts
```diff
@@ -15,6 +15,63 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
 import { AbstractPopover } from './abstract-popover';
 import { POPOVER_HOST_ANCHOR } from './popover-host-anchor.token';
 
+@Component({
```
coryrylan: To get TS working with TS projects and storybook the tsconfigs needed to be updated to run the tests as es2015. This speeds up the tests as a side effect but forced this code to be reordered since true ES2015 classes are not hoisted like the compiled ES5 equivalent.

src/clr-angular/typography/_typography.scss
```diff
@@ -0,0 +1,220 @@
+// Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
```
coryrylan: For the first version of core we are just shipping the components. I removed the core global styles from pulling in the typography and grid until we have time to evaluate how we want core users to adopt those styles.

src/clr-core/.storybook/addons.js
```diff
@@ -0,0 +1,25 @@
+import { addons } from '@storybook/addons';
+import '@storybook/addon-docs/register';
+import '@storybook/addon-knobs/register';
+import '@storybook/addon-actions/register';
```
coryrylan: This file registers the plugins needed for storybook.

src/clr-core/.storybook/config.js
```diff
@@ -0,0 +1,49 @@
...
+  ],
+  module
+);
+
+const req = require.context('../', true, /\.stories\.(ts|mdx)$/);
+if (module.hot) {
+  module.hot.accept(req.id, () => {
+    const currentLocationHref = window.location.href;
+    window.history.pushState(null, null, currentLocationHref);
+    window.location.reload();
```
coryrylan: Storybook supports web components now but does not yet have hot module reloading. So when the source code changes it has to refresh the iframe. (similar to Angular CLI)

src/clr-core/.storybook/main.js
```diff
@@ -0,0 +1,30 @@
+const { TsConfigPathsPlugin } = require('awesome-typescript-loader');
+const path = require('path');
+
+module.exports = {
```
coryrylan: This file allows us to configure how storybook processes our story files. This should be kept minimal and is just to allow us to write stories in TypeScript.

src/clr-core/.storybook/public/demo.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+h2 {
+  font-size: 1.4rem;
+}
+
+p {
+  color: hsl(0, 0%, 40%);
+  font-size: 0.7rem;
+}
```
coryrylan: These are some temporary basic styles for typography until we can plan how core users should consume clarity typography styles.

src/clr-core/.storybook/public/demo.css
```diff
@@ -0,0 +1,65 @@
...
+
+.docblock-propstable-body li {
+  list-style-position: outside;
+}
+
+.demo-inverse {
+  padding: 0.2rem;
+  display: inline-block;
+  background: hsl(0, 0%, 34%);
+}
```
coryrylan: These are some basic styles for a few of the demo/getting started pages and are scoped to `.doc`

src/clr-core/badge/badge.element.scss
```diff
@@ -12,59 +14,57 @@
...
-
-  background-color: var(--clr-badge-background-color);
-  color: var(--clr-badge-color);
+  background: var(--background);
+  color: var(--color);
 }
 
 :host(:visited) {
-  --clr-badge-color: var(--clr-color-on-neutral-600);
+  --color: #{$clr-color-on-neutral-600};
```
coryrylan: In depth description of global vs local vars is here https://github.com/vmware/clarity/issues/4061 but mainly to prevent components from relying on global state which can get really weird when apps start embedding other apps/components.

src/clr-core/badge/badge.stories.ts
```diff
@@ -0,0 +1,91 @@
...
+      gray: 'gray',
+      purple: 'purple',
+      blue: 'blue',
+      orange: 'orange',
+      'light-blue': 'light-blue',
+    },
+    undefined,
+    propertiesGroup
+  );
+  const textColor = colorKnob('--color', undefined, cssGroup);
```
coryrylan: Not a fan of prettier here. The options for the select can be a simpler string array but I have to do the label object style so I can set defaults of undefined to remove the attrs on the components.

src/clr-core/badge/badge.stories.ts
```diff
@@ -0,0 +1,91 @@
...
+  const background = colorKnob('--background', undefined, cssGroup);
+
+  return html`
+    <style>
+      cwc-badge {
+        ${setStyles({
+          '--color': textColor,
+          '--background': background,
+        })}
+    </style>
```
coryrylan: The `setStyles` makes it easy to hook the storybook knobs into dynamic styles that can change the CSS properties in the storybook UI

src/clr-core/button/button.element.ts
```diff
@@ -34,162 +32,58 @@ export enum ClrLoadingState {
...
+ * @cssprop --box-shadow-color
+ * @cssprop --border-radius
+ * @cssprop --border-width
+ * @cssprop --font-size
+ * @cssprop --font-weight
+ * @cssprop --border-color
+ * @cssprop --color
+ * @cssprop --background
+ * @cssprop --line-height
+ * @cssprop --letter-spacing
```
coryrylan: I applied the technique we started to use to modify different states with CSS properties in badges/tags to the button component. This made the API much surface much smaller. https://github.com/vmware/clarity/issues/4062

src/clr-core/button/button.element.ts
```diff
@@ -34,162 +32,58 @@ export enum ClrLoadingState {
...
-  status: 'default' | 'primary' | 'inverse' | StoplightStatusTypes;
+  status: 'primary' | 'success' | 'danger' | 'inverse';
 
   /**
    * Sets the overall height and width of the button based on the following string values:
-   * 'default', 'sm'
    */
   @property({ type: String })
-  size: DefaultSizes;
+  size: 'sm' | 'md' | 'icon';
```
coryrylan: I wasnt sure about this change, seeing the API in storybook it seemed to make more sense that `icon` is a size option. This prevented impossible/unsupported states.

src/clr-core/button/button.element.ts
```diff
@@ -34,162 +32,58 @@ export enum ClrLoadingState {
...
-  action: 'default' | 'outline' | 'link';
+  action: 'solid' | 'outline' | 'flat';
 
   /**
    * Sets the color of the button to match the following string statuses
-   * 'default', 'primary', 'inverse', 'success', 'warning', 'danger'
    */
   @property({ type: String })
-  status: 'default' | 'primary' | 'inverse' | StoplightStatusTypes;
+  status: 'primary' | 'success' | 'danger' | 'inverse';
```
coryrylan: Removed `StoplightStatusTypes` since buttons do not have a true warning color.

src/clr-core/button/button.element.ts
```diff
@@ -34,162 +32,58 @@ export enum ClrLoadingState {
...
+  /**
+   * Define the type of action the button triggers
+   *
+   * - `solid`: buttons direct the user’s attention to the primary action the application is suggesting that the user take.
+   * - `outline`: buttons indicate secondary actions that compliments a primary action or reduces visual noise when there are many actions on the page.
+   * - `flat`: buttons are used as tertiary buttons. Can also be used inline because they are different from content in style and recognizable as buttons alongside content.
+   */
   @property({ type: String })
-  action: 'default' | 'outline' | 'link';
+  action: 'solid' | 'outline' | 'flat';
```
coryrylan: Updated to match how the docs describe the buttons as well as what was discussed in slack.

src/clr-core/button/button.stories.ts
```diff
@@ -0,0 +1,235 @@
...
+ClarityIcons.addIcons(userIcon);
+
+export default {
+  title: 'Components|Button',
+  component: 'cwc-button',
+  parameters: {
+    options: { showPanel: true },
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/v2mkhzKQdhECXOx8BElgdA/Clarity-UI-Library---light-2.2.0?node-id=20%3A0',
```
coryrylan: Storybook we can link the related figma file and embed it in the API doc example

src/clr-core/common/decorators/property.ts
```diff
@@ -35,7 +35,14 @@ export function getDefaultOptions(
     case Object:
       return { reflect: false, ...options };
     case String:
-      return { reflect: true, attribute: camelCaseToKebabCase(propertyKey), ...options };
+      return {
+        reflect: true,
+        attribute: camelCaseToKebabCase(propertyKey),
+        converter: {
+          toAttribute: (value: string) => (!!value ? value : null),
```
coryrylan: If we get a falsy value remove the attribute from the element

src/clr-core/icon-shapes/icon.element.scss
```diff
@@ -29,33 +21,13 @@
...
 :host {
+  --color: #{$clr-color-neutral-700};
+  --badge-color: #{$clr-color-danger-700};
   display: inline-block;
   @include equilateral($clr_baselineRem_0_667);
   margin: 0;
   vertical-align: middle;
-  @include setIconColor(#{'var(--clr-icon-color-default, currentColor)'});
-
-  & > div {
```
coryrylan: Realized we didnt have the wrapper div anymore so this was all dead code.

src/clr-core/icon-shapes/icon.element.scss
```diff
@@ -132,16 +96,21 @@ svg {
...
 :host([inverse]) {
-  @include setIconColor(#{'var(--clr-icon-color-inverse, var(--clr-color-neutral-0,' + $clr-color-neutral-0 + '))'});
+  --color: #{$clr-color-neutral-0};
+}
+
+// DEPRECATED IN 3.0: is-info, is-inverse
+:host([inverse].is-info),
+:host([status='info'].is-inverse),
+:host([inverse][status='info']) {
+  --color: #{$clr-color-action-600};
```
coryrylan: I wasnt sure if this was a valid state? There was no inverse info icon color so when the icon was inverse with state value info it would be white instead of blue.

src/clr-core/icon-shapes/icon.element.ts
```diff
@@ -205,10 +202,3 @@ export class CwcIcon extends IconMixinClass {
   }
 }
 export interface CwcIcon extends IconMixinClass, UniqueId, CssHelpers {}
-registerElementSafely('cwc-icon', CwcIcon);
-
-declare global {
-  interface HTMLElementTagNameMap {
-    'cwc-icon': CwcIcon;
-  }
-}
```
coryrylan: Realized this was attempting to register twice. We already register in the `@clr/core/icon` entry point

src/clr-core/package.js
```diff
@@ -0,0 +1,32 @@
...
+
+del.sync(['./dist/clr-core/**/*.{tsbuildinfo,ngsummary.json}', './dist/clr-core/*.{tsbuildinfo,ngsummary.json}']);
+
+read('./dist/clr-core')
+  .filter(f => f.includes('package.json'))
+  .forEach(file => {
+    const data = fs.readJsonSync(file);
+    ['__processed_by_ivy_ngcc__', 'scripts'].forEach(p => delete data[p]);
+    fs.writeJsonSync(file, data, { spaces: 2 });
+  });
```
coryrylan: Angular CLI and TS enjoy adding random cache files and mutating our package json files during the build. This script cleans it all up after for deployment.

src/clr-core/tag/tag.element.scss
```diff
@@ -60,76 +55,76 @@
...
 }
 
 :host([color='purple']) {
-  --clr-tag-border-color: var(--clr-color-secondary-action-500);
+  --border: #{$clr-global-borderwidth} solid #{$clr-color-secondary-action-500};
 
   ::slotted(cwc-badge) {
-    --clr-badge-background-color: var(--clr-color-secondary-action-500);
-    --clr-badge-color: var(--clr-color-on-secondary-action-500);
+    --clr-badge-background: #{$clr-color-secondary-action-500};
```
coryrylan: todo: update

src/clr-core/tsconfig.json
```diff
@@ -17,38 +19,18 @@
     "noUnusedParameters": true,
     "strictFunctionTypes": true,
     "noFallthroughCasesInSwitch": true,
-    "strictNullChecks": true,
+    "strictNullChecks": false,
```
coryrylan: Disabled temporary for tests, once ng v9 is merged going to re-enable and fixup tests

tsconfig.json
```diff
@@ -22,8 +22,8 @@
       "@clr/angular": ["src/clr-angular"],
       "@clr/icons": ["src/clr-icons"],
       "@clr/icons/shapes/*": ["src/clr-icons/shapes", "src/clr-icons/shapes/*"],
-      "@clr/core": ["src/clr-core"],
-      "@clr/core/*": ["src/clr-core/*"]
+      "@clr/core": ["dist/clr-core"],
+      "@clr/core/*": ["dist/clr-core/*"]
```
coryrylan: Always have projects point to the dist when using core. The ng9 PR has this same update.

package.json
```diff
@@ -81,29 +81,28 @@
...
     "ui:lint:fix": "stylelint --config ./src/clr-core/.stylelintrc --fix \"src/clr-ui/**/*.scss\"",
-    "core:start": "npm-run-all core:build -p core:build:sass:watch 'core:build:ts -- -w' 'ng serve dev-core'",
-    "core:dev:build": "ng build dev-core --aot --sourceMap=false",
-    "core:dev:build:prod": "ng build dev-core --prod",
-    "core:lint": "npm run prelint && tslint -p . -c ./src/clr-core/tslint.json \"src/clr-core/**/*.ts\" -e \"src/clr-core/**/*.css.ts\" && tslint -p . -c ./src/dev-core/tslint.json \"src/dev-core/**/*.ts\" && stylelint --config ./src/clr-core/.stylelintrc \"src/clr-core/**/*.scss\" \"src/dev-core/**/*.scss\"",
-    "core:lint:fix": "npm run prelint && tslint --fix -p . -c ./src/clr-core/tslint.json \"src/clr-core/**/*.ts\" && tslint --fix -p . -c ./src/dev-core/tslint.json \"src/dev-core/**/*.ts\" && stylelint --config ./src/clr-core/.stylelintrc --fix \"src/clr-core/**/*.scss\" \"src/dev-core/**/*.scss\"",
+    "core:start": "npm-run-all core:build -p core:build:watch storybook",
+    "core:start:old": "npm-run-all core:build -p core:build:watch 'ng serve dev-core'",
```
jeeyun: will `core:start:old` hang around for a while? or did you mean to delete it before the merge?

package.json
```diff
@@ -81,29 +81,28 @@
...
     "core:test:unit": "ng test clr-core --watch=false",
     "core:test:treeshaking": "webpack --config src/clr-core/test-bundles/webpack.config.js && bundlesize --config src/clr-core/test-bundles/bundlesize.json",
     "core:build": "npm-run-all core:build:sass -p core:build:package 'core:build:ts -- --force' core:build:api core:build:clean",
-    "core:build:clean": "del ./dist/clr-core/**/*.{tsbuildinfo,ngsummary.json} && del ./dist/clr-core/*.{tsbuildinfo,ngsummary.json}",
+    "core:build:watch": "npm-run-all -p core:build:sass:watch 'core:build:ts -- -w'",
+    "core:build:clean": "node ./src/clr-core/package.js",
     "core:build:package": "cpy '**/package.json' '../../dist/clr-core/' --cwd=src/clr-core --parents && cpy ./src/clr-core/README.md ./dist/clr-core/; replace '@VERSION' $npm_package_version ./dist/clr-core/package.json",
     "core:build:ts": "tsc --b src/clr-core/tsconfig.project.json",
-    "core:build:src": "npm-run-all core:build:sass:components",
+    "core:build:src": "npm-run-all core:build:sass:components 'core:build:ts -- --force'",
```
jeeyun: typo (first `--`)?

src/clr-angular/button/_properties.buttons.scss
```diff
@@ -4,8 +4,7 @@
 
 @include exports('button.properties') {
   @if $clr-use-custom-properties == true {
-    :root,
-    :host {
+    :root {
```
jeeyun: Do you remember why we did both `:root` and `:host`? I vaguely remember adding this for a specific use case (was it browser support?) And why do we not need it anymore?

package.json
coryrylan: I was planning on leaving it until I can open a separate PR to remove the dev app. It wont build it during the CI build but still lets us run it in dev mode. I didnt want to change more than I had to in a single PR.

package.json
coryrylan: This is a bash syntax I think? In the run-all docs it described it as a way to pass a flag into another flag. The `core:build:ts` is technically a flag value of npm-run-all so its a way to pass `--force` down into `core:build:ts`. This was nice since I didnt have to write an additional npm script to just have a flag on.

src/clr-angular/button/_properties.buttons.scss
coryrylan: So we originally had both for two reasons. First root was for clr-ui since everything is global. Root works for some cases with the web component but if in the web component we refer and try to reassign a var it can break down.

```scss
:host(state='danger') {
  --background: red; // this override is scoped to :host not :root first
}
```

So when we wanted to change the var based on the component state global vars dont work very well. So we added both `:root` and `:host` to the selector so all properties would apply to both global `:root` scope and local `:host` scope. This actually after debugging got a bit weird since it was sometimes hard to tell which property was local or global in the dev tools.

This worked but after refactoring to follow the 'modify var locally based on component state pattern' like (badges/tags) it didn't seem like we even needed most of the stateful global clr-ui vars in the core component. The component could have all its vars scoped directly to its `:host`. There was no need to have the duplicate selector and pull in all those vars when the component only used a small subset of vars that clr-ui used.

docs/CODING_GUIDELINES_CORE.md
```diff
@@ -274,7 +376,7 @@ adhere to the following guidelines for our Core Web Component codebase.
 * We ship modern es2015 modules as the default.
 * Use a module bundler to dedupe and handle peer dependencies.
 * Share peer deps to prevent multiple versions running at one time.
-* `@clr/core/common` is private API do not use.
+* `@clr/core/common` is private internal API do not use.
```
coryrylan: Should probably reword this. Its internal for core but a contributor can use it for new components internally and component consumers should not.

package.json
jeeyun: Ah interesting. I wasn't sure if that was intentional. Thanks for explanation! 👍

src/clr-core/.storybook/addons.js
```diff
@@ -0,0 +1,25 @@
...
+  options: { showPanel: true },
+  theme: create({
+    base: 'light',
+    brandTitle: 'Clarity Design - Web Components',
+    brandUrl: 'https://clarity.design',
+    brandImage: './assets/images/clarity-logo.svg',
+  }),
+});
+
+setTimeout(() => {
```
mathisscott: LOL. `setTimeout` is like the American Express of JavaScript ("it's everywhere you wanna be!")

src/clr-core/badge/badge.element.scss
```diff
@@ -1,7 +1,9 @@
 @import './../styles/mixins/mixins';
-@import './properties';
 
 :host {
+  --background: #{$clr-color-neutral-600};
```
mathisscott: I feel pretty strongly that we should use `var(--clr-color-neutral-600)` in these cases instead of forcing users to rely on SASS variables that they won't be able to override in our next major version...

src/clr-core/badge/badge.element.scss
mathisscott: My issue is reliance on the SASS variables. We should start considering these to be constants and consider how that impacts users that want to update the color palette instead of component variables at a micro level.

src/clr-core/common/decorators/property.ts
mathisscott: I may be able to reuse this now in a couple of instances in icons where i had to use setters to basically implement this logic.

src/clr-core/icon-shapes/icon.element.scss
mathisscott: It should be a valid state, making the icon the dark theme info color for lack of a better construct...

mathisscott: For reference that color would be `clr-color-action-400`

src/clr-core/icon-shapes/icon.element.scss
```diff
@@ -366,19 +312,13 @@ svg {
...
+  --badge-color: #{$clr-color-danger-500};
 }
 
 // DEPRECATED IN 3.0: is-highlight, is-inverse
 :host([inverse].is-highlight),
 :host(.is-highlight.is-inverse) {
-  @include setIconColor(
-    #{'var(--clr-icon-color-inverse-info, var(--clr-color-action-400,' + $clr-color-action-400 + '))'}
-  );
+  --color: $clr-color-action-400;
```
mathisscott: I'm really against how we are relying on SASS variables throughout all of this code. 

In either case, this line is missing the `#{}`

src/clr-core/icon-shapes/icon.element.spec.ts
```diff
@@ -265,6 +265,8 @@ describe('icon element', () => {
         .map((val, index, arry) => {
           if (index !== 0 && index < arry.length - 1) {
             return val;
+          } else {
+            return undefined;
```
mathisscott: I don't trust `undefined`. Have you considered `void 0`?

src/clr-core/tag/tag.element.scss
```diff
@@ -1,34 +1,29 @@
...
-  @include css-var(letter-spacing, clr-tag-letter-spacing, $clr-label-letter-spacing);
-
-  // line-height needed to be adjusted to force center alignment
-  line-height: $clr-label-font-size + $clr_baselineRem_1px;
+  --font-size: #{$clr-p7-font-size};
+  --font-weight: #{$clr-p7-font-weight};
+  --letter-spacing: #{$clr-p7-letter-spacing};
+  --border-radius: #{$clr_baselineRem_0_5};
+  --background: none;
+  --background-hover: #{$clr-color-neutral-200};
```
mathisscott: I think I understand why you want to do this. Have you considered...
`var(--clr-color-neutral-200, #{$clr-color-neutral-200})`

So that way you get the default in there while allowing for an override?

src/clr-core/tag/tag.element.scss
```diff
@@ -42,14 +37,14 @@
...
+      background-color: var(--background-hover);
       cursor: pointer;
     }
 
     &:host(:active) {
       // OldEdge/IE
-      box-shadow: 0 $clr_baselineRem_1px 0 0 var(--clr-tag-bg-color) inset;
-      box-shadow: 0 $clr-label-borderwidth 0 0 var(--clr-tag-bg-color) inset;
+      box-shadow: 0 $clr_baselineRem_1px 0 0 var(--background) inset;
+      box-shadow: 0 $clr-global-borderwidth 0 0 var(--background) inset;
```
mathisscott: Why are we duplicating here? This looks like it was carried forward when it shouldn't have been.

Also `$clr-global-borderwidth` is not overrideable starting in 4.0. That may be fine in this particular instance.

src/clr-core/.storybook/addons.js
coryrylan: Yeah that was the workaround in their github issue :/

src/clr-core/badge/badge.element.scss
coryrylan: Updated to use color vars with sass fallbacks.

docs/CODING_GUIDELINES_CORE.md
```diff
@@ -257,8 +278,89 @@ adhere to the following guidelines for our Core Web Component codebase.
...
+
+* Components are themed using CSS Custom Properties defined on the `:host` element.
+  This ensures the component is only styled explicitly by the API we define and
+  cannot accidentally be changed at a global scope.
+
+  ```scss
+  // cwc-badge
+  :host {
+    // public style API
+    --background: #{$clr-color-neutral-600};
```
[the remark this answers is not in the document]
coryrylan: Updated to use color vars with sass fallbacks.

src/clr-core/badge/badge.element.scss
coryrylan: Updated to use color vars with sass fallbacks.

src/clr-core/tag/tag.element.scss
coryrylan: Updated to use color vars with sass fallbacks.

Issue #4240, opened 2020-01-28
core: ssr check


coryrylan:
When Core registers a custom element it should check that it is running within a browser to prevent any errors for server side rendering like Angular Universal. Core wont be able to render views with SSR but we can prevent it from executing in node.

In the `registerElementSafely` function we can check if the code is running in a browser, if not then don't attempt to register. `if (typeof window !== 'undefined')`.

There wont be an easy way to unit test that I can think of but we should spin up an Angular Universal project with the latest Angular 9 and test that SSR works with our components.
https://next.angular.io/guide/universal

PR #4251, opened 2020-01-31
fix: adding window check for server-side rendering

mathisscott:
• added check for server-side rendering
• fixed an issue with the dev-core app icon sets page

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4240 

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/common/utils/register.ts
+++ b/src/clr-core/common/utils/register.ts
@@ -1,19 +1,25 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
 import curryN from 'ramda/es/curryN';
-import __ from './__';
-import { elementExists } from './exists';
+import { elementExists, existsInWindow, isBrowser } from './exists';
 
-const addElementToRegistry = curryN(3, (tagName: string, elementClass: any, registry = window.customElements) => {
-  if (elementExists(tagName)) {
-    console.warn(`${tagName} has already been registered`);
-  } else {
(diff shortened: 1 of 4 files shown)


src/clr-core/common/utils/register.ts
```diff
@@ -1,19 +1,32 @@
...
+    }
   }
-});
+);
 
-export const registerElementSafely = addElementToRegistry(__, __, window.customElements);
+export function registerElementSafely(tagName: string, elementClass: any) {
+  const guard = and(!isNil(window), existsInWindow(['customElements']));
+
+  if (!guard) {
```
Shijir: nit: can move that `addElementToRegistry` call below into this check but with `guard` equals truthy?

mathisscott: sure. makes sense. thanks!

src/clr-core/common/utils/register.ts
```diff
@@ -1,19 +1,32 @@
...
+    } else {
+      registry.define(tagName, elementClass);
+    }
   }
-});
+);
 
-export const registerElementSafely = addElementToRegistry(__, __, window.customElements);
+export function registerElementSafely(tagName: string, elementClass: any) {
+  const guard = and(!isNil(window), existsInWindow(['customElements']));
```
coryrylan: nit: Could this be extracted to a function like `isBrowser()`?

Is there an advantage of `and` vs `&&` if its not used as a curried func?

mathisscott: Re: and. Nope. I can change that. 

Re: shared function. Yup. Will do.

mathisscott: Re: and. Nope. I can change that. 

Re: shared function. Yup. Will do.

PR #4291, opened 2020-02-18
fix(a11y): type error on focus initialization

znpopov:
The key-focus component does not handle properly the focus initialization, when some of the focusable elements are being removed.
In case the focus was on an element, whose index is no longer available, the component throws a type error.
This change handles the case by initializing the focus on the last focusable element.

Signed-off-by: Zdravko Popov <zpopov@vmware.com>

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

https://stackblitz.com/edit/clarity-dark-theme-v2-zushxr?file=src%2Fapp%2Fapp.component.ts

TypeError

Issue Number: N/A

## What is the new behavior?
The focus is initialized on the last focusable element.

* [ ] Yes
* [x] No

## Other information
This bug was introduced with the addition of the key-focus component.
For this reason, the fix needs to be back-ported to v2 and v1 branches.

### DIFF
--- a/src/clr-angular/utils/focus/key-focus/key-focus.spec.ts
+++ b/src/clr-angular/utils/focus/key-focus/key-focus.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -16,7 +16,7 @@ import { KeyCodes } from '@clr/core/common';
     <div *ngIf="open" clrKeyFocus [clrDirection]="direction" [clrFocusOnLoad]="focusOnLoad" (clrFocusChange)="changed = true">
       <button clrKeyFocusItem>Button 1</button>
       <button clrKeyFocusItem>Button 2</button>
-      <button clrKeyFocusItem>Button 3</button>
+      <button *ngIf="showLast" clrKeyFocusItem>Button 3</button>
     </div>
   `,
 })
@@ -27,18 +27,22 @@ class TestComponent {
   changed: boolean = false;
   direction: string = 'vertical';
   focusOnLoad = true;
+  showLast: boolean = true;
 }
 
 @Component({
   template: `
-    <div [clrKeyFocus]="buttons">
+    <div [clrKeyFocus]="buttons" [clrFocusOnLoad]="focusOnLoad">
       <button>Button 1</button>
       <button>Button 2</button>
+      <button *ngIf="showLast">Button 3</button>
     </div>
   `,
 })
 class DOMTestComponent {
   buttons: any;
+  focusOnLoad: boolean = true;
+  showLast: boolean = true;
 }
 
 let fixture: ComponentFixture<any>;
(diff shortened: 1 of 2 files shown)


src/clr-angular/utils/focus/key-focus/key-focus.ts
```diff
@@ -108,7 +108,7 @@ export class ClrKeyFocus {
 
   private get currentItem() {
     if (this._current >= this.focusableItems.length) {
-      return null;
+      this._current = 0;
```
gnomeontherun: Why should the focus move to the first item? This might make sense if you are on the first item and it is removed, but if you are on the last item and it is removed I would think focus should move to the new last focusable item? 

Your demo shows the latter behavior, so I would expect focus to move to the `this._current = this.focusableItems.length - 1`. 

If any item in the list is removed except the last item, the current index remains the same and items shift over so that shouldn't be an issue.

znpopov: Thanks for the review.
I have updated the request with the corresponding changes.

znpopov:
It turns out that the key-focus component can be used by two different ways:
1. Providing a list of focusable elements.
2. If the list is not provided, the component  auto queries for clrKeyFocusItem.

I have addressed the requested changes for both cases and added corresponding tests.

src/clr-angular/utils/focus/key-focus/key-focus.spec.ts
```diff
@@ -280,7 +292,19 @@ describe('KeyFocus directive', () => {
       // An example of this is the Tabs component.
       domComponent.buttons = Array.from(fixture.nativeElement.querySelectorAll('button'));
       fixture.detectChanges();
-      expect(clarityDirective.focusableItems.length).toBe(2);
+      expect(clarityDirective.focusableItems.length).toBe(3);
+    });
+    it('current value updates, when elements are being removed', () => {
```
gnomeontherun: This is just a duplicate test of the above using the template instead of the API. It can be removed.

src/clr-angular/utils/focus/key-focus/key-focus.ts
```diff
@@ -108,7 +108,7 @@ export class ClrKeyFocus {
 
   private get currentItem() {
     if (this._current >= this.focusableItems.length) {
-      return null;
+      this._current = this.focusableItems.length - 1;
```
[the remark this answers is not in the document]
znpopov: Thanks for this.

I moved the logic into the `initializeFocus()` method, since changes can be handled from the `listenForItemUpdates()` method or from the `focusableItems` setter.

With this logic, if all items are removed temporarily, the getter will never be called.

src/clr-angular/utils/focus/key-focus/key-focus.spec.ts
znpopov: It is not an exact duplicate. 
In this scenario, the changes are handled by the `focusableItems` setter.
In the case of the API test, the changes are handled by the `listenForItemUpdates()` method.
In my opinion, both tests are needed.

znpopov:
Hi @gnomeontherun,
Do you need more input from my side, in order the PR to be merged?
Thanks!

src/clr-angular/utils/focus/key-focus/key-focus.spec.ts
gnomeontherun: What is triggering both tests to actually change is simply the toggling of `showLast`, which triggers the `listenForItemUpdates` to fire based on the QueryList change. These tests are just setting up the menu in two different ways ,but both asserting the result the same way (hence the duplication). 

Thinking about it more, this test should not be asserting that `current` is set but that the `document.activeElement` equals the expected button. If you change the test to assert that, I would be satisfied that its testing two different ways.

znpopov: Thanks. The test has been updated with the suggested assertions.

Issue #4263, opened 2020-02-10
core: align exposed public style APIs and update contribution guidelines


coryrylan:
## Proposal
Ensure the exposed public styles globally and locally at the component level are consistent and expose the appropriate styles. We should add this to the contribution guidelines.

## CSS Props Component
- only expose non layout styles (background, color, box-shadows)
- prefer exposing shorthand values unless the shorthand alters sizes like border
   `border-color: var(--border-color)` not `border: var(--border)`
- colors should use global vars and fall back to hard coded sass vars so components render independently without requiring global styles.
  ```scss
  :host {
    --color: var(--clr-color-neutral-0, #{$clr-color-neutral-0});
    --background: var(--clr-color-action-600, #{$clr-color-action-600});
    --border-color: var(--clr-color-action-600, #{$clr-color-action-600});
    --box-shadow: 1 1 1 var(--$clr-color-neutral-600, #{$clr-color-neutral-600});  
  }
  ```
- Exposed properties should match the native CSS property and nested elements prefixed to prevent var name collisions.
  ```scss
  :host {
    // --[optional template element]-[css property]-[optional element state]: value;
    --icon-color-hover: var(...);
    --icon-color: var(...);
    --color: var(...);
    color: var(--color);

    &:hover {
      --color: var(--color-hover); 
    }

    cwc-icon {
      --color: var(--icon-color);
    }
  }
  ```

PR #4300, opened 2020-02-19
core: api cleanup

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

## What is the new behavior?
- Renames core component prefix from `cwc-` to `cds-`
- Removes the core dev app from the project
- Update Core dependencies (storybook, web component analyzer)
- Remove font based CSS props from the component APIs
- Remove host styles from components. By removing styles from `:host` this ensures the styles are isolated in the shadow DOM and forces the consumer to use only the CSS vars defined on `:host`.
Example this prevents from being able to set a background color like this:
  ```scss
  cds-button {
    background: green; // this is invalid should not be able to style component on host tag
    --background: green; // this is valid should only be able to style via CSS props
  }
  ```
- Remove the `block` feature from button. This will be handled by the layout system.
- Fix button loading state bug. The loading button could get into a broken disabled state in the following scenario when loading state is set in this order:

1. default
2. loading
3. success
4. loading
5. success
6. button now default but stuck in disabled state

closes https://github.com/vmware/clarity/pull/4300
closes https://github.com/vmware/clarity/issues/4263

* [ ] Yes
* [x] No


src/clr-core/button/button.element.ts
```diff
@@ -84,82 +79,59 @@ export class CwcButton extends CwcBaseButton {
...
-
-    if (state === ClrLoadingState.ERROR) {
-      // in case of error, go back to default case; app should display relevant error message
-      this.loadingState = ClrLoadingState.DEFAULT;
+  private updateLoadingState(previousState?: ClrLoadingState) {
+    if (this.loadingState === ClrLoadingState.LOADING) {
+      this.privateHost.style.width = getElementWidth(this.privateHost);
+      this.disabled = true;
+    } else if (this.loadingState === ClrLoadingState.SUCCESS && previousState === ClrLoadingState.LOADING) {
+      setTimeout(() => {
```
jeeyun: Is it necessary to check for `previousState === ClrLoadingState.LOADING` here? What happens when the button enters `SUCCESS` from `DEFAULT` state? Shouldn't the button still be disabled in this case?

Maybe a broader discussion would be - do we make certain assumptions about what the previous state is? How do we want to handle cases like what I mentioned? I think the tricky part is that there is an aspect of the button that doesn't depend on previous state (disabled/enabled) and one that does (setting the width of the button to prevent button resizing).

src/clr-core/badge/badge.element.ts
```diff
@@ -41,7 +41,9 @@ export class CwcBadge extends LitElement {
 
   render() {
     return html`
-      <slot></slot>
+      <div class="private-host">
```
mathisscott: Why `.private-host`?

src/clr-core/button/button.element.scss
```diff
@@ -1,33 +1,37 @@
...
   --line-height: 1.7rem;
   --letter-spacing: 0.12em;
+  --padding: 0 #{$clr_baselineRem_0_5};
+
   background: var(--background);
   color: var(--color);
   border-color: var(--border-color);
   border-radius: var(--border-radius);
-  border-width: var(--border-width);
+  border-width: $clr_baselineRem_1px;
```
mathisscott: Right now we have a `--clr-global-borderwidth`. Do we want to drop that?

src/clr-core/button/button.element.ts
mathisscott: ➕ 1

Why do we care about the previous loading state? The only reason I can imagine us wanting to use it is in a situation where we don't want to do anything if the new state and the previous state are the same and, iirc, Lit handles that for us.

IMO, ln 118 should only check for `this.loadingState === ClrLoadingState.SUCCESS`. Managing the previous state or the cycle of states is out of scope for the type of state we want to build into these components.

src/clr-core/common/utils/dom.ts
```diff
@@ -14,18 +14,3 @@ export function getElementWidthUnless(element: HTMLElement, unless: boolean) {
   }
   return '';
 }
-export function getTranslateForChromeRenderingBugUnless(unless: boolean) {
```
mathisscott: Are these functions no longer used?

Seems we'd get some use out of the `toggleDisabledAttribute()` one. And I'm curious as to why `getTranslateForChromeRenderingBugUnless()` was in there and now not.

src/clr-core/icon/index.ts
```diff
@@ -7,10 +7,10 @@
 import { registerElementSafely } from '@clr/core/common';
 import { CwcIcon } from '@clr/core/icon-shapes';
 
-registerElementSafely('cwc-icon', CwcIcon);
+registerElementSafely('cds-icon', CwcIcon);
```
mathisscott: Are we good with the disconnect between `cds-*` and `Cwc*`?

src/clr-core/tag/tag.element.ts
```diff
@@ -40,16 +38,24 @@ export class CwcTag extends CwcBaseButton {
...
   static get styles() {
     return [baseStyles, styles];
   }
+
+  render() {
+    return html`
+    <div class="private-host">
+      <slot></slot>
+    </div>
+    ${this.hiddenButtonTemplate}
```
mathisscott: Does `hiddenButtonTemplate` always need to be here? My question/concern is that this item captures focus even when it's not clickable.

src/clr-core/icon/index.ts
coryrylan: Thanks good catch! I'll rename them.

src/clr-core/common/utils/dom.ts
coryrylan: The `toggleDisabledAttribute` I can add back but seems like it's not needed since we are using custom elements we can do `this.disabled = true|false;`. I think we usually will want to set properties internally so the attributes are reflected and both stay in sync.

The `getTranslateForChromeRenderingBugUnless` was being applied every time the instance was created and applying the same CSS style inline so I moved it to the SCSS file instead.

src/clr-core/button/button.element.scss
coryrylan: hmmm for core I would think so since we did not want border sizes to be adjustable to prevent reflow. I am not sure the implications for clr-ui

src/clr-core/badge/badge.element.ts
coryrylan: I did not know what to call this (maybe just `.badge`) but its the reasoning described in the PR. Basically we dont want styles directly on host otherwise they are exposed and can be overridden externally without using the css props. Just makes the styles more encapsulated/strict.

src/clr-core/button/button.element.ts
coryrylan: Ah sorry that should not be there, it was leftover from some debugging I believe. I'll get that removed

src/clr-core/badge/badge.element.ts
mathisscott: got it. Thanks!

src/clr-core/button/button.element.scss
mathisscott: Yeah. The problem is we are still using the SASS variable `$clr-global-borderwidth` in places. We should probably be consistent one way or the other.

src/clr-core/common/utils/dom.ts
mathisscott: I'm fine with getting rid of them. Happy about it, even. I just wanted to understand why they weren't being used. 😊

src/clr-core/tag/tag.element.ts
coryrylan: Hmm I'll double check this. Every button that extends the base button has this. It should definitely not capture focus ever and should be completely hidden. It's only used as a workaround for triggering form submit events.

coryrylan: I double checked and this works as expected and does not capture focus. If the button is set to readonly it completely removes it from the DOM.

PR #4306, opened 2020-02-21
fix: core log and track active versions

coryrylan:
## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

This adds a global API to track current active versions of Clarity Core. Core is susceptible to being loaded with multiple version types. We want a way to know if a user is trying to do this and warn them that they should only use a single version. This will start as a private undocumented API.

Issue Number: core: set global namespace version track

* [ ] Yes
* [x] No


src/clr-core/common/utils/global.ts
```diff
@@ -0,0 +1,71 @@
...
+
+function getVersion() {
+  const log: CDSLog = {
+    versions: window.CDS._version,
+    loadedElements: window.CDS._loadedElements,
+    angularVersion: getAngularVersion(),
+    userAgent: navigator.userAgent,
+  };
+
+  console.log(JSON.stringify(log, null, 2));
```
coryrylan: I log the value here so it is easy to copy from the console for devs

src/clr-core/common/utils/global.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    setupCDSGlobal();
+  });
+
+  describe('CDS.getVersion()', () => {
+    it('should log all Clarity versions found', () => {
+      window.CDS._version.push('1.0.0');
+      const log = window.CDS.getVersion();
+      expect(log.versions.length).toBe(2);
+      expect(log.versions[0]).toBe('@VERSION');
+      expect(log.versions[1]).toBe('1.0.0');
```
mathisscott: Can we test that it will catch more than one version if present?

src/clr-core/common/utils/global.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+declare global {
+  interface Window {
+    CDS: CDSGlobal;
+  }
+}
+
+export function setupCDSGlobal() {
+  if (isBrowser()) {
```
mathisscott: 😊 Aww! Someone is using one of my funky functions... ❤️

src/clr-core/common/utils/register.ts
```diff
@@ -14,12 +15,14 @@ const addElementToRegistry = curryN(
...
     } else {
       registry.define(tagName, elementClass);
+      window.CDS._loadedElements.push(tagName);
     }
   }
 );
 
 export function registerElementSafely(tagName: string, elementClass: any) {
   if (isBrowser() && existsInWindow(['customElements'])) {
+    setupCDSGlobal();
```
mathisscott: I'm wondering if it makes more sense to move `setupCDSGlobal()` here into `addElementToRegistry()`.

This function doesn't need CDSGlobal at all. And it's just state that's spread out a bit too much for my comfort zone.

src/clr-core/test-bundles/bundlesize.json
```diff
@@ -6,7 +6,7 @@
     },
     {
       "path": "./dist/test-bundles/webpack.bundle.js",
-      "maxSize": "10 kB"
+      "maxSize": "11 kB"
```
mathisscott: 😂 
We take over the world... 1 kilobyte at a time...

PR #4310, opened 2020-02-25
feat(a11y): introduce ClrAriaLiveService as public service for developers

### LINKED ISSUE #4266: Make Aria-Live service public
## Summary

Applications need to have an aria-live service to easily manage the lifecycle of messages in their applications. Clarity has attempted to handle this for applications but we have found that there are too many edge cases or uses cases to handle it correctly by default for everyone. 

* [ ] Make existing Aria Live service public for applications to use
* [ ] Remove existing internal implementations of aria-live
* [ ] Expand documentation as necessary to demonstrate known uses of aria-live where we've removed it

## Use case

Applications need aria live in certain contexts to alert the user about important changes that happen in the page, which are almost always driven by changes in the application itself (not Clarity).

## Examples

Angular Material has something similar.

## Workarounds

Apps can do it themselves, but this makes it easier and more consistent.

bdryanovski:
feat(a11y): introduce ClrAriaLiveService as public service for developers

  * Mark internal use of ClrAriaLiveService for deprecation
  * Create a public class to provide ClrAriaLiveService
  * Add documentation for ClrAriaLiveService
  * Add deprecation notice at some of the components

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4266 

* [ ] Yes
* [x] No

## Other information

Close #4266

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -112,7 +112,7 @@ export declare class ClrAlert implements OnInit, OnDestroy, AfterViewInit {
     alertTexts: QueryList<ElementRef>;
     set alertType(val: string);
     get alertType(): string;
-    get ariaLive(): AriaLivePoliteness;
+    get ariaLive(): ClrAriaLivePoliteness;
     assertive: boolean;
     closable: boolean;
     clrCloseButtonAriaLabel: string;
@@ -122,7 +122,7 @@ export declare class ClrAlert implements OnInit, OnDestroy, AfterViewInit {
     isSmall: boolean;
     off: boolean;
     polite: boolean;
-    constructor(iconService: AlertIconAndTypesService, cdr: ChangeDetectorRef, multiAlertService: MultiAlertService, commonStrings: ClrCommonStringsService, ariaLiveService: AriaLiveService);
+    constructor(iconService: AlertIconAndTypesService, cdr: ChangeDetectorRef, multiAlertService: MultiAlertService, commonStrings: ClrCommonStringsService, ariaLiveService: ClrAriaLiveService);
     close(): void;
     ngAfterViewInit(): void;
     ngOnDestroy(): void;
@@ -178,6 +178,19 @@ export declare enum ClrAlignment {
     START = 0,
     CENTER = 0.5,
     END = 1
+}
+
+export declare enum ClrAriaLivePoliteness {
+    off = "off",
+    polite = "polite",
+    assertive = "assertive"
+}
+
+export declare class ClrAriaLiveService implements OnDestroy {
+    get id(): string;
+    constructor(ngZone: NgZone, _document: any, platformId: Object);
+    announce(message: string | HTMLElement, politeness?: ClrAriaLivePoliteness): void;
+    ngOnDestroy(): void;
 }
 
 export declare enum ClrAxis {
@@ -286,7 +299,7 @@ export declare class ClrControlContainer extends ClrAbstractContainer {
 
 export declare class ClrControlError implements AfterViewInit {
     controlIdService: ControlIdService;
-    constructor(controlIdService: ControlIdService, ariaLiveService: AriaLiveService, el: ElementRef);
+    constructor(controlIdService: ControlIdService, ariaLiveService: ClrAriaLiveService, el: ElementRef);
     ngAfterViewInit(): void;
 }
 
@@ -728,7 +741,7 @@ export declare class ClrDaypicker implements AfterViewInit {
     get monthAttrString(): string;
     get updateAriaLiveYear(): string;
     get yearAttrString(): string;
-    constructor(_viewManagerService: ViewManagerService, _dateNavigationService: DateNavigationService, _localeHelperService: LocaleHelperService, commonStrings: ClrCommonStringsService, ariaLiveService: AriaLiveService);
+    constructor(_viewManagerService: ViewManagerService, _dateNavigationService: DateNavigationService, _localeHelperService: LocaleHelperService, commonStrings: ClrCommonStringsService, ariaLiveService: ClrAriaLiveService);
     changeToMonthView(): void;
     changeToYearView(): void;
     currentMonth(): void;
@@ -873,7 +886,7 @@ export declare class ClrForm {
     set labelSize(size: number);
     labels: QueryList<ClrLabel>;
     layoutService: LayoutService;
-    constructor(layoutService: LayoutService, markControlService: MarkControlService, platformId: Object, el: ElementRef, commonStrings: ClrCommonStringsService, ariaLiveService: AriaLiveService);
+    constructor(layoutService: LayoutService, markControlService: MarkControlService, platformId: Object, el: ElementRef, commonStrings: ClrCommonStringsService, ariaLiveService: ClrAriaLiveService);
     markAsDirty(updateAriaLiveText?: boolean): void;
     markAsTouched(updateAriaLiveText?: boolean): void;
     onFormSubmit(): void;
@@ -1167,7 +1180,7 @@ export declare class ClrPopoverToggleService {
 }
 
 export declare class ClrProgressBar {
-    get ariaLive(): AriaLivePoliteness;
+    get ariaLive(): ClrAriaLivePoliteness;
     assertive: boolean;
     set clrDanger(value: boolean | string);
     set clrFade(value: boolean | string);
@@ -1193,7 +1206,7 @@ export declare class ClrProgressBar {
     get successClass(): boolean;
     get value(): number;
     set value(value: number);
-    constructor(ariaLiveService: AriaLiveService);
+    constructor(ariaLiveService: ClrAriaLiveService);
     displayAriaLive(): boolean;
 }
 
@@ -1302,7 +1315,7 @@ export declare class ClrSignpostTrigger implements OnDestroy {
 }
 
 export declare class ClrSpinner implements AfterViewInit {
-    get ariaLive(): AriaLivePoliteness;
+    get ariaLive(): ClrAriaLivePoliteness;
     assertive: boolean;
     set clrInline(value: boolean | string);
     set clrInverse(value: boolean | string);
@@ -1314,7 +1327,7 @@ export declare class ClrSpinner implements AfterViewInit {
     off: boolean;
     get smallClass(): boolean;
     get spinnerClass(): boolean;
-    constructor(el: ElementRef, ariaLiveService: AriaLiveService);
+    constructor(el: ElementRef, ariaLiveService: ClrAriaLiveService);
     ngAfterViewInit(): void;
 }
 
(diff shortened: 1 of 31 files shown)


src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -49,7 +49,7 @@ export const ARIA_LIVE_TICK: number = 100;
  *
  */
 @Injectable()
```
gnomeontherun: This needs to be provided in root once so we don't get copies of it.

```
@Injectable({
  providedIn: 'root'
})
```

src/website/src/app/documentation/demos/a11y/accessibility.demo.html
```diff
@@ -0,0 +1,50 @@
...
+      </p>
+
+      <p>
+        There are limitations to how different screen readers and assistive technologies support <code class="clr-code">aria-live</code>, and we've provided a
+        service <code class="clr-code">ClrAriaLiveService</code> for your Angular application to use anytime you need to make an announcement in a fully
+        compatible way.
+      </p>
+
+      <p>
+        Clarity does not make any announcements by default, it is up to the application to do this when it makes sense. Here are a few examples of how the
```
gnomeontherun: This isn't true based on not removing aria-live.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
```
gnomeontherun: Can you add to the `deprecations.spec.ts` file 

```
describe('3.0', () => {
  it('should not do aria-live by default');
});
```

src/clr-angular/emphasis/alert/alert.ts
```diff
@@ -23,13 +23,13 @@ import { AlertIconAndTypesService } from './providers/icon-and-types.service';
...
 import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
-import { AriaLiveService, AriaLivePoliteness } from '../../utils/a11y/aria-live.service';
+import { ClrAriaLiveService, ClrAriaLivePoliteness } from '../../utils/a11y/aria-live.service';
 import { ClrAlertText } from './alert-text';
 import { Subscription } from 'rxjs';
 
 @Component({
   selector: 'clr-alert',
-  providers: [AlertIconAndTypesService, AriaLiveService],
+  providers: [AlertIconAndTypesService, ClrAriaLiveService],
```
gnomeontherun: We should remove it from all providers here, since it should be provided in root.

src/clr-angular/utils/a11y/aria-live.service.ts
Shijir: Yeah, this is necessary as we create a container element for each instantiation of this service. All `.announce` calls should source from the same container element. However, we also have `ngOnDestroy` which is a component bound lifecycle hook. That means if we make this service root module level service,  `ngOnDestroy` will probably no longer work. If that's the case, we can remove the ngOnDestroy hook even though that means we won't remove the container element. But I am not afraid of leaving one container element.

src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -370,7 +369,7 @@ export default function(): void {
           expand.expanded = true;
           expand.loading = true;
           context.detectChanges();
-          tick(ARIA_LIVE_TICK);
+          tick(100);
```
Shijir: nit: we need to export `ARIA_LIVE_TICK` again and use the same source value.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -128,7 +131,19 @@ export class AriaLiveService implements OnDestroy {
    */
   public ngOnDestroy() {
     clearTimeout(this.previousTimeout);
+    this.removeAriaLiveContainer();
+  }
+
+  /**
+   * Clear aria-live message from container
+   */
+  public clearContainer() {
```
Shijir: This method should be private.

src/clr-angular/utils/a11y/aria-live.service.ts
```diff
@@ -128,7 +131,19 @@ export class AriaLiveService implements OnDestroy {
...
+    if (isPlatformBrowser(this.platformId) && this.ariaLiveElement) {
+      this.ariaLiveElement.textContent = null;
+    }
+  }
 
+  public removeAriaLiveContainer() {
```
Shijir: This one should be private too.

golden/clr-angular.d.ts
```diff
@@ -180,6 +180,21 @@ export declare enum ClrAlignment {
...
 }
 
+export declare enum ClrAriaLivePoliteness {
+    off = "off",
+    polite = "polite",
+    assertive = "assertive"
+}
+
+export declare class ClrAriaLiveService implements OnDestroy {
+    get id(): string;
```
Shijir: Do we need to have `id` in this service? I don't see it serving any purpose.

Ideally, we should expose only announce method from this service.

src/clr-angular/utils/a11y/aria-live.service.ts
bdryanovski: ... Github crashed and lost my comment so I'm starting from scratch...

So I'm not sure that this is the best way to do it. And here are my reasons for that.

By adding this to the aria service
```ts
@Injectable({
  providedIn: 'root'
})
```
We gonna get only one singleton instance of the service and this will result in having only one aria-live container. The service is design to be a multiple instance singleton service and there is a reason for that. I will try to explain it below.

Having one aria-live container could have some strange effects on how SR sees and read messages inside. 

For example, let say we have this template
```html
<input name="username" />
<clr-error>Field is required</clr-error>
<input name="password" />
<clr-error>Field is required</clr-error>
```
What will happen here is that SR will read-only once the `Field is requried` the reason for that is that there two messages with the same content. The SR will ignore the second message because for him the container did not change. Angular detection change even won't throw an event that the container is changed cause ... in reality, it is the same content. 

So for messages that are equal to each other, the SR will ignore them even if they are made from different components.

The second issue will come with how a single container is seen by SR. For them, if the content change there is something to do. 
Let say we have two components with how different purpose. 
 
```html
<welcome-message>Long message a few lines more than two sentences</welcome-message>
<another-component>Loading user data</another-component>
```
Now if both components are rendered there a few things that could happen here. 
  * The long message will be cut at the point where the `another-component` message is sent from the `announce` method.  This could be in mid-sentence or even may not start at all
  * No matter what politeness type is set  - it will look like assertive - last message is read everything else is ignored or cut.

This change will make the politeness type absolutely unneeded  - we always will have `assertive` as type, it may seem to work as `polite` but this is only if there is no new message coming soon. 

AriaLiveService is designed to let every component handle it's own messages and not let any other component access or change them. If we share only one instance, every component will have to fight for time. This is making the work of the service unpredictable. 

This was the initial design and I may say it works pretty good for the moment if we ignore some corner cases where the component is rendered inside a template and it's hard to know where to broadcast the message. 

If we just add the `@Injectable` on top of the service and not remove the `providers`. there will be a singleton instance of the service that may not be used at all. 

`ngOnDestroy` is there to make implementation a little easier and to handle it's cleanup automatically. If we remove it we have to provide a public method for removing it from the DOM. In the case where an application is destroyed this DOM will be left hanging. When playing with the auto-refresh of Stackblitz example below, I notice that this could lead to losing the reference of the aria-live-container and throwing an error.

I'm removing the public methods below for cleaning container and removing because I have another idea for that. But we need to provide in any case a method to clean this container. I think this way the issue with rendering components in complex templates where a aria-live component is rendered before shown could be handled and solved. 

Another thing that I notice that there could be a way to force the service announcing identical messages. This right now is an issue for example in loaders or progress bars.
```ts
longProcess().subscribe((finish) => { 
  if (finish) { return this.ariaLiveService.announce('Loading is done'); }
  this.ariaLiveService.announce('Loading ...') 
})
```
Even if this is not the best practice it clearly shows that if we try to update an existing component, with the same message, we won't get what we expected.

Overall: 
  * Components could cut or suppress other components message
  * Politeness level is ignored
  * Broadcasting similar messages will be ignored 
 
I made a little demo page where we could see the effect of it https://stackblitz.com/edit/aria-live-angular-demo?file=src%2Fapp%2Fhello.component.ts

For example if we comment from `hello.component.ts`
```ts
providers: [ClrAriaLiveService],
```
We gonna have this one service for the how application. 

Pressing `One step` we could see how one message will start to be read and pressing it again will just cut it off and another message will replace it. 

Also with `Auto toggle` and `Rate` could be simulated if multiple components are broadcasting at the same time. 

Removing the comment will create multiple dom elements inside and let read one by one message. This is not clearly visible when is set to auto, because the component will be removed quickly this will make the SR stop reading.

This is the reason why this service is optional and every component that needs it to have to implement and handle it. Single instance won't work.

golden/clr-angular.d.ts
bdryanovski: `id` is the only way to know where and what is the id of the aria-container that you are using. And because we don't provide additional methods for cleaning there is no other way to clean it if you needed. 

For example, when you want to have the same message twice the only way will be to remove the content wait for a little and announce it again so the SR will pick the change.

bdryanovski: We don't handle that case and it's left to people having to handle it find a solution .. 

May be in later iteration this could be improved, but without the ID with the design for multiple instances you won't be able to find the container inside the DOM.

src/clr-angular/utils/a11y/aria-live.service.ts
gnomeontherun: Thank you for the well articulated and detailed details here. To expedite this, let me address each point.

1. Having one instance means that the Application and our components can share responsibility for what is being said. An application can silence or override our components by clearing the message and putting their own in, giving them potential control. We initially thought it made sense to isolate each usage internally, but that is the real problem we're facing that applications don't have easy access to manage a multitude of dynamic aria-live regions. One instance is really the goal based on my conversations with @chlane and this would put us there.
2. Your point about having the same message will only read once is fair, but that is also managed by the application. They can choose to make the messages different (and they should) or worst case scenario they can clear and reapply the message.
3. The point about screen readers competing for attention is true no matter what we do. By using a single instance, we give applications more control when they do want to clear a queue of messages. We also don't know when a message is done being read, so this problem exists already. Again, by letting applications manage this directly we give them the power to determine the best course of action and a single instance is easier to reason about and manage than a multitude of instances.
4. Providing in the root means the service can be tree shaken if it isn't used, this is part of the providedIn benefit as well. 
5. We can enhance this service for more ergonomics over time, such as helping stack messages. Applications don't have to use this service, and if their needs are different they can build a simple service of their own. We likely can't work for every use case and again I want to defer to the applications to manage messages.

So overall, the concerns are valid but we essentially block the application from controlling the messages of our internal components. The concerns don't go away really, but the responsibility and ability to control falls to the application to manage this.

gnomeontherun:
Also your commit message shouldn't include `BREAKING CHANGE` text, or it will get marked as one.

bdryanovski:
> Also your commit message shouldn't include `BREAKING CHANGE` text, or it will get marked as one.

Sure .. when rebasing will fix it.

golden/clr-angular.d.ts
Shijir: This is not a blocker issue. But currently, the service doesn't use to access the element. It will always keep the reference to the element. Currently, it doesn't do anything but kept in the service. And I don't think of any good reason for users to clean it up manually.

src/clr-angular/utils/a11y/aria-live.service.ts
Shijir: Don't we have that cancellable setTimeout just to prevent many consecutive messages from racing to be announced? 

If users need to announce multiple messages, they could do so by stitching those messages together and call the announce method on it. That's what we did with the form errors.

But I am hoping we could improve it later on iteratively without introducing breaking changes.

PR #4313, opened 2020-02-25
chore: update documentation and schematics for webcomponents polyfill

### LINKED ISSUE #4032: Update documentation on how to include webcomponents polyfill
## Describe the bug

According to the [documentation](https://v2.clarity.design/get-started), the webcomponents polyfill should be included in `angular.json` like this:

```
"scripts": [
  ... any existing scripts
  "../node_modules/@webcomponents/custom-elements/custom-elements.min.js",
  "../node_modules/@clr/icons/clr-icons.min.js"
]
```

When building like that, I was getting an error in IE11: 

```
customElements is undefined
```

## How to fix

I fixed the problem by instead including the webcomponents polyfill in `polyfills.ts` at the very end of the file:

```
... other polyfills
import '@webcomponents/custom-elements';
```

This is also the way I found that the polyfills are loaded in the stackblitz examples. If this is indeed the way to go and this was not some other configuration issue in my project, the documentation should be updated accordingly.

## Versions

**App**

* Angular: 8
* Clarity: 2

**Device:**

* Browser IE11

gnomeontherun:
fixes #4032

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4032

* [ ] Yes
* [ ] No

### DIFF
--- a/src/website/src/app/documentation/get-started/get-started.component.ts
+++ b/src/website/src/app/documentation/get-started/get-started.component.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -30,7 +30,8 @@ export class AppModule { }
 const HTML_IMPORTS = `
 <link rel="stylesheet" href="path/to/node_modules/@clr/ui/clr-ui.min.css">
 <link rel="stylesheet" href="path/to/node_modules/@clr/icons/clr-icons.min.css">
-<script src="path/to/node_modules/@webcomponents/custom-elements/custom-elements.min.js"></script>
+<script src="path/to/node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
+<script src="path/to/node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
 <script src="path/to/node_modules/@clr/icons/clr-icons.min.js"></script>
 `;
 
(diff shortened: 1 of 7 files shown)


.prettierignore
```diff
@@ -11,3 +11,5 @@ src/website/src/sitemap.xml
 src/clr-core/**/*.css.ts
 src/clr-angular/utils/_mixins.clarity.scss
 docs/core
+**/node_modules/
+src/schematics/
```
hippee-lee: Shouldn't we be formatting these files?

src/website/src/app/documentation/get-started/get-started.component.ts
```diff
@@ -30,7 +30,8 @@ export class AppModule { }
 const HTML_IMPORTS = `
 <link rel="stylesheet" href="path/to/node_modules/@clr/ui/clr-ui.min.css">
 <link rel="stylesheet" href="path/to/node_modules/@clr/icons/clr-icons.min.css">
-<script src="path/to/node_modules/@webcomponents/custom-elements/custom-elements.min.js"></script>
+<script src="path/to/node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
```
Shijir: I wasn't aware of these dependency changes. Do people need both of these dependencies in all cases?

gnomeontherun: This is the recommendation from @coryrylan that will be used with Core, so yes I think so.

gnomeontherun: Also, the schematic will fix this for anyone automatically, and updated the ng-add schematic too.

.prettierignore
gnomeontherun: I'm going to revert this.

PR #4323, opened 2020-02-28
fix(tabs): tab overflow and its focus management

Shijir:

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

The keyboard focus management in Tabs is failing with Angular 9 changes.

## What is the new behavior?

The keyboard focus management in Tabs works well with Angular 9 now and has other focus management improvements.

* [ ] Yes
* [x] No


src/clr-angular/utils/focus/key-focus/key-focus.ts
```diff
@@ -36,19 +45,50 @@ export class ClrKeyFocus {
...
+  get nativeElement(): HTMLElement {
+    return this.elementRef.nativeElement;
+  }
+
   get focusableItems() {
     if (this._focusableItems) {
       return this._focusableItems;
     } else {
-      return this.clrKeyFocusItems.toArray();
+      return this.clrKeyFocusItems && this.clrKeyFocusItems.toArray();
```
Jinnie: nit: Why do we need this? Empty QueryList is still a QueryList, just as an empty array is still an array. Do we use it somewhere before view has finished initializing?

src/clr-angular/layout/tabs/tab-link.directive.ts
```diff
@@ -101,4 +101,9 @@ export class ClrTabLink {
   get active() {
     return this.ifActiveService.current === this.id;
   }
+
+  @HostBinding('attr.tabindex')
```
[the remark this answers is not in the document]
Shijir: Yes, in our case, we are implementing tabs with manual activation. So we will have only one tab that should be in the tab sequence which is the active tab link itself.

src/clr-angular/layout/tabs/tabs.spec.ts
```diff
@@ -236,6 +238,95 @@ describe('Tabs', () => {
...
+      context.testComponent.inOverflow = true;
+      context.fixture.detectChanges();
+    });
+
+    afterEach(() => {
+      context.fixture.destroy();
+    });
+
+    it('reset keyFocus current to active tab index', () => {
+      const activeTabIndex = 1;
```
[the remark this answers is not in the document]
Shijir: Maybe, active tab position better?

src/clr-angular/layout/tabs/tabs.spec.ts
```diff
@@ -236,6 +238,95 @@ describe('Tabs', () => {
...
+    it('focuses last visible tab link when overflow is closed', () => {
+      context.clarityElement.querySelector('.dropdown-toggle').click();
+      context.detectChanges();
+      expect(context.clarityElement.querySelector('.dropdown-menu')).not.toBeNull();
+      context.clarityElement.querySelector('.dropdown-toggle').click();
+      context.detectChanges();
+      expect(document.activeElement).toBe(tabLinkEls[2]);
+    });
+
+    it('puts overflow trigger in tab sequence if it contains active tab link', () => {
```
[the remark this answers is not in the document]
Shijir: This is the case for when the focus is outside the tab list and the active tab link is in overflow content. So when the focus re-enters the tablist with tab key pressing, it should land on the overflow trigger and that should open the content. So this specific one is testing whether overflow trigger is tabbable or not. The next is testing the case when the focus lands on it whether it opens the content or not.

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -45,32 +40,45 @@ import { isPlatformBrowser } from '@angular/common';
...
                         </ng-container>
-                    </clr-tab-overflow-content>
+                    </div>
                 </div>
             </ng-container>
         </ul>
         <ng-container #tabContentViewContainer></ng-container>
     `,
+  host: {
+    '[class.tabs-vertical]': 'isVertical',
```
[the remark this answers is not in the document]
Shijir: oh I remember I removed the host binding when I was debugging. I can revert this change.

src/clr-angular/utils/focus/key-focus/key-focus.ts
Shijir: This was to prevent `moveTo` method from being called if `this.clrKeyFocusItems` is empty.  If the method is called while the `this.clrKeyFocusItems` is null, it would throw an error.

Shijir:
@Jinnie I've created the issue: https://github.com/vmware/clarity/issues/4332

src/clr-angular/layout/tabs/tabs.spec.ts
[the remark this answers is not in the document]
Jinnie: Yes, will be good. "Tab index" is precise, but conflicting, so "tab position" sounds like second best.

src/clr-angular/utils/focus/key-focus/key-focus.ts
Jinnie: Yes, but if it's empty it's not null. It's only null before ContentChildren has initialized. Or I'm missing something?
Just curious.

src/clr-angular/layout/tabs/tab-link.directive.ts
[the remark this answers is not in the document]
Jinnie: But when you move the focus with arrow keys to a non-active tab, you need to make this tab tabbable and need to remove focus from the active one. Or you end with two tabbable tabs.

src/clr-angular/utils/focus/key-focus/key-focus.ts
Shijir: Yes, but it's a guard against the case when `focusableItems` gets accessed even before the `KeyFocus` directive is able to register its QueryList because `QueryList` is `undefined` until `AfterContentInit` cycle. Before refactoring this code, I was running into that very error because at one point I was accessing focusable items too early. But I kept this guard because I found this preemptively necessary.

src/clr-angular/layout/tabs/tab-link.directive.ts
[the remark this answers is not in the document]
Shijir: Yes, that would have been the case if we were doing automatic activation based on a roving index. But in this case, we don't change the tabindex until a tab gets activated manually. So only an active tab link would be in tab sequence no matter where the focus is currently.
https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html

src/clr-angular/layout/tabs/tabs.ts
```diff
@@ -45,32 +40,45 @@ import { isPlatformBrowser } from '@angular/common';
...
+                        (focus)="openOverflowOnFocus()"
+                        (click)="toggleOverflowOnClick()">
                             <clr-icon shape="ellipsis-horizontal"
                               [class.is-info]="toggleService.open"
                               [attr.title]="commonStrings.keys.more"></clr-icon>
                         </button>
                     </li>
                     <!--tab links in overflow menu-->
-                    <clr-tab-overflow-content>
+                    <div class="dropdown-menu" #tabOverflow *ngIf="toggleService.open"
```
gnomeontherun: Not part of this, but we ought to move to smart popovers. 

This gets more interesting then because the popovers are in the body instead of inside the list.

src/clr-angular/utils/focus/key-focus/key-focus-item.ts
```diff
@@ -11,7 +11,7 @@ import { isPlatformBrowser } from '@angular/common';
   selector: '[clrKeyFocusItem]',
 })
 export class ClrKeyFocusItem {
-  @HostBinding('attr.tabindex') tabIndex = -1;
+  @HostBinding('attr.tabindex') tabIndex: number;
```
gnomeontherun: Does removing the default of `-1` make a difference?

src/clr-angular/layout/tabs/tabs.module.ts
```diff
@@ -16,15 +16,13 @@ import { TabsWillyWonka } from './chocolate/tabs-willy-wonka';
...
 import { ClrTabLink } from './tab-link.directive';
-import { ClrTabOverflowContent } from './tab-overflow-content';
 import { ClrTabs } from './tabs';
 import { ClrKeyFocusModule } from '../../utils/focus/key-focus/key-focus.module';
 
 export const CLR_TABS_DIRECTIVES: Type<any>[] = [
   ClrTabContent,
   ClrTab,
   ClrTabs,
-  ClrTabOverflowContent,
```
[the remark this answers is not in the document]
Shijir: Make sense. I will put it back in.

golden/clr-angular.d.ts
```diff
@@ -203,10 +203,10 @@ export declare class ClrButton implements LoadingListener {
     buttonInGroupService: ButtonInGroupService;
     get classNames(): string;
     set classNames(value: string);
-    get disabled(): any;
     set disabled(value: any);
-    get id(): string;
+    get disabled(): any;
```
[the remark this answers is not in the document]
Shijir: Yeah, I was surprised to see that some unrelated APIs changed. I will go over it again.

src/clr-angular/utils/focus/key-focus/key-focus-item.ts
Shijir: Yes. If users have some tabindexes specified on their elements, ours would override them by taking them out of the tab sequence.

Also, we needed to decouple tabindex management from focus management. They are two separate things that need to be managed differently. So let users manage their tabindex on their side, but we will take care of the focus management part on our side.

src/clr-angular/layout/tabs/tab-link.directive.ts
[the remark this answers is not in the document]
Jinnie: @gnomeontherun this is easily reproducible:
> this way you'll have two tabbable tabs in the tablist - the focused one, and the active one. This explains the observed behavior of being able to shift-tab back to the active item.

explaining the reproduction scenario in more detail:
- Press tab key to get focus on the *active* tab.
- Move focus away from active tab with arrow keys.
- Press Shift-Tab. 
  - It is expected to go back outside the tabs component.
  - Instead you are landing on the *active* tab, which is not what is expected.

I can demonstrate and we can talk about it live, so we can have some more traction on this PR.

src/clr-angular/layout/tabs/tabs.ts
Jinnie: There were some talks to deprecate the tabs overflow in the long run. Did we reach a decision in favor/against it? If we still may want to do it, maybe we don't want the smart popovers transition here.

src/clr-angular/utils/focus/key-focus/key-focus-item.ts
Jinnie: 1. I still have doubts if we do need to separate tabindex from focus. The main idea behind creating the KeyFocus directive was "separation of concern". I.e. separating the roving index pattern from the Tabs main business scenario. It made sense to be inside a directive, as focus in not handled in the angular view layer at all, and direct DOM manipulation was needed. Which is what directives are excelling at.
2. I agreed with Shijir that if this pattern is not accepted by the team, I am okay for it to go away. I just need to be sure that we understand what we are removing, why it was created this way, and that we refactor it for some obvious advantages.
3. I'd be happy to hear some comment from @coryrylan too if possible, who did the review of the directive initially and who created parts of the code in it, so he should have some good context about it.

That said - I don't mind the change. But if I have to do such a feature again, I'd still start the same way. I need to know why we prefer the "all control to the component" pattern.

src/clr-angular/utils/focus/key-focus/key-focus.ts
Jinnie: nit: Strange. I removed it locally and see no errors. But you said it was temporary. Was the early access part of a valid scenario? I'd consider the lack of check to be a "fail-fast" guard against unexpected access.
Otherwise we silently cover unexpected behavior.

src/clr-angular/layout/tabs/tab-link.directive.ts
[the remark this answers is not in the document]
Shijir: > Instead you are landing on the active tab, which is not what is expected.

Ivan, that's how it's working in the spec example.

src/clr-angular/utils/focus/key-focus/key-focus.ts
Shijir: ok I can revert it back.

src/clr-angular/layout/tabs/tab-link.directive.ts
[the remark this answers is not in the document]
Jinnie: OK, now as you're on the active tab try pressing -> arrow key again. You will either skip a tab, or get stuck.

src/clr-angular/utils/focus/key-focus/key-focus-item.ts
Shijir: If users have some tabindexes specified on their elements, ours would override them by taking them out of the tab sequence. We cannot just assume users would implement roving index on their items all the time. Tabindex management would vary from component to component. By decoupling tabindex management from focus management, we are allowing more customizability for users and making the ClrKeyFocus more generic as it has now the better separation of concerns.

Jinnie: Can you give me some practical examples of why users may need to set parallel tabindex management on the tabs? Not the content but on the tabs? Won't this be considered component hacking?
My example is that we now refuse to release modals without focus-grab, although users seem to ask for this once in a while. But we consider this to be a bad practice. Your point that the users _may_ want to have their own tab-index management of the tabs contradicts with our practices, as illustrated by the modals example.

Shijir: Why would it be considered component hacking? It's a separation of concerns; we are making it more generic by decoupling tabindex management. Why are we assuming a user would implement roving tabindex while using ClrKeyFocus? As a utility directive, it should do one thing and do it well, which in our case is to provide keyboard focus navigation. 

Again, users' tabindex management would vary from component to component. Users should have all the freedom to handle their own tabindex management. We shouldn't get in their way. In this specific case of tabs component, I noticed the exact problem where ClrKeyFocus provides roving tabindex, but conflicts with the tab component's manual activation.

As for the modal, that's a user component that is recommended to have a focus trap all the time. But ClrKeyFocus is a utility directive whose purpose is to provide keyboard focus navigation to other components. There is no authoring a11y guideline that says it has to come with its own roving tabindex management.

src/clr-angular/layout/tabs/tabs.ts
gnomeontherun: No plans currently to change overflow.

src/clr-angular/utils/focus/key-focus/key-focus-item.ts
gnomeontherun: This is an internal directive, and I'm of the mindset that generally a directive should be pretty limited in scope. The more work a particular entity does, the more complicated it is to reuse elsewhere. 

However, given it is internal and it worked fine in Angular 8, its not bad the way it is. Given the changes needed, I'm fine with this separation and think it might make it more reusable in other scenarios.

Jinnie: We only have 2 scenarios - roving index and active descendant. I initially started with active descendant and even had the two in parallel for some time. They have nothing in common :) unless the goal they want to achieve.
That said - I can confirm that the last fix addresses the issue with skipping or blocking arrow key presses after shift-tab. I also confirmed with Chris that the shift-tab intermediate focus step is acceptable (though not required) - some authoritative aria examples have it, some don't. So it gets an approval from me too.
Thanks for the fix and the discussion.

gnomeontherun:
@Shijir can you check with @chlane to get a 👍?

Issue #4319, opened 2020-02-28
Core breaks in Stackblitz


coryrylan:
## Describe the bug
In stackblitz it forces all content to compile down to ES5 and there is no way to disable this. Unfortunately this puts the code in a invalid state as TS compiles a spread operation in a Set which breaks the code. https://stackblitz.com/edit/js-hnuwde

Related issue https://stackoverflow.com/questions/41783499/does-typescript-have-an-equivalent-of-es6-sets

```
Error in /turbo_modules/@clr/core@3.0.0/common/utils/global.js (40:74)
(intermediate value).slice is not a function
```

https://github.com/vmware/clarity/blob/master/src/clr-core/common/utils/global.ts#L63

```typescript
// source
function setRunningVersion() {
  window.CDS._version = [...new Set([...window.CDS._version, '@VERSION'])];
  (document.querySelector('body') as HTMLElement).setAttribute('cds-version', window.CDS._version[0]);

  if (window.CDS._version.length > 1) {
    console.warn(
      'Running more than one version of Clarity can cause unexpected issues. Please ensure only one version is loaded.'
    );
  }
}
```

This is what stackblitz force compiles to which causes the exception
```javascript
function setRunningVersion() {
   // broken ES5 slice on Set
    window.CDS._version = new Set(window.CDS._version.concat(['3.0.0'])).slice();
    document.querySelector('body').setAttribute('cds-version', window.CDS._version[0]);
    if (window.CDS._version.length > 1) {
        console.warn('Running more than one version of Clarity can cause unexpected issues. Please ensure only one version is loaded.');
    }
}
```

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

## Expected behavior

_A clear and concise description of what you expected to happen._

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

_Add any other notes about the problem here._

mathisscott:
[...]

PR #4326, opened 2020-02-28
fix: downgrade es2015 code in version tracker

mathisscott:
• we can't count on having the proper version of typescript, so i got rid of sets
• i added a bunch of guards to over-protect vs. potential SSR breakages
• tested in ngUniversal SSR build; it works

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4319 

## What is the new behavior?

Stackblitz is strugglin'. They force the code editors to a version of typescript that doesn't support `Set` (or rather compiles Set into broken ES5.

As a result, I'm making our code less awesome so things will work in Stackblitz again. I also ran this through SSR. It works as of ng9.0.0.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/common/utils/global.ts
+++ b/src/clr-core/common/utils/global.ts
@@ -33,13 +33,11 @@ export function setupCDSGlobal() {
 }
 
 function initializeCDSGlobal() {
-  if (!window.CDS) {
-    window.CDS = {
-      _version: [],
-      _loadedElements: [],
-      getVersion,
-    };
-  }
+  window.CDS = window.CDS || {
+    _version: [],
+    _loadedElements: [],
+    getVersion,
+  };
 }
 
 function getVersion() {
@@ -49,18 +47,21 @@ function getVersion() {
     angularVersion: getAngularVersion(),
     userAgent: navigator.userAgent,
   };
-
   console.log(JSON.stringify(log, null, 2));
   return log;
 }
 
 function getAngularVersion() {
-  const appRoot = document.querySelector('[ng-version]');
+  const appRoot = document && document.querySelector('[ng-version]');
   return appRoot ? appRoot.getAttribute('ng-version') : undefined;
 }
 
 function setRunningVersion() {
-  window.CDS._version = [...new Set([...window.CDS._version, '@VERSION'])];
+  const loadedVersion = '@VERSION';
+  if (window.CDS._version.indexOf(loadedVersion) < 0) {
+    window.CDS._version.push(loadedVersion);
+  }
+
   (document.querySelector('body') as HTMLElement).setAttribute('cds-version', window.CDS._version[0]);
 
   if (window.CDS._version.length > 1) {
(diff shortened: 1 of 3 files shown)


src/clr-core/common/utils/global.ts
```diff
@@ -50,20 +54,28 @@ function getVersion() {
...
+  if (!window) {
+    return;
+  }
+  if (window.CDS._version.indexOf(loadedVersion) < 0) {
+    window.CDS._version.push(loadedVersion);
+  }
   (document.querySelector('body') as HTMLElement).setAttribute('cds-version', window.CDS._version[0]);
 
-  if (window.CDS._version.length > 1) {
+  if (console && console.warn && window.CDS._version.length > 1) {
```
coryrylan: I think console is available in standard node env so there is no need to check for it

src/clr-core/common/utils/global.ts
```diff
@@ -50,20 +54,28 @@ function getVersion() {
...
+  if (console && console.log) {
+    console.log(JSON.stringify(log, null, 2));
+  }
   return log;
 }
 
 function getAngularVersion() {
-  const appRoot = document.querySelector('[ng-version]');
+  const appRoot = document && document.querySelector('[ng-version]');
   return appRoot ? appRoot.getAttribute('ng-version') : undefined;
```
coryrylan: Since all this code is guarded by `isBrowser()` here https://github.com/vmware/clarity/blob/f4c2610acee7d37407c1cbe0dc4d125b7013721e/src/clr-core/common/utils/global.ts#L29 do we need have the window/document checks? Or could they all use the same `isBrowser()` function?

mathisscott: We might be able to use the isBrowser function. But I was sledgehammering it because I'm tired of fighting SSR. I'm willing to give isBrowser a shot. But I don't really want to have to come back and fix this over again.

mathisscott: Note also that this issue comes from Stackblitz and Vuepress work.

All of this works in ng-universal. Even before the PR...

coryrylan: Ok, no worries on the SSR then, if we can get the stackblitz issue in then we can go ahead get the PR merged. I can block off time to do a in-depth check with core and a handfull of SSR/node setups. Thanks for fixing this!

golden/clr-angular.d.ts
```diff
@@ -160,8 +160,8 @@ export declare class ClrAlertsPager implements OnInit, OnDestroy {
     set currentAlert(alert: ClrAlert);
     get currentAlert(): ClrAlert;
     currentAlertChange: EventEmitter<ClrAlert>;
-    set currentAlertIndex(index: number);
     get currentAlertIndex(): number;
+    set currentAlertIndex(index: number);
```
gnomeontherun: Please check typescript installed is correct, these changes shouldn't occur.

PR #4339, opened 2020-03-04
fix(dropdown): let user overwrite ClrDropdownItem id

### LINKED ISSUE #3795: Dropdown is overriding `id` provided to `clrDropdownItem`
## Describe the bug

In Clarity version 2, we can't provide custom `id` to dropdown item.
If we explicitly provide `id`, it overrides it with some autogenerated id. (for instance. `clr-id-12`)

## How to reproduce

1.  Go to 'https://stackblitz.com/edit/clarity-light-theme-v2-upskue'
2.  Note that `action1` id has been added to `Action 1` in HTML code (https://stackblitz.com/edit/clarity-light-theme-v2-upskue?file=src%2Fapp%2Fapp.component.html)
3.  Click on 'DROPDOWN' to open dropdown menu
4. Observe the `id` of `Action 1` dropdown item.
4. It has been overridden to another ID that starts with `clr-id-`

## Expected behavior

If `id` has been provided to any element, it should use that `id` instead of overriding.

## Versions

**App**

* Angular: ^8.0.0
* Clarity: ^2.0.0

## Additional notes

This behavior breaks our automation test suite and other business logic.

bdryanovski:
Move the code that set the `id` for `ClrDropdownItem` from `DropdownFocusHandler ` into the component. This way we could overwrite the `id` from the outside and keep the same functionality 

## PR Checklist

## PR Type

Currently, the `id` for every dropdown item is set by the focus service. This is making hard to overwrite it from outside. This change is moving this into the component itself to handle it. And the focus service is no longer responsible for setting the id.

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3795 

* [ ] Yes
* [x] No

## Other information

Close #3795

### DIFF
--- a/src/clr-angular/popover/dropdown/dropdown-item.spec.ts
+++ b/src/clr-angular/popover/dropdown/dropdown-item.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -33,6 +33,17 @@ export default function(): void {
     spec(ClrDropdownItem, SimpleTest, null, {
       // Dummy dropdown provider, I don't even need a single property or method on it at the moment.
       providers: [{ provide: ClrDropdown, useValue: {} }, ROOT_DROPDOWN_PROVIDER],
+    });
+
+    it('sets the id to the unique generated id', function(this: Context) {
+      const id = 'myId';
+      this.clarityElement.setAttribute('id', id);
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('id')).toBe(id);
+    });
+
+    it('should have id by default', function(this: Context) {
+      expect(this.clarityElement.getAttribute('id')).not.toBe(null);
     });
 
     it('adds the menuitem role to the host', function(this: Context) {
--- a/src/clr-angular/popover/dropdown/dropdown-item.ts
+++ b/src/clr-angular/popover/dropdown/dropdown-item.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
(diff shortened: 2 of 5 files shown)


src/clr-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -27,8 +29,11 @@ export class ClrDropdownItem implements AfterViewInit {
     private el: ElementRef<HTMLElement>,
     private _dropdownService: RootDropdownService,
     private renderer: Renderer2,
-    private focusableItem: FocusableItem
-  ) {}
+    private focusableItem: FocusableItem,
+    @Inject(UNIQUE_ID) public id: string
+  ) {
+    console.log(focusableItem);
```
Jinnie: nit: console.log forgotten?

bdryanovski: rm

src/clr-angular/popover/dropdown/dropdown-item.spec.ts
```diff
@@ -35,6 +36,17 @@ export default function(): void {
       providers: [{ provide: ClrDropdown, useValue: {} }, ROOT_DROPDOWN_PROVIDER],
     });
 
+    it('sets the id to the unique generated id', function(this: Context) {
+      const id = this.fixture.debugElement.injector.get(UNIQUE_ID, 'not_found');
+      this.clarityElement.setAttribute('id', id);
+      this.detectChanges();
+      expect(this.clarityElement.getAttribute('id')).toBe(id);
+    });
```
gnomeontherun: Let's not get the injector here and just make a static string. 

```
    it('sets the id to the unique generated id', function(this: Context) {
      const id = 'custom-id';
      this.clarityElement.setAttribute('id', id);
      this.detectChanges();
      expect(this.clarityElement.getAttribute('id')).toBe(id);
    });
```

src/clr-angular/popover/dropdown/dropdown-item.ts
```diff
@@ -27,7 +29,8 @@ export class ClrDropdownItem implements AfterViewInit {
     private el: ElementRef<HTMLElement>,
     private _dropdownService: RootDropdownService,
     private renderer: Renderer2,
-    private focusableItem: FocusableItem
+    private focusableItem: FocusableItem,
+    @Inject(UNIQUE_ID) public id: string
```
gnomeontherun: The dropdown focus handler service already has a unique_id defined there, do we need it here again? Can't we just lean on that instead of having two unique ids?

bdryanovski: You are right. I just didn't want to go to this rabbit hole of refactoring. 

I think that the`FocusableItem` doesn't need to have the `id` it's never used anywhere. And it's hiding it deep inside the codebase. It's a component responsibility to handle the HTML id. 

But because I don't want to make it a braking change now. Can we mark `FocusableItem.id` for depreciation in the next version? 

```ts
export abstract class FocusableItem {
  /** @deprecated since 3.0, remove in 4.0 */
  id: string;
  // ....
```

Right now the code below is there just to handle this unused anywhere else field. 
```ts
  @Input('id')
  set dropdownItemId(value: string) {
    this.focusableItem.id = value;
  }
  get dropdownItemId() {
    return this.focusableItem.id;
  }
```

PR #4352, opened 2020-03-10
fix: id="null" for clr-buttons

KingMario:
Signed-off-by: KingMario <gcyyq@hotmail.com>

## PR Checklist

* [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

There could be many buttons with the same id.

## What is the new behavior?

If ID is not set, the button will have no id attribute.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/button/button-group/button.spec.ts
+++ b/src/clr-angular/button/button-group/button.spec.ts
@@ -19,7 +19,6 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
   template: `
         <clr-button
             #button1
-            id="button1"
             type="button"
             name="button1"
             id="button1"
@@ -267,8 +266,8 @@ export default function(): void {
       });
 
       it('sets the id correctly', () => {
-        expect(buttons[0].id).toBe('null');
-        expect(buttons[1].id).toBe('null');
+        expect(buttons[0].id).toBe('');
+        expect(buttons[1].id).toBe('');
         expect(buttons[2].id).toBe('button3');
       });
 
--- a/src/clr-angular/button/button-group/button.ts
+++ b/src/clr-angular/button/button-group/button.ts
@@ -15,13 +15,13 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
   selector: 'clr-button',
   template: `
         <ng-template #buttonProjectedRef>
-            <button 
-                [class]="classNames" 
+            <button
+                [class]="classNames"
                 (click)="emitClick($event)"
                 [attr.type]="type"
                 [attr.name]="name"
                 [attr.disabled]="disabled"
-                [id]="id">
+                [attr.id]="id">
                 <span class="spinner spinner-inline" *ngIf="loading"></span>
                 <ng-content></ng-content>
             </button>


KingMario:
@bdryanovski

Generally, we'll set an ID for `input` so that it can be used for a `label[for]`, but rarely set an ID for a button. Using many `clr-button` without setting the ids will cause the buttons to have the same `id="null"`(no, it's not ignored by the template, see: https://stackblitz.com/edit/clarity-light-theme-v2-rxak4k), which violates the no-same-id rule for XML.

To fix it with the current codebase, one will have to set IDs for all the buttons.

Or we can set auto-generated IDs instead.

bdryanovski:
@KingMario, I don't see the need of having an auto-generated ID for buttons - The reason for that is if they are auto-generated it will be hard to know what are they without making a DOM search/selection. Also by default buttons don't require to have IDs unless you want to address them in some way if this is the case you will need to set it manually. 

But I see the problem with the `id="null"` - I think that we have to focus on removing the `null` from the value and adding additional ID that will not be used.

KingMario:
@bdryanovski 

I don't know whether there is an easy way to set an optional ID for the button element. If the ID is not set for the `clr-button` manually, the ID prop should be removed from the button element, and if the ID is set for the `clr-button` , set it to the button element.

If this is not approachable, I think an auto-generated ID is better than the static `null`.

KingMario:
@hippee-lee 

The `UNIQUE_ID` is normally used for components that don't have an `Input('id')`. I think using it will be a little more complicated in the case of `clr-button` component.

bdryanovski:
Like I said I think that the problem is not in having a uniq-id, but having `id=null` as ID. If there is no ID passed from outside, buttons don't have to create one from scratch and apply it.

bdryanovski:
@KingMario 

`UNIQUE_ID` is used to provide auto-incremental ID to components as a default ID or in the cases where we require to have ID for every element.

But they are auto-incremented - so you won't know what number you will get, it depends on the order of the calls.

KingMario:
> Like I said I think that the problem is not in having a uniq-id, but having id=null as ID. If there is no ID passed from outside, buttons don't have to create one from scratch and apply it.

It's not easy to achieve it, is it?

KingMario:
The problem is not in **NOT** having a uniq-id, but having id=null as ID.

But I guess having a uniq-id is the simplest resolution.

bdryanovski:
The simplest solution that I got with quick prototyping is from this:
```ts
private _id: string = null;
```
to this:
```ts
private _id: string = '';
```

Then this `document.querySelectorAll('#null')`, will result into `NodeList []`. But still will create an attribute `id` without any value

```html
<button id>something</button>
```

I think the issue here is that we set attributes to `null` to tell Angular to not add them, but it seems that `id` attribute is a bit special and it's cast to `String` so this is the reason for getting it into the template.

By changing the default value of `_id` you won't get `null` anymore.

KingMario:
But a simple HTML document like this still cannot pass [the HTML validator](https://validator.w3.org/#validate_by_upload).

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Document</title>
</head>
<body>
  <button class="btn" id>I have no ID?</button>
  <button class="btn" id>I have no ID?</button>
  <button class="btn" id>I have no ID?</button>
</body>
</html>
```

image

bdryanovski:
I know, this is not the perfect solution like I said you will get `<button id></button>`.

KingMario:
@bdryanovski, @hippee-lee 

Force pushed by using `UNIQUE_ID`.

For an unknown reason, I cannot run test locally.

Feel sleepy, and will see the CI job results tomorrow.

Thx

bdryanovski:
If you provide some errors I could help you with the tests

KingMario:
@bdryanovski Thanks.

Here are some of the errors:

```
22 import { KeyCodes } from '@clr/core/common';
                            ~~~~~~~~~~~~~~~~~~
src/clr-angular/data/tree-view/tree-node.ts:28:26 - error TS2307: Cannot find module '@clr/core/common'.

28 import { KeyCodes } from '@clr/core/common';
                            ~~~~~~~~~~~~~~~~~~
src/clr-angular/emphasis/alert/alert.ts:71:65 - error TS2339: Property 'keys' does not exist on type 'ClrCommonStringsService'.

71   @Input() clrCloseButtonAriaLabel: string = this.commonStrings.keys.alertCloseButtonAriaLabel;
                                                                   ~~~~
src/clr-angular/emphasis/alert/providers/icon-and-types.service.spec.ts:78:58 - error TS2339: Property 'keys' does not exist on type 'ClrCommonStringsService'.
```

bdryanovski:
You need to build `core` first and then you could run `test` without any issues. We will soon update the documentation about the topic, is not the first time that someone hit it.

Goals here are:
  * Not set ID if not needed (auto-generated IDs are not an option)
  * Prevent reflected IDs (double IDs)

So the change must be something like that.

Change `[id]="id"` to `[attr.id]="id"`
```ts
@Component({
  selector: 'clr-button',
  template: `
        <ng-template #buttonProjectedRef>
            <button 
               // ...
                [attr.id]="id">
               // ...
            </button>
        </ng-template>
    `,
})
```

We need to remove the ID from the `clr-button` and pass it down.

```ts
  private _id: string = null;
  @HostBinding('attr.id') hostId: string;
  @Input('id') 
  set id(value: string) {
    if (typeof value === 'string') {
      this._id = value;
      this.hostId = null
    }
  }
  get id(): string {
    return this._id;
  }
```
The code above will remove the ID from the host component and set it into the inner button. This way we won't have double IDs.

This solution won't force us to set the auto-generated ID.

[Here](https://stackblitz.com/edit/clarity-v3-light-theme-cacqsw) is a stackblitz demo.


PS.
Thanks to @Jinnie for the shot in the dark about `[attr.id]`.

KingMario:
@bdryanovski

neat solution! thanks

A build error occurs:

image

And if I suppress it by `// tslint:disable-next-line: no-unused-variable`, npm `format` script will fail:

image

KingMario:
@bdryanovski 

I found another way to do so:

remove `[id]` binding for the button and add a ref to the button, then:

```
  @Input('id')
  set id(value: string) {
    if (typeof value === 'string') {
      this.elm.nativeElement.id = value;
    }
  }
```

KingMario:
@bdryanovski @hippee-lee 

I tried and found that [attr.id] will do the trick.

Updated and please review.

Thanks

Issue #4196, opened 2020-01-13
Syntax Error in @clr/core when using Jest


whizkidwwe1217:
## Describe the bug

After the @clr/core dependency of @clr/angular was introduced in 3.0, Jest is now throwing a syntax error coming from the new package. 

```
node_modules\@clr\core\fesm2015\clr-core.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { __decorate, __metadata, __awaiter } from 'tslib';
                                                                                                    ^

    SyntaxError: Unexpected token {

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:471:17)    
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:513:25)
      at node_modules/@clr/angular/bundles/clr-angular.umd.js:2:156
      at Object.<anonymous> (node_modules/@clr/angular/bundles/clr-angular.umd.js:5:2)

Test Suites: 2 failed, 4 passed, 6 total
Tests:       16 passed, 16 total
```

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

## Expected behavior

_A clear and concise description of what you expected to happen._

## Versions

**App**

* Angular: [e.g. 6]
* Node: [e.g. 8.10.0]
* Clarity: [e.g. 0.12.5]

**Device:**

* Type: [e.g. MacBook]
* OS: [e.g. iOS]
* Browser [e.g. chrome, safari]
* Version [e.g. 22]

## Additional notes

_Add any other notes about the problem here._

mathisscott:
Hi @whizkidwwe1217 
Is it possible to reproduce this on a stackblitz to help us investigate what may be causing it?

Thanks,

Scott

whizkidwwe1217:
Hi @mathisscott I don't know how to run unit tests in StackBlitz but you can clone https://github.com/whizkidwwe1217/jest-test. It's just very simple Angular app generated using `ng new`. Removed all karma files and replaced with Jest. To run, simply execute `npm install` and then `npm test`. Thank you.

As you can see in the spec, @clr/core is throwing error when the ClarityModule is added to the imports. It was working fine in Clarity v2.
[...]

Issue #4094, opened 2019-12-06
TypeError: Class constructor CommonStringsService cannot be invoked without 'new'


spitfjre:
## Describe the bug

After updating to version `3.0.0-next.4` just running `ng serve` will result in no compile errors. But you get a runtime error with CommonStringsService.

```
Error: Uncaught (in promise): TypeError: Class constructor CommonStringsService cannot be invoked without 'new'
TypeError: Class constructor CommonStringsService cannot be invoked without 'new'
    at new ClrCommonStringsService (clr-angular.js:707)
    at ClrCommonStringsService_Factory (clr-angular.js:709)
    at _callFactory (core.js:20068)
    at _createProviderInstance (core.js:20026)
    at resolveNgModuleDep (core.js:20001)
    at NgModuleRef_.push../node_modules/@angular/core/fesm5/core.js.NgModuleRef_.get (core.js:20677)
    at resolveNgModuleDep (core.js:20006)
    at NgModuleRef_.push../node_modules/@angular/core/fesm5/core.js.NgModuleRef_.get (core.js:20677)
    at resolveDep (core.js:21048)
    at createClass (core.js:20922)
    at resolvePromise (zone.js:852)
    at resolvePromise (zone.js:809)
    at zone.js:913
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:423)
    at Object.onInvokeTask (core.js:25977)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:422)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.runTask (zone.js:195)
    at drainMicroTaskQueue (zone.js:601)
```

Is clarity `3.x` intended to work with angular versions pre `9.x`, e.g. `8.2.0`?

## How to reproduce

Run `ng serve` and and open browser console.

## Expected behavior

It should just work?

## Versions

**App**

* Angular: 8.2.0
* Node: 13.3.0
* Clarity: 3.0.0-next.4

coryrylan:
The current Clarity 3.0.0-next.4 will work with Angular 8.x until we finish upgrading to Angular 9. This should be working currently and looks like a bug. What version of the Angular CLI is your project running?

spitfjre:
I will come back to you on Monday.

jkleczkowski:
Hi. I have the same issue. My package.json
```json
{
  "name": "web",
[...]

PR #4364, opened 2020-03-12
fix: adding .js extension to exported *.css.js file

mathisscott:
• this was throwing off some versions/configurations of Jest
• the module loader was looking for a CSS file

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

We have to export the base element styles. These are generated `*.css.js` styles.

Some configs/implementations of Jest don't try to look for `*css.js`. They stop looking at `*.css`.

Issue Number: N/A

## What is the new behavior?

I added the full path. This is the only Clarity issue I've been able to find that impedes Jest configuration. Every other issue in #4094 and #4196 are (thus far) solvable by properly configuring an application.

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/common/index.ts
+++ b/src/clr-core/common/index.ts
@@ -24,4 +24,4 @@ export * from './mixins/css-helpers';
 export * from './mixins/unique-id';
 export * from './mixins/apply-mixins';
 export * from './interfaces';
-export { styles as baseStyles } from './base/base.element.css';
+export { styles as baseStyles } from './base/base.element.css.js';


mathisscott:
I have no idea if we want to leave #4094 and #4196 open after this is merged.

No other issues in those tickets are caused by or related to Clarity.

src/clr-core/common/index.ts
```diff
@@ -24,4 +24,4 @@ export * from './mixins/css-helpers';
 export * from './mixins/unique-id';
 export * from './mixins/apply-mixins';
 export * from './interfaces';
-export { styles as baseStyles } from './base/base.element.css';
+export { styles as baseStyles } from './base/base.element.css.js';
```
coryrylan: This will fix the issue temporarily for Jest but once we pull Core components into clr-angular it will break again. We should update all the relative import paths in core to have `.js` at the end like this one. This is a fundamental issue with TypeScripts output not including the `.js` file extension. See discussion here https://github.com/microsoft/TypeScript/issues/16577 

You can see here lit-html and lit-element do the same thing to get around this issue. 
https://github.com/Polymer/lit-element/blob/master/src/lit-element.ts#L18 

This would make this a pretty large PR so we can merge this to fix the Jest issue now and I can open a new issue describing the steps to address the rest or we tackle this now in this PR. Either way is fine.

mathisscott:
@gnomeontherun 
See #4365

PR #4368, opened 2020-03-14
A11y on Clarity Datagrids

ghost:
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Current Behavior enables a11y on Clarity Datagrids, allowing for keyboard resizability on Datagrid columns, along with screen reading functionality triggered on column resizing. As of now, we consider the feature to be in beta, but with time we can always improve the feature and the functionality overall.  

Issue Number: N/A

* [ ] Yes
* [x] No

## Other information

Feature was also a part of the Borathon Contest.

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-column-separator.ts
+++ b/src/clr-angular/data/datagrid/datagrid-column-separator.ts
@@ -1,25 +1,35 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import { DOCUMENT } from '@angular/common';
-import { Component, Inject, Renderer2 } from '@angular/core';
-
+import { AfterViewInit, Component, ElementRef, Inject, NgZone, Renderer2, ViewChild } from '@angular/core';
 import { ClrDragEvent } from '../../utils/drag-and-drop/drag-event';
 import { ColumnResizerService } from './providers/column-resizer.service';
 import { TableSizeService } from './providers/table-size.service';
-import { UNIQUE_ID_PROVIDER, UNIQUE_ID } from '../../utils/id-generator/id-generator.service';
+import { UNIQUE_ID, UNIQUE_ID_PROVIDER } from '../../utils/id-generator/id-generator.service';
+import { DragEventType } from '../../utils/drag-and-drop/interfaces/drag-event.interface';
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 
 @Component({
   selector: 'clr-dg-column-separator',
   template: `
-    <div class="datagrid-column-handle" aria-hidden="true"
-      clrDraggable 
-      [clrGroup]="columnSeparatorId" 
-      (clrDragStart)="showTracker(resizeTrackerEl)" 
-      (clrDragMove)="moveTracker($event, resizeTrackerEl)" 
-      (clrDragEnd)="hideTracker(resizeTrackerEl)"></div>
+    <button
+      type="button"
+      class="datagrid-column-handle"
+      [attr.aria-label]="commonString.keys.columnResizeHandlerAriaLabel"
+      [attr.aria-describedby]="descriptionId"
+      clrDraggable [clrGroup]="columnSeparatorId"
+      (clrDragStart)="showTracker(resizeTrackerEl)"
+      (clrDragMove)="moveTracker($event, resizeTrackerEl)"
+      (clrDragEnd)="hideTracker(resizeTrackerEl)"
+      #columnHandle>
+    </button>
+    <span class="clr-sr-only"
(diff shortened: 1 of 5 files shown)


Shijir:
Hi @aliitani, thanks a lot for this contribution. This is a really cool feature to have in Datagrid. I think it's worth bringing our Clarity designers and a11y experts' attention to this PR and discuss how it solves the problem. But the first thing that grabbed my attention in this PR is that you are incorporating aria-live to announce the updated widths of columns after resizing. What was the reason for that? The reason why I am asking this is that I see this feature as a convenience feature for only sighted keyboard users.

aliitani:
Hi @Shijir, thanks for giving the PR a look! That was my first impression as well. However, this feature was worked on for Borathon, and since this year it was a11y focused, we wanted to make sure we touched on all aspects of a11y (or possibly majority of them). I can factor it out to limit the scope of the PR?

Shijir:
Hi @aliitani, thank you. I had to consult with @chlane on this PR. When I went over it, there were quite a few things to update and reorganize. I hope you wouldn't mind following through them.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -1,43 +1,86 @@
...
-      (clrDragEnd)="hideTracker(resizeTrackerEl)"></div>
+    <div class="datagrid-column-handle" aria-live="polite" tabindex="0" #columnHandle
+      clrDraggable
+      [clrGroup]="columnSeparatorId"
+      (clrDragStart)="showTracker(resizeTrackerEl)"
+      (clrDragMove)="moveTracker($event, resizeTrackerEl)"
+      (clrDragEnd)="hideTracker(resizeTrackerEl)"
+      (keydown)="onKeyPress($event)"
+      (keyup)="onKeyUp($event)">
+      <span class="clr-sr-only" #ariaTextContainer></span>
```
Shijir, line 29: No need for `#ariaTextContainer` template id. We could simply add the following content in this span element:
```
<span class="clr-sr-only" [attr.id]="descriptionId">
        Not necessary to use this button
</span>
```
As we know, unsighted users will stumble upon the resize button, but we could let them know that they don't need to use the button with the help this off screen text.

src/clr-angular/data/datagrid/datagrid-column-separator.ts
```diff
@@ -1,43 +1,86 @@
...
-      (clrDragStart)="showTracker(resizeTrackerEl)" 
-      (clrDragMove)="moveTracker($event, resizeTrackerEl)" 
-      (clrDragEnd)="hideTracker(resizeTrackerEl)"></div>
+    <div class="datagrid-column-handle" aria-live="polite" tabindex="0" #columnHandle
+      clrDraggable
+      [clrGroup]="columnSeparatorId"
+      (clrDragStart)="showTracker(resizeTrackerEl)"
+      (clrDragMove)="moveTracker($event, resizeTrackerEl)"
+      (clrDragEnd)="hideTracker(resizeTrackerEl)"
+      (keydown)="onKeyPress($event)"
```
Shijir, line 27: Unfortunately, we cannot use these template even bindings on `keydown` and `keyup` events here as they would trigger change detection on each `keydown` event. We need to use `Renderer2.listen` with `NgZone.runOutisdeAngular`

Shijir (changes requested):
Hi @aliitani, sorry for this late review. This is my partial review. There are several things we need to address. I will be happy to follow through these changes with you and help to implement them. Thank you!

.npmrc
```diff
@@ -1 +1,2 @@
-save-exact=true
\ No newline at end of file
+save-exact=true
```
Shijir, line 1: Please remove changes in this file

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@clr/clarity",
-  "version": "3.0.0-next.6",
+  "version": "3.0.0-next.7",
```
Shijir, line 3: No need to submit package-lock.json changes

aliitani:
Hi Shijir - Thanks for taking the time! I updated the pull request with the changes you went over.

Shijir (changes requested):
@aliitani this is coming along great. I appreciate your work on this feature. We still have a few things to improve so we can make the feature even better. Please let me know if you have any questions.

aliitani:
Hi Shijir - I went over the changes you point out

src/clr-angular/data/datagrid/datagrid-column-separator.ts
chlane: In case we have to explain why - the screen reader cursor points to things in the DOM. CSS truncation doesn't remove anything from the DOM so this technique only works for sighted users

Shijir:
Hi @aliitani, thank you for the changes. I will add some code refactoring on top of your changes.

aliitani:
Hi @Shijir, thanks for considering my feat

gnomeontherun:
I'm going to close in favor of #4535, which incorporates this work. Thank you @aliitani!

PR #4375, opened 2020-03-16
build: update package-lock 

bdryanovski:
Update `acorn` & `minimist` to prevent security warning

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/gemini/package-lock.json
+++ b/gemini/package-lock.json
@@ -34,11 +34,6 @@
         "negotiator": "0.6.2"
       }
     },
-    "acorn": {
-      "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-1.2.2.tgz",
-      "integrity": "sha1-yM4n3grMdtiW0rH6099YjZ6C8BQ="
-    },
     "acorn-node": {
       "version": "1.8.2",
       "resolved": "https://registry.npmjs.org/acorn-node/-/acorn-node-1.8.2.tgz",
@@ -1649,9 +1644,9 @@
       },
       "dependencies": {
         "acorn": {
-          "version": "5.7.4",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.7.4.tgz",
-          "integrity": "sha512-1D++VG7BhrtvQpNbBzovKNc1FLGGEE/oGe7b9xJm/RFHMBeUaUGpluV9RLjZa47YFdPcDAenEYuq9pQPcMdLJg=="
+          "version": "7.1.1",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.1.1.tgz",
+          "integrity": "sha512-add7dgA5ppRPxCFJoAGfMDi7PIBXq1RtGo7BhbLaxwrXPOmw8gq48Y9ozT01hUKy9byMjlR20EJhu5zlkErEkg=="
         }
       }
     },
@@ -2167,6 +2162,13 @@
         "foreach": "^2.0.5",
         "isarray": "0.0.1",
         "object-keys": "^1.0.6"
+      },
+      "dependencies": {
+        "acorn": {
+          "version": "7.1.1",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.1.1.tgz",
+          "integrity": "sha512-add7dgA5ppRPxCFJoAGfMDi7PIBXq1RtGo7BhbLaxwrXPOmw8gq48Y9ozT01hUKy9byMjlR20EJhu5zlkErEkg=="
+        }
       }
     },
     "fast-deep-equal": {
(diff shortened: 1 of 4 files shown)


bdryanovski:
Will, have to check it ...

Also, `minimist` is another one that needs to be updated.

gemini/package-lock.json
```diff
@@ -3911,6 +3905,13 @@
         "connect-logger": "0.0.1",
         "lodash": "^4.11.1",
         "minimist": "1.2.0"
+      },
+      "dependencies": {
+        "minimist": {
+          "version": "1.2.0",
```
gnomeontherun: This appears to be the wrong version here.

package-lock.json
```diff
@@ -31250,6 +31256,12 @@
           "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
           "integrity": "sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A==",
           "dev": true
+        },
+        "minimist": {
+          "version": "1.2.0",
```
gnomeontherun: Same here, can try to run `npm dedupe`.

gemini/package-lock.json
gnomeontherun: Isn't this still the wrong version? Shouldn't it be greater than 1.2.0?

package-lock.json
bdryanovski: Again here there are packages that require the exact version for example
`cypress@3.4.1 > minimist@1.2.0`

gemini/package-lock.json
bdryanovski: `lite-server` requires this exact version.  

So to auto-update it we have to update `lite-server` version.  That is a dependency of gemini.

We are on the latest patch `7.4.1` the next version is `7.5.0` - I may try testing it tomorrow if it will work for us? 

Updating it by hand may be overwritten the next time someone updates some of the main packages. 

I will try to update as much as I can the packages but will take a little time (for testing).

bdryanovski:
It seems that there is no way to tell Github 'Don't show me this again' in the case that this are development dependencies that are with a minor security risk.

gnomeontherun:
This is the fun dance with dependencies. We might have to bump more. See if you can bump the other projects depending on acorns to fix this fully. Otherwise we'll still get the warnings in GitHub.

bdryanovski:
The latest version `gemini@7.5.2` still depends on the fix version of `minimist@1.2.0` so even updating the package will have no effect. 
 
Angular 9 has it as a dependency, fix version of minimist. Something like `0.8.x`. They have open [issue](https://github.com/angular/angular/issues/36104). 

I didn't want to use `npm-force-resolutions` but it seems that it's working best in our case. So if someone has another idea - feel free to try it.

What I did here is to use `npm-force-resolutions` to check and auto-patch the package-lock file for the minimum version for a given package. 

**THIS WORK ONLY ON `npm install`** , so running `npm ci` won't have any effect on the package-lock.json file.

`acorn` is already patched in the `root/package-lock.json` file and for some reason, `npm-force-resolutions` is breaking the tree for this package (is removing one instance of the package for some reason - or it could be some of my tests?!). This is the reason that the `acron` is mention only inside `gemini/package.json`.

If I want to be honest, we could even remove the `"preinstall": "npx npm-force-resolutions",` from both `package.json` cause I already generated patched version of the lock file. The `preinstall` step is there for the case that somewhere someone run `npm install` and this may create a new `package-lock.json` file without patched version. 

It seems that all effected packages out there have an open issue for both dependencies, it is a matter of time before they are patched.

gnomeontherun:
So you're saying we can't resolve this until all dependencies update? If so, let's hold off and do it once to minimize drift.

bdryanovski:
That is an option. 

We could not update the Gemini or few other packages that have a dependency on this exact version specify into there `package.json`. The only solution is to force to use the latest version of the package that we want, by replacing it into the `package-lock.json` file. But this is with the risk that by forcing another version they don't have breaking change and their internal API stays the same. At least from what I tested everything is working fine. All major npm tasks passed and no other changes were produced from running them.

The only issue that I see that this is depending on an external task `"preinstall": "npx npm-force-resolutions",` to keep the correct version into the `package-lock.json` file. 

For the idea of waiting - I'm ok with that but I'm not sure when this gonna happen? And will require us to update to the latest version of a few big packages when the time comes. And this could be a big one.

bdryanovski:
I'm closing this because right now we could not do much until some of the packages update their internal dependencies. No need for us to try to overwrite them without knowing what will be the impact on their code. Too much testing.

Also, the dependencies are development so there are not having any impact on the bundle packages

Issue #4274, opened 2020-02-14
Make date-picker more versatile by date filtering


spitfjre:
## Summary

* What is the change?
* Why should it go in Clarity?
In order to make the date picker more versatile it would be nice to have the ability to filter selectable dates beyond the scope of `min` and `max` option (as present in other datepickers like material and ng-pick-datetime).

* Does this change impact existing behaviors? If so how?
No

* If this change introduces a new behavior, is this behavior accessible?
Yes

## Use case
Having the possibility to only select "working days" (as in no Saturdays, no Sundays, no public holidays).

## Workarounds
Currently the you have to validate the selected date after selection, which is not nice.

gnomeontherun:
Thanks @spitfjre we actually designed the current `min/max` internals to support expanding this support. This would be something we can get help on to prioritize, but I'll put it into our backlog and list of anticipated form improvements.

gnomeontherun:
This is a duplicate of #3009 and closing with same resolution.

Issue #3009, opened 2019-01-09
Ability to disable particular dates in the date picker


mohammedzamakhan:
## Summary

Date picker should have the ability to disable particular dates. Probably adding a Input that takes a function that disables particular dates of the month, or even it can disable the months or years in the date picker selection.

* What is the change?
* Why should it go in Clarity?
* Does this change impact existing behaviors? If so how?
* If this change introduces a new behavior, is this behavior accessible?

## Use case

There can be so many use cases for this.

## Examples

_Are there any examples in other projects that demonstrate this feature or enhancement?_

## Workarounds

Don't know of any as of now

yogeshnagrath:
Hi all,

We are also facing one issue of similar type.

In our case, we want to block all the dates except the ones which are in particular range of dates.

For eg: Lets take january 2021, we want to enable only 5th-10th jan, 15th-20th jan and 25th-30th jan or 1st-1st jan(same start date and end date).

We need this for our project(Horizon Cloud) to schedule timely upgrades.

gnomeontherun:
The feature request here has been captured into our list and we’re going to take it into consideration as we develop Clarity Core capabilities. In an effort to clean up our backlog and focus our attention, I’m going to close this as captured in our feature requests. Please follow our development and releases to see when we release relevant components to make this possible. Future feature requests can be made in our [GitHub Discussions](https://github.com/vmware/clarity/discussions).

PR #4390, opened 2020-03-20
feat(date-picker): implement clrDateFilter to allow selection of specific dates

martinbrom:
## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #4274 

## What is the new behavior?

Datepicker now has a clrDateFilter input which allows the user to specify which dates should be selectable (using a function accepting a Date).

* [ ] Yes
* [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "clean": "del ./dist",
     "start": "npm run angular:dev:start",
     "build": "npm-run-all build:ci",
-    "build:ci": "npm-run-all clean format lint build:libs core:dev:build test",
+    "build:ci": "npm-run-all clean format lint build:libs core:dev:build",
     "build:libs": "npm-run-all -p ui:build icons:build core:build -s angular:build schematics:build",
     "build:apps": "npm-run-all -p angular:dev:build core:dev:build",
     "test": "npm-run-all -p angular:test core:test icons:test",
--- a/packages/angular/projects/clr-angular/src/forms/datepicker/calendar.ts
+++ b/packages/angular/projects/clr-angular/src/forms/datepicker/calendar.ts
@@ -96,7 +96,7 @@ export class ClrCalendar implements OnDestroy {
       this.focusedDay,
       this.today,
       this._localeHelperService.firstDayOfWeek,
-      this._dateIOService.disabledDates
+      this._dateIOService.excludedRanges
     );
   }
 
--- a/packages/angular/projects/clr-angular/src/forms/datepicker/date-input.spec.ts
+++ b/packages/angular/projects/clr-angular/src/forms/datepicker/date-input.spec.ts
@@ -161,20 +161,20 @@ export default function () {
           expect(context.clarityElement.className).not.toContain('clr-col-12');
         });
 
-        it('should handle big endien date strings for min inputs', () => {
+        it('should handle big endian date strings for min inputs', () => {
           // NOTE: big endian date format is yyyy-mm-dd
           const testComponent = context.fixture.componentInstance;
           const [minYear, minMonth, minDay] = testComponent.minDate.split('-').map(n => parseInt(n, 10));
           const testMinDateModel = new DayModel(minYear, minMonth - 1, minDay);
-          expect(testMinDateModel).toEqual(dateIOService.disabledDates.minDate);
+          expect(testMinDateModel.incrementBy(-1)).toEqual(dateIOService.minRange.maxDate);
         });
 
-        it('should handle big endien date strings for max inputs', () => {
+        it('should handle big endian date strings for max inputs', () => {
           // NOTE: big endian date format is yyyy-mm-dd
           const testComponent = context.fixture.componentInstance;
           const [maxYear, maxMonth, maxDay] = testComponent.maxDate.split('-').map(n => parseInt(n, 10));
           const testMaxDateModel = new DayModel(maxYear, maxMonth - 1, maxDay);
-          expect(testMaxDateModel).toEqual(dateIOService.disabledDates.maxDate);
+          expect(testMaxDateModel.incrementBy(1)).toEqual(dateIOService.maxRange.minDate);
         });
       });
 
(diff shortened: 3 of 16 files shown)


gnomeontherun:
Hi @martinbrom, thanks for taking a look at this feature. Before I review the code in depth, I wanted to talk about a different approach.

We had already an idea how we wanted to see this feature implemented when we added the `min` and `max` date support. Instead of passing a function (like you have done), you would be able to add new entries to the `excludedDates` array in the calendar model. Instead of running a function against every date to get a true/false reading, we're already checking if it is excluded in that range. So in theory, this could be as simple as adding an input binding that accepts an array of `DateRange` type and adding it to the `excludedDates` array. The challenge is mostly in ensuring that the `excludedDates` array is constructed properly with or without the `min` and `max` values set, or perhaps they can't be used at the same time. I'm not a fan of passing functions as input bindings as well, as it can lead towards side effects if done wrong, and is something we've avoided in Clarity. 

I can advise more specifics on this, but the result I would expect would be like the following (assuming some minor refactoring).

```html
<input type="data" clrDate [clrExcludeRanges]="myRanges" />
```

```typescript
  myRanges = [
    { min: '2020-01-01', max: '2020-01-31' }, // All of January
    { min: '2020-02-01, max: '2020-02-07' }, // First week of Feb
  ];
```

This gets us away from the filtering function and provides a clean input binding. What do you think?

martinbrom:
Seems like a good solution to me. I chose the function approach since a few other date-pickers also do it this way, but if an array of excluded dates is the way you want to go, I am of course fine with that.

gnomeontherun:
Thanks @martinbrom, it is helpful to have some external validation of my original plan. Do you want specific feedback at the moment or are you able to review the implementation and revise things first?

martinbrom:
I thought about this a bit more, and I arrived at the conclusion that the filter function has its advantages. Suppose that a developer wants to only allow selection of weekends, using the `clrExcludedRanges` he would have to create over 50 ranges just to cover a single year, while the filter function would just be a few-characters-long one-liner.
The advantage of `clrExcludedRanges` becomes apparent when the 'invalid' dates are non-regular, such as fully booked days for a business reservation calendar.

Would it be a bad design to allow `min`, `max`, `clrDateFilter` and also `clrExcludedRanges` to give the devs a choice?

PS: Please look over the edit on my previous comment for additional thoughts on the implementation I had.

michaelstone:
Hey guys, we've got a similar requirement for a date picker and would love to stick with Clarity as our component library.

I see that the excluded date range array has some advantages, but I think I agree with @martinbrom that a filter function might be a better option. The excluded weekends example is a common use case where allowing a filter function to be provided makes that a simple operation, but if you have to provide an array of all weekends then a lot more complexity is pushed to the component consumer.

@gnomeontherun What sort of side effects are you concerned about if an input function is provided incorrectly?

martinbrom:
Also found this issue (#3009) which mentions the function approach.

gnomeontherun:
@martinbrom I don't like having too many different ways to do the same thing, so I don't want to support more options than necessary. I'm also trying to talk with others about how we need to handle this behavior so it aligns to a future version of this component, please bear with me a little longer.

When you have a function in a binding, the one-way data flow becomes messed up, and depending on how things are implemented that function can be _evaluated_ with each change detection cycle. When you pass a reference to a function, you should be able to avoid this unless your function gets called during change detection processing. Its a pattern I worry about using in one place, because once you use it once it becomes harder to control it in other places.

gnomeontherun:
Hi @martinbrom I've talked with folks and have an updated guidance here. I'm going to ask @coryrylan to review my comments for accuracy and clarity.

When we build a web component implantation of a date picker, there is no way we can accept a function. The solution we've discussed is more flexible than just providing a default list of dates to disable.

### API

```
<clr-date-container (clrMonthChange)="currentMonth($event)" [clrDisabledDates]="disabledDates">
   <input type="date" clrDate name="date" [(ngModel)]="startDate" />
</clr-date-container>
```

The ClrDateContainer will have a new input `clrDisabledDates` which is an array of dates in `YYYY-mm-dd` format to be disabled. It could also include date ranges in the following formats.

```
['2020-01-01'] // Single days
[['2020-02-01', '2020-02-15']] // Date ranges
['2020-01-01', ['2020-02-01', '2020-02-15']] // Mixed
```

Then there will be a new output `clrMonthChange` which will pass the current month back to the application. Instead of a function being passed into the control like your initial proposal, the function stays on the outside and listens for this month change event. When it fires, the expectation is that the app can update the `clrDisabledDates` binding with a new set of dates for the current month. So we've kept the function itself from being pushed into Clarity, but its still easy to make a function on the application.

I haven't done an audit of all of the implementation requirements but was interested in your feedback on this approach. I think it gives us the best long term compatibility as well as allows you to leverage a function on the app side if needed.

martinbrom:
I didn't really like the non-function approach since it pushed a lot of extra complexity to each application in some cases, but now with the addition of the `clrMonthChange` output, creating the disabled dates array will be a bit easier. I am fine with your suggestion and will implement it this way.

gnomeontherun:
Thanks @martinbrom! This is important to make sure the API doesn't change dramatically in a future date, so I'm looking forward to your updated implementation!

gnomeontherun:
Hey @martinbrom just curious how things are going?

martinbrom:
@gnomeontherun I got a little sidetracked with the current situation and also had problems with building/running/etc. while only having access to my Windows machine, sorry about that. I've seen there have been changes done to the project structure, I will rebase my changes and push them ASAP so I can get some feedback. I have a few questions but I will ask them after I push the code so it's easier to explain my thoughts.

gnomeontherun:
If you don't want to rebase, I can help do that if we get the rest of the code in the right place.

martinbrom:
I've tried for a few evenings now to test my changes on my machine but have been unsuccessful so far. There are still a few tests missing and documentation has also not been adapted yet.

To continue the discussion about the implementation:
with the "requirement" to use only a single array for the minRange, maxRange and the excludedRanges comes the problem I've described few comments prior. 

your idea:
> The challenge is mostly in ensuring that the excludedDates array is constructed properly with or without the min and max values set, or perhaps they can't be used at the same time.

my comment:
> I just realized that the `min` and `max` inputs actually create an 'enabled' interval, not a 'disabled' one (it should not have been called `excludedDates` in the code). What do you think is the best approach?
> 
> * merging the intervals
> * not allowing min/max+this new directive at the same time
> * changing the new directive to `clrIncludedRanges` ?

Do you still think it is beneficial to have only a single array for all of the excluded ranges? Right now I invert the min and max interval so they can be merged with the intervals we get in the `clrDisabledDates`. This leads to not very "nice" code down the road.

PS: The local error I get is "This class is visible to consumers via ClarityModule -> ... but is not exported from the top-level library entrypoint".

tanmayray:
Hi @martinbrom . We are using clarity date-picker in our project and have a use case where we want to disable any weekend selection. This can't be achieved using min max. 
Please let me know if this feature can be used to achieve the same.

gnomeontherun:
Hi @martinbrom, would you mind sharing your updated branch with us to review? Its probably not exporting something correctly for that specific error.

I think it makes sense to either use `min`/`max` or the `clrDisabledDates` option at once, so if `clrDisabledDates` is present it would take precendence over `min/max`.

martinbrom:
I force pushed the new impl (or a part of it) over the old function approach implementation. I tried building a freshly cloned repo and I get the same issue.

@tanmayray It should be possible when this is resolved.
@gnomeontherun I just realized, to correctly disable dates, we would need a `clrYearChange` too, knowing the current month is not enough to calculate disabled dates.

tanmayray:
@martinbrom Thank you for the update. I would like to highlight one concern here. Looking at the comments it seems like the implementation is going to be a new directive (clrDisabledDates) which accepts an array of dates to be disabled.
But for a use case like disabling all weekends calculating that array might not be possible as the result size would be infinite. We would probably need an event to detect the current month being displayed and calculate the array for that particular month (or something else on similar lines, if that makes sense).

martinbrom:
Yes, I know what you mean and if you look deeper into the comment thread we briefly discussed this a while ago. We decided on a `clrMonthChange` which will emit a value whenever the "visible" month changes and I realized that is not enough to determine the disabled dates so we might also add a `clrYearChange` or some combination of those two observables.

mathisscott:
@gnomeontherun : what is our status here?

gnomeontherun:
We haven't seen an updated PR yet with the latest requests, so this is still waiting.

martinbrom:
Unfortunately, I haven't been able to build/run the Clarity repository locally since March so unless I figure out what is the problem on either of my workstations I will not be able to submit an updated PR, sorry.

mathisscott:
@gnomeontherun 
Do we want to close this PR?

gnomeontherun:
@martinbrom Do you anticipate continuing to work on this? If so what is the status, and if not we can close the PR.

martinbrom:
No, I don't see that happening anytime soon, sorry again. Please close the PR.

PR #4404, opened 2020-03-25
fix(i18n): restore angular common strings service

mathisscott:
• crossing the streams with a service shared b/w core and angular was causing nx and jest to lose their minds
• much of the problem was due to when typescript or jest would decide to compile an import down to ES5 or not
• recent discussions have uncovered the potential for more significant divergence in the scope of what we previously felt would be shared services
• in the future, a shared resource may need to be pulled into angular proper – as opposed to being brought in as a dependency
• i verified this approach solves breakages in previously breaking apps
• users still need to use `transformIgnorePatterns` in their jest config to allow babel to compile core, lit, and ramada
• users will NOT need to override ClrCommonStrings in their spec configs
• that was a band-aid that was going to break again soon in any case

Tested in:
✔︎ Chrome

Fixes: #4094

## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [ ] No

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -283,7 +283,12 @@ export declare class ClrCheckboxWrapper implements DynamicWrapper, OnInit, OnDes
 export declare class ClrCommonFormsModule {
 }
 
-export declare class ClrCommonStringsService extends CommonStringsServiceInternal {
+export declare class ClrCommonStringsService {
+    get keys(): Readonly<ClrCommonStrings>;
+    localize(overrides: Partial<ClrCommonStrings>): void;
+    parse(source: string, tokens?: {
+        [key: string]: string;
+    }): string;
 }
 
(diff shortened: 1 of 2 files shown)


aravindh-nagarajan:
@mathisscott Hi, Is there any minor release in pipeline with this fix ? We have been facing issues like https://github.com/vmware/clarity/issues/4094 since clarityv3 upgrade and looks like this one will resolve it?

mathisscott:
@aravindh-nagarajan 
it's merged into our master branch. it will be released in our next release which is likely going out next week. given that Core is introducing new layouts and design tokens, it will probably be the 3.1 release.

aravindh-nagarajan:
@mathisscott Thank you.

Issue #4410, opened 2020-03-26
[@clr/core]: Firefox is picking up webkit-appearance settings


mathisscott:
## Describe the bug

[On our docs](https://clarity.design/storybook/core/?path=/story/components-button--form), Firefox is picking up a `-webkit-appearance` styling that is being set by storybook or a CSS reset.

## How to reproduce

View the link above in firefox.

Image of reproduction...

![Screen Shot 2020-03-26 at 12 43 08 PM](https://user-images.githubusercontent.com/2728359/77679748-fb301f80-6f60-11ea-8ca0-1f5fe86e6e12.png)

## Expected behavior

We have styles in place to prevent this.

![Screen Shot 2020-03-26 at 12 43 51 PM](https://user-images.githubusercontent.com/2728359/77679798-0edb8600-6f61-11ea-8eda-cac80ae18f23.png)

## Versions

**App**

* Clarity: 3.0.1

**Device:**

* Type: MacBook
* OS: OSX
* Browser: Firefox
* Version: 74.0

## Additional notes

1. The "fixed" image above overrides the storybook/reset styles by applying styles to `cds-button`. We should apply them to `:host` inside the component. But I could not get that working in the code inspector.
2. We should also account for the other `type` values listed in the reset so that this doesn't come back in a different form.

PR #4425, opened 2020-03-27
fix: migrate buttons to use design tokens

mathisscott:
• per discussion, not using layouts here but tokens
• this is to preserve ease of atomically re-styling the buttons with custom props
• fixed inset box shadows on active states – was not working
• applied new layouts to all buttons demos
• fixed a very precise sizing issue with line height erasers
• had to use precise sizings and paddings for buttons though; it's voodoo math at the moment but the existing equations are missing a variable somehow
• i believe the combination of a LHE inside of a "button" is throwing off the LHEs
• f-fox was applying `-webkit-appearance: button` to the `cds-button` b/c it has an attribute of `type="button"` or `type="submit"`
• we should be careful in the future not to use common HTML attribute names
• we should consider changing this attr/prop to not be `type`

## PR Checklist

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

See notes above...

Issue Number: #4410 

## What is the new behavior?

See notes above...

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-core/button/button.element.scss
+++ b/src/clr-core/button/button.element.scss
@@ -1,163 +1,192 @@
-@import './../styles/mixins/utils';
+// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
 
-// TODO: progress spinner CSS deprecated, will be replaced by cds-progress
+@import './../styles/tokens/generated/index';
+@import './../styles/mixins/utils';
+@import './../styles/mixins/mixins';
+
+// TODO: progress spinner CSS deprecated, will be replaced by cds-circular-progress
 @import './../../clr-angular/progress/spinner/variables.spinner';
 @import './../../clr-angular/image/icons.clarity';
 @import './../../clr-angular/progress/spinner/spinner.clarity';
 
 :host {
-  --box-shadow-color: var(--clr-color-action-500, #{$clr-color-action-500});
-  --border-radius: #{$clr_baselineRem_3px};
-  --border-color: var(--clr-color-action-600, #{$clr-color-action-600});
-  --color: var(--clr-color-neutral-0, #{$clr-color-neutral-0});
-  --background: var(--clr-color-action-600, #{$clr-color-action-600});
+  --box-shadow-color: #{$cds-token-color-action-900};
+
+  --border-radius: #{$cds-token-global-border-radius};
+  --border-width: #{$cds-token-global-border-width};
+  --border-color: #{$cds-token-color-action-600};
+  --background: #{$cds-token-color-action-600};
+
+  --color: #{$cds-token-color-neutral-0};
+  --font-size: #{$cds-token-space-size-6};
+  --font-weight: #{$cds-token-typography-font-weight-semibold};
+  --font-family: #{$cds-token-typography-font-family};
+  --text-transform: uppercase;
+  --letter-spacing: 0.12em;
+
+  --padding: #{$cds-token-layout-space-sm};
+  --height: calc(#{$cds-token-space-size-11} + #{$cds-token-space-size-1});
+
   display: inline-block;
 }
 
 .private-host {
-  --font-size: #{$clr_baselineRem_0_5};
-  --line-height: 1.7rem;
-  --letter-spacing: 0.12em;
-  --padding: 0 #{$clr_baselineRem_0_5};
-
+  -webkit-appearance: none !important;
+  align-items: center;
   background: var(--background);
-  color: var(--color);
   border-color: var(--border-color);
   border-radius: var(--border-radius);
-  border-width: $clr-global-borderwidth;
-  font-size: var(--font-size);
-  font-weight: 500;
-  line-height: var(--line-height);
-  letter-spacing: var(--letter-spacing);
-  padding: var(--padding);
-  user-select: none;
-  display: block;
-  -webkit-appearance: none !important;
+  border-style: solid;
+  border-width: var(--border-width);
+  color: var(--color);
   cursor: pointer;
-  min-width: $clr_baselineRem_3;
+  display: inline-flex;
+  font-family: var(--font-family);
+  font-size: var(--font-size);
+  font-weight: var(--font-weight);
+  height: var(--height);
+  justify-content: center;
+  letter-spacing: var(--letter-spacing);
+  line-height: 1em;
+  // TODO: following is a fix for Firefox cutting off right edge of buttons. remove when no longer needed.
+  margin-right: #{$cds-token-space-size-1};
+  min-width: #{$cds-token-space-size-13};
+  overflow: visible;
+  padding: var(--padding);
+  position: relative;
   text-align: center;
   text-decoration: none;
-  text-transform: uppercase;
+  text-overflow: ellipsis;
+  text-transform: var(--text-transform);
+  transform: translateZ(0px); // for chrome rendering bug
+  user-select: none;
   vertical-align: middle;
   white-space: nowrap;
-  text-overflow: ellipsis;
-  overflow: visible;
-  border-style: solid;
-  position: relative;
-  transform: translateZ(0px); // for chrome rendering bug
-}
 
-:host(:hover) {
-  --background: var(--clr-color-action-800, #{$clr-color-action-800});
-}
+  &:active {
+    box-shadow: 0 #{$cds-token-space-size-1} 0 0 var(--box-shadow-color) inset;
+  }
 
-:host(:active) .private-host {
-  box-shadow: 0 $clr_baselineRem_1px 0 0 var(--box-shadow-color) inset;
-}
+  & > span {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    height: var(--font-size);
 
-:host([is-anchor]) {
-  .private-host {
-    padding: 0;
+    & > span {
+      @include line-height-eraser(0em);
+    }
   }
 }
 
+:host(:active) .private-host {
+  box-shadow: 0 #{$cds-token-space-size-1} 0 0 var(--box-shadow-color) inset;
+}
+
+:host(:hover) {
+  --background: #{$cds-token-color-action-800};
+}
+
 ::slotted {
-  line-height: var(--line-height);
+  line-height: 1em;
   color: inherit;
 }
 
 ::slotted(a) {
-  padding: 0 #{$clr_baselineRem_0_5};
   text-decoration: none !important;
   display: block;
   height: 100%;
   color: inherit;
-  line-height: var(--line-height);
 }
 
 ::slotted(cds-icon) {
+  $icon-overage-nudge: calc(#{$cds-token-space-size-2} * -1);
+
   --color: inherit;
+
+  @include equilateral(calc(#{$cds-token-space-size-6} + #{$cds-token-space-size-4}));
+  margin-top: $icon-overage-nudge;
+  margin-left: $icon-overage-nudge;
 }
 
 :host([action='outline']) {
   --background: transparent;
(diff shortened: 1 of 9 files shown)


src/clr-core/styles/mixins/_mixins.scss
```diff
@@ -187,47 +187,58 @@ $opp-directions: (
...
+  $push-nudge: 0.044em
 ) {
-  @return calc(((1em - #{$font-top-gap} - #{$font-ascender-height} - #{$font-x-height}) + #{$line-height-gap}) * -1);
+  @return calc(
+    (((1em - #{$font-top-gap} - #{$font-ascender-height} - #{$font-x-height}) + #{$line-height-gap}) * -1) - #{$push-nudge}
+  );
 }
 
 // default are for metropolis/clarity-city
 @mixin line-height-eraser(
```
coryrylan: So were we not able to use the line height eraser for the button or was it not needed? Do we need to have an inner span for projected content to use the line height eraser?

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
+// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@import './../styles/tokens/generated/index';
+
 @import './../styles/mixins/utils';
+@import './../styles/mixins/mixins';
```
coryrylan: The one mixin file and tokens file are already in the base component styles so they do not need to be imported. We should add the other mixin file to the base style or combine the mixin files. 

https://github.com/vmware/clarity/blob/master/src/clr-core/common/base/base.element.scss#L5

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
...
   border-color: var(--border-color);
   border-radius: var(--border-radius);
-  border-width: $clr-global-borderwidth;
+  border-width: var(--border-width);
   font-size: var(--font-size);
-  font-weight: 500;
-  line-height: var(--line-height);
+  font-weight: var(--font-weight);
+  height: var(--height);
+  line-height: 1em;
```
coryrylan: We have line height here but also it is set on the slotted elements on lines 93 and 104 do we need all three or just one?

src/clr-core/button/button.stories.ts
```diff
@@ -84,120 +84,143 @@ export const API = () => {
...
-    <cds-button>solid</cds-button>
-    <cds-button action="outline">outline</cds-button>
-    <cds-button action="flat">link</cds-button>
+    <div cds-layout="horizontal gap:xs">
+      <cds-button>solid</cds-button>
+      <cds-button action="outline">outline</cds-button>
+      <cds-button action="flat">link</cds-button>
+    </div>
   `;
 };
```
coryrylan: This is more of a fleeting thought, but having layout code in the component demos is there any concern of confusion for the devs copy/pasting it since this is the same code that renders in the docs? I was kind of on the fence on it so I wasnt sure if anyone had an opinion.

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
...
+  --height: #{$clr_baselineRem_1_5};
   display: inline-block;
+  -webkit-appearance: none !important;
+}
+
+@mixin button-padding() {
+  padding-top: calc(var(--padding-vertical) - var(--border-width) + 0.025rem);
+  padding-left: calc(var(--padding-horizontal) - var(--border-width));
+  padding-right: calc(var(--padding-horizontal) - var(--border-width));
+  padding-bottom: calc(var(--padding-vertical) - var(--border-width) - 0.025rem);
```
coryrylan: Why are the calcs necessary as well as the 0.025rem?

src/clr-core/button/button.element.scss
```diff
@@ -45,119 +65,129 @@
...
+    box-shadow: 0 $clr_baselineRem_1px 0 0 var(--box-shadow-color) inset;
+  }
+
+  .spinner {
+    margin-top: -1 * $clr_baselineRem_4px;
+  }
 }
 
 :host(:active) .private-host {
   box-shadow: 0 $clr_baselineRem_1px 0 0 var(--box-shadow-color) inset;
```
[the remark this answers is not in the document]
jeeyun: I like the suggestion to include other common sizes.. but what do we do if new sizes come up as we work on core versions of more components? Do we want to stick with something like the updated above, and change the sizing of some of the components if that happens?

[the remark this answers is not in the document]
coryrylan: Yeah that is a discussion I would like to have. The goal of the token system is to try to narrow down the number of spacing values used which helps consistency with spacing. But we also want to make sure we find that right balance of to few or too many and what is acceptable if we have slight adjustments to use what already exists as a token.

src/clr-core/button/button.stories.ts
jeeyun: I think, if the component that is being demoed is commonly expected to be in a layout container (e.g. form with a bunch form fields?), it's ok to leave it there because it's beneficial for the devs to copy/paste it including the layout.

If the button demos could render reasonably without it maybe it's better to not include the layout code? What do y'all think?

jeeyun: Looking at the button demo though, I can see it's pretty useful here to space out the buttons that are grouped together. So I'm really ok with leaving it as is too..

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
...
+  padding-top: calc(var(--padding-vertical) - var(--border-width) + 0.025rem);
+  padding-left: calc(var(--padding-horizontal) - var(--border-width));
+  padding-right: calc(var(--padding-horizontal) - var(--border-width));
+  padding-bottom: calc(var(--padding-vertical) - var(--border-width) - 0.025rem);
 }
 
 .private-host {
   --font-size: #{$clr_baselineRem_0_5};
-  --line-height: 1.7rem;
+  --font-weight: 500;
```
jeeyun: Should this be one of the font-weight tokens? Maybe `$cds-token-typography-font-weight-semibold`?

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
...
-  --background: var(--clr-color-action-600, #{$clr-color-action-600});
+  --box-shadow-color: #{$cds-token-color-action-900};
+  --border-radius: #{$cds-token-global-border-radius};
+  --border-width: #{$cds-token-global-border-width};
+  --border-color: #{$cds-token-color-action-600};
+  --color: #{$cds-token-color-neutral-0};
+  --background: #{$cds-token-color-action-600};
+  --height: #{$clr_baselineRem_1_5};
   display: inline-block;
+  -webkit-appearance: none !important;
```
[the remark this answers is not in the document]
mathisscott: Sounds good. Thanks!

src/clr-core/button/button.element.scss
```diff
@@ -1,35 +1,55 @@
...
-  --border-color: var(--clr-color-action-600, #{$clr-color-action-600});
-  --color: var(--clr-color-neutral-0, #{$clr-color-neutral-0});
-  --background: var(--clr-color-action-600, #{$clr-color-action-600});
+  --box-shadow-color: #{$cds-token-color-action-900};
+  --border-radius: #{$cds-token-global-border-radius};
+  --border-width: #{$cds-token-global-border-width};
+  --border-color: #{$cds-token-color-action-600};
+  --color: #{$cds-token-color-neutral-0};
+  --background: #{$cds-token-color-action-600};
+  --height: #{$clr_baselineRem_1_5};
```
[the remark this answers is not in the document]
mathisscott: When i was using LHEs, it was shrinking the height of the button. That got left in there. It may no longer be needed but I'm also going to circle back to the LHEs to see if I can get them going.

src/clr-core/button/button.element.scss
```diff
@@ -45,119 +65,129 @@
...
-}
 
-:host(:hover) {
-  --background: var(--clr-color-action-800, #{$clr-color-action-800});
+  &:active {
+    box-shadow: 0 $clr_baselineRem_1px 0 0 var(--box-shadow-color) inset;
+  }
+
+  .spinner {
+    margin-top: -1 * $clr_baselineRem_4px;
```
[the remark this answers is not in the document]
mathisscott: Yup.

src/clr-core/button/button.element.scss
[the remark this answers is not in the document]
mathisscott: Yes we can. As discussed, I'll try to use calc() to put together some of the odd values we're seeing...

src/clr-core/styles/mixins/_mixins.scss
mathisscott: LHEs were behaving erratically. I can try to find a workaround for that but badges weren't behaving much better. The difference appears to be precise vertical sizing versus flow. LHEs do flow but disrupt precise vertical sizing.

src/clr-core/button/button.element.scss
mathisscott: Result of playing battleship with the LHEs. May or may not be needed anymore.

src/clr-core/button/button.stories.ts
mathisscott: It was an easy way to dogfood layouts and solved the disappearing margins due to how browsers were handling "whitespace" between the buttons. Per our meeting today, we'll keep it in.

src/clr-core/button/button.element.scss
mathisscott: For specific sizing, we have to remove the border-width from the padding values. Otherwise, we get a 37.5px high button instead of the expected 36px.

The `0.025rem` is a magic number that nudges the button text just enough to make it perfectly centered along the vertical axis.

mathisscott: I'm revisiting the LHEs, tho. So if I can get that working for buttons, we may be able to get rid of this magic number.

src/clr-core/button/button.element.scss
```diff
@@ -45,119 +65,129 @@
...
 :host([size='sm']) .private-host {
-  --line-height: #{$clr_baselineRem_1 - $clr_baselineRem_1px};
   --letter-spacing: 0.073em;
-  --font-size: #{$clr_baselineRem_0_458};
-  --padding: #{0 $clr_baselineRem_0_5};
+  --padding-vertical: #{$cds-token-layout-space-xs};
+  --height: #{$clr_baselineRem_1};
 
   .spinner {
     @include min-equilateral($clr_baselineRem_0_5416);
```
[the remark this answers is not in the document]
mathisscott: Nope. Removing it per our discussion today.

src/clr-core/button/button.element.scss
mathisscott: Yep. I missed that. Good catch!

src/clr-core/button/button.element.scss
mathisscott: Apparently we need two of them but not all three...

src/clr-core/button/button.element.scss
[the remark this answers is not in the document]
mathisscott: I've verified that `--height` is not needed.

src/clr-core/styles/tokens/tokens.yml
```diff
@@ -44,17 +52,29 @@ props:
     type: 'px'
     category: 'space'
   - name: 'size-8'
-    value: '24px'
+    value: '20px'
```
coryrylan, line 55: As per slack do you want to bump this down to 18px then that will meet the quarter values of 24px? I think thats all we would need left for the values

PR #4431, opened 2020-03-31
feat(dnd): add clrDragStartDelay input property to clrDraggable

Shijir:

## PR Checklist

## PR Type

This PR introduces `clrDragStartDelay` input to `clrDraggable` directive. The input defines the hold duration for initiating the drag start event.

Example: 
```html
<button class="btn btn-primary" clrDraggable [clrDragStartDelay]="500">draggable</button>
```

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/utils/drag-and-drop/draggable-ghost.spec.ts
+++ b/src/clr-angular/utils/drag-and-drop/draggable-ghost.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -7,11 +7,9 @@ import { Component } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-
 import { DomAdapter } from '../dom-adapter/dom-adapter';
-
-import { ClrDraggableGhost } from './index';
 import { ClrDragAndDropModule } from './drag-and-drop.module';
+import { ClrDraggableGhost } from './draggable-ghost';
 import { DragEventListenerService } from './providers/drag-event-listener.service';
 import { MOCK_DRAG_EVENT_LISTENER_PROVIDER } from './providers/drag-event-listener.service.mock';
 import { DraggableSnapshotService } from './providers/draggable-snapshot.service';
@@ -57,7 +55,9 @@ export default function(): void {
       });
 
       it('should appear on the first drag move event', function() {
+        this.dragEventListener.dragStartPosition = { pageX: 0, pageY: 0 };
         const mockDragMoveEvent = { dragPosition: { pageX: 33, pageY: 44 } };
+
         this.dragEventListener.dragMoved.next(mockDragMoveEvent);
 
         expect(this.ghostElement.style.left).toBe(`${mockDragMoveEvent.dragPosition.pageX}px`);
@@ -66,6 +66,7 @@ export default function(): void {
       });
 
       it('should be dragged from mouse position on page', function() {
+        this.dragEventListener.dragStartPosition = { pageX: 0, pageY: 0 };
         const mockDragMoveEvent1 = { dragPosition: { pageX: 120, pageY: 60 } };
         const mockDragMoveEvent2 = { dragPosition: { pageX: 180, pageY: 120 } };
 
--- a/src/clr-angular/utils/drag-and-drop/draggable/draggable.ts
+++ b/src/clr-angular/utils/drag-and-drop/draggable/draggable.ts
@@ -72,6 +72,15 @@ export class ClrDraggable<T> implements AfterContentInit, OnDestroy {
     this.dragEventListener.group = value;
   }
 
+  @Input('clrDragStartDelay')
+  set dragStartDelay(value: number) {
+    if (typeof value === 'number') {
+      this.dragEventListener.dragStartDelay = value;
+    } else if (typeof value === 'string') {
+      this.dragEventListener.dragStartDelay = parseInt(value, 10) || 0;
+    }
+  }
+
   private createDefaultGhost(event: DragEventInterface<T>) {
     this.draggableSnapshot.capture(this.draggableEl, event);
     // NOTE: The default ghost element will appear
--- a/src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
+++ b/src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
@@ -6,24 +6,29 @@
 import { Injectable, NgZone, Renderer2 } from '@angular/core';
 import { Observable, Subject } from 'rxjs';
 
-import { DragEventInterface, DragEventType } from '../interfaces/drag-event.interface';
+import { DragEventInterface, DragEventType, DragPointPosition } from '../interfaces/drag-event.interface';
 import { DragAndDropEventBusService } from './drag-and-drop-event-bus.service';
 
 @Injectable()
 export class DragEventListenerService<T> {
   private draggableEl: any;
 
-  // contains the starting events such as mousedown and touchstart
-  private listeners: (() => void)[];
-  // contains the nested events that happens after/inside the starting events
+  // contains listeners for the starting events such as mousedown and touchstart
+  private listeners: (() => void)[] = [];
+  // contains listeners for nested events that happens after/inside the starting events
   // such as selectstart, mousemove/touchmove, mouseup/touchend
   private nestedListeners: (() => void)[];
+
+  // contains listener for mousemove/touchmove before delay
+  private checkDragStartBoundaryListener: () => void;
 
   private dragStart: Subject<DragEventInterface<T>> = new Subject<DragEventInterface<T>>();
   private dragMove: Subject<DragEventInterface<T>> = new Subject<DragEventInterface<T>>();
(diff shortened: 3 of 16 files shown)


src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
```diff
@@ -13,18 +13,23 @@ import { DragAndDropEventBusService } from './drag-and-drop-event-bus.service';
...
+  // contains listener for mousemove/touchmove before delay
+  private checkDragStartBoundaryListener: () => void;
+
   private dragStart: Subject<DragEventInterface<T>> = new Subject<DragEventInterface<T>>();
   private dragMove: Subject<DragEventInterface<T>> = new Subject<DragEventInterface<T>>();
   private dragEnd: Subject<DragEventInterface<T>> = new Subject<DragEventInterface<T>>();
 
   private hasDragStarted: boolean = false;
 
+  private timeout: ReturnType<typeof setTimeout>;
```
gnomeontherun, line 31: 👏 Nice typing.

src/clr-angular/utils/drag-and-drop/draggable-ghost.spec.ts
```diff
@@ -1,17 +1,15 @@
...
 import { Component } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-
 import { DomAdapter } from '../dom-adapter/dom-adapter';
-
-import { ClrDraggableGhost } from './index';
 import { ClrDragAndDropModule } from './drag-and-drop.module';
+import { ClrDraggableGhost } from './index';
```
gnomeontherun, line 12: Shouldn't import from `index` even in test.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
```diff
@@ -136,15 +151,36 @@ export class DragEventListenerService<T> {
...
+          clearTimeout(this.timeout);
+
           // We must remove the the nested listeners every time drag completes.
-          if (this.nestedListeners) {
-            this.nestedListeners.map(event => event());
-          }
+          this.nestedListeners.map(event => event());
+
+          // We must remove the event listener from checkDragStartBoundary
+          this.checkDragStartBoundaryListener();
```
gnomeontherun, line 160: It seems unlikely, but it would be good to also check if this is defined and call it OnDestroy just to be safe.

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
```diff
@@ -37,24 +42,27 @@ export class DragEventListenerService<T> {
...
+  }
+
   constructor(private ngZone: NgZone, private renderer: Renderer2, private eventBus: DragAndDropEventBusService<T>) {}
 
   private initialPosition: { pageX: number; pageY: number };
 
   // Draggable component sets these properties:
   public dragDataTransfer?: T;
   public group?: string | string[];
+  public dragStartDelay: number = 0;
```
gnomeontherun, line 56: Just do `dragStartDelay = 0` since type is inferred. Do we want to have a default delay, say 100?

src/clr-angular/utils/drag-and-drop/providers/drag-event-listener.service.ts
```diff
@@ -37,24 +42,27 @@ export class DragEventListenerService<T> {
     return this.dragEnd.asObservable();
   }
 
+  get dragStartPosition(): { pageX: number; pageY: number } {
```
gnomeontherun, line 45: See several uses of this, maybe move it to an interface?

src/clr-angular/utils/drag-and-drop/draggable/draggable.ts
```diff
@@ -72,6 +72,15 @@ export class ClrDraggable<T> implements AfterContentInit, OnDestroy {
     this.dragEventListener.group = value;
   }
 
+  @Input('clrDragStartDelay')
+  set dragStartDelay(value: number) {
+    if (typeof value === 'number') {
```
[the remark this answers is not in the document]
Shijir: In that case, TS will complain that `value` is not a string. So you have to cast its type to string, which I am trying to avoid.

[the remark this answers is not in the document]
bdryanovski: Then a better solution it will add `string` as part of the possible types 

```ts
function dragStartDelay(value: number | string) {
  return parseInt(value as string, 10) || 0
}
dragStartDelay("100");
dragStartDelay(100);
```

When TypeScript compiles the code it will remove the type annotation and leave the code as simple as it could be.

```js
"use strict";
function dragStartDelay(value) {
    return parseInt(value, 10) || 0;
}
dragStartDelay("100");
dragStartDelay(100);
```

For example here is the instruction when we branch it. 
```bash
   13 E> 0x3e49bd357fca @    0 : a5                StackCheck
   25 S> 0x3e49bd357fcb @    1 : 25 02             Ldar a0
         0x3e49bd357fcd @    3 : 73 00             TestTypeOf #0
         0x3e49bd357fcf @    5 : 99 05             JumpIfFalse [5] (0x3e49bd357fd4 @ 10)
   62 S> 0x3e49bd357fd1 @    7 : 25 02             Ldar a0
   75 S> 0x3e49bd357fd3 @    9 : a9                Return
   85 S> 0x3e49bd357fd4 @   10 : 25 02             Ldar a0
         0x3e49bd357fd6 @   12 : 73 01             TestTypeOf #1
         0x3e49bd357fd8 @   14 : 99 14             JumpIfFalse [20] (0x3e49bd357fec @ 34)
  122 S> 0x3e49bd357fda @   16 : 13 00 00          LdaGlobal [0], [0]
         0x3e49bd357fdd @   19 : 26 fb             Star r0
         0x3e49bd357fdf @   21 : 0c 0a             LdaSmi [10]
         0x3e49bd357fe1 @   23 : 26 f9             Star r2
  129 E> 0x3e49bd357fe3 @   25 : 5e fb 02 f9 02    CallUndefinedReceiver2 r0, a0, r2, [2]
         0x3e49bd357fe8 @   30 : 96 03             JumpIfToBooleanTrue [3] (0x3e49bd357feb @ 33)
         0x3e49bd357fea @   32 : 0b                LdaZero
  153 S> 0x3e49bd357feb @   33 : a9                Return
         0x3e49bd357fec @   34 : 0d                LdaUndefined
  158 S> 0x3e49bd357fed @   35 : a9                Return
```

And here is the simple just `return parseInt(value, 10) || 0`

```bash
   13 E> 0x3e60f82d7fca @    0 : a5                StackCheck
   25 S> 0x3e60f82d7fcb @    1 : 13 00 00          LdaGlobal [0], [0]
         0x3e60f82d7fce @    4 : 26 fb             Star r0
         0x3e60f82d7fd0 @    6 : 0c 0a             LdaSmi [10]
         0x3e60f82d7fd2 @    8 : 26 f9             Star r2
   32 E> 0x3e60f82d7fd4 @   10 : 5e fb 02 f9 02    CallUndefinedReceiver2 r0, a0, r2, [2]
         0x3e60f82d7fd9 @   15 : 96 03             JumpIfToBooleanTrue [3] (0x3e60f82d7fdc @ 18)
         0x3e60f82d7fdb @   17 : 0b                LdaZero
   56 S> 0x3e60f82d7fdc @   18 : a9                Return
```

I think in the case above, the annotation is only helping you when you develop but won't prevent you from doing anything stupid.  But branching the code just to force the `tsc` to accept it is a bit overkill for this super small getter.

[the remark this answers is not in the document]
Shijir: That's what I am avoiding: casting string type to the value just to use it in `parseInt`. And to do that, you have to change the accepted types to both `string` and `number`. I am not supporting string value intentionally there to prevent confusion. It should be clear that It only accepts a numeric value. But it's reasonable for users to expect the following to work: `clrDragStartDelay="500"` This is handled properly in its current implementation without suggesting we also accept strings. That's more important than having one-liner there. And I am not worried about how TS compiles and does its job in this case.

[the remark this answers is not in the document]
bdryanovski: It's not one-liner the point here. If you require `number` - require `number`, adding annotation for number, but handle all other types inside the method is a bit of a lie and defeats the purpose of types. 

The idea of types is to give you information about what a function will accept/return and work with. I see and I get why `number` as argument of component is done this way. There are a lot of cases like that when TypeScript creates a fake feeling of safety. And keep in mind that TS is working only until we run `tsc` everything after that is just vanilla JS. So this method will and may get everything from `string`, `number`, `object`, `array` whatever and nothing could stop it ;)

But we don't need to handle all of them, we expect `number` and will parse `strings` to number just to not force people to wrap everything into `[input]`. So why we don't annotate it - what it is? Is more or less like an easter egg, "Oy, it could work with `string` as well, d*mn I did not know that"

```ts
parseInt(value as string, 10)
``` 

This type of casting is not really adding anything to the code after all is used only to tell `tsc` ok boy, don't throw an error. It's excluded from the compiled code.

```js
parseInt(value, 10)
```

But branching with `if/else` and adding `typeof` as operations are having an effect on the code. 
I'm ok with a lot of things but don't see why we need to write JavaScript code to make some TypeScripts rules pass that are adding overhead but don't give anything in return. (The reason for the if is that `parseInt` is annotated that the first argument must be `string` but .. it will work with `number` and just will return the number - again misleading annotation )

The speed improvement is not that big https://jsperf.com/if-else-parseint/1 but we did a lot more for a lot less. Still think that `parseInt` only is a better solution, providing readability and cover a lot more with less. 

I think that I prove my point so at least for me .. will stop bragging about it. ;)

Issue #4406, opened 2020-03-25
core: css util performance


coryrylan:
If Core components use layout utils directly in the template then they must pull in the layout css as a dependency. This is shared across all components but still pulls in an excess amount of CSS. Investigate tools like purge CSS that can strip the unused CSS out of the bundle during the build.

https://purgecss.com

PR #4453, opened 2020-04-08
core: css perf/chores

coryrylan:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Internally in core components we use the layout utilities. We pull these into a common file for all components to share to help performance. However, this still pulls in all the layout utils into the internal bundle regardless if they are available globally due to shadow DOM. This adds an additional ~8kb gzip to the initial bundle of Core regardless of the component.

closes https://github.com/vmware/clarity/issues/4406 

## What is the new behavior?

In the final build step we now use a tool from Mozillia called Purge CSS to analyze our component templates and determine what utilities are used. Once determined we essentially treeshake all the unused layout utils from the final bundle. This reduces the initial bundle size by about 30% after gzip compression.

Our automated webpack bundle test pulls in the badge component and one icon.

**Before: ~25kb gzip
After: ~17kb gzip**

### Chores
- update to lit/lit-html
- update querySlot to support required elements
- add customElement decorator, same as lit-element but with our registration logic
- add missing typography tokens/docs
- add golden file track for core, not yet source controlled waiting on PR to land in web component analyzer 

* [ ] Yes
* [x] No


package.json
```diff
@@ -85,15 +85,16 @@
     "core:test:watch": "ng test clr-core --watch",
     "core:test:unit": "ng test clr-core --watch=false",
     "core:test:treeshaking": "webpack --config src/clr-core/test-bundles/webpack.config.js && bundlesize --config src/clr-core/test-bundles/bundlesize.json",
-    "core:build": "npm-run-all core:build:tokens core:build:sass -p 'core:build:ts -- --force' core:build:api -s core:build:package core:build:clean",
+    "core:test:bundles": "source-map-explorer dist/test-bundles/webpack.bundle.js",
+    "core:build": "npm-run-all core:build:tokens core:build:sass 'core:build:ts -- --force' core:build:sass:min -p core:build:api -s core:build:package core:build:clean",
```
coryrylan, line 89: The sass minification step must be after the TS builds now as the CSS is written into the JS output. After the TS is built we  use purge to "treeshake" the common css shared between components from the js file.

src/clr-core/styles/layout/_optimize.scss
```diff
@@ -6,15 +6,15 @@
...
+  --cllg: var(--cds-token-layout-space-lg, #{$cds-token-layout-space-lg-static});
+  --clxl: var(--cds-token-layout-space-xl, #{$cds-token-layout-space-xl-static});
 }
 
-$cds-token-layout-space-xs: var(--cdlsxs, $cds-token-layout-space-xs-static);
-$cds-token-layout-space-sm: var(--cdlssm, $cds-token-layout-space-sm-static);
-$cds-token-layout-space-md: var(--cdlsmd, $cds-token-layout-space-md-static);
-$cds-token-layout-space-lg: var(--cdlslg, $cds-token-layout-space-lg-static);
-$cds-token-layout-space-xl: var(--cdlsxl, $cds-token-layout-space-xl-static);
+$cds-token-layout-space-xs: var(--clxs);
```
coryrylan, line 16: By inverting the fallback we only have to define the fallback once instead for every single reference to the sass variables.

src/clr-core/styles/layout/docs/layout.stories.mdx
```diff
@@ -85,6 +85,25 @@ the component. [Read More: Grid Layout](?path=/docs/experimental-layout-grid--pa
...
 
+## Responsive
+
+The layout system provides many built-in responsive utilities to make it easy to
+manage layouts on many different screen sizes. All responsive utilities follow the same set of breakpoints.
+
+<Story id="experimental-design-tokens-stories--layout" />
+
+```css
+@media (min-width: 480px) { }
```
mathisscott, line 96: We may need to consider aligning these with the existing clarity breakpoints. In apps that use both Core and Clarity Angular/UI, responsive behavior would look weird as Core components would shift at different screen sizes from the other components.

I have no problem changing breakpoints globally but I think this is a bigger change that will have a hard time flying under the radar.

src/clr-core/test-bundles/bundlesize.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "path": "./dist/test-bundles/webpack.bundle.js",
-      "maxSize": "25.5 kB"
+      "maxSize": "17 kB"
```
mathisscott, line 29: 🏆

src/clr-core/styles/layout/docs/layout.stories.mdx
coryrylan: I can update them in this PR. I'll also update the example to pull from the token so they are not hard coded.

package.json
```diff
@@ -81,19 +81,23 @@
     "core:dev:build": "build-storybook -s ./src/clr-core/.storybook/public -c src/clr-core/.storybook -o dist/storybook/core",
     "core:lint": "npm run prelint && tslint -p . -c ./src/clr-core/tslint.json \"src/clr-core/**/*.ts\" -e \"src/clr-core/**/*.css.ts\" && stylelint --config ./src/clr-core/.stylelintrc \"src/clr-core/**/*.scss\"",
     "core:lint:fix": "npm run prelint && tslint --fix -p . -c ./src/clr-core/tslint.json \"src/clr-core/**/*.ts\" stylelint --config ./src/clr-core/.stylelintrc --fix \"src/clr-core/**/*.scss\"",
+    "core:golden": "npm run core:build:api",
+    "core:golden:fix": "npm run core:golden && cpy dist/clr-core/custom-elements.json ./src/clr-core/",
+    "core:golden:test": "diff -u src/clr-core/custom-elements.json dist/clr-core/custom-elements.json",
```
coryrylan, line 86: This sets up to use the API metadata for a golden file. Right now it is not source controlled until a PR in the web component analyzer lands.

src/clr-core/alert/alert-actions.element.ts
```diff
@@ -24,10 +24,9 @@ import { styles } from './alert-actions.element.css';
  * </cds-app-alert>
  * ```
  *
- * @beta 3.0
- * @element cds-alert-actions
+ * @beta
  */
-// @dynamic
+@customElement('cds-alert-actions')
```
coryrylan, line 29: Same API as lit-element but uses our registration code under the hood.

src/clr-core/internal/decorators/element.ts
```diff
@@ -0,0 +1,64 @@
...
+  initializer?: Function;
+  extras?: ClassElement[];
+  finisher?: <T>(classDef: Constructor<T>) => undefined | Constructor<T>;
+  descriptor?: PropertyDescriptor;
+}
+
+// tslint:disable-next-line:interface-over-type-literal
+export type Constructor<T> = {
+  new (...args: any[]): T;
+};
```
coryrylan, line 64: This will be removed in a future version of TypeScript once they implement the type defs natively.

src/clr-core/styles/tokens/tokens.yml
```diff
@@ -179,6 +179,66 @@ props:
     type: 'number'
     category: 'typography'
 
+  # global font sizes
```
coryrylan, line 182: Added these to match the weight tokens. These were useful for forms where there are labels and elements that dont directly match the content attributes applications would use.

docs/CODING_GUIDELINES_CORE.md
```diff
@@ -209,6 +209,19 @@ adhere to the following guidelines for our Core Web Component codebase.
   }
   ```
 
+  Content that is projected into a slot should not be styled by the host web
+  component. The component cannot make assumptions about the layout of the content
+  projected provided by the application. If the user needs certain layouts
+  defined within the projected content they can use layout utilities.
+
```
mathisscott, line 216: `content provided by the application` not `content projected provided by the application`

also, i'm not sure what we mean by `should not be styled`. maybe a little more clarification there. it seems like it is talking about laying out the content. but the phrasing here sounds like any styling...

src/clr-core/alert/alert-actions.element.ts
mathisscott: We had trouble in the past with Lit's `customElement` decorator. That's why we stopped using it. Are those issues no longer a problem for us?

src/clr-core/internal/decorators/query-slot.ts
```diff
@@ -20,21 +20,42 @@ const standardQuery = (descriptor: PropertyDescriptor, element: any) => ({
...
     // tslint:disable-next-line:no-any decorator
     name?: PropertyKey
   ): any => {
     const descriptor = {
       get(this: LitElement) {
-        return this.querySelector(selector);
+        const ref = this.querySelector(selector);
+
+        if (config) {
+          const message = config.requiredMessage
```
mathisscott, line 45: (nit) This can maybe be simplified to...

```
const message = config.requiredMessage || `The <${selector}> element is required to use <${this.tagName.toLocaleLowerCase()}>`;
```

src/clr-core/internal/decorators/query-slot.ts
```diff
@@ -20,21 +20,42 @@ const standardQuery = (descriptor: PropertyDescriptor, element: any) => ({
...
+            ? config.requiredMessage
+            : `The <${selector}> element is required to use <${this.tagName.toLocaleLowerCase()}>`;
+
+          if (!ref && config.required === 'error') {
+            throw new Error(message);
+          }
+
+          if (!ref && config.required === 'warning') {
```
mathisscott, line 53: so if `config.required` comes in as anything other than `warning` or `error` no message is shown? i know that we are using typescript but i don't think that prevents endusers from making a mistake. could we simplify this a bit and make `warning` the fallback? in the event that a value is passed but isn't expected?

```
if (!ref && config.required) {
  if (config.required === 'error') {
    throw new Error(message);
  } else {
    console.warn(message);
  }
}
```

src/clr-core/styles/layout/docs/layout.stories.mdx
```diff
@@ -85,6 +85,25 @@ the component. [Read More: Grid Layout](?path=/docs/experimental-layout-grid--pa
...
+
+The layout system provides many built-in responsive utilities to make it easy to
+manage layouts on many different screen sizes. All responsive utilities follow the same set of breakpoints.
+
+<Story id="experimental-design-tokens-stories--layout" />
+
+```css
+@media (min-width: 576px) { }
+
+@media (min-width: 768px) { }
```
mathisscott, line 98: ❤️

src/clr-core/styles/tokens/tokens.yml
mathisscott: Could we mark these as something other than `typography`? It sounds like we want to use them internally. But they are presented as an alternative to the typography tokens that we want applications to use.

Also, I had/have a need for `11px` in buttons. Do we go there? Or should I just calc() it? Where should we draw the line on how many of these we offer?

src/clr-core/styles/typography/typography.stories.ts
```diff
@@ -73,6 +73,21 @@ export const weights = () => {
...
 };
 
+export const sizes = () => {
+  return html`
+    <cds-demo cds-layout="vertical gap:md">
+      <p cds-text="body" style="font-size: var(--cds-token-typography-font-size-0)">The quick brown foxes jumps over the lazy dog.</p>  
+      <p cds-text="body" style="font-size: var(--cds-token-typography-font-size-1)">The quick brown foxes jumps over the lazy dog.</p>
+      <p cds-text="body" style="font-size: var(--cds-token-typography-font-size-2)">The quick brown foxes jumps over the lazy dog.</p>
+      <p cds-text="body" style="font-size: var(--cds-token-typography-font-size-3)">The quick brown foxes jumps over the lazy dog.</p>
+      <p cds-text="body" style="font-size: var(--cds-token-typography-font-size-4)">The quick brown foxes jumps over the lazy dog.</p>
```
mathisscott, line 83: Are we intending for end users to use these? It seems like a step back to me.

src/clr-core/styles/tokens/tokens.yml
```diff
@@ -179,6 +179,66 @@ props:
...
+  # global line-heights
+  - name: line-height-0
+    value: 1
+    type: 'number'
+    category: 'typography'
+  - name: line-height-1
+    value: 1.1
+    type: 'number'
+    category: 'typography'
+  - name: line-height-2
```
mathisscott, line 225: line-height should be in ems. I've run into problems using just the plain number (even though it's supposed to work as expected).

i also don't understand why these are necessary.

src/clr-core/styles/typography/typography.stories.ts
```diff
@@ -85,6 +100,25 @@ export const inline = () => {
...
+export const lineHeights = () => {
+  return html`
+    <style>
+      .line-heights-demo p {
+        background: var(--cds-token-color-neutral-300);
+        margin: 0
+      }
+    </style>
+    <cds-demo cds-layout="vertical gap:xs" class="line-heights-demo">
+      <p style="line-height: var(--cds-token-typography-line-height-0)">--cds-token-typography-line-height-0</p>
```
mathisscott, line 112: Are we intending for end users to use these? It seems like a step back to me.

src/clr-core/alert/alert-actions.element.ts
coryrylan: The issue with the lit element decorator is that it doesnt check if the element has already been registered so it will throw an exception if imported more than once. This decorator is the same api but instead of registering directly it calls our safe register func under the hood.

src/clr-core/styles/tokens/tokens.yml
coryrylan: I can make them all private and internal. Thats what they were intended for. I'll add 11px and 13px as well since those are used in a lot of places. If we mark these internal then we can change them at anytime.

src/clr-core/styles/typography/typography.stories.ts
coryrylan: Removing not meant to be public

src/clr-core/styles/tokens/tokens.yml
coryrylan: I dont think they are necessary thinking about it. There are cases where we need sizes internally but line heights look like they will be more specific to the use case so I'll pull them out.

src/clr-core/styles/typography/typography.stories.ts
coryrylan: nope will be internal, will remove the story

src/clr-core/styles/tokens/tokens-build.js
```diff
@@ -9,32 +9,52 @@
...
 
-function getPropValue(prop) {
-  return isRelativePxValue(prop) ? getRemValue(prop) : prop.value;
+function getPropValue(prop, base = 20) {
+  return isRelativePxValue(prop) ? getRemValue(prop, base) : prop.value;
+}
+
+function getPropString(prop, base = 20) {
+  // if the token is private but a relative value we generate a unique css custom prop
+  // to be able dynamically adjust it for apps using base 16px font size
```
mathisscott, line 34: Why are we focused specifically on base 16px as an alternative?

src/clr-core/styles/tokens/tokens.yml
```diff
@@ -158,26 +158,84 @@ props:
     value: 200
     type: 'number'
     category: 'typography'
+    private: true
   - name: font-weight-regular
     value: 400
     type: 'number'
     category: 'typography'
+    private: true
```
mathisscott, line 166: We did we make the font-weight typography helpers private?

src/clr-core/styles/tokens/tokens.yml
```diff
@@ -434,6 +492,7 @@ props:
     value: '32px'
     type: 'px'
     category: 'typography'
+    legacy: true
```
mathisscott, line 495: why didn't we mark the `h0-font-size` token as legacy?

src/clr-core/styles/tokens/tokens-build.js
coryrylan: This func in particular generates a unique id for relative size private tokens so we can still scale rem values easily but also discourage apps from using them since they are a random string value. So if the app is already using base 16 then they simply add one line change and the relative values auto scale properly for them without much work.

Example of generated code

```scss
// tokens/index.scss 
// Non relative private values are static and dont create a public css prop
$cds-token-layout-grid-cols: 12;

// relative private values create a private css prop to allow easy change of scale if neeeded
$cds-token-typography-font-size-2: var(--cds_alqnkoi63, 0.60rem);  // 12px base 20

// tokens/_public.scss
// consumers discouraged from using private vars but still can easily use base 16 if migrating to Clarity by applying the attribute
[cds-base-font="16"] {
  --cds_alqnkoi63: 0.7500rem; // 12px base 16
}
```

src/clr-core/styles/tokens/tokens.yml
coryrylan: The attr utils no so I think this is a mistake on my part

docs/CODING_GUIDELINES_CORE.md
```diff
@@ -60,16 +60,16 @@ adhere to the following guidelines for our Core Web Component codebase.
   `clr-core/my-toggle/my-toggle.element.ts`.
 
 * Stateless Utilities should be named appropriately in the `utils` directory.
-  `clr-core/common/utils/register.ts`.
+  `clr-core/internal/utils/register.ts`.
```
jeeyun, line 63: I was making these changes in my PR too. :) Thank you for updating these.

PR #4461, opened 2020-04-10
feat: introduce clr-form-placeholder solid color and set disabled opacity to 0.55

bdryanovski:
Adding a `clr-form-placeholder` SCSS and CSS variable to set the input/textarea/select and every other component placeholder color to a solid one and remove the `opacity` option on it. In addition, change the opacity to all disabled items from 0.54 to 0.55. 

The changes try to improve the accessibility of the colors to be more closely to 3:1 ration - and try to avoid the use of opacity as a form of color shading.

The breaking change from this PR is the change of 0.01 opacity for placeholder color and disabled form elements.

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: N/A

* [x] Yes
* [ ] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/forms/styles/_containers.clarity.scss
+++ b/packages/angular/projects/clr-angular/src/forms/styles/_containers.clarity.scss
@@ -11,7 +11,52 @@
   }
 
   .clr-form-control-disabled {
-    @include css-var(opacity, clr-forms-disabled-opacity, $clr-forms-disabled-opacity, $clr-use-custom-properties);
+    @include css-var(color, clr-forms-text-disabled-color, $clr-forms-text-disabled-color, $clr-use-custom-properties);
+
+    label,
+    .clr-control-label {
+      @include css-var(
+        color,
+        clr-forms-label-disabled-color,
+        $clr-forms-label-disabled-color,
+        $clr-use-custom-properties
+      );
+    }
+
+    .clr-input,
+    .clr-textarea,
+    .clr-select {
+      @include css-var(
+        color,
+        clr-forms-text-disabled-color,
+        $clr-forms-text-disabled-color,
+        $clr-use-custom-properties
+      );
+      @include css-var(
+        border-bottom-color,
+        clr-forms-border-disabled-color,
+        $clr-forms-border-disabled-color,
+        $clr-use-custom-properties
+      );
+    }
+
+    input[type='range']::-webkit-slider-thumb {
+      @include css-var(
+        background-color,
+        clr-forms-border-disabled-color,
+        $clr-forms-border-disabled-color,
+        $clr-use-custom-properties
+      );
+    }
+
+    .clr-subtext {
+      @include css-var(
+        color,
+        cclr-forms-subtext-disabled-color,
+        $clr-forms-subtext-disabled-color,
+        $clr-use-custom-properties
+      );
+    }
   }
 
   .clr-form-control-multi {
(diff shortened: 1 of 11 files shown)


src/clr-angular/forms/styles/_variables.forms.scss
```diff
@@ -12,9 +12,10 @@ $clr-forms-label-color: $clr-color-neutral-800 !default;
 $clr-forms-text-color: $clr-color-neutral-1000 !default;
 $clr-forms-invalid-color: $clr-color-danger-800 !default;
 $clr-forms-subtext-color: $clr-color-neutral-600 !default;
+$clr-forms-placeholder-color: $clr-color-neutral-600 !default;
 $clr-forms-border-color: $clr-color-neutral-500 !default;
 $clr-forms-focused-color: $clr-color-action-600 !default;
-$clr-forms-disabled-opacity: 0.54 !default;
+$clr-forms-disabled-opacity: 0.55 !default;
```
hippee-lee, line 18: IMO - this shouldn't be hard-coded. 
Could this update the generic forms variable to 0.55 and use that instead?
```--clr-forms-disabled-opacity: #{$clr-forms-disabled-opacity};``` => 0.54. 
It seems where ever $clr-forms-disabled-opacity is needed would also need this enhancement.

Comment (bdryanovski) [reply src/clr-angular/forms/styles/_variables.forms.scss:18]: Sure, will make the needed changes tomorrow and will ping you :)

Comment (gnomeontherun) [reply src/clr-angular/forms/styles/_variables.forms.scss:18]: Actually, I want to _remove_ the opacity and shade the label/helper text colors. Right now, this is still inaccessible. https://github.com/vmware/clarity/issues/4378

Comment (bdryanovski) [reply src/clr-angular/forms/styles/_variables.forms.scss:18]: After checking this again, it seems that is already done 

```scss
--clr-forms-disabled-opacity: #{$clr-forms-disabled-opacity}; // 0.55
```

My initial idea was to drop overall the opacity as a way of creating colors and use solid versions of them. But we gonna need more than one color for that. For example, the `border-bottom` of the inputs and the placeholder text color is not the same color. I had one more reason for not doing that when I tested it .. but could not remember right now.

In feature PR could try to add this change but right now the focus of it was the placeholder text inside the input/textarea/selects and to align both the solid color and opacity level to the same value from the design spec.

gnomeontherun:
The problem with opacity is that it was a shortcut and doesn't let us control the individual colors of text accurately. The fix here should be to remove opacity and rely on color tokens to set the colors in both light and dark. I don't want to just change the opacity by .01 and persist the problems.

bdryanovski:
I have to wait for the Color Tokens until I could add the solid colors properly tested. But it seems that it won't take that much work that I initially expected.

gnomeontherun:
I'm not sure what you mean about color tokens, the design tokens are only in Core for now and we can fix this by either using existing colors or adding new variables as necessary. Hopefully not necessary, but I want to get this closed out.

bdryanovski:
Screen Shot 2020-06-04 at 4 01 31 PM

Screen Shot 2020-06-04 at 4 01 06 PM

colinreedmiller:
The color system that will be implemented in Core will provide the proper values for these colors without using transparency to create disabled states.

This work is more or less complete and will be finalized in concert with the Core color system implementation. The current agreement is that once implemented and fine tuned, the system will provide the basis for a similar approach for Clarity Angular.

bdryanovski:
Sure will address only the color change mention here #4753 for the border. Right now in this PR from what I quickly checked it was 500 ..not 700

Issue #3953, opened 2019-10-30
Core Component: Modal


coryrylan:
# Core Modal
When implimenting the modal it may be a good time to check to see if the modal can internally use the native dialog element. It provides some needed functionality as well as a11y. There is a supported polyfill for the dialog API.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
- https://github.com/GoogleChrome/dialog-polyfill
- https://gist.github.com/samthor/babe9fad4a65625b301ba482dad284d1

## API Proposal (Angular Example Usage)

```html
<cds-modal>
  Content...
</cds-modal>

<cds-modal (closedChange)="closed = $event.detail" size="lg">
  <cds-modal-header>I have a nice title</cds-modal-header>
  <cds-modal-content>
    <p>But not much to say...</p>
  </cds-modal-content>
  <cds-modal-actions>
    <div cds-layout="horizontal gap:sm align:right">
      <cds-button (click)="modal.close()">Cancel</cds-button>
      <cds-button (click)="modal.save()">Save</cds-button>
    </div>
  </cds-modal-actions>
</cds-modal>
```

### Properties
`size` - Adjust the modal size styles using tshirt style size options
`closable` - use false to disable the close button and background click, defaults to true

### Events
`closedChange` - event for when modal's close button is clicked or the ESC key is pressed

### Slots
`cds-modal-header` - header content
`cds-modal-header-actions` - header action content
`cds-modal-content` - body content
`cds-modal-actions` - actions content

mathisscott:
Tech specs are in progress. Notes from early discussions...

The modal will need the following utilities...

1) focusTrap for a11y
2) focusReturn for a11y
3) scrollKiller for a11y
4) backdrop (which may or may not be the scrollKiller as well)

PR #4479, opened 2020-04-16
feat(modal): core modal

jeeyun:

## PR Checklist

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: #3953 

* [ ] Yes
* [x] No


src/clr-core/alert/alert.stories.ts
```diff
@@ -65,7 +65,7 @@ export const API = () => {
       .iconTitle=${iconTitle}
       .size=${size} 
       .status=${alertStatus}
-      @closedChange=${action('closeChanged')}>
+      @closedChange=${action('closedChange')}>
```
jeeyun, line 68: I'm wondering if we should deviate from Angular event name and call it "clickedClose" because we aren't actually going to change the state of the component to closed. Thoughts on this? Same question for alerts...

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    if (document.activeElement) {
+      this.previousFocus = document.activeElement;
+    }
+
+    // add class to prevent scroll on body
+    document.body.classList.add('no-scrolling');
```
jeeyun, line 30: Angular version had this as a separate `ScrollingService` class, but it was only being used by modal. I moved this into this base class instead. Let me know if you think if this line of logic (with the corresponding remove) makes more sense to live in Modal class instead.

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+      // we completely delete all traces of the removed DOM objects.
+      delete this.topReboundEl;
+      delete this.bottomReboundEl;
+    }
+  }
+
+  protected render() {
+    return html`<slot></slot>`;
+  }
+}
```
mathisscott, line 117: Does the focus-trap have unit tests?

src/clr-core/modal/modal.element.scss
```diff
@@ -0,0 +1,120 @@
...
+
+.private-host {
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  z-index: 1050; // TODO: restore using map-get to get the z-index
+  display: flex;
+  flex-direction: column;
```
mathisscott, line 42: We should be using layouts instead of flex styles.

src/clr-core/modal/modal.element.scss
```diff
@@ -0,0 +1,120 @@
...
+.modal-dialog {
+  position: relative;
+  z-index: 1050; // TODO: restore using map-get to get the z-index
+  width: var(--width);
+  max-width: 100%;
+
+  $modal-boxshadow-y: $cds-token-space-size-1;
+  $modal-boxshadow-spread: $cds-token-space-size-2;
+
+  .modal-content {
```
mathisscott, line 73: the `.modal-content` selector does not need to be nested inside the `.modal-dialog` selector.

Why do we need the classnames?

src/clr-core/modal/modal.element.scss
```diff
@@ -0,0 +1,120 @@
...
+    border-radius: var(--border-radius);
+    box-shadow: 0 $modal-boxshadow-y $modal-boxshadow-spread $modal-boxshadow-spread
+      rgba(var(--content-box-shadow-color), 0.2);
+    line-height: 0; // this is so that what goes inside content doesn't have weird extra spaces..
+  }
+}
+
+.modal-title {
+  color: var(--title-color);
+  font-size: $cds-token-typography-h3-font-size;
```
mathisscott, line 84: If users can't change the size of the title's font, it seems like the rest of the customizations are less relevant. Could we allow them to change the font-size here (maybe they want an h1 or h4? and then set the line-height to a sane value they can override with CSS if the unthinkable occurs and they want a 2em line-height with h4 text? 😂 

A sane line-height for header text (from Colin's recommendations) appears to be `1.1em` or thereabouts. These titles shouldn't wrap often.

mathisscott: Also, why are not using the line-height erasers here?

src/clr-core/modal/modal-header.element.scss
```diff
@@ -0,0 +1 @@
+@import './../styles/mixins/utils';
```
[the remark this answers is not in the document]
mathisscott: How are we doing the layout in the modal-header?

src/clr-core/modal/modal-actions.element.scss
```diff
@@ -0,0 +1 @@
+@import './../styles/mixins/utils';
```
mathisscott, line 1: How are we doing the layout in the modal-footer/actions area?

src/clr-core/alert/alert.stories.ts
coryrylan: This is not exclusive to Angular most frameworks encourage a 1:1 mapping between a property and event as well as events being present tense. Since this is also tied to keyboard users so click I don't think is quite right as well. I would keep the same convention and update it to be `closeChange`. This will help ensure our properties and events are consistent as well as keep `close` available as a property if we need it in the future (hopefully not). The convention will be especially important in the future with tooling support for things like React wrappers which have to parse the events.

```html
<!-- Vue -->
<cds-modal v-bind:close="closeValue" v-on:closeChange="doSomething()">
<cds-modal :close="closeValue" @closeChange="doSomething()">

<!-- React Wrapper -->
<CdsModal close={closeValue} onClose={() => this.doSomething()} />

<!-- Svelte native & html -->
<CdsModal bind:close="closeValue" on:close={doSomething} />
<cds-modal bind:close="closeValue" on:closeChange={doSomething}>

<!-- Angular -->
<cds-modal [close]="closeValue" (closeChange)="doSomething()">

<!-- Angular JS -->
<cds-modal close="$ctrl.closeValue" on-close-change="$ctrl.doSomething()">
```

src/clr-core/internal/base/focus-trap.base.ts
coryrylan: Should update to use cds-layout attr described below

src/clr-core/modal/modal-actions.element.scss
coryrylan: The decision discussed earlier was projected content from an API point we avoid managing layouts to give them more flexibility. So bother actions and header dont need any styles so I think both the `scss` files could be removed.

There is a outline about it here
https://github.com/vmware/clarity/blob/master/docs/CODING_GUIDELINES_CORE.md#web-component-conventions

src/clr-core/modal/modal.element.scss
coryrylan: If I remember I think from our previous discussion we decided to treat the header title the same as layouts and dont apply styles. This will make the styles API smaller/simpler as well as give them more flexibility. Example use case they can add a heading with sub heading or easily control layout with additional actions.

```html
<cds-modal>
  <cds-modal-header>
    <h3 cds-text="title">My Modal</h3>
  </cds-modal-header>
</cds-modal>
```

I think that would be the most flexible option. If we dont want to go that way then all the title styles should exist under `cds-modal-header`.

```scss
// modal-header.element.scss
:host {
   --color: $cds-token...;
   --font-size: $cds-token...;
   --font-weight: $cds-token...;
   --line-height': $cds-token...;
   --letter-spacing': $cds-token...;
   // font family automatically inherits from the app so it does not need to be a exposed prop
}
```

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /** If false, the modal will not render the close button.  */
+  @property({ type: Boolean })
+  closable = true;
+
+  /** Sets the overall height and width of the modal and icon based on value */
+  @property({ type: String })
+  size: 'default' | 'sm' | 'lg' | 'xl';
+
+  private ariaLabel = `aria-${this._idPrefix}${this._uniqueId}`;
```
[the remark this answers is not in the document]
coryrylan: Probably wont get to the required option on the property decorator before this gets merged. When I update the decorator I'll go back through all the components and update them.

src/clr-core/alert/alert.stories.ts
jeeyun: Thank you both. Good point about it also mapping to keyboard event. I'll update modal and alerts to `closeChange` (from `closedChange` in the case of alert).

src/clr-core/styles/utils/_a11y.scss
```diff
@@ -0,0 +1,23 @@
...
+  border: 0;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  white-space: nowrap;
+  top: 0;
+  left: 0;
+}
+
+@include exports('a11y.clarity') {
```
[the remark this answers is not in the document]
jeeyun: Artifact from angular package. Will remove the exports.

src/clr-core/internal/base/focus-trap.base.ts
jeeyun: Same logic as the button base, we test it out by testing the components that extend it.

src/clr-core/modal/modal-actions.element.scss
jeeyun: I think this may apply to alert subcomponents as well. Should I create a separate ticket for alerts?

coryrylan: Yeah anything we find related to other components we should make a separate ticket I think. That way someone can go in and evaluate all of the components at once without slowing down an open PR.

src/clr-core/modal/modal.element.scss
jeeyun: Moved it out! 

I don't know if we need to keep these sass variables (carried over from Angular). I can get rid of them and use the tokens directly. Or are you asking why `.modal-dialog` and `.modal-content` classes exist?

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+      return;
+    }
+
+    if ($event.target && this.hostElement) {
+      if (
+        !this.hostElement.contains($event.target) ||
+        $event.target === this.topReboundEl ||
+        $event.target === this.bottomReboundEl
+      ) {
+        this.hostElement.focus();
```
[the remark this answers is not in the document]
jeeyun: It focuses back to the host element if you try to focus on the rebound element.

src/clr-core/modal/modal-header.element.scss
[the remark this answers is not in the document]
jeeyun: Removing the scss files in the subcomponents. I'll refactor the header a bit more to be less opinionated in dictating its layout per your and @coryrylan's comment.

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+    window.addEventListener('keydown', this.fireEventOnEscape);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    window.removeEventListener('keydown', this.fireEventOnEscape);
+  }
+
+  closeModal() {
+    this.closedChange.emit(true);
```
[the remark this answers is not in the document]
mathisscott: Yeah. I tracked that down and actually deleted this comment. Yet it was still posted in the review. 🤷‍♂️

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+                    <slot name="modal-header"></slot>
+                  </span>
+              </div>
+              <div cds-layout="align:right">
+                <slot name="modal-header-actions"></slot>   
+              </div>
+              ${
+                this.closable
+                  ? html`
+                <button
```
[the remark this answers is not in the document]
jeeyun: We went back/forth on this a bit. Before the `cds-icon-button` was a separate component, I started out with `cds-button` but the default styling and functionality didn't quite fit well in modal and we discussed and changed to a plain button.  We are already addressing a11y concerns / features in this component. Do you see other benefits of using our button component here?

src/clr-core/modal/modal.element.scss
mathisscott: If we have `cds-modal-content` and other components, why do we need to use the `.modal-content` classname?

I don't see internal wrappers in those components (like i've used to help with layouts and LHEs in tags and buttons) so I don't understand why we can't use the component tagName as the selector.

src/clr-core/modal/modal-actions.element.scss
mathisscott: I'm doing an alerts refactor to use layouts after i get the PR up for buttons, tags, and badges. i can take care of this if you wouldn't mind creating a github issue and assigning it to me...

src/clr-core/styles/utils/_a11y.scss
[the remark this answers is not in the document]
jeeyun: Actually removing this file completely as we don't need it anymore.

src/clr-core/internal/base/focus-trap.base.ts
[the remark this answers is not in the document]
mathisscott: I see now. thanks. might be a good datapoint for the comment above about abstracting/simplifying the conditionals...

src/clr-core/modal/modal.element.ts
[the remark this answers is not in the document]
jeeyun: Renaming to not conflict with the standard existing `aria-label` attribute.

src/clr-core/modal/modal-actions.element.scss
jeeyun: Filed #4484 and assigned to @mathisscott.

src/clr-core/styles/utils/_a11y.scss
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
jeeyun: Removed the file and updated.

src/clr-core/modal/modal.element.scss
jeeyun: I think your suggestion makes sense to keep in line with our guidelines. I'm taking out the styling (and customizability) of the title so that the end user can define the look and feel of the header text using appropriate typography.

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+    super.disconnectedCallback();
+    window.removeEventListener('keydown', this.fireEventOnEscape);
+  }
+
+  closeModal() {
+    this.closedChange.emit(true);
+  }
+
+  private fireEventOnEscape = (e: KeyboardEvent) => {
+    if (e.keyCode === ESC || e.key === 'Esc') {
```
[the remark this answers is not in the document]
jeeyun: As discussed, created a separate ticket (#4487) to address this in both modal and base button.

src/clr-core/modal/modal.element.scss
jeeyun: since these classnames are not direct parent of each selector, I think they are still needed in these specific cases. Take a look and let me know if you think otherwise.

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    super.disconnectedCallback();
+  }
+
+  protected onFocusIn($event: any) {
+    if (FocusTrapTracker.getCurrent() !== this.hostElement) {
+      return;
+    }
+
+    if ($event.target && this.hostElement) {
```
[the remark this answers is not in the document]
mathisscott: Thoughts on this?

src/clr-core/internal/utils/dom.ts
```diff
@@ -17,3 +19,39 @@ export function getElementWidthUnless(element: HTMLElement, unless: boolean) {
...
+
+export function isHTMLElement(el: any) {
+  if (el) {
+    return el instanceof HTMLElement;
+  }
+  return false;
+}
+
+export function addAttributeValue(element: HTMLElement, attr: string, value: string) {
+  if (element) {
```
mathisscott, line 31: It looks like this is appending an attributeValue. so if I have an element like '<div title="ohai">` and I pass in....

```
addAttributeValue(thatDiv, 'title', 'kthxbye')
```
...the result would be...
```
<div title="ohai kthxbye">
```
...instead of...
```
<div title="kthxbye">
```
But most HTML attrs aren't like that. This would seem to be most useful for adding classnames or (in the case of our layouts) adding layout designations.

Is this intentional?

src/clr-core/modal/modal.element.scss
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.modal-content {
+  $modal-boxshadow-y: $cds-token-space-size-1;
+  $modal-boxshadow-spread: $cds-token-space-size-2;
+
+  background-color: var(--background-color);
+  border-radius: var(--border-radius);
+  box-shadow: 0 $modal-boxshadow-y $modal-boxshadow-spread $modal-boxshadow-spread
+    rgba(var(--content-box-shadow-color), 0.2);
```
mathisscott, line 70: My experience with this pattern has not been a positive one `rgba(var(--content-box-shadow-color), 0.2)`. You may want to verify it in Firefox and Edge...

src/clr-core/modal/modal.element.ts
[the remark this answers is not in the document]
mathisscott: At present, this button is being vertically centered based on line-height. There's also no real way of knowing how wide the button is, although I suspect it is just wide enough to enclose the icon.

There's CSS in `modal.element.scss` on lines 40, 41, and 52 that would not need to be there if we used an icon-button. Additionally, we could move `cds-icon-button.close` CSS into the icon button so it could be shared across components.

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+            <div class="modal-body">
+              <slot></slot>
+            </div>
+            <div cds-layout="p-t:sm p-t@md:md">
+              <slot name="modal-actions" cds-layout="align-stretch"></slot>             
+            </div>
+          </div>
+          <div class="clr-sr-only">${CommonStringsService.keys.modalContentEnd}</div>
+        </div>
+        <div class="modal-backdrop" aria-hidden="true"></div>
```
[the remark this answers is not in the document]
mathisscott: Thoughts here?

src/clr-core/styles/utils/_close.scss
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+// scss-lint:disable QualifyingElement
+button.close {
```
mathisscott, line 6: Why is our preference here to create global CSS styles over either reusing/extending another component or making a component like `cds-close-button` that we can re-use?

This seems inconsistent with how we've approached things thus far.

src/clr-core/modal/modal.element.ts
```diff
@@ -0,0 +1,140 @@
...
+  private idForAriaLabel = `${this._idPrefix}${this._uniqueId}`;
+
+  render() {
+    return html`
+      <div class="private-host" cds-layout="horizontal p:sm p@md:lg align:center">
+        <div class="modal-dialog" 
+              role="dialog" 
+              aria-modal="true" 
+              aria-labelledby="${this.idForAriaLabel}">
+          <div cds-layout="display:screen-reader-only">${CommonStringsService.keys.modalContentStart}</div>
```
mathisscott, line 73: All of these commonstrings will need to be replaced by the end user to something meaningful and unique to the modal. I don't think we want to handle i18n the same way in core as we do in angular. the strategy is to provide some default strings using commonstrings and warn end users that if they don't switch them out then they risk not being accessible.

because all of these sr-only and aria values are always and only strings, we could make them properties/attributes.

src/clr-core/internal/base/focus-trap.base.ts
mathisscott: I'm not a huge fan of testing everything through integration, although that should maybe be a deeper discussion outside of this context.

I'd rather know that the focus-trap is breaking and be able to zero in quickly to fix it than see that a change I made broke the datepicker, wizard, modal, combobox, signposts, and dropdowns and have to do the hop-skip-jump to realize what I really broke was the focus-trap.

When I test the modal, I care about things that are specific to the modal.

We should discuss this with @coryrylan at some point because we're going to be frequently stomping on each other's toes if we have philosophically opposed views on unit testing.

src/clr-core/modal/modal.element.scss
mathisscott: We don't have to carryover the SASS variables.

src/clr-core/internal/utils/dom.ts
jeeyun: Yup, this was intentional. As you correctly pointed out:

> This would seem to be most useful for adding classnames or (in the case of our layouts) adding layout designations.

This function is used to add and remove the `no-scrolling` by changing the `cds-layout` attr on body. An earlier suggestion in the review was to leverage our `cds-layout` in our focus trap base class to achieve this. We shouldn't remove other values that are set by the application as it pertains to our layout, and this function facilitates that by only adding or removing the value that's passed to this function. That's why the function name is called `addAttributeValue` as opposed to `setAttributeValue` (besides, this functionality is already available natively and wouldn't need this function).  

Do you have a suggestion to make this point clearer? Maybe a better name? Constrain this to layout attr? Or do you disagree that we should be using `cds-layout` attribute for this, and suggest we revert to using classnames?

src/clr-core/modal/modal.element.ts
jeeyun: Yes, I believe the consensus was to address this separately after this PR? 😁

src/clr-core/internal/utils/dom.ts
mathisscott: maybe `appendAttributeValue`? although I see no issues with `addAttributeValue` tbh. i just wanted to be sure i understood the intention. makes total sense in the case you describe. i don't think there's a need to change the name.

src/clr-core/modal/modal.element.ts
mathisscott: maybe... it _is_ Wednesday. so things might be getting foggy for me...

src/clr-core/styles/utils/_close.scss
jeeyun: I think creating a reusable `cds-close-button` makes sense here. 👍 Will work on this after the latest button PR (with icon button) is merged!

src/clr-core/modal/modal.element.scss
jeeyun: Removing the variables!

src/clr-core/styles/utils/_close.scss
coryrylan: Pulling in CSS into a component style sheet is not shared its duplicated across every component because of how sass compiles out the CSS. Because of this we should avoid patterns like this if we can.

Best bet would be to get `<cds-icon-button>` in and see if that meets our needs for a close button.
However we could get internal components up and running. I outlined what would need to be done. Since its non trivial amount of work it could be done as a follow up PR to this one. https://github.com/vmware/clarity/issues/4501

src/clr-core/modal/modal.element.ts
coryrylan: Yeah we decided to keep that a separate PR to work through and apply to all components. We may need a hybrid of the two but its still undecided how or what the API may look like. I was still planning on opening an issue with a couple of ideas we could discuss. 

```typescript
@property({ type: String, { required: 'warning', requiredMessage: 'must be set for a11y' } })
closeLabel = CommonStringsService.keys.close;
```

```html
<cds-modal close-label="Close Sign Up Modal">
  <h1>Sign Up</h1>
</cds-modal>
```

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (
+        !this.hostElement.contains($event.target) ||
+        $event.target === this.topReboundEl ||
+        $event.target === this.bottomReboundEl
```
Shijir, line 67: We don't need these checks right? Because the rebounder elements are outside of the host element so the first check will cover these cases too. In our Angular version, we have these check as we have strict focus trap config.

src/clr-core/internal/services/focus-trap-tracker.service.ts
```diff
@@ -0,0 +1,25 @@
...
+* This software is released under MIT license.
+* The full license information can be found in LICENSE in the root directory of this project.
+*/
+
+/**
+ * FocusTrapTracker is a static class that keeps track of the active element with focus trap,
+ * in case there are multiple in a given page.
+ */
+export class FocusTrapTracker {
+  private static focusTrapElements: Array<any> = [];
```
Shijir, line 12: nit: I see some element variable/properties are using `any` or `HTMLElement`. Should we just stick to `HTMLElement`?

src/clr-core/internal/services/focus-trap-tracker.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  static activatePreviousCurrent() {
+    this.focusTrapElements.shift();
+  }
+
+  static getCurrent() {
+    return this.focusTrapElements[0];
+  }
+}
```
Shijir, line 25: Awesome 👍

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+  private bottomReboundEl: any;
+  private onFocusInEvent: any;
+
+  @property({ attribute: false })
+  private hostElement = this;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    if (document.activeElement) {
```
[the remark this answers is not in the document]
mathisscott: That may not work though. AoT is very finicky when it comes to this stuff. We would want to verify that approach wouldn't break server-side rendering. I know in the past I had to put these guards around each call.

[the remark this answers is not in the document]
coryrylan: We do not need to have `isBrowser` checks within the custom elements at all, as they do not execute or get loaded on the server. The `registerElementSafely` will not attempt to register the element if not in the browser. Custom elements can not be rendered dynamically in node since the custom element registry does not exist in node but the outer HTML definition can.

Example I can render this bit of HTML in node fine

```html
<cds-modal>
  This content is SSR and is in the static HTML 
</cds-modal>
```

So for Angular Universal you must import the custom elements within the browser module only. 

Demo repo Angular Universal + Core:
https://github.com/coryrylan/core-demos/tree/master/angular-universal 

Load custom element in browser module
https://github.com/coryrylan/core-demos/blob/master/angular-universal/src/app/app.browser.module.ts#L2 // 

Use in angular component templates
https://github.com/coryrylan/core-demos/blob/master/angular-universal/src/app/app.component.html#L10 

The Vuepress docs for the new website has a similar setup

```javascript
core.js
import '@clr/core/alert';
import '@clr/core/button';
import '@clr/core/icon';
import { ClarityIcons, homeIcon, angleDoubleIcon, angleIcon} from '@clr/core/icon-shapes';

ClarityIcons.addIcons(angleDoubleIcon, angleIcon, homeIcon);
``` 

```javascript
// enhanceApp.js 
import './styles/index.scss';

export default ({
  Vue, // the version of Vue being used in the VuePress app
  options, // the options for the root Vue instance
  router, // the router instance for the app
  siteData, // site metadata,
  isServer
}) => {
  if (!isServer) {
    import('./core');
  }
}
```

You can render custom elements dynamically in node with additional setup on the server side either using a custom elements node shim like [JSDom](https://github.com/jsdom/jsdom)  which recently added [custom element support](https://github.com/jsdom/jsdom/pull/2548).

Or you can use a prerender with something like [puppeteer](https://developers.google.com/web/tools/puppeteer). This is how the Angular static site renderer Scully works https://github.com/scullyio/scully.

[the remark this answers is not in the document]
Shijir: I see. As you mentioned, `registerElementSafely` is not attempting to work during server-side rendering. It doesn't look like it's chocking during the pre-rendering process. So in this case, do we need to be checking if it's running in a browser?

src/clr-core/modal/modal.element.scss
```diff
@@ -0,0 +1,103 @@
...
+  --backdrop-opacity: 0.85;
+  --backdrop-color: #{$cds-token-color-neutral-900};
+  --box-shadow-color: #{$cds-token-color-action-500};
+  --border-radius: #{$cds-token-global-border-radius};
+  --border-color: #{$cds-token-color-action-600};
+  --background-color: #{$cds-token-color-neutral-0};
+  --close-icon-color: #{$cds-token-color-neutral-600};
+  --close-icon-color-hover: #{$cds-token-color-neutral-700};
+  --content-box-shadow-color: #{$cds-token-color-neutral-1000};
+  --width: calc(8 * #{$cds-token-space-size-13});
```
coryrylan, line 15: Some nits on the css api to make it a bit more flexible and follow the naming convention

`--backdrop-color` => `--backdrop-background`
`--box-shadow-color` => `--box-shadow`
`--background-color` => `--background`
`--border-color` => `--border`
`--content-box-shadow-color` => `--content-box-shadow`

We should lean towards shorthand props `--background` that way consumers have full control and we dont have to expose additional background-* props.

src/clr-core/modal/modal.element.scss
jeeyun: Thank you for pointing this out. Updated how we do this per discussion.

src/clr-core/styles/utils/_close.scss
jeeyun: Resolving so that this is addressed separately after this PR is merged.

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+import { html, LitElement } from 'lit-element';
+import { property } from '../decorators/property';
+import { FocusTrapTracker } from '../services/focus-trap-tracker.service';
+import { addAttributeValue, isHTMLElement, removeAttributeValue } from '../utils/dom';
+import { isBrowser } from '../utils/exists';
+
+export class CdsBaseFocusTrap extends LitElement {
+  private previousFocus: Element;
+  private topReboundEl: any;
+  private bottomReboundEl: any;
```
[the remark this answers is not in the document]
jeeyun: The `topReboundEl` and `bottomReboundEl` can be so I'll update those. Not the previousFocus though because `document.activeElement`'s return type is `Element`.

src/clr-core/internal/base/focus-trap.base.ts
jeeyun: Rebound elements are inside the host element in modal actually so it won't pass with the first check.

src/clr-core/modal/modal.element.scss
jeeyun: I'll make the changes.. `--content-box-shadow-color` is a bit different though as it literally only lets you customize the color:

```
:host {
  --content-box-shadow-color: rgba(0, 0, 0, 0.2);
}

.modal-content {
  box-shadow: 0 $cds-token-space-size-1 $cds-token-space-size-2 $cds-token-space-size-2 var(--content-box-shadow-color);
}
```

src/clr-core/internal/services/focus-trap-tracker.service.ts
jeeyun: I'll change it to type `Element` 👍

src/clr-core/internal/base/focus-trap.base.ts
Shijir: We cannot have the rebound elements inside the host element though. If they are outside, they will longer be able to catch the outside focus and bounce it back into the host element.

src/clr-core/internal/base/focus-trap.base.ts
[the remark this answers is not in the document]
coryrylan: For pre-rendering using with Scully or Puppeteer it works without checks since it uses a headless browser. Pre-rendering with Angular Universal or anything that renders to a string follows the same pattern as server side rendering. It must load the custom elements only in the browser.

The Universal demo I linked it could be updated to prerender. The code of where the custom elements are imported would stay the same. So in the pre-render scenarios we would not need to check if we are running in the browser.

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+  protected createFocusableOffScreenEl(): any {
+    if (isBrowser()) {
+      const offScreenSpan = document.createElement('span');
+      offScreenSpan.setAttribute('tabindex', '0');
+      offScreenSpan.classList.add('offscreen-focus-rebounder');
+      return offScreenSpan;
+    }
+  }
+
+  protected addReboundElements() {
```
coryrylan, line 83: With a `FocusTrapElement` interface this could be refactored out into a stand alone function 

```typescript
export function addReboundElementsToFocusTrapElement(focusTrapElement: FocusTrapElement) {
  focusTrapElement.topReboundElement = createFocusTrapReboundElement();
  focusTrapElement.bottomReboundElement = createFocusTrapReboundElement();

  if (focusTrapElement.parentElement) {
    focusTrapElement.parentElement.insertBefore(focusTrapElement.topReboundElement, focusTrapElement);
    if (focusTrapElement.nextSibling) {
      focusTrapElement.parentElement.insertBefore(focusTrapElement.bottomReboundElement, focusTrapElement.nextSibling);
    } else {
      focusTrapElement.parentElement.appendChild(focusTrapElement.bottomReboundElement);
    }
  }
}
```

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+  }
+
+  protected createFocusableOffScreenEl(): any {
+    if (isBrowser()) {
+      const offScreenSpan = document.createElement('span');
+      offScreenSpan.setAttribute('tabindex', '0');
+      offScreenSpan.classList.add('offscreen-focus-rebounder');
+      return offScreenSpan;
```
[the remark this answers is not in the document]
mathisscott: Great suggestions, Cory.

Only thing I'd like to point out is that in this function...

```
export function focusElementIfInCurrentFocusTrapElement(focusedElement: HTMLElement, focusTrapElement: FocusTrapElement) {
  if (FocusTrapTracker.getCurrent() === focusTrapElement && elementIsOutsideFocusTrapElement(focusedElement, focusTrapElement)) {
    focusTrapElement.focus();
  }
}
```

We are asking the function to be aware of `FocusTrapTracker`.  It shouldn't have knowledge of anything outside of itself. So maybe passing FocusTrapTracker's current element as an argument would be advised?

```
export function focusElementIfInCurrentFocusTrapElement(focusedElement: HTMLElement, focusTrapElement: FocusTrapElement, currentFocusTrapElement: HTMLElement) {
  if (currentFocusTrapElement === focusTrapElement && elementIsOutsideFocusTrapElement(focusedElement, focusTrapElement)) {
    focusTrapElement.focus();
  }
}
```

src/clr-angular/layout/main-container/_layout.clarity.scss
```diff
@@ -79,11 +79,12 @@
...
 
-  body.no-scrolling .main-container .content-container .content-area {
+  body.no-scrolling,
+  body[cds-layout='no-scrolling'] {
     overflow: hidden;
-  }
 
-  body.no-scrolling {
-    overflow: hidden;
+    .main-container .content-container .content-area {
```
mathisscott, line 86: We should note that this selector specifically targets Clarity-UI and is here only for integration between the two. Core will not be carrying this DOM structure over.

src/clr-core/internal/base/focus-trap.base.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+// register the element just for testing purposes
+registerElementSafely('cds-base-focus-trap', CdsBaseFocusTrap);
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'cds-base-focus-trap': CdsBaseFocusTrap;
+  }
+}
+
+xdescribe('modal element', () => {
```
mathisscott, line 19: Do we want to leave this xdescribe in? Our build will probably flag it...

src/clr-core/internal/base/focus-trap.base.ts
```diff
@@ -0,0 +1,35 @@
...
+export class CdsBaseFocusTrap extends LitElement {
+  protected focusTrap: FocusTrap;
+
+  @property({ type: Boolean })
+  __demoMode = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    if (!this.__demoMode) {
```
mathisscott, line 19: Even though it's internal, we want some tests on this demo-mode flag.

src/clr-core/internal/utils/dom.ts
```diff
@@ -18,6 +20,13 @@ export function getElementWidthUnless(element: HTMLElement, unless: boolean) {
   return '';
 }
 
+export function isHTMLElement(el: any) {
+  if (el) {
+    return el instanceof HTMLElement;
```
mathisscott, line 25: Just a minor thing. We could simplify this further to...

`return !!el && el instanceof HTMLElement`

src/clr-core/modal/modal.stories.ts
```diff
@@ -0,0 +1,234 @@
...
+          '--close-icon-color-hover': closeIconColorHover,
+          '--content-box-shadow-color': contentBoxShadowColor,
+          '--width': width,
+        })}
+      .sbdocs-preview > div[class^="css-"], .sbdocs-preview > div[class^="css-"] > div[class^="css-"]:first-child {
+        height: 500px;
+      }
+      body[cds-layout='no-scrolling'] {
+        overflow: auto !important;
+      }
```
coryrylan, line 75: nit: I dont think the body no scroll overrides in each story demo is needed now that the focus trap can be disabled.

src/clr-angular/layout/main-container/_layout.clarity.scss
jeeyun: Thank you. Yes, I removed that in the core's scss file. I'll put a comment here as well for extra clarity.

src/clr-core/internal/base/focus-trap.base.spec.ts
jeeyun: Ah I left it in by accident. Thanks for catching that.

PR #4482, opened 2020-04-17
fix: remove ng-cli test runner for core

coryrylan:
## PR Checklist

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Build related changes
* [x] CI related changes

## What is the new behavior?
- This removes the dependency between the web component tests and the Angular CLI.
- Improves Core dev build times with hot module reloading
- Fixes unit test coverage
- Adds required support for `@property` decorator as well as env hints
- Adds Vue and React to debug log output


src/clr-core/icon-shapes/icon.element.ts
```diff
@@ -55,6 +55,7 @@ export class CdsIcon extends IconMixinClass {
   private _shape: string;
   private _size: string;
 
+  @property({ type: String })
```
coryrylan, line 58: Found looking at lit docs that the prop has to be on the getter or it can cause unexpected behavior.

src/clr-core/import-map.importmap
```diff
@@ -0,0 +1,29 @@
+  
+{
+  "imports": {
+    "tslib": "/base/node_modules/tslib/tslib.es6.js",
```
coryrylan, line 4: This file tells chrome how to load the module natively. I think in the near future I will be able to automate this part using some existing 3rd party tools but for now its manual. This will also set us up to easily support pure JS/HTML/CDN setups with no build step.

src/clr-core/internal/css-vars/css-vars.spec.ts
```diff
@@ -9,13 +9,13 @@ import { cssVarsDefaultConfig, runCssVarsPolyfill } from './css-vars.js';
 
 describe('Utilities - CssVarsPolyfill ', () => {
   describe('runCssVarsPolyfill() ', () => {
-    it('calls cssVars() with default config', () => {
+    xit('calls cssVars() with default config', () => {
```
coryrylan, line 12: Need to circle back and fix these.

src/clr-core/internal/utils/global.ts
```diff
@@ -37,7 +37,6 @@ function getVersion() {
     angularVersion: getAngularVersion(),
     userAgent: navigator.userAgent,
   };
-  console.log(JSON.stringify(log, null, 2));
```
coryrylan, line 40: Need to circle back and make this run but not during tests since it gets noisy.

src/clr-core/package.js
```diff
@@ -21,7 +21,10 @@ const read = dir =>
       []
     );
 
-del.sync(['./dist/clr-core/**/*.{tsbuildinfo,ngsummary.json}', './dist/clr-core/*.{tsbuildinfo,ngsummary.json}']);
+del.sync([
+  './dist/clr-core/**/*.{tsbuildinfo,ngsummary.json,spec.js,spec.js.map,spec.d.ts}',
```
coryrylan, line 25: This setup runs the unit tests side by side with the code we ship which should help reduce any build differences between tests and lib code. When we run a full build this deletes any test artifacts.

src/clr-core/tsconfig.spec.json
```diff
@@ -1,11 +1,9 @@
...
-  "extends": "./tsconfig.json",
+  "extends": "./tsconfig.project.json",
   "compilerOptions": {
-    "outDir": "../../out-tsc/spec",
-    "types": ["jasmine", "node"],
-    "target": "es2015",
-    "noImplicitAny": false
+    "outDir": "../../dist/clr-core",
+    "types": ["jasmine"],
+    "strictNullChecks": false
```
coryrylan, line 6: I left this as false for the unit tests as it was a lot of work to change. Theres checks due to dom queries always possibly being null in the return signature.

package.json
```diff
@@ -149,6 +150,7 @@
     "@commitlint/cli": "8.0.0",
     "@commitlint/config-conventional": "8.0.0",
     "@ng-select/ng-select": "3.7.2",
+    "@open-wc/karma-esm": "2.13.23",
```
coryrylan, line 153: This package is a web component tool that adds native es module to karma.

mathisscott: 🏆

src/clr-core/alert/alert-actions.element.ts
```diff
@@ -6,7 +6,7 @@
 
 import { baseStyles, registerElementSafely } from '@clr/core/internal';
 import { html, LitElement } from 'lit-element';
-import { styles } from './alert-actions.element.css';
+import { styles } from './alert-actions.element.css.js';
```
mathisscott, line 9: ❤️

src/clr-core/test/utils.ts
```diff
@@ -55,7 +55,7 @@ function retry(fn: any, maxTries = 10, promise?: Promise<any>, promiseObject?: {
     });
 
   fn()
-    .then(result => {
+    .then((result: any) => {
```
mathisscott, line 58: LOL. I was hitting this issue earlier and did the same thing. 😂

src/clr-core/internal/utils/register.ts
```diff
@@ -26,3 +29,7 @@ export function registerElementSafely(tagName: string, elementClass: any) {
     addElementToRegistry(tagName, elementClass, window.customElements);
   }
 }
+
+function isStorybook() {
+  return window.location.href.includes('localhost:6006');
```
mathisscott, line 34: is this the best way to know if we're running inside storybook? it seems a little brittle...

src/clr-core/.storybook/main.js
```diff
@@ -26,13 +25,15 @@ module.exports = {
     });
 
     config.resolve.extensions.push('.ts');
-    config.resolve.extensions.push('.js');
-    config.resolve.alias = {
-      '@clr/core': path.resolve(__dirname, '../../../dist/clr-core'),
-      '@clr/core/*': path.resolve(__dirname, '../../../dist/clr-core/*'),
-    };
```
coryrylan, line 33: We dont need storybook to resolve these files as TS does this already with its path mapping

src/clr-core/internal/css-vars/css-vars.spec.ts
coryrylan: After some research it looks like it is not possible to spy on objects external to a native JS module. Even though we are used to it in angular it was due to everything being bundled or transpiled causing a unintentional side effect.

src/clr-core/internal/decorators/element.spec.ts
```diff
@@ -11,9 +11,22 @@ import { customElement } from './element.js';
...
     const component = customElements.get('test-custom-element-decorator');
     expect(component).toBe(TestElement);
   });
+
+  it('should support native decorator API proposal', () => {
+    const desc = { kind: 'class', elements: [''] };
+    customElement('test-custom-element-decorator-legacy')(desc as any).finisher(LegacyTestElement);
+    const component = customElements.get('test-custom-element-decorator-legacy');
+    expect(component).toBe(LegacyTestElement);
+  });
```
coryrylan, line 31: Each of our custom decorators has two logic branches, one for legacy TS behavior and another for the current TC39 decorator proposal syntax. This checks that the expected return matches the native signature for the browser.

src/clr-core/internal/decorators/property.ts
```diff
@@ -58,16 +65,56 @@ export function getDefaultOptions(
...
+  const tag = tagName.toLocaleLowerCase();
+  return (
+    `${capitalizeFirstLetter(
+      level
+    )}: ${propertyName} is required to use ${tag} component. Set the JS Property or HTML Attribute.\n\n` +
+    `${getAngularVersion() ? `Angular: <${tag} [${propertyName}]="..."></${tag}>\n` : ''}` +
+    `${getVueVersion() ? `Vue: <${tag} :${propertyName}="..."></${tag}>\n` : ''}` +
+    `${getReactVersion() ? `React: <${kebabCaseToPascalCase(tag)} ${propertyName}={...} />\n` : ''}` +
+    `${`HTML: <${tag} ${camelCaseToKebabCase(propertyName)}="..."></${tag}>\n`}` +
+    `${`JavaScript: document.querySelector('${tag}').${propertyName} = '...';\n\n`}`
```
coryrylan, line 99: Here if a property is required and missing we can warn with a code hint. This will adjust based on possible assumed application env.

core-warning

src/clr-core/internal/utils/exists.spec.ts
```diff
@@ -44,8 +44,8 @@ describe('Functional Helper: ', () => {
           levelThree: {
             notTrue: false,
             notStrung: '',
-            nullified: null,
-            notDefined: undefined,
+            nullified: null as any,
+            notDefined: undefined as any,
```
coryrylan, line 48: eslint fix

src/clr-core/internal/utils/framework.ts
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+// DO NOT USE for conditionally applying behavior based on the host framework.
+// These utilities are only for detecting host frameworks for debugging and logging.
+// Host framework specific code should exist outside of the @clr/core package.
```
coryrylan, line 9: I want to make it clear here we do not change logical behavior in Core for external frameworks. This is to help our logging tool for Github issues. `window.CDS.getVersion()`. Right now we can only get the active version for Angular. React and Vue we can detect but currently do not expose the running version.

src/clr-core/internal/utils/register.ts
```diff
@@ -26,3 +29,7 @@ export function registerElementSafely(tagName: string, elementClass: any) {
...
+}
```
coryrylan, line 35: This should be moved with the other env detection logic.

src/clr-core/internal/utils/register.ts
coryrylan: Yeah I didnt like this one either. I can dig a bit more. There isnt much exposed to the runtime code since they put the stories in iframes to sandbox them.

src/clr-core/.storybook/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": []
```
jeeyun, line 2: Will this remain empty?

src/clr-core/internal/utils/register.ts
```diff
@@ -7,16 +7,20 @@
 import curryN from 'ramda/es/curryN';
 import { elementExists, existsInWindow, isBrowser } from './exists.js';
 import { setupCDSGlobal } from './global.js';
+import { isStorybook } from './framework.js';
 
 const addElementToRegistry = curryN(
   3,
   (tagName: string, elementClass: any, registry: { define: (a: string, b: any) => {} }) => {
-    if (elementExists(tagName)) {
+    if (elementExists(tagName) && !isStorybook()) {
```
jeeyun, line 15: This is great! 👍

src/clr-core/import-map.importmap
jeeyun: Will the build process fail if we miss a 3r party import here? I can already see that I'll need to add a couple for the modal PR, for example.

src/clr-core/.storybook/.babelrc
coryrylan: Yeah, this removes the default presets that storybook applies to our code.

src/clr-core/internal/utils/string.ts
```diff
@@ -26,6 +26,18 @@ export function camelCaseToKebabCase(value: string) {
...
+export function kebabCaseToCamelCase(str: string) {
+  return str
+    .split('-')
+    .map((item, index) => (index ? item.charAt(0).toUpperCase() + item.slice(1).toLowerCase() : item))
+    .join('');
+}
+
+export function kebabCaseToPascalCase(string: string) {
+  const camelCase = kebabCaseToCamelCase(string);
+  return camelCase[0].toUpperCase() + camelCase.substr(1);
```
jeeyun, line 38: This line is logically equivalent to `capitalizeFirstLetter` function below. Curious if there's a reason for using `substr` in this one and `slice` in the other one?

src/clr-core/import-map.importmap
coryrylan: Yeah this will fail during the build if a file is missing and not listed here. Will be automated away in the near future.

src/clr-core/internal/utils/string.ts
coryrylan: Updated to use the same function

PR #4498, opened 2020-04-22
fix: add aria-selected to selected tree node link

Shijir:

## PR Checklist

## PR Type

The PR adds `aria-selected` attribute and sets its value depending on whether the tree node-link belongs to the currently activated page or not. There are two types of node links: one that depends on `RouterLinkActive` and the other one doesn't. If it depends on `RouterLinkActive`, tree-node-link automatically handles `aria-selected`. If it doesn't, a user needs to manually set `aria-selected` which is reflected in the document.

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes

Issue Number: https://github.com/vmware/clarity/issues/3571

* [ ] Yes
* [x] No

### DIFF
--- a/src/clr-angular/data/tree-view/tree-node-link.ts
+++ b/src/clr-angular/data/tree-view/tree-node-link.ts
@@ -1,16 +1,18 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Directive, ElementRef } from '@angular/core';
+import { Directive, ElementRef, Optional } from '@angular/core';
+import { RouterLinkActive } from '@angular/router';
 
 @Directive({
   selector: '.clr-treenode-link',
+  host: { '[attr.aria-selected]': 'routerLinkActive?.isActive' },
 })
 export class ClrTreeNodeLink {
-  constructor(private el: ElementRef) {}
+  constructor(private el: ElementRef, @Optional() public routerLinkActive: RouterLinkActive) {}
 
   activate() {
     if (this.el.nativeElement && this.el.nativeElement.click) {
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -1696,7 +1696,8 @@ export declare class ClrTreeNode<T> implements OnInit, OnDestroy {
 }
 
 export declare class ClrTreeNodeLink {
-    constructor(el: ElementRef);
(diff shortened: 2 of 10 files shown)


src/clr-angular/data/tree-view/tree-node-link.ts
```diff
@@ -1,16 +1,18 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Directive, ElementRef } from '@angular/core';
+import { Directive, ElementRef, Optional } from '@angular/core';
+import { RouterLinkActive } from '@angular/router';
```
gnomeontherun, line 8: I've resisted this before, because it now puts a dependency on the router. Why don't we just have the apps put `aria-selected` themselves so we aren't tying to the router? Some folks use a different router.

Shijir: That's why I used `@Optional()` for the `RouterLinkActive` injection. If it's present, we will manage it. If it's not present, we won't be involved. Also, if the user is using `RouterLinkActive`, it's better for them not to be involved in setting the value of `aria-selected` as it should be always in sync with `RouterLinkActive`'s current state. If they are not relying on `RouterLinkActive`, by all means, they can manage the `aria-selected` on their side. And this behavior is reflected in the website example.

gnomeontherun: Using `@Optional` only works for injection, not for importing. That is why this will fail if you use Tree View without first installing `@angular/router`.

Shijir: oh shoot you are absolutely right! It will make even simple tree nodes dependent on angular router module. I will fix it. Thank you!

gnomeontherun: If we can bump to TS 3.8, it might be possible to use `import()` https://github.com/tc39/proposal-dynamic-import https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#-es2020-for-target-and-module

gnomeontherun: So Angular 10 requires TS 3.8, so we could use that but then we can't backport it.

Shijir: Yeah I think it's too early to adopt dynamic imports. I will just offer doc guidelines as a solution for now.

Shijir:
Closing this PR as we are planning to offer a different solution to the issue.

PR #4555, opened 2020-05-11
feat(a11y): expandable datagrid-row (backport to v3)

bdryanovski:
Backport #4163 to v3

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-row.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-row.spec.ts
@@ -316,6 +316,11 @@ export default function(): void {
         expect(context.clarityElement.querySelector('.spinner')).not.toBeNull();
       });
 
+      it('button must contain aria-controls', function() {
+        const button = context.clarityElement.querySelector('.datagrid-expandable-caret-button');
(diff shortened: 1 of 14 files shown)


src/clr-angular/data/datagrid/datagrid-row.spec.ts
```diff
@@ -3,518 +3,62 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
```
gnomeontherun, line 4: What happened here?

src/clr-angular/data/datagrid/datagrid-row.ts
```diff
@@ -199,9 +203,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
   ngAfterContentInit() {
     this.dgCells.changes.subscribe(() => {
       this.dgCells.forEach(cell => {
-        if (!cell._view.destroyed) {
```
Shijir, line 202: You are reverting changes; it needs cleaner backport.

bdryanovski:
@Shijir @gnomeontherun thanks for catching it .. I didn't notice that my cherry-pick had done that 😭

PR #4601, opened 2020-05-13
fix(form): export Layouts enum for correct input type

twittwer:
Export the Layouts enum to enable consumers with strict template checking to pass the correct input type.

Signed-off-by: Tobias Wittwer <t.wittwer95@gmail.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently I'm not able to import the `Layouts` enum and when using plain strings like `clrLayout="vertical"` the strict template checker complains about it, because the string is not assignable to the enum.

## What is the new behavior?

In my current proposal the `Layouts` enum gets exported. So it possible to pass the real enum value as input `[clrLayout]="Layouts.VERTICAL"`.

## Alternative

An alternative would be to use the [input setter coercion](https://angular.io/guide/template-typecheck#input-setter-coercion). Therefore, I would add a static property to the `ClrLayout` directive:  
`public static ngAcceptInputType_layout: Layouts | 'vertical' | 'horizontal' | 'compact';`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -2050,6 +2050,12 @@ export declare const IS_TOGGLE_PROVIDER: {
 
 export declare function isToggleFactory(): BehaviorSubject<boolean>;
 
+export declare enum Layouts {
+    VERTICAL = "vertical",
+    HORIZONTAL = "horizontal",
+    COMPACT = "compact"
+}
+
 export declare abstract class LoadingListener {
     abstract loadingStateChange(state: ClrLoadingState): void;
 }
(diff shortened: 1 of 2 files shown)


gnomeontherun:
The `Layouts` enum is internal, and we only want to make items public if they are prefixed with `Clr`. 

If we need to make it public, then we should rename `Layouts` to `ClrFormsLayout`.

bdryanovski:
Poke @twittwer

twittwer:
@gnomeontherun Should the enum itself be renamed or the export in the index.ts?

gnomeontherun:
I would prefer renaming the enum everywhere instead of reexporting it as a new name.

bdryanovski:
Poke @twittwer

mathisscott:
I'm assuming this is not making it into 4.0

gnomeontherun:
Closed by #4994

PR #4607, opened 2020-05-13
feat: adding smaller values to core layout spacing

mathisscott:
• added "2px" spacer
• added "4px" spacer
• these are needed for sizing layouts in compact versions of our components
• shifted all existing values up a size
• sizes now go from xxs to xxl


coryrylan:
Theres a couple places within the component templates that will need to be updated ex [example modal](https://github.com/vmware/clarity/blob/master/packages/core/src/modal/modal.element.ts#L72) and the templates in storybook probably will have to be adjusted for the new sizes as well.

PR #4680, opened 2020-05-23
style: switch explicit-function-return-type TS rule to error for product code of core

amalitsky:
Signed-off-by: Alex Malitsky <1457274+amalitsky@users.noreply.github.com>

## PR Type

- [X] Code style update (formatting, local variables)

Currently `@typescript-eslint/explicit-function-return-type` rule is set to _off_ for all files inside _packages/core_. That means that providing function/method return value type is **not required**.

Issue Number: No issue at this point, please let me know if I need to create one.

## What is the new behavior?
`@typescript-eslint/explicit-function-return-type` rule is set to _error_ for all files besides _.stories.ts_ and _spec.ts_. All linting errors caused by enabling the rule were fixed as well.

Once this PR is merged providing type for function/method return value will be **required**.

I will look into how many occurrences need to be fixed for _spec_ and _story_ files and might update this PR with more commits or raise another PR. 

- [ ] Yes
- [X] No

## References
- rule description: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -2,14 +2,14 @@ const OFF = 0;
 const WARN = 1;
 const ERROR = 2;
 
-// some temporary relaxation for spec and story files
+// temporary relaxations for story files
 const overrideRules = {
-  '@typescript-eslint/explicit-function-return-type': 0, // TOO MUCH WORK AT THE MOMENT ;)
+  '@typescript-eslint/explicit-function-return-type': 0,
 };
 
 const overrides = [
   {
-    files: ['*.stories.ts', '*.spec.ts'],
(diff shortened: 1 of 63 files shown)


coryrylan:
Thank you for your contribution. We are currently moving a lot of code around in preparation for v4 and project mono repo/tooling re-work, so we will have to put this PR on hold until things settle down a bit. If you come across any issues/cleanup improvements, feel free to open an issue about it first, so make sure we coordinate and collaborate on it.

amalitsky:
Git it, Cory, no worries. Is it smth worth rebasing once v4 is released?

Issue #48, opened 2016-11-14
Active .nav-item box-shadow does not overlap with the box-shadow of .nav


adityarb88:
**Select one ...**  (check one with "x")
```
[ x] bug
[ ] feature request
```

### Expected behavior

The box-shadow of `.nav` should overlap with the box-shadow of an active `.nav-item`

![image](https://cloud.githubusercontent.com/assets/1426805/20280726/7d0bdae8-aa62-11e6-9832-f34426821046.png)

### Actual behavior

The box-shadow of `.nav` is not overlapping with the box-shadow of an active `.nav-item`

![image](https://cloud.githubusercontent.com/assets/1426805/20280752/94245ee4-aa62-11e6-804e-8aeb2f57e4fd.png)

### Environment details

* **Angular version:** 2.0.X

* **Clarity version:** 0.6.2 and above

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

PR #4687, opened 2020-05-28
refactor(a11y): deprecating ClrAriaLiveService from internal use

### LINKED ISSUE #4222: AriaLiveService is not announcing errors inside clrInputContainer
## Describe the bug

`clr-error` when into the context of `clr-input-container` is rendered and then hidden. This is preventing `AriaLiveService` of updating and announcing the message inside the error. Also preventing the native angular lifecycles of firing.

## How to reproduce

```html        
<clr-input-container>
  <label>Field 1 label</label>
  <input clrInput type="text" [(ngModel)]="fieldLabel" name="example" required />
  <clr-control-helper>Helper text that shows while it is pristine and valid</clr-control-helper>
  <clr-control-error>Error message that appears after focus is lost and control is invalid</clr-control-error>
</clr-input-container>
```

* `clr-control-error` will be rendered and `aria-live` container will be created
* `clr-control-error` is now visible
*  no destroy livecycle is called for `clr-control-error` so the `aria-live` container is still inside the DOM
*  when `clr-control-error` is visible -  no update to the `aria-live` is made so no new annoncement from `AriaLiveService`

The only way to remove the `aria-live` DOM element is to detach `clr-input-container` from the DOM.

Content Projection is rendering the components before moving them inside the `clr-input-container` template. This is making some of the lifecycles behave in a strange way. For example `onDestroy` is not called when a component is hidden/shown with `*ngIf`. Angular is keeping the components alive until the parent is destroyed and reusing the same instance. 

## Expected behavior

Error content must be read from SR when it's rendered on the screen. As many times as the error happens. 

## Versions

All versions are affected.

## Additional notes

A solution could be the use of `ngTemplateOutlet` but this will result in a breaking change. Another option could be providing a service that will provide both the container and the error with a message when the error is visible and when not, to force `AriaLiveService` to announce the message or remove it.

### LINKED ISSUE #4671: Remove internal use of aria-live
## Summary

After deliberation, it was decided that Clarity is not able to do `aria-live` announcements in a predictable and consistent way that works for all use cases.

## Task: Remove aria live

* [ ] Alert
* [ ] Form Error
* [ ] Date picker (need to confirm)
* [ ] Progress
* [ ] Spinner

## Task: Update documentation to help identify common places applications should announce.

bdryanovski:
#4671

Drop ClrAriaLiveService from internal component and let the developer decide how and when they want to announce to the user for changes on the page. 

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4671

- [ ] Yes
- [x] No

## Other information
Close #4671 
Close #4222


packages/angular/projects/clr-angular/src/progress/progress-bars/progress-bar.ts
```diff
@@ -135,11 +128,12 @@ export class ClrProgressBar {
...
 
-  // Aria Live
-  /** @deprecated since 3.0, remove in 4.0 */
-  @Input('clrAssertive') assertive: boolean;
-  /** @deprecated since 3.0, remove in 4.0 */
-  @Input('clrOff') off: boolean;
+  /**
+   * Display aria-live only when there is value and it's not 0 or equal to the max value
+   */
+  displayStringValue() {
```
hippee-lee, line 134: This is for the optional value text after the progress bar, right? Why would it be hidden if it was 0 or max?

packages/angular/projects/clr-angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -61,13 +61,8 @@ export const commonStringsDefault: ClrCommonStrings = {
   datepickerCurrentDecade: 'Current decade',
```
hippee-lee, line 61: Do we also need to remove the datagrid aria strings (line #48 - #51)?

packages/angular/projects/clr-angular/src/progress/progress-bars/progress-bar.ts
bdryanovski: Yes is for showing the optional value after the progress bar .. but I don't have an idea why it is written this way?! 

I'm trying to figure it out. But it seems that the check is "pointless". I think that there is a need for the `ngIf` but the check has to be something else. 

```ts
// show value on the right only when the class label is set to true
displayStringValue() { return this._labeled }
```

bdryanovski: It seems that this was something that no one notice .. thanks for catching it

packages/angular/projects/clr-angular/src/utils/i18n/common-strings.default.ts
bdryanovski: If you are talking about this lines 

```ts
  singleSelectionAriaLabel: 'Single selection header',
  singleActionableAriaLabel: 'Single actionable header',
  detailExpandableAriaLabel: 'Toggle more row content',
  datagridFilterAriaLabel: 'Toggle column filter',
```
I think that not, there used inside the `datagrid.html` as aria-labels for some of the header buttons also has there own `@Input` for overwritting them if needed.

Issue #4484, opened 2020-04-17
Remove styling from subcomponents in core alert


jeeyun:
## Describe the bug

Core components leave it up to the application to define and control the layout of sub-components inside a component. Examples of this include modal and alerts. This ticket is to remove scss files for `cds-alert-actions` component that is used inside `cds-alert`.

Coding guidelines for core: https://github.com/vmware/clarity/blob/master/docs/CODING_GUIDELINES_CORE.md#web-component-conventions

Issue #4460, opened 2020-04-09
Alert text not vertically centered in the banner


YihongS:
## Describe the bug

The text within a standard alert banner is not vertically centered, which makes it looks weird when there's only one line in the alert.
![image (3)](https://user-images.githubusercontent.com/32996499/78930952-d4daab80-7a59-11ea-8cc8-fe04d1ef7863.png)
The small alert banner doesn't have this problem.

## How to reproduce

Just go to Clarity documentation page: https://v2.clarity.design/alerts

## Expected behavior

The text should be vertically-centered in the alert banner.

## Versions

**Device:**

* Type: MacBook Pro
* OS: macOS
* Browser: Safari
* Version: 13.1

bdryanovski:
It seems that this issue is contained only on the website.

```css
.alert-items {
  flex: 1 1 auto;
  flex-flow: column nowrap;
  padding: .4rem .55rem;
  display: flex;
}
```
The padding must be `padding: .5rem .55rem;`

sis0k0:
Reopening as the Angular sample still isn't centered - https://clarity.design/angular-components/alert/#size-2.

Issue #4405, opened 2020-03-25
core: update component internals to use tokens/layouts


coryrylan:
Update core components to use token system and layouts once merged.

- [ ] alert https://github.com/vmware/clarity/issues/4343 
- [ ] app alert
- [ ] badge
- [ ] button
- [ ] tag

Issue #4343, opened 2020-03-05
core: refactor alert private host


coryrylan:
Refactor the alerts components to use the private host DOM structure to ensure proper CSS encapsulation. See cds-button or cds-tag for examples.

mathisscott:
Should we tackle this when we refactor for layout?

PR #4707, opened 2020-06-05
fix: core - refactoring alerts

mathisscott:
• updated alert-actions to display link-style buttons in light and default alerts per current design
• added min-width and text-decoration custom properties to buttons for internal use inside of alerts
• added close button internal component
• added inline buttons
• hiding alert-actions inside a lightweight alert
• moving icon and badge up to the base button because other button types needed them
• added styles and code to support links in icon buttons
• added styles and code to support links in inline buttons
• added stories for links in inline buttons
• added stories for links in icon buttons
• added loading status to alerts
• added compact alerts (specifically compact lightweight alerts)
• getting close-button internal component to 100% test coverage
• boosting branch coverage to over 90%
• improving css-vars test coverage
• added alert group element
• moved alert group types and alert status type interfaces to internal
• added lightweight alert group
• added compact alert groups
• verified close change event emits from alerts alert group
• verified that alert-group status does not override custom icon shapes in alerts
• verified that alerts with a loading status are not overridden by the alert-group status
• created css utility functions
• refactored the CssHelpers mixin to use the css utility functions
• made it so that only alerts care about their children
  • alert-groups don't know anything about alert-actions
• removing old alert-content component
• removing old alert scss file
• added aria and unit tests for a11y
• fixed issue with sr-only elements inside of cds-icon that was revealing them in Safari
• setting up app-level alert groups
• removing aria-live declarations from alerts
• refactoring alerts to slot cds-icons as custom icons instead of using icon-shape attr
• removing old app-alert files
• refactoring app-level alert group type to banner
• styles for banner alerts loading status
• limiting compact sizing to just default alerts
• adding compact sizing to banner alerts is not possible...
• ...because we do not have outline buttons that are smaller than 24px height
• adding verification for app-alert pager
• adding code that allows for banners to be center aligned without a pager and left aligned with a pager
• adding banner alert statuses
• implementing functionality that forces banner alert groups to update when their child alert's status changes
• removing old app-alert styles
• increased branch code coverage threshold to 90%
• added tests for new layouts introduced by banner alert group pager
• added `assignSlotNames` helper to centralize work of adding slot attrs in connectedCallback
• fixed issue where light alert groups were not getting status color on their icons as expected
• story on alert page said "link button" instead of inline button
• fixed Safari width issue on alerts
• CSS hacks for alert actions across shadowDom in Safari

## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: #4484 #4343 #4405 #4460 #3791 

- [x] Yes
- [ ] No

### Default  Alerts

Screen Shot 2020-06-04 at 4 11 58 PM

### Lightweight  Alerts

Screen Shot 2020-06-04 at 4 12 12 PM

### Compact Alerts

Screen Shot 2020-06-04 at 4 12 28 PM

### Banner Alerts with No Pager

Screen Shot 2020-06-04 at 4 12 45 PM

### Banner Alerts with Pager Placeholder

Screen Shot 2020-06-04 at 4 13 09 PM


packages/core/src/alert/alert-actions.element.scss
```diff
@@ -1,8 +1,66 @@
...
+}
+
+::slotted(cds-button.alert-btn:hover) {
+  --color: var(--action-hover-text-color, #{$cds-token-color-neutral-1000-static});
+  --border-color: var(--action-hover-text-color, #{$cds-token-color-neutral-1000-static});
+}
+
+:host([type='default']) {
+  --action-size: calc(#{$cds-token-space-size-9} - #{$cds-token-space-size-3});
+  padding-top: #{$cds-token-space-size-2};
```
jeeyun, line 41: should the `padding-top` be defined by its container rather than in the component?

mathisscott: The container is outside of the slot while the component here is projected into the slot for alert actions.

When I put padding on the container, it was like it then ignored the layouts in the slot (pushing everything down much farther than we wanted). It's also worth noting that this particular fix was related to issues I was seeing in Safari. It was a late entry because things looked fine everywhere else but Safari was not handling padding on the container and then negative margin on a layout projected inside the slot gracefully.

Safari was like "oh.. you must not want any negative margins then!" on the slot content. 😬

mathisscott: For future reference... we should check how things are laying out in Safari. It's no longer a to-be-trusted platform. Things honestly look great and consistent across firefox, (new) edge, and chrome. Safari is quickly becoming a problem...

jeeyun: Thank you for the explanation. It's too bad that Safari has become the new "outlier browser."  😞  I'll be looking out for that in my work too, to be extra vigilant in checking that it looks as expected in Safari.

packages/core/src/icon-shapes/icon.element.scss
```diff
@@ -22,13 +22,13 @@
 }
 
 :host {
-  --color: #{$cds-token-color-neutral-700};
-  --badge-color: #{$cds-token-color-danger-700};
+  // --color: var(--color, #{$cds-token-color-neutral-700});
+  // --badge-color: var(--badge-color, #{$cds-token-color-danger-700});
```
jeeyun, line 26: clean up: remove?

packages/core/src/internal/base/button.base.ts
```diff
@@ -80,13 +83,6 @@ export class CdsBaseButton extends LitElement {
     }
   }
 
-  private setupAnchorFocus() {
-    if (this.anchor) {
-      this.anchor.addEventListener('focus', () => this.setAttribute('focused', ''));
-      this.anchor.addEventListener('blur', () => this.removeAttribute('focused'));
-    }
-  }
-
```
coryrylan, line 89: Not sure but I don't think this was supposed to be removed? That is what allows the button to apply the focus styles when the user tabs to a button that is a link and I think some other use cases as well. On prod it looks broken when focusing not sure about locally if it has been fixed https://clarity.design/storybook/core/?path=/story/components-button-stories--links

It is also what is breaking the commented test here https://github.com/vmware/clarity/pull/4707/files#diff-84e6a67e36b8a8cef08d1aba22c89482R243

packages/core/src/internal/interfaces/index.ts
```diff
@@ -16,6 +16,10 @@ export type StoplightStatusTypes = 'success' | 'warning' | 'danger';
 
 export type StatusTypes = 'info' | StoplightStatusTypes;
 
+export type AlertStatusTypes = 'info' | 'success' | 'warning' | 'danger' | 'unknown' | 'loading';
+
+export type AlertGroupTypes = 'default' | 'banner' | 'light';
```
coryrylan, line 21: nit: component specific interfaces and code should probably stay within its directory

packages/core/src/alert/alert-group.element.scss
```diff
@@ -0,0 +1,157 @@
...
+:host([type='banner']) .private-host {
+  --background: #{$cds-token-color-action-600};
+  --border-color: #{$cds-token-color-action-600};
+  --border-radius: 0;
+}
+
+:host([type='banner'][status='warning']) .private-host {
+  // color carried over from clr-angular
+  --background: hsl(26, 100%, 38%);
+  --border-color: hsl(26, 100%, 38%);
```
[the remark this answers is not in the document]
mathisscott: i avoided doing that because the colors are so different. additionally, `warning-700` fails a11y with white text while the other color does not.

i didn't feel comfortable with a change of that magnitude (deep burnt orange to muddy yellow and white text to black text). so i opted to leave it.

packages/core/src/alert/alert.element.ts
```diff
@@ -4,27 +4,182 @@
...
+    case 'danger':
+      return [ClarityIcons.getIconNameFromShape(errorStandardIcon), commonstrings.danger];
+    case 'unknown':
+      return [ClarityIcons.getIconNameFromShape(helpIcon), commonstrings.info];
+    case 'loading':
+      return ['loading', commonstrings.loading];
+    default:
+      return [ClarityIcons.getIconNameFromShape(infoStandardIcon), commonstrings.info];
+  }
+}
```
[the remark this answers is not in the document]
mathisscott: sounds good. i've been having a little too much fun with switch statements lately.

packages/core/src/alert/alert.element.ts
```diff
@@ -4,27 +4,182 @@
...
+ *     <cds-alert-actions>
+ *       buttons, links
+ *     </cds-alert-actions>
+ *   </cds-alert>
  * ```
  *
  * @beta
  * @element cds-alert
- * @slot default - Content slot for inside the alert. Usually will contain at least a <cds-alert-content> component.
+ * @slot default - Content slot for inside the alert !!! TODO... OTHER SLOTS?
```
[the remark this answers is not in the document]
mathisscott: good catch. it will be removed.

packages/core/src/alert/alert.element.ts
```diff
@@ -33,16 +188,171 @@ import { styles } from './alert.element.css.js';
...
+          ? html`<span class="alert-spacer" cds-layout="align:stretch">&nbsp;</span>`
+          : html``}
+        ${this.alertGroupType !== 'light' && this.closable
+          ? html`<span class="alert-close-wrapper"
+              ><cds-internal-close-button
+                @click="${() => this.closeAlert()}"
+                icon-shape="${this.closeIconShape}"
+                aria-label="${this.closeIconTitle}"
+                icon-size="${getCloseIconSize(this.alertGroupType, this.closeIconSize)}"
+              ></cds-internal-close-button
```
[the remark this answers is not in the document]
mathisscott: good call. i can do that.

packages/core/src/button/button.element.spec.ts
```diff
@@ -183,14 +199,14 @@ describe('button element', () => {
...
-    let anchor: HTMLAnchorElement;
+    // let anchor: HTMLAnchorElement;
 
     beforeEach(async () => {
       testLinkElement = createTestElement();
       testLinkElement.innerHTML = `<cds-button><a href="about">About</a></cds-button>`;
       await waitForComponent('cds-button');
       componentLink = testLinkElement.querySelector<CdsButton>('cds-button');
-      anchor = testLinkElement.querySelector<HTMLAnchorElement>('a');
+      // anchor = testLinkElement.querySelector<HTMLAnchorElement>('a');
```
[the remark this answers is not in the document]
mathisscott: i think so. i'll verify though.

packages/core/src/internal/base/button.base.ts
mathisscott: i don't recall removing it. so it wasn't something that happened recently. but i'll look into that.

packages/core/src/internal/interfaces/index.ts
mathisscott: they were put here early on because we initially thought we'd have alerts and app-alerts as separate components. they were intended to be shared. i'll relocate them.

packages/core/src/alert/alert-group.element.scss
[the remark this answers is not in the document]
colinreedmiller: Agreed - this is a problematic spot with regard to a11y vs aesthetics and should be addressed by a larger group and with a more system wide analysis.

packages/core/src/internal/base/button.base.ts
coryrylan: I dug into this it looks like the broken styles on the prod docs are were fixed in this commit on master https://github.com/vmware/clarity/commit/458b8aaa3592367ad137b328fbe0e290077a63a7

PR #4740, opened 2020-06-18
WIP feat: integrate Chromatic

bdryanovski:
Work in progress don't merge until fully tested

NOTE to self - change commit message to be "build: ..." when rebasing

#### Reason for not using `chromaui/action@v1`

So below is the version using the `chromaui/action@v1` Github action plugin.
```yml
jobs:
  chromatic:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - uses: actions/setup-node@v1
        with:
          node-version: '12.x'
      - run: yarn
      - run: yarn build:ci
      - uses: chromaui/action@v1
        with:
          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
          storybookBuildDir: ./packages/core/dist/storybook/core
    env:
      CI: true
``` 
This will throw an error that Chromatic could not find storybook version. This is because the Github action plugin (GAP) is running from the root of the repository and there is no way to change the working directory. 

A workaround that I found is to trick the GAP that there is storybook by adding a dependency to the root package.json in our case this is:

```json
"@storybook/web-components": "6.0.0-alpha.30",
```

Deep inside the plugin there is this executable `get-info.js` 
```js
// node_modules/chromatic/bin/storybook/get-info.js

const viewLayers = [
  'react',
  'vue',
  'angular',
  'html',
  'web-components',
  'polymer',
  'ember',
  'marko',
  'mithril',
  'riot',
  'svelte',
  'preact',
  'rax',
];
```
In this case, `viewLayers` is the list of storybook packages that are been checked. 

So to summarize if we want to use the plugin we have to keep the `root/package.json` and `packages/core/package.json` dependency in sync. Not sure is the plugin depending on the version or just checking to see if it exist but still, this is something that we have to track by hand.

I did not find also any benefit of running the plugin or the direct command.

##### Loading icon trigger change.
On the components that have loading animation (spinner) Chromatic is tracking it as change because it's not manging the take the screenshot at the same time. Sometimes tests that includes this type of element pass without any error. 

The quick and dirty solution to that is adding `class="chromatic-ignore"` to the DOM elements that are dynamic. This is not an optimal solution because is introducing unwanted classes to components.

From the [Chromatic Documentation](https://www.chromatic.com/docs/animations) there is a way to pause the animation and then take a screenshot but it had no effect and still, the test failed 
```json
chromatic: { pauseAnimationAtEnd: true }
```

## PR Type

- [x] Build related changes
- [x] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,9 @@ jobs:
           node-version: '12.x'
       - run: yarn
       - run: yarn build:ci
+      - name: Deploy Core Storybook to Chromatic
+        working-directory: ./packages/core
+        run: npx chromatic --storybook-build-dir=./dist/storybook/core --exit-zero-on-changes --ci
 
     env:
       CI: true
--- a/packages/core/src/alert/alert-group.stories.ts
+++ b/packages/core/src/alert/alert-group.stories.ts
@@ -227,7 +227,7 @@ export const bannerGroupStatus = () => {
       </cds-alert-group>
 
       <cds-alert-group type="banner">
-        <cds-alert status="loading">
+        <cds-alert class="chromatic-ignore" status="loading">
           This example is an alert with alert actions and a status of "loading" inside a banner alert group.
           <cds-alert-actions>
             <cds-button>Button 1</cds-button>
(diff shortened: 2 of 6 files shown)


bdryanovski:
Chromatic why are you the way you are!

gnomeontherun:
@bdryanovski Are we going to get a review for this soon? I'd like to either move forward to close this out.

bdryanovski:
I gonna have few meetings this week to talk about it .. after that will update the PR.

bdryanovski:
I'm closing this PR for now.

PR #4774, opened 2020-07-09
fix(toggle): popover toggle prevents default behavior of arrow key events

bbogdanov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3622

## What is the new behavior?

When the container is too large and an arrow key initiates page scrolling, but at the same time initiates a popover opening, the popover doesn't close and stays open.

- [ ] Yes
- [x] No

## Other information

It is somehow tricky to create a unit test for this because it is part of the popovers and they are represented as directives and services, and at the same time, it requires a parent with a template that has a scroll and a popover child component in it. It is not a good idea to add the unit test in the popover but at the same time it is not a good idea to set a single unit test for example only in dropdown cause this can be the same issue in the tooltips.

### DIFF
--- a/packages/angular/projects/clr-angular/src/utils/popover/providers/popover-toggle.service.spec.ts
+++ b/packages/angular/projects/clr-angular/src/utils/popover/providers/popover-toggle.service.spec.ts
@@ -11,6 +11,7 @@ import { Observable } from 'rxjs';
 import { ClrPopoverEventsService } from './popover-events.service';
 import { ClrPopoverPositionService } from './popover-position.service';
 import { ClrPopoverToggleService } from './popover-toggle.service';
+import { KeyCodes } from '../../enums/key-codes.enum';
 
 @Component({
   selector: 'test-host',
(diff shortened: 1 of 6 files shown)


packages/angular/projects/clr-angular/src/utils/popover/providers/popover-toggle.service.ts
```diff
@@ -57,6 +58,8 @@ export class ClrPopoverToggleService {
    * This is for instance the case of components that open on a click, but close on a click outside.
    */
   public toggleWithEvent(event: any) {
+    preventArrowKeyScroll(event);
```
hippee-lee, line 61: This looks like the correct solution to the issue that was reported. 

For a unit test I would like to see a spec added to https://github.com/vmware/clarity/blob/master/packages/angular/projects/clr-angular/src/utils/popover/popover-close-button.spec.ts In the `Template API` suite. 

Something like 
```it('does not close the popover when called with an arrow key event', function (this: Context) { ...})```

Just create the specific arrow keypresses that the util function checks for and assert that the popover stays open. 

An example of creating a specific event for a spec and then testing something with it is here: https://github.com/vmware/clarity/blob/master/packages/angular/projects/clr-angular/src/utils/popover/providers/popover-events.service.spec.ts#L83

File: packages/angular/projects/clr-angular/src/utils/popover/providers/popover-toggle.service.spec.ts [hunk 2]
```diff
@@ -94,6 +95,30 @@ export default function (): void {
         expect(this.toggleService.open).toBeFalse();
         expect(this.toggleService.openEvent).toBe(closeClickEvent);
       });
+
+      Object.keys(KeyCodes)
+        .filter(key => key.search('Arrow') > -1)
```
bbogdanov, line 100: @hippee-lee @coryrylan I am not really satisfied with getting all the arrow keys based on a hardcoded string in a filter,  but I couldn't think of a better way at this point. If you have suggestions I am open to improving the filter.

hippee-lee: It works, I'm ok with the approach. 

What do you think about combining into one pass instead of two? e.g 
```typescript
Object.keys(KeyCodes)
  .forEach(key => {
    if(key.search('Arrow') > -1) {
      // Arrow keys are ignored to prevent content from closing the popover content
      it(`prevents the default toggle action for the ${key} key`, function (this: TestContext) {
        const arrowKeyEvent = new KeyboardEvent(key, { key });
        spyOn(arrowKeyEvent, 'preventDefault');
        this.toggleService.toggleWithEvent(arrowKeyEvent);
        expect(arrowKeyEvent.preventDefault).toHaveBeenCalled();
      }
    } else if(key.search('Arrow') === -1) {
      it(`does not prevent the default toggle action for the ${key} key`, function (this: TestContext) {
        const arrowKeyEvent = new KeyboardEvent(key, { key });
        spyOn(arrowKeyEvent, 'preventDefault');
        this.toggleService.toggleWithEvent(arrowKeyEvent);
        expect(arrowKeyEvent.preventDefault).not.toHaveBeenCalled();
      }
    }
  });
```

bbogdanov: Done.

Issue #4528, opened 2020-05-01
Datagrid selection: async clrDgSelected with immutable array not working


Ergnos:
## Describe the bug

When an immutable object is used (in example a ngrx/store state element) the selection breaks when 'item' is pushed to this.current:
```
app.error-handler.ts:19 TypeError: Cannot add property 0, object is not extensible
    at Array.push (<anonymous>)
    at Selection.selectItem (clr-angular.js:11544)
    at Selection.setSelected (clr-angular.js:11581)
    at ClrDatagridRow.set selected [as selected] (clr-angular.js:11815)
    at ClrDatagridRow.toggle (clr-angular.js:11827)
    at ClrDatagridRow_ng_template_5_div_4_Template_input_ngModelChange_1_listener (clr-angular.js:538)
    at executeListenerWithErrorHandling (core.js:21772)
    at wrapListenerIn_markDirtyAndPreventDefault (core.js:21814)
    at SafeSubscriber.schedulerFn [as _next] (core.js:37027)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:183)
```

## How to reproduce

1.  See stackblitz https://stackblitz.com/edit/clarity-light-theme-v2-cpfinp
2.  Select an item
3.  See console for error message

## Expected behavior
Selection should work with async selection from immutable array.

Solution could be to replace current code:
```
private selectItem(item: T): void {
    this.current.push(item);
    if (this._items.trackBy) {
      // Push selected ref onto array
      const lookup = this._items.all.findIndex(maybe => maybe === item);
      this.prevSelectionRefs.push(this._items.trackBy(lookup, item));
    }
```
to: 
```
private selectItem(item: T): void {
    this.current.concat(item);
    if (this._items.trackBy) {
      // Push selected ref onto array
      const lookup = this._items.all.findIndex(maybe => maybe === item);
      this.prevSelectionRefs.push(this._items.trackBy(lookup, item));
    }
```

Additionally set current(value) can be changed to since otherwise the change event is emitted twice.

```
    set current(value) {
        this.updateCurrent(value, false);
    }
```

## Versions

**App**

* Angular: [8.2.14 & 9.1.1]
* Node: [13.8.0]
* Clarity: [2.4.1 & 3.1.1 (tested with modified clr-angular.js due to issue #4504) ]
* Ngrx/store: [9.1.0]

**Device:**

* Type: [Laptop]
* OS: [Windows 10]
* Browser [Chrome]
[...]

PR #4781, opened 2020-07-13
fix(datagrid): current selection immutability

skipgeldens:
The current selection is now immutable, so that selection can directly be used with @ngrx/store strict checking.

Signed-off-by: Skip Geldens <skip@skipgeldens.nl>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently I cannot use [clrDgSelected] directly from a ngrx/store async variable because the datagrids mutates the content of the array of selected items instead of replacing the complete array. Shown in: https://stackblitz.com/edit/clarity-light-theme-v2-cpfinp

Issue Number: https://github.com/vmware/clarity/issues/4528

## What is the new behavior?

The `_current` array in the selection provider is not mutated but replaced on change.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts
@@ -3,8 +3,8 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { TrackByFunction } from '@angular/core';
-import { fakeAsync, tick } from '@angular/core/testing';
+import { TrackByFunction, NgZone } from '@angular/core';
+import { fakeAsync, tick, TestBed } from '@angular/core/testing';
 import { Subject } from 'rxjs';
 
 import { ClrDatagridFilterInterface } from '../interfaces/filter.interface';
@@ -38,8 +38,8 @@ export default function (): void {
         itemsInstance = new Items(filtersInstance, sortInstance, pageInstance);
         itemsInstance.smartenUp();
         itemsInstance.all = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
-
-        selectionInstance = new Selection(itemsInstance, filtersInstance);
+        const ngZone = TestBed.get(NgZone);
+        selectionInstance = new Selection(itemsInstance, filtersInstance, ngZone);
       });
 
       afterEach(function () {
@@ -193,7 +193,7 @@ export default function (): void {
         expect(currentSelection.sort(numberSort)).toEqual(itemsInstance.displayed);
         selectionInstance.toggleAll();
         expect(currentSelection).toEqual([]);
-        expect(nbChanges).toBe(3);
+        expect(nbChanges).toBe(4);
       });
 
       it('exposes an Observable to follow selection changes in single selection type', function () {
@@ -347,6 +347,27 @@ export default function (): void {
 
         expect(selectionInstance.current).toEqual([4, 2]);
       });
+
+      it('does not mutate the selection array when selecting or deselecting', function () {
+        selectionInstance.selectionType = SelectionType.Multi;
+        const selection = [4, 2];
+        selectionInstance.current = selection;
+
+        selectionInstance.setSelected(1, true);
+        expect(selectionInstance.current === selection).toBe(false);
+
+        selectionInstance.setSelected(1, false);
+        expect(selectionInstance.current === selection).toBe(false);
+      });
+
+      it('does not mutate the selection array when clearing the selection', function () {
+        selectionInstance.selectionType = SelectionType.Multi;
+        const selection = [4, 2];
+        selectionInstance.current = selection;
+
+        selectionInstance.clearSelection();
+        expect(selectionInstance.current === selection).toBe(false);
+      });
     });
 
     describe('clrDgPreserveSelection', () => {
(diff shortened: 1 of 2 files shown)


mathisscott:
This change broke a test in one of the datagrid services. That is not surprising because the test is looking for mutability. I'm pinging @hippee-lee and @gnomeontherun on this one. It will be a few days before @gnomeontherun can review this, though, because he is currently on PTO.

packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.ts
```diff
@@ -237,8 +238,10 @@ export class Selection<T = any> {
 
   public updateCurrent(value: T[], emit: boolean) {
     this._current = value;
-    if (emit) {
+    if (emit && !this.debounce) {
       this.emitChange();
+      this.debounce = true;
+      setTimeout(() => (this.debounce = false));
```
gnomeontherun, line 244: This pattern raises a flag for me, what exactly is happening that requires this? I can't tell from the code review what this solves.

skipgeldens: I added this because otherwise the change was emitted with every item that was added to the selection (within the same change detection cycle). This made the following test fail: https://github.com/vmware/clarity/blob/9c5523481c1ed574a3b6d57de99df2dc6d8637a2/packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts#L182 

I saw that this debounce was also in previous implementations of Selection.

Jinnie: We're usually unwilling to use setTimeout, unless unavoidable, as it is patched from the Angular framework and forces a change detection cycle on every call.
The possible alternatives we've been using are either use an observable with debounceTime(0) (as in popover-content.ts), or call the setTimeout in an ngZone's runOutsideAngular, as done in datagrid-action-overflow.ts
The first may be more suitable in this case, as it's also about debouncing "nearly simultaneous" events, while the second we usually use with focus management, which really isn't covered well in Angular.

skipgeldens: I tried implementing debounceTime(0), but that caused another test to fail: https://github.com/vmware/clarity/blob/ebb6f552485fe82d1f103f568548cf32c3d7ca09/packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts#L214

It seems to emit too late for the test to check the amount of changes. But all other tests do succeed. Is this an issue with the test?

skipgeldens:
Is it possible to add this to the next release candidate? We rely heavily on the immutability in our project, and I think more users do, as ngrx is enforcing immutability.

mathisscott:
@gnomeontherun : what's the temperature on this for v4?

packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts
```diff
@@ -528,7 +549,8 @@ export default function (): void {
 
         it('does not apply trackBy to single selection with no items', () => {
           const emptyItems = new Items(filtersInstance, sortInstance, pageInstance);
-          const selection = new Selection(emptyItems, filtersInstance);
+          const ngZone = TestBed.get(NgZone);
```
gnomeontherun, line 552: nit: ngZone should be moved up into the beforeEach

skipgeldens: Okay, I'll move up the ngZone in the test! But, as @mathisscott said, the test is now checking for mutability. Should I change the test to make it pass? Normally I don't like to change other tests as they have of course been written with a purpose.

Jinnie (approved):
It's okay with me, as code.
Though, as Jeremy mentioned - the intention of the debounce is not entirely clear looking at it, so it will be great if a comment could be added, describing this intention and the use-case we're solving.

Jinnie:
Thanks for the quickly updated comment.

@gnomeontherun, ping - requested changes done here, and comment added to debounce line.

gnomeontherun:
@Jinnie can you review and merge if all is good? Needs backport to v4.

packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.spec.ts
```diff
@@ -193,7 +193,7 @@ export default function (): void {
         expect(currentSelection.sort(numberSort)).toEqual(itemsInstance.displayed);
         selectionInstance.toggleAll();
         expect(currentSelection).toEqual([]);
-        expect(nbChanges).toBe(3);
+        expect(nbChanges).toBe(4);
```
Jinnie, line 196: nit: Merging back this PR, I noticed this line and realized I have no idea where the extra change comes from. Any suggestions?

Jinnie: Found it. Two `emitChanges()` calls - one in setSelected and one more in updateCurrent(), which is chain-called by setSelected.
Will fix it in the v3/v2 merge directly and then make a separate PR for v4/master

Jinnie: done

PR #4787, opened 2020-07-14
Fix: backward compatibility for non Ivy build related to Datagrid

bdryanovski:
This is backward compatibility check for non-Ivy build to call or not to call a method.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #4692

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Close: #4692

PR #4788, opened 2020-07-14
fix: backport datagrid missing cells issue

bdryanovski:
Backport of this PR #4787

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/wrapped-cell.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/wrapped-cell.ts
@@ -6,6 +6,7 @@
 import { AfterViewInit, Component, EmbeddedViewRef, OnDestroy, TemplateRef, ViewChild } from '@angular/core';
 
 import { DynamicWrapper } from '../../utils/host-wrapping/dynamic-wrapper';
+import { ifIvyEnabled } from '../../utils/ivy/if-ivy-enabled';
 
 @Component({
   selector: 'dg-wrapped-cell',
@@ -25,6 +26,15 @@ export class WrappedCell implements DynamicWrapper, AfterViewInit, OnDestroy {
   }
 
   ngOnDestroy() {
-    this.cellView.destroy();
+    /**
+     * @TODO
+     * Check to see if Ivy is enabled and if so call destroy method.
+     * This is related to this issues #4692 #4232
+     *
+     * This check is for backward compatibility for NON-Ivy builds
+     */
+    ifIvyEnabled(() => {
+      this.cellView.destroy();
+    });
   }
 }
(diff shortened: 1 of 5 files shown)


packages/angular/projects/clr-angular/src/data/datagrid/wrapped-column.ts
```diff
@@ -27,6 +27,15 @@ export class WrappedColumn implements DynamicWrapper, AfterViewInit, OnDestroy {
...
-    this.columnView.destroy();
+    /**
+     * @TODO
+     * Check to see if Ivy is enabled and if so call destroy method.
+     * This is related to this issues #4692 #4232
+     *
+     * This check is for backward compatibility for NON-Ivy builds
+     */
+    if (this[`__ngContext__`]) {
+      this.columnView.destroy();
```
mathisscott, line 38: Would it make sense to make this a separate helper function as opposed to repeating the conditional in three places?

That would also make it easier to unit test in isolation...

```
function checkIfIvyAndThen(context, fn) {
    if(context[`__ngContext__`] {
       fn();
    }
}
```

...and...

```
ngOnDestroy() {
    checkIfIvyAndThen(this, () => {
        this.columnView.destroy();
    }
}
```

bdryanovski: Give me a few minutes .. to find a place to put it in...

packages/angular/projects/clr-angular/src/data/datagrid/wrapped-cell.ts
```diff
@@ -25,6 +25,15 @@ export class WrappedCell implements DynamicWrapper, AfterViewInit, OnDestroy {
...
   ngOnDestroy() {
-    this.cellView.destroy();
+    /**
+     * @TODO
+     * Check to see if Ivy is enabled and if so call destroy method.
+     * This is related to this issues #4692 #4232
+     *
+     * This check is for backward compatibility for NON-Ivy builds
+     */
+    if (this[`__ngContext__`]) {
```
Shijir, line 35: I looked in the Angular source to see whether they offer a way for users to check Ivy is enabled or not. Apparently they do have [privately export module](https://github.com/angular/angular/blob/master/packages/core/src/core_private_export.ts#L21) for that, which means it's not exactly public API. But it's still better than what we are doing here with `__ngContext__`.  So basically we could do:

```ts
import { ..., ɵivyEnabled as ivyEnabled  } from '@angular/core';

...

ngOnDestroy() {
    if(ivyEnabled) {
      this.rowView.destroy();
    }
}

Comment (mathisscott) [reply packages/angular/projects/clr-angular/src/data/datagrid/wrapped-cell.ts:35]: If that's the case, then the utility function is an even easier path...

Comment (mathisscott) [reply packages/angular/projects/clr-angular/src/data/datagrid/wrapped-cell.ts:35]:
```
import { eivyEnabled as ivyEnabled } from '@angular/core';

function checkIfIvyAndThen(fn) {
    if(ivyEnabled) {
       fn();
    }
}
```
...and...

```
ngOnDestroy() {
    checkIfIvyAndThen(() => {
        this.columnView.destroy();
    }
}
```

Shijir: Well in both cases, we are relying on Angular's private implementation detail. But the change I am suggesting relies on the exported variable that tells us directly what we want to know. `__ngContext__`, on the other hand, is completely private property.

bdryanovski:
quickest merge .. ;)

Issue #4487, opened 2020-04-20
create a util function for checking keyCodes in core


jeeyun:
## Describe the bug

We have a couple of places where we are doing checks to see if a key pressed is a certain key (ESC, ENTER, etc). Let's refactor this into a reusable function.

## Expected behavior
Instead of having checks like:
 ```
if (e.keyCode === ESC || e.key === 'Esc') {
   ...
}
```
where `e` is of type `KeyboardEvent` we could have a reusable function, e.g.:
```
if (keyEquals(e, Keys.Escape)) {
   ...
}
```
And the function would check against our dictionary of tuples to check the key in a cross-browser compatible way.

mathisscott:
We need to get #4479 merged before we do this...

PR #4789, opened 2020-07-14
fix: keycode utilities

mathisscott:
• removed keycode enums
• added keycode service to manage registry of keycodes
• allows users to add their own keycodes
• added onKey, onAnyKey, and onKeyCombo utility functions
• added isStringAndNotNilOrEmpty utility function
• added Maybe monadic data structure to help with defensive coding
• updated button and modal to use new keycode utilities
• adding deepClone utility fn
• fixed issue with Maybe mappings trying to turn bad function returns Just
• added isMap utility function to identity.ts
• improved deepClone() utility function to handle maps, arrays, and objects
• cleaning up some warnings and issues with clr/react

## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Bugfix
- [x] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Previously we were duplicating keycode event checks across multiple files. The logic was implemented in an inconsistent fashion (see clr/angular) and what's worse is we were checking for the wrong things when it came to spacebars in some situations.

This change consolidates that logic into one place. DRYs it all up and creates a potential integration point with clr/angular.

Issue Number: #4487 

- [ ] Yes
- [x] No


packages/core/src/internal/services/keycodes.service.ts
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Maybe } from '../data-structures/maybe.js';
+import { deepClone } from '../utils/identity.js';
+
+export enum KeyCodeTypes {
```
bbogdanov, line 10: What do you think about using `const enum` instead of `enum`?

mathisscott: Digging through it, it appears we don't need a lookup table so using `const` would be fine. In addition, I found that this enum is only used inside this file so we don't need to export it.

packages/core/src/internal/services/keycodes.service.ts
```diff
@@ -0,0 +1,139 @@
...
+export function getKeycodeFromRegistry(
+  codeToLookup: string,
+  whichCodeType: 'code' | 'ie-code' = KeyCodeTypes.DEFAULT,
+  registry = keyCodeRegistry
+): string {
+  return Maybe.of(registry.get(codeToLookup))
+    .map(mbMapVal => {
+      return (mbMapVal.get && mbMapVal.get(whichCodeType)) || null;
+    })
+    .valueOr('').unsafeValue;
```
coryrylan, line 138: nit: could this be simplified if the maybe works with undefined?

```typescript
 return Maybe.of(registry.get(codeToLookup))
    .map(mbMapVal => mbMapVal?.get(whichCodeType))
    .valueOr('').unsafeValue;
```

mathisscott: Yep. Good suggestion!

packages/core/src/internal/utils/keycodes.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (keyArray.length === 0) {
+    return true;
+  }
+
+  const curriedChecker = curryN(2, keyWasEvented);
+  const myCheckerFn = curriedChecker(evt);
+
+  return keyArray.reduce((acc: boolean, currentKey: string) => {
+    return acc !== false && myCheckerFn(currentKey);
+  }, true);
```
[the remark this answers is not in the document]
mathisscott: Yeah, I think that will work. I will give it a try. Thanks!

PR #4799, opened 2020-07-16
fix(a11y): hidden button into vertical-nav from been tabbable

bdryanovski:
Hide a helper button from tabbing to prevent noise from SR

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #4117 

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

## Other information
Close #4117

PR #4800, opened 2020-07-16
fix(a11y): hidden button into vertical-nav from been tabbable backpor…

bdryanovski:
Backporting from v4 #4799


bdryanovski:
> Thanks

No, I thank you!

PR #4816, opened 2020-07-22
core: demo apps, alert fix, API fixes

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

## What is the new behavior?
- Added angularjs demo app
- All core demos now use the local copy of core + yarn
- Fixed Alert lifecycle issue when alerts were out of sync with group in most frameworks
- Added lit-analyzer (same tool that generates API docs) to lint for common component bugs
- Added 4.0 changelog
- Updated Core dependencies
- Improved react version detection
- Fixed duplicate icon service with @clr/react

- [ ] Yes
- [ ] No


packages/core/src/icon/index.ts
```diff
@@ -420,13 +420,17 @@ export { trailerIcon } from './shapes/trailer.js';
...
+export { loadMediaIconSet, mediaCollectionIcons, mediaCollectionAliases } from './collections/media.js';
+export { loadMiniIconSet, miniCollectionIcons, miniCollectionAliases } from './collections/mini.js';
+export {
+  loadTechnologyIconSet,
+  technologyCollectionIcons,
+  technologyCollectionAliases,
+} from './collections/technology.js';
+export { loadSocialIconSet, socialCollectionIcons, socialCollectionAliases } from './collections/social.js';
+export { loadTextEditIconSet, textEditCollectionIcons, textEditCollectionAliases } from './collections/text-edit.js';
+export { loadTravelIconSet, travelCollectionIcons, travelCollectionAliases } from './collections/travel.js';
```
coryrylan, line 436: fixes perf issue with Parcel imports

packages/core/src/internal/utils/framework.spec.ts
```diff
@@ -4,19 +4,34 @@
...
     document.body.setAttribute('data-reactroot', '');
     expect(getReactVersion(false)).toBe('unknown version');
     document.body.removeAttribute('data-reactroot');
+    window.CDS._react.version = '0.0.0';
+    expect(getReactVersion(false)).toBe('0.0.0');
+    window.CDS._react.version = undefined;
+    const temp = window.CDS;
+    window.CDS = undefined;
+    expect(getReactVersion(false)).toBe(undefined);
+    window.CDS = temp;
```
coryrylan, line 34: Add improved support for react and angularjs version detection

packages/core/src/alert/alert-actions.element.scss
```diff
@@ -14,51 +14,41 @@
...
 }
 
 :host([type='banner']) {
   --action-size: #{$cds-token-space-size-9};
-
-  ::slotted(cds-button) {
-    // slight nudges in alert-actions adjusting for styling across slots.
-    // !important necessary to override specificity of layout styles.
-    margin-top: #{$cds-token-space-size-1} !important;
-  }
```
coryrylan, line 54: Moving the layout of the actions to be managed by the alert removed some of the slot weirdness. Was able to cross compare with master and keep exact rendered layout and white space.

packages/core/src/alert/alert-actions.element.ts
```diff
@@ -78,35 +26,40 @@ function updateLayoutsAndActionButtonStyles(alertActionsComponent: CdsAlertActio
...
+  }
 
   connectedCallback() {
     super.connectedCallback();
     assignSlotNames([this, 'actions']);
-    updateLayoutsAndActionButtonStyles(this, this.parentElement as CdsAlert);
   }
 
-  updated() {
-    updateLayoutsAndActionButtonStyles(this, this.parentElement as CdsAlert);
```
coryrylan, line 105: The main source of the lifecycle issues is when the child component refers to the parent. Depending on how the framework renders and inserts the element into the DOM it could be inversed order which breaks the lookup. This was happening in vuepress and angular. I removed a similar parentElement reference in forms.

For the cases of needing to pass data to the child we can use the `@internalProperty` decorator which will work regardless of render order. It will also mark it as an internal API omitting it from the docs.

packages/core/src/alert/alert-group.element.ts
```diff
@@ -154,6 +100,19 @@ export class CdsAlertGroup extends LitElement {
...
+    super.updated(props);
+    await Promise.all(Array.from(this.alerts).map(a => a.updateComplete));
+
+    Array.from(this.alerts).forEach(alert =>
+      syncProps(alert, this, {
+        status: props.has('status') && this.type !== 'light' && alert.status !== 'loading',
+        type: props.has('type'),
+        size: props.has('size'),
+      })
+    );
```
coryrylan, line 113: Added a util func similar to the one I used in forms. It lets us define a set of common props between a parent and child component and the condition when it should set them.

packages/core/src/alert/alert-group.stories.ts
```diff
@@ -58,7 +58,9 @@ export const alertGroup = () => {
           of "info".
           <cds-alert-actions>
             <cds-button>Button 1</cds-button>
-            <cds-button>Button 2</cds-button>
+            <cds-button>
+              <a href="#">Link 1</a>
+            </cds-button>
```
coryrylan, line 60: Added some link demos as that was a question with regards to the new website

packages/core/src/alert/alert-group.stories.ts
```diff
@@ -137,50 +147,56 @@ export const alertGroup = () => {
...
+      <cds-alert-group type="banner" status="success">
+        <cds-alert closable>
           This example is a closable alert with a status of "success" inside a banner alert group.
         </cds-alert>
       </cds-alert-group>
 
-      <cds-alert-group type="banner">
-        <cds-alert status="warning" closable>
+      <cds-alert-group type="banner" status="warning">
+        <cds-alert closable>
```
coryrylan, line 160: Similar to angular data flow has to be top down with the prop setting since we cant control the render order. So the group can set the child elements but not the child. If we want the child to alert a parent state then we would need to use some additional custom events.

packages/core/src/alert/alert.element.ts
```diff
@@ -230,46 +193,41 @@ export class CdsAlert extends AlertMixinClass {
...
 
-  updated() {
-    this.updateSlots(this.alertGroupType);
-
-    if (this.alertGroupType === 'banner') {
-      (this.parentElement as CdsAlertGroup).updateBannerAlertGroupStatus();
-    }
+  updated(props: Map<string, any>) {
+    super.updated(props);
+    syncDefinedProps(props, this, [this.alertActions]);
```
coryrylan, line 211: The `syncDefinedProps` handles most of the syncing logic above.

apps/core-parcel-js/src/index.js
```diff
@@ -1,4 +1,4 @@
-import '@clr/core/button/register.js';
+import '@clr/core/button/register';
```
coryrylan, line 1: not sure why this got removed? I can add back

apps/core-create-react-app/package.json
```diff
@@ -1,25 +1,28 @@
...
     "react": "^16.13.0",
     "react-dom": "^16.13.0",
     "react-scripts": "3.4.1",
     "typescript": "^3.7.5"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
+    "build": "SKIP_PREFLIGHT_CHECK=true react-scripts build",
```
coryrylan, line 25: Create react app does not like how yarn resolves jest. This skips is validation check. I figured that would be ok for this demo since we are not adding tests here.

apps/core-create-react-app/src/index.tsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import './index.css';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
 
+import 'normalize.css/normalize.css';
+import '@clr/core/global.css'; // pre-minified version breaks
```
coryrylan, line 7: Not sure why but the version of postcss create react app uses does not like our pre-minified version of the global CSS.

packages/core/src/forms/control/control.element.ts
```diff
@@ -273,7 +280,11 @@ export class CdsControl extends LitElement {
     this.inputControl.addEventListener('focusout', () => (this.focused = false));
     this.disabled = this.inputControl.disabled;
     this.observers.push(
-      listenForAttributeChange(this.inputControl, 'disabled', () => (this.disabled = this.inputControl.disabled))
+      listenForAttributeChange(this.inputControl, 'disabled', () => {
+        // if (this.inputControl.disabled !== this.disabled) {
+        this.disabled = this.inputControl.disabled;
+        // }
```
coryrylan, line 286: Forgot to clean this back up. Will remove commented code

packages/core/src/index.ts
```diff
@@ -6,3 +6,18 @@
...
+    'cds-text': string;
+    'cds-layout': string;
+    'cds-list': string;
+  }
+
+  interface HTMLElementTagNameMap {
+    'cds-card': HTMLElement;
+    'cds-demo': HTMLElement;
+    'cds-placeholder': HTMLElement;
+  }
```
coryrylan, line 22: The type checker for the templates would complain about these since they are not defined elements. I'm still looking around see if there is a better spot these could exist. Maybe a types.d.ts in the directory would work?

packages/core/tsconfig.json
```diff
@@ -30,7 +30,18 @@
...
+      {
+        "name": "ts-lit-plugin",
+        "strict": true,
+        "rules": {
+          "no-missing-import": "off",
+          "no-unknown-attribute": "off",
+          "globalAttributes": ["cds-layout", "cds-list", "cds-text"]
+        }
+      }
+    ]
```
coryrylan, line 44: I disabled two of the rules for the lit-element linter. The `no-missing-import` is not picking up the `register` files. The `no-unknown-attribute` is not picking up any kebab case attributes. I think both of these are open issues so once the tool fixes them we can enable the checks.

apps/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "clarity-demo-apps",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
```
mathisscott, line 5: 🏆

packages/core/docs/changelog.stories.mdx
```diff
@@ -4,6 +4,75 @@ import { Meta, Story, Preview, API } from '@storybook/addon-docs/blocks';
...
+```javascript
+import '@clr/core/button'; // 3.0
+
+import '@clr/core/button/register'; // 4.0+
+```
+
+- ### Alerts / Alert Groups
+
+  The Alert APIs are revamped to support new features including alert groups/actions as
+  well as lightweight alerts which are now exclusivly in Core. Alerts APIs will
```
mathisscott, line 70: `exclusively` not `exclusivly`

packages/core/src/icon/icon.stories.ts
```diff
@@ -389,10 +391,10 @@ export const legacy = () => {
     <clr-icon shape="info-circle" style="width: 72px; height: 72px;"></clr-icon>
 
     <h2>Direction</h2>
-    <clr-icon shape="caret" dir="up"></clr-icon>
+    <!-- <clr-icon shape="caret" dir="up"></clr-icon>
     <clr-icon shape="caret" dir="right"></clr-icon>
     <clr-icon shape="caret" dir="down"></clr-icon>
-    <clr-icon shape="caret" dir="left"></clr-icon>
+    <clr-icon shape="caret" dir="left"></clr-icon> -->
     <clr-icon shape="caret" style="transform: rotate(0deg);"></clr-icon>
```
mathisscott, line 398: Why did we make this change here?

packages/core/src/internal-components/close-button/register.ts
```diff
@@ -7,7 +7,7 @@
 import '@clr/core/icon/register.js';
 import '@clr/core/button/register.js';
 import { registerElementSafely } from '@clr/core/internal';
-import { CdsInternalCloseButton } from '@clr/core/internal-components/close-button';
+import { CdsInternalCloseButton } from './close-button.element.js';
```
mathisscott, line 10: 👍

packages/react/package.json
```diff
@@ -17,7 +17,7 @@
     "test:snap:update": "jest --updateSnapshot"
   },
   "dependencies": {
-    "@clr/core": "^3.1.0",
+    "@clr/core": "^4.0.0-rc.0",
```
mathisscott, line 20: Why is the `4.0.0-rc.1` version of `@clr/react` relying on the `4.0.0-rc.0` version of `@clr/core`?

packages/core/docs/changelog.stories.mdx
```diff
@@ -4,6 +4,75 @@ import { Meta, Story, Preview, API } from '@storybook/addon-docs/blocks';
...
+Several bugfixes have been made to improve performance and treeshaking of components.
+
+### React Support
+
+React is now officially supported with our new `@clr/react` package. Follow our
+[React Documentation](./?path=/story/documentation-react--page) to get started.
+
+```bash
+npm install @clr/react
+```
```
jeeyun, line 20: should we mention `@clr/core` as an optional peer dependency? Or is that implied since this is in the dox for core?

coryrylan: Good point! In the package json of the react package we have it listed as a direct dependency but it really should be moved to peer deps. That way as people build on top of core they share the same version in one app. 

```json
{
  "dependencies": {
    "@clr/core": "4.0.0",
    "@clr/react": "4.0.0",
    "@clr/city": "1.1.0",
    "super-cool-lib-using-core": "1.2.3"
  } 
}
```

So in this example app we would have all these deps installed, then `@clr/react` and `super-cool-lib-using-core` would have their `peerDependencies` list `@clr/core`. This will ensure all apps share the same version like angular libraries. I'll get this fixed if this sounds right.

jeeyun: Should it remain under `dependencies` for `src/packages/react` for development (since it is a strict dependency for developing react wrappers), and we move it to `peerDependencies` for the `package.json` of the react package when publishing?

packages/core/src/icon/icon.element.ts
```diff
@@ -159,7 +158,7 @@ export class CdsIcon extends IconMixinClass {
    * Setting the badge to 'null' removes the attribute from the DOM.
```
jeeyun, line 158: nit: update the comment about setting it to null?

packages/react/tsconfig.build.json
```diff
@@ -29,7 +29,7 @@
     "baseUrl": "./",
     "rootDir": "./src",
     "paths": {
-      "@clr/core/*": ["../../dist/core/*"]
+      "@clr/core/*": ["../core/dist/core/*"]
```
jeeyun, line 32: Something to double check here is make sure that both versions are the same. I initially pointed it to the parent-level `dist` folder because that's what gets published to npm.

packages/react/tsconfig.json
```diff
@@ -28,7 +28,7 @@
     "baseUrl": "./",
     "rootDir": "./src",
     "paths": {
-      "@clr/core/*": ["../../dist/core/*"] // This mapping is relative to "baseUrl"
+      "@clr/core/*": ["../core/dist/core/*"] // This mapping is relative to "baseUrl"
```
jeeyun, line 31: Same comment as above.

packages/react/tsconfig.build.json
coryrylan: Reverted this change. I think once we finish setting up the yarn workspaces we wont need TS path mapping

packages/core/src/forms/control-label/control-label.element.scss
```diff
@@ -22,23 +23,23 @@
...
-  font-size: $cds-token-typography-secondary-font-size;
-  font-weight: $cds-token-typography-secondary-font-weight;
-  color: $cds-token-typography-secondary-color;
-  line-height: $cds-token-typography-secondary-line-height;
-  letter-spacing: $cds-token-typography-secondary-letter-spacing;
+  font-size: $cds-token-typography-secondary-font-size !important;
+  font-weight: $cds-token-typography-secondary-font-weight !important;
+  color: $cds-token-typography-secondary-color !important;
+  line-height: $cds-token-typography-secondary-line-height !important;
+  letter-spacing: $cds-token-typography-secondary-letter-spacing !important;
```
coryrylan, line 30: When using core in a app that has a global CSS framework defined already global elements like label/input collide. The important ensure slotted elements are consistent.

packages/core/src/modal/modal-content.element.ts
```diff
@@ -32,7 +32,9 @@ import { html, LitElement } from 'lit-element';
  */
 export class CdsModalContent extends LitElement {
   render() {
-    return html` <slot></slot> `;
+    return html`<slot
+      cds-layout="${this.hasAttribute('cds-layout') ? 'display:contents' : 'vertical gap:lg p-y:xs'}"
+    ></slot>`;
```
coryrylan, line 37: This check lets users use `cds-layout` on the host element directly

packages/core/src/styles/layout/_optimize.scss
```diff
@@ -5,7 +5,8 @@
 // This is a performance optimization that aliases the token to
 // a "minified" value to reduce overall bundle size.
 
-:root {
+:root,
+:host {
```
coryrylan, line 9: The optimized shorter variables were missing from the base component styles causing styles to fail when global CSS is not used.

packages/core/src/internal/base/button.base.ts
```diff
@@ -73,6 +75,11 @@ export class CdsBaseButton extends LitElement {
     `;
   }
 
+  connectedCallback() {
+    super.connectedCallback();
+    this.tabIndex = 0; // initialize immediately so button can be focused synchronously
```
coryrylan, line 80: This allows users to reference and call `focus()` on an element immediately. Later in the lifecycle if the button is set to readonly the `tabindex` is removed.

packages/core/docs/changelog.stories.mdx
coryrylan: Updated the package.js script to set core as a peer dependency when building for publish. Local developement its set as a regular dependency.

PR #4863, opened 2020-08-03
fix: core icon color, website core icon load

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
Fixes an issue where cds-icon does not apply fill color based on legacy clr-ui behavior. Swaps the website icons to use cds-icon legacy mode instead of clr-icon. This reduces the home page landing JS from 417kb gzip to about 218kb gzip.

Screen Shot 2020-08-03 at 3 55 57 PM

Screen Shot 2020-08-03 at 3 55 47 PM

- [ ] Yes
- [x] No

### DIFF
--- a/apps/website/.vuepress/public/demos/sidenav/icons-css.html
+++ b/apps/website/.vuepress/public/demos/sidenav/icons-css.html
@@ -8,8 +8,8 @@
         <a class="nav-link nav-icon"><clr-icon shape="home"></clr-icon> Home</a>
         <a class="nav-link"><clr-icon shape="cog"></clr-icon> Settings</a>
         <a class="nav-link active"><clr-icon shape="user"></clr-icon> Users</a>
-        <a class="nav-link"><clr-icon shape="Folder-open"></clr-icon> Files</a>
-        <a class="nav-link"><clr-icon shape="Cloud"></clr-icon> Data</a>
+        <a class="nav-link"><clr-icon shape="folder-open"></clr-icon> Files</a>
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "clean:modules": "del ./**/node_modules",
     "clean": "del ./dist",
     "start": "yarn angular:dev:start",
-    "build": "npm-run-all build:ci",
+    "build": "npm-run-all build:ci build:website",
```
hippee-lee, line 31: I'm not sure about adding this here. imo we should separate lib builds from website builds because we have the gitub action that builds and deploys website. This includes seperate tasks for vuepress, sb:angular & sb:core. The separation of website/storybook-ng/storybook-core lets us see when one of them fails.

coryrylan: Oops sorry I meant to leave this change out of this PR. The `build` task is not called by the CI. Github calls [`build:ci` instead](https://github.com/vmware/clarity/blob/master/.github/workflows/build.yml#L15). Before the old site was removed from the repo the [`build` script was just a short cut](https://github.com/vmware/clarity/blob/ec782d2859e1a0e499c12acd1376d6027e1ac846/package.json#L10 ) to run the `build:ci` and `build:website` commands. It was helpful for devs starting out in the repo. As long as they ran `build` and it passes then its unlikely the CI would fail. I can pull it out of this commit tho so the git history is clean.

coryrylan: Reverted the script change 👍

PR #4887, opened 2020-08-10
revert: check is Ivy enabled or disabled

bdryanovski:
The check is no longer needed it seems that everything is working now and no leftover elements are being uncleaned 

Also, there is no longer a need to check for ivy enabled/disabled it seems that this is handle inside the `ngcc`, and the correct version of the code is provided to the parent application.

I created two versions one including the check and one without and tested it inside a real application with enabled and disabled Ivy build - no error was thrown and no performance issues with the datagrid was found.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #4841

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #4888, opened 2020-08-10
revert: check is Ivy enabled or disabled (backport from v4)

### LINKED ISSUE #4841: Memory leak in datagrids in Clarity 3.1.5
## Describe the bug
Grid performance degraded a lot when we upgraded to 3.1.5 (from 3.1.3). This seems to be because of a memory leak; before we upgraded, we could leave a screen running with a grid of our vApps for over a day and memory would grow to about ~1GB from ~300MB when the app first launched. We knew that wasn't great but most people don't leave the app running for days.

Some of our screen are constantly polling the server to get updates so the rows may be re-rendering every two seconds (when there are busy entities) or every 10 seconds (when entities aren't busy)

After the upgrade, our app starts using 1GB of memory within an hour, sometimes less.

## How to reproduce

I was able to reproduce the problem by creating a grid that updates more than once a second and letting it run for a while. 

* If I run https://stackblitz.com/edit/clarity-v3-grid-polling-314 (which uses Clarity 3.1.4), the frame running the app does not crash if left up for a day and memory is stable after GC
Screen Shot 2020-07-29 at 6 56 35 PM

* If I run https://stackblitz.com/edit/clarity-v3-grid-polling-315 (which uses Clarity 3.1.5), the frame running the app crashes within 5 minutes (when memory reaches 4GB) with the current `rowCount`/`columnCount`/`updateInterval` settings
Screen Shot 2020-07-29 at 6 55 45 PM

[Clarity 3.1.3](https://clarity-v3-grid-polling-313.stackblitz.io/) seems to have the same correct behavior as Clarity 3.1.4

## Expected behavior
I would expect neither stackblitz to crash after running for a day, and I would the memory to stay at similar levels after GC

## Versions

**App**

- Angular/core : 9.1.0
- Node:  v12.16.1
- Clarity: 3.1.5 (compared to performance of 3.1.4)

**Device:**

- Type: Tested on both MacBook and Linux
- OS: Mac High Sierra and Ubuntu
- Browser: Chrome
- Version: Version 84.0.4147.89 (Official Build) (64-bit)

## Additional notes

* @gnomeontherun is under the impression it may be caused by https://github.com/vmware/clarity/commit/4c7753d8dc9892de217bbe6683108776e33cc8a0

### On Using `trackBy`
* Not having a `trackBy` on `<clr-dg-column *ngFor="">` also causes a memory leak (on 3.1.3 and 3.1.5)
* Not having a `trackBy` on `<clr-dg-cell *ngFor="">` causes a layout bug (3.1.3 and 3.1.5) 
image but does not cause a memory leak
* ⚠️ Not having a `trackBy` on `<clr-dg-row  *ngFor="">` causes an even worse bug, the rows are appended to the DOM with every refresh. Notice how small the scrollbar is in the screenshot below 
image

* Forcing no extra creation/destruction of `clr-dg-row` nodes by returning the `index` from the `trackBy` function makes 3.1.5 seem like it doesn't have a problem, but it real case scenarios, records/nodes will be removed/added causing memory leaks

### Workaround (kind of)
* Using a proper implementation of `trackBy`  on `<clr-dg-row>` that doesn't destroy/create components as often (only when records are removed/added to view) probably makes our app usable again for 24 hours, unless users are doing something that would trigger creation/destruction of nodes.

### Playing with both stackblitz
The stackblitz `app.component.ts` file has properties to easily configure how many rows, columns and how often to refresh to test different configurations.

### Measuring memory
Be sure to open the performance monitor from the https://clarity-v3-grid-polling-313.stackblitz.io/ URLs since the `/edit` has the editor and can be confusing with all the frames

bdryanovski:
Backport of #4887

Close #4841


gnomeontherun:
Please update commit message to use `revert` like we see here https://www.npmjs.com/package/conventional-changelog-angular

PR #4886, opened 2020-08-10
fix(datagrid): disable whitespace wrap in column title

bbogdanov:
Signed-off-by: Bogdan Bogdanov <bbogdanov@vmware.com>

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #4872

## What is the new behavior?

The title doesn't wrap on column resize.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Before
image

After (horizontal scroll available)
image

After (min-width => title content)
image

PR #4890, opened 2020-08-10
fix(datagrid): disable whitespace wrap in column title (port)

bbogdanov:

port from #4886 (v3)

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
@@ -839,6 +839,14 @@
         align-self: center;
         display: flex;
 
+        /**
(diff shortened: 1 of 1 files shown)


bbogdanov:
> This solution breaks our column resizing. Line breaks in Datagrid columns are valid and useful for many users. This kind of solution should be offered as a workaround to a specific user's need. We need to discuss it further.
> 
> 
> 
> This is this PR's build:
> 
> https://5f316e16d10050bc5c1adfe9--vmware-clarity.netlify.app/datagrid/column-sizing

What do you suggest? I already pushed this code in 'v3' btw

gnomeontherun:
Please revert the v3 change, and then we need to come up with a new solution.

bbogdanov:
> Please revert the v3 change, and then we need to come up with a new solution.

#4893 Revert commit

PR #4898, opened 2020-08-13
core: browser layout fixes

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
- add missing responsive alignment doc examples
- aligns slot layout behavior with child combinator behavior in horizontal/vertical layouts
- improve slot behavior consistency between all browsers
- several fixes with layout util inconsistencies in Safari and IE
- fix tree shaking issue with layout untils internal to components

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/alert/alert.element.scss
+++ b/packages/core/src/alert/alert.element.scss
@@ -22,11 +22,6 @@ $lightweight-alert-line-height: $cds-token-typography-body-line-height;
   min-height: var(--min-height);
 }
 
-slot[name='close-button'],
-slot[name='alert-icon'] {
-  line-height: 0;
-}
-
 :host([type='default']) {
   --font-size: #{$cds-token-typography-font-size-3};
 }
@@ -125,8 +120,7 @@ cds-internal-close-button {
 }
 
 :host([type='banner']) {
-  .alert-content,
-  .alert-content slot {
+  .alert-content {
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
--- a/packages/core/src/forms/control-group/control-group.element.ts
+++ b/packages/core/src/forms/control-group/control-group.element.ts
@@ -105,9 +105,9 @@ export class CdsInternalControlGroup extends LitElement {
   /** @private */
   @query('cds-internal-control-label[action=primary]') controlLabel: CdsInternalControlLabel;
 
-  @query('slot[name=controls]') private controlSlot: HTMLElement;
+  @query('.controls') private controlSlot: HTMLElement;
 
-  @query('slot[name=message]') private messageSlot: HTMLElement;
+  @query('.messages') private messageSlot: HTMLElement;
 
   @id() private groupLabelId: string;
 
@@ -127,7 +127,9 @@ export class CdsInternalControlGroup extends LitElement {
     return html`
       <div cds-layout="horizontal align:shrink ${this.messages?.length ? 'gap:sm' : ''}" class="messages-container">
         ${!this.isInlineControlGroup ? getStatusIcon(this.status) : ''}
-        <slot class="messages" name="message"></slot>
+        <div class="messages">
+          <slot name="message"></slot>
+        </div>
       </div>
     `;
   }
@@ -136,13 +138,17 @@ export class CdsInternalControlGroup extends LitElement {
     return this.isInlineControlGroup
       ? html`
           <div cds-layout="horizontal gap:sm align:horizontal-stretch" class="input-container">
-            <slot name="controls" cds-layout="horizontal align:horizontal-stretch"></slot>
+            <div class="controls" cds-layout="horizontal align:horizontal-stretch">
+              <slot name="controls"></slot>
+            </div>
             ${getStatusIcon(this.status)}
           </div>
         `
       : html`
           <div cds-layout="horizontal align:shrink" class="input-container">
-            <slot name="controls" cds-layout="${this.inlineControlLayout}"></slot>
+            <div class="controls" cds-layout="${this.inlineControlLayout}">
+              <slot name="controls"></slot>
+            </div>
           </div>
         `;
   }
--- a/packages/core/src/forms/forms.stories.ts
+++ b/packages/core/src/forms/forms.stories.ts
@@ -1851,7 +1851,7 @@ export const responsiveCheckoutForm = () => {
     const iframe = document.querySelector<HTMLIFrameElement>('#complex-form-demo');
     iframe.onload = () => {
       const resizeObserver = new ResizeObserver(
-        () => (iframe.style.height = `${iframe.contentWindow.document.body.scrollHeight}px`)
+        () => (iframe.style.height = `${iframe.contentWindow?.document.body.scrollHeight}px`)
       );
       resizeObserver.observe(iframe);
     };
--- a/packages/core/src/internal/base/base.element.scss
+++ b/packages/core/src/internal/base/base.element.scss
@@ -29,7 +29,7 @@
 
 slot {
   font-family: $cds-token-typography-font-family;
-  display: block;
+  display: contents !important;
 }
 
 // normalize focus styles
--- a/packages/core/src/modal/modal-actions.element.ts
+++ b/packages/core/src/modal/modal-actions.element.ts
@@ -38,9 +38,9 @@ export class CdsModalActions extends LitElement {
   }
 
   render() {
-    return html`<slot
-      cds-layout="${this.hasAttribute('cds-layout') ? 'display:contents' : 'horizontal gap:sm align:right'}"
-    ></slot>`;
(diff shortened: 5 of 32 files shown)


packages/core/src/forms/control-group/control-group.element.ts
```diff
@@ -105,9 +105,9 @@ export class CdsInternalControlGroup extends LitElement {
   /** @private */
   @query('cds-internal-control-label[action=primary]') controlLabel: CdsInternalControlLabel;
 
-  @query('slot[name=controls]') private controlSlot: HTMLElement;
+  @query('.controls') private controlSlot: HTMLElement;
 
-  @query('slot[name=message]') private messageSlot: HTMLElement;
+  @query('.messages') private messageSlot: HTMLElement;
```
coryrylan, line 110: Since safari defaults slot to display:contents and ie11 polyfill removes the slot element its best to not use the slot as a selector in JS or Scss

packages/core/src/alert/alert.element.scss
```diff
@@ -22,11 +22,6 @@ $lightweight-alert-line-height: $cds-token-typography-body-line-height;
   min-height: var(--min-height);
 }
 
-slot[name='close-button'],
-slot[name='alert-icon'] {
-  line-height: 0;
-}
```
coryrylan, line 28: Using display: contents fixes the various line height issues with slots

packages/core/src/forms/forms.stories.ts
```diff
@@ -1851,7 +1851,7 @@ export const responsiveCheckoutForm = () => {
     const iframe = document.querySelector<HTMLIFrameElement>('#complex-form-demo');
     iframe.onload = () => {
       const resizeObserver = new ResizeObserver(
-        () => (iframe.style.height = `${iframe.contentWindow.document.body.scrollHeight}px`)
+        () => (iframe.style.height = `${iframe.contentWindow?.document.body.scrollHeight}px`)
```
coryrylan, line 1854: Depending on the browser sometimes storybook loads this in the background and can cause errors if not checked

packages/core/src/styles/layout/_type-grid.scss
```diff
@@ -5,13 +5,11 @@
...
   @for $i from 1 through $cds-token-layout-grid-cols-static {
     // implicit columns/rows
-    [#{$layout}*='#{$cols}#{$breakpoint}:#{$i}'] > *,
-    [#{$layout}*='#{$cols}#{$breakpoint}:#{$i}']::slotted(*) {
+    [#{$layout}*='#{$cols}#{$breakpoint}:#{$i}'] > * {
       grid-column: span $i / span $i;
     }
 
-    [#{$layout}*='#{$rows}#{$breakpoint}:#{$i}'] > *,
-    [#{$layout}*='#{$rows}#{$breakpoint}:#{$i}']::slotted(*) {
```
coryrylan, line 14: Removed internal slot support for grid since we dont use them. Likely in the future will want to avoid viewport based media queries internally in favor of container queries with ResizeObserver.

packages/core/src/styles/layout/_type-horizontal.scss
```diff
@@ -2,135 +2,134 @@
...
+    & > ::slotted([#{$layout}*='#{$align}#{$breakpoint}:#{$vertical-center}']) {
       align-self: center;
     }
 
     & > [#{$layout}*='#{$align}#{$breakpoint}:#{$center}'],
     & > [#{$layout}*='#{$align}#{$breakpoint}:#{$horizontal-center}'],
-    &::slotted([#{$layout}*='#{$align}#{$breakpoint}:#{$center}']),
-    &::slotted([#{$layout}*='#{$align}#{$breakpoint}:#{$horizontal-center}']) {
+    & > ::slotted([#{$layout}*='#{$align}#{$breakpoint}:#{$center}']),
+    & > ::slotted([#{$layout}*='#{$align}#{$breakpoint}:#{$horizontal-center}']) {
```
coryrylan, line 91: This makes the slot selector behave much like the regular light DOM `>`. This seems to help improve consistency with the slot behavior in layouts

packages/core/src/styles/layout/_type-horizontal.scss
```diff
@@ -2,135 +2,134 @@
...
+}
+
 [#{$layout}~='#{$horizontal}'] {
   position: relative;
   display: flex;
   flex-direction: row;
   flex-wrap: wrap;
   justify-content: flex-start;
   align-content: flex-start;
+  align-items: flex-start;
```
coryrylan, line 132: Safari flexbox has different defaults than the rest of the browsers.

packages/core/src/styles/layout/shims/_gap.scss
```diff
@@ -27,14 +27,14 @@
...
+        margin-top: $sizeValue;
       }
 
       & > *:first-child,
-      &::slotted(*:first-child),
-      &::slotted([slot]:first-of-type) {
-        margin-top: 0 !important;
+      & > ::slotted(*:first-child),
+      & > ::slotted([slot]:first-of-type) {
+        margin-top: 0;
```
coryrylan, line 37: The specificity had to be lowered for Safari with the gap shim. This is so the `align:*` utils can override the margin.

packages/core/src/internal/base/base.element.scss
```diff
@@ -29,7 +29,7 @@
 
 slot {
   font-family: $cds-token-typography-font-family;
-  display: block;
+  display: contents !important;
```
mathisscott, line 32: 👍

packages/core/src/styles/layout/docs/layout.stories.mdx
```diff
@@ -86,20 +86,25 @@ manage layouts on many different screen sizes. All responsive utilities follow t
 
 <Story id="foundation-design-tokens-stories--layout" />
 
-```css
+```scss
 @media (min-width: 576px) {
```
mathisscott, line 90: We may want to consider setting these sizes in SASS variables or a SASS map like we used to so people can get to them more easily. Preferably they could be SASS variables generated by tokens.yml.

Out of scope with what you're doing here. But plz open a :octocat: issue if you agree.

packages/core/src/styles/layout/layout.stories.ts
```diff
@@ -1463,6 +1505,7 @@ export const utilitiesContainers = () => {
       <cds-demo layout cds-layout="container:md">container:md (992px)</cds-demo>
       <cds-demo layout cds-layout="container:lg">container:lg (1200px)</cds-demo>
       <cds-demo layout cds-layout="container:xl">container:xl (1440px)</cds-demo>
+      <cds-demo layout cds-layout="container:fill">'container:fill' or 'fill' (width 100%)</cds-demo>
```
mathisscott, line 1508: are this `container:*` values new? i don't remember seeing them before...

coryrylan: They have been there for a while I think. They are under the utils part of the layout docs. I think we use them in the demo layout patterns as well

packages/core/src/styles/layout/docs/layout.stories.mdx
coryrylan: We do have them in the tokens file and a sass map here
https://github.com/vmware/clarity/blob/master/packages/core/tokens.yml#L111 

The map is using the generated vars from the tokens file
https://github.com/vmware/clarity/blob/master/packages/core/src/styles/layout/mixins/_mixins.scss#L16 

I think this was here for documentation on what the breakpoint values were since we cant use custom props in the media queries.

packages/core/src/input/input.stories.ts
```diff
@@ -75,7 +75,7 @@ export const horizontal = () => {
   return html`
     <cds-form-group layout="horizontal">
       <cds-input layout="horizontal">
-        <label>label</label>
+        <label>label text here</label>
```
coryrylan, line 78: todo: remove, was from testing

packages/core/src/modal/modal-actions.element.ts
```diff
@@ -38,11 +38,9 @@ export class CdsModalActions extends LitElement {
...
 
   render() {
-    return html`
-      <div cds-layout="${this.hasAttribute('cds-layout') ? 'display:contents' : 'horizontal gap:sm align:right'}">
-        <slot></slot>
-      </div>
-    `;
+    return this.hasAttribute('cds-layout')
+      ? html`<slot></slot>`
+      : html`<div cds-layout="horizontal gap:sm align:right"><slot></slot></div>`;
```
coryrylan, line 43: If there is a cds-layout on host use the default slot with `display: contents`. This fixes the issue in IE since the slot tag is removed and the content will then become the direct child of the host element.

PR #4907, opened 2020-08-14
fix(core): form alignment issues

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
- allow icon whitespace to be offset
- adjust form to fix corrected whitespace in labels
- fix missing button lifecycle check

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/forms/control-action/control-action.element.scss
+++ b/packages/core/src/forms/control-action/control-action.element.scss
@@ -17,14 +17,11 @@
 }
 
 .private-host {
-  height: calc(#{$cds-token-space-size-9} - #{$cds-token-space-size-2});
   color: $cds-token-color-neutral-600;
   font-size: $cds-token-typography-font-size-3;
   display: flex;
   justify-content: center;
-}
-
-slot {
+  align-items: center;
   line-height: calc(#{$cds-token-space-size-9} - #{$cds-token-space-size-2});
 }
 
@@ -33,14 +30,6 @@ slot {
   height: calc(#{$cds-token-space-size-9} - #{$cds-token-space-size-2});
   margin-bottom: $cds-token-space-size-1;
   pointer-events: none;
-}
-
-::slotted(cds-icon[shape='exclamation-circle']),
-::slotted(cds-icon[shape='check-circle']),
-::slotted(cds-icon[shape='info-circle']) {
-  @include equilateral($cds-token-space-size-9);
-  margin: calc(-1 * #{$cds-token-space-size-3});
-  margin-bottom: 0;
 }
 
 ::slotted(cds-icon[shape='angle']) {
--- a/packages/core/src/icon/icon.element.ts
+++ b/packages/core/src/icon/icon.element.ts
@@ -11,6 +11,7 @@ import {
   hasStringPropertyChangedAndNotNil,
   Orientations,
   property,
+  internalProperty,
   StatusTypes,
   id,
 } from '@clr/core/internal';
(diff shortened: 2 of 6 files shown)


packages/core/src/icon/icon.element.ts
```diff
@@ -167,6 +175,13 @@ export class CdsIcon extends LitElement {
     if (props.has('title')) {
       this.updateSVGAriaLabel();
     }
+
+    if (props.has('innerOffset') && this.innerOffset > 0) {
+      const dimension = `calc(100% + ${this.innerOffset * 2}px)`;
+      this.svg.style.width = dimension;
+      this.svg.style.height = dimension;
+      this.svg.style.margin = `-${this.innerOffset} 0 0 -${this.innerOffset}`;
+    }
```
coryrylan, line 184: Some of the icons have whitespace around the svg itself which causes excess whitespace between elements. This lets us define a whitespace offset so if we define a icon to be 16px the icon is edge to edge of the 16px host cds-icon element with no extra white space.

Maybe this is something we could fix in the future with the SVGs as a breaking change? Or make it more automated to opt into whitespace removal? For now I kept this as a internal API.

packages/core/src/input/input-groups.stories.ts
```diff
@@ -22,7 +22,7 @@ export default {
 export const group = () => {
   return html`
     <div cds-layout="vertical gap:lg">
-      <cds-input-group>
+      <cds-input-group layout="vertical">
```
coryrylan, line 25: Fixed missing layouts when the default was switched

packages/core/src/internal/base/button.base.ts
```diff
@@ -82,17 +82,13 @@ export class CdsBaseButton extends LitElement {
...
     super.firstUpdated(props);
-    this.updateButtonAttributes();
     this.setupAnchorFocus();
     this.setupNativeButtonBehavior();
   }
 
   protected updated(props: Map<string, any>) {
     super.updated(props);
-    // if readonly or disabled attribute was updated, button attributes might need updating
-    if (props.has('readonly') || props.has('disabled')) {
```
coryrylan, line 93: This check was preventing the attrs from being updated when inheriting the base button class. Some of the components like tags had role of button even though they were set to readonly. I'm not 100% what was causing this but running the check every time resolves the issue.

packages/core/src/forms/control-action/control-action.element.scss
```diff
@@ -35,14 +32,6 @@ slot {
...
 }
 
-::slotted(cds-icon[shape='exclamation-circle']),
-::slotted(cds-icon[shape='check-circle']),
-::slotted(cds-icon[shape='info-circle']) {
-  @include equilateral($cds-token-space-size-9);
-  margin: calc(-1 * #{$cds-token-space-size-3});
-  margin-bottom: 0;
-}
-
```
coryrylan, line 45: With the whitespace offset removal in the cds-icon itself the styles become simpler as we dont have to compensate for the misalignment within the svg.

packages/core/src/icon/icon.element.ts
mathisscott: The SVGs all have some white space around the edge. This is because curved paths at the very edge of the Vbox will be cut off. I don't think we can get rid of that.

The white space is consistent, however. It's a ring of 1px White space at the full 36 x 36 px. So that ratio would scale up or down depending on the size of the icon.

I think this is fine how it is being done here. I like it better than the CSS hack. But it is absolutely possible for us to guesstimate a default offset. `needFnToRoundToNearestHundredth(0.27778 * intendedSizeOfIcon)`;

Issue #3936, opened 2019-10-22
Datagrid uses wrong trackBy and breaks selection


rootix:
## Describe the bug

We have a dynamic datagrid with a configuration source for the columns. Because of this, we use a ngFor-loop for the `clr-dg-column` and utilize a trackBy function to improve performance.

We experience the issue that if the grid is initialized with selected items, the first selected item is not selected but the last item of the list. By debugging i found out, that the selection feature is using the Items<T>.trackBy function and that this one is the one from the column-loop instead of the row-loop. This happens as there is `ClrDatagridItemsTrackBy<T>` which sets the trackBy function of the Items Object to the trackBy of any ngForTrackBy that is found in the grid.

I got things working by removing every trackBy in the whole grid, but this isn't really a good solution for this problem.

## How to reproduce

I added a minimal example here: https://stackblitz.com/edit/clarity-light-theme-v2-bzpjmm

- You can see in the Component, that the first and second item are selected programatically, but the selection switches to the second and third (last) item.
- In the console you can see that the trackBy function of `ClrDatagrid.items.trackBy` is the column-trackBy
- If you remove the trackBy of the column-loop, selection works again

## Expected behavior

The selection feature should only use the trackBy function of *clrDgItems or ngForOf of the `clr-dg-row` component and not any other unconstrained ngForOf within the grid.

## Versions

**App**

* Angular: 8
* Node: 10.16
* Clarity: 2.3.0

gnomeontherun:
I think the issue here is that the selector is generic for `ClrDatagridItemsTrackBy`: `'[ngForTrackBy]'` so its picking up the last trackBy to be rendered in the datagrid. We should be able to either make the selector more specific, or figure out how to filter it someway.

itimie:
Hi, I’m just following up on the issue raised, is there any update on a fix?
Thanks

dkanev:
Hi,
do you have a workaround for this problem?
I am using <clr-dg-action-bar> in our datagrid and have *ngFor for the buttons in it. After I set a trackBy function on the  *ngFor  for the action bar buttons a weird  selection on a last row appeared in my grid. I investigated and saw that the problem is the same as  the one described in this issue. 
Do you have a plan to fix this issue?

bbogdanov:
@rootix The team looked into the scenario you provided and [a little bit further.](https://github.com/vmware/clarity/pull/4913) We came into the conclusion `trackBy` set on anything else than the `clr-dg-row` generation loop is just not needed.

[...]

PR #4913, opened 2020-08-17
fix(datagrid): trackBy items by provided trackBy option

bbogdanov:
When we have two for loops in a single `<clr-datagrid>` but one of them has a `trackBy` in simple `ngFor` and another one is inside `*clrDgItems` `ClrDatagridItemsTrackBy` selector only selects the `trackBy` of the simple `ngFor` because this is what we told Angular to search for and Angular never went searching for `clrDgItemsTrackBy` directive which represents the `trackBy` in `clrDgItems`. 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3936

## What is the new behavior?

The correct trackBy function is applied when using `*clrDgItems` 

- [ ] Yes
- [x] No

### DIFF
--- a/src/clr-angular/data/datagrid/datagrid-items-trackby.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-items-trackby.spec.ts
@@ -15,7 +15,7 @@ import { Sort } from './providers/sort';
 import { StateDebouncer } from './providers/state-debouncer.provider';
 
 export default function(): void {
-  describe('DatagridItemsTrackby directive', function() {
+  describe('DatagridItemsTrackBy directive', function() {
     beforeEach(function() {
       /*
              * Since the DatagridItems element is a template that isn't rendered in the DOM,
@@ -29,6 +29,8 @@ export default function(): void {
       this.fixture = TestBed.createComponent(FullTest);
       this.fixture.detectChanges();
       this.testComponent = this.fixture.componentInstance;
+      this.testComponent.trackBy = (index: number, item: number) => index;
+      this.testComponent.trackByNgFor = (index: number, item: number) => index;
       this.itemsProvider = TestBed.get(Items);
     });
 
@@ -38,18 +40,29 @@ export default function(): void {
 
     it('receives an input for the trackBy option', function() {
       expect(this.itemsProvider.trackBy).toBeUndefined();
-      this.testComponent.trackBy = (index: number, item: number) => index;
       this.fixture.detectChanges();
       expect(this.itemsProvider.trackBy).toBe(this.testComponent.trackBy);
+    });
+
+    it("doesn't take the ngFor trackBy option", function() {
+      expect(this.itemsProvider.trackBy).toBeUndefined();
+      this.fixture.detectChanges();
+      expect(this.itemsProvider.trackBy).not.toBe(this.testComponent.trackByNgFor);
     });
   });
 }
 
-@Component({ template: `<div *ngFor="let n of numbers; trackBy: trackBy">{{n}}</div>` })
+@Component({
+  template: `
+  <div *ngFor="let n of numbers; trackBy: trackByNgFor">{{n}}</div>
+  <div *clrDgItems="let n of numbers; trackBy: trackBy">{{n}}</div>
+`,
+})
 class FullTest {
   @ViewChild(ClrDatagridItems) datagridItems: ClrDatagridItems<number>;
 
   numbers = [1, 2, 3, 4, 5];
 
   trackBy: (index: number, item: number) => any;
+  trackByNgFor: (index: number, item: number) => any;
 }
--- a/src/clr-angular/data/datagrid/datagrid-items-trackby.spec.ts
+++ b/src/clr-angular/data/datagrid/datagrid-items-trackby.spec.ts
@@ -15,7 +15,7 @@ import { Sort } from './providers/sort';
 import { StateDebouncer } from './providers/state-debouncer.provider';
 
 export default function(): void {
-  describe('DatagridItemsTrackBy directive', function() {
+  describe('DatagridItemsTrackby directive', function() {
     beforeEach(function() {
       /*
              * Since the DatagridItems element is a template that isn't rendered in the DOM,
@@ -29,8 +29,6 @@ export default function(): void {
       this.fixture = TestBed.createComponent(FullTest);
       this.fixture.detectChanges();
       this.testComponent = this.fixture.componentInstance;
-      this.testComponent.trackBy = (index: number, item: number) => index;
-      this.testComponent.trackByNgFor = (index: number, item: number) => index;
       this.itemsProvider = TestBed.get(Items);
     });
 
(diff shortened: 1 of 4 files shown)


Jinnie:
It would be nice to have some tests with the specific reported setup, with two loops with trackby, and make sure there is no interference both ways.

bdryanovski:
I also think that some tests could be helpful here ..

gnomeontherun:
I'm not sure this won't break the ngFor support for trackby. Please make sure we have a way to test both clrDgItems and ngFor use cases and that we pick up the right trackBy.

Jinnie (approved):
LGTM

bdryanovski (approved):

gnomeontherun:
Also we should first make this PR against `master` and then work on backports.

bbogdanov:
The issue and its example were quite misleading so it took me more time to figure out what's happening.

The `ClrDatagridItemsTrackBy` actually never goes to `clrDgItems` to take the provided `trackBy`. It will get only `ngFor` and the last declared `trackBy` will be assigned to `Items`. it works Many(Direcvites) -> One(grid items) and the last one wins.

`Items` directive needs to get the `trackBy` function in order the [Selection<T>](https://github.com/vmware/clarity/blob/881d93b39dd530f9d4afa1fea76d4a99f59f4296/packages/angular/projects/clr-angular/src/data/datagrid/providers/selection.ts) directive to use it to track/select the correct items.

I think this is the only thing we need to fix here from our side related to `trackBy` in `Datagrid`. Anything else that provides `trackBy` inside the datagrid I think is a sign of poor implementation and needs a redesign.

Another solution that comes into my mind rn will be to remove the `ClrDatagridItemsTrackBy` directive and provide an `@Input` in `ClrDatagrid` component with `trackBy` function for the items. That way we will get rid of Many -> One relation and let the users use as many `ngFor` and `trackBy` as they want in the datagrid.

gnomeontherun:
This looks like a better solution here, but adding @hippee-lee as a reviewer to double check this since he made the change to the iterable proxy setup (granted this was a while ago). 

We used to do this, but it was refactored and seems to mean we don't capture trackby properly. I think this fix is good, but wanted to check on side effects. https://github.com/vmware/clarity/blame/f6a4874497655fbf6859f8a1417dc92d1e48f984/src/clr-angular/data/datagrid/datagrid-items.ts#L51

hippee-lee (approved):
It has been a while. I think this is better than adding a new input to the datagrid. I recall that the tests for mutating the array were a warning indicator  that something wasn't working correctly with the trackby fn provided. From waht I recall, as long as those tests are still passing this is the best fix for the issue linked to this PR.

Shijir:
@bbogdanov after some investigation, I don't think this PR is solving #3936.

The trackyBy function on columns is being mistaken for the row items' trackby. So the culprit is in the assignment we make in `datagrid-items.tracky.ts` where we first instantiate and overload Angular’s default `[ngForTrackBy]`. That means any `trackyBy` function whether it’s on row items or some other datagrid component element will be used for row items `trackBy`. So we have to do that assignment of `this._items.trackBy = value` only after we make sure that the `trackyBy` is being used on row items. But that’s really hard… I just don’t see any way to check and make sure ngForTrackBy is being used on rows at this moment. 

The workaround to the issue is not to use `ngForTrackBy` anywhere except on row items AFAIS. Here, our first mistake was to overload a default Angular directive and assumed it would be used only on row items. As Matt suggested above, we could add another `@input` in `clr-dg-row` for its `trackBy` function.

bbogdanov:
@Shijir Actually this will solve the issue because `clrDgItems` will apply the `trackBy` because it is the last instance. Anyway I am not actually trying to solve the entire situation with `ngForTrackBy` directive, because there is no point in doing that and it is something we would like to avoid supporting. What we will do is to encourage everyone to avoid using trackBy on anything else than `clr-dg-row` or in `clrDgItems` because there is no benefit at all to have a `trackBy` on anything else in the datagrid. 

P.S. Matt is saying actually the opposite of what you thought he is saying.

Shijir (commented):
@bbogdanov yes, that will be the case only if a user is using `clrDgItems`, and the `ngForTrackBy` on rows happens to be the last one in Datagrid. But as I said, the culprit is in overloading the default Angular directive and not fixed by this PR. Without fixing the culprit, we are only addressing that specific user's case. You can see that other cases where ngFor and ngForTrackBy are being used on actionable buttons. Or if these directives are used in row detail or detail pane, etc., we will still see similar issues.

Sorry, I misread what Matt wrote. But the point is that we don't have to overload `[ngForTrackBy]` to accept the trackBy function; it could be simply through `trackBy` input on `clr-dg-row`.

bbogdanov:
@Shijir From your last comment, I understand that you would like to remove the `ClrDatagridItemsTrackBy ` directive. I would like to do that too but it is a breaking change. I suggest we at least give the users a major release period of time to change their approaches from using `ngFor` to `clrDgItems` and then remove the directive. 

About the input on `clr-dg-row` I think it getting back to the same approach we have with the directive. We will face the same issue. The last `clr-dg-row` with a `trackBy` will win. (If I understand you correctly)

Jinnie:
Not overloading the directive is the right solution for the long term, but this will require us to discard the use of ngFor, which will be a big breaking change. It will be great solution, too, as using clrDgItems vs ngFor is a common source of confusion. Nevertheless, we have this as tech debt and I don't think it's the right time for this at this point of the product lifecycle. I hope we won't have the same problem in Core.
Also, having trackby in ngFor used for columns is useless - how many items do you have there, why would you want to update them dynamically and preserve state, what state? Hhaving it in action buttons - the same.
In internal ngFors is also a sign of bad architecture and design, though we may see it occasionally. It would be great to be able to avoid it on the Clarity side, but I don't think it's critical and worth the heavy lifting effort at this moment.
In the same time, having clrDgItems silently drop the trackby is a trap that got several devs into and required extensive research and debugging. It's an outstanding problem and its fix is not breaking change.

So, my suggestion is to split the overload scenario from the current case and file new issue for the overload, so we can plan it for some major update. I confirm my approval review for the currently presented commit.

Shijir:
> Not overloading the directive is the right solution for the long term, but this will require us to discard the use of ngFor

I think we should try to remove the overloaded instance of `ngForTrackBy` in one of the next major releases. But I don't quite understand how that will force us to discard the use of `ngFor`.

> Also, having trackby in ngFor used for columns is useless

Yes, I agree! Because of that, I thought, we should just recommend not to use `ngForTrackyBy` on anywhere except exactly 
 on rows until we solve the root cause of the issue once and for all.

> my suggestion is to split the overload scenario from the current case and file new issue (or discussion) for the overload, so we can plan it for some major update

The dilemma I am facing here is that this PR solves just one symptom of the issue. But at the same time as we haven't addressed the root cause, we will see other reports of similar or related symptoms. Actually, in the comment section of the issue, we've already received them.

I agree that it's probably better to separate the overload issue from the case scenario this PR is solving. So I will just remove my block on this PR. But we also need to file another issue to track for the overloaded `ngForTrackBy`.

Shijir (approved):

bbogdanov:
#5044

gnomeontherun (approved):
Thanks all, glad we got a consensus.

PR #4914, opened 2020-08-17
docs: core dev contributions

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
This breaks up the core contribution guidelines into smaller markdown files that storybook can use. Storybook makes it a bit easier to organize share but if there is a strong reason against it I can put it back into a single file. I also added a few new sections on reasons why certain things are built/created/organized. There are a few todos in there for additional content I'd like to add in later follow up PRs.

If there is anything incorrect or I forgot please let me know. Happy to polish this up as much as needed.

- [ ] Yes
- [x] No


packages/core/docs/development/templates-and-styles.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+```scss
+// cds-badge
+:host {
+  // public style API
+  --background: #{$clr-color-neutral-600};
+  --color: #{$clr-color-on-neutral-600};
+}
+
+// Styles other than the default display should not be applied directly to the
+// `:host` selector. This ensures only customizations only done through
```
mathisscott, line 75: nit:

`This ensures that customizations are only done through...`

packages/core/docs/development/using-icons.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+```typescript
+import { ClarityIcons } from '@clr/core/icon/icon.service.js';
+import { successStandardIcon } from '@clr/core/icon/shapes/success-standard.js';
+```
+
+By importing the direct path, we bypass the `index .js` file to prevent
+a large module resolution when using native ESM. This is an internal-only
+optimization, and users of the icons should continue to use the root `@clr/core/icon` import.
+
+We may be able to change this in the future but as its a possibly a breaking change
```
mathisscott, line 36: nit:

`it's` not `its`

mathisscott: also wording:

`as it's possibly a...`

mathisscott:
@coryrylan 
Do we want to get this in for the v4 release? Please add the v4 label and let @gnomeontherun know if so.

Issue #4872, opened 2020-08-05
Fixed-height datagrid header obscures the first row if the header is bigger


hristozov:
## Describe the bug

When the datagrid header is bigger (e.g. a column caption is wrapped) on a fixed-height datagrid, it obscures the first row.

## How to reproduce

1.  Go to https://clarity.design/documentation/datagrid/fixed-height
2. Shrink the "Creation date" column enough to wrap "date" on the next line.
3. The first row is partially obscured.

<img width="897" alt="Screenshot 2020-08-05 at 17 58 24" src="https://user-images.githubusercontent.com/197483/89428643-4c3cf500-d745-11ea-83ca-812b628598e6.png">

## Expected behavior

The first row should be fully visible. Note this is the behavior in at least 3.1.3 and earlier.

## Versions

**App**

- Angular: 9.1.9
- Node: 10.x
- Clarity: 3.1.5

**Device:**

- Type: Macbook
- OS: Mojave
- Browser: Firefox 79 / Chrome 84

## Additional notes

This is a regression I found when upgrading from 3.1.3 to 3.1.5.

whizkidwwe1217:
I'm also having the same issue. It happens when the column header text overflows.
![image](https://user-images.githubusercontent.com/1977391/89483095-e0529f00-d7cd-11ea-891e-3b330ebc664a.png)

bbogdanov:
#4886 seems to fix particular cases. We will proceed with a different fix, cause the current one breaks the design in some cases when resizing the column.

@hristozov #4886 might be helpful until we do a better fix.

Jinnie:
A commit in the v4 branch and release in v4 is needed before we can close this.

hristozov:
Any plans for a minor 3.x update with this fix?

mathisscott:
@hristozov 
Yep. We're working on that.

PR #4919, opened 2020-08-18
fix(datagrid): header visually goes over the data

bbogdanov:
When there are two or more words in column title and the column size gets smaller the text starts to wrap. When the text starts to wrap, it forces the header to shrinks vertically and goes over the data in the grid.

The `min-height` with `position:sticky` make the .datagrid-header to keep its height and even when the content wraps the `height` of the .datagrid-header stays the same. That's why the header content goes over the data.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4872

## What is the new behavior?

The header of datagrid doesn't hide some of the data based on column headers content size.

More 

- [ ] Yes
- [x] No

## Additional

Before
image

After
image

### DIFF
--- a/src/clr-angular/data/datagrid/_datagrid.clarity.scss
+++ b/src/clr-angular/data/datagrid/_datagrid.clarity.scss
@@ -600,7 +600,6 @@
     // bug(popover): prevents action-overflow from being on top (first row).
     // Needed to keep select/radio and expand svgs underneath header on scrolling
     z-index: map-get($clr-layers, datagrid-header);
-    min-height: $clr_baselineRem_1_5;
     width: auto;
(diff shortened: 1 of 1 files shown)


bbogdanov:
> hmm interesting. I think these styles in `.datagrid-header` need to be revisited. I am specifically suspicious of why `position: sticky` is used here.
> 
> These styles could be hack used to fix that action-overflow issue in Datagrid as the comment suggest. Have you checked your solution in Datagrid with action-overflow?

After some scrolling and opening the actions.
image

LGTM

PR #4949, opened 2020-08-26
refactor(schematics): update dependencies, add unit tests, stop using fs for file updates 

sis0k0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

- Adds unit tests for the `ng-add` schematic
- Adds VSCode setup for debugging the package
- Updates to the latest Angular packages
- Stops using `fs` for updating files
- Cleans up the AST utilities and reuses some of the utils exported from `@angular-devkit`

- [ ] Yes
- [X] No


packages/schematics/.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
bdryanovski, line 1: AS this seems helpful is creating one small issue every time that someone wants to change or store some local settings for the project will trigger git change that can't be committed by mistake or not suitable for everyone. 

We have the `.vscode` folder already tracked that are creating a lot of thous issues mentioned above.

So I'm ok to discuss it ... but I think that editor specific configuration doesn't have to be part of the repository. 

A better solution will be to have a section into the documentation for improving the development experience for people but not force them to use it.

packages/schematics/.vscode/tasks.json
```diff
@@ -0,0 +1,13 @@
+{
```
bdryanovski, line 1: Same as `.vscode/launch.json`

packages/schematics/package.json
```diff
@@ -12,9 +12,11 @@
   },
   "scripts": {
     "build": "npm-run-all schematics:build:ts schematics:build:collection",
+    "build:test": "tsc -p tsconfig.test.json",
+    "clean": "del src/**/*.{js,js.map}",
     "schematics:build:ts": "tsc -p tsconfig.json",
     "schematics:build:collection": "cpy --parents --cwd='src/' '**/*.json' ../../../dist/clr-angular/schematics",
-    "test": "yarn build && jasmine **/*_spec.js"
+    "test": "npm run build:test && jasmine **/*_spec.js"
```
bdryanovski, line 19: Can we run `yarn run build:test` or there is some other reason?

packages/schematics/src/add/index.ts
```diff
@@ -4,22 +4,24 @@
...
  */
 
-// @TODO REWRITE THIS
-// This is a mess, because it was written early on and could be much better.
-// Also, no real docs existed then or now, so good luck brave soldier.
-
-import { normalize, Path } from '@angular-devkit/core';
 import { chain, Rule, SchematicContext, SchematicsException, Tree } from '@angular-devkit/schematics';
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import * as fs from 'fs';
```
bdryanovski, line 9:
```ts
import { existsSync } from 'fs';
```

Later on, we could use it like that

```ts
if (existsSync(join(__dirname, '../../package.json'))) {
  // ...
}
```

packages/schematics/src/add/index.ts
```diff
@@ -28,180 +30,138 @@ if (fs.existsSync(join(__dirname, '../../package.json'))) {
...
+      '@angular/platform-browser/animations',
+      projectSettings.module
+    );
+}
+
+function addClarityDependencies(host: Tree) {
+  updateJsonFile(host, 'package.json', json => {
+    const version = getVersion(json.dependencies['@angular/core'], corePackage.version);
+
+    json.dependencies['@clr/angular'] = `${version}`;
```
bdryanovski, line 88: Is there a reason to use string literals. From what I'm seeing `getVersion` return already a string?

packages/schematics/src/add/index.ts
```diff
@@ -28,180 +30,138 @@ if (fs.existsSync(join(__dirname, '../../package.json'))) {
...
+    json.dependencies['@clr/ui'] = `${version}`;
+    json.dependencies['@clr/icons'] = `${version}`;
+
+    const packages = Object.keys(json.dependencies);
+    if (!packages.includes('@webcomponents/webcomponentsjs')) {
```
bdryanovski, line 93: We could skip the `packages` assignment we don't use it after the if.

```ts
if (!Object.keys(json.dependencies).includes('@webcomponents/webcomponentsjs')) {
  // ..
}
```

packages/schematics/src/add/index.ts
```diff
@@ -28,180 +30,138 @@ if (fs.existsSync(join(__dirname, '../../package.json'))) {
...
-    if (text === null) {
-      throw new SchematicsException(`File ${modulePath} does not exist.`);
-    }
-    const sourceText = text.toString('utf-8');
-    const source = ts.createSourceFile(modulePath, sourceText, ts.ScriptTarget.Latest, true);
-
-    const declarationChanges = addSymbolToNgModuleMetadata(source, modulePath, imports, moduleName, packageName);
+function addAssetsToConfigFile(host: Tree) {
+  updateJsonFile(host, CONFIG_FILE_NAME, json => {
+    const projects = Object.keys(json.projects);
```
bdryanovski, line 116: Again here `projects` is no longer used so the assignment is not needed

packages/schematics/src/add/index.ts
```diff
@@ -28,180 +30,138 @@ if (fs.existsSync(join(__dirname, '../../package.json'))) {
...
+    const project = projects.find(key => key === projectSettings.project);
 
-    const declarationRecorder = host.beginUpdate(modulePath);
-    for (const change of declarationChanges) {
-      if (change instanceof InsertChange) {
-        declarationRecorder.insertLeft(change.pos, change.toAdd);
-      }
+    if (!project) {
+      console.info(`Could not update CLI config file to add scripts and styles. You'll have to add them manually.`);
```
bdryanovski, line 120: Why don't we use `SchematicsException` here?

Seems that this the end of the road if the project is missing? Or this is the happy path and this is just to say that "OK nothing to do more here"?

Not sure is the change required but it seems like a `info` may look to someone that "the task had done some work and finished successfully "

packages/schematics/src/add/index_spec.ts
```diff
@@ -1,14 +1,116 @@
-import { Tree } from '@angular-devkit/schematics';
-import { SchematicTestRunner } from '@angular-devkit/schematics/testing';
+import { Tree, HostTree, SchematicsException } from '@angular-devkit/schematics';
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
 import * as path from 'path';
```
bdryanovski, line 3:
```ts
import { join } from 'fs';
```

packages/schematics/src/utility/file-not-found-exception.ts
```diff
@@ -0,0 +1,6 @@
+export class FileNotFoundException extends Error {
```
bdryanovski, line 1: I think that there is already an Exception for that out-of-the-box;

https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/core/src/exception/exception.ts#L22

```ts
export class FileDoesNotExistException extends BaseException {
  constructor(path: string) { super(`Path "${path}" does not exist.`); }
}
```

bdryanovski (changes requested):

packages/schematics/src/add/index.ts
hippee-lee: Maybe `warn` is a better method here? Throwing an exception seems a little to heavey here unless we consider this an error that angular.json arrays could not be updated.

packages/schematics/src/add/index_spec.ts
```diff
@@ -1,14 +1,116 @@
...
+      workspaceTree = await setupProject(new UnitTestTree(new HostTree()), runner, PROJECT_NAME);
+    });
+
+    it('should add dependencies to Clarity packages', async () => {
+      const tree = await runner.runSchematicAsync('ng-add', {}, workspaceTree).toPromise();
+
+      const packageJsonPath = '/package.json';
+      expect(tree.files).toContain(packageJsonPath);
+      const packageJson = JSON.parse(getFileContent(tree, packageJsonPath));
+      const { dependencies } = packageJson;
```
hippee-lee, line 37: Could this assign the JSON.parse output directly here and elminiate packageJson?

packages/schematics/src/utility/add-module-import-to-module.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param host {Tree} The source tree.
+ * @param importedModuleName {String} The name of the imported module.
+ * @param importedModulePath {String} The location of the imported module.
+ * @param moduleToImportIn {String} The location of the module to import in.
+ */
+export function addModuleImportToModule(
+  host: Tree,
+  importedModuleName: string,
+  importedModulePath: string,
+  moduleToImportIn?: string
```
hippee-lee, line 17: Why make this optional if we havea  guard on line #20 that throws an exception?

packages/schematics/src/utility/add-module-import-to-module.ts
```diff
@@ -0,0 +1,36 @@
...
+) {
+  if (!moduleToImportIn) {
+    throw new SchematicsException(`Entry module not found.`);
+  }
+
+  const moduleSource = getTypeScriptSourceFile(host, moduleToImportIn);
+  if (!moduleSource) {
+    throw new SchematicsException(`Module not found: ${moduleToImportIn}`);
```
hippee-lee, line 25: Similar question for this guard. If the function arg `moduleSource` is required but not provided by the caller ~couldn't~ shouldn't we depend on `tsc` to fail with a syntax error? 

The reason for asking is that this seems to look like un-necesarry code but I've not tested it locally to make sure.

packages/schematics/src/utility/setup-project.ts
```diff
@@ -0,0 +1,25 @@
+import { UnitTestTree, SchematicTestRunner } from '@angular-devkit/schematics/testing';
+
+export async function setupProject(tree: UnitTestTree, schematicRunner: SchematicTestRunner, name: string) {
+  tree = await schematicRunner
+    .runExternalSchematicAsync('@schematics/angular', 'workspace', {
+      name: 'workspace',
+      version: '10.0.0',
```
hippee-lee, line 7: It looks like we are hard coding the Angular version here. Does that mean it will need to be updated each time we upgrade the Anger version?

packages/schematics/src/utility/setup-project.ts
```diff
@@ -0,0 +1,25 @@
...
+      'application',
+      {
+        name,
+        projectRoot: '',
+      },
+      tree
+    )
+    .toPromise();
+
+  return tree;
```
hippee-lee, line 24: Should the function have a `UnitTestTree` return type ?

packages/schematics/tsconfig.json
```diff
@@ -9,8 +9,6 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitAny": true,
     "noImplicitThis": true,
-    "noUnusedParameters": true,
```
hippee-lee, line 12: Why were these removed?

packages/schematics/.vscode/tasks.json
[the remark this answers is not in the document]
bdryanovski: More or less you are right Matt about the vscode. The issue that I see is that some plugins sometimes try to write there  settings inside the `.vscode` folder. 

I have for example plugin that color every project in different colors. It stores this inside `.vscode` folder so it will trigger a change. Sometimes even small tweaks in `.vscode` will result in some property inside the folder to change. And this is a conflict between developers. 

I'm ok with `tasks.json` and `launch.json` files to exist but inside a documentation file and people could create and modify them as they wish if needed.

packages/schematics/src/add/index.ts
bdryanovski: I'm ok with a `warn` just the message seems more like "Oops, this task could not finish successfully" .

But `warn` is fine ...

packages/schematics/.vscode/launch.json
sis0k0: I see what's the issue with tracking these configuration files. I'll document the setup instead.

packages/schematics/src/add/index.ts
sis0k0: I'll switch the `info` for `warn`.

packages/schematics/src/utility/add-module-import-to-module.ts
sis0k0: `moduleSource` might be an empty string

packages/schematics/tsconfig.json
sis0k0: I've removed them during the development process. I'll revert.

packages/schematics/src/utility/setup-project.ts
sis0k0: I think we won't have to bump the version unless there are drastic changes in the  Angular project structure. Regardless, I'll check for other options.

sis0k0: I've just checked the Angular schematics tests. The version there's hardcoded to `6.0`.
https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/component/index_spec.ts#L38
https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/service/index_spec.ts#L27

bdryanovski (approved):
😃

PR #4951, opened 2020-08-26
fix(form): display the correct error when validating

### LINKED ISSUE #4929: [Clarity v4] All clr-control-errors shown when form control is invalid
## Describe the bug

When using multiple validators on a form control with separate `clr-control-error *clrIfError` elements, all error messages are shown when the control is invalid

## How to reproduce

[Link to Stackblitz](https://stackblitz.com/edit/clarity-v4-multiple-errors)

An input with 3 different validations - required, maxlength and pattern

1.  Click on the Required input
2.  Focus out (to trigger the required validation)
3.  See all three control errors show up

## Expected behavior

Only the **This is a required field** error is displayed

## Versions

**App**

- Angular: 10
- Clarity: 4.0.0

**Device:**

- Type: MacBook
- OS: macOS
- Browser: chrome, firefox, safari

## Additional notes

The three validation messages disappear together when they are all satisfied - with value 'aa' for the stackblitz example

bdryanovski:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4929 

- [ ] Yes
- [x] No

## Other information

Close #4929

### DIFF
--- a/packages/angular/projects/clr-angular/src/forms/common/if-control-state/if-error.ts
+++ b/packages/angular/projects/clr-angular/src/forms/common/if-control-state/if-error.ts
@@ -28,17 +28,24 @@ export class ClrIfError extends AbstractIfState {
    * @param state CONTROL_STATE
    */
   protected handleState(state: CONTROL_STATE) {
-    const isInvalid = CONTROL_STATE.INVALID === state;
-
-    if (isInvalid && this.displayedContent === false) {
-      let options = {};
-      if (this.error && this.control && this.control.hasError(this.error)) {
-        options = { error: this.control.getError(this.error) };
-      }
-      this.container.createEmbeddedView(this.template, options);
-    } else if (!isInvalid && this.container) {
-      this.container.clear();
+    if (this.error && this.control) {
+      this.displayError(this.control.hasError(this.error));
+    } else {
+      this.displayError(CONTROL_STATE.INVALID === state);
     }
-    this.displayedContent = isInvalid;
+  }
+
+  private displayError(invalid: boolean) {
+    /* if no container do nothing */
+    if (!this.container) {
+      return;
+    }
+    if (invalid && this.displayedContent === false) {
+      this.container.createEmbeddedView(this.template, { error: this.control.getError(this.error) });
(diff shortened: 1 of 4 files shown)


packages/angular/projects/clr-angular/src/forms/common/if-control-state/if-error.ts
```diff
@@ -28,17 +28,20 @@ export class ClrIfError extends AbstractIfState {
...
 
-    if (isInvalid && this.displayedContent === false) {
-      let options = {};
-      if (this.error && this.control && this.control.hasError(this.error)) {
-        options = { error: this.control.getError(this.error) };
-      }
-      this.container.createEmbeddedView(this.template, options);
-    } else if (!isInvalid && this.container) {
+  private displayError(invalid: boolean) {
+    if (invalid && this.displayedContent === false) {
```
gnomeontherun, line 39: Do we need to do a check for the container before trying to create a view?

bdryanovski: I don't think so. At least I could not trigger an error for that.

In the case above, `this.container` is the `host` container so it exists.
In the previews implementation, the `displayError` was call in the constructor stage so there may be cases where the `host` is not ready. But this is not the case anymore.

I try a few different combinations and did not manage to trigger the `TypeError: Cannot read property`.
From what I suspect the `else if` check for the container is for cases when the directive is destroyed and at the same time a `displayError` is called (because it's a callback and could be called async) to prevent an unnecessary code execution. 

I could put the additional check to be double sure?

gnomeontherun: I think it would be useful to bail out if container is not present, just to avoid lifecycle issues and simplify the logic.

noce2:
@gnomeontherun any chance you'd be able to re-review this please? One of my projects is on hold while this fix is being worked on 😅 .

gnomeontherun (approved):

hippee-lee (approved):

PR #4962, opened 2020-08-26
fix: build script for react wrapper

jeeyun:
- updated react package to use built core file like demo apps
- updated the parcel based dev app to use cache
- updated react package's @clr/core dependency to point to locally built one

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/package.js
+++ b/packages/react/package.js
@@ -42,7 +42,8 @@ read('./dist/react')
 
     if (file === 'dist/react/package.json') {
       ['alias', 'browserslist', 'scripts', 'devDependencies'].forEach(p => delete packageFile[p]);
-      packageFile.peerDependencies = { '@clr/core': packageFile.dependencies['@clr/core'] };
+      // move @clr/core from dependencies to peerDependencies and assign version number that matches the package.json
+      packageFile.peerDependencies = { '@clr/core': `^${packageFile.version}` };
       delete packageFile.dependencies['@clr/core'];
     }
 
--- a/packages/react/tsconfig.json
+++ b/packages/react/tsconfig.json
@@ -26,9 +26,6 @@
     "importHelpers": true,
     "composite": true,
     "baseUrl": "./",
-    "rootDir": "./src",
-    "paths": {
-      "@clr/core/*": ["../../dist/core/*"] // This mapping is relative to "baseUrl"
-    }
+    "rootDir": "./src"
   }
 }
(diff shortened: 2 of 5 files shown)


packages/react/package.js
```diff
@@ -42,7 +42,8 @@ read('./dist/react')
 
     if (file === 'dist/react/package.json') {
       ['alias', 'browserslist', 'scripts', 'devDependencies'].forEach(p => delete packageFile[p]);
-      packageFile.peerDependencies = { '@clr/core': packageFile.dependencies['@clr/core'] };
+      // move @clr/core from dependencies to peerDependencies and assign version number that matches the package.json
+      packageFile.peerDependencies = { '@clr/core': packageFile.version };
```
jeeyun, line 46: how strict should the version be for `@clr/core` here? Currently I have it to match exactly.

The above script will copy the version number in its package.json. For example, if the version number is 4.0.1 for this package it will list:

```
  "peerDependencies": {
    "@clr/core": "4.0.1"
  },
```

packages/react/tsconfig.json
```diff
@@ -26,9 +26,6 @@
     "importHelpers": true,
     "composite": true,
     "baseUrl": "./",
-    "rootDir": "./src",
-    "paths": {
-      "@clr/core/*": ["../../dist/core/*"] // This mapping is relative to "baseUrl"
-    }
```
jeeyun, line 32: this `paths` is not needed anymore since we're copying over the `@clr/core` to this package's `node_modules`.

packages/react/package.js
coryrylan: As long as we are not breaking semver we could probably do something like

```javascript
{ '@clr/core': `^${packageFile.version}` }
```

packages/react/tsconfig.json
coryrylan: It looks like the tests are failing because it cant find the package. I think its because of this in the `jest.config.js`
I dont think thats needed anymore and will fix the tests.

```
  moduleNameMapper: {
    '@clr/(.*)': '<rootDir>/../core/dist/$1',
  },
```

Also it looks like there is a similar path mapping in the `tsconfig.build.json` file that probably be removed as well.

packages/react/package.js
jeeyun: I think that's reasonable. will update.

packages/react/tsconfig.json
jeeyun: Just a note: I removed them from both `tsconfig.json` and `tsconfig.build.json` but kept the one in jest config. Removing it from jest config caused the build to fail even though test runs fine locally. Local env might be relying on a pre-condition that build isn't replicating.

PR #4996, opened 2020-09-02
fix: datagrid-row correct way of assigning clrDgSelectable

### LINKED ISSUE #4993: [clrDgSelectable]="false" in clr-grid does not disable select in v-4
## Describe the bug

using [clrDgSelectable]="false" in clr-grid does not disable select in v-4

## How to reproduce

https://stackblitz.com/edit/clarity-v4-light-theme-k7dk5i?file=src%2Fapp%2Fapp.component.html

## Versions

**App**

- Angular: angular 10
- Node: [e.g. 8.10.0]
- Clarity: 4.0

**Device:**

- Type: any device
- OS: any oc
- Browser any

## Additional notes

bdryanovski:
Same as `colDgType`, `clrDgSelectable` is called in some cases before `clrDgItem` is defined so there is no way to lock it or unlock it. 

Moving this inside `ngOnChanges` and checking it there solve this. 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4993

- [ ] Yes
- [x] No

## Other information

Close #4993

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.spec.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.spec.ts
@@ -111,7 +111,41 @@ export default function (): void {
         expect(context.clarityDirective.selected).toBe(true);
       });
 
-      it('should be able you toggle the state of selection when clrDgSelectable is false', () => {
+      it('should be able to toggle selection through API when clrDgSelectable is false', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.clarityDirective.toggle(true);
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(false);
+      });
+    });
+
+    describe('Conditional selection switch order or arguments', function () {
+      let context: TestContext<ClrDatagridRow, SelectableRowOrder>;
+      let selectionProvider: Selection;
+      let checkbox: HTMLElement;
+
+      beforeEach(function () {
+        context = this.create(ClrDatagridRow, SelectableRowOrder, DATAGRID_SPEC_PROVIDERS);
+        selectionProvider = TestBed.get(Selection);
+        TestBed.get(Items).all = [{ id: 1 }, { id: 2 }];
+      });
+
+      it('should toggle when clrDgSelectable is false for type  SelectionType.Multi', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        checkbox = context.clarityElement.querySelector("input[type='checkbox']");
+
+        expect(checkbox.getAttribute('disabled')).toBe('true');
+        expect(checkbox.getAttribute('aria-disabled')).toBe('true');
+
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(true);
+      });
+
+      it('should be able to toggle selection through API when clrDgSelectable is false', () => {
         selectionProvider.selectionType = SelectionType.Multi;
         context.clarityDirective.toggle(true);
         context.testComponent.clrDgSelectable = false;
@@ -452,10 +486,18 @@ export default function (): void {
 class ProjectionTest {}
 
 @Component({
-  template: `<clr-dg-row [clrDgSelectable]="clrDgSelectable" [clrDgItem]="item">None</clr-dg-row>`,
+  template: `<clr-dg-row [clrDgItem]="item" [clrDgSelectable]="clrDgSelectable">None</clr-dg-row>`,
 })
 class SelectableRow {
   clrDgSelectable = true;
+  item: Item = { id: 42 };
+}
+
+@Component({
+  template: `<clr-dg-row [clrDgSelectable]="clrDgSelectable" [clrDgItem]="item">None</clr-dg-row>`,
+})
+class SelectableRowOrder {
+  clrDgSelectable = undefined;
   item: Item = { id: 42 };
 }
 
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
@@ -132,7 +132,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit 
     }
   }
 
-  // By default every item is selectable
+  // By default every item is selectable; it becomes not selectable only if it's explicitly set to false
   @Input('clrDgSelectable')
   public set clrDgSelectable(value: boolean) {
     this.selection.lockItem(this.item, value === false);
@@ -261,6 +261,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit 
 
   ngOnInit() {
     this.wrappedInjector = new HostWrapper(WrappedRow, this.vcr);
+    this.selection.lockItem(this.item, this.clrDgSelectable === false);
   }
 
   public get _view() {
--- a/packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
+++ b/packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
@@ -6,6 +6,48 @@
 
 <h2>Datagrid: Conditional selection</h2>
 
+<h3>Lock the how table</h3>
+
+<button class="btn btn-primary" (click)="canItBeSelected = !canItBeSelected">
+  {{ canItBeSelected ? 'Lock' : 'UnLock' }} Table
+</button>
+
+<clr-datagrid [(clrDgSingleSelected)]="selected">
+  <clr-dg-column>User ID</clr-dg-column>
+  <clr-dg-column>Name</clr-dg-column>
+  <clr-dg-column>Creation date</clr-dg-column>
+  <clr-dg-column>Favorite color</clr-dg-column>
+
+  <clr-dg-row [clrDgSelectable]="canItBeSelected" *clrDgItems="let user of users" [clrDgItem]="user">
+    <clr-dg-cell>{{user.id}}</clr-dg-cell>
+    <clr-dg-cell>{{user.name}}</clr-dg-cell>
+    <clr-dg-cell>{{user.creation | date}}</clr-dg-cell>
+    <clr-dg-cell>
+      <span class="color-square" [style.backgroundColor]="user.color"></span>
+    </clr-dg-cell>
+  </clr-dg-row>
+
+  <clr-dg-footer>{{users.length}} users</clr-dg-footer>
+</clr-datagrid>
+
+<clr-datagrid [(clrDgSelected)]="selectedRows">
+  <clr-dg-column>User ID</clr-dg-column>
+  <clr-dg-column>Name</clr-dg-column>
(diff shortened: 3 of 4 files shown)


packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
```diff
@@ -248,6 +247,15 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
     this.subscriptions.forEach((sub: Subscription) => sub.unsubscribe());
   }
 
+  ngOnChanges(changes: any) {
+    /**
+     * When `clrDgSelectable` make sure  to lock or unlock the given row
+     */
+    if (changes.clrDgSelectable && changes.clrDgSelectable.currentValue !== changes.clrDgSelectable.previousValue) {
```
gnomeontherun, line 254: Ok this is going to run _every change detection cycle_ for _every row_. Can we avoid this with some other approach?

bdryanovski: Are we talking about the `if`  or the body of the `if`? 

But you are right it's called for every row, at least once. `ngOnChanges` is called after all `inputs` are updated so the good thing is that won't run for every input update.

It's not that different than having it inside the `@Input` getter. 

I could think of something else as a way to monitor the change of  `clrDgSelectable` but right now I don't have any idea :) 

Open for suggestions.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
```diff
@@ -133,12 +135,9 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
   }
 
   // By default every item is selectable
-  @Input('clrDgSelectable')
-  public set clrDgSelectable(value: boolean) {
-    this.selection.lockItem(this.item, value === false);
-  }
+  @Input('clrDgSelectable') clrDgSelectable: boolean;
```
jeeyun, line 138: If every item is selectable by default, should this be set to `true` by default?

bdryanovski: The reason for not adding a default value is that - we only care if it's `false` and everything else is `true` for us. But I could change it it's an easy one ;)

jeeyun: Yeah, I guess logically it makes no difference. =) It was a bit harder to parse that mentally upon initial glance. In some parts of our code I think we rely on falsy value as equivalent of setting it to false (which is not true in this case).

Maybe we don't need to change, and just edit the comment?
```
// By default every item is selectable; it becomes not selectable only if it's explicitly set to false
```

bdryanovski:
Can I get some review on that @gnomeontherun , @jeeyun, @Shijir  :)

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
Shijir, line 254: The issue we are trying to solve was specifically in the **initial** input setting order.  And `ngOnInit` is what is there for us to coordinate the input properties because it runs once after all input changes.

So the fix could have been simply adding the following one line in the `ngOnInit` cycle instead of doing checks on every input value change:
```ts
this.selection.lockItem(this.item, !this.clrDgSelectable);
```

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
```diff
@@ -132,13 +132,13 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
...
 
-  // By default every item is selectable
+  // By default every item is selectable; it becomes not selectable only if it's explicitly set to false
   @Input('clrDgSelectable')
   public set clrDgSelectable(value: boolean) {
     this.selection.lockItem(this.item, value === false);
   }
 
-  public get clrDgSelectable() {
+  public get isSelectable() {
```
Shijir, line 141: I don't think we should change this public property name. It's basically a breaking change.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
```diff
@@ -261,6 +261,7 @@ export class ClrDatagridRow<T = any> implements AfterContentInit, AfterViewInit
 
   ngOnInit() {
     this.wrappedInjector = new HostWrapper(WrappedRow, this.vcr);
+    this.selection.lockItem(this.item, this.clrDgSelectable === false);
```
Shijir, line 264: - You are accessing `this.clrDgSelectable`, but you changed it to `this.selection`. But as I said above, it's better to keep `this.clrDgSelectable` to avoid a breaking change.

- This 2nd point is a nit. Why not `!this.clrDgSelectable` instead of `this.clrDgSelectable === false`? Because as we already know `this.selection.isLocked(this.item)` is the falsy/opposite value of this.clrDgSelectable.

packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
```diff
@@ -6,6 +6,46 @@
...
+
+<button class="btn btn-primary" (click)="locked = !locked || false">Lock Table</button>
+
+<clr-datagrid [(clrDgSingleSelected)]="selected">
+  <clr-dg-column>User ID</clr-dg-column>
+  <clr-dg-column>Name</clr-dg-column>
+  <clr-dg-column>Creation date</clr-dg-column>
+  <clr-dg-column>Favorite color</clr-dg-column>
+
+  <clr-dg-row [clrDgSelectable]="locked" *clrDgItems="let user of users" [clrDgItem]="user">
```
Shijir, line 19: nit: Shouldn't we have `!locked` here? As I mentioned above, `locked` should represent the opposite of `selectable`.

Shijir (changes requested):

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
bdryanovski: I will start from bottom to top ;) 

the `this.clrDgSelectable === false` is for readdability - It's to be clear that only false is important. 

the design of this attribute is a bit misleading from the start - the idea is to disabled from selection - but you need to send it `false` as an argument. So keeping the expression more verbose is to let people calculate the end result faster (there is a comment above it ) 

And for the first one, `the.selection` is the `Selection` service that keeps track of selected element and the one that holds the methods for locking and selecting items. It's not a breaking change.

packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
bdryanovski: This is copy paste of the Stackblitz demo design to test the different order of arguments in the feature this demo could be removed or not - it has no effect of change...

I could rename it from `locked` to `state` and pass `false` or `true`

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
bdryanovski: This is a super internal API that is really hard to access and if it's not used inside the template I will prefer to be private. You as a developer should not have access to it and use the `Selection` and another service to change the state. But this is a limitation of  TypeScript and Angular integration when it comes to accessing variables inside the template (or linter setting)

It's improving a lot the readability when it comes to reading the template (the only place where this is used). Now is clear that if this is `true` then the `ngIf` will be displayed. With the strange behavior of `clrDgSelectable` that was not clear at first. 

So I think that this is a minor unnoticeable change that could improve the code at first. (It's a bad design on my part a year ago naming it like that)

packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
Shijir: No, `state` would make it much vaguer. I think `locked` is just fine. I was pointing out that if `locked` is for `clrDgSelectable `, it should represent the opposite value of itself. Right now, it looks like both `locked` and `clrDgSelectable` are using the same exact value.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.ts
Shijir: No, how do you know it's super internal API? If there are teams and users already making use of this property, we will break their app once they upgrade to our next release. We even prefixed the property with `clr` to indicate it's a public API property.

packages/angular/projects/dev/src/app/datagrid/conditional-selection/conditional-selection.html
bdryanovski: The value of `locked` is fine, it's controlled from the button above and it's designed to be able to lock and unlock the rows by pressing it (the button) for demo purpose. No need to invert it or change it.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.spec.ts
```diff
@@ -121,6 +121,40 @@ export default function (): void {
...
+      });
+
+      it('should be able you toggle the state of selection when clrDgSelectable is false', () => {
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.clarityDirective.toggle(true);
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(false);
+      });
```
Shijir, line 155: This test actually seems to be testing the API part I mentioned in the previous comment. But the description should be improved.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.spec.ts
```diff
@@ -121,6 +121,40 @@ export default function (): void {
...
+        selectionProvider.selectionType = SelectionType.Multi;
+        context.testComponent.clrDgSelectable = false;
+        context.detectChanges();
+        checkbox = context.clarityElement.querySelector("input[type='checkbox']");
+
+        expect(checkbox.getAttribute('disabled')).toBe('true');
+        expect(checkbox.getAttribute('aria-disabled')).toBe('true');
+
+        context.clarityDirective.toggle();
+        expect(context.clarityDirective.selected).toBe(true);
```
Shijir, line 145: Two things here:

1. The description looks off. When `clrDgSelectable` is false, you **shouldn't be able to toggle** through the selection checkbox's UI because it's disabled. I see you've tested whether it's getting `disabled` attributes.

2. The lines I've highlighted here doesn't belong here. These lines are testing the API part. And It looks like you are testing it in the test right below it.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.spec.ts
```diff
@@ -121,6 +121,40 @@ export default function (): void {
...
+      });
+
+      it('should be able you toggle the state of selection when clrDgSelectable is false', () => {
```
Shijir, line 148: nit: should be able to toggle selection through API when clrDgSelectable is false

Shijir (approved):
Approved with small change requests in the tests

jeeyun (approved):

PR #5024, opened 2020-09-09
Allow className to be passed down to the underlying component

astorije:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The [Storybook page about the Icon API](https://clarity.design/storybook/core/?path=/story/components-icon-getting-started--page#api) shows that it should be possible to change the color with `--color`.

In practice, I could not find a way to apply a `className` to the `CdsIcon` React component because the `className` property was being swallowed. [Here is a sandbox](https://codesandbox.io/s/great-swanson-kdb39?file=/src/App.tsx) to demonstrate the issue.

Issue Number: N/A

## What is the new behavior?

After removing this, this would be the expected behavior:

Screen Shot 2020-09-09 at 4 00 05 PM

(I couldn't figure out how to patch dependencies on CodeSandbox)

- [ ] Yes
- [x] Probably?
- [ ] No

## Other information

I'm not sure if this is the right way to do that, and I'm more than happy to edit my PR if there is a better way, but I figured I would share what I tried and what worked in my specific use case.
Hopefully this is just a PEBCAK and there is a simple way to pass down a CSS class that I missed. Let me know how I can help!

### DIFF
--- a/packages/react/src/converter/react-wrapper.test.tsx
+++ b/packages/react/src/converter/react-wrapper.test.tsx
@@ -79,4 +79,16 @@ describe('CdsTestComponent', () => {
     );
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('correctly passes down className to the underlying web component', () => {
+    const wrapper = mount(
+      <div>
+        <CdsTestComponent className="test-classname">Hello World</CdsTestComponent>
+      </div>
+    );
+
+    const { classList } = wrapper.find('ReactWrapperComponent').getDOMNode();
+    expect(classList).toHaveLength(1);
+    expect(classList).toContain('test-classname');
+  });
 });
--- a/packages/react/src/converter/react-wrapper.tsx
+++ b/packages/react/src/converter/react-wrapper.tsx
@@ -44,7 +44,7 @@ export function createReactComponent<BaseComponent extends HTMLElement>(elementN
     }
 
     _propIsReservedReactProp(prop) {
-      const reactProperties = ['children', 'localName', 'ref', 'style', 'className'];
+      const reactProperties = ['children', 'localName', 'ref', 'style'];
       return reactProperties.indexOf(prop) !== -1;
     }
 


jeeyun (approved):

packages/react/src/converter/react-wrapper.tsx
```diff
@@ -44,7 +44,7 @@ export function createReactComponent<BaseComponent extends HTMLElement>(elementN
     }
 
     _propIsReservedReactProp(prop) {
-      const reactProperties = ['children', 'localName', 'ref', 'style', 'className'];
+      const reactProperties = ['children', 'localName', 'ref', 'style'];
```
mathisscott, line 47: I feel like the only thing we are missing is a test here. Both to verify the fix is working as expected as well as to ensure this isn't broken in the future.

mathisscott (changes requested):

packages/react/src/converter/react-wrapper.tsx
astorije: @mathisscott, I added a test, but it's _definitely_ a bad one: it passes before and after my fix. Would you be able to help me figure out a better test case for this?

packages/react/src/converter/react-wrapper.tsx
```diff
@@ -44,7 +44,8 @@ export function createReactComponent<BaseComponent extends HTMLElement>(elementN
     }
 
     _propIsReservedReactProp(prop) {
-      const reactProperties = ['children', 'localName', 'ref', 'style'];
+      const reactProperties = ['children', 'localName', 'ref', 'style', 'className'];
+      // const reactProperties = ['children', 'localName', 'ref', 'style'];
```
hippee-lee, line 48: Is this an extra comment?

astorije: Oops, my bad, `git add -p` gone wrong! I removed it, still not sure what to do about the test case however.

hippee-lee: No worries. Thanks for the quick fix 🥇

packages/react/src/converter/react-wrapper.test.tsx
```diff
@@ -79,4 +79,15 @@ describe('CdsTestComponent', () => {
...
+
+  it('correctly passes down className to the underlying component', () => {
+    const wrapper = shallow(
+      <div>
+        <CdsTestComponent className="test-classname">Hello World</CdsTestComponent>
+      </div>
+    );
+
+    const renderedComponent = wrapper.find('ReactWrapperComponent');
+    expect(renderedComponent.prop('className')).toEqual('test-classname');
```
mathisscott, line 91: I think I would expect the test to look at whether the classname was passed down to the wrapped component. Maybe I'm missing something here? Once the component is rendered, you should be able to look in the DOM and see if the web component has the classname in its classList. I would think that test would fail without the fix.

astorije: That makes total sense, and that worked indeed! I could verify that the test failed without my change, and passed afterwards. Thanks for the pointer and sorry for being slow, it's the first time I work with web components (and Clarity) so still adjusting.

packages/react/src/converter/react-wrapper.tsx
mathisscott: @astorije 
I pulled down your branch and verified locally that applying a classname to the wrapper component does, in fact, update the CSS custom properties as expected inside the web component. My assumption was that this is a big part of the use case for this functionality. So given that I've verified that and we now have a test in place, I think we are 👍

mathisscott (approved):
I think all that we need to do now is rebase on current master, squash the two commits, and push back up. Once that's done, we can merge into the codebase.

Thanks for the contribution, @astorije !

astorije:
@mathisscott, thanks, done!

astorije:
Awesome, thanks @mathisscott!

PR #5031, opened 2020-09-11
docs: add components overview page

bbogdanov:

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

## Other information

Added also the `new` badge 

image

### DIFF
--- a/apps/website/.vuepress/theme/global-components/ComponentsOverview.vue
+++ b/apps/website/.vuepress/theme/global-components/ComponentsOverview.vue
@@ -1,0 +1,12 @@
+<template>
+  <div>
+    <ItemOverviewList :items="componentList" />
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'ComponentsOverview',
+  props: ['componentList'],
+};
+</script>
--- a/apps/website/components/README.md
+++ b/apps/website/components/README.md
@@ -1,1 +1,3 @@
 # Component Overview
+
+<ComponentsOverview :componentList="$site.themeConfig.sidebar.find(x => x.title === 'Components').children" />
--- a/apps/website/.vuepress/theme/global-components/ItemOverview.vue
+++ b/apps/website/.vuepress/theme/global-components/ItemOverview.vue
@@ -1,0 +1,55 @@
+<template>
+  <section class="item-overview">
+    <h2>
+      <a :href="path" :target="target"> {{ title }} </a>
+      <span v-if="isNew" class="badge badge-info">new</span>
+    </h2>
+    <div ro0e="img">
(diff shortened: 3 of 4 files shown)


apps/website/components/README.md
```diff
@@ -1 +1,3 @@
 # Component Overview
+
+<ComponentsOverview :componentList="$site.themeConfig.sidebar.find(x => x.title === 'Components').children" />
```
hippee-lee, line 3: I would prefer to have the function/logic in a computed property so that the template stays simple.

apps/website/.vuepress/theme/global-components/ComponentsOverview.vue
```diff
@@ -0,0 +1,12 @@
+<template>
+  <div>
+    <ItemOverviewList :items="componentList" />
```
hippee-lee, line 3: It looks like this component only exists to pass the list down to the ItemOverview component. Does it make sense to move this up one level and only have the top level overview / list ad the list item template?

apps/website/components/README.md
hippee-lee: I was also thinking that this would have a more general naame, e.g:

```<DocItemsOverview :items="'Components">```

That way it could easily be re-used for 

```<DocItemsOverview :items="'Foundations">```

hippee-lee (changes requested):
Thanks for sarting this. The approach looks like its going in the right direction. I left some ideas to consider. 

imo - this PR should be closed for now. We are still waiting on the final design and this PR isn't ready to be merged into master without that work.

apps/website/.vuepress/theme/global-components/ComponentsOverview.vue
bbogdanov: I think it is a great fit in case we would like to change something when we get the list of the component before passing them to the `ItemOverviewList`

apps/website/components/README.md
bbogdanov: @hippee-lee  The thing with using the plural for me is that it is casual to misread the name and think of it as a single element. When you have the `List` in the end - there is no way you would think this component is responsible to display only a single item. Let me know what do you think.

apps/website/.vuepress/theme/global-components/ComponentsOverview.vue
bbogdanov: After playing around a little more with setting the footer for each component I believe the overview list is not a great fit. I already started to move some of the code into `ComponentOverview` component. Will push soon.

PR #5054, opened 2020-09-16
fix: make datagrid header sticky no matter how many rows

### LINKED ISSUE #4889: Datagrid header row won't stick to the top with fixed height
## Describe the bug

When styling the datagrid with a fixed height and selecting page size of e.g. 100, the header row scrolls along after a while.
 
## How to reproduce

1. Go to clarity's [datagrid component with pagination](https://clarity.design/documentation/datagrid/pagination)
2. Select page size 100
3. Inspect the datagrid DOM element and give it a fixed height of 500px
4. Scroll down

## Expected behavior

Header row stays on top

## Versions

**App**

- Clarity: 3.1.5 

**Device:**

- Type: MacBook
- OS: macOs
- Browser: Chrome
- Version 84.0.4147.89

## Additional notes

I noticed on inspecting that div.datagrid-table-wrapper and div.datagrid-table don't have the full height als the row items, which is why the sticky header row stops sticking. Compare to older datagrid verions, you seem to have added new wrapper elements, maybe it's a styling bug?

image

hippee-lee:
This change fixes a regression that breaks the sticky behavior of datagrid column headers.

Closes #4889

## PR Checklist

## PR Type

There was a regression introduced that broke the sticky column (header row) behavior in datagrid. This reverts that change. 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently sticky headers will scroll out of view when the last visible row overflows the container element. This was happening because the `.datagrid` container had a display: flex on it. 
The code removed indicates that it was added to fix content stretching in Safari but I couldn't see what that issue was when it was removed. 

Issue Number: #4889 

## What is the new behavior?
Datagrid headers stay in view when row content overflows no matter how much overflow there is. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
@@ -40,8 +40,6 @@
     overflow: auto;
     flex: 1 1 auto;
     margin-top: $clr_baselineRem_0_5;
-    // Force content stretch on Safari
-    display: flex;
   }
 
   .datagrid-container {
@@ -645,6 +643,7 @@
   .datagrid-table-wrapper {
     display: flex;
     flex: 1 1 auto;
+    min-height: 100%;
   }
 
   .datagrid-table {
@@ -652,9 +651,7 @@
     flex-direction: column;
     flex: 1 1 auto;
     align-content: flex-start;
-    min-height: $clr_baselineRem_3; // Includes header and space for a single (non existing) row
     position: relative;
-    height: 100%;
 
     .datagrid-body {
       width: auto;


bdryanovski (approved):

Jinnie (changes requested):
This simply unfixes the previous fix and will return the problem with no inner border in "Details" view when there are very few items in the list.
I researched more how we can fix the latter issue, without using flexbox on the parent and came with this solution, in which the flexbox is not needed:
Screen Shot 2020-09-17 at 1 07 14 PM
which should be better, as we're resetting all the parents of the header individually, using only min-width recommendation, not forcing a strict width.
I checked if the datagrid looks OK with 0 and 1 items (because of the commented previous min-width) and it looks good.
It probably has to be tested for other regressions before commit.

Jinnie (commented):
This is an example of the missing border issue I'm talking about. Can be tested at http://localhost:4200/datagrid/detail

Screen Shot 2020-09-17 at 1 31 07 PM

hippee-lee:
Thanks @Jinnie - thats why I added you as a reviewer ... it wasn't clear to me why the flexbox had been added to the `.datagrid` container and I didn't see anything obviously wrong with the dev grids. I was hoping to get your insight to catch anything that wasn't obvious. Fwiw - the sticky header + sticky controls column is a little tricky and depends on the exact nested flex structure that we ended up with so that we can manage it with css (native browser) only. 

This highlights how important it is to get our visual regression testing back. When I refactored the datagrid for sticky header/sticky first (row controls) columns I relied heavily on our old Gemini testing to warn me of the side effects the refactor was having on the visual styles. 

Thanks again for double checking, I'll push changes up soon.

packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
```diff
@@ -645,16 +643,15 @@
...
     flex: 1 1 auto;
+    min-height: 100%;
   }
 
   .datagrid-table {
     display: flex;
     flex-direction: column;
     flex: 1 1 auto;
     align-content: flex-start;
-    min-height: $clr_baselineRem_3; // Includes header and space for a single (non existing) row
```
Jinnie, line 655: It seems `min-height: 100%`, which I was suggesting, is not really needed here, that's good. 
We can probably return this safeguard to `$clr_baselineRem_3`?
I really don't know why it was included, and I don't like removing things like that, unless unavoidable. Or unexpected regressions may appear.

I am approving this, as it's not a blocker, just a precaution. If you decide (or know the exact origin of the safeguard) you can either return it or commit like this. Thanks for the fix and the fix of the fix.

Jinnie (approved):

PR #5100, opened 2020-09-30
feat: core accordion

jeeyun:
Core implementation of the Clarity accordion.

### Notes
- Core accordion doesn't manage state and therefore does not automatically close a panel if another is open, etc. This logic, along with any other logic (e.g. intercept when user expands / closes a panel) can be implemented on the application side
- Slide animation for expand / close (animating from `height: 0` to `height: auto`) is not included in this PR. Our current goal, if possible, is to not use javascript in our animation. If we were to use javascript to grab the content's height there are some challenges with responding to dynamic changes in accordion content's height at runtime. We'll have to revisit the animation part especially in the context of Clarity motion and how they might fit in with the rest of the components in the system.

### Reference links for expand / collapse animation
- https://developers.google.com/web/updates/2017/03/performant-expand-and-collapse
- https://jdsteinbach.com/css/holy-grail-css-animation/
- https://github.com/tweenjs/tween.js/blob/master/dist/tween.esm.js
- https://stackoverflow.com/questions/54520554/custom-element-getrootnode-closest-function-crossing-multiple-parent-shadowd

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/accordion/accordion-content.element.ts
+++ b/packages/core/src/accordion/accordion-content.element.ts
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles } from '@clr/core/internal';
+import { html, LitElement } from 'lit-element';
+
+/**
+ * Web component accordion content to be used inside an accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded>
+ *     <cds-accordion-header>Item 1</cds-accordion-header>
+ *     <cds-accordion-content>Content 1</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel>
+ *     <cds-accordion-header>Item 2</cds-accordion-header>
+ *     <cds-accordion-content>Content 2</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel disabled>
+ *     <cds-accordion-header>Item 3</cds-accordion-header>
+ *     <cds-accordion-content>Content 3</cds-accordion-content>
+ *   </cds-accordion-panel>
+ * </cds-accordion>
+ * ```
+ *
+ * @beta
+ * @element cds-accordion-content
+ */
+export class CdsAccordionContent extends LitElement {
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('slot', 'accordion-content');
+  }
+  render() {
+    return html`<slot></slot>`;
+  }
+
+  static get styles() {
+    return [baseStyles];
+  }
+}
--- a/packages/core/src/accordion/accordion-header.element.ts
+++ b/packages/core/src/accordion/accordion-header.element.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles } from '@clr/core/internal';
+import { html, LitElement } from 'lit-element';
+
+/**
+ * Web component accordion to be used inside an accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded>
+ *     <cds-accordion-header>Item 1</cds-accordion-header>
+ *     <cds-accordion-content>Content 1</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel>
+ *     <cds-accordion-header>Item 2</cds-accordion-header>
+ *     <cds-accordion-content>Content 2</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel disabled>
+ *     <cds-accordion-header>Item 3</cds-accordion-header>
+ *     <cds-accordion-content>Content 3</cds-accordion-content>
+ *   </cds-accordion-panel>
+ * </cds-accordion>
+ * ```
+ *
+ * @beta
+ * @element cds-accordion-header
+ */
+export class CdsAccordionHeader extends LitElement {
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('slot', 'accordion-header');
+  }
+
+  render() {
+    return html`<slot></slot>`;
+  }
+
+  static get styles() {
+    return [baseStyles];
+  }
+}
--- a/packages/core/src/accordion/accordion-panel.element.scss
+++ b/packages/core/src/accordion/accordion-panel.element.scss
@@ -1,0 +1,84 @@
+@import './../styles/tokens/generated/index';
+@import './../styles/mixins/mixins';
+
+:host {
+  --color: #{$cds-token-color-neutral-800};
+  --icon-color: #{$cds-token-color-neutral-700};
+  --font-size: #{$cds-token-typography-font-size-4};
+  --font-weight: #{$cds-token-typography-font-weight-regular};
+  --header-background: #{$cds-token-color-neutral-50};
+  --header-color: #{$cds-token-color-neutral-800};
+  --header-font-size: #{$cds-token-typography-font-size-3};
+  --header-font-weight: #{$cds-token-typography-font-weight-semibold};
+  --header-open-background: #{$cds-token-global-selection-color};
+  --transition-duration: 200ms;
+}
+
+::slotted(cds-accordion-header) {
+  background: var(--header-background);
+  color: var(--header-color);
+  font-size: var(--header-font-size);
+  font-weight: var(--header-font-weight);
+}
+
+::slotted(cds-accordion-content) {
+  color: var(--color);
+  font-size: var(--font-size);
+  font-weight: var(--font-weight);
+}
+
+// todo: set max-height for content
+.accordion-content-wrapper {
+  height: 0;
+  transition: transform height var(--transition-duration);
+  transition-timing-function: cubic-bezier(0, 0.99, 0, 0.99);
+
+  overflow-y: hidden;
+
+  &.open {
+    height: auto;
+    border-top: var(--border-width) solid var(--border-color);
+  }
+}
+
+.accordion-header {
+  background: var(--header-background);
+
+  .accordion-angle {
+    --color: var(--icon-color);
+  }
+
+  .accordion-angle,
+  .accordion-header-button,
+  .accordion-status,
+  ::slotted([slot='accordion-header']) {
+    align-items: center;
+    cursor: pointer;
+  }
+
+  &.disabled {
+    .accordion-angle,
+    .accordion-header-button,
+    .accordion-status,
+    ::slotted([slot='accordion-header']) {
+      cursor: not-allowed;
+    }
+  }
+
+  .accordion-angle-wrapper {
+    transition: transform var(--transition-duration);
+  }
+
+  &.open {
+    --header-background: var(--header-open-background);
+
+    .accordion-angle-wrapper {
+      transform: rotate(90deg);
+    }
+  }
+}
+
+.accordion-header-button {
+  background: transparent;
+  border: 0;
+}
--- a/packages/core/src/accordion/accordion-panel.element.spec.ts
+++ b/packages/core/src/accordion/accordion-panel.element.spec.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { html } from 'lit-html';
+import '@clr/core/accordion/register.js';
+import { CdsAccordionPanel } from '@clr/core/accordion';
+import { componentIsStable, createTestElement, getComponentSlotContent, removeTestElement } from '@clr/core/test/utils';
+
+describe('accordion-panel element', () => {
+  let testElement: HTMLElement;
+  let component: CdsAccordionPanel;
+  const placeholderHeader = 'Accordion Header Placeholder';
+  const placeholderContent = 'Accordion Content Placeholder';
+
+  beforeEach(async () => {
+    testElement = await createTestElement(html`
+      <cds-accordion-panel>
+        <cds-accordion-header>${placeholderHeader}</cds-accordion-header>
+        <cds-accordion-content>${placeholderContent}</cds-accordion-content>
+      </cds-accordion-panel>
+    `);
+    component = testElement.querySelector<CdsAccordionPanel>('cds-accordion-panel');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  it('should create the component', async () => {
+    await componentIsStable(component);
+    const slots = getComponentSlotContent(component);
+    expect(slots['accordion-header']).toBe(
+      `<cds-accordion-header slot="accordion-header"><!---->${placeholderHeader}<!----></cds-accordion-header>`
+    );
+    expect(slots['accordion-content']).toBe(
+      `<cds-accordion-content slot="accordion-content"><!---->${placeholderContent}<!----></cds-accordion-content>`
+    );
+  });
+
+  it('should emit an expandedChange event when header is clicked', async done => {
+    let value: any;
+    await componentIsStable(component);
+    component.addEventListener<any>('expandedChange', (e: CustomEvent) => {
+      value = e.detail;
+      expect(value).toBe(true);
+      done();
+    });
+
+    const button = component.shadowRoot.querySelector<HTMLButtonElement>('.accordion-header-button');
+    expect(button).toBeDefined();
+    button.click();
+  });
+});
--- a/packages/core/src/accordion/accordion-panel.element.ts
+++ b/packages/core/src/accordion/accordion-panel.element.ts
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { html, LitElement } from 'lit-element';
+import { baseStyles, event, property, EventEmitter } from '@clr/core/internal';
+import { styles } from './accordion-panel.element.css.js';
+
+/**
+ * Web component accordion panel to be used inside an accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded>
+ *     <cds-accordion-header>Item 1</cds-accordion-header>
+ *     <cds-accordion-content>Content 1</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel>
+ *     <cds-accordion-header>Item 2</cds-accordion-header>
+ *     <cds-accordion-content>Content 2</cds-accordion-content>
+ *   </cds-accordion-panel>
+ *   <cds-accordion-panel disabled>
+ *     <cds-accordion-header>Item 3</cds-accordion-header>
+ *     <cds-accordion-content>Content 3</cds-accordion-content>
+ *   </cds-accordion-panel>
+ * </cds-accordion>
+ * ```
+ *
+ * @element cds-accordion-panel
+ * @slot
+ * @slot cds-accordion-header
+ * @slot cds-accordion-content
+ * @event expandedChange - notify when the user has clicked the panel header
+ * @cssprop --border-color
+ * @cssprop --border-width
+ * @cssprop --color
+ * @cssprop --font-size
+ * @cssprop --font-weight
+ * @cssprop --header-background
+ * @cssprop --header-color
+ * @cssprop --header-font-size
+ * @cssprop --header-font-weight
+ * @cssprop --header-open-background
+ * @cssprop --transition-duration
+ */
+export class CdsAccordionPanel extends LitElement {
+  @property({ type: Boolean }) disabled = false;
+
+  @property({ type: Boolean }) expanded = false;
+
+  @event() protected expandedChange: EventEmitter<boolean>;
+
+  connectedCallback() {
+    super.connectedCallback();
+  }
+
+  private toggle() {
+    this.expandedChange.emit(!this.expanded);
+  }
+
+  render() {
+    return html` <div class="private-host" role="group">
+      <div class="accordion-header${this.expanded ? ' open' : ''}${this.disabled ? ' disabled' : ''}">
+        <button
+          cds-layout="horizontal p-x:md p-y:sm"
+          type="button"
+          @click="${() => this.toggle()}"
+          class="accordion-header-button"
+          ?disabled="${this.disabled}"
+          aria-disabled="${this.disabled}"
+          aria-expanded="${this.expanded}"
+        >
+          <span class="accordion-status" cds-layout="vertical align:vertical-center p-r:sm">
+            <span class="accordion-angle-wrapper">
+              <cds-icon class="accordion-angle" shape="angle" direction="right" size="14"></cds-icon>
+            </span>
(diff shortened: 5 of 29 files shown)


packages/core/src/internal/base/accordion-panel.base.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { html, LitElement } from 'lit-element';
+import { event, EventEmitter } from '../decorators/event.js';
+import { property } from '../decorators/property.js';
+export class CdsBaseAccordionPanel extends LitElement {
```
mathisscott: I'm not sure how I feel about such a sparse base class. I think this is bringing us too close to the line of inheritance versus composition – a.k.a. the murder-robot-dog problem.

What if we needed an accordion panel that was also a focus trap? With this approach, I don't see how we could do that without rewriting parts of one or the other in the same class. And in this case I feel like the accordion base class loses out because it is so limited.

Did we consider making an accordion component and extending _that_ to make things like the stepper?

packages/core/tokens.yml
```diff
@@ -21,6 +21,10 @@ props:
     value: 'hsl(197, 68%, 65%)'
     type: 'hsl'
     category: 'global'
+  - name: selection-color
```
mathisscott, line 24: 👍

packages/core/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+      </cds-accordion-panel>
+    </cds-accordion>
+  `;
+};
+
+export const basicAccordion = () => {
+  return html`
+    <cds-accordion>
+      <cds-accordion-panel expanded>
+        <cds-accordion-header>Item 1</cds-accordion-header>
```
mathisscott: Instead of `Item1`, `Item2` and `Item3` can we have the titles reflect what each panel is demoing?

• Expanded accordion
• Disabled accordion
• Collapsed accordion

I think it will help answer some common questions.

packages/core/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+        <cds-accordion-content>Content 2</cds-accordion-content>
+      </cds-accordion-panel>
+      <cds-accordion-panel>
+        <cds-accordion-header>Item 3</cds-accordion-header>
+        <cds-accordion-content>Content 3</cds-accordion-content>
+      </cds-accordion-panel>
+    </cds-accordion>
+  `;
+};
+
```
mathisscott, line 45: What would an accordion that asynchronously loaded its content look like? 


As in, a user clicks on the title, the panel expands, and what?

While I would think that this is a pattern, it may make for a good demo.

packages/core/src/accordion/accordion.element.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @element cds-accordion
+ * @slot - Content slot for inside the accordion
+ * @cssprop --border-color
+ * @cssprop --border-width
+ * @cssprop --border-radius
+ */
+export class CdsAccordion extends LitElement {
+  render() {
+    return html` <slot name="accordion-panel"></slot> `;
```
mathisscott: Why does this slot need to be named?

packages/core/src/accordion/accordion.element.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Web component accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded='panel1Expanded' (expandedChange)='panel1Expanded = $event.detail'>
```
mathisscott: this sample code looks like Angular. do you think that would confuse anyone?

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private toggle() {
+    this.expanded = !this.expanded;
+    this.expandedChange.emit(this.expanded);
+  }
+
+  render() {
+    return html` <div class="private-host" role="group">
+      <div class="accordion-header${this.expanded ? ' open' : ''}${this.disabled ? ' disabled' : ''}">
```
mathisscott, line 69: We seem to have several approaches to boolean attributes. 

Here: `${this.disabled ? ' disabled' : ''}`

`control.element.ts, line 208`: `.disabled="${this.disabled}"`

`file.element.ts, line 45` (also line 75 here): `?disabled=${this.disabled}`

It appears that `.disabled` has won out more often than not. But I think we should make a case for which is the best way and converge around that.

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,102 @@
...
+          type="button"
+          @click="${() => this.toggle()}"
+          class="accordion-header-button"
+          ?disabled="${this.disabled}"
+          aria-disabled="${this.disabled}"
+          aria-expanded="${this.expanded}"
+        >
+          <span class="accordion-status" cds-layout="vertical align:vertical-center p-r:sm">
+            <span class="accordion-angle-wrapper">
+              <cds-icon class="accordion-angle" shape="angle" dir="right" size="14"></cds-icon>
```
mathisscott: Does the accordion angle change direction when open? 

`dir` is also something we are considering deprecating. i recommend using `direction` instead. same values, longer name. 😁 

Also, can people put their own icon here? How could we enable them to do that?

mathisscott: I think we are also missing some aria here for a11y...

packages/core/src/accordion/entrypoint.tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.lib.json",
+  "compilerOptions": {
+    "composite": true
+  },
+  "references": [{ "path": "../internal/entrypoint.tsconfig.json" }]
```
mathisscott: since accordion is using icons internally, i think we also need the icons path here.

See `./core/src/forms/entrypoint.tsconfig.json`

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @cssprop --header-open-background-color
+ * @cssprop --transition-duration
+ */
+export class CdsAccordionPanel extends CdsBaseAccordionPanel {
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('slot', 'accordion-panel');
+  }
+
+  private toggle() {
```
mathisscott, line 63: How will end users escape hatch this?

packages/core/src/accordion/accordion-panel.element.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+  });
+
+  it('should have a slot attribute of value `accordion-panel`', async () => {
+    await componentIsStable(component);
+    expect(component.hasAttribute('slot')).toBe(true);
+    expect(component.getAttribute('slot')).toEqual('accordion-panel');
+  });
+
+  it('should emit an expandedChange event when header is clicked', async done => {
```
mathisscott, line 42: The click is also changing whether the panel is expanded or not. I don't know if that's the path we want to go down. But if it is, we should write a unit test to verify that state is managed.

packages/core/src/accordion/accordion-panel.element.scss
```diff
@@ -0,0 +1,78 @@
...
+// todo: set max-height for content
+.accordion-content-wrapper {
+  height: 0;
+  transition: transform height var(--transition-duration);
+  transition-timing-function: cubic-bezier(0, 0.99, 0, 0.99);
+
+  overflow-y: hidden;
+
+  &.open {
+    transform: scaleY(1);
```
mathisscott: is it our intent to keep the transform here? because it will break overlays and we don't seem to be making use of it anywhere else.

packages/core/src/accordion/accordion-panel.element.scss
```diff
@@ -0,0 +1,78 @@
...
+    ::slotted([slot='accordion-header']) {
+      cursor: not-allowed;
+    }
+  }
+
+  &.open {
+    --header-background-color: var(--header-open-background-color);
+
+    .accordion-angle-wrapper {
+      transform: rotate(90deg);
```
mathisscott, line 76: this seems related to my comment on giving users the option to customize the accordion icon below.

packages/core/src/accordion/accordion-header.element.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Web component accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded='panel1Expanded' (expandedChange)='panel1Expanded = $event.detail'>
```
mathisscott: Looks like more angular code. i'm not sure we need a full accordion code example here...

packages/core/src/accordion/accordion-content.element.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Web component accordion
+ *
+ * ```typescript
+ * import '@clr/core/accordion/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-accordion>
+ *   <cds-accordion-panel expanded='panel1Expanded' (expandedChange)='panel1Expanded = $event.detail'>
```
mathisscott: Same question here re: Angular looking code and whether we need a full accordion code example.

mathisscott (changes requested):

packages/core/src/internal/base/accordion-panel.base.ts
jeeyun: Thank you for that feedback. Our Angular component actually follows this and stepper extends from accordion (including sub-components). I found it to be a bit hard to follow on the template side, but I don't think we have to keep that the same in core so let me take a look. I think your points are valid and I could probably come up with a decent solution.

packages/core/src/accordion/accordion-panel.element.scss
```diff
@@ -0,0 +1,78 @@
...
+  --font-weight: #{$cds-token-typography-font-weight-regular};
+  --header-background-color: #{$cds-token-color-neutral-50};
+  --header-color: #{$cds-token-color-neutral-800};
+  --header-font-size: #{$cds-token-typography-font-size-3};
+  --header-font-weight: #{$cds-token-typography-font-weight-semibold};
+  --header-open-background-color: #{$cds-token-global-selection-color};
+  --transition-duration: 200ms;
+}
+
+::slotted(cds-accordion-header) {
```
mathisscott, line 17: A challenge we've had with using `cds-*` element selectors is that the ShadyDOM polyfill will rewrite the elements. Basically turning them into spans with classnames.

For this reason, we have tried to avoid using the custom elements in our CSS selectors. Because we can't count on them being there.

I don't know if this still applies after the IE11 announcement this month. @coryrylan may have more info on how this does or doesn't affect Safari.

But just a heads up that all these custom element selectors may need to change...

packages/core/src/accordion/accordion-panel.element.scss
mathisscott: Another issue here is that there is an animation when the panel opens but no animation when it closes. Is that intentional?

packages/core/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+export const customStyles = () => {
+  return html`
+    <style>
+      cds-accordion-panel.app-custom {
+        --border-color: midnightblue;
+        --header-color: aliceblue;
```
mathisscott: Could we include changing the color of the caret/angle icon here? I think it would be helpful for people to see/know they can do that.

packages/core/src/accordion/accordion.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+## Accordion
+
+<Preview>
+  <Story id="components-accordion-stories--basic-accordion" />
+</Preview>
+
+## Custom accordion styles
+
+<Preview>
+  <Story id="components-alert-stories--custom-styles" />
```
mathisscott: Should this point to the accordion stories?

packages/core/src/accordion/accordion.stories.mdx
```diff
@@ -0,0 +1,51 @@
+import { Meta, Props, Story, Preview, API } from '@storybook/addon-docs/blocks';
+import { html, LitElement } from 'lit-element';
+
+<Meta title="Components/Accordion/Getting Started" />
```
mathisscott: The stories for the accordion need to go into the components preview section. New components go through one full version preview.

mathisscott (changes requested):

packages/core/src/accordion/accordion.element.ts
jeeyun: it doesn't. removing. 😄

packages/core/src/accordion/accordion.element.ts
jeeyun: yeah that's a good point. i think initially we populated the sample codes in Angular and I probably carried it over. I'll update this.

packages/core/src/accordion/accordion-panel.element.ts
jeeyun: I think those are different use cases and all valid actually: https://stackoverflow.com/questions/61490521/dot-with-html-elements-attribute-property

- boolean attribute: `?disabled=${this.disabled}`

- property: `.disabled=${this.disabled}`

- attribute value: `<div class="${...}></div>`

In this component we are using the boolean attribute on an html button, and the conditional statement for the div class.

In the case of `control.element.ts` it's actually a property (line 105) so it'd be correct to use the `.disabled="${this.disabled}"` there.

packages/core/src/accordion/accordion-panel.element.ts
jeeyun: The accordion angle does change direction when open but via transform and not by changing the attribute value. I'll still update to use `direction`. 

Can you tell me what may be missing from a11y point of view? I vetted it against our Angular implementation and also did screenreader testing locally. It's possible I missed something though.

packages/core/src/accordion/accordion-panel.element.ts
mathisscott: ah. i see. maybe that's a good candidate for the contribution guidelines if it isn't there already. i'll make a note to check and add that to them if they aren't there. just looking at it, it looks like we can't make up our mind how we want to do it. but understanding it's actually three different things...

packages/core/src/accordion/accordion-panel.element.ts
mathisscott: An `aria-label` on the icon is probably called for...

packages/core/src/accordion/accordion-panel.element.ts
jeeyun: You are right; this should be the responsibility of the application. Removing the auto toggling from component logic. Thanks for catching that.

packages/core/src/accordion/accordion-panel.element.spec.ts
jeeyun: Yup. Updating the logic. It should be handled by the component.

packages/core/src/accordion/accordion-panel.element.scss
jeeyun: Removing.

packages/core/src/internal/base/accordion-panel.base.ts
jeeyun: Updated to get rid of the base class!

packages/core/src/accordion/accordion-panel.element.scss
jeeyun: Fixed to animate on both expand and collapse.

packages/core/src/accordion/accordion-panel.element.ts
jeeyun: The icon by itself is not actionable and is somewhat decorative. I've taken a look at couple of other aria accordions as well. Maybe it makes more sense to set `aria-hidden=true` so that it's hidden from SR user?

packages/core/src/accordion/accordion-panel.element.scss
coryrylan: With ie11 drop we no longer have to worry about this and can use element selectors 👍

jeeyun: Awesome! 🎉

packages/core/src/accordion/accordion-panel.element.scss
```diff
@@ -0,0 +1,85 @@
...
+  --border-width: #{$cds-token-global-border-width};
+  --color: #{$cds-token-color-neutral-800};
+  --icon-color: #{$cds-token-color-neutral-700};
+  --font-size: #{$cds-token-typography-font-size-4};
+  --font-weight: #{$cds-token-typography-font-weight-regular};
+  --header-background-color: #{$cds-token-color-neutral-50};
+  --header-color: #{$cds-token-color-neutral-800};
+  --header-font-size: #{$cds-token-typography-font-size-3};
+  --header-font-weight: #{$cds-token-typography-font-weight-semibold};
+  --header-open-background-color: #{$cds-token-global-selection-color};
```
coryrylan: May not be worth the effort to adjust templates but we could move these to the `cds-accordion-header` element similar to alerts/alert groups. Then the component doesnt need nested css props on the public API and may not need conditional classes in the component template.

```scss
// cds-accordion-header css
:root {
  --background: #{$cds-token-color-neutral-50};
  --color: #{$cds-token-color-neutral-800};
  --font-size: #{$cds-token-typography-font-size-3};
  --font-weight: #{$cds-token-typography-font-weight-semibold};
}
```

```scss
// cds-accordion-panel css
:host([expanded]) ::slotted(cds-accordion-header) {
  --background: #{$cds-token-global-selection-color};
}
```

```css
cds-accordion-header {
  --background: ...
  --color: ...
  --font-size: ...
  --font-weight: ...
}
```

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @slot
+ * @slot cds-accordion-header
+ * @slot cds-accordion-content
+ * @event expandedChange - notify when the user has clicked the panel header
+ * @cssprop --border-color
+ * @cssprop --border-width
+ * @cssprop --color
+ * @cssprop --font-size
+ * @cssprop --font-weight
+ * @cssprop --header-background-color
```
coryrylan: nit: using/assigning to `background` instead of `background-color` gives a bit more flexibility with theming

coryrylan (approved):

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,105 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles, event, EventEmitter } from '@clr/core/internal';
+import { ClarityIcons, angleIcon } from '@clr/core/icon';
+import { html, LitElement, property } from 'lit-element';
+import { styles } from './accordion-panel.element.css.js';
+
+ClarityIcons.addIcons(angleIcon);
```
coryrylan: Will want to move this to the `register.ts` so it is not removed by tree shaking

coryrylan (changes requested):

packages/core/src/accordion/accordion-panel.element.ts
jeeyun: Ah yes, thank you for catching that. Will move this.

packages/core/src/accordion/accordion.stories.ts
jeeyun: As discussed, we'll incorporate this into the pattern work since this is more of a "how to use this component in an application" rather than component's API.

mathisscott: Because we have no place to put "patterns" and because we have no current plan in place as to how we are tackling patterns, documenting them, or storing them for users to get to, I'm concerned that this is pushing a common ask/use case into the indefinite future. I feel like we're going to get asked this and have to answer it over and over. 

I've erred on the side of including a "pattern" or two in the stories for now (see: accessible circular progress, nested/layered overlays, etc.) so that users have working examples now. Given the limitations of storybook, there's a bit of a leap between the story and what a user would implement. But the story demonstrates that something is possible and the raw mechanics of how it would be accomplished.

packages/core/src/accordion/accordion-panel.element.ts
mathisscott: `aria-hidden="true'` makes a lot of sense in that case...

packages/core/src/accordion/accordion-panel.element.scss
jeeyun: I started out trying to refactor but I think I'll leave it as is for now.

Animation also relies on the conditional css class and some parts of the header template rely on both `expanded` and `disabled` states that are available in the panel, making the refactor a bit less clean-cut than initially anticipated. I'm not opposed to improving things though, so if you think it might still be worth the effort let me know and maybe we can pair on this?

coryrylan: Going to resolve since I'm not sure if its really worth the effort especially with the new theming coming soon after.

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @cssprop --transition-duration
+ */
+export class CdsAccordionPanel extends CdsBaseAccordionPanel {
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('slot', 'accordion-panel');
+  }
+
+  private toggle() {
+    this.expanded = !this.expanded;
```
coryrylan: Not sure if this was removed based on the comments above but looks like its still needs to be removed to prevent the stateful escape hatch.

packages/core/src/accordion/accordion.stories.ts
coryrylan: I think something like this will work in storybook

```html
<cds-accordion>
  <cds-accordion-panel expanded>
    <cds-accordion-header>Header 1</cds-accordion-header>
    <cds-accordion-content>Content 1</cds-accordion-content>
  </cds-accordion-panel>
  <cds-accordion-panel id="async-panel">
    <cds-accordion-header>Header 1</cds-accordion-header>
    <cds-accordion-content>
      
    </cds-accordion-content>
  </cds-accordion-panel>
</cds-accordion>
<script>
  const asyncPanel = document.querySelector('#async-panel');
  const asyncPanelContent = asyncPanel.querySelector('cds-accordion-content');

  asyncPanel.addEventListener('expandedChange', () => {
    asyncPanel.expanded = !asyncPanel.expanded;
    asyncPanelContent.innerHTML = `<cds-progress-circle></cds-progress-circle>`;

    if (asyncPanel.expanded) {
      setTimeout(() => asyncPanelContent.innerHTML = `<p>hello world</p>`, 3000);
    }
  });
</script>
```

packages/core/src/accordion/accordion-panel.element.ts
coryrylan: nvm I see it :)

packages/core/src/internal/utils/metadata.ts
```diff
@@ -18,7 +18,7 @@ export function getElementStorybookArgs(args: any) {
 
 function getCustomPropsStorybookArgs(customElementMetadata: any) {
   return customElementMetadata.cssProperties?.reduce((obj: any, prop: { name: string }) => {
-    const type = prop?.name?.includes('color') || prop.name === '--background' ? 'color' : 'text';
+    const type = prop?.name?.includes('color') || prop.name?.endsWith('background') ? 'color' : 'text';
```
mathisscott: wouldn't we need this to be...

```
const type = prop?.name?.includes('color') || prop?.name?.endsWith('background') ? 'color' : 'text';
```

If `prop` can be null the first time around, I don't see why it couldn't be null the second.

mathisscott (changes requested):

coryrylan (approved):

packages/core/src/accordion/accordion.stories.ts
mathisscott: I had a hard time tracking this down but I see the example in `accordion-panel.stories.ts`

Thanks for doing that!

Issue #4925, opened 2020-08-20
core: remove icon deprecations


coryrylan:
In the cds-icon styles we have several legacy icon classes that are marked for deprecation and can be removed for 5.0

https://github.com/vmware/clarity/blob/master/packages/core/src/icon/icon.element.scss#L66

PR #5110, opened 2020-10-02
fix: remove core icons deprecated styles and methods

bbogdanov:
Removing the deprecated methods and styles introduced in previous versions and marked as deprecated for v5 in Icons.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4925

## What is the new behavior?

Removed deprecated code

- [x] Yes
- [ ] No

Anybody who uses the old icon sizes styles, `dir` property of `cds-icon` and `get`, `add` and `alias` methods from `icon.service` will be affected.

### DIFF
--- a/packages/core/src/icon/icon.element.scss
+++ b/packages/core/src/icon/icon.element.scss
@@ -37,121 +37,83 @@ svg {
 }
 
 // sizing
-// DEPRECATED IN 4.0: all legacy sizing classnames
-// will be replaced with [size="sm|md|lg|xl|xxl"] selectors!
-:host(.clr-i-size-sm) {
+:host([size='sm']) {
   // weirdly, the default... 16px
   @include equilateral(#{$cds-token-space-size-7});
 }
 
-:host(.clr-i-size-md) {
+:host([size='md']) {
   // 24px
   @include equilateral(#{$cds-token-space-size-9});
 }
 
-:host(.clr-i-size-lg) {
+:host([size='lg']) {
   // 36px
   @include equilateral(#{$cds-token-space-size-11});
 }
 
-:host(.clr-i-size-xl) {
+:host([size='xl']) {
   // 48px
   @include equilateral(#{$cds-token-space-size-12});
 }
 
-:host(.clr-i-size-xxl) {
+:host([size='xxl']) {
   // 64px
   @include equilateral(calc(#{$cds-token-space-size-13} - #{$cds-token-space-size-5}));
 }
 
 // colors
-// DEPRECATED IN 3.0: is-green, .is-success
-:host(.is-green),
-:host(.is-success),
 :host([status='success']) {
   --color: #{$cds-token-color-success-700};
 }
 
-// DEPRECATED IN 3.0: is-red, is-error, is-danger
-:host(.is-red),
-:host(.is-danger),
-:host(.is-error),
 :host([status='danger']) {
   --color: #{$cds-token-color-danger-700};
 }
 
-// DEPRECATED IN 3.0: is-warning
-:host(.is-warning),
 :host([status='warning']) {
   --color: #{$cds-token-color-warning-900};
 }
 
-// DEPRECATED IN 3.0: is-blue, is-info, is-highlight
-:host(.is-blue),
-:host(.is-info),
-:host(.is-highlight),
 :host([status='info']) {
   --color: #{$cds-token-color-action-600};
 }
 
 // INVERSE COLORS
 
-// DEPRECATED IN 3.0: is-white, is-inverse
-:host(.is-white),
-:host(.is-inverse),
 :host([inverse]) {
   --color: #{$cds-token-color-neutral-0};
 }
 
-// DEPRECATED IN 3.0: is-info, is-inverse
-:host([inverse].is-info),
-:host([status='info'].is-inverse),
 :host([inverse][status='info']) {
   --color: #{$cds-token-color-action-400};
 }
 
-// DEPRECATED IN 3.0: is-success, is-inverse
-:host([inverse].is-success),
-:host([status='success'].is-inverse),
 :host([inverse][status='success']) {
   --color: #{$cds-token-color-success-400};
 }
 
-// DEPRECATED IN 3.0: is-error, is-danger, is-inverse
-:host(.is-inverse.is-error),
-:host(.is-inverse.is-danger),
-:host([status='danger'].is-inverse),
-:host([inverse].is-error),
-:host([inverse].is-danger),
 :host([inverse][status='danger']) {
   --color: #{$cds-token-color-danger-700};
 }
 
-// DEPRECATED IN 3.0: is-warning, is-inverse
-:host(.is-inverse.is-warning),
-:host([status='warning'].is-inverse),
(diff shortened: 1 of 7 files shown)


packages/core/src/icon/icon.element.scss
```diff
@@ -36,105 +36,6 @@ svg {
   display: block;
 }
 
-// sizing
-// DEPRECATED IN 4.0: all legacy sizing classnames
-// will be replaced with [size="sm|md|lg|xl|xxl"] selectors!
-:host(.clr-i-size-sm) {
```
mathisscott, line 42: The `.clir-i-size-*` classes have been removed but have not been replaced. This will break sizing across all icons.

Please fix.

Also, there are a lot of deprecations in this file that were missed (e.g. `has-badge`, `has-alert`, etc.)

packages/core/src/icon/icon.element.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('icon element', () => {
   let component: CdsIcon;
 
   beforeAll(() => {
-    ClarityIcons.add({ testing: testIcon });
+    ClarityIcons.addIcons(['testing', testIcon]);
```
mathisscott, line 21: In this file, the `.clr-i-size-*` unit tests are still here, despite being removed from CSS.

Please correct.

packages/core/src/icon/icon.element.ts
```diff
@@ -87,16 +87,6 @@ export class CdsIcon extends LitElement {
...
 
-  /**
-   * @deprecated
-   * Takes a directional value (up|down|left|right) that rotates the icon 90° with the
-   * top of the icon pointing in the specified direction.
-   *
-   * Deprecated in 3.0. Use `direction` instead. `dir` will be removed in 4.0!
-   */
-  @property({ type: String })
-  dir: Directions;
```
mathisscott, line 98: This property is removed in the `element.ts` file but is still in the CSS. Please fix.

packages/core/src/icon/utils/icon.classnames.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('Icon classname helpers: ', () => {
     let component: CdsIcon;
 
     beforeAll(() => {
-      ClarityIcons.add({ testing: testIcon });
+      ClarityIcons.addIcons(['testing', testIcon]);
```
mathisscott, line 102: There's a lot of code in this spec and corresponding utility that is no longer needed. Please remove.

mathisscott (changes requested):

packages/core/src/icon/icon.element.scss
bbogdanov: What do you mean by `not been replaced`? 

Do you mean `:host([size="sm"])` instead of `:host(.clr-i-size-sm)`,  `:host([size="md"])` instead of `:host(.clr-i-size-md)`, etc ?

mathisscott: Yes. Also the code that "translates" `size="sm"` into the classnames is no longer needed.

packages/core/src/icon/icon.element.scss
```diff
@@ -36,126 +36,6 @@ svg {
...
-:host(.clr-i-size-xxl) {
-  // 64px
-  @include equilateral(calc(#{$cds-token-space-size-13} - #{$cds-token-space-size-5}));
-}
-
-// colors
-// DEPRECATED IN 3.0: is-green, .is-success
-:host(.is-green),
-:host(.is-success),
-:host([status='success']) {
```
bbogdanov, line 71: @mathisscott When I went through the changes I think lines like 71, 79, 85, etc actually don't need to be removed. Am I wrong?

mathisscott: You are correct. Good catch. we do still need the `:host([status='...'])` styles...

packages/core/src/icon/icon.element.scss
```diff
@@ -36,126 +36,6 @@ svg {
...
-:host(.is-inverse.is-warning),
-:host([status='warning'].is-inverse),
-:host([inverse].is-warning),
-:host([inverse][status='warning']) {
-  --color: #{$cds-token-color-warning-500};
-}
-
-// directional
-:host([dir='up']),
-:host([direction='up']) {
```
mathisscott, line 140: We also need to keep the `:host([direction='...'])` styles

mathisscott (changes requested):

packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -131,20 +91,15 @@ export function getUpdateSizeStrategy(size: string) {
 
 export function updateIconSizeStyleOrClassnames(el: CdsIcon, size: string) {
```
bbogdanov, line 92: @mathisscott The name of the function can be renamed. I didn't do it because I am not sure if someone out there is not using it for some reason.

mathisscott: These internal utility functions aren't documented so there is low risk that someone outside of the Clarity project is using them.

packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -118,7 +78,7 @@ export function getUpdateSizeStrategy(size: string) {
     return SizeUpdateStrategies.NilSizeValue;
   }
 
-  if (isString(size) && isIconTshirtSizeClassname(size)) {
+  if (isString(size) && isTshirtSize(size)) {
```
mathisscott, line 81: `isTshirtSize()` already guards against non-strings and nil values. I think it is safe to remove `isString()` from this conditional...

packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -131,20 +91,15 @@ export function getUpdateSizeStrategy(size: string) {
 
 export function updateIconSizeStyleOrClassnames(el: CdsIcon, size: string) {
   const updateStrategy = getUpdateSizeStrategy(size);
-  const newTshirtSize = getIconTshirtSizeClassname(size);
```
mathisscott, line 134: This all looks so much better with the legacy code removed. Thank you!

mathisscott (changes requested):

packages/core/src/icon/utils/icon.classnames.ts
bbogdanov: At the moment only TypeScript guards it on compile time. It will loop through the array of sizes when it is executed in JavaScript despite the parameter type. On the other hand, we expect the value to be one of the t-shirt sizes in the array so if the value is not found in the array it will return `false`. We are safe.

bbogdanov: Line 77 guards it only against `nil` and empty strings.

mathisscott (dismissed):

coryrylan (approved):

mathisscott (approved):

coryrylan (approved):

PR #5116, opened 2020-10-06
feat: eslint plugin for reporting the usage of clarity Angular elements

sis0k0:
Signed-off-by: Stanimira Vlaeva <svlaeva@vmware.com>

## @clr/eslint-plugin-clarity-migration

This PR adds a new package to the repository - `@clr/eslint-plugin-clarity-migration`. This ESLint plugin can be used to discover the usage of deprecated components in Angular and non-Angular projects and report it as a warning or error, depending on the configuration in the project.

Currently, the plugin contains a single rule - `no-clr-button`. This rule reports the usage of `<button class="btn btn-primary"></button>` inside HTML files or inside inlined Angular components templates (TS files).

For parsing the TS files in the project, the plugin uses [`@typescript-eslint/plugin`](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin). Then, it parses the HTML within the component template with [`node-html-parser`](https://www.npmjs.com/package/node-html-parser). Using the AST tree provided from `node-html-parser` it detects the usage of `<button class="btn btn-primary">`.

For parsing the HTML files, the plugin uses [`eslint-html-parser`](https://www.npmjs.com/package/eslint-html-parser). Unfortunately, that parser cannot handle more than one root element in the HTML file. It doesn't report the rest of the tree. In the example below, only the `div` element will be traversed.

```html
<div>First</div>
<button class="btn btn-primary">Second</button>
```

We can gradually add rules for all Clarity Angular components. We can also point to docs links describing how to migrate the used component to a Clarity Core component. For some cases, we can also include a fixer with the rule that automatically performs the migration to the Core component.

## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [X] No

### DIFF
--- a/packages/eslint-plugin-clarity-migration/package.json
+++ b/packages/eslint-plugin-clarity-migration/package.json
@@ -1,0 +1,69 @@
+{
+  "name": "@clr/eslint-plugin-clarity-migration",
+  "version": "0.0.0",
+  "description": "A set of ESLint rules for Clarity migration.",
+  "main": "dist/src/index.js",
+  "typings": "dist/src/index.d.ts",
+  "keywords": [
+    "clarity",
+    "eslint-plugin",
+    "eslint"
+  ],
+  "author": "clarity",
+  "license": "MIT",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w",
+    "lint": "npm run lint:src && npm run lint:test",
+    "lint:src": "tslint --project ./tsconfig.json -c ./tslint.json 'src/**/*.ts'",
+    "lint:test": "tslint -c ./tslint.json 'test/**/*.ts'",
+    "lint:fix": "npm run lint:src:fix && npm run lint:test:fix",
+    "lint:src:fix": "tslint --project ./tsconfig.json -c ./tslint.json 'src/**/*.ts' --fix",
+    "lint:test:fix": "tslint -c ./tslint.json 'test/**/*.ts' --fix",
+    "test": "jest",
+    "test:debug": "node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand --coverage false",
+    "prepublishOnly": "npm run build && npm run lint && npm run test",
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s && git add CHANGELOG.md"
+  },
+  "dependencies": {
+    "@typescript-eslint/experimental-utils": "^4.1.1",
+    "node-html-parser": "^1.2.20"
+  },
+  "devDependencies": {
+    "@types/eslint": "^7.2.2",
+    "@types/jest": "^26.0.13",
+    "@types/node": "^14.0.2",
+    "@typescript-eslint/eslint-plugin": "^4.1.1",
+    "@typescript-eslint/eslint-plugin-tslint": "^4.1.1",
+    "@typescript-eslint/parser": "^4.1.1",
+    "@typescript-eslint/typescript-estree": "^4.1.1",
+    "conventional-changelog-cli": "^2.0.25",
+    "eslint": "^7.9.0",
+    "eslint-html-parser": "^6.8.0",
+    "jest": "^26.4.2",
+    "ts-jest": "^26.3.0",
+    "tslint": "^6.1.3",
+    "typescript": "^3.7.5"
+  },
+  "jest": {
+    "testEnvironment": "node",
+    "moduleFileExtensions": [
+      "js",
+      "ts"
+    ],
+    "collectCoverage": true,
+    "collectCoverageFrom": [
+      "<rootDir>/src/**"
+    ],
+    "coverageDirectory": "<rootDir>/reports/coverage",
+    "transform": {
+      "\\.(ts)$": "ts-jest"
+    },
+    "testRegex": "/test/.*\\.spec\\.ts$",
+    "preset": "ts-jest",
+    "testMatch": null
+  }
+}
--- a/packages/eslint-plugin-clarity-migration/src/rules/no-clr-button/index.ts
+++ b/packages/eslint-plugin-clarity-migration/src/rules/no-clr-button/index.ts
@@ -1,0 +1,61 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
+import { parse } from 'node-html-parser';
+import { getDecoratorPropertyValue } from '../utils';
+import { HTMLElement } from 'eslint-html-parser';
+
+export const createESLintRule = ESLintUtils.RuleCreator(() => ``);
+
+export type MessageIds = 'clrButtonFailure';
+
+export default createESLintRule({
+  name: 'no-clr-button',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow use of clr-button',
+      category: 'Best Practices',
+      recommended: 'warn',
+    },
+    fixable: 'code',
+    messages: {
+      clrButtonFailure: 'Using clr-button is not allowed!',
+    },
+    schema: [{}],
+  },
+  defaultOptions: [{}],
+  create(context) {
+    return {
+      'HTMLElement[tagName="button"]'(node: HTMLElement) {
+        const classNode = node.attributes?.find(attribute => attribute.attributeName.value === 'class');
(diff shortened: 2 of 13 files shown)


packages/eslint-plugin-clarity-migration/test/test-helper.ts
```diff
@@ -0,0 +1,78 @@
...
+      if (typeof test !== 'string') {
+        if (test.parser === parser) {
+          throw new Error(errorMessage);
+        }
+        if (!test.filename) {
+          (<any>test).filename = this.filename;
+        }
+      }
+    });
+    tests.invalid.forEach(test => {
```
bbogdanov, line 67: Can we just set here `((test: any) => ...)` instead of typecasting like it is done on line 71?

packages/eslint-plugin-clarity-migration/test/no-clr-button.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+                  const myButton = \`
+                    <button class="btn btn-primary custom-class">Primary</button>
+                  \`;
+              }
+            `,
+  ],
+});
+
+function getInvalidTest(code: string, output?: string) {
+  const invalidTest: any = {
```
bbogdanov, line 87: We can create an interface `InvalidTest` and use it  instead of `any`

```
interface InvalidTest {
  code: string;
  errors: { messegeId: string }[] 
  output?: string;
}
```

packages/eslint-plugin-clarity-migration/test/test-helper.ts
```diff
@@ -0,0 +1,78 @@
...
+            code: test,
+            filename: this.filename,
+          };
+        }
+
+        return test;
+      });
+    }
+
+    tests.valid.forEach(test => {
```
bbogdanov, line 57: `(test: any) =>` like the one bellow.

packages/eslint-plugin-clarity-migration/src/rules/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+export const getDecoratorPropertyValue = (
+  decorator: TSESTree.Decorator,
+  name: string
+): TSESTree.Expression | TSESTree.Literal | undefined => {
+  const property = getDecoratorProperty(decorator, name);
+  if (!property) {
+    return undefined;
+  }
+
+  return property.value as any;
```
bbogdanov, line 45: I am curious - Why we need to cast? I mean we already specify what type of values we expect in line 39.

sis0k0: Basically - mismatching types from `@typescript-eslint` and `estree`.

bbogdanov (approved):

packages/eslint-plugin-clarity-migration/src/rules/no-clr-button/index.ts
```diff
@@ -0,0 +1,61 @@
...
+export type MessageIds = 'clrButtonFailure';
+
+export default createESLintRule({
+  name: 'no-clr-button',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow use of clr-button',
+      category: 'Best Practices',
+      recommended: 'warn',
```
hippee-lee, line 17: Any thoughts as to where we might put url's to written guides? E.g - github wiki or website release notes?

hippee-lee (commented):
Overall the rule lgtm.

packages/eslint-plugin-clarity-migration/src/rules/no-clr-button/index.ts
sis0k0: I think these should be part of the error messages.
```js
messages: {
	clrButtonFailure: 'Using clr-button is not allowed! See _link_.'
}
```

packages/eslint-plugin-clarity-migration/package.json
```diff
@@ -0,0 +1,88 @@
+{
+  "name": "@clr/eslint-plugin-clarity-migration",
```
gnomeontherun, line 2: Do we need to publish this as `eslint-plugin-clarity-migration` instead of with `@clr` prefix?

packages/eslint-plugin-clarity-migration/package.json
```diff
@@ -0,0 +1,88 @@
...
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w",
+    "lint": "npm run lint:src && npm run lint:test",
+    "lint:src": "tslint --project ./tsconfig.json -c ./tslint.json 'src/**/*.ts'",
+    "lint:test": "tslint -c ./tslint.json 'test/**/*.ts'",
+    "lint:fix": "npm run lint:src:fix && npm run lint:test:fix",
+    "lint:src:fix": "tslint --project ./tsconfig.json -c ./tslint.json 'src/**/*.ts' --fix",
+    "lint:test:fix": "tslint -c ./tslint.json 'test/**/*.ts' --fix",
+    "test": "jest",
```
gnomeontherun, line 26: We use karma/jasmine for our other testing, I'm hesitant to see a new testing framework introduced. What do we get from Jest?

packages/eslint-plugin-clarity-migration/package.json
```diff
@@ -0,0 +1,88 @@
...
+    "testRegex": "/test/.*\\.spec\\.ts$",
+    "preset": "ts-jest",
+    "testMatch": null
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
```
gnomeontherun, line 78: I think these husky/lint rules don't need to be here, they apply at the top level project.

packages/eslint-plugin-clarity-migration/src/rules/no-clr-button/index.ts
gnomeontherun: We don't have links yet but I agree this is how we'd do it generally speaking.

gnomeontherun (commented):
We have a lot more than just the primary button so we'll need to extend this, but this is a good start overall.

packages/eslint-plugin-clarity-migration/package.json
bbogdanov: I agree that we need to stick using only one framework. On the other hand, Jest is way faster than karma. In the last project with 2000+ tests, I worked on we migrated from karma to Jest and we reduced the test run time by 70%. The migration was also pretty easy. Just sharing.

packages/eslint-plugin-clarity-migration/package.json
sis0k0: Right now the package is configured to be published under the `@clr` scope - `@clr/eslint-plugin-clarity-migration`. If we don't want that, we can also publish it as `eslint-plugin-clarity-migration`.

packages/eslint-plugin-clarity-migration/package.json
sis0k0: Initially, I set up the POC project tests with Jest. While transferring the plugin to this repo, I noticed that the `@clr/react` package also uses Jest and that's why I left it.

packages/eslint-plugin-clarity-migration/package.json
sis0k0: Thanks for noting that! I'll update the configuration.

hippee-lee (approved):

hippee-lee:
@sis0k0 - fyi I updated the default branch for this PR and re-requested reviewers.

bbogdanov (dismissed):

packages/eslint-plugin-clarity-migration/package.json
sis0k0: @gnomeontherun, is it ok to leave Jest or should I swap it out for Karma+Jasmine?

gnomeontherun: That is fair, I leave it to the team to decide but multiple options seems harder to maintain long term.

bbogdanov (approved):

Jinnie (approved):

bbogdanov (approved):

Jinnie (approved):

Issue #5085, opened 2020-09-28
Datagrid filter / sorting not working


kaliopane:
## Describe the bug

When we have a checkbox input that defines a filter and it is turned on, this breaks the datagrid sorting and filtering.

## How to reproduce

[Link to Stackblitz](https://stackblitz.com/edit/clarity-v3-datagrid-selection-asnknn?file=src/app/app.component.ts)

1. Click on the button
2. Start sorting and / or filtering columns
3. Observe that no elements are present in the datagrid

## Expected behavior

All datagrid entries should be visible. Regression introduced after upgrade from Clarity v1.

## Versions

**App**

-   Angular: [9.1.12]
-   Node: [10.19] - locally
-   Clarity: [3.1.7]

**Device:**

-   Type: [MacBook]
-    OS: [Mojave]
-    Browser [FF]
-    Version [80.0.1]

sis0k0:
@kaliopane, it seems that the behavior is reproducible even without the filters. Sorting + pagination in the datagrid is enough. Here's a minimal reproduction - https://stackblitz.com/edit/clarity-v3-datagrid-selection-bcncz1?file=src/app/app.component.ts.

We're working on resolving the issue.

kaliopane:
Thanks for the updates guys!
It seems we ought to switch to using Ivy as well, as Angular is not really being helpful in this particular situation.

PR #5117, opened 2020-10-06
fix: (datagrid) filtering not working well with Ivy off

Jinnie:
issue #5085

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When Ivy is turned off, the displayedRows list accumulates old ghost items, which mess the proper child views visualization. Thus, after been filtered out, rows are not shown back when the filter is removed.

Issue Number: #5085 

## What is the new behavior?

Works fine both with and without Ivy

- [ ] Yes
- [x] No

## Other information

Please note, that because of the complicated setup we can't make a test for the no-Ivy scenario. We can't have a test run with and without Ivy unless we run a new task with new test setup.

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid.ts
@@ -226,6 +226,13 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
         if (!this.items.smart) {
           this.items.all = this.rows.map((row: ClrDatagridRow<T>) => row.item);
         }
+        // Remove any projected rows from the displayedRows container
+        // Necessary with Ivy off. See https://github.com/vmware/clarity/issues/4692
+        for (let i = this._displayedRows.length - 1; i >= 0; i--) {
+          if (this._displayedRows.get(i).destroyed) {
+            this._displayedRows.remove(i);
+          }
+        }
         this.rows.forEach(row => {
           this._displayedRows.insert(row._view);
         });


bbogdanov (dismissed):

sis0k0 (approved):

gnomeontherun (commented):
We've had a regression from fixes here before. I want to be 100% certain before we merge that this is correct. How have we tested this?

Jinnie:
As I said, automated tests are not possible, so:
1. Checked with enabled Ivy that it does not break any current tests. I didn't test with Ivy off, as many of our other tests are failing with Ivy off.
2. Manually tested the following scenarios (both Ivy on and off):
    - no interference with the trackby functionalities - it keeps the proper items selected.
    - made sure the recently fixed issue of keeping detail pane open on refresh is okay.
3. Made sure we're not skipping the view destruction, which was previously leading to memory leaks.
4. Debugged the previously offending data list, which was accumulating dead records, thus leading to possible memory leaks.

Any other testing ideas or manually performed testing by the reviewers are welcome. I won't merge this for a couple of day more to allow for more reviews.

Jinnie:
After today's tests I made a more precise selection of what should be removed - only the views marked as "destroyed". Without it - it was possible to get "pseudo-randomly" rearranged rows after update.

gnomeontherun:
Thanks, I wanted to be as certain as we can. It helps to put more of this detail into the issue PR description as well in the future.

Jinnie:
Thanks form me too - it was good to have one more round - we found an issue. I'm postponing the unblocking with 1 more day.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid.ts
```diff
@@ -226,6 +226,13 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
         if (!this.items.smart) {
           this.items.all = this.rows.map((row: ClrDatagridRow<T>) => row.item);
         }
+        // Remove any projected rows from the displayedRows container
+        // Necessary with Ivy off. See https://github.com/vmware/clarity/issues/4692
+        for (let i = this._displayedRows.length - 1; i >= 0; i--) {
+          if (this._displayedRows.get(i).destroyed) {
+            this._displayedRows.detach(i);
```
Shijir, line 233: I think this was due to an Angular bug—something along the lines of this: https://github.com/angular/angular/issues/38648. 

When `ViewRef` destroys itself independently without going through `ViewContainerRef` as we do in our `WrappedRow`'s `OnDestroy` cycle, the `ViewContainerRef` still holds the reference to the `ViewRef`. And that results in out-of-sync views and `ViewContainerRef`.

Shijir (dismissed):

packages/angular/projects/clr-angular/src/data/datagrid/datagrid.ts
Jinnie: I think I started with the `ViewContainerRef.remove()` at first, but it was having some side effects. Then I saw that in other part of this same TS file we're using detach() and it worked better. But I'm not sure if I was doing the `.destroyed` check back then, so I'll give it a try, out of curiosity at least. 
Thanks for the detailed review, Shijir!

Jinnie: You're right, though it probably gets reclaimed after being detached, using remove gives us finer level of control and better protection against memory leaks. Updated all 3 PRs to use remove()

hippee-lee:
@Jinnie - fyi I updated the base branch to reflect the new default for the repo.

Shijir (approved):
I will approve it again. I am only a bit apprehensive we couldn't test it automatically. But I am sure @Jinnie tested well with both Ivy and Ivy off.

bbogdanov (approved):
What do you think about creating an issue or a discussion to revisit this code at some point in order to covert it with tests?

Issue #4756, opened 2020-06-29
Accessibility gaps found in datagrid example.


pavandixit93:
## Describe the bug

Below are the few accessibility issues found in datagrid.

1. aria-label is missing for `clr-dg-string-filter` component. `clr-dg-string-filter` can be activated in datagrid built in filters example. When user clicks on filter button `input` element is shown and aria-label is not added in `clr-dg-string-filter` component.
For you information, `clr-dg-number-field`  component has given aria-label but for `clr-dg-string-filter` aria-label is missing.
Issue can be found in https://clarity.design/documentation/datagrid/built-in-filters , Click on the filter icon button('Pokeman').

2. datagrid-action-overflow has role="menu" but when we open datagrid-action-overflow menu its children has no any role.
The childrens are mssing role="menuitem" role. this issue is for https://clarity.design/documentation/datagrid/single-action example. Click on the menu bar present before the User Id column.

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

## Expected behavior

_A clear and concise description of what you expected to happen._

## Versions

**App**

- Angular: [e.g. 6]
- Node: [e.g. 8.10.0]
- Clarity: [e.g. 0.12.5]

**Device:**

- Type: [e.g. MacBook]

## Additional notes

_Add any other notes about the problem here._

pavandixit93:
Hi Team, Both issues are straight forward, Please let me know if any input needed from me.

mathisscott:
@pavandixit93
Thanks for the continued interest in this issue. It is one we will be addressing but our priorities are focused on delivering 4.0 right now. If our near term objectives do not align with your product team's wants, you could submit a contribution. Updating aria attributes seems like a good candidate for helping out if you need this effort resolved quickly.

If/when you do submit a PR, though, I'd recommend pinging @chlane so that our a11y lead has a chance to vet the approach/effort. Thanks!

chlane:
@pavandixit93 Hi pavandixit93, we've reported the first issue previously in https://github.com/vmware/clarity/issues/4664

chlane:
@pavandixit93 Regarding the second issue, I can't find any elements with role="menu" on https://clarity.design/documentation/datagrid/single-action.  See the relevant HTML below.

```
[...]

Issue #4986, opened 2020-08-31
In Correct content order for clarity signpost component read by screen reader.


rahultokase:
## Describe the accessibility issue

Clarity signpost content element create the following two element which are not in order hence the correct order is not been read by the screen reader.

`<clr-signpost-content _ngcontent-tli-c4="" class="top-middle signpost-content" ng-reflect-position="top-middle" id="clr-id-5" style="position: absolute; top: 0px; bottom: auto; left: 0px; right: auto; transform: translateX(-25px) translateY(36px);"><div class="signpost-wrap"><div class="popover-pointer"></div><div class="signpost-content-body"> Lorem ipsum... </div><div class="signpost-content-header"><button class="signpost-action close" type="button" aria-label="Signpost Close" aria-controls="clr-id-5"><clr-icon shape="close" title="Close"></clr-icon></button></div></div></clr-signpost-content>`

the clarity signpost element is rendering the .signpost-content-body element first and then .signpost-action close element. Hence the content is not read in the order visible to the human eyes. The correct order should be close button-> clarity signpost body by press Down arrow key.

_If this is a VMware product, do not fill this out! Instead consult with the accessibility team about the issue so we can consolidate all VMware related issues and track them separately._

_Provide clear description of the accessibility issue and the impact to the user. Is this only in certain use cases? Is it confusing or broken?_
Vision-impaired users may get confused when signpost content is been shown to user.

## WCAG Guideline

_What WCAG guideline is being violated here and how? You should reference a guideline from https://www.w3.org/TR/WCAG21/_
WCAG 2.1 -> 1.3.1 Info and Relationships (Level A):  
Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.  

## Recommended Solution:  
Move the <div class="signpost-content-header"> up, to make it as the first element in <div class="signpost-content-body">

## How to reproduce

_Please provide a link to a reproduction scenario using one of the [Clarity Stackblitz templates](https://stackblitz.com/@clr-team). Reports without a clear reproduction may not be prioritized until one is provided._
https://stackblitz.com/edit/angular-8-app-clarity-signpost-issue?file=src%2Fapp%2Fapp.component.html

1.  Go to '...'
2.  Click on '....'
3.  Scroll down to '....'
4.  See error

## Expected behavior

_A clear and concise description of what you expected to happen._
\*Recommended Solution:  
Move the <div class="signpost-content-header"> up, to make it as the first element in <div class="signpost-content-body">

## Versions

[...]

PR #5131, opened 2020-10-09
fix(datagrid): add aria-label to string filter input

amellnik:
Add a new common string for "Search" and use it to add an aria-label to the datagrid default string filter input.  This string can probably be used elsewhere in the future.

Partially closes: #4756

~WIP to resolve some build issues that I think are unrelated.~

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The datagrid default string filter does not have an aria-label or placeholder.  

Issue Number: #4986 

## What is the new behavior?

This adds the aria label via a new common string.  It doesn't populate the placeholder but that would be easy to do.  

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -380,6 +380,7 @@ export interface ClrCommonStrings {
     detailPaneEnd: string;
     detailPaneStart: string;
     expand: string;
+    filterItems: string;
     firstPage: string;
     hide: string;
     info: string;
@@ -2149,6 +2150,7 @@ export declare class DatagridPropertyStringFilter<T = any> implements ClrDatagri
 }
 
 export declare class DatagridStringFilter<T = any> extends DatagridFilterRegistrar<T, DatagridStringFilterImpl<T>> implements CustomFilter, AfterViewInit, OnDestroy {
+    commonStrings: ClrCommonStringsService;
     set customStringFilter(value: ClrDatagridStringFilterInterface<T> | RegisteredFilter<T, DatagridStringFilterImpl<T>>);
     filterContainer: ClrDatagridFilter<T>;
     filterValueChange: EventEmitter<any>;
@@ -2156,7 +2158,7 @@ export declare class DatagridStringFilter<T = any> extends DatagridFilterRegistr
     open: boolean;
     get value(): string;
     set value(value: string);
-    constructor(filters: FiltersProvider<T>, domAdapter: DomAdapter, smartToggleService: ClrPopoverToggleService);
+    constructor(filters: FiltersProvider<T>, domAdapter: DomAdapter, commonStrings: ClrCommonStringsService, smartToggleService: ClrPopoverToggleService);
     ngAfterViewInit(): void;
     ngOnDestroy(): void;
 }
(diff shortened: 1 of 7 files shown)


bdryanovski (changes requested):
You gonna need to update the `clr-angular.d.ts` file by running 

```bash
$ cd packages/angular
$ yarn golden:fix
```
To fix the issue with the failing test.

Also a unit-test will be nice to cover the use of the new `attr.aria-label` and the use of `commonStrings.keys.search`

gnomeontherun:
I'm not opposed to this, but it doesn't really solve the need that we have for the labels to be fully descriptive. Just 'Search' is not helpful for screen readers and we need to come up with a solution that allows for customized labels per use. 

Tagging @chlane and @scroniser for visibility.

chlane:
@gnomeontherun Agreed, the label should say what is being searched. That content should come from the Clarity consumer.

hippee-lee:
Hi @amellnik - fyi we changed the default branch for the Clarity repo to `next`. 
I updated this pull request for the new default branch. Please reach out if you run into any issues.

Thanks for another contribution!!

amellnik:
Should we add this as the default behavior and add an @Input to for the user to override the placeholder text?  Is there a better default phrase that we should use than just "Search"?  The numeric filter currently has "Min value" and "Max value" for placeholders.

amellnik:
I had a really hard time getting this to build, it turned out that I needed to delete my local copy, reclone, `yarn install` and then run `golden:fix` prior to running any build steps in order to get it to work, not sure why.  

Added a test for the `aria-label`, waiting on feedback on:
* Adding a placeholder to go with the aria-label
* Adding an `@Input` so downstream users can override it
* Change `Search` to something else, maybe `Filter` or something?

bdryanovski (changes requested):
Thanks, @amellnik for the contribution, 

I didn't find anywhere the `@Input` change that you mention - or this is no longer relevant? 

I have few minnor changes to add.

Like @gnomeontherun mention, the `search` key is not that well named and may point people in the wrong direction. In this case, we are opening filter so maybe the default text must be something close to `Filter items` then `Search`.

In the current version of the `fix` changing the `CommonStrings` value, it will affect every Datagrid and Filter at the same time. As developers, we could provide additional information for example `Filter users by there first name` (maybe over-kill). So for doing this we gonna need some way to pass this value `@Input` is an option.  And still have some default value for the cases that we can't or we don't want to provide additional information.

amellnik:
Hi @bdryanovski,

I didn't add the @Input in the first pass, but will do so now if we're ok with that approach.  I'll also mention and show it in the documentation.  `Filter items` does seem clearer, I'll update the default string to that.

gnomeontherun:
I don't want to add a new @Input for the override string, we need a different pattern but that is beyond the scope of this PR. I think we can focus on just having it use the global string and worry about the localized override later.

amellnik:
Updated to use "Filter items" for the aria-label.  This is still not populating a placeholder but that could easily be added if we want.

amellnik:
As per Jeremy, not adding an `@Input` for now.

bdryanovski (approved):

amellnik:
If you found this helpful I'd appreciate it if you would add the hacktoberfest-accepted -- I'm attempting to mine a shirt. Cheers -A

gnomeontherun (approved):

PR #5137, opened 2020-10-12
fix(datagrid): give all action-item items the menuitem role

amellnik:
Currently the documentation shows adding buttons with this class, but not giving them the `menuitem` role which is required since the wrapping container has a `menu role`.  We could update the documentation to do this, but this approach adds a tiny directive that does it for the user.  Also added a test for this.

Partially fixes:  #4756

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Documentation for the datagrid single action shows adding buttons to the menu, but they don't have the `menuitem` role which is a 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-action-item.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-action-item.ts
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive } from '@angular/core';
+
+/**
+ * @remark
+ * This directive is used to identify action buttons within the single row action menu and to give
+ * them the 'menuitem' aria role.
+ */
+@Directive({
+  selector: '[clrDgActionItem]',
+  host: {
+    '[class.action-item]': 'true',
+    role: 'menuitem',
+  },
+})
+export class ClrDatagridActionItem {}
--- a/apps/website/.vuepress/public/demos/datagrid/single-action.html
+++ b/apps/website/.vuepress/public/demos/datagrid/single-action.html
@@ -10,8 +10,8 @@
   this pattern in both selectable and non-selectable datagrids. Add a
   <code class="clr-code">clr-dg-action-overflow</code> inside a <code class="clr-code">clr-dg-row</code>. The content
   inside of it will be projected as an action menu which will toggle when the user clicks on the ellipsis icon as shown
-  below. We recommend that the menu items be buttons with a class <code class="clr-code">.action-item</code> as in the
-  example.
+  below. We recommend that the menu items be buttons with the <code class="clr-code">clrDgActionItem</code> directive as
+  in the example.
 </p>
 
 <p>
@@ -41,8 +41,8 @@
 
   <clr-dg-row *clrDgItems="let user of users" [clrDgItem]="user">
     <clr-dg-action-overflow>
-      <button class="action-item" (click)="onEdit(user)">Edit</button>
-      <button class="action-item" (click)="onDelete(user)">Delete</button>
+      <button clrDgActionItem (click)="onEdit(user)">Edit</button>
+      <button clrDgActionItem (click)="onDelete(user)">Delete</button>
     </clr-dg-action-overflow>
     <clr-dg-cell>{{user.id}}</clr-dg-cell>
     <clr-dg-cell>{{user.name}}</clr-dg-cell>
--- a/apps/website/components/datagrid/README.md
+++ b/apps/website/components/datagrid/README.md
@@ -674,7 +674,7 @@ Depending on the role of certain batch actions, you can choose to break button b
 
 ### Single Action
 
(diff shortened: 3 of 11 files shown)


amellnik:
This is failing the netlify presubmit checks but I don't have access to the logs, not sure how to debug them.

mathisscott:
@amellnik 
I don't know why you are seeing this. We aren't usually running Netlify as a build check.

@hippee-lee : Do you know what's going on?

hippee-lee:
Yeah - I saw this yesterday, here is what I think is going on. The Netlify checks get picked up when it (Netlify) tries to build the website with that CI/CD infrastructure. This started after I updated the default branch and set up the new site in netlify. However, on github, those checks are not required in the `next` branch protections so it won't block the PR once it has two :shipit:'s 

_I think_ I removed the notifications coming from Netlify yesterday so these checks shouldn't show up in future PR's. I'll continue to monitor to make sure nothing gets blocked unnecessarily.

amellnik:
Is there anything I should do to retrigger the CI/CD pipeline or is it OK for now?

hippee-lee:
Hi @amellnik - CI/CD should be ok. The checks that were picked up by mistake are not required for merging. Once there are two :shipit:'s it should be ok. Ping me if you find different.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-action-item.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { Directive, ElementRef } from '@angular/core';
+
+/**
+ * @remark
+ * This directive is used to identify action buttons within the single row action menu and to give
+ * them the 'menuitem' aria role.
+ */
+@Directive({
+  selector: '.action-item',
```
hippee-lee: I think we need to be careful with this. Even though we only use `action-item` in the main-renderer.spec.ts on line 334 for the Clarity codebase, this could cause consumers problems if they also have elements with `class="menu-item"` on anything. If we go this route we might have to prefix it to try and create a namespace around it and then change the classes inside the datagrid.

gnomeontherun: Good call Matt. Unfortunately Angular doesn't allow you to use any CSS selector, so it would not be possible to make this directive work better without targetting potentially any `class="action-item"`.

I think a better solution here is to make this an attribute directive, which would use host bindings to set the class and the role instead of trying to reach in and use the native Element below (which would break SSR currently).

This is untested, but basically this would be the class:

```
@Directive({
  selector: '[clrDgActionItem]',
  host: {
    '[class.action-item]': true,
    role: 'menuitem',
  }
})
export class ClrDatagridActionItem {}
```

Then the API would be 

```
<clr-dg-action-overflow>
        <button clrDgActionItem (click)="onEdit(user)">Edit</button>
        <button clrDgActionItem (click)="onDelete(user)">Delete</button>
</clr-dg-action-overflow>
```

This is not breaking the existing API, but for anyone who wants this a11y fix they'd update their usage.

gnomeontherun (changes requested):

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-action-item.ts
amellnik: I was also thinking about the generic class names but unfortunately it's hard to change them at this point.  I'm also worried about relying on users to do anything to fix a11y issues.  

What about having `clr-dg-action-item` dynamically add the menuitem role to all buttons within it using `@ContentChildren`?  If that seems too invasive I can take the approach above.

gnomeontherun: Frankly, this bug is easily resolved by apps who can just add the right role to their buttons. Anything we do is a bonus.

This approach might also work, but needs to be careful of dynamic rendering scenarios. It must watch for changes to the list. Also, the direct nativeElement access below will break in server side rendering, so we need to use the Angular renderer.

I worry that this might have a problem or two around dynamic items because of lifecycles. We see this in a few places, like list of tabs are rendered too late and not picked up by the screen reader. If that is the case, then it would be better for apps to directly apply the roles since that would render the menu items in regular time vs in a potential later cycle (due to renderer).

amellnik: All good points, updated with your suggestion above.

bdryanovski (changes requested):
I think that we need to add a unit-test for `'[clrDgActionItem]'` - I know it's a super small directive but it will be helpful when we start refactoring ...

Also notice that there is a merge conflict with one of the popovers files.

If exclude that everything for me is OK :)

mathisscott:
@gnomeontherun : What is the status with this PR?

amellnik:
I'm afraid this fell off my radar -- I'm currently in the process of moving
but can finish it up after I'm reunited with my computer unless someone
beats me to it.

gnomeontherun:
Hi @amellnik, given the time this has been open it is time to consider it stale and close for housekeeping. If you are able to update and refresh the PR, we are happy to take another look!

PR #5190, opened 2020-10-22
icons: replace angular icons with core ones

bbogdanov:
We will need to update the website and provide guidelines on how to migrate to `cds-icon`. I prefer doing that after the code changes are approved. 

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [x] Build related changes
- [ ] CI related changes

The clarity angular icons are used everywhere in the components

## What is the new behavior?

All angular icons are replaced with the ones provided by `core`

- [ ] Yes
- [x] No, maybe 🤷‍♂️


packages/angular/projects/clr-angular/src/data/datagrid/_datagrid.clarity.scss
```diff
@@ -59,7 +59,7 @@
     }
 
     .datagrid-expandable-caret-icon {
-      @include css-var(color, clr-datagrid-icon-color, $clr-datagrid-icon-color, $clr-use-custom-properties);
+      @include css-var(--color, clr-datagrid-icon-color, $clr-datagrid-icon-color, $clr-use-custom-properties);
```
mathisscott: The `css-var` does not take `--` in front of its first argument.

mathisscott (changes requested):

packages/angular/projects/clr-angular/src/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -192,12 +194,19 @@
       text-align: left;
 
       .nav-group-trigger-icon {
+        //cds-icon {
```
mathisscott: Why was this commented code added? Is it still needed?

packages/angular/projects/clr-angular/src/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -192,12 +194,19 @@
...
+        //      clr-vertical-nav-item-color
+        //      $clr-vertical-nav-item-color,
+        //      $clr-use-custom-properties
+        //  );
+        //}
         //Display
         flex-shrink: 0;
 
         //Dims
-        height: $clr-vertical-nav-item-height;
```
mathisscott, line 199: Why were these sizing styles removed?

packages/angular/projects/clr-angular/src/utils/_theme.dark.icons.clarity.scss
```diff
@@ -1,8 +1,8 @@
-// Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
-clr-icon {
+cds-icon {
   &.is-green,
```
mathisscott: These classnames don't exist in 5.0. They are legacy clr-icon classes.

packages/angular/projects/clr-angular/src/wizard/wizard-page.spec.ts
```diff
@@ -133,18 +133,18 @@ class TemplateTestComponent {
...
+        <cds-icon shape="cloud" solid></cds-icon>
       </clr-wizard-header-action>
 
       <clr-wizard-page #viewTestWizardPageOne [id]="testId" [clrWizardPagePreventDefaultCancel]="preventCancel">
         <ng-template clrPageTitle>View Page 1</ng-template>
 
         <ng-template clrPageHeaderActions>
           <clr-wizard-header-action (actionClicked)="headerActionClicked($event)" id="bell">
-            <clr-icon shape="bell" class="has-badge"></clr-icon>
+            <cds-icon shape="bell" class="has-badge"></cds-icon>
```
mathisscott: `has-badge` class is no longer in 5.0

packages/angular/projects/dev/src/app/button-group/static/icon-buttons/icon-button-group.ts
```diff
@@ -1,12 +1,12 @@
...
- * Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import { Component } from '@angular/core';
 
 @Component({
-  selector: 'clr-icon-button-group-demo',
+  selector: 'cds-icon-button-group-demo',
```
mathisscott: does this selector need to change? the `cds-icon-button` is a different component from the `cds-icon`

packages/angular/projects/dev/src/app/drag-and-drop/custom-ghost-and-handle.demo.html
```diff
@@ -33,12 +33,12 @@ <h3 class="card-title">Custom ghost and handle card</h3>
     <div class="clr-col-lg-8 clr-col-md-12 clr-col-sm-12 clr-col-12">
       <div class="card" clrDraggable>
         <clr-draggable-ghost *clrIfDragged>
-          <clr-icon
+          <cds-icon
             shape="event"
             size="72"
             class="is-solid is-success"
```
mathisscott: `is-solid` and `is-success` are no longer in 5.0

packages/angular/projects/dev/src/app/dropdown/dropdown-angular-close-item-false.demo.html
```diff
@@ -7,8 +7,8 @@
 <div class="clr-example squeeze dropdown-demo">
   <clr-dropdown [clrCloseMenuOnItemClick]="false">
     <button clrDropdownTrigger aria-label="dropdown demo button">
-      <clr-icon shape="error" class="is-error" size="24"></clr-icon>
-      <clr-icon shape="caret down"></clr-icon>
+      <cds-icon shape="error" class="is-error" size="24"></cds-icon>
```
mathisscott: `is-error` is not in 5.0

packages/angular/projects/dev/src/app/dropdown/dropdown-static-icon-toggle.demo.html
```diff
@@ -7,8 +7,8 @@
 <div class="clr-example squeeze demo-menu-height dropdown-demo">
   <div class="dropdown bottom-left open">
     <button class="dropdown-toggle">
-      <clr-icon shape="error" class="is-error" size="24"></clr-icon>
-      <clr-icon shape="caret down"></clr-icon>
+      <cds-icon shape="error" class="is-error" size="24"></cds-icon>
```
mathisscott: `is-error`  is not in 5.0

packages/angular/projects/dev/src/app/iconography/icon-colors.demo.html
```diff
@@ -1,14 +1,14 @@
...
 
 <span class="clr-example">
-  <clr-icon shape="info-circle"></clr-icon>
-  <clr-icon shape="info-circle" class="is-highlight"></clr-icon>
-  <clr-icon shape="info-circle" class="is-error"></clr-icon>
-  <clr-icon shape="info-circle" class="is-warning"></clr-icon>
-  <clr-icon shape="info-circle" class="is-success"></clr-icon>
-  <clr-icon shape="info-circle" class="is-info"></clr-icon>
+  <cds-icon shape="info-circle"></cds-icon>
+  <cds-icon shape="info-circle" class="is-highlight"></cds-icon>
```
mathisscott: all of these classes are not in 5.0

packages/angular/projects/dev/src/app/iconography/icon-inverse-color.demo.html
```diff
@@ -1,9 +1,9 @@
 <!--
-  ~ Copyright (c) 2016-2018 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
 
 <span class="clr-example inverse">
-  <clr-icon shape="info-circle" class="is-inverse"></clr-icon>
+  <cds-icon shape="info-circle" class="is-inverse"></cds-icon>
```
mathisscott: `is-inverse`  is not in 5.0

packages/angular/projects/dev/src/app/iconography/icon-orientation.demo.html
```diff
@@ -7,25 +7,25 @@
...
 </span>
 
 <h6>B. Rotated: Set the icon orientation in the <code>dir</code> attribute</h6>
 
 <span class="clr-example">
-  <clr-icon shape="caret" dir="up"></clr-icon>
-  <clr-icon shape="caret" dir="right"></clr-icon>
-  <clr-icon shape="caret" dir="down"></clr-icon>
-  <clr-icon shape="caret" dir="left"></clr-icon>
+  <cds-icon shape="angle" dir="up"></cds-icon>
```
mathisscott: `dir`  is not in 5.0

packages/angular/projects/dev/src/app/iconography/icon-variants.demo.html
```diff
@@ -1,40 +1,40 @@
...
 
 <span class="clr-example">
-  <clr-icon shape="bell"></clr-icon>
-  <clr-icon shape="bell" class="has-badge"></clr-icon>
-  <clr-icon shape="bell" class="has-badge--success"></clr-icon>
-  <clr-icon shape="bell" class="is-info has-badge--error"></clr-icon>
-  <clr-icon shape="bell" class="is-solid"></clr-icon>
-  <clr-icon shape="bell" class="is-solid has-badge--info"></clr-icon>
+  <cds-icon shape="bell"></cds-icon>
+  <cds-icon shape="bell" class="has-badge"></cds-icon>
```
mathisscott: all of these classes were removed in 5.0

packages/angular/projects/dev/src/app/vertical-nav/direct-icon/vertical-nav-icon.demo.html
```diff
@@ -20,15 +20,15 @@
...
+          <cds-icon shape="cog"></cds-icon>
         </a>
       </div>
     </header>
     <div class="content-container">
       <clr-vertical-nav [clrVerticalNavCollapsible]="true" [(clrVerticalNavCollapsed)]="collapse">
         <ng-container *ngFor="let item of case.items">
           <a clrVerticalNavLink href="javascript:void(0)" *ngIf="!item['children']">
-            <clr-icon [attr.shape]="item.icon" *ngIf="item['icon']" clrVerticalNavIcon></clr-icon>
+            <cds-icon [attr.shape]="item.icon" *ngIf="item['icon']" clrVerticalNavIcon></cds-icon>
```
mathisscott, line 31: Is `[attr.shape]` still necessary? I was under the impression that it was not.

packages/angular/projects/dev/src/app/wizard/wizard-header-actions.demo.html
```diff
@@ -55,14 +55,14 @@
 
     <ng-template clrPageHeaderActions>
       <clr-wizard-header-action (actionClicked)="headerActionClicked($event)" id="user">
-        <clr-icon shape="user" [class.has-badge]="userActive" [class.is-solid]="userActive"></clr-icon>
+        <cds-icon shape="user" [class.has-badge]="userActive" [class.is-solid]="userActive"></cds-icon>
```
mathisscott: These classes are not in 5.0

packages/angular/projects/dev/src/main.ts
```diff
@@ -5,14 +5,11 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
-// import { runCssVarsPolyfill } from '@clr/core';
 
 if (environment.production) {
   enableProdMode();
 }
 
-// runCssVarsPolyfill();
```
mathisscott: 👍

mathisscott (changes requested):

packages/angular/projects/dev/src/app/vertical-nav/direct-icon/vertical-nav-icon.demo.html
bbogdanov: PR is still in progress. It is a Draft one. I will go through every single `cds-icon` attribute tomorrow so I remove the unnecessary usages or replace them with the proper ones. For example, one thing I did today was to replace `class="is-solid"` with `solid` property.

packages/angular/projects/dev/src/app/button-group/static/icon-buttons/icon-button-group.ts
bbogdanov: Agree! Thanks!

packages/angular/src/stories/card/basic.html
```diff
@@ -41,7 +41,7 @@
               <div class="dropdown top-left" [class.open]="ddOpen">
                 <button class="dropdown-toggle btn btn-sm btn-link">
                   Dropdown Actions
-                  <clr-icon shape="caret down"></clr-icon>
+                  <cds-icon shape="caret down"></cds-icon>
```
mathisscott: `down` is not in 5.0

mathisscott (changes requested):

packages/angular/projects/dev/src/app/vertical-nav/direct-icon/vertical-nav-icon.demo.html
bbogdanov: @mathisscott  The `[attr.shape]` is needed because if we do`[shape]` the angular compiler says the `[shape] input` doesn't exist. Another way to doing this is `shape="{{ icon name }}"` but I believe the `[attr.shape]` is much better.

packages/angular/projects/clr-angular/src/clr-angular.module.ts
```diff
@@ -48,4 +72,32 @@ import { ClrTimelineModule } from './timeline/timeline.module';
     ClrTimelineModule,
   ],
 })
-export class ClarityModule {}
+export class ClarityModule {
```
bbogdanov: @gnomeontherun Need a review on this one.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -39,14 +39,15 @@ import { ClrPopoverPositionService } from '../../utils/popover/providers/popover
 import { ClrPopoverEventsService } from '../../utils/popover/providers/popover-events.service';
 import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
 import { DetailService } from './providers/detail.service';
+import { Directions } from '@clr/core/internal';
```
bbogdanov: @gnomeontherun Need a special review for this file.

It brings a little breaking change in case someone extends the `sort` function and changes the `shape` icon of the sorting. I hope nobody does that but it is a case.

packages/angular/projects/clr-angular/src/layout/vertical-nav/_vertical-nav.clarity.scss
bbogdanov: image
Take a look at the `angle`. This is how it looks like when the height and width are applied. Looks fine without them.

packages/angular/projects/clr-angular/src/timeline/enums/timeline-step-state.enum.ts
```diff
@@ -6,8 +6,8 @@
 
 export enum ClrTimelineStepState {
   NOT_STARTED = 'not-started',
-  CURRENT = 'current',
+  CURRENT = 'info',
   PROCESSING = 'processing',
```
bbogdanov: I am not sure what color the `processing` has to be, maybe is `info` but I need advice here.

mathisscott: The current step is blue. Info is blue. But these colors are driven by the CSS. So I would think you would want to update those selectors? Or the styles? I don't think you need to update the enum. That's not going to solve it.

packages/angular/projects/clr-angular/package.json
```diff
@@ -11,6 +11,7 @@
   "peerDependencies": {
     "@angular/common": "^10.0.0",
     "@angular/core": "^10.0.0",
+    "@clr/core": "@VERSION",
```
bbogdanov: Update to minimum version usage

packages/angular/projects/clr-angular/src/clr-angular.module.ts
```diff
@@ -48,4 +84,45 @@ import { ClrTimelineModule } from './timeline/timeline.module';
...
+  constructor() {
+    // Icons used in the built-in components
+    ClarityIcons.addIcons(
```
bbogdanov: Per component module

packages/angular/projects/clr-angular/src/timeline/enums/timeline-step-state.enum.ts
bbogdanov: https://github.com/bbogdanov/clarity/blob/migrate-to-core-icons/packages/angular/projects/clr-angular/src/timeline/timeline-step.ts#L21

https://github.com/bbogdanov/clarity/blob/migrate-to-core-icons/packages/angular/projects/clr-angular/src/timeline/timeline-step.ts#L40

I believe two will lines will give some clarity on why I change it like that. There can be a better way but I might introduce a breaking change. What do you think?

packages/angular/projects/clr-angular/src/accordion/accordion.module.ts
```diff
@@ -31,4 +32,8 @@ const declarations = [
   declarations: [...declarations],
   exports: [...declarations],
 })
-export class ClrAccordionModule {}
+export class ClrAccordionModule {
+  constructor() {
+    ClarityIcons.addIcons(angleIcon, exclamationCircleIcon, checkCircleIcon);
```
hippee-lee, line 37: 👍🏼

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -322,7 +330,15 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
...
   @Output('clrDgSortOrderChange') public sortOrderChange = new EventEmitter<ClrDatagridSortOrder>();
 
+  /**
+   * @deprecated
+   *
+   * Use `sortDirection` to indentify the sort direction
+   */
   public sortIcon: string | null;
+
+  public sortDirection: Directions | null;
```
hippee-lee: Directions has some values that don't quite make sense if a consumer were to use this. They could still pass 'left' or 'right'.

Perhaps just keep it simple and use only the string literal types we want here + null. e.g

```
public sortDirection: 'up' | 'down' | null;
```

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -336,7 +352,8 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
     // setting the private variable to not retrigger the setter logic
     this._sortOrder = this._sort.reverse ? ClrDatagridSortOrder.DESC : ClrDatagridSortOrder.ASC;
     // Sets the correct icon for current sort order
-    this.sortIcon = this._sortOrder === ClrDatagridSortOrder.DESC ? 'arrow down' : 'arrow';
+    this.sortDirection = this._sortOrder === ClrDatagridSortOrder.DESC ? 'down' : 'up';
```
hippee-lee: This looks ok to me. The api for calling this function is the same. the sort fn is using `this.sortDirection` internally and the only reason I can see for making `this.sortDirection` public is to use it in the template. 👍🏼

packages/angular/projects/clr-angular/src/forms/range/range-container.ts
```diff
@@ -22,13 +22,20 @@ import { IfControlStateService } from '../common/if-control-state/if-control-sta
       <div class="clr-range-wrapper" [class.progress-fill]="hasProgress">
         <ng-content select="[clrRange]"></ng-content>
         <span *ngIf="hasProgress" class="fill-input" [style.width]="getRangeProgressFillWidth()"></span>
-        <clr-icon
+        <cds-icon
           *ngIf="showInvalid"
           class="clr-validate-icon"
+          status="danger"
           shape="exclamation-circle"
```
hippee-lee, line 29: Why isn't this module adding them like the others do?
```
export class ClrRangeModule {
  constructor() {
    ClarityIcons.addIcons(exclamationCircleIcon, checkCircleIcon);
  }
}
```

packages/angular/projects/clr-angular/src/timeline/enums/timeline-step-state.enum.ts
hippee-lee: Processing is a `clr-spinner` component when the timeline step state is `procesing`. I would expect that the color will be updated there when I get tothat part ofthe review :) 

I agree with Scott. I don't think these values need to change. If they do, we should update the Enum constant value.

packages/angular/projects/clr-angular/src/utils/popover/README.md
```diff
@@ -41,7 +41,7 @@ This is what it looks like from the implementing component perspective:
   <header class="header-4" role="heading">
     Header
     <button role="button" class="btn btn-link" clrPopoverCloseButton>
-      <clr-icon shape="close" size="36" class="is-inverse"></clr-icon>
+      <clr-icon shape="window-close" size="36" class="is-inverse"></clr-icon>
```
hippee-lee: SHould this be `<cds-icon shape="window-close" size="36" class="is-inverse"></cds-icon>`?

hippee-lee (changes requested):
Hey Bogdan, this is a huge effort. Thanks for managing it. 

I fond a few things in the angular codebase the I believe need attention and added some context to previous discusiion in a few places. 

Overall the angular part looks pretty close, jsut a few things.

I also went through the dev app and story book code but didn't run them locally yet. I will do that next week and update here if I see anything. 

One thing I do when the git actions stop running on github for me =like I see them at today is to _change_ the history with `git commit --amend` and then force push that. It ususally fixes the actions on gh.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
bbogdanov: Agree!

packages/angular/projects/clr-angular/src/forms/range/range-container.ts
bbogdanov: Somehow I missed that module. Thank you for noticing it.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
bbogdanov: I see even a better solution at the moment. What about if we make a `get` for the `sortDirection` and change the property to `private _sortDirection` inside the component? That way we will be sure the property will not be changed from outside. (yea when used in JavaScript everything is possible but you know what I mean xD)

packages/angular/projects/clr-angular/src/timeline/enums/timeline-step-state.enum.ts
bbogdanov: [Commit](https://github.com/vmware/clarity/pull/5190/commits/a32306bf19bc025b2a3424567f35865ed8702860) I believe this solution we will agree is much better than the previous one

packages/angular/projects/clr-angular/src/timeline/providers/timeline-icon-attribute.service.ts
```diff
@@ -43,4 +48,8 @@ export class TimelineIconAttributeService {
   public getIconShape(step: ClrTimelineStepState): string {
     return this.attributeMap.get(step).iconShape;
   }
+
+  public getIconStatus(step: ClrTimelineStepState): string {
+    throw this.attributeMap.get(step).iconStatus;
```
bdryanovski: xa-xa-xa

packages/angular/projects/clr-angular/src/layout/vertical-nav/_vertical-nav.clarity.scss
```diff
@@ -153,15 +153,18 @@
...
       //Others
-      @include css-var(
-        color,
-        clr-vertical-nav-toggle-icon-color,
-        $clr-vertical-nav-toggle-icon-color,
-        $clr-use-custom-properties
-      );
+      cds-icon[shape='angle-double'] {
+        @include css-var(
+          --color,
```
mathisscott: Having the `--` in front of the variable name may not work as you'd expect...

packages/angular/projects/clr-angular/src/timeline/interface/icon-attribute.iterface.ts
```diff
@@ -4,7 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
```
mathisscott, line 4: Filename is `icon-attribute.iterface.ts` instead of `icon-attribute.interface.ts`

packages/angular/projects/dev/src/app/dropdown/dropdown-static-icon-toggle.demo.html
```diff
@@ -7,8 +7,8 @@
 <div class="clr-example squeeze demo-menu-height dropdown-demo">
   <div class="dropdown bottom-left open">
     <button class="dropdown-toggle">
-      <clr-icon shape="error" class="is-error" size="24"></clr-icon>
-      <clr-icon shape="caret down"></clr-icon>
+      <cds-icon shape="error" status="danger" size="18"></cds-icon>
```
mathisscott, line 10: Is the change from size 24 to 18 expected?

mathisscott (changes requested):

mathisscott (approved):

packages/angular/projects/dev/src/app/dropdown/dropdown-static-icon-toggle.demo.html
bbogdanov: Yea, as far as I remember the visual tool mark that the `clr-icon` one was a little bit smaller than the update one so I just lower the size to satisfy the visual checker

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -336,7 +336,11 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
    */
   public sortIcon: string | null;
 
-  public sortDirection: 'up' | 'down' | null;
+  private _sortDirection: 'up' | 'down' | null;
+
+  public get sortDirection(): 'up' | 'down' | null {
```
hippee-lee, line 341: 👍🏼

hippee-lee (approved):

packages/angular/package.json
```diff
@@ -49,6 +49,7 @@
     "@angular/platform-browser-dynamic": "11.0.0-rc.2",
     "@angular/router": "11.0.0-rc.2",
     "css-vars-ponyfill": "2.3.1",
+    "ramda": "^0.27.0",
```
bbogdanov: @mathisscott @hippee-lee take a look at this one. I am pretty sure there can be a better fix for this one. Currently, the angular application refers to `@clr/core/icon` and depends on typescript to resolve from where to get the files `dist/` or `packages/core/src`.  It takes the files from `pacakges/core/src` so when the angular build starts it expects the cds-icon files references to packages to point to `packages/angular/node_modules` instead to `packages/core/node_modules` and in angular node_modules there is no `ramda` package installed

mathisscott: `ramda` is a dependency for core. think of it like `lodash` or `underscore`. why wouldn't you have it as a dependency? you have `css-vars-ponyfill` listed on the line above it. same sort of thing.

bbogdanov: When `angular:build` script starts it also builds the `@clr/core/icon` files. When it starts to build the `core` files the build cannot find a `ramda` package and the build fails. I agree this is a dirty fix and that's why I am asking for little assistance to find a better fix. I've been struggling to find a fix to satisfy the build step in GitHub actions for nearly 4 days and this fix is one of the solutions

bbogdanov: What I aiming to happen now is to let the typescript when it builds the `@clr/angular` and sees import from `@clr/icon` the path of `@clr/core` to be `/dist/core` instead of `packages/core/src/*` like it is declared in [here](https://github.com/vmware/clarity/blob/next/tsconfig.json#L30)

coryrylan: All dependencies of Core would also become dependencies of Angular. Typically applications do not need to worry about this because yarn/npm will resolve those automatically for them. Yarn would typically handle this in apps but in the monorepo we have to use `nohoist` for the angular projects since it does not play well with yarn. I think`@clr/angular` can take the same strategy as `@clr/react`.

Currently projects like `@clr/react` which depend on Core, will point to the dist directly like this https://github.com/vmware/clarity/blob/next/packages/react/package.json#L20 This gets around all the weird tsconfig/yarn resolution issues usually.

To make yarn happy when installing prior to the build we create a empty directory in the pre-install
https://github.com/vmware/clarity/blob/next/package.json#L85 

During the core build we copy the build into the node_modules to keep it updated without running yarn again. You can see this copy step here https://github.com/vmware/clarity/blob/next/packages/core/package.js#L86

bbogdanov: I guess I am doing it wrong. I updated the `package.json` of `@clr/angular` to get the `@clr/core` from "./packages/core/dist/core", also updated the `package.js` of `@clr/core` to also copy the files in `node_modules` of `@clr/angular`. Still facing the same issue as previous builds

coryrylan: I think I know why, the angular cli projects dont hoist or share dependencies with the rest of the Repo which is why the react apps work but angular does not. Its a tricky problem with the monorepo we havent really solved in a generic way yet. I can dig into this some more after the workshop.

bbogdanov: Ok, I believe for now the best way to connect core with everything else is to provide `@clr/core` as a dep. I believe also that providing a package manually inside node_modules is just counterproductive. 

What I figure out yesterday when I did the experiment on what @coryrylan suggested is that:
1. Running the build of `@clr/angular`
2. The builder gets to some `@clr/core/icon` import line
3. The builder tries to resolve the imports inside `@clr/core/icons` by looking into `packages/angular/node_modules`
4. Builder fails because the packages that `@clr/core` depends on are no provided in `@clr/angular` node_modules

We have the same issue inside the website and what we did there is a workaround with webpack where we manually provide where to look for certain packages but this solution is also tricky and leaves a room the build to fail if the `core` package deps grow at some point.

In the end, when we publish the `@clr/angular` package it expends on peerDependency of `@clr/core` so the user has to install `@clr/core` anyway. Having this in mind I believe having it as `dependency` inside our internal `angular` package is fine for now cause we only use this dependency internally and we can update it at any time without affecting our users.

[here](https://github.com/vmware/clarity/pull/5190/files#diff-d08139eef05218816a5597154199c170727c5bcd22546a9e6367735e6b223fb0R51) is the solution I am referring to.

gnomeontherun:
Is this blocked still or just needs to have conflicts resolved?

bbogdanov:
> Is this blocked still or just needs to have conflicts resolved?

Conflicts have to be resolved.

Issue #4754, opened 2020-06-26
Screen Reader dialog boundaries are not communicated for datagrid show/hide column


pratikgala:
@
## Describe the bug

Accessibility Issue.
Dialog boundaries are not communicated for datagrid show/hide column,
Example of this include:
- Show Columns dialog. The div for the dialog contains role of a dialog. However, it does not provide name.

[User Impact]
When this is not done, it is not clear to screen readers when they are leaving or entering a dialog when using the virtual cursor.

[Code Reference]
`<div class="column-switch clr-popover-content ng-star-inserted" clrfocustrap="" role="dialog" tabindex="0" id="clr-id-833" style="top: 334.5px; left: 313px;">...</div>`

## How to reproduce

https://clarity.design/documentation/datagrid/hide-show

1.  Use a screen reader like JAWS or NVDA.
2.  Go to clarity datagrid
3.  Click on show/hide
4.  Scroll down to the datagrid
5.  Click open the dialog to show/hide column
 Inspect the dialog code, the dialog code does not provide role name for the dialog.
 The dialog does not provided aria-labelledby and aria-label attributes.

## Expected behavior

[Recommendation]
Ensure that the dialog is properly labeled. Make use of role=dialog and ensure that the dialog is named by using aria-label (if no visual title is present) or aria-labelledby (if a visual title is rendered within the dialog).

[Compliant Code Example]
aria-label:
```
<div role="dialog" aria-label="My Dialog Title">
 <p>Lorem ipsum dolor set... /p>
</div>
```

aria-labelledby:
```
<div role="dialog" aria-labelledby="myDialogTitle">
 <h3 id="myDialogTitle">Dialog Title</h3>
 <p>Lorem ipsum dolor set... /p>
</div>
```

## Versions

**App**

- Angular: [8.1.0]
- Node: [10]
- Clarity: [2.4.2]

**Device:**
- Browser [chrome]
- Version [84]

## Additional notes

![Screenshot 2020-06-26 at 15 00 56](https://user-images.githubusercontent.com/3613248/85905644-c7af9c00-b7c0-11ea-8338-db213ba868aa.png)

gnomeontherun:
Hello @pratikgala is this in relation to a VMware product?

pratikgala:
@gnomeontherun yes this is VMware product, the issue I have reported in on the clarity website component, but is the same for products using those components as well.

Jinnie:
I'll backport this to v4

PR #5195, opened 2020-10-23
fix(datagrid): add an aria-label to the column chooser dialog

amellnik:
Add an aria-label to the column chooser dialog.

Closes: #4754

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The column chooser dialog has the dialog role, but there's no label associated with it.  There is a invisible div inside the dialog that describes it, but since it's not an aria-label its not read again as the screen reader virtual focus exits the dialog.

## What is the new behavior?

Removes the invisible div and moves the message to the aria-label of the dialog.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column-toggle.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column-toggle.ts
@@ -38,14 +38,12 @@ import { ClrDatagridColumnToggleButton } from './datagrid-column-toggle-button';
     <div
       class="column-switch"
       role="dialog"
+      [attr.aria-label]="commonStrings.keys.showColumnsMenuDescription"
       [id]="popoverId"
       clrFocusTrap
       *clrPopoverContent="openState; at: smartPosition; outsideClickToClose: true; scrollToClose: true"
     >
       <div class="switch-header">
-        <div class="clr-sr-only" tabindex="-1" #menuDescription>
-          {{ commonStrings.keys.showColumnsMenuDescription }}
-        </div>
         <div class="clr-sr-only" tabindex="-1" #allSelected>{{ commonStrings.keys.allColumnsSelected }}</div>
         <ng-container *ngIf="!customToggleTitle">{{ commonStrings.keys.showColumns }}</ng-container>
         <ng-content select="clr-dg-column-toggle-title"></ng-content>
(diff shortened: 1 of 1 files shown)


packages/angular/projects/clr-angular/src/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -38,14 +38,12 @@ import { ClrDatagridColumnToggleButton } from './datagrid-column-toggle-button';
...
       role="dialog"
+      [attr.aria-label]="commonStrings.keys.showColumnsMenuDescription"
       [id]="popoverId"
       clrFocusTrap
       *clrPopoverContent="openState; at: smartPosition; outsideClickToClose: true; scrollToClose: true"
     >
       <div class="switch-header">
-        <div class="clr-sr-only" tabindex="-1" #menuDescription>
-          {{ commonStrings.keys.showColumnsMenuDescription }}
-        </div>
```
Jinnie, line 48: Where does this and related removals come from? I can't find them as part of the linked issue.

amellnik: Previously, the description for the column toggle panel was being done with this invisible div that was programatically focused when the column chooser was opened.  The problem with this is that while it's read when the panel is opened, when the user exits the panel the screen reader will tell them that they have exited a dialog but not what dialog it was that they exited.  This change moves the description to the dialog itself so that it's read both when the user enters it and leaves it.  

There's still another one of these invisible divs in the dialog after this change, but a better way of handling that would probably to use the Angular CDK's aria-live service rather than trying to get the same effect by shifting focus to an invisible div.

Jinnie: Sounds good. Btw, we're staying away from making our own aria-live implementations or services, because we don't want to impose any such implementations to the user. So, your solution is more in sync with our current strategy. I like it when code gets removed.

hippee-lee (approved):

amellnik:
If you found this helpful I'd appreciate it if you would add the hacktoberfest-accepted -- I'm attempting to mine a shirt. Thanks -A

gnomeontherun (approved):

Jinnie:
I've not forgotten this PR, I am trying to fix in parallel some other related issues with the same dialog and I want to see what interference might there be both ways.

Jinnie (approved):
OK, I don't think it will be conflicting with the other changes I'll be doing here.
Looks good to me, too.

PR #5200, opened 2020-10-23
feat(dropdown): overlay component (core)

mathisscott:
• adding cds-internal-overlay component
• added dynamic overlay test
• extended createId() helper to accept prefixes
• added unit tests for createId()
• override focus-trap-id in the overlay to add a prefix to the id
• need this so we can differentiate between focus traps and overlays
• updated content of overlay test demo
• added test stylings to overlay test demo
• locked down container height to prevent scrolling when overlay is present
• improved overlay test demo
• added backdrop to the overlay
• add custom properties for the backdrop
• worked with z-index and positioning to get layering working
• added aria as per fl specs
• added active property to FocusTrap
• allows us to tell if a FocusTrap is enabled or not
• added CSS to handle layered backdrops/overlays
• added code to the overlay to handle nested/layered overlays
• added a layered overlay example
• refactored focus traps to key off html root instead of body
• this addresses a jumping issue caused by applying a height and overflow: hidden to the body
• fixed focus trap unit tests broken by the refactor
• split up the focus trap no-scroller and utility no-scroller attribute value
• having them using the same styles was causing problems for the utility attr
• added closeChange event
• added functionality to fire closeChange when backdrop is clicked
• refactored overlay stories to use new closeChange event
• added Escape key functionality
• guarded against close events being fired by overlay that were not the currently active one
• unit tests for overlay
• clean-up todos
• added cds-first-focus functionality to move focus to an element inside a focus trap when it is enabled
• cleaned up overlay css
• changed isLayered to a getter
• cleaned up comments in overlay element
• refactor cds-modal to use the new overlay component
• refactor overlay to use a renderOverlay() method so shared HTML can be reused
• updated modal unit tests to account for meta close event output
• updated overlay and modal based on what I was seeing in react
• improved unit testing of modal sub components
• re-factored overlay backdrop rendering and interaction behaviors to account for framework synchronicity
• added dox and new overlay stories

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

A generic "overlay" component that can be shared by modals, wizards, dropdowns, signposts, etc.

- [ ] Yes
- [x] No


packages/core/src/internal/services/focus-trap-tracker.service.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('Focus Trap Tracker Service - ', () => {
     it('should return something', () => {
       const myDocroot: HTMLElement = service.getDocroot();
       expect(myDocroot).not.toBeNull('Docroot node should exist');
-      expect(myDocroot.tagName.toLowerCase()).toBe('body', 'should return the body node');
+      expect(myDocroot.tagName.toLowerCase()).toBe('html', 'should return the documentElement(html) node');
```
mathisscott, line 29: Placing this on body caused the underlying page to jump to the top.

packages/core/src/internal/services/focus-trap-tracker.service.ts
```diff
@@ -17,7 +17,7 @@ export const CDS_FOCUS_TRAP_DOCUMENT_ATTR = 'cds-focus-trap-ids';
 
 export class FocusTrapTracker {
   static getDocroot(): HTMLElement {
-    return document.body as HTMLElement;
+    return document.documentElement as HTMLElement;
```
mathisscott, line 20: `documentElement` returns the `html` tag

packages/core/src/internal/utils/identity.ts
```diff
@@ -58,8 +58,8 @@ export function getEnumValues(enumeration: any) {
   return Object.values(enumeration);
 }
 
-export function createId() {
-  return `_${Math.random().toString(36).substr(2, 9)}`;
+export function createId(prefix = '_') {
+  return `${prefix}${Math.random().toString(36).substr(2, 9)}`;
```
mathisscott, line 62: allowing for a prefix gives us the flexibility to use these identifiers to identify types of things if we want them to.

packages/core/src/styles/layout/_container.scss
```diff
@@ -2,8 +2,13 @@
...
 
-body[cds-focus-trap-ids*='_'],
-body[cds-layout='no-scrolling'] {
+html[cds-focus-trap-ids*='_'] {
+  overflow: hidden !important;
+  width: 100vw;
+  height: 100vh;
+}
+
+[cds-layout='no-scrolling'] {
```
mathisscott, line 5: tying these together brought along a lot of baggage for `no-scrolling`. splitting them apart allows `no-scrolling` to be used on containers.

packages/core/src/styles/mixins/_mixins.scss
```diff
@@ -22,19 +22,20 @@ $imported-modules: () !default;
...
-  }
-
-  @media (-webkit-min-device-pixel-ratio: 0) {
-    &:focus-within,
-    &.focus-within {
-      outline: $clr_baselineRem_5px auto -webkit-focus-ring-color;
-    }
-  }
-}
+// this mixin is not being used atm
```
mathisscott: this mixin is not being used in core anywhere. i wasn't sure if i could remove it or not.

packages/core/test-bundles/bundlesize.json
```diff
@@ -2,7 +2,7 @@
   "files": [
     {
       "path": "./dist/core/global.min.css",
-      "maxSize": "7 kB",
+      "maxSize": "7.2 kB",
```
mathisscott: global.min.css jumped up to 7kb. i increased it to 7.2.

packages/core/src/styles/layout/_container.scss
coryrylan: Some teams do not use the layout global setup but we still recommend the minimum reset so `cds-focus-trap-ids` might be better in the `module.reset.scss` file.

packages/core/src/styles/mixins/_mixins.scss
coryrylan: I think its safe to remove. We can always find it back in source control.

packages/core/src/internal-components/overlay/overlay.element.scss
```diff
@@ -0,0 +1,38 @@
+@import './../../styles/tokens/generated/index';
+@import './../../styles/mixins/mixins';
+
+:host {
+  // TODO: backdrop appearing needs transitions; what should they be?
+
+  --backdrop-opacity: 0.85;
+  --layered-backdrop-opacity: 0;
+  --backdrop-background: #{$cds-token-color-neutral-900};
```
coryrylan: The new theme system will help expose these with the object tokens 👍

jeeyun (approved):

hippee-lee (approved):

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -0,0 +1,167 @@
...
+ * </cds-internal-overlay>
+ * ```
+ *
+ * @beta
+ * @element cds-internal-overlay
+ * @slot - Content slot for the content inside the overlay's panel
+ * @event closeChange - Notify user when close event has occurred
+ * @cssprop --backdrop-background
+ * @cssprop --backdrop-opacity
+ * @cssprop --layered-backdrop-opacity
```
coryrylan: Once this is rebased with the new themes you may not need these anymore since they are part of the theme under the object tokens.

packages/core/src/internal/interfaces/index.ts
```diff
@@ -22,3 +22,117 @@ export type AriaLivePolitenessSettings = 'off' | 'polite' | 'assertive';
 export type Directions = 'up' | 'down' | 'left' | 'right';
 
 export type Orientations = 'horizontal' | 'vertical';
+
+// A11y
+
+export type AriaRoles =
```
coryrylan, line 28: 💯

packages/core/src/modal/modal.element.ts
```diff
@@ -44,8 +37,9 @@ import { styles } from './modal.element.css.js';
  * @slot cds-modal-header
  * @slot cds-modal-actions
  * @event closeChange - notify when the user has clicked the close button
- * @cssprop --backdrop-opacity
- * @cssprop --backdrop-background
+ * @cssprop --backdrop-opacity - inherited from the internal overlay component
+ * @cssprop --backdrop-background - inherited from the internal overlay component
+ * @cssprop --layered-backdrop-opacity - inherited from the internal overlay component
```
coryrylan: May not need these anymore once rebased onto themes with the object tokens http://localhost:6006/?path=/story/foundation-object-styles--page#object-layers

packages/core/src/modal/modal.element.ts
```diff
@@ -55,13 +49,11 @@ import { styles } from './modal.element.css.js';
  * @cssprop --content-box-shadow-color
  * @cssprop --width
  */
-export class CdsModal extends CdsBaseFocusTrap {
+export class CdsModal extends CdsInternalOverlay {
   static get styles() {
-    return [baseStyles, styles];
+    return [baseStyles, overlayStyles, styles];
```
coryrylan: You can inherit base component styles without needing to import with this syntax

```typescript
  static get styles() {
    return [...super.styles, styles];
  }
```

coryrylan (approved):

packages/core/src/modal/modal.stories.ts
```diff
@@ -23,118 +24,265 @@ export default {
...
-          ${args['cds-modal-actions']}
-        </cds-modal-actions>
+        <cds-modal-actions> ${args['cds-modal-actions']} </cds-modal-actions>
       </cds-modal>
     </cds-demo>
   `;
 }
 
-/** @website */
 export function defaultSize() {
```
coryrylan: Would the interactive demos be better as new additional stories? They are good and helpful for focus trap testing. I think the main demos should remain stateless since it helps show the component differences and the code snippets for the docs. The code is marked for the website docs so we likely would want them as plain html when possible. Will the `__demo-mode` continue to work for the stateless demos? It will be needed for story screenshot testing and maybe the website as well?

packages/core/src/modal/modal.stories.ts
```diff
@@ -23,118 +24,265 @@ export default {
...
+      <cds-modal-actions>
+        <cds-button @click=${hideXLargeModal} action="outline">Cancel</cds-button>
+        <cds-button @click=${hideXLargeModal}>Ok</cds-button>
+      </cds-modal-actions>
+    </cds-modal>
   `;
 }
 
 /** @website */
 export function focusTrap() {
```
coryrylan: Looking at the interactive demos this one seems to have a layout issue for the demo https://5fda9d2e0ea1383e0d1a73c2--vmware-clarity.netlify.app/storybook/core/?path=/story/components-modal--page#managing-focus-in-a-modal

packages/core/test-bundles/index.js
```diff
@@ -1,11 +1,11 @@
 import '@cds/core/badge/register.js';
 import '@cds/core/icon/register.js';
 import { ClarityIcons, userIcon } from '@cds/core/icon';
-import { CommonStringsService } from '@cds/core';
+import { I18nService } from '@cds/core/internal';
```
coryrylan: This import is a public api so it wouldn't need to be imported from internal. The deep import might be what caused the base bundle size to go up

`import { I18nService } from '@cds/core';`

PR #5283, opened 2020-11-10
fix: updated a11y of forms and form input groups

mathisscott:
• updated cds-control-actions so they could be announced by screen readers
• duplicated screen readers styles because reset styles aren't being loaded internally to components
• updated placeholder text values to "place holder text"
• the rationale here was that VoiceOver was pronouncing it "plass-holder text"
• VoiceOver now pronounces the value as intended
• added a reset style for a fieldset that uses cds-layout
• legends inside a reset fieldset are likewise reset
• updated checkout form demo to show expected use of fieldset
• fieldsets + legends behave slightly odd inside a layout; I got it working but :-\
• new reset styles increased reset bundlesize
• removed unused imports from icon.stories.ts

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/forms/control-action/control-action.element.ts
+++ b/packages/core/src/forms/control-action/control-action.element.ts
@@ -6,7 +6,14 @@
 
 import { html, property } from 'lit-element';
 import { CdsIcon } from '@clr/core/icon/icon.element.js';
-import { baseStyles, CdsBaseButton, querySlot, setAttributes, assignSlotNames } from '@clr/core/internal';
+import {
+  assignSlotNames,
+  baseStyles,
+  CdsBaseButton,
+  hasAttributeAndIsNotEmpty,
+  querySlot,
+  setOrRemoveAttribute,
+} from '@clr/core/internal';
 import { styles } from './control-action.element.css.js';
 import { LogService } from '@clr/core/internal';
 
@@ -30,6 +37,10 @@ export class CdsControlAction extends CdsBaseButton {
   /** Set the action type placement within the supporting input control */
   @property({ type: String }) action: 'label' | 'prefix' | 'suffix';
 
+  @property({ type: Boolean }) readonly = false;
+
+  @property({ type: String }) ariaLabel = '';
+
   @querySlot('cds-icon') protected icon: CdsIcon;
 
   static get styles() {
@@ -47,7 +58,16 @@ export class CdsControlAction extends CdsBaseButton {
 
   connectedCallback() {
     super.connectedCallback();
-    setAttributes(this, ['aria-hidden', 'true']);
+    this.syncAria();
+  }
+
+  // TODO: TESTME
+  private syncAria() {
+    const iAmReadonly = this.readonly;
+    const iHaveAriaLabel = hasAttributeAndIsNotEmpty(this, 'aria-label');
+    setOrRemoveAttribute(this, ['aria-hidden', 'true'], () => {
+      return iAmReadonly && !iHaveAriaLabel;
+    });
   }
 
   updated(props: Map<string, any>) {
@@ -56,8 +76,9 @@ export class CdsControlAction extends CdsBaseButton {
       this.setSlotLocation();
     }
 
-    if (props.has('readonly')) {
+    if (props.has('readonly') || props.has('ariaLabel')) {
       this.validateAriaLabel();
+      this.syncAria();
     }
   }
 
--- a/packages/core/src/forms/control-action/control-action.element.ts
+++ b/packages/core/src/forms/control-action/control-action.element.ts
@@ -61,7 +61,6 @@ export class CdsControlAction extends CdsBaseButton {
     this.syncAria();
   }
 
-  // TODO: TESTME
   private syncAria() {
     const iAmReadonly = this.readonly;
     const iHaveAriaLabel = hasAttributeAndIsNotEmpty(this, 'aria-label');
--- a/packages/core/src/forms/forms.stories.ts
+++ b/packages/core/src/forms/forms.stories.ts
@@ -37,7 +37,7 @@ export const nativeHTML5ValidationSingle = () => {
   return html`
     <cds-input validate>
       <label>text input (required)</label>
-      <input placeholder="placeholder text" required />
+      <input placeholder="place holder text" required />
       <cds-control-message error="valueMissing">required</cds-control-message>
     </cds-input>
   `;
@@ -99,7 +99,7 @@ export const form = () => {
       <cds-form-group control-width="shrink">
         <cds-input>
           <label>text label</label>
-          <input placeholder="placeholder text" />
+          <input placeholder="place holder text" />
         </cds-input>
 
         <cds-select>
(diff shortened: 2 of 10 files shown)


packages/core/src/forms/control-action/control-action.element.ts
```diff
@@ -47,7 +58,16 @@ export class CdsControlAction extends CdsBaseButton {
 
   connectedCallback() {
     super.connectedCallback();
-    setAttributes(this, ['aria-hidden', 'true']);
+    this.syncAria();
+  }
+
+  // TODO: TESTME
```
mathisscott: Gah! I need some tests.

packages/core/src/forms/control-action/control-action.element.ts
```diff
@@ -56,8 +76,9 @@ export class CdsControlAction extends CdsBaseButton {
       this.setSlotLocation();
     }
 
-    if (props.has('readonly')) {
+    if (props.has('readonly') || props.has('ariaLabel')) {
```
mathisscott, line 78: Allows for readonly control actions to still be read. Also allows for readonly control actions to be skipped when reading them would be redundant.

packages/core/src/forms/forms.stories.ts
```diff
@@ -37,7 +37,7 @@ export const nativeHTML5ValidationSingle = () => {
   return html`
     <cds-input validate>
       <label>text input (required)</label>
-      <input placeholder="placeholder text" required />
+      <input placeholder="place holder text" required />
```
mathisscott, line 40: Screen readers now read this as `PLACE holder`.

Previously, it was being read as `PLASS holder`

packages/core/src/forms/forms.stories.ts
```diff
@@ -1723,125 +1733,130 @@ export const multiColumn = () => {
...
-
-        <cds-select cds-layout="col@xs:6 col@sm:4">
-          <label>State</label>
-          <select>
-            <option>choose...</option>
-            <option>California</option>
-          </select>
-        </cds-select>
+      <div cds-layout="vertical gap:xl">
+        <fieldset cds-layout="vertical gap:md">
```
mathisscott: Fieldsets and legends don't fully behave in layouts. I'm not 100% sure why. But you can see I took an alternate strategy of defining my smallest vertical whitespace as the "gap" and adding whitespace with our padding tokens.

packages/core/src/icon/icon.stories.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import '@clr/core/icon/register.js';
-import { CdsIcon, ClarityIcons, imageIcon, userIcon } from '@clr/core/icon';
```
mathisscott, line 8: Build was complaining about `CdsIcon` and `registerElementSafely` in this file.

packages/core/src/internal/utils/dom.ts
```diff
@@ -32,6 +32,17 @@ export function hasAttributeAndIsNotEmpty(element: HTMLElement | null, attribute
   return !!element && element.hasAttribute(attribute) && isStringAndNotNilOrEmpty(element.getAttribute(attribute));
 }
 
+// TODO: TESTME!
```
mathisscott: Gah! I need some tests.

packages/core/src/internal/utils/dom.ts
```diff
@@ -32,6 +32,17 @@ export function hasAttributeAndIsNotEmpty(element: HTMLElement | null, attribute
...
+export function setOrRemoveAttribute(element: HTMLElement, attrTuple: HTMLAttributeTuple, test: () => boolean) {
+  const [attribute, value] = attrTuple;
+  console.log('ohai: element = ', element, '; tuple = ', attribute, ', ', value, '; did i pass my test? ', test());
```
mathisscott: What rolls down stairs
alone or in pairs,
and over your neighbor's dog?
What's great for a snack,
And fits on your back?
It's log, log, log

It's log, it's log,
It's big, it's heavy, it's wood.
It's log, it's log, it's better than bad, it's good."

Everyone wants a log
You're gonna love it, log
Come on and get your log
Everyone needs a log
log log log

packages/core/src/styles/layout/_display.scss
```diff
@@ -55,3 +55,21 @@
 @media (min-width: $cds-token-layout-width-xl-static) {
   @include display('xl');
 }
+
+// duplicating here b/c the form fields can't pull it in from
```
mathisscott, line 59: Comment reveals all

packages/core/src/forms/forms.stories.ts
coryrylan: I ran into this when working on forms. Fieldset and legend dont follow standard CSS layout and will break anything that uses flexbox or css grid and also has several cross browser behavior issues. Overall its not used often because it breaks most CSS. The issue applies to almost all browsers currently. I think we should hold off on this change until we find a robust solution otherwise we will have lots of weird/broken layouts in general in clarity regardless of `cds-layout` usage.

Related links:
https://github.com/jensimmons/cssremedy/issues/50 
https://stackoverflow.com/questions/28078681/why-cant-fieldset-be-flex-containers
https://github.com/philipwalton/flexbugs#flexbug-9 
https://bugs.chromium.org/p/chromium/issues/detail?id=375693
https://bugzilla.mozilla.org/show_bug.cgi?id=984869 

Some options, we rely on role group and aria-labeledby which is what some of the links above suggest. Another option is we create a some kind of wrapper component like `<cds-fieldset>`. Or possibly maybe the most robust solution is to have the current `cds-form-group` create an inner `<fieldset>` if a `<legend>` element is slotted. Then its encapsulated and wont break regular css layouts. Form group is a connivence component to make it easy to layout a group of controls. There can be more than one in a form so it does already closely relate to a fieldset.

```html
<cds-form-group>
  <legend>...</legend>
  <cds-input>...</cds-input>
  <cds-input>...</cds-input>
</cds-form-group>
```

coryrylan (changes requested):

packages/core/src/forms/forms.stories.ts
mathisscott: Due to the nature and time constraints of this effort, I am inclined to use groups and aria-labelledby. Creating another wrapper component or an inner encapsulation for field sets would take a lot longer. And I honestly don't know if that would fix the issues I've seen.

coryrylan: I'd vote to use groups and aria-labeledby as well for now. That way we dont push something out that would need to be rolled back or cause unexpected behavior like legend/fieldset do.

Something like this I would think would be a reasonable solution until we better figure out how to handle fieldset

```html
    <div cds-layout="vertical gap:lg container:sm">
      <cds-form-group role="group" aria-labelledby="billing-address-group" layout="vertical">
        <p id="billing-address-group" cds-text="section">Billing Address</p>
        ...
      </cds-form-group>

      <cds-form-group role="group" aria-labelledby="payment-group" layout="vertical">
        <p id="payment-group" cds-text="section">Payment</p>
        ...
      </cds-form-group>
    </div>
```

bdryanovski (approved):

hippee-lee (approved):

coryrylan (approved):

PR #5285, opened 2020-11-11
build: change package scope clr -> cds

hippee-lee:
Signed-off-by: Matt Hippely <mhippely@vmware.com>

## PR Type

What kind of change does this PR introduce?

Updates the build to publish `@cds/core, @cds/react`. This includes updating all core apps, the new website and Angular dev app. It also updates `@clr/city` dependencies to use `@cds/city`. 

Verification process includes:
1. Serving and building all core apps
2. serving and building the website app
3. serving and building the Angular dev app
4. publishing `@cds/core`, `@cds/react' and `@cds/city` to a dockerized Verdaccio proxy for npm
5. Creating an app with create-react-app that pulls packages from the local Verdaccio packages and setting up a simple app with CdsModal

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [x] CI related changes
- [x] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?
The build publishes all artifacts to the `@clr` scope and all apps with dependencies consume from the `@clr` scope. 

Issue Number: N/A

## What is the new behavior?
The build publishes only `core and react` artifacts to the `@cds` scope and all apps with dependencies consume those packages from the `@cds` scope. One additional change is to also consume `@cds/city` even though that is published from a different repo. 

## Does this PR introduce a breaking change?

- [x] Yes 
- [ ] No

Consumers will need to install core, react wrappers and clarity city font from the new scope:

```
yarn add @cds/core @cds/react @cds/city
```

The core apps depend on [this PR](https://github.com/vmware/clarity-city/pull/6).

**The build is failing b/c of yarn.lock. I'll need to update that after `@cds/city` is published on npm.**

PR #5288, opened 2020-11-11
build: core ng modules

Shijir:

This pull request generates angular modules for each Core component so that users don't need to rely on `CUSTOM_ELEMENTS_SCHEMA`. Every time Angular builds, it generates those core angular modules first. I have added sample core components to the dev app for the purpose of visually testing. I've also used the core angular module(`CdsModule`) in the `core-angular-cli` app. Finally, it also generates a test file `core-ng-module.spec.ts` in `core-angular-cli` app. Its purpose is to test every core component directive from the angular build so it basically mimics an integration test.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/angular/projects/clr-angular/src/index.ts
```diff
@@ -21,3 +21,5 @@ export * from './accordion/stepper/index';
 export * from './progress/spinner/index';
 export * from './progress/progress-bars/index';
 export * from './timeline/index';
+
+export * from './cds/components/index';
```
gnomeontherun: We don't want to ship this as part of `clr-angular`, it should output to a new directory in the dist directory.

packages/angular/projects/dev/src/app/cds/cds.demo.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { Component } from '@angular/core';
+import { FormGroup, FormBuilder, Validators } from '@angular/forms';
+import { startWith } from 'rxjs/operators';
+import { Observable } from 'rxjs';
+
+import { cloudIcon, ClarityIcons } from '@clr/core/icon';
```
gnomeontherun: Any Core registrations should happen inside of the generated CdsModule so that apps don't have to manually consider it.

scripts/core-ng-module-generator.js
```diff
@@ -0,0 +1,172 @@
...
+const INTEGRATION_TEST_FILE_NAME = 'core-ng-module.spec.ts';
+
+const directiveTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/directive.ts.mustache', 'utf8');
+const moduleTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/module.ts.mustache', 'utf8');
+const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.mustache', 'utf8');
+const rootIndexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/root-index.ts.mustache', 'utf8');
+const rootModuleTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/root-module.ts.mustache', 'utf8');
+const testTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/test.ts.mustache', 'utf8');
+
+function getDirectiveClassName(tagName, prefix, suffix) {
```
gnomeontherun, line 39: Some documentation of what happens in this file would go a long way to support future development.

yarn.lock
```diff
@@ -3299,35 +3299,19 @@
...
 
+"@clr/angular@./packages/angular/dist/clr-angular":
+  version "0.0.0"
+
 "@clr/city@1.1.0", "@clr/city@^1.1.0":
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/@clr/city/-/city-1.1.0.tgz#4a444cd12c626d66ffadab27e9d09bb3c8eca352"
   integrity sha512-R+C4uywmXoTD01LINOt3O0cBRviQdbAVNxdVvOyuO3+rM9bvFroF7UZY0R1ue/xvKXlqJrEkNKZQODeKjzaAhA==
 
 "@clr/core@./packages/core/dist/core":
```
gnomeontherun: Is this right?

yarn.lock
```diff
@@ -8585,16 +8569,6 @@
     "@webassemblyjs/wast-parser" "1.9.0"
     "@xtuc/long" "4.2.2"
 
-"@webcomponents/custom-elements@^1.4.2":
```
gnomeontherun: Are these due to changes for IE that we didn't get updated correctly in the lock file?

gnomeontherun (changes requested):

scripts/core-ng-module-generator.js
```diff
@@ -0,0 +1,172 @@
...
+    tagName
+      .split('-')
+      .map(w => w.charAt(0).toUpperCase() + w.slice(1))
+      .join('') +
+    suffix
+  );
+}
+
+function getDiretoryName(tagPath) {
+  return tagPath.split('/')[1];
```
bdryanovski: Same as 
```js
functioon getDirectoryName(path) {
  return path.dirname(path);
}
```

scripts/core-ng-module-generator.js
```diff
@@ -0,0 +1,172 @@
...
+
+    createFileOnTemplate(rootModuleTemplate, rootModuleTemplateData, 'cds.module.ts', TARGET_PATH);
+
+    return moduleData;
+  }
+}
+
+function createDirectives(componentDirectories) {
+  if (componentDirectories) {
+    const customElementsData = JSON.parse(fs.readFileSync(SOURCE_PATH));
```
bdryanovski, line 113: `JSON.parse` has the tendency to throw SyntaxError when something gets even close to wrong. Is it better to have `try/catch`

scripts/core-ng-module-generator.js
```diff
@@ -0,0 +1,172 @@
...
+      }
+
+      return moduleData;
+    }, {});
+  }
+}
+
+function createComponentDirectories() {
+  if (fs.existsSync(SOURCE_PATH)) {
+    const customElementsData = JSON.parse(fs.readFileSync(SOURCE_PATH));
```
bdryanovski, line 142: same here `JSON.parse` could throw `SyntaxError` and kill the `core-ng-module-generator` task

bdryanovski (approved):
Some minor comments, nothing to block the PR. 

About the `yarn.lock` I think running it against `next` could fix it. There were a lot of changes lately into the package and three versions so it could generate strange lock files.

scripts/core-ng-module-generator.js
coryrylan: Wouldnt we want that to throw and kill the task so we know something is wrong? Or catch and throw a specific exception?

packages/angular/projects/clr-angular/src/index.ts
coryrylan: Maybe something like `@cds/angular`  since `@clr/angular` will remain under its current scope?

gnomeontherun: Right, that is the intention and https://github.com/vmware/clarity/pull/5285 is the change to follow.

gnomeontherun:
@Shijir whats the status here?

Shijir:
@gnomeontherun this is completed except the build issue the's being addressed in https://github.com/vmware/clarity/pull/5310

.gitignore
```diff
@@ -18,6 +21,7 @@ src/clr-angular/.ng_build
 src/clr-angular/.ng_pkg_build
 
 /reports
+packages/eslint-plugin-clarity-migration/reports
```
gnomeontherun: This is an old artifact, shouldn't be needed anymore.

packages/angular/projects/dev/src/app/cds/cds.demo.module.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { CoreModule } from '@cds/angular';
+import { ClarityModule } from '@clr/angular';
```
gnomeontherun: Do we need both modules?

scripts/core-ng-module-generator.js
gnomeontherun: I think this is alright, if the file is missing you'll get an exception and that works fine for this situation since its a script.

yarn.lock
```diff
@@ -46,6 +46,14 @@
     "@angular-devkit/core" "11.0.1"
     rxjs "6.6.3"
 
+"@angular-devkit/architect@^0.1001.0":
```
gnomeontherun: Seems like a lot of changes here, can we update only the necessary things?

packages/angular/projects/cds-angular/src/cds/_stubs/root-module.ts.mustache
```diff
@@ -0,0 +1,24 @@
...
+{{/modules}}
+
+@NgModule({
+  exports: [
+    {{#modules}}
+    {{moduleClassName}},
+    {{/modules}}
+  ],
+})
+export class CoreModule {}
```
gnomeontherun: Let's make this `CdsModule`

packages/angular/projects/cds-angular/src/cds/_stubs/module.ts.mustache
```diff
@@ -0,0 +1,29 @@
...
+    {{directiveClassName}},
+    {{/directives}}
+  ],
+  exports: [
+    {{#directives}}
+    {{directiveClassName}},
+    {{/directives}}
+  ],
+})
+export class {{moduleClassName}} {}
```
coryrylan, line 29: It looks like in the code this will generate one NgModule per component correct? I want to double check since Angular can only tree shake down to the NgModule level for us to keep the per component tree shaking apps should import each component NgModule instead of the root when possible.

```typescript
// app module
@NgModule({
  imports: [
    CommonModule,
    CdsIconModule,
    CdsAlertModule,
    CdsModalModule
  ],
```

Shijir: Yes, each core component will have its own module. And then there will be a single root module `CdsModule` that includes all of those component modules.

packages/angular/projects/cds-angular/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "@cds/angular",
+  "version": "@VERSION",
+  "description": "Core component modules for Clarity Angular",
+  "keywords": ["ng-add", "clarity", "angular", "components", "core"],
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:vmware/clarity.git"
+  },
+  "peerDependencies": {
+    "@angular/common": "^11.0.0",
```
gnomeontherun: We should soften this version if possible, I don't know how far back we could go? It might be possible to support back to v4?

yarn.lock
Shijir: Fixed.

packages/angular/projects/cds-angular/src/cds/_stubs/root-module.ts.mustache
Shijir: Fixed

packages/angular/projects/cds-angular/README.md
```diff
@@ -0,0 +1,13 @@
+## Installing Clarity Angular [![npm version](https://badge.fury.io/js/%40clr%2Fangular.svg)](https://badge.fury.io/js/%40clr%2Fangular)
```
gnomeontherun: Badge shows @clr/angular instead of @cds/angular.

packages/angular/projects/cds-angular/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@cds/angular",
+  "version": "@VERSION",
+  "description": "Core component modules for Clarity Angular",
+  "keywords": ["ng-add", "clarity", "angular", "components", "core"],
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:vmware/clarity.git"
```
gnomeontherun: "url": "https://github.com/vmware/clarity.git"

gnomeontherun (approved):
Just a few comments.

Issue #4502, opened 2020-04-22
core: common strings


coryrylan:
# Core: i18n & a11y internal strings

Prototype: https://stackblitz.com/edit/typescript-kbwbop 

Many but not all components need to have a specific label for a11y. Example, a modal close button should give more context to a screen reader user than just "close modal". Instead it should specify its intent. Example, "Close VM setup modal". Other components such as the date-picker will internally need strings that can be switched out for i18n but may not be specific to the context usage.

I18n also has similar needs in that these internal strings need to be set and dynamically changed. This API proposal unifies both into a single API that allow customization of internal strings at a global and component instance level.

## API

Each component expose a `i18n` property that can will contain an object of all i18n strings
for that component. This object would be initialized with default values from the global `CommonStringsService`. This would allow all i18n values to be overridden at a global level or component level and not require individual attrs/props to be added or removed from the component public API.

```typescript
export class CdsDatepicker extends LitHtml {
  @i18n() i18n = CommonStringsService.keys.datepicker; 
}
```

```html
<cds-datepicker cds-i18n="{ nextMonth: 'next available month', toggle: 'toggle schedule calendar' }">
  <label>Today</label>
  <input tyle="date" />
</cds-datepicker>
```

This API allows i18n strings to be customized at a component level and global level. They can be customized for i18n or a11y specific reasons with no change in the public API.

## i18n decorator

The `@i18n` decorator uses the `@property` decorator internally but provides additional behavior. 

1. Use component default values and merge any custom overridden value via the component or `ComonStringsService`.
2. Provide a way to mark certain strings as required per instance for a11y.  

The lit elements property decorator which i18n will use automatically supports simple flat objects in attributes or JS properties. https://lit-element.polymer-project.org/guide/properties#initialize-property-values-from-attributes-in-markup 

```javascript
const datepicker = document.querySelector('cds-datepicker');
datepicker.i18n = myCustomI18nValues;
```

The flat object styles make the API stable and easy to customize but also easy to support i18n with server side rendering systems to serialize to the component

```cshtml
// C# razor view
<cds-datepicker @Html.Attr("cds-i18n", model.i18n)></cds-datepicker>
```

## Validation

Some i18n strings could be required to be customized per the usage for a11y. The i18n decorator could take a list of required strings to warn the developer to set per component. This may need to be a future feature to add on depending on a11y requirements.

```typescript
export class CdsModal extends LitHtml {
[...]

PR #5295, opened 2020-11-12
feat: core i18n

jeeyun:
The I18nService is a static class that gives users the ability to use and override strings within the components for internationalization / globalization. One can override default values globally for their application or override per component instance as needed. See https://github.com/vmware/clarity/issues/4502 for more details.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/vmware/clarity/issues/4502

- [ ] Yes
- [x] No


packages/core/src/internal/services/i18n.service.ts
```diff
@@ -0,0 +1,223 @@
...
+    nextMonth: 'Next month',
+    previousDecade: 'Previous decade',
+    nextDecade: 'Next decade',
+    currentDecade: 'Current decade',
+    selectMonthText: 'Select month, the current month is {CALENDAR_MONTH}',
+    selectYearText: 'Select year, the current year is {CALENDAR_YEAR}',
+    daypickerSRCurrentMonthPhrase: 'The current month is {CURRENT_MONTH}',
+    daypickerSRCurrentYearPhrase: 'The current year is {CURRENT_YEAR}',
+    daypickerSRCurrentDecadePhrase: 'The current decade is {DECADE_RANGE}',
+  },
```
jeeyun: Should I remove all the strings that were carried over from the Angular library and re-add them as needed we implement the components? In particular, the datepicker has these string values where a part of the value is meant to be dynamic. I don't believe we'll need this behavior with this I18nService, so at the very least I expect these values to change. Some of these keys might go away too depending on how we design / implement future components.

coryrylan: That sounds good. Re-adding them as needed will give us time to review/audit them as we go so we don't include anything that isn't needed.

packages/core/src/alert/alert-group.stories.ts
```diff
@@ -4,14 +4,18 @@
...
 import { action } from '@storybook/addon-actions';
 import { html } from 'lit-html';
 import { ClarityIcons, headphonesIcon, nodeGroupIcon, timesCircleIcon } from '@clr/core/icon';
 import customElements from '../../dist/core/custom-elements.json';
 
 ClarityIcons.addIcons(headphonesIcon, nodeGroupIcon, timesCircleIcon);
 
+I18nService.localize({
+  alert: { closeButtonAriaLabel: 'close this alert' },
+});
```
coryrylan, line 18: Small nit, we should probably import the `I18nService` from `@clr/core` here just for helping set an example for anyone looking at the example code in storybook

packages/core/src/alert/alert-group.stories.ts
```diff
@@ -43,8 +47,12 @@ export const alertGroup = () => {
   return html`
     <div cds-layout="vertical gap:sm">
       <cds-alert-group status="info" aria-label="This is an example info alert group">
-        <cds-alert closable>
-          This example is an alert a user may be able to close inside an alert group with a status of "info".
+        <cds-alert
+          closable
+          aria-label="This is an example info alert group"
+          cds-i18n='{ "closeButtonAriaLabel": "close my example alert"}'
```
mathisscott, line 54: fwiw, an example that also showed how to change the default string of the alert icon would make accessibility very happy.

mathisscott (approved):

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  alert: {
+    closeButtonAriaLabel: string;
+    loading: string;
+    info: string;
+    success: string;
+    warning: string;
+    danger: string;
+  };
+  datagrid: {
```
coryrylan: Are we still planning on  removing these and only adding them as the component is added?

jeeyun: Yup, I will remove all the extra ones that aren't currently used in core.

coryrylan (approved):

PR #5300, opened 2020-11-16
chore: improve build times on local and CI

bdryanovski:
### Improve Build runtime at local and Github Action CI

Dropping the `yarn install` on packages that are not needed for the build/website workflow

- Fewer packages to install - less runtime
- Not installing packages that we don't use like for example skipping `rollup` is a huge boost.
- Small cache size for Github CI Actions will let us have a faster pack/unpack of the cache before the run.
- Did not find any side-effects of it. Internal scripts will continue to run as expected.
- Less chance of having to update the `yarn.lock` file (having less packages to track will mean fewer packages that could require updating)

When we have `yarn.lock` change we have to drop the cache from Github CI Actions and rebuild it from scratch,  this is adding additional run-time (deleting cache, installing packages, zipping packages into cache). 

⚠️ In the case when we have two PR one with `yarn.lock` file changes and one without. Will force the rebuilding of the cache for every run.  

⚠️ PR doesn't make difference between v4, v5, next, and so on, so the cache is also rebuilt when we have PR for v4 and next they have different `yarn.lock` file but have the same cache key. This could be fixed by adding `version` as part of the cache key. 

⚠️ There is [Github Cache Limit](https://github.com/actions/cache#cache-limits) around (5000mb) - if we keep big cache files Github automatically will delete older cache files.

Cache size before the changes: **~2202 MB (2309315668 B)**
Cache size after the changes:  **~1182 MB (1239547915 B)**

By adding `github.base_ref` as part of the key - there will be cache file for every branch that is a target for our case `v4` & `next` that won't change that much so we will have much more chance to not have to rebuild the cache when multiple PR is running.  

| Workflows| Old  time |New time |Time gain|
|-----------|:---------:|:---------:|:---------:|
| Website    |  21m 20s~  | 10m 37s~   | 11m~ | 
| Dev           |    23m 13~  |   7m 17s~   | 16m~ |
| Build         | 18m 22s~   |   9m 34s~  | 9m~   |

❗️ time depends on a lot of things like networks times to NPM or how much load is right now on Github Actions servers so they could go up/down
❗️ task work in parallel so the slowest task is the one overall time.

### Yarn Zero-Install

This is also an option, not sure what will be the speed benefit of it - there are few small issues when trying to run it.

- Yarn Zero-Install is part of the Yarn 2.x right now the latest version was `2.0.0-rc0`
- Yarn is throwing some errors when trying to run it  - from what I saw (did not spend much time on it) some of the direct paths to packages were the issues, also notice that it does not execute the `preinstall` hook for some reason?
- Yarn 2 is required to be installed - did not find an easy way to have both yarn 1.x and yarn 2.x at the same time without having issues. (will require to update v4 as well)

The basic idea of `zero-install` is to add `.yarn/cache` folder as part of the repository. This way yarn will have to load and run from the cache and won't install anything. But I'm afraid that this will make the `git checkout` a slow process. Over time the `cache` folder will change and all the changes will be tracked by `git`  - this will make the deltas and checkout process really slow. As well the git-repository size will become huge. Other issues could be packages that have `postinstall` scripts, on paper Yarn will keep them as separate packages and run the `postinstall` when loading the cache, but there are disclaimers in the documentation to keep an eye on it.

⚠️ Not sure how we gonna handle the git conflicts related to the `.yarn/cache` folder?

```
// .gitignore

.yarn/*
!.yarn/cache
!.yarn/releases
!.yarn/plugins
!.yarn/sdks
!.yarn/versions
```

So for that to work, we gonna need to invest some time that this PR won't cover. Just adding my finding with the idea that this could be helpful in feature research.

I tested most of the command tasks, run core/angular, and website build and did not find anything that is not working - there could be something ... and may say that everything is working! 

⚠️⚠️⚠️ The side effect of these changes is that the demo projects inside `apps/` will have to run `yarn install` on their own - no longer the packages in their `package.json` file will be pick by running `yarn` into the root folder. Did not find any scripts to depend on them but I'm willing to patch them if someone knows about potential issues with them. 

<hr/>

### ⚠️ Update ⚠️

So the overview of the update is as follow: 

- Packages inside apps and tools are no longer auto-install if not needed (exception is the website as it is a part of the Build)
- Demo apps require to manually run `yarn install` inside the directory if needed to test something - but no longer are auto-installed on the CI as part of the build tasks.
- Some missed dependencies are installed to their packages so every package must be built outside the monorepo without getting errors for missing dependencies.
- Clear list of what package is used where and what versions will be installed. Let us manage dependencies and backtrace build errors.
- Require half of the space before so Github Actions Cache could store more than 2 Builds at a time.
- Double the speed (or close to double) of installing Yarn packages and running Github Actions.
- Uncompressing and compressing Yarn Cache now take less time and space.
- Small steps to solving the issue with copy-paste dist-packages around the project so we could have internal package dependency

### Overall  stats

Now the packages and `node_modules` folder are the ones that are the minimum requirement to run the Github CI task and start developing.

```bash
# Before changes (25m 6s ±) next branch
180K	./tools/audit-website/node_modules
180K	./tools/link-checker/node_modules
 56K	./tools/visual-tester/node_modules
2.1G	./node_modules
3.9M	./packages/ui/node_modules
742M	./packages/core/node_modules
1.2G	./packages/angular/node_modules
198M	./packages/eslint-plugin-clarity-adoption/node_modules
 69M	./packages/icons/node_modules
 75M	./packages/schematics/node_modules
 67M	./packages/react/node_modules
1.3G	./apps/core-angular-universal/node_modules
 60M	./apps/core-snowpack/node_modules
 11M	./apps/core-angular-js/node_modules
190M	./apps/core-rollup-js/node_modules
 16M	./apps/website/node_modules
 92M	./apps/core-create-react-app/node_modules
1.1G	./apps/core-angular-cli/node_modules
9.4M	./apps/core-parcel-js/node_modules
9.4M	./apps/core-vue-cli/node_modules
7.2G	total

Cache Size: ~2468 MB (2587418242 B)

# After changes (7m 24s ±) build-changes branch
1.0G	./node_modules
 13M	./packages/ui/node_modules
763M	./packages/core/node_modules
1.2G	./packages/angular/node_modules
143M	./packages/eslint-plugin-clarity-adoption/node_modules
 68M	./packages/icons/node_modules
 75M	./packages/schematics/node_modules
 66M	./packages/react/node_modules
474M	./apps/website/node_modules
3.8G	total

Cache Size: ~1477 MB (1548926817 B)
```

As shown above, the size of the package that we install now is `3.8G`, and then compress it is around `1.3G` as the cached file. This lets us have more than 2 cache files inside Github Actions, and thus minimizes the need of rebuilding caches. 

The issue with the cache files that we create is that are too close to the internal limit provided by Github for the size limit. More here: [https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#:~:text=There is no limit on, is less than 5 GB](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#:~:text=There%20is%20no%20limit%20on,is%20less%20than%205%20GB)

So when we have two PR and a backport for example this will always drop cache files that we have. So it's making caching not helping a lot - even making times worse - an additional task is run every time when we update the cache (tgziping node_modules). Now with more compact `node_modules` we could have more cached stores before the need of dropping some of them is needed. 

Another issue that is made steps for resolving is unnamed dependencies: This is an issue cause we are not sure what package is used when running the code. The reason for that is yarn is trying to move packages into the root of the project to reuse them in places where equal versions are needed. More on how Yarn workspaces are working could be found here [https://www.notion.so/Improve-Build-c848ce3422ee46d2b88411db8c1d910c#8f1499c3dae54213b73c3a0f4bac9a40](https://www.notion.so/Improve-Build-c848ce3422ee46d2b88411db8c1d910c#8f1499c3dae54213b73c3a0f4bac9a40)

A quick explanation is that the workspace package or any package inside the monorepo could request and use any package found inside the monorepo even if not mention inside his own `package.json`. This is a nice option to have and making build "work" is introducing an issue with what version of the package you will get when asking for it. Also, where will be this package located - will it be in the `node_modules` inside the sub-package or into the root-level? The reality is that it depends on other packages and their version and how often you use one version or do you have multiple versions and so on. 

So for example: `cds/core` use inside `babel-loader` but it does not mention it inside the `package.json` but because is part of the monorepo it will get some version that Yarn already installed for some other package. You need at minimum two packages with the same version to be able to share between packages. From this there could be two bad outcomes:

- One of these two packages change the version of `babel-loader` in that case Yarn will install two versions of `babel-loader` inside their local `node_module` and from now on `babel-loader` will no longer be accessible for `cds/core` and it will fail.
- Both packages change their version number - updating/downgrading - in that case, `cds/core` will get whatever version is set - This could create a case when update at one place - could break things into another package without even having any relations. This could lead to cryptic errors or unstable builds.

This change must not affect the day-to-day workflow that you have at the moment - but only provide a little more stable dependency tree and better resource use.

❗️ To test these changes here are the steps needed for you to run locally.

```bash
# Test website build
yarn clean:modules;
yarn; 
yarn build:libs; 
yarn build:website; 
yarn build:storybook:core

# Test Build
yarn clean:modules;
yarn;
yarn build:ci

# Test Dev build
yarn clean:modules;
yarn;
yarn build:libs;
yarn build:angular:dev;

# Testing everything at once:
yarn clean:modules; yarn; yarn build:libs; yarn build:ci; yarn build:website; yarn build:storybook:core;
```

As a result of this you must get: 

- ✅ All tasks must pass without errors.
- ✅ YOU MUST NOT GET ANY CHANGES for `yarn.lock` - if you get that means that there was some change into the dependency.


package.json
```diff
@@ -5,7 +5,7 @@
   "workspaces": {
     "packages": [
       "packages/*",
-      "apps/*",
+      "apps/website",
```
coryrylan: This will break the demo apps as they rely on the local built clarity packages for dependencies. This will prevent local installs from working which we use for day to day testing/dev on core. Let's hold off on changing this line. There are a couple of related PRs and work that I am going through that this would effect.

coryrylan (changes requested):

package.json
coryrylan: @gnomeontherun I think had some experience testing out the yarn zero install, he might be able to add some details.

bdryanovski: I tested the `zero-install` for yarn - it's not working that smooth as they promote it - also like I mentioned in the description there are issues with linking directly by the path and also keeping the `.yarn/cache` inside the repository will add a huge drawback on checkouts and overall git speed. 

Also, there some issues with the yarn 1x & 2x that are making them incompatible or at least with this state of linking packages between sub-projects. It was super hard to test it. 

I think that `yarn zero-install` is something that we could investigate next year after they release the stable version and provide better support.

coryrylan: We can make this change to exclude the apps from the yarn workspace but to keep the apps working for local installs in each app update the `@clr/core` package path (excluding website)

from `"@clr/core": "./packages/core/dist/core"`
to `"@clr/core": "../../dist/core"`

This will allow each app to be able to run yarn independently and still install. When they were part of the workspace the install only worked if the path started at the root of the workspace.

bdryanovski: Gonna update the `package.json` files to address the paths and run a few tests to see if all the packages are building correctly. 

Also will create a subtask to install the dependencies for each of the sub-projects.

gnomeontherun: Make sure this aligns with @bbogdanov's work please.

coryrylan: @bbogdanov showed me a really good solution for our local package usage that should clean up a lot of our build code. I think this change will still work fine for his update but +1 to double check.

bdryanovski: Yap, we had a chat with @bbogdanov and sync our approach on the issue. Manage to mark some point where we need to sync. 

So by mentioning it, we decide the one that manages to merge his PR last will have to solve the potential `git-conflict` inside the version inside `apps/`. (I'm not changing it right now - cause I'm not sure what will be the final version of the package that we gonna put (`@latest` or `=4.0.6` "exact version") - and wanted to keep this PR working at the moment)

And by saying that .. I hope to merge mine before him! 😆 

Update `apps/` packages to have their own `yarn.lock` file and be stand-alone apps. I had the idea to have `pre-hook` for installing `yarn` packages but notice that `yarn install` will be called without the need of doing it on every `build` call and even that the apps are pretty small there is no need for that (also could result of creating different `yarn.lock` file that had to be commit/track).

bbogdanov (approved):

bdryanovski:
After merging #5322 will revert some of my changes here related to `paths`

package.json
coryrylan: Now that this has rebased onto `next` I think this particular line will still break the demo apps. It excludes them from the workspace so when yarn is ran at the top level of the repo or in the specific demo app directory it will fail.

bdryanovski: @coryrylan update to address the changed path of the `core` dist

coryrylan: The path update looks good to me, I think the cds rename is all thats left

coryrylan (approved):
Looks good to me, all that would be left is rebase to the new @cds change I think

mathisscott (approved):

gnomeontherun:
@bdryanovski What is the status here?

bdryanovski:
I need to fix this yarn.lock merge conflict before that. I had some issues with it. Will try to fix it soon.

gnomeontherun:
What is the status here?

bdryanovski:
Will update the PR on Monday. 

When we drop `apps/*` from workspaces some of the packages stop building. The reason for that is that our dependence tree is having missing packages. Before that, they got their package as a dependency of some of the apps inside `apps/` and yarn resolve it for us by installing it into `<root>/node_modules` and linking back to local workspaces. 

This is an issue for two reasons:
  * Now packages have an unnamed dependency that could change without anyone noticing. Or you have a package with an unknown version. For example, the website depends on webpack - but its use whatever version of webpack is found into `<root>/node_modules/` that is provided from one of the `demo/apps` 
  * Building packages now depend on `demo/apps` that may be updated/removed/changed without notice. 

It a bit slow cause I need to drop/install all packages every time to make checks if `build:ci` pass and then run packages by hand to be sure!

The good thing here is that dropping `apps/*` (except website) is having visibly big improvement on installing packages and overall runtime on the build.

bdryanovski:
The pull request is updated and ready for review, also please spend some time reading the description of the PR and if something is not clear - poke me - so I could update it.

coryrylan (approved):

sis0k0 (approved):

hippee-lee (approved):

PR #5305, opened 2020-11-17
🥳 chore: updating Angular to 11.0.0

bdryanovski:
Update Angular to the latest stable release.


sis0k0 (approved):

packages/angular/package.json
```diff
@@ -42,25 +42,25 @@
...
     "zone.js": "0.10.3"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.1100.0-rc.2",
+    "@angular-devkit/build-angular": "0.1100.1",
     "@angular-devkit/build-ng-packagr": "0.1002.0",
-    "@angular/cli": "11.0.0-rc.2",
-    "@angular/compiler-cli": "11.0.0-rc.2",
-    "@angular/language-service": "11.0.0-rc.2",
+    "@angular/cli": "11.0.0",
```
hippee-lee, line 61: nit, the latest version at time of my review is `11.0.1`

hippee-lee (approved):

packages/angular/package.json
bdryanovski: Yap, you are right but will let it stay this version for now. It's only a dev dependency but it's always overwritten if there is a global install of this package.

Also from the changelog [here](https://github.com/angular/angular-cli/releases/tag/v11.0.1), it seems that there is nothing that critical

PR #5310, opened 2020-11-17
chore: add tooling for yarn link support across the monorepo

bbogdanov:
Providing tooling to support `yarn link` across the monorepo and remove the `tsconfig` `paths` usage inside each package so we make sure every single package we develop handles its own dependencies.

By removing `tsconfig` paths in our packages we are making sure every package handles its own dependencies.

By using `yarn link` and the provided tooling in this PR we are making sure every package that is used inside another package in the monorepo is using the local builds(the ones in `dist` folder) while developing instead of the published ones in NPM.

What we getting mainly from this approach is removing the necessary package to package linking by relative paths and copy/pasting the dist folder of each package inside another package `node_modules` so we can build the packages inside the monorepo. 

Outside of the monorepo, our published packages depend on `peerDependencies` and our users will need to install all needed packages.

The tooling will work even when we bring new packages to the repo.
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, we do not have a standard way of using packages inside packages and also testing them with their local builds.

## What is the new behavior?

Providing a workflow for linking packages together inside the monorepo without affecting end-users.

- [ ] Yes
- [x] No


package.json
```diff
@@ -27,12 +27,13 @@
...
   "scripts": {
     "ng": "ng",
+    "clean:workspace": "del ./**/dist && yarn clean:modules",
     "clean:modules": "del ./**/node_modules",
     "clean": "del ./dist && yarn run schematics:clean",
     "start": "yarn angular:dev:start",
     "build": "npm-run-all build:ci build:website build:storybook:angular build:storybook:core angular:dev:build",
     "build:ci": "npm-run-all clean format lint build:libs test",
-    "build:libs": "npm-run-all -p ui:build icons:build core:build -s angular:build schematics:build react:build",
+    "build:libs": "npm-run-all -p ui:build icons:build core:build -s link:packages angular:build schematics:build react:build link:packages",
```
bbogdanov: The first call is needed here to make sure when the `@clr/core` is built - it can be used inside other packages that depend on it and the last one is needed to `link` the rest of the packages cause in the first run they weren't ready to link

I believe also that this `link` actually can be skipped and when we need to interact with local builds we are going to run the task described on line `49` to make the local package linking.

scripts/link-packages.js
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Extract recusively all `package.json`s inside directory
+ *
+ * @param {string} directory absolute directory location
+ * @param {Array<{ name, file, dir }>} files list of package.json
+ * @return {Array<{ name, file, dir }>} list of package.json
+ */
+function extractRootPackageJsons(directory, files) {
+  // Excludes others than the built packages
+  if (directory.indexOf('node_modules') >= 0) {
```
bdryanovski:
```ts
directory.includes('node_modules')
```

package.json
coryrylan: A possible future improvement but not needed for this PR is to have the link script take an argument and each project can call it directly so each project would build then link itself. That might add a bit more time than linking them all at once but it would ensure each is built & linked in the correct order every time. Something like this maybe

```json
"build": "ng build --prod && node scripts/link-packages @clr/angular"
```

packages/angular/tsconfig.json
```diff
@@ -20,11 +20,7 @@
     "typeRoots": ["node_modules/@types"],
     "paths": {
       "@clr/angular": ["./projects/clr-angular/src/index.ts"],
-      "@clr/angular/*": ["./projects/clr-angular/src/*"],
-      "@clr/icons": ["../../dist/clr-icons"],
-      "@clr/icons/shapes/*": ["../../dist/clr-icons/shapes/*"],
-      "@clr/core": ["../../dist/core"],
-      "@clr/core/*": ["../../dist/core/*"]
```
coryrylan, line 27: 👍

packages/react/package.json
```diff
@@ -17,10 +17,12 @@
     "clean": "del ./dist .parcel-cache"
   },
   "dependencies": {
-    "@clr/core": "./packages/core/dist/core",
     "react": "^16.13.1",
     "react-dom": "^16.13.1"
   },
+  "peerDependencies": {
+    "@clr/core": "latest"
+  },
```
coryrylan, line 25: Thanks for fixing this 👍

apps/website/.vuepress/config.js
```diff
@@ -103,45 +103,4 @@ module.exports = {
...
-        'ramda/es/includes': path.resolve('../../packages/core/node_modules/ramda/es/includes.js'),
-        'ramda/es/curryN': path.resolve('../../packages/core/node_modules/ramda/es/curryN.js'),
-        'ramda/es/path': path.resolve('../../packages/core/node_modules/ramda/es/path.js'),
-        'ramda/es/has': path.resolve('../../packages/core/node_modules/ramda/es/has.js'),
-        'ramda/es/is': path.resolve('../../packages/core/node_modules/ramda/es/is.js'),
-        'ramda/es/isEmpty': path.resolve('../../packages/core/node_modules/ramda/es/isEmpty.js'),
-        'ramda/es/without': path.resolve('../../packages/core/node_modules/ramda/es/without.js'),
-      },
-    },
-  },
```
coryrylan, line 146: Its good to see that we can remove a lot of the custom path mapping code and tsconfigs

apps/core-angular-cli/package.json
```diff
@@ -22,7 +22,7 @@
     "@angular/platform-browser-dynamic": "~10.0.3",
     "@angular/router": "~10.0.3",
     "@clr/city": "^1.1.0",
-    "@clr/core": "./packages/core/dist/core",
+    "@clr/core": "latest",
```
coryrylan, line 25: Just to double check, does the build still properly swap and set the package versions in the final dist output for publication?

apps/website/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@sentry/browser": "5.15.5",
     "@sentry/integrations": "5.15.5",
+    "@clr/core": "^4.0.0",
```
coryrylan: Why was this one set to 4.0 instead of latest like the others?

scripts/link-packages.js
```diff
@@ -0,0 +1,176 @@
...
+   * e.g. @clr/angular -> @clr/core
+   */
+  distPackages.forEach(distPkg => {
+    if (!packageJson.dependencies[distPkg.name]) {
+      return;
+    }
+
+    packageConnector(distPkg, pkg);
+  });
+});
```
coryrylan: Do we ever have to worry about unlinking or are the link removed when the dist is deleted?

package.json
```diff
@@ -27,12 +27,13 @@
   },
   "scripts": {
     "ng": "ng",
+    "clean:workspace": "del ./**/dist && yarn clean:modules",
     "clean:modules": "del ./**/node_modules",
     "clean": "del ./dist && yarn run schematics:clean",
```
coryrylan: To simplify a bit I think `clean:workspace` and `clean:modules` could be combined into just `clean:workspace`. If you run `clean:modules` you should rebuild anyways since new dependencies would be installed.

scripts/link-packages.js
bbogdanov: the link is removed when the `dist` is deleted, yarn even notifies that when the build script runs with something like `link for package {smth} was removed`

package.json
bdryanovski: I could suggest using `pre` and `post` hooks

```
"prebuild": "link something"
"postbuild": "link something"

Comment (bbogdanov) [reply package.json]: I thought about this and in the end, when you want to publish the package, do you really want the build to run `link`?

File: apps/website/package.json (hunk 7 above)
Comment (bbogdanov) [reply apps/website/package.json]: I tried something with it but forgot to turn in back, thanks for seeing it!

File: packages/angular/tsconfig.json [hunk 10]
```diff
@@ -20,11 +20,7 @@
...
+      "@clr/angular/*": ["./projects/clr-angular/src/*"]
```
bdryanovski, line 23: I think that we could go without the last line. Have to be tested 

```
"@clr/angular/*": ["./projects/clr-angular/src/*"]
```

Is letting us access private API that is something that I prefer not to be able - cause it is super easy to create demos or code that is not gonna compile outside the Dev app.

If I remember correctly, I fixed all of them with the Angular 11-rc PR.

scripts/link-packages.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+const packageJsons = extractRootPackageJsons(process.cwd(), []); // Collect every root package.json inside the monorepo
+
+if (!packageJsons) {
+  console.log('Warning: No packages available!');
+  return;
+}
+
+const distPackages = packageJsons.filter(x => x.dir.indexOf(distFolder) >= 0);
```
bdryanovski: Same here .. `indexOf` is a bit cryptic `>= 0` or `< 0`

scripts/link-packages.js
```diff
@@ -0,0 +1,176 @@
...
+  if (!pkg.name) {
+    throw new ReferenceError('Package name is missing!');
+  }
+
+  const spawn = spawnSync('yarn', ['link'], {
+    cwd: pkg.dir,
+  });
+
+  if (spawn.stdout.exit === 1) {
+    console.log(`${pkg.name} failed to link!`);
```
bdryanovski: I think that we have to throw an error and fail how task. 

Why?:
 * If one or more links could not be created we are not sure that the next build/compile task will pass at all (it may .. but it may not)
 * if there is no error `yarn/npm` won't throw an error and will suppress it inside the log.

I think this task is critical for the build so if it's not working for any reason there is no real reason for starting anything else.

scripts/link-packages.js
```diff
@@ -0,0 +1,176 @@
...
+  const spawn = spawnSync('yarn', ['link', linkPkg.name], {
+    cwd: targetPkg.dir,
+  });
+
+  const spawnResponse = spawn.stdout.toString('utf8');
+
+  const message = `${linkPkg.name} => ${targetPkg.name}`;
+
+  if (spawnResponse.indexOf('yarnpkg.com') >= 0) {
+    console.error(`Error: ${message} `);
```
bdryanovski: I think here is the same as the above method. If we fail to link we won't know cause if this task did not return `process.code 1` (can't remember 0 or 1 was the bad one)

package.json
coryrylan: Ideally, no, but I think it's helpful when the build is consistent as possible. Link might add a bit of extra work, but it simplifies the workflow and reduces the chance that the build output is inconsistent. If we want to publish, see if a local branch will pass CI, or test for a production bug, it's always just run `build`.

The package build+link step for each project independently is an optimization that can be added later since this will help unblock a few other PRs.

apps/core-angular-cli/package.json
bbogdanov: Yea, you can verify that by seeing the package.json version inside `node_modules` folder of the target package.

gnomeontherun:
I see some errors thrown in the logs but it isn't stopping the builds. Check the logs carefully here.

scripts/link-packages.js
```diff
@@ -0,0 +1,218 @@
...
+
+  return files;
+}
+
+function main() {
+  const packageJsons = extractRootPackageJsons(process.cwd()); // Collect `package.json` every root package inside the monorepo
+
+  if (!packageJsons) {
+    console.log('Warning: No packages available!');
+    exit(0);
```
hippee-lee, line 197: If there are no packges in the repo, should this be an error?

scripts/link-packages.js
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  const distPackages = packageJsons.filter(x => x.dir.includes(distFolder)); // Get all package.jsons inside the the `distFolder`
+
+  if (!distPackages.length) {
+    console.log('No packages available to link!');
+
+    exit(0); // nothing to link
```
hippee-lee, line 205: This also seems like there should be an error. This script is run after the build:libs script so there should always be packages available to link, right?

scripts/link-packages.js
```diff
@@ -0,0 +1,218 @@
...
+
+  const spawn = spawnSync('yarn', ['link'], {
+    cwd: pkg.dir,
+  });
+
+  const errorMessage = spawn.stderr.toString('utf8');
+
+  const errorMessageIncludesWarning = errorMessage.includes('warning');
+
+  if (errorMessage && !errorMessageIncludesWarning) {
```
hippee-lee, line 105: Is this checking for error messages when the thread runs `yarn link` and ignoring all that have 'warning' in the string?

Would it be better to whitelist the `success Registered ${pkg.name}` and error if that isn't found?

hippee-lee (changes requested):
Thanks for this programmatic approach to linking to the builds. Here are some observations and notes from testing: 

I ran the build commands one at t time, there are some issues I see after the following:

`yarn test`
```
Generating browser application bundles (phase: building)...warning There's already a package called "@clr/core" registered. ...
Cannot create directory `/Users/mhippely/Projects/clarity/vmw-clarity/apps/core-angular-cli/node_modules/@clr/core`: ENOENT: no such file or directory, mkdir
```
This seems to indicate a problem to me. The concern here is two fold, this error didn't fail the build, e.g it still exited with (0). How do we know that it is not really an issue? If we are sure it is not an issue does that indicate that it should not happen?

`yarn build:website`
```
Module not found: Error: Can't resolve '@clr/core/toggle/register.js' in '/Users/mhippely/Projects/clarity/vmw-clarity/apps/website/.vuepress/theme'
resolve '@clr/core/toggle/register.js' in '/Users/mhippely/Projects/clarity/vmw-clarity/apps/website/.vuepress/theme'
```

This is the symlink in the websites node modules:
```core -> ../../../../../../../.config/yarn/link/@clr/core```
and this is whtas in the config (core is missing)
```
vmw-clarity on  pr/bbogdanov/5310 [$] is 📦 v4.0.0 via ⬢ v12.16.0 
•100% ➜ ll ~/.config/yarn/link/@clr     
total 0
drwxr-xr-x  6 mhippely  staff  192 Nov 18 16:55 .
drwxr-xr-x  4 mhippely  staff  128 Nov  6 14:06 ..
lrwxr-xr-x  1 mhippely  staff   57 Nov 18 15:48 angular -> ../../../../Projects/clarity/vmw-clarity/dist/clr-angular
lrwxr-xr-x  1 mhippely  staff   55 Nov 18 15:48 icons -> ../../../../Projects/clarity/vmw-clarity/dist/clr-icons
lrwxr-xr-x  1 mhippely  staff   51 Nov 18 15:48 react -> ../../../../Projects/clarity/vmw-clarity/dist/react
lrwxr-xr-x  1 mhippely  staff   52 Nov 18 15:48 ui -> ../../../../Projects/clarity/vmw-clarity/dist/clr-ui
```

That is as far as I've gotten with manual testing the effects of the changes here. 
I'm uncomfortable putting this in right now for two reasons:

1) what we have, while not ideal,  worked with all of the complex moving parts of the build. if these changes have any issues it means debugging this instead of developing, releasing or publishing. Given where we are with v5 I am not sure it is the right time to introduce a change like this. 
2) this seems like a good programmatic approach but there are a lot of edge cases and error handling that seems to be missing: 
    - the downstream scripts after build:libs (website/storybook) are not building for me when I run them one at at ime and it seems to indicate that something isn't working as intended. 
    - these changes don't seem to be reversible ... in that I can link things programmatically but there is no way to unlink and clean everything up. 
    - what if I have multiple repos (e.g: fork, vmw-clarity, vmw-clarity-v{2,3,4,5} won't they all use the same yarn link?

bbogdanov:
> Thanks for this programmatic approach to linking to the builds. Here are some observations and notes from testing: 
> 
> 
> 
> I ran the build commands one at t time, there are some issues I see after the following:
> 
> 
> 
> `yarn test`
> 
> ```
> 
> Generating browser application bundles (phase: building)...warning There's already a package called "@clr/core" registered. ...
> 
> Cannot create directory `/Users/mhippely/Projects/clarity/vmw-clarity/apps/core-angular-cli/node_modules/@clr/core`: ENOENT: no such file or directory, mkdir
> 
> ```
> 
> This seems to indicate a problem to me. The concern here is two fold, this error didn't fail the build, e.g it still exited with (0). How do we know that it is not really an issue? If we are sure it is not an issue does that indicate that it should not happen?
> 
> 
> 
> `yarn build:website`
> 
> ```
> 
> Module not found: Error: Can't resolve '@clr/core/toggle/register.js' in '/Users/mhippely/Projects/clarity/vmw-clarity/apps/website/.vuepress/theme'
> 
> resolve '@clr/core/toggle/register.js' in '/Users/mhippely/Projects/clarity/vmw-clarity/apps/website/.vuepress/theme'
> 
> ```
> 
> 
> 
> This is the symlink in the websites node modules:
> 
> ```core -> ../../../../../../../.config/yarn/link/@clr/core```
> 
> and this is whtas in the config (core is missing)
> 
> ```
> 
> vmw-clarity on  pr/bbogdanov/5310 [$] is 📦 v4.0.0 via ⬢ v12.16.0 
> 
> •100% ➜ ll ~/.config/yarn/link/@clr     
> 
> total 0
> 
> drwxr-xr-x  6 mhippely  staff  192 Nov 18 16:55 .
> 
> drwxr-xr-x  4 mhippely  staff  128 Nov  6 14:06 ..
> 
> lrwxr-xr-x  1 mhippely  staff   57 Nov 18 15:48 angular -> ../../../../Projects/clarity/vmw-clarity/dist/clr-angular
> 
> lrwxr-xr-x  1 mhippely  staff   55 Nov 18 15:48 icons -> ../../../../Projects/clarity/vmw-clarity/dist/clr-icons
> 
> lrwxr-xr-x  1 mhippely  staff   51 Nov 18 15:48 react -> ../../../../Projects/clarity/vmw-clarity/dist/react
> 
> lrwxr-xr-x  1 mhippely  staff   52 Nov 18 15:48 ui -> ../../../../Projects/clarity/vmw-clarity/dist/clr-ui
> 
> ```
> 
> 
> 
> That is as far as I've gotten with manual testing the effects of the changes here. 
> 
> I'm uncomfortable putting this in right now for two reasons:
> 
> 
> 
> 1) what we have, while not ideal,  worked with all of the complex moving parts of the build. if these changes have any issues it means debugging this instead of developing, releasing or publishing. Given where we are with v5 I am not sure it is the right time to introduce a change like this. 
> 
> 2) this seems like a good programmatic approach but there are a lot of edge cases and error handling that seems to be missing: 
> 
>     - the downstream scripts after build:libs (website/storybook) are not building for me when I run them one at at ime and it seems to indicate that something isn't working as intended. 
> 
>     - these changes don't seem to be reversible ... in that I can link things programmatically but there is no way to unlink and clean everything up. 
> 
>     - what if I have multiple repos (e.g: fork, vmw-clarity, vmw-clarity-v{2,3,4,5} won't they all use the same yarn link? 
> 
> 
> 
> 

Unfortunately, I am agree with you. After some thinking and discussion with Bozhidar this solution seems to bring more issues than it solves. I think I have another solution for the the problem we face. I prefer to discuss it before trying it cause it will take some time to prove it will work.

PR #5322, opened 2020-11-19
fix: local build package

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

This is a very similar but slightly different option to the solution in https://github.com/vmware/clarity/pull/5310 in case the yarn link does not work out. This still needs more testing to be sure but I wanted to put up what code I had so far as I wont be able look into this until tomorrow or next Monday.

We need a way to resolve and install locally built dependencies in the monorepo. This issue is a blocker for 5.0. Some examples of these new and existing dependency chains:

- core-react-app => @clr/react => @clr/core => lit-element, lit-html...
- angular-dev-app => @clr/angular => @cds/angular => @clr/core => lit-element, lit-html...
- website => @clr/core => lit-element, lit-html...

We have gotten around this so far with inconsistent build scripts copying files (@clr/core/react) or custom webpack setups (website). However with Angular packages now depending on Core there are new issues.

We cannot copy the build of Core for the Angular CLI projects to use like we do with other projects. Angular CLI has not worked well with the package hoisting feature that yarn has provided. Because of this every Angular project installs all
of its dependencies in isolation, dependencies like lit-html are not available unless explicitly listed and installed by yarn. However our library should not be listing upstream dependencies only direct dependencies like `@clr/core`.
This can cause dependency mismatch issues by not allowing normal dependency resolution by npm and yarn.

So this leaves two possible solutions that I am aware of. The first is outlined in https://github.com/vmware/clarity/pull/5310. Yarn link is how yarn would handle local package linking without using features of yarn 2. This would make some local development easier as we can help ensure consistency between the dependencies as they all point to the same local build output directory. We have had issues with yarn link with the Angular CLI in the past but some of those seem to have been improved.

This PR is what I had come up with that is a bit less than ideal but may be the option if yarn does not work. This uses a copy build strategy similar to what core does today but in a generic way like https://github.com/vmware/clarity/pull/5310.

The trick to making this work is two steps. First, we must copy the package.json files of our packages (@clr/react, @clr/core) into the root dist during the yarn preinstall.

```json
mkdir -p ../../dist/core/ && cp ./package.json ../../dist/core/
```

This will allow projects to refer to them locally in their package.json.

```json
{
  "@clr/core": "./dist/core"
}
```

This step is important as it allows yarn to install without building and ensure all upstream dependencies are installed. Example clr-angular needs core which needs lit-html installed.

The second step is after a package builds it calls a generic script which finds all the projects in the repo that have a dependency on the package and copies the build output into the node_modules directory of the project.

This strategy will allow all projects to refer dependencies in a consistent way and remove custom copy scripts, webpack configs and tsconfig mappings. This is similar to https://github.com/vmware/clarity/pull/5310 but swapping yarn link for a copy step. There may be a better way to do this as this is just extending what Core does today in a generic way for all packages.

~~This is missing another step however. When we build to publish we need to make sure we update all the dependencies with the correct version number. Currently this is being done with various inconsistent build scripts.~~

https://github.com/vmware/clarity/blob/next/packages/angular/projects/clr-angular/package.json#L14
https://github.com/vmware/clarity/blob/next/packages/ui/package.json#L11
https://github.com/vmware/clarity/blob/next/packages/icons/package.json#L10
https://github.com/vmware/clarity/blob/next/packages/react/package.js#L46

This is now updated to properly update and copy the version number when distributing the build as listed above. This also fixes an issue of building libraries multiple times.

- [ ] Yes
- [x] No


packages/react/package.json
```diff
@@ -4,10 +4,12 @@
   "scripts": {
     "start": "npm-run-all core:build dev:start",
     "dev:start": "yarn run clean && parcel index.html --cache-dir=./.parcel-cache",
-    "build": "npm-run-all -s build:ts build:package build:copy",
+    "build": "npm-run-all -s build:ts build:package build:copy build:distribute",
+    "build:distribute": "cd ../../scripts/ && node ./package-distribute.js @clr/react",
```
Shijir: You couldn't directly call the script without changing directories?
```
node ../../scripts/package-distribute.js @clr/react
```

Shijir (approved):

packages/react/package.json
coryrylan: I think I can with proper use of the path resolve I just havent had time to finish this up yet. It needs to resolve the paths regardless if the script is executed in the project or root repo directory.

hippee-lee (approved):

bbogdanov (approved):

bdryanovski (approved):
I tested it to one of my blocked PR and seems to work. 

So I think that we need to merge it soon so we could unblock the other PR ;)

coryrylan:
Working through this still. There were some issues with the paths aligning and version numbers. Will rebase and re-request reviews when I push the update soon.

PR #5337, opened 2020-11-30
feat(icons): export icon names in entrypoint of @clr/core/icon

### LINKED ISSUE #5335: Expose icon names in `core/src/icon/index.ts`
## Summary

**What is the change?**

Re-exporting icon names (e.g. `infoStandardIconName`) in `core/src/icon/index.ts` and not only `@clr/core/icon/shapes/<icon>`

**Why should it go in Clarity?**

When working in TypeScript, since the `shape` prop of the `CdsIcon` component is typed as a `string`, it's very easy to make a typo that goes unchecked by `tsc`, the test runner, etc. For example `<CdsIcon shape="info-standand" />`

There are 2 ways this could be addressed by the library, on the top of my head:

- Change the `shape` prop type from `string` to a union type: `... | 'info-standard' | ...`
- Exposing icon names and recommending their usage: `<CdsIcon shape={infoStandardIconName} />`

Both options can be combined for maximum type safety.
The first option is how Font Awesome works (EDIT: I forgot that Font Awesome _also_ supports the second option), for instance, and it has been helpful to us many times. But it's a bigger change in the library, and somewhat a breaking change (but if it breaks one's app, it was probably a typo). More importantly, it does not ensure that the `info-standard` icon was actually loaded by `ClarityIcons.addIcons()` so, while safer at compile time, it might still silently break at runtime.
The second option would be trivial to support here since you already expose the icon names in the `shapes` files, it would just be a bunch of extra exports in `icon/index.ts`. And it would also be completely backwards compatible, just an extra line [in the doc](https://clarity.design/storybook/core/?path=/story/components-icon-getting-started--page) to recommend TypeScript users to optionally use the icon name instead of the string.

I would be more than happy to open a PR if you think this is a good addition to the library.

## Examples

```tsx
import { ClarityIcons, infoStandardIcon, infoStandardIconName } from '@clr/core/icon';
import { CdsIcon } from '@clr/react/icon';

ClarityIcons.addIcons(infoStandardIcon);

<CdsIcon shape={infoStandardIconName} />
```

## Workarounds

At the moment, it can be done by loading the `shapes` files, but it's extra (and undocumented) boilerplate that does not make it very discoverable:

```tsx
import { ClarityIcons, infoStandardIcon } from '@clr/core/icon';
import { infoStandardIconName } from '@clr/core/icon/shapes/info-standard';
import { CdsIcon } from '@clr/react/icon';

ClarityIcons.addIcons(infoStandardIcon);

<CdsIcon shape={infoStandardIconName} />
```

astorije:
## PR Checklist

- _N/A_ Tests for the changes have been added (for bug fixes / features)
- _N/A_ If applicable, have a visual design approval

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/vmware/clarity/issues/5335

TypeScript users had to load `import { xyzIconName } from '@clr/core/icon/shapes/xyz'; in order to use type checked icon names.

## What is the new behavior?

TypeScript users can now load directly from: `import { xyzIcon, xyzIconName } from '@clr/core/icon';`

- [ ] Yes
- [x] No


packages/core/src/icon/icon.stories.mdx
```diff
@@ -24,6 +24,17 @@ ClarityIcons.addIcons(userIcon);
 <cds-icon shape="user"></cds-icon>
 ```
 
+TypeScript users may use the following in contexts where the markup is being type-checked (e.g. in TSX):
```
astorije: This wording is a little arbitrary, please feel free to ask me for corrections!

bbogdanov: Actually in can be used in JSX as well.

I would suggest something similar to 

```
To use the default by Clarity icons shapes import the icon name from `@clr/core/icon`
```

mathisscott: +1

mathisscott (changes requested):
LGTM, outside of @bbogdanov 's request on the documentation. The names weren't exposed previously because they were included in the icon shape tuple. When tuples get better support in Javascript, it will be easier to use the icon shape for this. But it meets your use case in the near term. Thanks, @astorije

packages/core/src/icon/icon.stories.mdx
astorije: Thank you very much, @bbogdanov and @mathisscott, done!

@bbogdanov, I wasn't entirely following the meaning of your sentence, so I worded it a little differently while taking your feedback into consideration. Does this work for you, or do you want me to edit further?

bbogdanov (approved):

packages/core/src/icon/icon.element.ts
```diff
@@ -35,6 +35,17 @@ import { hasIcon } from './utils/icon.service-helpers.js';
...
+ *
+ * ```tsx
+ * import { ClarityIcons, userIcon, userIconName } from '@clr/core/icon';
+ * import { CdsIcon } from '@clr/react/icon';
+ *
+ * ClarityIcons.addIcons(userIcon);
+ *
+ * <CdsIcon shape={userIconName} />
+ * ```
+ *
```
coryrylan: Let's move this comment block. The one in the MDX file is good however this is React/JSX specific syntax on the custom element that would not apply to all users. We can move this to the React specific wrapper component. You should be able to add a jsdoc comment right above line 6 here

https://github.com/vmware/clarity/blob/next/packages/react/src/icon/index.tsx#L6

astorije:
@mathisscott, anything else you need me to change on this? :)

packages/core/src/icon/icon.stories.mdx
```diff
@@ -24,6 +24,17 @@ ClarityIcons.addIcons(userIcon);
 <cds-icon shape="user"></cds-icon>
 ```
 
+In JavaScript, import the icon name from `@clr/core/icon` instead. This also provides type safety when used in TypeScript:
```
coryrylan: nit: let's clarify that this is JSX/React specific syntax so users are not confused on how to use the component. Something like this maybe,

```
If using JSX or TSX, import the icon name from `@clr/core/icon` and include it in the `shape` prop to improve type safety.
```

coryrylan (changes requested):

packages/core/src/icon/icon.stories.mdx
astorije: Done!

packages/core/src/icon/icon.element.ts
astorije: Done, is that how you meant?

bdryanovski (approved):
Rebase and it's ready to merge :)

mathisscott (approved):

astorije:
@bdryanovski, done, rebased on `next` and squashed! It just needs a review from @coryrylan to lift the block :D

coryrylan (approved):

PR #5347, opened 2020-12-02
fix: (combobox) simplify control state management

### LINKED ISSUE #5101: Stepper resets when having a combobox inside.
## Describe the bug

Stepper resets when having a combobox inside.

## How to reproduce

https://stackblitz.com/edit/clarity-v4-light-theme-zaowaw

1.  Go to https://stackblitz.com/edit/clarity-v4-light-theme-zaowaw
2.  Click on Next
3.  Select a value from combobox
4.  Stepper resets to first panel

If you change anything in first step to make the form non-pristine, it works as expected.

## Expected behavior

Stepper should not reset

## Versions

**App**

- Angular: 10.0.6
- Node: 8
- Clarity: 4.0.0-rc.1 (inside stackblitz, as I can't update there), but also on latest 4.0.2

**Device:**

- Type: MacBook
- OS: macOS
- Browser chrome, safari

## Additional notes

The listener in stepper for form status changes resets the panel, as the form is pristine.
https://github.com/vmware/clarity/blob/master/packages/angular/projects/clr-angular/src/accordion/stepper/stepper.ts#L76

Seems that the combobox control update which sets the form to non-pristine is after the check above.

Jinnie:
It was leading to lifecycle calls overhead in incorrect intermediate states.
closes: #5101

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There are intermediate pristine state changes, which get caught by the Stepper and it resets itself.

Issue Number: #5101 

## What is the new behavior?

All direct control-manipulations are removed and communication with the forms API goes only through the ControlValueAccessor interface. Stepper does not reset.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/forms/combobox/combobox.ts
+++ b/packages/angular/projects/clr-angular/src/forms/combobox/combobox.ts
@@ -256,15 +256,7 @@ export class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContainer>
         if (!this.multiSelect && newSelection && !newSelection.isEmpty()) {
           this.toggleService.open = false;
         }
-
         this.updateControlValue();
-        if (this.control && this.control.control) {
-          this.control.control.updateValueAndValidity();
-        }
-        // for the purposes of validation
-        if (this.onChangeCallback) {
-          this.onChangeCallback(this.optionSelectionService.selectionModel.model);
-        }
       })
     );
 
@@ -318,8 +310,8 @@ export class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContainer>
   }
 
   private updateControlValue() {
-    if (this.control && this.control.control) {
-      this.control.control.setValue(this.optionSelectionService.selectionModel.model);
+    if (this.onChangeCallback) {
+      this.onChangeCallback(this.optionSelectionService.selectionModel.model);
     }
   }
 


packages/angular/projects/clr-angular/src/forms/combobox/combobox.ts
```diff
@@ -318,8 +310,8 @@ export class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContainer>
   }
 
   private updateControlValue() {
-    if (this.control && this.control.control) {
-      this.control.control.setValue(this.optionSelectionService.selectionModel.model);
+    if (this.onChangeCallback) {
```
bdryanovski, line 313: nit: 

I will change here is `onChangeCallback` a function before trying to call like one.

We use `any` to define the type so there is no guarantee that this won't throw `is not a function error`

bdryanovski (approved):

packages/angular/projects/clr-angular/src/forms/combobox/combobox.ts
Jinnie: Thanks! It's only public, because the interface handler is, but it's not supposed to be overwritten by the users. It gets injected by the forms API.
Not that the suggestion is bad, but I don't expect it to lead to problems. If someone overrides with something else - he'll get no worse exception than the one I could give him.

bbogdanov (approved):

sis0k0 (approved):

PR #5373, opened 2020-12-09
fix: file input a11y improvement

jeeyun:

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When you click on the "X" button to clear the file chosen, there's no shift in focus.

## What is the new behavior?

When you click on the "X" button to clear the file chosen, the focus moves back to the "browse" button. This allows the SR to read out the button informing that the previously chosen file is no longer there.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/file/file.element.ts
+++ b/packages/core/src/file/file.element.ts
@@ -55,7 +55,7 @@ export class CdsFile extends CdsControl {
 
   protected get clearFiles() {
     return this.inputControl.files?.length
-      ? html` <cds-control-action @click="${() => this.updateLabel()}" aria-label="${this.i18n.removeFile}">
+      ? html` <cds-control-action @click="${() => this.updateLabelAndFocus()}" aria-label="${this.i18n.removeFile}">
           <cds-icon shape="times"></cds-icon>
         </cds-control-action>`
       : html``;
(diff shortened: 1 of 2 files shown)


packages/core/src/file/file.element.ts
```diff
@@ -64,15 +64,21 @@ export class CdsFile extends CdsControl {
...
 
-  private updateLabel(files?: FileList) {
+  private updateLabelAndFocus(files?: FileList) {
     if (files && files.length) {
       this.buttonLabel = files.length > 1 ? `${files.length} ${CommonStringsService.keys.files}` : files[0].name;
     } else {
       this.buttonLabel = CommonStringsService.keys.browse;
       this.inputControl.value = '';
+
+      const browseButton = this.shadowRoot?.querySelector('cds-button');
```
bbogdanov, line 76: What do you think about setting a `const enum` with our root element selectors e.g. `cds-button`, `cds-control`, `cds-icon`, etc.? I think it will be helpful to change the selector in the enum instead of search/replace. Some possible situation for this is for example changing the prefix of the library.

coryrylan (approved):

mathisscott (approved):

packages/core/src/file/file.element.ts
jeeyun: Not a bad idea and we can consider it in a separate PR! This PR contains unit tests that would fail if the prefix changed so it wouldn't go unnoticed.

PR #5388, opened 2020-12-15
fix: core form alignment with range and input icon

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There was some minor misalignments with the input status icons and range input handle in chrome and firefox.

## What is the new behavior?
Both alignments are now fixed.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/icon/icon.element.ts
+++ b/packages/core/src/icon/icon.element.ts
@@ -14,6 +14,7 @@ import {
   internalProperty,
   StatusTypes,
   isString,
+  pxToRem,
 } from '@cds/core/internal';
 import { html, LitElement, query, svg } from 'lit-element';
 import { styles } from './icon.element.css.js';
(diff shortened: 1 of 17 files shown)


packages/core/src/icon/icon.element.ts
```diff
@@ -155,7 +155,7 @@ export class CdsIcon extends LitElement {
       const dimension = `calc(100% + ${this.innerOffset * 2}px)`;
       this.svg.style.width = dimension;
       this.svg.style.height = dimension;
-      this.svg.style.margin = `-${this.innerOffset} 0 0 -${this.innerOffset}`;
+      this.svg.style.margin = `-${this.innerOffset}px 0 0 -${this.innerOffset}px`;
```
mathisscott: I'm concerned about how robust these px values here will be considering we are basing everything in rem.

Maybe we could add a function that takes an expected pixel value and converts it to a base-20 rem value?

mathisscott (changes requested):

hippee-lee (approved):

packages/core/src/icon/icon.element.ts
coryrylan: Good idea, I think I can also have it read the base font value from the css custom property. That way if it ever changes it will calculate correctly.

mathisscott: I don't know if we need to change relative to the basefont value tho. all of our rem values are base-20. so if these shift based on changes to the basefont size, that means they will get out of sync with all the other sizings that are still base-20.

coryrylan: Fixed up. I found a handful of other places we were using px instead of rem in component logic so fixed those up and made a `pxToRem` utill function.

mathisscott (approved):

PR #5390, opened 2020-12-15
fix: add new fallback font for core

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
This adds Metropolis as a fallback font if Clarity City fails to load in Core. Many teams using Clarity Core and Clarity UI will have both available so this will make the fallback more robust.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/build/tokens.ts
+++ b/packages/core/build/tokens.ts
@@ -339,8 +339,8 @@ const typography = {
   },
   baseFontSize: token('125%'),
   baseFontSizePx: token(20),
-  fontFamily: token("'Clarity City', 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
-  headerFontFamily: token("'Clarity City', 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
+  fontFamily: token("'Clarity City', Metropolis, 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
+  headerFontFamily: token("'Clarity City', Metropolis, 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
   topGapHeight: token('0.1475em'), // line-height eraser
   ascenderHeight: token('0.1703em'), // line-height eraser
   xHeight: token('0.517em'), // line-height eraser


packages/core/build/tokens.ts
```diff
@@ -339,8 +339,8 @@ const typography = {
   },
   baseFontSize: token('125%'),
   baseFontSizePx: token(20),
-  fontFamily: token("'Clarity City', 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
-  headerFontFamily: token("'Clarity City', 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
+  fontFamily: token("'Clarity City', Metropolis, 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
+  headerFontFamily: token("'Clarity City', Metropolis, 'Avenir Next', 'Helvetica Neue', Arial, sans-serif"),
```
mathisscott, line 343: There are bugs with the glyph alignment in Metropolis which would cause this to break layouts. Has this change been visually tested?

mathisscott (changes requested):

packages/core/build/tokens.ts
coryrylan: Yeah there are issues. I was thinking of adding since it's the fallback and breaks less than the other native font fallbacks. I had to do this in the recent hackathon since I couldn't get city to load so it was the next best fallback. I'll double check though to make sure I didn't miss anything.

coryrylan: So testing Metropolis as a fallback is visually an improvement than the system fallback fonts. However thinking over this more, allowing it as a fallback might discourage teams from correcting the source issue of properly loading City in their apps. Also we don't know what version of Metropolis it might load, it may not be the latest that is in clr-ui which could cause more confusion. Going to close out this PR. Here are some of the screenshots 

### City
city

### Metropolis
metropollis  

### Avenir Next
avenir-next

### Helvetica Neue
helvetica-neue

### Arial
arial

PR #5406, opened 2020-12-21
fix(checkbox): poc for setting checkbox values in react - DO NOT MERGE

### LINKED ISSUE #5397: CdsCheckbox indeterminate not working in React
## Describe the bug

using the `indeterminate` property in the React wrapper `CdsCheckbox` doesn't work

## How to reproduce

Stackblitz:

https://stackblitz.com/edit/react-ts-hraphp?file=index.tsx

1. Create a checkbox with refs on both the `CdsCheckbox` and `input` components
2. apply `indeterminite` to the input ref in `useEffect` (after the nativeElement promise has resolved)
3. Nothing happens

## Expected behavior

You can't apply the `indeterminate` attribute on the input directly in the JSX, you must do it via a ref.  You'll see how this works in the stackblitz with an unwrapped input, but nothing happens to the Clarity Checkboxes. 

## Versions

**App**

- React: 16.14
- Clarity: 4.0.7 (I also tried with v5)

## Additional notes

I dug into the code and confirmed that the input `ref` and `this.inputControl` from [checkbox.element.ts](https://github.com/vmware/clarity/blob/next/packages/core/src/checkbox/checkbox.element.ts#L56) are the same component. The mutation observer is being added to the element, but the callback is never being hit.

I tried adding a MutationObserver manually in my stackblitz and struggled to get that to fire on a vanilla input either though, so this may be an issue with react.

mathisscott:
• DO NOT MERGE
• if this is the way to fix this, we have a bigger refactor ahead with form controls
• opened up access to checked/indeterminate properties on CdsCheckbox
• Added syncProps to CdsCheckbox to sync incoming values on the wrapped input
• Added a manual test on the cds/react demo app to verify
• verified against code in the stackblitz of the reporting issue as well

Fixes: #5397 

I've verified this addresses the issue in Clarity React (#5397 ). But it makes some big changes to the underlying CdsCheckbox architecture and I want to make sure that @coryrylan has a chance to look into it before we decide to go this route.

Because these changes would likely impact other form controls at an architectural level...

### DIFF
--- a/packages/core/src/checkbox/checkbox.element.ts
+++ b/packages/core/src/checkbox/checkbox.element.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { internalProperty, listenForAttributeChange } from '@cds/core/internal';
+import { property, listenForAttributeChange, syncProps } from '@cds/core/internal';
 import { CdsInternalControlInline } from '@cds/core/forms';
 import { styles } from './checkbox.element.css.js';
 
@@ -31,12 +31,24 @@ import { styles } from './checkbox.element.css.js';
  * @cssprop --border-radius
  */
 export class CdsCheckbox extends CdsInternalControlInline {
-  @internalProperty({ type: Boolean, reflect: true }) protected checked = false;
+  @property({ type: Boolean }) checked = false;
 
-  @internalProperty({ type: Boolean, reflect: true }) protected indeterminate = false;
+  @property({ type: Boolean }) indeterminate = false;
 
   static get styles() {
     return [...super.styles, styles];
+  }
+
+  updated(props: Map<string, any>) {
+    super.updated(props);
+
+    if (props.has('checked') && this.inputControl.checked !== this.checked) {
+      syncProps(this.inputControl, this, { checked: true });
+    }
+
+    if (props.has('indeterminate') && this.inputControl.indeterminate !== this.indeterminate) {
+      syncProps(this.inputControl, this, { indeterminate: true });
+    }
   }
 
   firstUpdated(props: Map<string, any>) {
--- a/packages/react/App.tsx
+++ b/packages/react/App.tsx
@@ -31,12 +31,19 @@ interface AppState {
   modal2Open: boolean;
 }
 
+let isInd = false;
+
+function toggleIndeterminate(self: React.Component) {
+  isInd = !isInd;
+  self.setState({ isInd: isInd });
+}
+
 export default class App extends React.Component<{}, AppState> {
   buttonRef: React.RefObject<CdsButton>;
 
   constructor(props: any) {
     super(props);
-    this.state = { modalOpen: false, modal2Open: false };
+    this.state = { modalOpen: false, modal2Open: false, isInd: false };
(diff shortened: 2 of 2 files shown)


coryrylan:
Sorry, here's an update of where I got last week. I should have posted my notes sooner. I got a potential fix eod last week but didn't get time to open a PR or write a unit test. The issue filed not react specific as I thought but it is an issue that happens with all frameworks and plain JS. The source issue is somewhat of a CSS bug on our end as well as tricky browser behavior. 

Checkbox inputs have a `value` attr/prop as well as a `checked` attr/prop. The `value` represents what the boolean state is referring so `<input type="checkbox" value="enableDarkTheme" />` for example, while checked represents boolean checked state. We add an internal `checked` attr to our web component when the checkbox changes purely for CSS. 

React, and any other framework should able to interact with the native checkbox as this is the benefit of the API not being framework specific. The bug comes in if the `checked` attribute or property is set programmatically our web component does not see the update, so it doesn't apply our custom CSS to show the checked state. Our component was only listening for the change user input event.

Checkboxes have the `checked` attribute and the `checked` javascript property, but they don't stay in sync with each other. The attr can default the checkbox to true, but the attr remains if the JS checked property is set. For our component to properly apply the CSS, we need to know when the checked property or the checked attribute of the native checkbox has changed. 

To solve this, we can follow the same strategy similar to the indeterminate state attribute, but extend the checked input setter so whenever the checkbox `checked` property is set our css will reflect that.

This is what I had working but did not complete or get any tests written up

```typescript
  firstUpdated(props: Map<string, any>) {
    super.firstUpdated(props);
    this.inputControl.addEventListener('change', () => {
      this.checked = this.inputControl.checked;
      this.indeterminate = false;
    });

    this.listenForCheckedState();
    this.listenForIndeterminateState();
  }

  private listenForCheckedState() {
    // initialize the css checked attr hook
    this.checked = this.inputControl.checked || this.inputControl.hasAttribute('checked');

    // if checkbox checked attr changes update css attr hook
    this.observers.push(listenForAttributeChange(this.inputControl, 'checked', val => (this.checked = val !== null)));

    // if checkbox checked property changes update css attr hook
    const checkedProp = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'checked');
    Object.defineProperty(this.inputControl, 'checked', {
      get : checkedProp.get,
      set: (val) => {
        this.checked = val;
        checkedProp.set.call(this.inputControl, val);
      },
    }); 
  }

  private listenForIndeterminateState() {
    this.indeterminate = this.inputControl.indeterminate || this.inputControl.hasAttribute('indeterminate');
    this.inputControl.indeterminate = this.indeterminate;
    this.observers.push(listenForAttributeChange(this.inputControl, 'indeterminate', val => (this.indeterminate = val !== null)));
  }
  ```

mathisscott:
Thanks for the detailed response, @coryrylan 

Can we discuss this when you are back? I don't think there is an easy fix. I have a lot of other things to get done before the holidays and I'm not sure that the proposed fix above would work because I tried something similar and it didn't address the use case in issue #5397 

This PR _does_ work, however, and I'm unclear why one is preferred over the other. I think we both need to take a look at this when we are back.

PR #5400, opened 2020-12-18
fix: core themes

coryrylan:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the new behavior?
- fixes several core theme bugs and aligns closer to the design spec
- align status colors across components and fix missing states
- refactor theme status tokens to only require a single set vs primary/secondary improving performance 
- cleanup status and alert stories to better reflect statuses
- add status demo to storybook to easily see all components using status colors

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #5418, opened 2021-01-04
fix: core disabled/neural states theming

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
Fixes a few contrast issues with disabled/neutral states in core components such as toggles and object interaction styles in the base themes.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/build/tokens.ts
+++ b/packages/core/build/tokens.ts
@@ -451,7 +451,7 @@ const aliases = {
         hover: token(color.construction[1000]),
         active: token(color.construction[1000]),
         selected: token(color.construction[700]),
-        disabled: token(color.construction[200]),
+        disabled: token(color.construction[300]),
       },
     },
     app: {
@@ -479,8 +479,9 @@ const aliases = {
     },
     warning: {
       value: token(color.ochre[500]),
-      tint: token(color.ochre[50]),
+      tint: token(color.ochre[100]),
       shade: token(color.ochre[600]),
+      dark: token(color.ochre[800]),
     },
     danger: {
       value: token(color.red[700]),
--- a/packages/core/src/button/base-button.element.scss
+++ b/packages/core/src/button/base-button.element.scss
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
@@ -26,7 +26,6 @@
   @include vertical-center-content;
 
   -webkit-appearance: none !important;
-  background: var(--background);
   border-color: var(--border-color);
   border-radius: var(--border-radius);
   border-style: solid;
@@ -48,10 +47,7 @@
   user-select: none;
   vertical-align: middle;
   white-space: nowrap;
-
-  &:active {
-    box-shadow: 0 #{$cds-global-space-1} 0 0 var(--box-shadow-color) inset;
-  }
+  width: 100%;
 
   & > span {
     @include center-content('block');
@@ -74,50 +70,57 @@
   }
 }
 
-:host([contains-anchor]) .private-host {
-  padding: 0;
+:host(:active) .private-host,
+:host(:active) .private-host::after {
+  box-shadow: 0 #{$cds-global-space-2} 0 0 var(--box-shadow-color) inset;
 }
 
-:host(:active) .private-host {
-  box-shadow: 0 #{$cds-global-space-1} 0 0 var(--box-shadow-color) inset;
+.private-host::after {
+  --offset: calc(-1 * #{$cds-global-space-1});
+  background: var(--background);
+  border-radius: var(--border-radius);
+  position: absolute;
+  content: '';
+  top: var(--offset);
+  left: var(--offset);
+  bottom: var(--offset);
+  right: var(--offset);
+  inset: var(--offset); // once safari catches up positions can be deleted for inset
+  z-index: -1;
 }
 
 :host(:hover),
 :host(:focus),
 :host(:active) {
   .private-host::after {
-    background: var(--background);
-    border-radius: var(--border-radius);
-    filter: brightness(95%);
-    position: absolute;
-    content: '';
-    top: 0;
-    left: 0;
-    bottom: 0;
-    right: 0;
-    inset: 0; // once safari catches up positions can be deleted for inset
-    z-index: -1;
+    filter: brightness(90%);
   }
+}
+
+:host([action='outline']) {
+  .private-host::after {
+    filter: brightness(100%);
+    opacity: 0;
+  }
+}
+
+:host([action='outline']:hover),
+:host([action='outline']:focus),
+:host([action='outline']:active) {
+  .private-host::after {
+    opacity: 0.1;
+  }
+}
+
+:host([status='inverse']:hover),
+:host([status='inverse']:focus),
+:host([status='inverse']:active) {
+  --background: #{$cds-alias-status-disabled-tint};
 }
 
 ::slotted {
   line-height: 1em !important;
   color: inherit !important;
-}
-
-/** @deprecated slotted anchor deprecated in 4.0 in favor of wrapping element */
-::slotted(a) {
-  text-decoration: none !important;
-  display: block !important;
-  color: inherit !important;
-  font-size: inherit !important;
-  font-family: inherit !important;
-  margin: calc(var(--padding) * -1) !important;
-  padding: var(--padding) !important;
-  min-width: var(--min-width) !important;
-  line-height: 1em !important;
-  box-sizing: border-box !important;
-  outline: 0 !important;
 }
 
 ::slotted(cds-icon) {
@@ -134,9 +137,8 @@
 }
 
 :host([action='outline']) {
-  --background: #{$cds-alias-object-opacity-0};
-  --border-color: var(--color);
   --color: #{$cds-alias-status-info};
+  --border-color: var(--color);
 }
 
 :host([status='success'][action='outline']) {
@@ -157,15 +159,7 @@
   --background: #{$cds-alias-object-opacity-0};
   --border-color: #{$cds-alias-object-opacity-0};
   --color: #{$cds-alias-status-info};
-
-  .private-host {
-    box-shadow: none;
-  }
-}
-
-:host([action='flat']) .private-host:hover {
-  --background: #{$cds-alias-object-opacity-0};
-  --color: #{$cds-alias-status-info};
+  --box-shadow-color: #{$cds-alias-object-opacity-0};
 }
 
 :host([size='sm']) {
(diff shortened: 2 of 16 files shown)


packages/core/src/progress-circle/progress-circle.element.scss
```diff
@@ -21,6 +21,7 @@ svg {
 .private-host {
   @include center-content('block');
   @include equilateral(100%);
+  filter: brightness(95%);
```
coryrylan: This is a temp fix I think, Colin and I are going to make an audit through the themes for any other potential issues I may have missed.

mathisscott: What is adding a brightness filter attempting to fix?

mathisscott (changes requested):

packages/core/src/progress-circle/progress-circle.element.scss
coryrylan: There is a contrast issue with the background of the ring, it's slightly too light. This kind of was a bandaid fix, but I'm cool with removing it until I get a proper fix in with the theme status colors.

mathisscott: I see. Does the ring's background track matter that much? I would think the contrast of the foreground fill/rotation would be what a11y cares about most. The background track is decorative and doesn't convey any information whatsoever.

mathisscott (approved):

packages/core/src/progress-circle/progress-circle.element.scss
coryrylan: Agree, going to remove for now

coryrylan:
Pushed some doc fixes around neutral states and status component demos

coryrylan:
I missed some stashed changes in the PR and should have pushed them. Fixes for theming/contrast as well as button state fixes.

packages/core/build/tokens.ts
```diff
@@ -479,8 +479,9 @@ const aliases = {
     },
     warning: {
       value: token(color.ochre[500]),
-      tint: token(color.ochre[50]),
+      tint: token(color.ochre[100]),
       shade: token(color.ochre[600]),
+      dark: token(color.ochre[800]),
```
coryrylan, line 484: To meet contrast requirements warning needs an additional darker color. For now I think it will be the exception.

packages/core/src/button/base-button.element.scss
```diff
@@ -48,10 +48,7 @@
   user-select: none;
   vertical-align: middle;
   white-space: nowrap;
-
-  &:active {
-    box-shadow: 0 #{$cds-global-space-1} 0 0 var(--box-shadow-color) inset;
-  }
+  width: 100%;
```
coryrylan, line 50: This fixes the bug of the button not filling its space when displayed as a flex child. When using layout or flexbox and stretching the host would stretch but the private-host element did not fill its space correctly.

packages/core/src/button/base-button.element.scss
```diff
@@ -74,12 +71,8 @@
...
 }
 
-:host([contains-anchor]) .private-host {
-  padding: 0;
-}
-
-:host(:active) .private-host {
-  box-shadow: 0 #{$cds-global-space-1} 0 0 var(--box-shadow-color) inset;
+:host(:active) .private-host::after {
+  box-shadow: 0 #{$cds-global-space-2} 0 0 var(--box-shadow-color) inset;
```
mathisscott, line 75: Does this have any negative effect on the compact buttons? The reason we'd used the smaller spacing previously was because a larger box-shadow effect introduced visual issues.

packages/core/src/button/button.stories.ts
```diff
@@ -229,10 +229,10 @@ export function sizes() {
...
   return html`
     <div cds-layout="vertical gap:sm align:horizontal-stretch">
-      <cds-button block>Default ('md')</cds-button>
-      <cds-button block action="outline">Default ('md')</cds-button>
-      <cds-button block size="sm">Compact ('sm')</cds-button>
-      <cds-button block action="outline" size="sm">Compact ('sm')</cds-button>
+      <cds-button>Default ('md')</cds-button>
+      <cds-button action="outline">Default ('md')</cds-button>
+      <cds-button size="sm">Compact ('sm')</cds-button>
+      <cds-button action="outline" size="sm">Compact ('sm')</cds-button>
```
coryrylan: With the flex bug fix, block is not needed as the button now behaves like native dom elements when placed as flex children.

packages/core/src/button/button.element.ts
```diff
@@ -14,7 +14,7 @@ import { styles } from './button.element.css.js';
 
 export const iconSpinner = (size: string) => {
   const spinnerSize = size === 'sm' ? '12' : '18';
-  return html`<span class="button-status-icon" cds-layout="horizontal align:center"><cds-progress-circle class="button-spinner" size="${spinnerSize}"></cds-progress-circle></span></span>`;
+  return html`<span class="button-status-icon" cds-layout="horizontal align:center"><cds-progress-circle class="button-spinner" size="${spinnerSize}" status="info"></cds-progress-circle></span></span>`;
```
mathisscott, line 17: 👍

packages/core/build/tokens.ts
```diff
@@ -451,7 +451,7 @@ const aliases = {
         hover: token(color.construction[1000]),
         active: token(color.construction[1000]),
         selected: token(color.construction[700]),
-        disabled: token(color.construction[200]),
+        disabled: token(color.construction[300]),
```
coryrylan, line 454: Bumping up contrast for disabled text colors

packages/core/src/button/button.stories.ts
```diff
@@ -229,10 +229,10 @@ export function sizes() {
 export function block() {
   return html`
     <div cds-layout="vertical gap:sm align:horizontal-stretch">
-      <cds-button block>Default ('md')</cds-button>
-      <cds-button block action="outline">Default ('md')</cds-button>
-      <cds-button block size="sm">Compact ('sm')</cds-button>
-      <cds-button block action="outline" size="sm">Compact ('sm')</cds-button>
+      <cds-button>Default ('md')</cds-button>
```
mathisscott: Are we no longer supporting `block` buttons?

packages/core/src/button/icon-button.stories.ts
```diff
@@ -96,10 +96,10 @@ export function sizes() {
 export function block() {
   return html`
     <div cds-layout="vertical gap:sm align:horizontal-stretch">
-      <cds-icon-button block><cds-icon shape="user"></cds-icon></cds-icon-button>
-      <cds-icon-button block action="outline"><cds-icon shape="user"></cds-icon></cds-icon-button>
-      <cds-icon-button block size="sm"><cds-icon shape="user"></cds-icon></cds-icon-button>
-      <cds-icon-button block action="outline" size="sm"><cds-icon shape="user"></cds-icon></cds-icon-button>
+      <cds-icon-button><cds-icon shape="user"></cds-icon></cds-icon-button>
```
mathisscott: Same here. What's happening with block buttons?

packages/core/src/button/base-button.element.scss
```diff
@@ -135,16 +113,25 @@
...
 }
 
 :host([status='success'][action='outline']) {
   --color: #{$cds-alias-status-success};
+  --background-hover: #{$cds-alias-status-success-tint};
 }
 
 :host([status='danger'][action='outline']) {
   --color: #{$cds-alias-status-danger};
+  --background-hover: #{$cds-alias-status-danger-tint};
```
coryrylan: Missing hover states for outline buttons.

packages/core/src/button/button.element.scss
```diff
@@ -61,14 +61,17 @@
...
+    --border-color: #{$cds-alias-status-disabled} !important;
+    --color: #{$cds-alias-status-disabled} !important;
+  }
+}
+
 :host([disabled][status='inverse']),
 :host([disabled][action='outline']) {
-  --background: #{$cds-alias-object-opacity-0} !important;
-  --color: #{$cds-alias-status-disabled-tint} !important;
-  --border-color: #{$cds-alias-status-disabled-tint} !important;
```
coryrylan, line 68: This was duplicated in both bas button styles and button styles

mathisscott (changes requested):

packages/core/src/button/base-button.element.scss
coryrylan: I'll double check, probably need a selector for each it sounds like

packages/core/src/button/button.stories.ts
coryrylan: We could leave the attr in the css for now as a deprecation. The demos work without block because the buttons work correctly with flexbox and the layout utils

packages/core/src/button/base-button.element.scss
coryrylan: I see now, reverted change.

packages/core/src/button/button.stories.ts
coryrylan: Reverted the block attr removal then updated the demos to use block and not the layout stretch util

packages/core/src/button/icon-button.stories.ts
coryrylan: Reverted the block attr removal then updated the demos to use block and not the layout stretch util

mathisscott (approved):

packages/core/src/button/base-button.element.scss
jeeyun: Is there no hover color for the inverse outline button? Or does the default for outline (`$cds-alias-status-info-tint`) work for inverse?

packages/core/src/toggle/toggle.element.scss
```diff
@@ -1,11 +1,11 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
 @import './../styles/tokens/generated/index';
 
 :host {
-  --background: #{$cds-alias-status-neutral-tint};
+  --background: #{$cds-global-color-gray-400};
```
jeeyun: Does this mean that the background always be this color (in dark theme and other themes)?

packages/core/src/button/base-button.element.scss
coryrylan: Good catch, fixed!

packages/core/src/toggle/toggle.element.scss
coryrylan: Fixed, I think that was from testing various contrast levels, it should of been a updated status alias color

jeeyun (approved):

mathisscott (approved):

astorije:
Perfect, I was about to open a ticket with the following screencast but woosh it's already fixed, thank you so much @coryrylan  and @mathisscott! I know it seems small, but since we already migrated from `next.0` to `rc.0`, any chance you could release `rc.1` with this? 🙏

link-hover

(Posting the screencast here in case others noticed this)

EDIT: Actually, my bad, I think I read too fast and this PR does not fix this, which was introduced [here](https://github.com/vmware/clarity/pull/5400/files#diff-b70673935e9b71cfe543d77fab1dd87f4e0719d9f338178542239a8a32aa07e7L472-R472)

PR #5421, opened 2021-01-04
fix: core layout css gap shim

### LINKED ISSUE #5420: core: safari technical preview layouts
## Describe the bug

Next major release of Safari ads flex gap support while good causes a problem with the current shim for the current Safari release. With the next Safari version the css shim and native gap double up and cause excess gap space in layouts.

## How to reproduce

View Core layouts in Safari Technical Preview

## Expected behavior

Gap sizes should be correct.

## Versions
Core 5.x

**Device:**

- Safari

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The CSS gap shim for safari breaks in the next major Safari release.

Issue Number: 
fixes https://github.com/vmware/clarity/issues/5420 

## What is the new behavior?
Adds a `cds-supports` attr to easily set css values for certain feature support in browsers. For the current Safari version this ensures the css gap shim is only applied if flex gap is not supported. This will ensure the shim does not run on the next Safari version which does support CSS Gap. You can see this behavior using the Safari Technical Preview. The attr is needed instead of the [CSS @support selector](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) due to Safari reporting a false positive supporting gap due to css gap in grid not flexbox.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/internal/utils/supports.spec.ts
+++ b/packages/core/src/internal/utils/supports.spec.ts
@@ -1,16 +1,13 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { supportsFlexGap, supportsResizeObserver } from './supports.js';
-
 describe('browser support feature checks', () => {
-  it('should check if browser supports ResizeObserver', () => {
-    expect(supportsResizeObserver()).toBe(true);
-  });
-  it('should check if browser supports FlexGap', () => {
-    expect(supportsFlexGap()).toBe(true);
+  it('should check browser support value strings in cds-support attr', () => {
+    const supportValues = document.body.getAttribute('cds-supports');
+    expect(supportValues).toContain('js');
+    expect(supportValues).toContain('flex-gap');
   });
 });
--- a/packages/core/src/styles/module.shims.scss
+++ b/packages/core/src/styles/module.shims.scss
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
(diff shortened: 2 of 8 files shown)


packages/core/src/internal/utils/supports.spec.ts
```diff
@@ -1,16 +1,11 @@
...
 
-import { supportsFlexGap, supportsResizeObserver } from './supports.js';
-
 describe('browser support feature checks', () => {
   it('should check if browser supports ResizeObserver', () => {
-    expect(supportsResizeObserver()).toBe(true);
-  });
-  it('should check if browser supports FlexGap', () => {
-    expect(supportsFlexGap()).toBe(true);
+    expect(document.body.getAttribute('cds-supports')).toBe('js flex-gap');
```
mathisscott: It seems like this test is more brittle than it needs to be. Can we test if the attribute contains those two strings? Instead of exact matching them in that order?

packages/core/src/internal/utils/supports.spec.ts
```diff
@@ -1,16 +1,11 @@
...
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { supportsFlexGap, supportsResizeObserver } from './supports.js';
-
 describe('browser support feature checks', () => {
   it('should check if browser supports ResizeObserver', () => {
```
mathisscott: Is the title of this test correct? The test assertion doesn't seem to answer it.

packages/core/src/styles/module.shims.scss
```diff
@@ -10,6 +10,6 @@
 @import './layout/shims/gap';
 
 // older safari
-@media not all and (min-resolution: 0.001dpcm) {
+[cds-supports~='no-flex-gap'] {
```
mathisscott, line 13: 👍

mathisscott (changes requested):

packages/core/src/internal/utils/supports.spec.ts
coryrylan: nope, will fix

coryrylan: fixed

packages/core/src/internal/utils/supports.spec.ts
coryrylan: Split into two assertions

mathisscott (approved):

jeeyun (approved):

PR #5431, opened 2021-01-07
fix: core form style hooks, private host styles

### LINKED ISSUE #5397: CdsCheckbox indeterminate not working in React

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Fix Part 1

For some Core form components such as checkbox, toggle and radio we need to know when the state of the native input changes. We must know when these components are changed so we can update the CSS to reflect those changes.

In clr-ui for custom checkboxes we would use a combination of `:checked` and `::before`/`::after` to create the custom checkbox look. However in Core with shadow dom [Safari does not properly support these selectors for slotted elements](https://bugs.webkit.org/show_bug.cgi?id=178237).
To fix this we have a div in the component template that is styled base on if the component is in one of the possible states the native input can be in.

The current version would listen to the `change` event of the native input and update our CSS base on that update. However I missed the use case of a developer programmatically updating a native input. Example:

```typescript
const checkbox = document.querySelector('cds-checkbox input');
checkbox.checked = true; // checkbox would not show the checked styles
``` 

Our component would not see that `checked` was set only `change` events. This caused the bug reported in https://github.com/vmware/clarity/issues/5397. You can find a plain JS example of the bug here https://stackblitz.com/edit/typescript-nbrgi7 as well as in React here https://stackblitz.com/edit/react-ts-yaoseg 

The fix for this bug is to override the setter on the native input as well as listen for attr changes on the native element.
This way if the checkbox is updated by the user, or its checked property and attr is set our component will update.
This PR adds a util function called `getElementUpdates()` which given an element and property name will notify when the property or its corresponding attribute has changed.

## Fix Part 2

The second part of this PR is to help some of the confusion in https://github.com/vmware/clarity/issues/5397. When we style our components/theme based on component state we often use host attr selectors.

```html
<cds-button status="success">...</cds-button>
```
```css
:host([status='success']) {
  --color: green;
}
```

This works well, however, at times we need to use host attrs to set internal styles. Sometimes these internal style hooks can cause confusion with the public API. Example form components should be disabled by setting the `disabled` attr or property on the native input not `<cds-input>`.

```html
<cds-input>
  <label>disabled input</label>
  <input type="text" disabled />
</cds-input>
```

Our component will add a `disabled` attr to `cds-input` so internally we can easily update the css custom properties based on the native input state.

```html
<cds-input disabled> <!-- added/removed based on native input -->
  <label>disabled input</label>
  <input type="text" disabled />
</cds-input>
```

This is a point of confusion for our APIs as when debugging. It creates similarly named attrs in the DOM and uncertainty of what should be updated.

We could remove host attrs for internal styles but this often breaks or complicates the CSS custom property themes. Example given this css in a component the `.active` class creates a new instance of `--color` in different scope.

```css
  :host {
    --color: blue;
  }

  p {
    color: var(--color);
  }

  .active {
    --color: red;
  }
```

If a user attempts to override `--color` and `.active` is applied they will no longer be able to customize `--color` because it refers to the element with the active class and not the `--color` defined on the host. Other web component libraries have this same issue. Ionic solves it by keeping the style hooks on host to avoid this. See this example for more detail https://stackblitz.com/edit/lit-element-example-gf19vu 

So to allow internal host attrs for internal styles without breaking the theme API this PR follows something similar to Ionic by keeping the style hooks on the host element but prefixing or marking them as internal. Now `@internalProperty()` will prefix host attributes with an `_` if `reflect: true` is set and no custom attr name is provided.

```typescript
@internalProperty({ type: Boolean; reflect: true }) type = 'banner';
```
```
<my-element _type="banner"></my-element>
```

So now in our example of `cds-input` when the native text input disabled is set the output HTML will look like the following:

```html
<cds-input __disabled>
  <label>disabled input</label>
  <input type="text" disabled />
</cds-input>
```

While still not the most ideal this hopefully is a step in the right direction and does seem to align with some of the other web component library patterns. (See ionic toggle component for a good example) This is an internal change so no breaking changes and could be iterated on without worry of future breaking changes.

closes #5397 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/checkbox/checkbox.stories.ts
+++ b/packages/core/src/checkbox/checkbox.stories.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -27,7 +27,7 @@ export function API(args: any) {
     <cds-checkbox ...="${spreadProps(getElementStorybookArgs(args))}">
       <label>checked</label>
       <input type="checkbox" checked />
-      <cds-control-message .status=${args.status}>message message</cds-control-message>
+      <cds-control-message .status=${args.status}>message text</cds-control-message>
     </cds-checkbox>
   `;
 }
@@ -38,7 +38,7 @@ export function checkbox() {
     <cds-checkbox>
       <label>checkbox</label>
       <input type="checkbox" checked />
-      <cds-control-message>message message</cds-control-message>
+      <cds-control-message>message text</cds-control-message>
     </cds-checkbox>
   `;
 }
--- a/packages/core/src/forms/control-inline/control-inline.element.ts
+++ b/packages/core/src/forms/control-inline/control-inline.element.ts
@@ -1,11 +1,11 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { html, internalProperty } from 'lit-element';
-import { property } from '@cds/core/internal';
+import { html } from 'lit-element';
+import { EventEmitter, property, event, getElementUpdates, internalProperty } from '@cds/core/internal';
 import { styles } from './control-inline.element.css.js';
 import { CdsControl } from '../control/control.element.js';
 import { getStatusIcon } from '../utils/index.js';
@@ -32,6 +32,15 @@ export class CdsInternalControlInline extends CdsControl {
 
   /** @private */
   @internalProperty() isControlGroup: boolean;
+
+  /** @private */
+  @internalProperty({ type: Boolean, reflect: true }) protected checked = false;
+
+  /** @private */
+  @internalProperty({ type: Boolean, reflect: true }) protected indeterminate = false;
+
+  /** @private */
+  @event() protected checkedChange: EventEmitter<boolean>;
 
   protected supportsPrefixSuffixActions = false;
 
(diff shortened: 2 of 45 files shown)


packages/core/src/checkbox/checkbox.stories.ts
```diff
@@ -37,8 +37,8 @@ export function checkbox() {
   return html`
     <cds-checkbox>
       <label>checkbox</label>
-      <input type="checkbox" checked />
-      <cds-control-message>message message</cds-control-message>
+      <input type="checkbox" />
```
mathisscott: Are we no longer demo'ing how `checked` works from HTML?

packages/core/src/forms/control-inline/control-inline.element.ts
```diff
@@ -67,8 +76,29 @@ export class CdsInternalControlInline extends CdsControl {
     `;
   }
 
-  connectedCallback() {
-    super.connectedCallback();
-    this.setAttribute('cds-control-inline', '');
+  firstUpdated(props: Map<string, any>) {
+    super.firstUpdated(props);
+    this.inputControl.addEventListener('change', () => (this.checked = this.inputControl.checked));
+    this.observers.push(
```
mathisscott, line 82: Do we want to watch for `disabled` as well?

mathisscott (changes requested):

packages/core/src/checkbox/checkbox.stories.ts
coryrylan: Typo I think will fix 👍

packages/core/src/forms/control-inline/control-inline.element.ts
coryrylan: Yeah good catch, interestingly I realized clr-angular misses this use case as well it looks like.

coryrylan: Fixed

packages/core/src/checkbox/checkbox.stories.ts
coryrylan: fixed

mathisscott (approved):

jeeyun (approved):
A clean and elegant solution for the issue at hand. Thank you for your work on this!

PR #5448, opened 2021-01-14
feat: turn-on strictTemplates for Angular

### LINKED ISSUE #4744: Datepicker with reactive form and strictTemplates throws Type 'string' is not assignable to type 'Date'
## Describe the bug

The datepicker component in its reactive form implementation as shown in the [official documentation](https://clarity.design/documentation/datepicker) at the *Reactive Form Demo* section, throws the following typescript compilation error, when `strictTemplates: true` in `tsconfig.json`:

```
error TS2322: Type 'string' is not assignable to type 'Date'.
```

The app is compiled successfully when `strictTemplates` is disabled. 

## How to reproduce

1. Ensure `strictTemplates` is enabled `tsconfig.json`
```tsconfig.json
{
...
  "angularCompilerOptions": {
    ...
    "strictTemplates": true
  }
}
```
2. Use the example code, exactly as found in the [official documentation](https://clarity.design/documentation/datepicker) at the *Reactive Form Demo* section.

3. See typescript compilation error
```
ERROR in ...component.html:5:26 - error TS2322: Type 'string' is not assignable to type 'Date'.
```

## Expected behavior

The app should compile successfully as it does when `strictTemplates` is not enabled.

## Versions

**App**

- Angular: [9.1.11]
- Node: [14.4.0]
- Clarity: [3.1.4]

**Device:**

- Type: [MacBook]
- OS: [iOS > docker > node:14-alpine]
- Browser [chrome]
- Version [latest]

bdryanovski:
Enabled Angular compiler options for strictTemplates and resolving known errors.

When enabling the

```
"angularCompilerOptions": {
  "strictTemplates": true
}
```

Some changes were required to pass all the validation on build time.

Side-Effects (or lack of them): 
  *  No breaking changes are required to consumers of the `clr-angular` if your code was passing the `strictTemplates` before that.
  * Now by-default we will run with `strictTemplates: true` to prevent feature fixes.
  * There is a noticeable performance drop when running the build task.

Time with `strictTemplates: true` - ✨ **Done in 110.16s.**
Time with `strictTemplates: false` - ✨ Done in 73.92s.

So next time when we write an Angular Input that we expect to be used only as an argument for example:
```html
<div clrProp></div>
```
And this argument is something different than a `string` such as `number`, `boolean` etc. We have to know that `@Input` by default always returns `string` if they are not bound before that so we have to add a second type `string` and handle it inside our code.

```ts
@Input('clrProp') prop: boolean | string
```

I try to write an extra decorator to handle this for us, the issue is that we are assigning types to the `@Input` decorator and we still have to add the second type so there was no much use of just a decorator to convert any value to `boolean` or `number` 

```js
@Input('clrProp') 
@InputBoolean()
prop: boolean | string
```
The only bonus here is that `prop` will always be of type `boolean`, but to pass the compiler we still have to mention `string` as a possible value. This is not much more cleaner - so I dropped the idea.

There is an open issue for that in Angular - [it's from 2017 so I don't have big hopes for it.](https://github.com/angular/angular/issues/14761)
A proposal for solving this [issue](https://github.com/angular/angular/issues/21143) and one more [here](https://github.com/angular/angular/issues/19929)

We have an open issue for one of the errors so this PR must resolve it as well.

Fixes #4744

To test it run inside `packages/angular`
```bash
yarn build && yarn dev 
```
Everything must pass without errors;

## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -68,7 +68,7 @@ export declare abstract class ClrAbstractContainer implements DynamicWrapper, On
 }
 
 export declare class ClrAccordion implements OnInit, OnChanges, AfterViewInit, OnDestroy {
-    multiPanel: boolean;
+    multiPanel: boolean | string;
     panels: QueryList<ClrAccordionPanel>;
     subscriptions: Subscription[];
     constructor(accordionService: AccordionService);
@@ -306,8 +306,8 @@ export declare class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContai
     get id(): string;
     protected index: number;
     invalid: boolean;
-    set multiSelect(value: boolean);
-    get multiSelect(): boolean;
+    set multiSelect(value: boolean | string);
+    get multiSelect(): boolean | string;
     get openState(): boolean;
     optionSelected: ClrOptionSelected<T>;
     optionSelectionService: OptionSelectionService<T>;
@@ -660,7 +660,7 @@ export declare class ClrDatagridHideableColumn implements OnDestroy {
     set clrDgHidden(hidden: boolean);
     set clrDgHideableColumn(value: {
         hidden: boolean;
-    });
+    } | string);
     hiddenChange: EventEmitter<boolean>;
     constructor(titleTemplateRef: TemplateRef<any>, viewContainerRef: ViewContainerRef, columnsService: ColumnsService, columnState: BehaviorSubject<ColumnState>);
     ngOnDestroy(): void;
@@ -871,7 +871,7 @@ export declare class ClrDateContainer implements DynamicWrapper, OnDestroy, Afte
 
 export declare class ClrDateInput extends WrappedFormControl<ClrDateContainer> implements OnInit, AfterViewInit, OnDestroy {
     protected control: NgControl;
-    set date(date: Date);
+    set date(date: Date | string);
     dateChange: EventEmitter<Date>;
     set disabled(value: boolean | string);
     get disabled(): boolean | string;
@@ -889,7 +889,7 @@ export declare class ClrDateInput extends WrappedFormControl<ClrDateContainer> i
     onValueChange(target: HTMLInputElement): void;
     setFocusStates(): void;
     triggerValidation(): void;
-    static ngAcceptInputType_date: Date | null;
+    static ngAcceptInputType_date: Date | null | string;
 }
 
 export declare class ClrDatepickerModule {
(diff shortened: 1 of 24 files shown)


packages/angular/golden/clr-angular.d.ts
```diff
@@ -889,7 +889,7 @@ export declare class ClrDateInput extends WrappedFormControl<ClrDateContainer> i
     onValueChange(target: HTMLInputElement): void;
     setFocusStates(): void;
     triggerValidation(): void;
-    static ngAcceptInputType_date: Date | null;
+    static ngAcceptInputType_date: Date | null | string;
```
Jinnie, line 892: That's not the most supportable feature I've seen :( 
Still don't have a clue why we would need it.

Jinnie (approved):

packages/angular/projects/dev/src/app/datagrid/pagination/pagination.ts
```diff
@@ -25,8 +25,8 @@ export class DatagridPaginationDemo {
     this.users = inventory.all;
   }
 
-  pageChange($event: Event) {
-    console.log($event);
+  pageChange(pageNumber: number) {
+    console.log(pageNumber);
```
bbogdanov, line 29: Do we need this `console.log`?

packages/angular/golden/clr-angular.d.ts
```diff
@@ -1125,8 +1125,8 @@ export declare class ClrIfError extends AbstractIfState {
 }
 
 export declare class ClrIfExpanded implements OnInit, OnDestroy {
-    get expanded(): boolean;
-    set expanded(value: boolean);
+    get expanded(): boolean | string;
```
bbogdanov, line 1128: The getter is not part of an `@Input` so do we need the `string` type?

packages/angular/projects/dev/src/app/datagrid/pagination/pagination.ts
bdryanovski: This is for demo purpose inside the devApp, so yes :)

packages/angular/golden/clr-angular.d.ts
bdryanovski: I don't understand the question?

bdryanovski: getter and setters must return and work with the same type so they are coupled to each other.

```ts
 get expanded(): boolean | string {
    return this._expanded;
  }

  @Input('clrIfExpanded')
  set expanded(value: boolean | string) {
    if (typeof value === 'boolean') {
      this.expand.expanded = value;
      this._expanded = value;
    }
  }
```

Even that the getter method is on top of the `Input` they are linked

bbogdanov (approved):

PR #5460, opened 2021-01-18
fix: adoption guide update

bdryanovski:
* updating getting started page
* adding adoption guide to the website.yml (Github Action)
* adding netlify configuration
* Moving eslint documentation into getting started at one single page
* Small ui fixes

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -280,6 +280,16 @@
   to = "/storybook/core"
   status = 200
 
+[[redirects]]
+  from = "/adoption-guide"
+  to = "/adoption-guide/"
+  status = 200
+
+[[redirects]]
+  from = "/adoption-guide/*"
+  to = "/adoption-guide/"
+  status = 200
+
 # Support CORS on JSON files
 [[headers]]
   for = "/*.json"
--- a/packages/angular/projects/clarity-adoption/src/index.html
+++ b/packages/angular/projects/clarity-adoption/src/index.html
@@ -10,4 +10,21 @@
   <body>
     <app-root></app-root>
   </body>
+
+  <!-- Hotjar Tracking Code for clarity.design/adoption-guide -->
+  <script>
+    (function (h, o, t, j, a, r) {
+      h.hj =
+        h.hj ||
+        function () {
+          (h.hj.q = h.hj.q || []).push(arguments);
+        };
+      h._hjSettings = { hjid: 1762865, hjsv: 6 };
+      a = o.getElementsByTagName('head')[0];
+      r = o.createElement('script');
+      r.async = 1;
+      r.src = t + h._hjSettings.hjid + j + h._hjSettings.hjsv;
+      a.appendChild(r);
+    })(window, document, 'https://static.hotjar.com/c/hotjar-', '.js?sv=');
+  </script>
 </html>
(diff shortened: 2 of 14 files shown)


bbogdanov (approved):

netlify.toml
```diff
@@ -280,6 +280,11 @@
   to = "/storybook/core"
   status = 200
 
+[[redirects]]
+  from = "/adoption-guide/*"
+  to = "/adoption-guide"
```
hippee-lee: this needs to have a trailing slash

```to = "/adoption-guide/"```

You can also add a second redirect rule for `/adoption-guide` e.g:

```
[[redirects]]
  from = "/adoption-guide"
  to = "/adoption-guide/"
```

File: yarn.lock [hunk 2]
```diff
@@ -3329,7 +3329,7 @@
   integrity sha512-S9K+Q39BGOghyLHmR0Wdcmu1i1noSUk8HcvMj+3IaohZw02WFd99aPTQDHJeseXrXZP3CNovaSlePI0R11NcFg==
 
 "@cds/core@./dist/core":
-  version "5.0.0-rc.0"
+  version "5.0.0-rc.1"
```
hippee-lee, line 3332: Should this and the other locked versions below be `5.0.1`?

hippee-lee (changes requested):
I tried to access the app from the deploy preview but there seems to be an error. Doe sit work for you [here](https://6005b6c2ab6d727cd7aa7405--vmware-clarity.netlify.app/adoption-guide/)?

yarn.lock
bdryanovski: Probably need just to merge with `next` .. it's an old PR - before the `v5` release.

packages/angular/projects/clarity-adoption/src/index.html
```diff
@@ -10,4 +10,21 @@
   <body>
     <app-root></app-root>
   </body>
+
+  <!-- Hotjar Tracking Code for clarity.design/adoption-guide -->
+  <script>
```
hippee-lee, line 15: I would rather not have the feedback form on anything except the website. We don't use it on storybook either. 
If we need analytics GA is a better option.

hippee-lee (commented):
A few general thoughts:

1) since this is a standalone app it should live in `<root>/apps/`. I know we also build and publish the core storybook but that is heavily integrated with the core codebase. 
2) I'm not clear how folks will get to the application 
    - does it need something on the website so that people who want to see this example can find it?
3) will it be updated with more examples in the future? 

I'm sorry to keep blocking this ... I'm still unsure of the apps place as part of the website (and subdomains).

gnomeontherun:
What is happening here?

bdryanovski:
I Will try to update it on Monday, @hippee-lee is probably right that we need to see where we gonna locate the Adoption Guide and how we gonna promote it. Will try to draft some ideas and with pros and cons and will get back to it.

When it comes to the location, if it stays an Angular App I prefer to have it as part of the `packages/angular` so it will reuse the same angular version and tools that we use for the Angular and not have to maintain multiple versions. 

Gonna try to update or close it soon.

hippee-lee:
We don't have to block this b/c of navigation to or promotion. 
I do think it should live in the apps directory though. The dev app there is a legacy holdover form the very early days and it _should_ eventually be deleted. As I mentioned, storybook is a different story but as a demonstration app this seems to be more similar to whats in the apps directory and I think that there is a script that syncs the build artifacts so that they do not need to be manually updated to use `@latest`. Shouldn't the adoption app use the current build of the angular libs?

bdryanovski:
I gonna close this PR for now. 

Until then I will address the comments above.

PR #5467, opened 2021-01-19
fix(website): header docs in tablet view

bbogdanov:

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] clarity.design website / infrastructure changes

- [ ] Yes
- [x] No

### DIFF
--- a/apps/website/.vuepress/theme/layouts/Layout.vue
+++ b/apps/website/.vuepress/theme/layouts/Layout.vue
@@ -20,7 +20,11 @@
       <Sidebar :items="sidebarItems" :isSidebarOpen="isSidebarOpen" @isSidebarOpenChange="toggleSidebar()" />
       <div id="content-area" tabindex="-1" :class="{ 'content-area': true, 'home-page': $page.frontmatter.home }">
         <Home v-if="$page.frontmatter.home" class="make-it-scrollable" />
-        <div v-if="is404" class="make-it-scrollable" cds-layout="horizontal gap:md align:horizontal-center">
+        <div
+          v-if="is404"
+          class="make-it-scrollable"
+          cds-layout="horizontal gap:md p-t@sm:xl p-t:sm align:horizontal-center"
+        >
           <slot></slot>
         </div>
         <div class="page-wrapper" v-if="!$page.frontmatter.home && !is404">


bbogdanov:
@hippee-lee  I thought that the `do/dont` component was bringing the issue but I found out that the angular component demos are written manually so this is what is needed for the headers to look okay on a tablet.

apps/website/.vuepress/theme/global-components/DocInset.vue
```diff
@@ -25,6 +25,18 @@ export default {
   justify-content: center;
   align-items: center;
   flex-direction: column;
+
+  @media (max-width: 1024px) {
+    width: 100%;
+    overflow: auto;
+
+    header {
```
bbogdanov: @hippee-lee I am open to moving this one to another component or if we have some kind of global style file(didn't find one)

apps/website/angular-components/header/README.md
```diff
@@ -43,6 +43,8 @@ If you’re using icons on their own in the header, using tooltips will help use
 <div class="clr-col-sm-12 doc-do">
 <div class="doc-example">
 
+::: inset
```
hippee-lee: We don't want to uase inset here. THe original demo is using `.doc-wrapper` to set the gray background and container styles. 

The ported angular demos may not have retained their responsiveness from the old site but the way to address this is to use `./apps/website/.vuepress/theme/styles/overrides.scss` for any css needed to override built in behavior that is causing a problem as in inline demo on the site. Keep in mind that in the long term the angular components and demos (from the original docs site) will be removed in favor of the core components and I don't want to make changes to the theme for the Angular components. IMO it is better to use an override and leave a clear `// comment about angular header overside styles` so that it can be easily removed in the future.

hippee-lee (changes requested):
I left more details in the heasder docs but we don't want to modify the theme like this, its better to use the `override.scss` to document what /why for writing the css so it can be removed when the angular demo isn't needed any more.

apps/website/angular-components/header/README.md
bbogdanov: The issue we are facing here is that the `doc-example` has `::after` selector that adds the `Do` / `Dont` label and if I apply the overflow to the `doc-example` the `Do` / `Dont` label fits into the scrollable area and becomes invisible until you scroll. That's why I need a wrapper on the content so I can make the content wrapper scrollable without hiding the labels

I am up to set the styles in the `overrides.scss` but definitely need a wrapper for the content. Also applying the overflow to the `inset` helps all other examples be scrollable on a tablet view

PR #5516, opened 2021-01-26
fix: core visual fixes / tokens

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

## What is the new behavior?
- token adjustments from Design. Includes some color fixes, hsla use, missing link states, input group alignment
- fix with input group alignments, had a small extra gap when no cds-control-message was provided
- improve global bundle size by removing unnecessary fallback value with css vars when generating global tokens

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/build/build.tokens.ts
+++ b/packages/core/build/build.tokens.ts
@@ -3,23 +3,29 @@ import { Token, CdsTheme } from './token-utils';
 import * as fs from 'fs';
 
 fs.mkdir('../src/styles/tokens/generated', { recursive: true }, () => console.log('Building Tokens...'));
-fs.writeFileSync('../src/styles/tokens/generated/tokens.json', JSON.stringify(flattenTheme(baseTheme), null, 2));
+fs.writeFileSync('../src/styles/tokens/generated/tokens.json', buildJSONTokens(baseTheme));
+fs.writeFileSync('../src/styles/tokens/generated/tokens.css', buildCSSTokens(baseTheme));
 fs.writeFileSync('../src/styles/tokens/generated/_index.scss', buildSassTokens(baseTheme));
-fs.writeFileSync('../src/styles/tokens/generated/_public.scss', buildCSSTokens(baseTheme));
+
+function buildJSONTokens(baseTheme) {
+  return JSON.stringify(flattenTokens(baseTheme), null, 2);
+}
 
 function buildSassTokens(theme: CdsTheme) {
-  return Object.entries(flattenTheme(theme))
+  return Object.entries(flattenTokens(theme))
     .map(tokenEntry => `${tokenToSass(tokenEntry[1])}\n`)
     .join('')
     .trim();
 }
 
 function buildCSSTokens(theme: CdsTheme) {
-  const tokens = Object.entries(flattenTheme(theme));
+  const tokens = Object.entries(flattenTokens(theme));
   return `
 :root {
   ${tokens
-    .map(tokenEntry => `  --cds-${camelCaseToKebab(tokenEntry[1].name)}: ${convertCSSValue(tokenEntry[1])};\n`)
+    .map(
+      tokenEntry => `  --cds-${camelCaseToKebab(tokenEntry[1].name)}: ${convertCSSValue(tokenEntry[1], 20, false)};\n`
+    )
     .join('')
     .trim()}
 }
@@ -50,7 +56,7 @@ function tokenToSass(token: Token) {
   return `${dynamicVar}${staticVar}`;
 }
 
-function flattenTheme(theme: any) {
+function flattenTokens(theme: any) {
   function flatten(config: any, parent = ''): { [key: string]: Token } {
     return Object.entries(config)
       .map(([key, value]: [string, any]) => {
@@ -69,7 +75,7 @@ function flattenTheme(theme: any) {
   return flatten(theme);
 }
 
-function convertCSSValue(token: Token, base = 20) {
+function convertCSSValue(token: Token, base = 20, fallback = true) {
   let value = token.value;
 
   if (token.config.absolute) {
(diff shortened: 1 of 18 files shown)


packages/core/build/build.tokens.ts
```diff
@@ -80,7 +86,9 @@ function convertCSSValue(token: Token, base = 20) {
     // hsl
     value = `hsl(${token.value[0]}, ${token.value[1]}%, ${token.value[2]}%)`;
   } else if (token.value instanceof Token) {
-    value = `var(--cds-${camelCaseToKebab(token.value.name)}, ${convertCSSValue(token.value)})`; // reference
+    value = `var(--cds-${camelCaseToKebab(token.value.name)}${
+      fallback ? `, ${convertCSSValue(token.value, 20, fallback)}` : ''
+    })`; // reference
```
coryrylan, line 91: When generating the alias tokens we can omit the fallback value since it is defined within the same global file.

mathisscott (approved):

packages/core/src/styles/typography/_typography.scss
```diff
@@ -182,13 +182,18 @@
...
   line-height: inherit !important;
   font-size: inherit !important;
 
   &:hover {
-    color: $cds-alias-status-info-shade !important;
+    filter: brightness(85%);
+  }
+
+  &:visited:not([cds-text~='static']),
+  &[cds-text~='visited']:not([cds-text~='static']) {
```
coryrylan: This would be considered a small but new API, `static` disabling the visited color and `visited` for forcing the visited color regardless if the link was clicked. This is useful for applications where the navlinks dont need visited state or where the visited state needs to be controlled by the host router.

mathisscott: Do we need to add documentation about that to the website?

coryrylan: Yeah eventually it should. It falls under typography/foundation for core which is not yet in the website. The website points to storybook right now until those are added. I did update the storybook demo docs to show examples of each https://clarity.design/foundation/typography/

coryrylan: Also added note, open to different names, I wasn't sure what to call the disabled visited style option. Maybe `no-visit` would be more clear?

jeeyun: Is that a link that has been visited by the user but currently disabled?

coryrylan: Its for the use case of styling the link but don't change the color if it has been visited. Example if there are links being used in some kind of navigation sometimes apps don't want to show the visited purple color since eventually all the links would be visited.

packages/core/src/styles/typography/_typography.scss
```diff
@@ -182,13 +182,18 @@
...
 [cds-text~='link'] {
-  color: $cds-alias-status-info !important;
+  color: $cds-global-typography-link-color !important;
   text-decoration: underline !important;
   line-height: inherit !important;
   font-size: inherit !important;
 
   &:hover {
-    color: $cds-alias-status-info-shade !important;
+    filter: brightness(85%);
```
jeeyun: Is there a reason for switching to using filter here? My concern is needing different numeric values depending on themes (dark theme or another custom theme). In the past we had encountered issue with using a percentage like this because one value did not work across different themes.

coryrylan: Good point, I think I'll need to add a new token for hover states. We could expose the brightness as a property define specific hover tokens Maybe something like this

```typescript
// tokens
link: {
    color: {
      value: token(color.blue[800]),
      hover: token(color.blue[900]),
      visited: token(color.lavender[600]),
      visitedHover: token(color.lavender[700]),
    },
  },
```

```scss
--cds-global-typography-link-hover-brightness: brightness(85%) // new

or

--cds-global-typography-link-color: #{$cds-global-color-blue-400};
--cds-global-typography-link-color-hover: #{$cds-global-color-blue-500}; // new
--cds-global-typography-link-color-visited: #{$cds-global-color-lavender-300};
--cds-global-typography-link-color-visited-hover: #{$cds-global-color-lavender-400}; // new
```

packages/core/src/styles/typography/_typography.scss
jeeyun: Gotcha. I think `static` is ok there; `no-visit` seems more ambiguous. Also open to different / clearer names but don't have any suggestions atm. :)

packages/core/src/styles/typography/_typography.scss
jeeyun: Thank you! I think that would work great. I remember a case where the color contrast not being enough depending on the base color (when using a percentage). Being able to customize like this would future proof this. 👍

coryrylan: Added new hover tokens for links

bbogdanov (approved):

hippee-lee (approved):

astorije:
@coryrylan, just double-checking, is it intentional that hover and selected interactions are the same now? (for reference: hover used to be `gray[100]` and selected used to be `blue[50]`, now both are `blue[50]`)

Screen Shot 2021-02-05 at 12 34 18 AM

We are styling our custom datagrid implementation to mimic Clarity and the v5.0.0 version seemed nicer:

v5.0.0 | v5.0.1
--- | ---
before | after

Additionally, could you clarify when to use `active` versus `selected`? In the GIFs above, we use `selected` since it's literally a selection, but visually `active` would differentiate it, albeit being semantically incorrect.

Thanks! 😃

coryrylan:
@astorije The selected state is for when a component is selected but not in an active/hover state. So example would be a selected accordion header or vertical tab. https://clarity.design/storybook/core/?path=/story/foundation-interaction-styles--page#accordion-interaction 

I'll double check on the hover/selected color state to make sure that is correct.

astorije:
Oooh gotcha, active as in the CSS active state, i.e. when the mouse button is being pressed, that makes sense, thanks.
And thanks for double-checking on hover/selected!

packages/core/build/tokens.ts
```diff
@@ -344,6 +345,16 @@ const typography = {
...
   xHeight: token('0.517em'), // line-height eraser
+  link: {
+    color: {
+      value: token(color.blue[800]),
+      hover: token(color.blue[900]),
+      visited: {
+        value: token(color.lavender[600]),
+        hover: token(color.lavender[700]),
+      },
+    },
```
astorije, line 356: I believe the `cds-inline-button` colors have not been updated accordingly:

Screen Shot 2021-02-05 at 3 08 28 AM

PR #5524, opened 2021-01-27
fix: core safari flex shim

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

We use CSS Flex Gap for many of our layout utilities. Unfortunately as of Safari version 14 this is not supported. We provide a shim of sorts that provides a fallback using margins for safari. This already is supported in our global CSS. This bug fixes an issue when the web component uses the layout utils but does not have the css gap shim provided. Because of the shadow DOM the components do not have access to the global layout utils by default.

Issue Number: N/A

## What is the new behavior?

The fix for this pr is to dynamically check if flex gap is supported then to dynamically apply the css gap shim to the component. Because this is a runtime check in the browser, every component needs to know if the css should be applied for safari. Rather than adding a base class or function call to every component, I created a small base class that is applied when the component is registered and only if the browser needs it.

This is a short term fix as Safari Technical Preview does support Flex Gap. So the next release of Safari it will be supported and the shim will be no longer needed.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/internal/base/css-gap.base.spec.ts
+++ b/packages/core/src/internal/base/css-gap.base.spec.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { LitElement, html } from 'lit-element';
+import { createTestElement, removeTestElement, componentIsStable } from '@cds/core/test/utils';
+import { applyCSSGapShim } from './css-gap.base.js';
+import { browserFeatures } from '../utils/supports.js';
+
+class GapShim extends LitElement {}
+class GapShimSupports extends LitElement {}
+
+window.customElements.define('gap-shim-support', applyCSSGapShim(GapShimSupports));
+
+(browserFeatures as any).supports.flexGap = false; // force false flex gap support
+const shimClass = applyCSSGapShim(GapShim);
+(browserFeatures as any).supports.flexGap = true;
+window.customElements.define('gap-shim', shimClass);
+
+describe('CSS Flex Gap Shim', () => {
+  let testElement: HTMLElement;
+  let component: GapShim;
+  let componentSupports: GapShimSupports;
+
+  beforeEach(async () => {
+    testElement = await createTestElement(html`
+      <gap-shim></gap-shim>
+      <gap-shim-support></gap-shim-support>
+    `);
+    component = testElement.querySelector<GapShim>('gap-shim');
+    componentSupports = testElement.querySelector<GapShimSupports>('gap-shim-support');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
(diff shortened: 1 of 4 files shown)


packages/core/src/internal/base/css-gap.base.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    component = testElement.querySelector<DemoComponent>('cds-css-gap-shim');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  it('should only apply css gap shim attr if not supported', async () => {
+    await componentIsStable(component);
+    expect(component.getAttribute('_nfg')).toBe(null);
```
coryrylan: I have a test to ensure the shim is not applied by default but I couldn't think of an easy way to test for when it should be applied since the check is specific to the browser.

packages/core/src/internal/base/css-gap.base.ts
```diff
@@ -0,0 +1,30 @@
...
+export function applyCSSGapShim(elementClass: any) {
+  class GapShim extends elementClass {
+    connectedCallback() {
+      super.connectedCallback();
+      if (!browserFeatures.supports.flexGap) {
+        this.setAttribute('_nfg', '');
+      }
+    }
+  }
+  return browserFeatures.supports.flexGap ? elementClass : GapShim;
```
coryrylan, line 27: Not a big fan of this, as it adds a layer of inheritance but also not a fan of adding a check for the shim to all the components.

packages/core/src/internal/base/css-gap.base.spec.ts
mathisscott: Could you mock `browserFeatures.supports.flexGap` in a test?

mathisscott (approved):

packages/core/src/internal/base/css-gap.base.spec.ts
coryrylan: Was able to cast/override the support value to check the attr being set

mathisscott (approved):

bbogdanov (approved):

hippee-lee (approved):

packages/core/src/internal/base/css-gap.base.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it('should not apply css gap shim attr if not supported', async () => {
+    await componentIsStable(component);
+    expect(component.getAttribute('_nfg')).toBe(null);
+  });
+
+  it('should apply css gap shim attr if not supported', async () => {
+    (browserFeatures as any).supports.flexGap = false; // force false flex gap support
+    await componentIsStable(component);
+    expect(component.getAttribute('_nfg')).toBe(null);
```
jeeyun: Is this test case identical to the one above? Shouldn't the `_nfg` attribute have empty value (`''`) instead of `null` if it does not support flex gap?

jeeyun (approved):

PR #5533, opened 2021-01-28
fix: tag a11y enhancement

jeeyun:
Our tags can sometimes be a composite element, containing other elements inside such as an icon or a badge. In order to better inform a screen reader that those items are related, a11y team recommends we add a `role='group'` to the containing parent element (tag), in combination with an `aria-labelledby` assigned to the span inside the tag that usually contains the text.

Two possible solutions were discussed for this: 

1. Clarity core auto-assigning those attributes on tags.
2. Application developer doing this manually for each tag.

We have opted for the first solution, auto-assigning the attributes ourselves. In the case that a tag is a simple element with no children this is unnecessary since it doesn't contain other elements. As of publishing of this PR, the a11y team considers this the acceptable and preferred solution, since the gain we get outweighs it - making the tags in every application compliant and accessible without code changes on their part.

Additionally, we updated our documentation to add `role="image"` with `alt` attribute for the tags containing an icon. This does not change our API or component internals; this is expected to be added via the application developer for additional clarity for screen reader users.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/internal/base/button.base.ts
+++ b/packages/core/src/internal/base/button.base.ts
@@ -52,14 +52,9 @@ export class CdsBaseButton extends LitElement {
     this.tabIndex = 0; // initialize immediately so button can be focused synchronously
   }
 
-  protected firstUpdated(props: Map<string, any>) {
-    super.firstUpdated(props);
-    this.setupNativeButtonBehavior();
-  }
-
   protected updated(props: Map<string, any>) {
     super.updated(props);
-    this.updateButtonAttributes();
+    this.updateButtonAttributes(props);
   }
 
   /** This mimics the mouse-click visual behavior for keyboard only users and screen readers.
@@ -79,15 +74,20 @@ export class CdsBaseButton extends LitElement {
     }, 300);
   }
 
+  private setupNativeButtonBehavior(readonly: boolean) {
+    if (readonly) {
+      this.removeEventListener('click', this.triggerNativeButtonBehavior);
+      this.removeEventListener('keydown', this.emulateKeyBoardEventBehavior);
+    } else {
+      this.addEventListener('click', this.triggerNativeButtonBehavior);
+      this.addEventListener('keydown', this.emulateKeyBoardEventBehavior);
+    }
+  }
+
   /**
    * We have to append a hidden button outside the web component in the light DOM
    * This allows us to trigger native submit events within a form element.
    */
-  private setupNativeButtonBehavior() {
-    this.addEventListener('click', this.triggerNativeButtonBehavior);
-    this.addEventListener('keydown', this.emulateKeyBoardEventBehavior);
-  }
-
   private triggerNativeButtonBehavior(event: Event) {
     if (!this.readonly) {
       if (this.disabled) {
@@ -120,7 +120,7 @@ export class CdsBaseButton extends LitElement {
     });
   }
 
-  private updateButtonAttributes() {
+  private updateButtonAttributes(props: Map<string, any>) {
     this.isAnchor = this.parentElement?.tagName === 'A';
 
     if (this.isAnchor && this.parentElement) {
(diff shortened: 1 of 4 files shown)


packages/core/src/internal/base/button.base.ts
```diff
@@ -129,14 +129,19 @@ export class CdsBaseButton extends LitElement {
...
       this.ariaDisabled = null;
     } else {
+      this.role = 'button';
       this.tabIndexAttr = this.disabled ? -1 : 0;
       this.ariaDisabled = this.disabled ? 'true' : 'false';
     }
+
+    if (props.has('readonly')) {
+      this.setupNativeButtonBehavior(this.readonly);
+    }
```
jeeyun, line 145: The updating of attributes that happens in the first if/else goes through two `updated` cycles in Lit Element to complete. I'm not sure if some of the attributes have unique properly that makes it so, but this made it impossible to combine this if/else with the separate logic right below it (setting up the native button behavior) since the condition for `props.has('readonly')` was false in the second cycle resulting in inconsistent / incorrect behavior.

packages/core/src/internal/base/button.base.ts
```diff
@@ -79,15 +74,20 @@ export class CdsBaseButton extends LitElement {
     }, 300);
   }
 
+  private setupNativeButtonBehavior(readonly: boolean) {
+    if (readonly) {
```
mathisscott, line 78: 👍

packages/core/src/internal/base/button.base.ts
mathisscott: that's possibly caused by the update of the properties in this function.

mathisscott: i ran into something similar with animations where i updated a LitElement property and it almost immediately went through another update loop.

packages/core/src/internal/base/button.base.ts
```diff
@@ -79,15 +74,20 @@ export class CdsBaseButton extends LitElement {
...
+      this.removeEventListener('click', this.triggerNativeButtonBehavior);
+      this.removeEventListener('keydown', this.emulateKeyBoardEventBehavior);
+    } else {
+      this.addEventListener('click', this.triggerNativeButtonBehavior);
+      this.addEventListener('keydown', this.emulateKeyBoardEventBehavior);
+    }
```
mathisscott, line 84: Did we add any unit tests for all this? Or did we modify existing ones?

mathisscott (approved):

packages/core/src/internal/base/button.base.ts
jeeyun: These were already present and applied always, but now conditionally applying to buttons that are not readonly.

Looks like we never created a spec file for base button? I can create some tests for the base button in general and include coverage for this part.

mathisscott: I think we were testing through `button.spec.ts` to avoid duplication. @coryrylan , plz correct me if i'm wrong on that.

It may be enough to verify this is still adequately tested through those unit tests. Verify code coverage, etc.

coryrylan: Yeah thats correct, the base was being covered by the button.spec.ts so we could make sure we have a test there that checks that the event handler is removed in readonly. Eventually we could move the base button logic as its own spec file so the button.spec.ts only tests the cds-button specific behavior. That I think can be a separate PR for another time.

packages/core/src/tag/tag.element.ts
```diff
@@ -76,9 +78,14 @@ export class CdsTag extends CdsBaseButton {
...
-      <div class="private-host" cds-layout="horizontal align:vertical-center">
+      <div
+        class="private-host"
+        role="group"
+        aria-labelledby="${this.groupLabelId}"
+        cds-layout="horizontal align:vertical-center"
+      >
         <slot name="tag-icon"></slot>
-        <span class="tag-content" cds-text="lhe"><slot></slot></span>
+        <span id="${this.groupLabelId}" class="tag-content" cds-text="lhe"><slot></slot></span>
```
coryrylan, line 88: ❤️

packages/core/src/tag/tag.stories.ts
```diff
@@ -82,7 +81,6 @@ export function badgesStatus() {
       <cds-tag readonly status="danger"
         >Danger <cds-badge aria-label="notification 12" status="danger">12</cds-badge></cds-tag
       >
-      <cds-tag disabled status="info">Disabled <cds-badge aria-label="notification 12">12</cds-badge></cds-tag>
```
coryrylan, line 85: Thanks for removing, the disabled was a bit confusing in the readonly status colors. 👍

packages/core/src/internal/base/button.base.ts
jeeyun: Added unit test to check add / remove event handlers on `readonly` updates.

mathisscott (approved):

coryrylan (approved):

PR #4931, opened 2020-08-20
Fix typo and add functional wrapper for Clarity Core components

lukeshiru:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

- There was a typo in the exported button components which exported all as `cds-button`
- Currently the Core components are only wrapped with classes.

## What is the new behavior?

- The exported button components where fixed to include `cds-icon-button` and `cds-inline-button`.
- The new wrapper (`wrapCustomElement`):
  - Returns a functional component instead of a class.
  - Has a custom name for every component for react dev tools `CustomElement(tag-name)`, instead of having `ReactWrapperComponent`.
  - Doesn't require the `nativeElement` "hack", it uses `ref` directly (using `forwardRef` internally).
  - Syntax can be cleaner:
```tsx
// Instead of this:
export class CdsAlert extends createReactComponent<CdsAlertType>('cds-alert') {}

// We can just do this:
export const CdsAlert = wrapCustomElement<CdsAlertType>('cds-alert');
```

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #5535, opened 2021-01-28
fix: react update base wrapper to lit implementation

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current React wrapper had an issue with not being able to use a ref to directly access the underlying custom element. This is an issue when you need to programmatically interact with an element. Example calling `focus()` on a `cds-button`. The initial workaround was providing an alternative `nativeElement` API that returns a promise with the reference once the element has rendered.

Issue Number: N/A

## What is the new behavior?
This PR updates the base to use the new lit-element wrapper with some modifications. This wrapper is not yet published on NPM but once published we should be able to use it directly. There are some slight modifications to improve some of the typing. This version primarily fixes our issue of refs not working. The wrapper uses a middle component between the ref and custom element so there is always a component to refer, fixing our issue of the ref not being populated.

This is not a breaking change however it does improve the TypeScript typing so it may catch bugs that once were silent errors.
- [ ] Yes
- [x] No


packages/react/App.tsx
```diff
@@ -32,18 +32,20 @@ interface AppState {
...
 
   componentDidMount() {
-    this.buttonRef.current.nativeElement.then(element => {
-      element.focus();
-    });
+    // (this.buttonRef.current as any).nativeElement.then((element: any) => {
+    //   element.focus();
+    // });
+
+    this.buttonRef.current.focus();
```
coryrylan, line 48: Refs now work as expected. To prevent a breaking change there is still a `nativeElement` property available on the API.

packages/react/src/accordion/__snapshots__/index.test.tsx.snap
```diff
@@ -3,24 +3,49 @@
...
 <div>
   <CdsAccordion>
-    <cds-accordion>
-      <CdsAccordionPanel
-        expanded={true}
+    <cds-accordion(CustomElement)
+      __forwardedRef={null}
+    >
+      <cds-accordion
+        __forwardedRef={null}
```
coryrylan, line 10: The snapshots are updated due to a middle layer created from the lit wrapper fixing the ref issue.

packages/react/src/converter/react-wrapper.tsx
```diff
@@ -1,74 +1,251 @@
...
     }
-  };
-}
+  }
+
+  const element = React.forwardRef((props?: UserProps, ref?: React.Ref<unknown>) =>
+    createElement(ReactComponent, { ...props, __forwardedRef: ref } as any, props?.children)
+  );
+
+  element.displayName = `${componentName}`; // edit to preserve component name in DevTools - Cory
```
coryrylan, line 249: I added a comment to any lines I specifically changed from the original lit wrapper.

mathisscott (approved):
Will this change allow us to close out #4931 ?

bbogdanov (approved):

jeeyun (approved):

PR #5538, opened 2021-01-29
chore(panel): add internal panel component

bbogdanov:
Introducing the `panel` component.


bbogdanov:
#5563

packages/core/src/internal-components/overlay/overlay.element.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('Overlay helper functions: ', () => {
   });
 });
 
-describe('Overlay element: ', () => {
+fdescribe('Overlay element: ', () => {
```
mathisscott: 🤔

packages/core/src/card/card.stories.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+- `cds-card`
+
+## Slots
+
+- `default`: defines card content section
+
+### CSS Custom Properties
+
+[Panel CSS properties]('/packages/core/src/card/panel.stories.mdx') with some changes on default values
```
mathisscott: Why `/packages/core/src/card/panel.stories.mdx`? Shouldn't there be a `/packages/core/src/card/card.stories.mdx`?

packages/core/src/internal-components/panel/panel.element.scss
```diff
@@ -0,0 +1,4 @@
+:host {
+  width: 100%;
```
mathisscott: might need a bit more in the way of styling. there are going to be some css custom properties that we need (background, border, box-shadow, padding... all come to mind). And those will likely need to be applied to the `.private-host` wrapper.

You will also want to test out this sizing against various types of containers and in all the supported browsers. This may be fine. Or you may find you need min/max heights and/or flexbox.

Also, what happens if a panel's content exceeds its height??? Do we want an overflow on the panel wrapper itself? Or would it be better to have an attribute that designates which part of the panel we'd like to overflow? Or is that something best left to product teams? Thoughts, @coryrylan?

packages/core/src/internal-components/panel/panel.element.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  describe('Container filling ', () => {
+    it('should fill the container space', async () => {
+      await componentIsStable(component);
+      expect(component.clientWidth).toBe(wrapperElement.clientWidth);
```
mathisscott: Love the unit tests on the height... Check the other simple components like the accordion header element spec. We often do a sanity check test to also make sure the component renders and projects content as expected.

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -0,0 +1,21 @@
...
+export const CdsInternalPanelTagName = 'cds-internal-panel';
+
+export class CdsInternalPanel extends LitElement {
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  render() {
+    return this.hasAttribute('cds-layout')
+      ? html`<slot></slot>`
```
mathisscott: I think we should always have the `.private-host` wrapper... But maybe not in this instance... I'm not sure what conversations you've had on this with other team members.

packages/core/src/card/card.stories.mdx
bbogdanov: This link is actually a link to the panel. I updated the path.

packages/core/import-map.importmap
```diff
@@ -66,6 +66,8 @@
     "@cds/core/internal-components/close-button/register.js": "/base/dist/core/internal-components/close-button/register.js",
     "@cds/core/internal-components/overlay": "/base/dist/core/internal-components/overlay/index.js",
     "@cds/core/internal-components/overlay/register.js": "/base/dist/core/internal-components/overlay/register.js",
+    "@cds/core/internal-components/panel": "/base/dist/core/internal-components/panel/index.js",
+    "@cds/core/internal-components/panel/register.js": "/base/dist/core/internal-components/panel/register.js",
```
bbogdanov, line 71: @mathisscott  What I was actually missing to run the unit tests. I thought at first I needed only to specify the `index.js` then I saw the overlay also specified the `register.js` and turns out this was the missing piece.

packages/core/src/internal-components/panel/panel.element.spec.ts
bbogdanov: Do we want to have `connectedCallback` as the `accordion-header` [does](https://github.com/vmware/clarity/blob/be9bc1eb71ee20c8e0668c79c7f810f99335c097/packages/core/src/accordion/accordion-header.element.ts#L43)?

packages/core/src/internal-components/panel/panel.element.scss
bbogdanov: If we place this one only to `.private-host` the `<cds-internal-panel>` will not fill the parent container.  Verified.

packages/core/src/internal-components/panel/panel.element.scss
```diff
@@ -0,0 +1,34 @@
...
+  --height: 100%;
+  --color: initial;
+  --overflow: visible; // default html behaviour
+  --overflow-x: visible; // default html behaviour
+  --overflow-y: visible; // default html behaviour
+  --background: #{$cds-alias-object-container-background};
+  --border-color: initial;
+  --border-width: #{$cds-alias-object-border-width-100};
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --padding-horizontal: #{$cds-global-space-0};
```
mathisscott: Instead of doing things like `padding-horizontal` and `padding-vertical` we use `--padding` where we can. This is to reduce the scope of the API.

packages/core/src/internal-components/panel/panel.element.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  describe('create the panel component', () => {
+    it('should create the component', async () => {
+      await componentIsStable(component);
+      expect(component.innerText).toBe(placeholderText);
```
mathisscott, line 23: 👍

packages/core/src/internal-components/panel/panel.stories.ts
```diff
@@ -0,0 +1,162 @@
...
+    Vivamus aliquet sem massa, eget sodales sapien dictum eu.
+    Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos.
+    Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos.
+    Suspendisse non interdum sem, nec egestas nulla.
+    Duis vitae leo justo. Fusce ante enim, tincidunt nec faucibus sed, fringilla et ipsum.
+    Proin quis diam elit. Cras nec finibus libero. Sed bibendum lectus eget magna ultricies eleifend.
+    Duis ut molestie urna. Proin cursus dolor ac ex rutrum, eu condimentum elit tincidunt.
+    In a dolor faucibus, vehicula ipsum eu, viverra lorem. Aenean congue fermentum commodo.
+    Sed ex purus, consectetur mollis urna eget, vestibulum mollis dolor.
+    Donec sodales, felis sit amet gravida ornare, magna dolor ullamcorper turpis, vel facilisis odio ipsum semper sem.
```
mathisscott, line 34: 🏆

mathisscott (changes requested):

packages/core/src/card/card.stories.mdx
```diff
@@ -0,0 +1,60 @@
...
+  <cds-divider></cds-divider>
+
+  <div>{ main content }</div>
+
+  <cds-divider></cds-divider>
+
+  <cds-progress-bar></cds-progress-bar>
+
+  <footer>{ footer }</footer>
+</cds-card>
```
coryrylan: Is this file supposed to be in the PR? This looks like card component specific not panel

packages/core/src/internal-components/panel/panel.element.scss
```diff
@@ -0,0 +1,33 @@
...
+  --width: 100%;
+  --height: 100%;
+  --color: initial;
+  --overflow: visible; // default html behaviour
+  --overflow-x: visible; // default html behaviour
+  --overflow-y: visible; // default html behaviour
+  --background: #{$cds-alias-object-container-background};
+  --border-color: initial;
+  --border-width: #{$cds-alias-object-border-width-100};
+  --border-radius: #{$cds-alias-object-border-radius-100};
```
coryrylan, line 12: You could simplify style API by combining the border color and width properties as one, that will also allow border type to be overridden.

```css
--border:  #{$cds-alias-object-border-width-100} solid initial;
```

packages/core/src/internal-components/panel/panel.stories.mdx
```diff
@@ -0,0 +1,42 @@
...
+- `--border-color`
+- `--border-width`
+- `--border-radius`
+- `--padding`
+- `--box-shadow`
+- `--overflow`
+- `--overflow-x`
+- `--overflow-y`
+- `--width`
+- `--height`
```
coryrylan: You don't have to list these out, instead you can generate a API table with storybook

```tsx

## API

<Props of={'cds-internal-panel'} />
```

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @cssprop --border-color
+ * @cssprop --border-width
+ * @cssprop --border-radius
+ * @cssprop --padding
+ * @cssprop --box-shadow
+ */
+export class CdsInternalPanel extends LitElement {
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('slot', 'panel');
```
coryrylan: I don't think we want to add a default named slot to the panel. If this were to be projected into another component without a slot named `panel` or if needs to project multiple panels into different slots it will wont be able to render.  Components only get projected into another components default unnamed `<slot></slot>` if they don't have a slot name.

packages/core/src/card/card.stories.mdx
bbogdanov: This is just the API proposal of the card. If you think it is better to be added as a separate PR, let me know.

packages/core/src/internal-components/panel/panel.element.scss
bbogdanov: That one was something I thought about and then forgot to raised it as a "how to" question in the channel and eventually forgot about it. Thank you! Awesome fix!

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @cssprop --box-shadow
+ */
+export class CdsInternalPanel extends LitElement {
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  render() {
+    return this.hasAttribute('cds-layout')
+      ? html`<div class="private-host"><slot></slot></div>`
```
mathisscott: You may want to test this. While layouts will work across slots, I'm not sure they will work across a slot inside a wrapper element. In the event it does not, you might consider having a `layout` property that projects onto `.private-host`. In the event it does, you're good.

mathisscott (changes requested):

gnomeontherun:
If this is only an internal component, then it is not a `feat` commit message. It would be a `chore` I think.

packages/core/src/card/card.stories.mdx
coryrylan: Yeah, we should keep it separate. Storybook is still the primary docs for Core so this would show up on the site with the next release.

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -0,0 +1,48 @@
...
+export class CdsInternalPanel extends LitElement {
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  render() {
+    const layout = this.getAttribute('cds-layout') || 'vertical';
+
+    return html`
+      <div class="private-host" cds-layout="${layout}">
```
mathisscott: Since this is across the shadow boundary, I'm curious as to how this translates. Thoughts, @coryrylan ?

mathisscott (changes requested):

packages/core/src/internal-components/panel/panel.element.ts
coryrylan: Yeah this will be an issue. The layout util cannot be dynamically assigned like this since the unused CSS is compiled/treeshaken out of the templates. Also the CSS can only apply to one level, it cant cross boundaries easily between shadow/light DOM. It would need to be something like this https://github.com/vmware/clarity/blob/next/packages/core/src/modal/modal-actions.element.ts#L40

bbogdanov: Seems like we are getting back to the initial implementation of `render` method? Also if we do what we did in the `modal-action` element we will get back to the previous situation where the component won't have a `private-host` class.

mathisscott: `.private-host` in this component can still be useful, even if it just a wrapper for a `slot`.

In theory, people could style the background, border, border-radius and padding of the panel through the CSS custom properties. All of those would/should be applied to `.private-host`.

bbogdanov: If i understand you correctly we will get back to the concern that was mentioned [here](https://github.com/vmware/clarity/pull/5538#discussion_r580437399). 

When I verified the concern is on point I updated the implementation to resolve the issue we were about to face and I end up with the current implementation. 

Example of the issues with the previous implementation 

```
 <cds-panel cds-layout="horizontal">
    <div class="private-host> // <- the only element `cds-layout` will apply styles to
       //content 
    </div>
 </cds-panel>
```

I might be missing something but this is how I see it currently.

mathisscott: No, that's not what I'm getting at. What I'm saying is...

1) have the `<cds-panel>` greedily take up any space (height/width) available to it
2) have _no cds-layout_ specified on the `cds-panel` component
3) have a `div.private-host` internal to the `cds-panel`. Why? so that people could potentially use CSS or cds-layout padding utilities to force the `.private-host` to have, basically, a margin of its own.
4) End users would then use `cds-layout` on the contents that are then rendered in the slot inside `.private-host`. So if they want it to be vertical/horizontal, cool. Go for it.

As for how this translates to cards, you'd likely want to do either what we are doing in `form-group.element.ts` with the `layout` property. But I don't think it's needed for a panel. Panels are more generic. Cards are more specific and a little more complicated. I could see forcing a layout on panels via a property as complicating the development of things like drawers on the vertical axes or even the detail pane of a datagrid (both of which should/would use the panel as a base).

bbogdanov: Got it!

coryrylan (approved):

mathisscott (approved):

mathisscott (approved):

mathisscott (approved):

PR #5574, opened 2021-02-04
fix: datagrids in forms with checkboxes

### LINKED ISSUE #4289: Extra padding on datagrid checkbox when its inside clrForm
## Describe the bug

In data grid, checkbox column gets extra padding when it is wrapped inside a clrForm
Root cause - checkbox label gets additional classes "clr-col-12 clr-col-md-2" when wrapped inside "clrForm"

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v2-qkdlgk

## Expected behavior

No extra padding should be added for checkbox inside the datagrid

## Versions

**App**

* Angular: [8.2.0]
* Clarity: [2.1.0]

hippee-lee:
This fixes an issue when datagrid is placed inside the clrForm. Because ClrLabel (a directive that overloads the [label] element) checks for the ClrForm services and a label with one or more clr-col classes on an element inside the ClrForm instance, we need to prevent that when a consumer wraps a datagrid in a form. The addition of  class does that without adding any actual styles to the selection checkboxes.

closes #4289

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When consumers embed a datagrid inside a form the ClrLabel directive adds `clr-col-12 clr-col-2` classes to the label and this pushed the select all checkbox column over. 

Issue Number: #4289 

## What is the new behavior?
The change adds a `clr-col-null` class to the inputs for select all and row checkboxes to prevent the unwanted classes being added. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.html
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.html
@@ -42,16 +42,21 @@
           [ngClass]="{ 'clr-form-control-disabled': !clrDgSelectable }"
           role="gridcell"
         >
-          <input
-            clrCheckbox
-            type="checkbox"
-            [ngModel]="selected"
-            (ngModelChange)="toggle($event)"
-            [id]="checkboxId"
-            [attr.disabled]="clrDgSelectable ? null : true"
-            [attr.aria-disabled]="clrDgSelectable ? null : true"
-            [attr.aria-label]="commonStrings.keys.select"
-          />
+          <div class="clr-checkbox-wrapper">
+            <input
+              type="checkbox"
+              [ngModel]="selected"
+              (ngModelChange)="toggle($event)"
+              [id]="checkboxId"
+              [attr.disabled]="clrDgSelectable ? null : true"
+              [attr.aria-disabled]="clrDgSelectable ? null : true"
+              [attr.aria-label]="commonStrings.keys.select"
+            />
+            <!-- Usage of class clr-col-null here prevents clr-col-* classes from being added when a datagrid is wrapped inside clrForm -->
+            <label [for]="checkboxId" class="clr-control-label clr-col-null">
+              <span class="clr-sr-only">{{commonStrings.keys.select}}</span>
+            </label>
+          </div>
         </div>
         <div
           *ngIf="selection.selectionType === SELECTION_TYPE.Single"
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid.html
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid.html
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+  ~ Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
   ~ This software is released under MIT license.
   ~ The full license information can be found in LICENSE in the root directory of this project.
   -->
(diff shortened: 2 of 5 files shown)


packages/angular/projects/clr-angular/src/data/datagrid/datagrid.html
```diff
@@ -20,14 +20,19 @@
...
+                    <div class="clr-checkbox-wrapper">
                       <input
-                        clrCheckbox
                         type="checkbox"
+                        id="dg-select-all"
                         [(ngModel)]="allSelected"
                         [attr.aria-label]="commonStrings.keys.selectAll"
                       />
-                    </span>
+                      <!-- Usage of class clr-col-null here prevents clr-col-* classes from being added when a datagrid is wrapped inside clrForm -->
```
mathisscott, line 30: 👍

mathisscott (approved):

packages/angular/projects/clr-angular/src/data/datagrid/datagrid-row.html
```diff
@@ -42,16 +42,21 @@
...
+              type="checkbox"
+              [ngModel]="selected"
+              (ngModelChange)="toggle($event)"
+              [id]="checkboxId"
+              [attr.disabled]="clrDgSelectable ? null : true"
+              [attr.aria-disabled]="clrDgSelectable ? null : true"
+              [attr.aria-label]="commonStrings.keys.select"
+            />
+            <!-- Usage of class clr-col-null here prevents clr-col-* classes from being added when a datagrid is wrapped inside clrForm -->
+            <label [for]="checkboxId" class="clr-control-label clr-col-null">
```
mathisscott, line 56: Where is `.clr-col-null` defined? I don't see it in this PR so I'm wondering if it already existed.

hippee-lee: It isn't defined, thats why I chose it - it prevents the ClrLabel `@Directive` from adding `clr-col-12 and clr-col-2` classes. See [here](https://github.com/vmware/clarity/blob/next/packages/angular/projects/clr-angular/src/forms/common/label.ts#L45). Thats also why I didn't write a datagrid test for this change - there is testing coverage for this functionality in the label spec.

hippee-lee: What I really wanted to do was use the `disableGrid()` fn below but its not exposed and I didn't want to go the route of adding a reference and calling it in the right spot for the datagrid lifecycle.

packages/angular/projects/clr-angular/src/data/datagrid/datagrid.html
```diff
@@ -20,14 +20,19 @@
                     class="datagrid-column datagrid-select datagrid-fixed-column"
                     *ngIf="selection.selectionType === SELECTION_TYPE.Multi"
                   >
-                    <span class="datagrid-column-title">
+                    <div class="clr-checkbox-wrapper">
                       <input
-                        clrCheckbox
                         type="checkbox"
+                        id="dg-select-all"
```
Shijir: So shouldn't we use a unique id here instead of a hardcorded id? This hardcoded `id` would be repeated in case there are multiple datagrids.

Shijir (changes requested):

packages/angular/projects/clr-angular/src/data/datagrid/datagrid.html
hippee-lee: Right, I forgot to change that. Good catch.

Shijir (approved):

PR #5579, opened 2021-02-05
fix(icons): backwards compatibility for the static components in v5

bbogdanov:
In order to keep the icons backward compatibility in v5, we need to get back the styles that affect the deprecated icons so people have time to switch to Clarity Core Icons. 

This fix only affects people that are using the static versions of the components we have. 

We saw such behavior in the timeline component in our website that we had to fix by switching to `cds-icon` in #5539

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Static implementations will receive strange icon behavior because of missing CSS.

## What is the new behavior?

Static implementations will look as they should be in versions 4, 3, and 2.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/popover/tooltip/_tooltips.clarity.scss
+++ b/packages/angular/projects/clr-angular/src/popover/tooltip/_tooltips.clarity.scss
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
@@ -215,8 +215,10 @@
       margin-right: 0;
     }
 
-    cds-icon > svg {
-      pointer-events: none;
+    & {
+      > svg {
+        pointer-events: none;
+      }
     }
   }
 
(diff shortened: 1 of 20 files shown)


packages/angular/projects/clr-angular/src/popover/tooltip/_tooltips.clarity.scss
```diff
@@ -215,8 +215,11 @@
       margin-right: 0;
     }
 
-    cds-icon > svg {
-      pointer-events: none;
+    cds-icon,
+    clr-icon {
+      svg {
```
mathisscott: This should probably be `& > svg` to better reflect the original SASS. I don't think it makes that much difference. But just calling out that...

```
cds-icon { 
    svg {
```

...is not the same as....

`cds-icon > svg {`

mathisscott (approved):

packages/angular/projects/clr-angular/src/popover/tooltip/_tooltips.clarity.scss
bbogdanov: Thank you for catching that!

packages/angular/projects/clr-angular/src/utils/_theme.dark.icons.clarity.scss
```diff
@@ -1,7 +1,38 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
+/**
+ * TODO: Remove after v7 is released
+ *
+ * This one is left for backwards compatibility
```
hippee-lee, line 8: We might want to add an `@deprecated` here but thats not a blocker.

hippee-lee (approved):

PR #5593, opened 2021-02-09
feat: core enable token formats

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
This PR adds beta/experimental formats for design tokens
- json
- js/ts
- scss,
- xml (android)
- swift (ios)

This will also make it easier for storybook/website to pull in the token metadata for demo use cases. Marked these formats as beta/experimental so we can get feedback on the format output and can adjust as needed.

- [ ] Yes
- [x] No


packages/core/build/build.tokens.ts
```diff
@@ -1,57 +1,155 @@
...
+// Public API Tokens
+buildCSSTokens('../dist/core/styles/module.tokens.css');
+buildJSONTokens('../dist/core/tokens/tokens.json');
+buildJSTokens('../dist/core/tokens/tokens.ts');
+buildAndroidXMLTokens('../dist/core/tokens/tokens.android.xml');
+buildIOSSwiftTokens('../dist/core/tokens/tokens.ios.swift');
+buildSassTokens('../dist/core/tokens/tokens.scss');
+
+// Internal API Tokens for custom elements with fallback values
+buildInternalSassTokens('../src/styles/tokens/generated/_index.scss');
```
coryrylan, line 22: The internal sass tokens are for the custom elements to provide the vars with a inlined fallback in case the global tokens fail.
`var(--token-color, #fff)`

hippee-lee (approved):

packages/core/build/build.tokens.ts
```diff
@@ -97,3 +194,22 @@ function convertCSSValue(token: Token, base = 20, fallback = true) {
 function camelCaseToKebab(val: string) {
   return val.replace(/([A-Z]|\d+)/g, '-$1').toLocaleLowerCase();
 }
+
+function isHSL(value: any) {
+  return Array.isArray(value) && value.length === 3;
```
mathisscott, line 199: This is probably fine for now but I'm wondering what it would do if it were passed an array of rgb values? HSL is `[0-359][0-100][0-100]`. RGB is `[0-255][0-255][0-255]`

I don't see how this would catch that.

Perhaps leave a note for a possible future cleanup? Or would it be unnecessary here because we are controlling all values?

mathisscott (approved):

packages/core/build/build.tokens.ts
coryrylan: Yeah since we control the values it wasnt something I thought needed to be checked. However I think in the future we can work towards making this script more generic to the point of it being its own cds package with the token formats.

Issue #5419, opened 2021-01-04
core: reduced motion theme


coryrylan:
_Please review our [Contribution guide](https://github.com/vmware/clarity/blob/master/CONTRIBUTING.md) for more details about how to contribute effectively._

## Summary

_Describe the change or feature in detail. Try to answer the following questions._

- What is the change?
- Why should it go in Clarity?
- Does this change impact existing behaviors? If so how?
- If this change introduces a new behavior, is this behavior accessible?

## Use case

We have some animation timing values that could be moved into the tokens file so they are easily changed by a theme.

## Examples
Refactoring the animation timing in the CSS would help make a simple reduced motion theme. I have a prototype of this on a branch from the core theming work I rebased https://github.com/coryrylan/clarity/commit/e1b9eb384f920796b610992f16b23ed91ddfb1c5 . Wanted to open a proposal on it since its closely related to https://github.com/vmware/clarity/issues/5395

## Public API
```css
// load theme and follow same convention as dark theme in core
// by keeping it a external CSS file it allows teams to dynamically load themes as needed
@import '@cds/core/styles/theme.a11y.min.css';
```

```javascript
// enable theme based on user preferences
if (window.matchMedia('(prefers-reduced-motion: reduce)')).matches) {
  body.setAttribute('cds-theme', 'reduced-motion');
}
```

## Implementation

```typescript
// tokens.ts
motion: {
    duration: {
      0: token('0'),
      100: token('200ms'),
      200: token('300ms'),
      300: token('800ms'),
    },
  },
```

```scss
// theme.a11y.scss
[cds-theme~='reduced-motion'] {
  --cds-alias-motion-duration-100: #{$cds-alias-motion-duration-0};
  --cds-alias-motion-duration-200: #{$cds-alias-motion-duration-0};
  --cds-alias-motion-duration-300: 2s;
}
```

coryrylan:
Visibility @mathisscott

Issue #5395, opened 2020-12-17
core: adding clarity motion, specifically to modals


mathisscott:
Now that overlays and the modal refactor to use this is complete (#5200), we plan to introduce the motion library work via the modal.

Motion will be a decorator that can be applied to a component or possibly a property on a component. Users will be able to turn motion off via the API. Motion will be largely driven by CSS transitions and Animations.

PR #5657, opened 2021-02-25
feat: clarity motion

mathisscott:
• brought over animate decorator code
• built ClarityMotion service
• broke motion utilities into their own file
• wired up overlay to use the animate decorator
• introduced Animatable interface
• allows us to expect a component will have cdsMotion and motionChange
• add _animations internal property to store our animation configs
• verified that the decorator can access component properties
• verified that changing them in the decorator kicks off an update loop
• added AnimationStatus enum
• added modal-enter animation
• added way to retrieve animations from ClarityMotion service
• created animation run utility
• added error logging
• added target retrieval
• added ability to play any defined animation in reverse
• updated animation runner to deep clone animations so that the object reference remains untouched
• built out promise architecture to support sequential and concurrent animation runs
• fixed race condition with the CSS vs. animation
• added guard to prevent animation from running on host element with cds-motion 'off' or demo-mode set
• added 'ifOnly' conditional property to animation to make them easier to turn on/off via (other) host element properties
• checks host element properties/attributes for expected values
• only runs if all checks are true
• improved modal animation configs
• added custom event firing to start/end of animation run
• updated internal animation status attr at start/end of animation run
• pushed cds-motion 'off' into the duration code
• same end result but it integrates better with the web animation api
• set an onlyIf on the overlay backdrop animation so that it doesn't run on nested/layered overlays
• fixed issue with _demo-mode overlays appearing on top of dynamic overlays
• fixed bug in the nestedOverlay logic that wasn't catching when a focusTrapId had been removed from trapIds
• this was a problem that was highlighted by animations
• because overlays are removed from trapIds before or during animating out
• rem'd the overlay animation
• created enums and consts for common animation string values
• fixed typescript issue with possible null config property
• added animation-duration CSS custom prop to overlays
• added animation-easing CSS custom prop to overlays
• added animation tokens for duration and easing to design tokens
• tied animation tokens into overlay CSS properties
• used design tokens and CSS to toggle animation timing in the JS api
• updated modal-enter animation to use CSS custom property lookups
• added routine to lookup custom properties if passed a custom property name in animation config
• added getCssPropertyValue() utility
• added isCssPropertyName() utility
• added getNumericValueFromCssSecondsStyleValue() utility
• added getMillisecondsFromSeconds() utility
• updated other components to use getCssPropertyValue() utility
• re-organized motion files to more closely mirror i18n and icons
• made it so we export animation configs the same way we export icons
• this will allow us to keep our animation configs separate
• broke up the animation utility mono-function into a set of smaller utility functions that will be easier to test and maintain
• added `queryChildFromLightOrShadowDom()` utility to dom utils
• added `anyOrAllPropertiesPass()` utility to identity utils
• added `isPrefixedBy()` and `isSuffixedBy()` to string utils
• added `convertStringPropValuePairsToTuple()` to string utils
• general code cleanup
• unit tests for new utility functions
• unit tests for service
• unit tests for decorator
• unit test coverage over 90%
• fixed issue with animations keying off old property vals
• working on using animate decorator with accordions
• fixed issue with borders and expanded accordions
• added animated async accordion demo
• added static accordion demos to prove out the "off" scenario
• added hinge and shake animation configs to demonstrate how we might expand animations in the future
• added ability to size width/height in animation config based on another element in the host
• this is to help mostly with height-based animations
• not as canonical as using scaleY transforms but visually more coherent
• updated accordion panel open animation to use the new height option
• consolidated isPrefixedBy() and isSuffixedBy() into a simpler parent fn()
• added ability to remove suffixes and prefixes from strings with a family of Fns
• updated accordion unit tests post cds-motion inclusion
• added unit tests for the new 'from:' sizing functionality
• added unit tests for new prefix/suffix utility Fns
• increasing bundlesize to account for new animation tokens
• reorganized motion exports so they now come from one index file
• updating all existing transitions/animations to use Clarity Motion design tokens
• fixing inconsistencies with component API dox and supported CSS custom properties
• removed unnecessary layout wrapper from modal
• added entrance animations to modal
• made transition speed for circular progress configurable via CSS custom property
• this will help with low motion theming and circular progress
• broke up keyframes for reuse
• added basic support for inline overrides
• added story to test/demo overrides
• updated animation tokens to use UX recommended labels
• refactored all animated components to use new animation tokens
• added docs for clarity motion
• added comments to clarity animation utils and services
• added low-motion theme for a11y
• added stories to demo low-motion theme in use
• per feedback, docs broken up into public vs. internal apis
• build fixes

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #5395 
Issue Number: #5419

- [ ] Yes
- [x] No


packages/core/build/tokens.ts
```diff
@@ -417,6 +417,25 @@ const typography = {
...
+    secondary: token('0.3s'),
+    primary: token('0.4s'),
+    slow: token('0.5s'),
+    slower: token('0.7s'),
+    slowest: token('0.8s'),
+  },
+  easing: {
+    primary: token('cubic-bezier(0,.99,0,.99)'),
+    secondary: token('cubic-bezier(0, 1.5, 0.5, 1)'),
+    loop: token('cubic-bezier(0.17, 0.4, 0.8, 0.79)'),
```
mathisscott, line 435: Names for animation tokens came from @colinreedmiller

packages/core/import-map.importmap
```diff
@@ -8,6 +8,7 @@
     "lit-html/lib/shady-render.js": "/base/node_modules/lit-html/lib/shady-render.js",
     "lit-html/directives/if-defined.js": "/base/node_modules/lit-html/directives/if-defined.js",
     "ramda/es/anyPass.js": "/base/node_modules/ramda/es/anyPass.js",
+    "ramda/es/clone.js": "/base/node_modules/ramda/es/clone.js",
```
mathisscott, line 11: We use clone to make copies of animation objects so we don't pass around references and unintentionally mutate animation configs.

packages/core/src/accordion/accordion-panel.element.scss
```diff
@@ -8,32 +8,52 @@
...
   padding: 0;
   width: 100%;
 }
 
 .accordion-content {
-  height: 0;
-  transition: transform height var(--transition-duration);
-  transition-timing-function: cubic-bezier(0, 0.99, 0, 0.99);
   overflow-y: hidden;
+  box-shadow: 0 calc(-1 * var(--border-width)) 0 0 var(--border-color);
```
mathisscott, line 24: Had to use a box shadow for the content border because border was causing the animation to nudge out at the very end.

packages/core/src/accordion/register.ts
```diff
@@ -18,7 +23,7 @@ registerElementSafely('cds-accordion-content', CdsAccordionContent);
 registerElementSafely('cds-accordion-header', CdsAccordionHeader);
 
 ClarityIcons.addIcons(angleIcon);
-
+ClarityMotion.add(AnimationAccordionPanelOpenName, AnimationAccordionPanelOpenConfig);
```
mathisscott, line 26: Animations get added in a similar fashion to how icons are added...

packages/core/src/forms/form-group/form-group.element.spec.ts
```diff
@@ -96,7 +97,7 @@ describe('cds-form-group', () => {
     await componentIsStable(formGroup); // firstUpdated
 
     expect(controls[0].querySelector('label').getBoundingClientRect().width).toBe(200);
-    expect(getComputedStyle(formGroup).getPropertyValue('--internal-label-min-width')).toBe('10rem');
+    expect(getCssPropertyValue('--internal-label-min-width', formGroup)).toBe('10rem');
```
mathisscott, line 100: I added a `getCssPropertyValue()` utility function

packages/core/src/input/input.element.scss
```diff
@@ -17,7 +17,7 @@
   --font-size: #{$cds-global-typography-font-size-3};
   --color: #{$cds-global-typography-color-500};
   --line-height: #{$cds-global-space-9};
-  --transition: background-size 0.2s ease;
+  --transition: background-size #{$cds-global-animation-duration-quick} ease;
```
mathisscott, line 20: All pre-existing animations now use the animation tokens from Nick and Colin

packages/core/src/internal-components/overlay/overlay.element.scss
```diff
@@ -36,10 +38,21 @@
 
 :host([_demo-mode]) {
   position: absolute;
+  z-index: 100000;
```
mathisscott, line 41: This fixes an issue where overlays in the demo blocks were showing up above the interactive overlays.

packages/core/src/internal-components/overlay/overlay.element.spec.ts
```diff
@@ -15,6 +15,14 @@ describe('Overlay helper functions: ', () => {
       expect(isNestedOverlay('ohai_2', 'ohai_', ['ohai_1', 'ohai_2', 'ohai_3'])).toBe(true, 'middle one checks');
     });
 
+    // this situation happens during an update loop when the focus trap list is updated just
+    // prior to the rest of the component and the component still needs to think of itself as
+    // "nested" or "layered". a good example is running an exit animation asynchronously
+    // outside of the update loop.
```
mathisscott, line 21: This comment explains a need for a re-factoring of the focus traps. Animations revealed a bad state they could get into.

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -14,12 +16,26 @@ import {
...
+  myId: string,
+  overlayPrefix: string,
+  trapIds: string[],
+  previousValue?: boolean
+): boolean {
   const overlayIds = trapIds.filter(id => id.indexOf(overlayPrefix) > -1);
+
+  if (previousValue === true && trapIds.indexOf(myId) < 0) {
+    // handling situation where focusTrapIds remove our overlay from the list and we still need to consider
+    // this overlay as nested. this happens when an overlay is being closed/hidden
```
mathisscott, line 35: This is the bad state that animations revealed. It could've happened with any a synchronous update of a focus trap. Animations just revealed it first.

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -56,8 +72,25 @@ type CloseChangeSources = 'backdrop-click' | 'escape-keypress' | 'close-button-c
  * @event closeChange - Notify user when close event has occurred
  * @cssprop --backdrop-background
  * @cssprop --layered-backdrop-background
+ * @cssprop --animation-duration
+ * @cssprop --animation-easing
+ *
+ * KNOWN ISSUE: Safari jumps through the exit animation but only when the ESC key is pressed.
```
mathisscott, line 78: We have a known issue where safari jumps to the end of an exit animation but only when the escape key is pressed.

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -56,8 +72,25 @@ type CloseChangeSources = 'backdrop-click' | 'escape-keypress' | 'close-button-c
...
+ *
  */
-export class CdsInternalOverlay extends CdsBaseFocusTrap {
+@animate({
+  hidden: {
+    true: reverseAnimation(AnimationModalEnterName),
+    false: AnimationModalEnterName,
+  },
+})
```
mathisscott, line 86: This is how you animate a component! ✨

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -87,8 +120,12 @@ export class CdsInternalOverlay extends CdsBaseFocusTrap {
...
     super.updated(props);
     const oldLayered = this.isLayered;
-    const isNested = isNestedOverlay(this.focusTrapId, this.overlayIdPrefix, FocusTrapTracker.getTrapIds());
-    const newLayered = this.focusTrap.active && isNested;
+    const newLayered = isNestedOverlay(
+      this.focusTrapId,
+      this.overlayIdPrefix,
+      FocusTrapTracker.getTrapIds(),
+      oldLayered
+    );
```
mathisscott, line 128: This is the code impact of the focus trap refactor. If we make the focus trap a headless component in the future it would be helpful to migrate this into that headless component. At present, we only have one family of components that require the focus trap...

packages/core/src/internal/interfaces/index.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
```
mathisscott, line 5: Originally I had the motion interfaces in this file. I migrated them into their own directory inside of cds internal

packages/core/src/internal/motion/motion.stories.ts
```diff
@@ -0,0 +1,20 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import '@cds/core/button/register.js';
+import '@cds/core/internal-components/overlay/register.js';
+import '@cds/core/accordion/register.js';
+// import { ClarityMotion } from '@cds/core/internal/motion/index.js';
+// import { html } from 'lit-html';
+// import customElements from '../../dist/core/custom-elements.json';
```
mathisscott: This file may not be needed. Do you think I should remove it, @coryrylan  ?

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+/**
+ * runPropertyAnimations() is a utility function called by the @animate(). It is a single point of entry
+ * to execute the property-driven animations found in components like the accordion and overlay.
+ *
+ * Other utilities here are exported to break up the unit tests covering the animation/motion
+ * library. While they could be called or used in isolation, it is not recommended for most use cases.
+ *
+ * @internal
+ */
+export async function runPropertyAnimations(props: Map<string, any>, hostEl: AnimatableElement) {
```
mathisscott: This function drives a family of utilities that does all the work for the animations. It amounts to a bunch of transformations on the animation config objects and finals down to a list of promises.

You can follow the comments for a pretty good guide as to what state the transformation is in at each major milestones.

packages/core/src/internal/utils/css.ts
```diff
@@ -39,9 +41,24 @@ export function updateElementStyles(el: HTMLElement, ...styleTuples: [string, st
...
+    return getComputedStyle(el, pseudoSelectorModifier).getPropertyValue(propertyName).trim();
+  } catch (e) {
+    LogService.warn('Container element passed to getCustomPropertyValue must be an element.');
+    return '';
+  }
+}
+
+export function isCssPropertyName(str: any): boolean {
+  return !!str && isString(str) && str.slice(0, 2) === '--';
+}
```
mathisscott, line 64: Two new utilities added, both centered on CSS properties.

packages/core/src/internal/utils/dom.ts
```diff
@@ -158,3 +158,7 @@ export function spanWrapper(nodeList: NodeListOf<ChildNode>): void {
       spanWrapper.appendChild(node);
     });
 }
+
+export function queryChildFromLightOrShadowDom(selector: string, hostEl: Element): Element | null {
+  return hostEl.querySelector(selector) || hostEl?.shadowRoot?.querySelector(selector) || null;
+}
```
mathisscott, line 168: This utility is extremely useful in our components.

packages/core/src/internal/utils/identity.ts
```diff
@@ -81,3 +82,43 @@ export function deepClone(obj: any) {
...
+  }
+}
+
+export function allPropertiesPass(obj: any, propValuePairs: string): boolean {
+  return anyOrAllPropertiesPass(obj, propValuePairs, 'all');
+}
+
+export function anyPropertiesPass(obj: any, propValuePairs: string): boolean {
+  return anyOrAllPropertiesPass(obj, propValuePairs, 'any');
+}
```
mathisscott, line 120: These utility functions pave the way for us to get more mileage out of the property value keys I'm using in the web animations api extensions. We could see more utility out of the `propertyName:testValue` style of checks in our code.

packages/core/src/internal/utils/string.ts
```diff
@@ -66,3 +68,103 @@ export const cssGroup = 'CSS Custom Properties';
...
+  }
+}
+
+export function removePrefix(str: string, prefix: string): string {
+  return removePrefixOrSuffix(str, prefix, 'prefix');
+}
+
+export function removeSuffix(str: string, suffix: string): string {
+  return removePrefixOrSuffix(str, suffix, 'suffix');
+}
```
mathisscott, line 123: There are a lot of utilities here for managing, testing for, and removing prefixes and suffixes. Raising awareness that they are here in case they are needed in the future.

packages/core/src/internal/utils/string.ts
```diff
@@ -66,3 +68,103 @@ export const cssGroup = 'CSS Custom Properties';
...
+  | [string, null]
+  | [string, undefined]
+  | [string, number];
+
+// this utility is a little restrictive on its inputs. it expects a specific format.
+// going outside of that format (like passing an object or function as the value) can yield unintended results.
+// this does NOT eval anything. that would be bad.
+// 'isValid:true status:success': string =>
+// [['isValide', true], ['status', 'success']]: [string, string | number | boolean][]
+export function convertStringPropValuePairsToTuple(propValString: string): ObjectPropertyNameAndValueTuples[] {
```
mathisscott, line 138: This is useful for the `propertyName:testValue` formats. I don't think it would be used for anything beyond that. But the array of tuples makes it easy to loop through them with a map or something else.

packages/core/src/modal/modal.element.ts
```diff
@@ -46,7 +56,13 @@ import { styles } from './modal.element.css.js';
  * @cssprop --box-shadow
  * @cssprop --width
  */
-export class CdsModal extends CdsInternalOverlay {
+@animate({
+  hidden: {
+    true: reverseAnimation(AnimationModalEnterName),
+    false: AnimationModalEnterName,
+  },
+})
```
mathisscott, line 64: Animations are not inherited because decorators are applied when a new class instance is created. This is probably the best argument for making the decorator a headless component.

packages/core/src/modal/modal.element.ts
```diff
@@ -88,28 +104,26 @@ export class CdsModal extends CdsInternalOverlay {
...
-          <div class="modal-content" cds-layout="vertical gap:md gap@md:lg align:stretch">
-            <div cds-layout="horizontal gap:md wrap:none align:vertical-center p-x:lg p-t:lg">
-              <div>
-                <slot name="modal-header"></slot>
-              </div>
-              <div cds-layout="align:right">
-                <slot name="modal-header-actions"></slot>
-              </div>
-              <slot name="close-button"></slot>
+      <div class="modal-dialog private-host" tabindex="-1" cds-layout="m:md m@md:xl">
```
mathisscott, line 107: We had an unnecessary wrapper around the modal. It's possible it had utility in a previous iteration. But I have verified that it was vestigial after the overlay re-factor.

packages/core/src/styles/theme.low-motion.scss
```diff
@@ -0,0 +1,17 @@
...
+[cds-theme~='low-motion'] {
+  --cds-global-animation-duration-instant: 0;
+  --cds-global-animation-duration-quickest: 0;
+  --cds-global-animation-duration-quicker: 0;
+  --cds-global-animation-duration-quick: 0;
+  --cds-global-animation-duration-secondary: 0;
+  --cds-global-animation-duration-primary: 0;
+  --cds-global-animation-duration-slow: 2s;
+  --cds-global-animation-duration-slower: 2s;
+  --cds-global-animation-duration-slowest: 2s;
```
mathisscott, line 16: We will need to verify these low-motion speeds with @scroniser. IIRC, they come from Sheri and Chris.

packages/core/src/internal-components/overlay/overlay.element.scss
hippee-lee, line 41: In `@clr/angular` we used a map ($clr-layers) to organize and clarify different uses of z-index accross components. Does core need (or already have) anything like that?

packages/core/src/internal/motion/utils.ts
hippee-lee: Spec questions. In the `describe('runPropertyAnimations()', () => {...});` block there is only one test for the log warning. With my Angular diven model for testing it seems like we should also have tests for it:
  
- runsPropAnimationsOnAnimatableElement (generic / basic it just works sanity check) or an checks for multiple animations b/c of the for loop on line num 46
- doesNotRunWhenUpdateSentButPropertyNotChanged - this guard is called out as an unlikely state ... should we know if something ever triggers it?
- adds / remove and correctly manages the super secret `_cds-animation-status` class
- emits the correct motionChange string 
- detectsReverseAnimations
- detectsWhenCdsMotionTurnedOff
- detectsNoAnimationRoutines
- createsArrayOfPromises
- handlesOnlyIfPropValue properly `onlyIf prop:value`

packages/core/src/internal/motion/motion.stories.mdx
```diff
@@ -0,0 +1,468 @@
...
+ * @animate({
+ *   // the property name goes here
+ *   propertyName: {
+ *     // any value can be used as an object key; the value associated with it
+ *     // must be an animation loaded into ClarityMotion. If not, then no
+ *     // animation will run.
+ *     "property-value": "name-of-animation-config-loaded-into-clarity-motion"
+ *   },
+ *   // multiple properties can be defined
+ *   {
```
hippee-lee: should this example have a `propertyName: {...}` for the booleans?

packages/core/src/internal/motion/motion.stories.mdx
```diff
@@ -0,0 +1,468 @@
...
+
+- Overriding an animation value in the Clarity Motion registry
+- Assigning animations to a component by subclassing an existing component
+- Overriding a component animation with an inline override
+- Customizing the timing or easing of an animation with CSS custom properties
+- Customizing the timing or easing of all animations with a theme
+
+### Override Animation Definition in Clarity Motion
+
+An animation in the Clarity Motion registry can be overridden anywhere in your application where are you could add an icon to Clarity Core or add a new animation to Clarity Motion.
```
hippee-lee: Typo? "where `are` you could add"

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -40,11 +49,22 @@ import { styles } from './accordion-panel.element.css.js';
...
+    true: AnimationAccordionPanelOpenName,
+    false: reverseAnimation(AnimationAccordionPanelOpenName),
+  },
+})
+export class CdsAccordionPanel extends LitElement implements Animatable {
+  @property({ type: String })
+  cdsMotion = 'on';
+
+  @event()
+  motionChange: EventEmitter<string>;
```
coryrylan: Since motion is a common API across components and we prefix it like `cds-layout`, `cds-theme` and `cds-i18n` seems like the event should correspond as well `cdsMotion` and `cdsMotionChange`.

packages/core/src/internal/motion/motion.service.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  describe('add()', () => {
+    // add basic functionality tested in get()
+    it('early returns if passed a falsy animation name', () => {
+      spyOn(ClarityMotion, 'add');
+      ClarityMotion.add(null, setConfig);
+      ClarityMotion.add(void 0, setConfig);
+      ClarityMotion.add('', setConfig);
+      expect(ClarityMotion.add).not.toThrowError();
```
hippee-lee, line 54: Is testing that it doesn't throw an error the same as an early return? 
Could a check for a return of `undefined` also be used here?
Same comment for the test below.

```
const undf =  ClarityMotion.add(null, setConfig);
expect(undf).toBeUndefined();
```

packages/core/src/internal/motion/motion.service.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { MotionRegistry, TargetedAnimation } from './interfaces.js';
+
+const registry: MotionRegistry = {};
+
+const motionRegistry = new Proxy(registry, {
```
hippee-lee, line 11: Cool. I didn't know about Proxy objects.

packages/core/.storybook/preview.js
```diff
@@ -27,16 +27,7 @@ export const parameters = {
...
-          'Design Tokens',
-          'Typography',
-          'Spacing',
-          'Color',
-          'Object Styles',
-          'Interaction Styles',
-          'Status',
-          'Platforms',
-        ],
+        ['Color', 'Object Styles', 'Interaction Styles', 'Status', 'Platforms', 'Motion'],
```
coryrylan: It looks like the first three paths were accidentally removed

```
'Design Tokens',	
'Typography',	
'Spacing',
```

packages/core/src/internal/motion/motion.stories.ts
coryrylan: Yeah if theres no motion specific stories then it can be removed. It might be breaking the other stories as well since it has a title of `title: 'Stories/Typography'`

packages/core/src/internal/decorators/animate.ts
```diff
@@ -0,0 +1,37 @@
...
+
+      updated(props: Map<string, any>) {
+        super.updated(props);
+
+        const self = (this as unknown) as Element & Animatable & { _animations: PropertyDrivenAnimation };
+        self._animations = config || void 0;
+
+        if (!self.hasAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME)) {
+          self.setAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME, AnimationStatus.ready);
+        } else if (!self.hasAttribute('_demo-mode')) {
```
hippee-lee, line 30: Is it worth adding a test to make sure this doesn't run for _demo-mode?

hippee-lee:
This is my first pass, I still need to read through the changes to existing components.

packages/core/src/internal/decorators/animate.ts
mathisscott: IMO, not really. The mocking involved would yield low benefits. Karma struggles with spying on functions, so you'd have to mock it all the way down into the supporting functions. Per code-coverage/istanbul, that branch already has `23x` coverage. So I'm confident it's handled elsewhere.

packages/core/src/internal/motion/motion.service.ts
mathisscott: Yeah. I wasn't too familiar with them either. @coryrylan introduced them to the library. I like to learn new things!

packages/core/src/internal/motion/motion.service.spec.ts
mathisscott: The problem there is it returns undefined either way. So testing to see if it returns undefined would pass even if it threw an error.

packages/core/src/internal/motion/motion.stories.mdx
mathisscott: Good catch. I will fix that.

packages/core/src/internal/motion/motion.stories.mdx
mathisscott: Yes it should. Another good catch. Thanks!

packages/core/src/internal-components/overlay/overlay.element.scss
mathisscott: Right now, we don't need it. We don't have that many Z indexes. When we start building things like toasts, notifications, tooltips, and the header, that's when more Z indexes would come into play. They would likely live in design token land. But right now, adding them would just create one more thing to document.

packages/core/src/internal/motion/utils.ts
mathisscott: Right now, the utils have really good code coverage. One of the benefits of a more functional approach is the ability to test things from the inside out. You can test the functions that drive meta-functions like `runPropertyAnimations()` in isolation. This wasn't possible in the angular library because we took an OOP approach, meaning if we had utility functions they were often private or unavailable outside of the angular component.

This meant we often had to test a component's functionality from the outside in – we had to test the results, as opposed to the pieces that did the actual work.

Another benefit of this approach is that we will have a clearer picture of what is failing if a test breaks. Testing from the outside in, you may see a test fail but have no idea what the cause of that failure might be.

packages/core/src/internal/motion/motion.stories.ts
mathisscott: Makes sense. Will do.

packages/core/.storybook/preview.js
mathisscott: Yeah, this was weird because I had those in my branch but, when I re-based, the re-base wanted to remove them. I went digging and figured that the removal was intentional. Happy to add them back in.

packages/core/src/accordion/accordion-panel.element.ts
mathisscott: Yeah, that makes sense. I can do that.

packages/core/src/internal/motion/utils.ts
hippee-lee: Sounds good. fwiw I was focused on the control flow logic of `runPropertyAnimations` and the things it was doing, not the calls to the delegated functions. e.g if line 86 was modified by accident were removed by accident and the call on line happened incorrectly or didn't happen when it needs to how would we know?

packages/core/src/internal/motion/motion.service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  static has(name: string): boolean {
+    return !!name && !!motionRegistry[name];
+  }
+
+  static get(name: string): TargetedAnimation[] {
+    return motionRegistry[name] || [];
+  }
+
+  static add(animationName: string, animationConfig: TargetedAnimation[]) {
```
hippee-lee, line 41: ClarityIcons uses addIcon, since these (has & get) methods are similar should they be more explicit? e.g  `addAnimation`

hippee-lee (approved):
One last comment about naming. 
Everything lgtm.

packages/core/src/internal/motion/motion.service.ts
mathisscott: `addIcon` is only `addIcon` because 3.0's API was backwards compatible with `@clr/icon`.

In the future, `addIcon` will become `add` and be deprecated. We just aren't in a hurry to do that and will likely keep it hanging around as an alias for `ClarityIcons.add` for the long term.

packages/core/src/internal/motion/utils.ts
mathisscott: oh i see. that is a valid concern. the limiting factor is how poorly karma allows us to spy on just plain functions. methods on objects? great! a standalone function coming from a module? 😱 

the easiest way to unit test that control flow is to either... break it out into its own function(s) that returns our promises and animations so we can inspect them. or build all of this into some kind of static class so we can spy on the methods. could/should the utilities be part of a general service? either `ClarityMotion` or another service to keep `ClarityMotion` light?

these are good questions i don't have good answers to. thoughts @hippee-lee and @coryrylan ?

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+            resolve('animation finished');
+            animationPlayer.removeEventListener('finish', listener);
+          };
+
+          animationPlayer.addEventListener('finish', listener);
+        });
+      });
+
+      // wait for all the concurrent property animations to finish
+      await Promise.all(animations);
```
coryrylan: (optional) This could be extracted into a function

```typescript
const animations = getAnimationPromiseInstructions(hostEl, animationConfig);
await Promise.all(animations);
```

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+        if (config.onlyIf && !allPropertiesPass(hostEl, config.onlyIf)) {
+          return true;
+        }
+
+        // accesses and manipulates the private animation status attr;
+        if (hostEl.getAttribute('_cds-animation-status') !== AnimationStatus.active) {
+          hostEl.setAttribute('_cds-animation-status', AnimationStatus.active);
+          // this is here so we only emit the start event once, right when the whole animation starts...
+          hostEl.motionChange.emit(`${animatedPropertyValueAsString} animation ${AnimationStatus.start}`);
+        }
```
coryrylan: (optional) This could be extracted into a function

```typescript
setAnimationStartStatus(hostEl);
```

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+      }
+
+      if (cdsMotionValue && cdsMotionValue === 'off') {
+        // cdsMotion is turned off, so zero out the duration
+        motionForMyValue = zeroOutAnimationConfig(motionForMyValue);
+      } else {
+        // sets duration and easing based on CSS properties, passed values, or a global fallback
+        motionForMyValue = setAnimationDuration(motionForMyValue, hostEl);
+        motionForMyValue = setAnimationEasing(motionForMyValue, hostEl);
+      }
```
coryrylan: (optional) This could be extracted into a function

```typescript
motionForMyValue = setAnimationConfigOptions(motionName, motionForMyValue, hostEl);
```

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+
+      // loops through animation config and runs them; assigns the promises to an array we can Promise.all() at the end
+      const animations = motionForMyValue.map(config => {
+        // does not allow animation to run if the onlyIf prop:value string doesn't let us
+        if (config.onlyIf && !allPropertiesPass(hostEl, config.onlyIf)) {
+          return true;
+        }
```
coryrylan: Could this be a filter rather than an early return? With the early return would `animations` have multiple types in the array, promises and `true`?


```typescript
const animations = motionForMyValue
  .filter(!config.onlyIf && allPropertiesPass(hostEl, config.onlyIf)
  .map(config => {
     ...
  });
```

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+      });
+
+      // wait for all the concurrent property animations to finish
+      await Promise.all(animations);
+
+      // sets super secret animation attr back to 'ready'
+      hostEl.setAttribute('_cds-animation-status', AnimationStatus.ready); // A
+
+      // emits the name of the animation and that it ended
+      hostEl.motionChange.emit(`${animatedPropertyValueAsString} animation ${AnimationStatus.end}`); // A
```
coryrylan: (optional) This could be extracted to a function

```typescript
setAnimationEndStatus(animatedPropertyValueAsString, hostEl);
```

packages/core/src/internal/motion/utils.ts
```diff
@@ -0,0 +1,372 @@
...
+  config: TargetedAnimation[],
+  staticFallback: string | number,
+  valueConverter?: (arg: string) => string | number
+) {
+  return config.map((anim: TargetedAnimation) => {
+    if (anim.options) {
+      if (!(anim.options as { [key: string]: string | number })[propertyName]) {
+        (anim.options as { [key: string]: string | number })[propertyName] = staticFallback;
+      } else if (isCssPropertyName((anim.options as { [key: string]: any })[propertyName])) {
+        const myConfigOption = (anim.options as { [key: string]: string | number })[propertyName];
```
coryrylan: Could we make a new type for the inline type expression?

```typescript
{ [key: string]: string | number }

type AnimationOptions = { [key: string]: string | number };
```

packages/core/src/internal/motion/utils.ts
hippee-lee: What about leaving a comment for future test enhancements? 

I didn't block because its working for me and I understand it. Adding tests will provide a verification check that it always does what I think it does but I'm not sure how hard that is at this point. Since you are comfortable I didn't want to block on my hang up but It is worth calling out that there could be a few places in the function logic that are hard to track down b/c they aren't tested , yet.

imo - not a blocker but worth leaving a note or a TODO for future devs.

coryrylan: I would lean towards smaller functions over a class that has to be spied on. Because the test coverage is high I wont block on this either. I added some comments of possible ideas of ways to break the function up into smaller functions. Each comment block looked like a possible opportunity for extracting to a function. Even if they are not exported for tests it might make it easier to read/follow.

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -40,11 +49,22 @@ import { styles } from './accordion-panel.element.css.js';
  * @event expandedChange - notify when the user has clicked the panel header
  * @cssprop --border-color
  * @cssprop --border-width
- * @cssprop --color
- * @cssprop --font-size
- * @cssprop --font-weight
+ * @cssprop --animation-duratino
```
jeeyun: typo: should be `--animation-duration` :)

packages/core/src/internal-components/overlay/overlay.stories.ts
```diff
@@ -87,6 +90,9 @@ export const interactive = () => {
       myOverlay.addEventListener('closeChange', () => {
         myOverlay.setAttribute('hidden', 'true');
       });
+      myOverlay.addEventListener('motionChange', (e: any) => {
+        console.log(e.detail);
```
jeeyun, line 94: I assume the `console.log` here is for testing? Just checking that you weren't intending to do something additional in this event listener.

packages/core/src/internal-components/overlay/overlay.stories.ts
```diff
@@ -87,6 +90,9 @@ export const interactive = () => {
       myOverlay.addEventListener('closeChange', () => {
         myOverlay.setAttribute('hidden', 'true');
```
jeeyun: for consistency, should we set attribute value to empty quotes here as well?

jeeyun: The stories below all follow this pattern so I won't add additional comments. I figured they are all going to stay the same or be updated together. :)

packages/core/src/internal-components/overlay/overlay.stories.ts
mathisscott: Yeah. That's just for demo purposes.

packages/core/src/internal-components/overlay/overlay.stories.ts
mathisscott: maybe? i'm not sure. i'd have to test it against the property-driven animation to make sure that worked as expected. i think it would but 🤷‍♂️

mathisscott: it works! change made. thanks!

coryrylan (approved):

PR #5666, opened 2021-02-26
feat(modal): reduce change detection cycles for the modal body

arturovt:
This PR reduces change detection cycles for the modal body by replacing
`HostListener` with `Renderer.listen` outside of the zone. Updating `tabindex`
on the host element can be done through the DOM API and that would not require
Angular to run `ApplicationRef.tick()`. Note that the `HostListener` wraps the
actual listener under the hood into the internal Angular function which runs

## PR Checklist

## PR Type
- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/modal/modal-body.ts
+++ b/packages/angular/projects/clr-angular/src/modal/modal-body.ts
@@ -1,10 +1,10 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Directive, HostListener } from '@angular/core';
+import { Directive, ElementRef, NgZone, OnDestroy, Renderer2 } from '@angular/core';
 
 /**
  * Allows modal overflow area to be scrollable via keyboard.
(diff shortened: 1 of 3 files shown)


bbogdanov (approved):

arturovt:
Hey @bbogdanov . I've run `build:ci` and this fails locally:

image

But this PR doesn't have changes to datepicker, not sure how to tackle this.

UPD: it also fails on the `next` branch.

bbogdanov:
@arturovt seems like you need to update also the `golden` file

```
cd packages/angular
yarn golden:fix
```

bbogdanov (approved):

gnomeontherun:
@bdryanovski can you take a look at this too?

packages/angular/projects/clr-angular/src/modal/modal-body.ts
```diff
@@ -1,36 +1,43 @@
...
-  selector: '.modal-body',
-  host: {
-    '[attr.tabindex]': 'tabindex',
-  },
-})
-export class ClrModalBody {
-  tabindex = 0;
+@Directive({ selector: '.modal-body' })
+export class ClrModalBody implements OnDestroy {
+  private tabindex = '0';
```
Jinnie, line 17: nit: not sure we need this variable. It's never redefined, nor we save it before being removed. I'd just hardcode it to 0 in the listener.

Jinnie (approved):
Looks great to me!
I love the way the API gets simpler and we move the lifecycle irrelevant events outside the Angular zone.
I'm fine with even not changing the "nit" note below and merging it as is.
Thanks for the contribution

PR #5673, opened 2021-03-01
feat(datagrid): reduce change detection cycles for the `clrOutsideClick`

arturovt:
This PR reduces change detection cycles for the `clrOutsideClick` by replacing
`HostListener` with `Renderer.listen` outside of the zone. Basically, it will not
run change detection when the click event happened on the host element, since the
listener will just `return`, it will run `ApplicationRef.tick()` only if the click
event happens outside of the host element. Note that the `HostListener` wraps the
actual listener under the hood into the internal Angular function which runs

## PR Checklist

## PR Type
- [x] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/utils/outside-click/outside-click.spec.ts
+++ b/packages/angular/projects/clr-angular/src/utils/outside-click/outside-click.spec.ts
@@ -1,52 +1,69 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component } from '@angular/core';
-import { TestBed } from '@angular/core/testing';
+import { ApplicationRef, Component } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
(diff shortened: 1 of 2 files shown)


gnomeontherun:
@bdryanovski can you take an initial look this week?

bdryanovski (approved):
LGTM 

But you gonna need to re-run the Github Actions so the build task will pass. 

Better will be the merge with `vmware:next` before doing that - we have a fix for this type of failing actions.

mathisscott (approved):

gnomeontherun:
@Shijir can you review why this is failing and see if we can close this out?

Shijir (commented):
HI @arturovt, could you run `yarn run format:file` command to fix some formatting issues. After that, I believe, the builds will pass successfully.

bdryanovski:
@Shijir @gnomeontherun @arturovt 

The reason is not unformatted files.

The reason is failing tests. This is something that we addressed here #5773 (I think this is the fix)

To solve this there are two options: 

  * merge/rebase with `next` branch and get the fix from there
  * re-run `build` Github Action until it passes - this could take a few tries

bdryanovski:
@arturovt can please re-run the `Build / build (pull_request)` so we could close this PR

hippee-lee:
I'm closing this in favor of this PR which adopts these changes but allows me to run the CI/CD and fix any formatting issues that might exist. 

The new PR is: https://github.com/vmware/clarity/pull/6035

PR #5557, opened 2021-02-03
fix: assign correct described-by id to form input and control helpers

bdryanovski:
* provide another service for creating container/control ID
 * set correct role to elements
 * set describe by id and link it to control helper/error/success

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] clarity.design website / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

## Other information

Fixes #5429 
Fixes #4650

PR #5678, opened 2021-03-02
Migrate Datepicker to extend ClrAbstractContainer as all other containers

bdryanovski:
Datepicker container is the only one that is not extending ClrAbstractContainer - this creates the need for code duplications. With this change, this duplication is no longer needed and now we could have a more compact container code.

NOTE: This PR extends the work here #5557 so when merged then this could be changed from Draft to PR.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -859,32 +859,21 @@ export declare class ClrDatalistModule {
 export declare class ClrDataModule {
 }
 
-export declare class ClrDateContainer implements DynamicWrapper, OnDestroy, AfterViewInit, AfterContentInit {
-    _dynamic: boolean;
+export declare class ClrDateContainer extends ClrAbstractContainer implements AfterViewInit {
     set actionButton(button: ElementRef);
     set clrPosition(position: string);
     commonStrings: ClrCommonStringsService;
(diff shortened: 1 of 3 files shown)


gnomeontherun:
Why was this done? Was there a bug?

bdryanovski:
The change is really small is primary here `packages/angular/projects/clr-angular/src/forms/datepicker/date-container.ts` - but because is on top of #5557 that is not already merged it looks huge. 

The issue is that the `DateContainer` is requiring to duplicate a lot of the code that we have for other containers - there is no real reason for us to have it separate. This change is making `DateContainer` part of the Input Containers Family so all logic and helper functions that we have there are accessible to it. Before this, every change had to be made twice and tested twice. 

One of the big changes here will be that any fix on how and when Success, Error, or Helper component work and behave will be done in one place. Something that is not true at the moment.

After merging #5557 will change from Draft to PR and the changes will be more clear. I hope to merge it by the end of the week.

Jinnie (approved):
I love it when we remove code!

sis0k0 (approved):

PR #5690, opened 2021-03-05
fix: core icon render performance

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When rendering icons it can cause unnecessary style calculations for size causing performance slow down.

Before:
before

After:
after

## What is the new behavior?
This PR prevents unnecessary style computation when not specifying numeric size. Also will cache font size calculation in the html attr after first check. This improves average render time by 3x

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/icon/utils/icon.classnames.ts
+++ b/packages/core/src/icon/utils/icon.classnames.ts
@@ -33,10 +33,11 @@ export function getUpdateSizeStrategy(size: string) {
 
 export function updateIconSizeStyle(el: CdsIcon, size: string) {
   const updateStrategy = getUpdateSizeStrategy(size);
-  const val = pxToRem(parseInt(size));
+  let val = '';
 
   switch (updateStrategy) {
     case SizeUpdateStrategies.ValidNumericString:
+      val = pxToRem(parseInt(size)); // set val in block to run expensive call only when needed
       updateElementStyles(el, ['width', val], ['height', val]);
       return;
     case SizeUpdateStrategies.ValidSizeString:
--- a/packages/core/src/internal/utils/css.spec.ts
+++ b/packages/core/src/internal/utils/css.spec.ts
@@ -98,6 +98,12 @@ describe('Css utility functions - ', () => {
     it('should convert px to rem values from base font size token', () => {
       expect(pxToRem(10)).toBe('0.5rem');
     });
+
+    it('should cache the rem calculation to the HTML cds-font-size attr', () => {
+      document.documentElement.setAttribute('cds-base-font', null);
+      expect(pxToRem(20)).toBe('1rem');
+      expect(document.documentElement.getAttribute('cds-base-font')).toBe('20');
+    });
   });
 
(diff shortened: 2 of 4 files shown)


packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -33,10 +33,9 @@ export function getUpdateSizeStrategy(size: string) {
 
 export function updateIconSizeStyle(el: CdsIcon, size: string) {
   const updateStrategy = getUpdateSizeStrategy(size);
-  const val = pxToRem(parseInt(size));
-
   switch (updateStrategy) {
     case SizeUpdateStrategies.ValidNumericString:
+      const val = pxToRem(parseInt(size));
```
mathisscott: Might need to move `val` into a `let` on line 36 and then assign it to `val` here.

packages/core/src/internal/utils/css.spec.ts
```diff
@@ -96,5 +96,11 @@ describe('Css utility functions - ', () => {
     it('should convert px to rem values from base font size token', () => {
       expect(pxToRem(10)).toBe('0.5rem');
     });
+
+    it('should cache the rem calculation to the HTML cds-font-size attr', () => {
```
mathisscott, line 102: 👍

packages/core/src/internal/utils/css.ts
```diff
@@ -39,9 +39,15 @@ export function updateElementStyles(el: HTMLElement, ...styleTuples: [string, st
 }
 
 export function pxToRem(pxValue: number) {
-  const baseProp = window
-    .getComputedStyle(document.body, null)
-    .getPropertyValue('--cds-global-typography-base-font-size');
-  const baseFontSize = (16 * parseInt(baseProp !== '' ? baseProp : '100%')) / 100;
-  return `${pxValue / baseFontSize}rem`;
+  let base = parseInt(document.documentElement.getAttribute('cds-base-font') as any);
```
mathisscott: Would this attribute really comeback as `any`? Seems like it would be either a `string | undefined | null`.

mathisscott (changes requested):

packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -33,9 +33,11 @@ export function getUpdateSizeStrategy(size: string) {
 
 export function updateIconSizeStyle(el: CdsIcon, size: string) {
   const updateStrategy = getUpdateSizeStrategy(size);
+  let val = '';
+
   switch (updateStrategy) {
     case SizeUpdateStrategies.ValidNumericString:
-      const val = pxToRem(parseInt(size));
+      val = pxToRem(parseInt(size)); // set val in block to run expensive call only when needed
```
mathisscott, line 40: 🏆

mathisscott (approved):

hippee-lee (approved):

PR #5706, opened 2021-03-09
feat: core pagination

Shijir:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/src/pagination/pagination-button.element.scss
```diff
@@ -0,0 +1,51 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --height: #{$cds-global-space-9};
+  --min-width: #{$cds-global-space-9};
+  --padding: 0;
+  --border-color: var(--color);
+  --border-width: 0;
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --box-shadow-color: #{$cds-alias-object-opacity-100};
```
mathisscott: We should probably make this just `--box-shadow` and put the full style from line 27 in it. that would give users more flexibility to customize it or remove it.

packages/core/src/pagination/register.ts
```diff
@@ -0,0 +1,24 @@
...
+import '@cds/core/icon/register.js';
+import { registerElementSafely } from '@cds/core/internal';
+import { CdsPaginationButton } from './pagination-button.element.js';
+import { CdsPaginationNumber } from './pagination-number.element.js';
+import { CdsPagination } from './pagination.element.js';
+
+registerElementSafely('cds-pagination', CdsPagination);
+registerElementSafely('cds-pagination-number', CdsPaginationNumber);
+registerElementSafely('cds-pagination-button', CdsPaginationButton);
+
```
mathisscott, line 19: We need to do `ClarityIcons.addIcons()` for all the icons we are using in the pagination button. That's usually done in this file.

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles, CdsBaseButton, property, querySlot } from '@cds/core/internal';
+import { html } from 'lit-element';
+import { styles } from './pagination-button.element.css.js';
+
+export type CdsPaginationButtonAction = 'prev' | 'next' | 'first' | 'last';
```
mathisscott: If `CdsPaginationButtonAction` is only being used in this file, we probably don't need to export it.

packages/core/src/pagination/pagination-number.element.scss
```diff
@@ -0,0 +1,29 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --height: #{$cds-global-space-9};
```
mathisscott: People can set the height but they can't set the width?

packages/core/src/pagination/pagination-number.element.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles } from '@cds/core/internal';
+import { html, LitElement } from 'lit-element';
+import { styles } from './pagination-number.element.css.js';
+
+export class CdsPaginationNumber extends LitElement {
```
mathisscott: If I am reading this correctly, this component is like the textual content of the pagination? Could it be something other than a number? If so, we may want to consider renaming it to `CdsPaginationContent` instead...

packages/core/src/pagination/pagination.element.scss
```diff
@@ -0,0 +1,11 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --background: none;
+  --color: #{$cds-global-typography-color-500};
+  --font-size: #{$cds-global-typography-font-size-4};
+  --disabled-color: #{$cds-alias-status-disabled};
```
mathisscott: I see a number of custom properties here but it's not clear where or if they are used. I see `--color` being used on line 9. What about the others?

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,23 @@
...
+import { html, LitElement } from 'lit-element';
+import { styles } from './pagination.element.css.js';
+
+export const PAGINATION_LAYOUT = 'horizontal gap:md align:center';
+
+export class CdsPagination extends LitElement {
+  render() {
+    return this.hasAttribute('cds-layout')
+      ? html`<slot></slot>`
+      : html`<div cds-layout="${PAGINATION_LAYOUT}"><slot></slot></div>`;
```
mathisscott: I get why we made the layout a constant but note that it also creates a reference in the template the Lit Element will try to keep track of. Since this is kind of a one-and-done application of the `PAGINATION_LAYOUT` const, I would consider wrapping it in a `cache()` directive. You will need to import it through lit-html, though, not lit-element.

https://lit-html.polymer-project.org/guide/writing-templates#caching-template-results:-the-cache-directive

packages/core/src/pagination/pagination.stories.mdx
```diff
@@ -0,0 +1,68 @@
+import { Meta, Props, Story, Preview, API } from '@storybook/addon-docs/blocks';
+import { html, LitElement } from 'lit-element';
+
+<Meta title="Components (Preview)/Pagination" />
+
+# Pagination
+
+Pagination is for paginating.
```
mathisscott: 🤔

mathisscott (commented):
Overall, great job. I'm going to be OOO next week so I am going to submit my feedback as comments instead of blocking it by requesting changes.

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class CdsPaginationButton extends CdsBaseButton {
+  @property({ type: String }) action: CdsPaginationButtonAction;
+
+  /** @private */
+  @querySlot('cds-icon', { assign: 'cds-icon-slot' }) cdsIcon: HTMLElement;
```
coryrylan, line 67: This can use the TS `private` keyword instead of the JS doc `@private`

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -0,0 +1,40 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { baseStyles, CdsBaseButton, property, querySlot } from '@cds/core/internal';
+import { html } from 'lit-element';
+import { styles } from './pagination-button.element.css.js';
+
+export type CdsPaginationButtonAction = 'prev' | 'next' | 'first' | 'last';
+
+export class CdsPaginationButton extends CdsBaseButton {
```
bbogdanov, line 55: https://github.com/vmware/clarity/blob/0300acd8e1c40b7b480e400271ce3306c74cb568/packages/core/src/button/button.element.ts#L35-L63 

The comment will help the storybook page to show the available css props

packages/core/src/pagination/pagination-number.element.ts
Shijir: I agree that `cds-pagination-number` could be a poorly named component. But I am also avoiding `cds-pagination-content` as users might assume they could place/project their content only within that component. They could actually place their content anywhere within the `cds-pagination` like this:

```html
<cds-pagination>
    Lorem Ipsum
    <cds-pagination-button action="prev"></cds-pagination-button>
    <cds-pagination-number>1 / 3<cds-pagination-number>
    <cds-pagination-button action="next"></cds-pagination-button>
    Bacon Ipsum
</cds-pagination>
```

But I still think we could/should definitely rename it differently than `cds-pagination`. 

Also before renaming it, I would like to step back and provide some context in why we need `cds-pagination-number`. The goal of `cds-pagination-number` is to provide a **default** wrapper style to the current/total page numbers. Users can apply custom styles and control the spacing of the current/total pages numbers while keeping the pagination items' symmetrical look. For Example:

```html
<cds-pagination>
    <cds-pagination-button action="prev"></cds-pagination-button>
    <cds-pagination-number cds-layout="horizontal m-x:lg">1 / 3<cds-pagination-number>
    <cds-pagination-button action="next"></cds-pagination-button>
</cds-pagination>
```

But most importantly, the component also allows us to provide a default input style:

```html
<cds-pagination>
    <cds-pagination-button action="prev"></cds-pagination-button>
    <cds-pagination-number><input type="text" value="1" size="1" /> / 3</cds-pagination-number>
    <cds-pagination-button action="next"></cds-pagination-button>
</cds-pagination>
```

So without `cds-pagination-number`, we both lose the symmetrical look and the ability to target that particular input for the current page.

I also brainstormed some ideas with Cory. One was to use `cds-input` without a wrapper component such as like `cds-pagination-number`. However, after some investigation, it might just complicate the API more. For example:


```html
<cds-pagination>
    <cds-pagination-button action="prev"></cds-pagination-button>
    <cds-input box control-width="shrink">
      <input type="text" value="1" size="1" aria-label="Edit current page" />
    </cds-input> / 3
    <cds-pagination-button action="next"></cds-pagination-button>
</cds-pagination>
```

And even with this API, we might need a wrapper component to keep the symmetry when the current and total page numbers are used together. Another consideration is what if users might want to use not just one but multiple `cds-input` within `cds-pagination`. All of them could get the current page number input style. But I will continue my investigation.

gnomeontherun:
Where are we with this @Shijir ?

packages/core/src/pagination/pagination-number.element.scss
Shijir: Removed `cds-pagination-number` as it's no longer necessary.

packages/core/src/pagination/pagination.element.scss
Shijir: Removed most of the css properties except `--color` and `--font-size`.

packages/core/src/forms/control/control.element.ts
```diff
@@ -337,7 +344,7 @@ export class CdsControl extends LitElement {
   }
 
   private setupHiddenLabel() {
-    if (this.label.getAttribute('cds-layout')?.includes('display:screen-reader-only')) {
+    if (this.label?.getAttribute('cds-layout')?.includes('display:screen-reader-only')) {
       this.hiddenLabel = true;
```
mathisscott, line 347: According to karma, this line here isn't covered in our unit tests. Any way we could address that?

packages/core/src/pagination/pagination.element.ts
mathisscott: Any ideas on this comment?

packages/core/src/pagination/pagination.stories.ts
```diff
@@ -0,0 +1,312 @@
...
+  parameters: {
+    options: { showPanel: true },
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/v2mkhzKQdhECXOx8BElgdA/Clarity-UI-Library---light-2.2.0?node-id=51%3A673',
+    },
+  },
+};
+// pager, pagination, paginator
+export const API = (args: any) => {
```
mathisscott: Do we need `args` here?

mathisscott (changes requested):
It's looking pretty close. I have a couple of questions/requests but after that I think we are good to go.

packages/core/src/forms/control/control.element.ts
Shijir: Yes, will take a look.

packages/core/src/pagination/pagination.stories.ts
Shijir: No. I was meaning to remove but it slipped through. Thanks for catching it.

packages/core/src/pagination/pagination.element.ts
Shijir: Is it possible to use that `cache` directive within `lit-element` templates? I tried it before. It seems to work only with `lit-html` templates.

mathisscott: You should be able to...

https://lit-element.polymer-project.org/guide/templates#import-and-use-a-lit-html-directive

mathisscott (approved):

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -33,6 +33,13 @@ export interface I18nStrings {
     showButtonAriaLabel: string;
     hideButtonAriaLabel: string;
   };
+  pagination: {
+    paginationAriaLabel: string;
+    firstButtonAriaLabel: string;
+    prevButtonAriaLabel: string;
+    nextButtonAriaLabel: string;
+    lastButtonAriaLabel: string;
+  };
```
jeeyun: Hi Shijir,
There is no need to add these to i18n service unless the aria labels are INSIDE the template of the component and thus inaccessible to customize by our application developers. In the pagination case, they can easily do this without us having these as properties in the component. For example, this can be done without it being an explicit property of the pagination / pagination button components:

```
 <cds-pagination aria-label="pagination">
      <cds-pagination-button aria-label="go to first page" action="first" disabled></cds-pagination-button>
      <cds-pagination-button aria-label="go to previous page" action="prev" disabled></cds-pagination-button>
      <span aria-label="current page">1 / 3</span>
      <cds-pagination-button aria-label="go to next page" action="next"></cds-pagination-button>
      <cds-pagination-button aria-label="go to last page" action="last"></cds-pagination-button>
</cds-pagination>
```

If a11y team recommends we have a default value, we should maybe have a broader discussion since we don't always follow this pattern of pre-setting a default value. In any case, this can be achieved without using the i18n service since these aria-labels are on the host element and not inside the template.

packages/core/src/forms/control/control.element.ts
```diff
@@ -128,8 +128,11 @@ export class CdsControl extends LitElement {
 
   @querySlot('label', {
     required: 'error',
-    requiredMessage: 'To meet a11y standards a <label> should be provided.',
+    requiredMessage: 'To meet a11y standards either a <label> or input[aria-label] should be provided.',
     assign: 'label',
+    exemptOn: _this => {
+      return _this.inputControl?.hasAttribute('aria-label');
+    },
```
jeeyun, line 137: Great job with adding this to the decorator. I do think this lends to cleaner logic + testability. 👍

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    return [baseStyles, styles];
+  }
+
+  firstUpdated() {
+    this.setupNumberInput();
+  }
+
+  updated(props: Map<string, any>) {
+    super.updated(props);
+  }
```
jeeyun: Is this needed? Seems like we aren't executing any additional custom logic here.

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    return html`${cache(this.hasAttribute('cds-layout') ? userLayout : defaultLayout)}`;
+  }
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  firstUpdated() {
+    this.setupNumberInput();
+  }
```
jeeyun: probably need to call `super.firstUpdated()` in here as well.

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    super.updated(props);
+  }
+
+  private setupNumberInput() {
+    const numberInput = this.querySelector<CdsInput>('[cds-pagination-number]');
+    if (numberInput) {
+      numberInput.controlWidth = 'shrink';
+      numberInput.layout = 'compact';
+    }
+  }
```
jeeyun: To be consistent with the rest of the library let's use `@query` (import from 'lit-element'):

You can mark it as private so it won't show up on the dox:

```
export class CdsPagination extends LitElement {
  ...
  
  /** @private */
  @query('cds-input[cds-pagination-number]') numberInput: CdsInput;
  
  ...

  private setupNumberInput() {
      if (this.numberInput) {
        this.numberInput.controlWidth = 'shrink';
        this.numberInput.layout = 'compact';
      }
    }
  ```

File: packages/core/src/pagination/pagination-button.element.ts [hunk 18]
```diff
@@ -0,0 +1,121 @@
...
+          this.ariaLabel = I18nService.keys.pagination.nextButtonAriaLabel;
+          break;
+        case CdsPaginationButtonAction.Last:
+          this.ariaLabel = I18nService.keys.pagination.lastButtonAriaLabel;
+          break;
+        default:
+        // do nothing
+      }
+    }
+  }
```
jeeyun: I think this logic may have an unintented bug.

Let's say the component's property such as `disabled` (inherited from `CdsBaseButton`) is changed sometime in its lifecycle. I believe this will trigger the aria-label to be re-assigned to a pre-assigned value from our service and would override a previously set custom aria-label if the user supplied one through the attribute (i.e. `<cds-pagination-button aria-label='my-custom-label'>...</cds-pagination-button>`). Of course this isn't a problem if they used the `cds-i18n` attribute to set the custom label but I think this goes back to the question about whether to use the service in this component. I'd say no since this is a property directly on the host and not inside the template as mentioned in my other comment in this PR).

packages/core/src/pagination/pagination-button.element.scss
```diff
@@ -0,0 +1,58 @@
...
+
+:host(:active) .private-host {
+  box-shadow: var(--box-shadow);
+}
+
+:host(:hover),
+:host(:focus),
+:host(:active) {
+  .private-host {
+    filter: brightness(90%);
```
jeeyun, line 36: We had talked about tokenizing values like this with percentage in them because the same percentage might not work for all colors when you apply a custom theme on core components, etc.

Do we have any examples of this @coryrylan? I forgot the context / component in which we had this discussion before.

packages/core/src/internal/services/i18n.service.ts
Shijir: This is an important question. Let's discuss this on the chat.

packages/core/src/pagination/pagination.element.ts
Shijir: No, actually it should have been in the firstUpdated. Thank you!

packages/core/src/pagination/pagination-button.element.ts
Shijir: Since we've configured the property to reflect the aria-label attribute, that should be fine, right?

```
@property({ type: String, reflect: true }) ariaLabel
```

I have also set a custom `aria-label` value on the button and tested whether that user-provided text is being overridden or not:

https://github.com/vmware/clarity/pull/5706/files#diff-9299cc10edb451e7e097efac8e6b7ea84c1d95dd3a5ad996f3f98399028bb0aaR116-R125

mathisscott: I think @jeeyun is correct about an unintended bug. It goes something like this...

```
Everytime the lit element update cycle runs and "ariaLabel" is not present in the Map of updated properties this switch statement will reassign "ariaLabel" even if it was already assigned a value.
```

But the more important matter is that I don't think we need this. We don't need any default set here. We only need for "ariaLabel" to be a required property so that people are warned if it is not present.

Shijir: Ooh I see now. But yes, we no longer need default ariaLabel values.

packages/core/src/pagination/pagination-button.element.scss
Shijir: As we didn't need the entire base button style, I've only included this part from the base button style. 

And I agree. I also noticed it works differently for bright and dark colors.

coryrylan: Yeah we talked about building out a cds-action-button component for all the common action style button that datagrid uses (filtering, multi-action). I think this works for now and once the action button is complete it could inherit it without any breaking api change to the pagination components.

packages/core/src/forms/control/control.element.ts
```diff
@@ -290,6 +293,10 @@ export class CdsControl extends LitElement {
     this.observers.forEach(o => o.disconnect());
   }
 
+  get inputHasAriaLabel() {
+    return typeof this.inputControl?.hasAttribute === 'function' && this.inputControl.hasAttribute('aria-label');
```
coryrylan: Can we move this logic into `setupHiddenLabel`? This will ensure it only runs this check once after the first render. If its a getter it will query the DOM with each re-render.

```typescript
if (this.label?.getAttribute('cds-layout')?.includes('display:screen-reader-only') || this.inputControl?.hasAttribute('aria-label')) {

}
```

packages/core/src/internal/decorators/query-slot.ts
```diff
@@ -41,7 +46,9 @@ export function querySlot(selector: string, config?: QuerySlotConfig) {
     function firstUpdated(this: any): void {
       const ref = this.querySelector(selector);
 
-      if (!ref && config?.required) {
+      const shouldExempt = config?.exemptOn && config?.exemptOn(this);
+
+      if (!ref && config?.required && !shouldExempt) {
```
coryrylan: To help prevent development time validation logic running in prod environment can we add an additional condition here?

We can export the `notProductionEnvironment` from the log service file so components can access it through `@cds/internal`. The log service runs this check to not log but we also would want to make sure the components are not querying DOM for dev time validation checks if possible. 

https://github.com/vmware/clarity/blob/next/packages/core/src/internal/services/log.service.ts#L33 

```typescript
if (!ref && notProductionEnvironment() && config?.required && !shouldExempt) {
```

packages/core/src/pagination/pagination-button.element.scss
```diff
@@ -0,0 +1,58 @@
...
+
+:host {
+  --background: none;
+  --padding: 0;
+  --border-color: var(--color);
+  --border-width: 0;
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --box-shadow: 0 #{$cds-global-space-2} 0 0 #{$cds-alias-object-opacity-100} inset;
+  --disabled-background: none;
+  --disabled-color: #{$cds-alias-status-disabled};
```
coryrylan: I would omit `--disabled-background` and `--disabled-color` typically if we can get away with it any property that can match to attribute state we leave off. This reduces the [API surface and maintenance](https://clarity.design/storybook/core/?path=/story/internal-documentation-templates-and-styles--page#component-styles) for us but also leaves it customizable for consumers. This is similar to alerts/buttons statuses where the css properties are changed based on host attributes without introducing new props to represent each attr state.

```css
cds-pagination-button[disabled] {
  --color: green;
  --background: blue;
}
```

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  updated(props: Map<string, any>) {
+    super.updated(props);
+    this.setupNumberInput();
+    this.validateAriaLabel();
```
coryrylan, line 54: These probably can run in `firstUpdated` that way they are guaranteed to run once and not re-run during property updates.

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private setupNumberInput() {
+    if (this.numberInput) {
+      this.numberInput.controlWidth = 'shrink';
+      this.numberInput.layout = 'compact';
+    }
+  }
+
+  private validateAriaLabel() {
+    if (!this.getAttribute('aria-label')?.length) {
```
coryrylan: Similar to above we can add a prod check so dev time warnings dont run in prod env 

```typescript
if (notProductionEnvironment() && !this.getAttribute('aria-label')?.length) {
```

packages/core/src/pagination/pagination.stories.ts
```diff
@@ -0,0 +1,312 @@
...
+};
+
+export const API = () => {
+  return html`
+    <cds-pagination aria-label="pagination">
+      <cds-pagination-button aria-label="go to first" action="first" disabled></cds-pagination-button>
+      <cds-pagination-button aria-label="go to previous" action="prev" disabled></cds-pagination-button>
+      <span aria-label="current page">1 / 3</span>
+      <cds-pagination-button aria-label="go to next" action="next" aria-label="go to next"></cds-pagination-button>
+      <cds-pagination-button aria-label="go to last" action="last" aria-label="go to last"></cds-pagination-button>
```
coryrylan: duplicate aria-labels

packages/core/src/pagination/register.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import '@cds/core/button/register.js';
+import '@cds/core/icon/register.js';
+import '@cds/core/input/register.js';
+import { registerElementSafely } from '@cds/core/internal';
+import { ClarityIcons } from '../icon/icon.service.js';
+import { angleIcon, stepForward2Icon } from '../icon/index.js';
```
coryrylan: For internal icon imports load directly from the icon file. This helps with native esm loading performance not eagerly loading all icons in the index file. Also this helps treeshaking as each component is its own entrypoint/package. https://clarity.design/storybook/core/?path=/story/internal-documentation-using-icons--page 

```typescript
import { ClarityIcons } from '@cds/core/icon/icon.service.js';
import { angleIcon } from '@cds/core/icon/shapes/angle.js';
import { stepForward2Icon } from '@cds/core/icon/shapes/step-forward-2.js';
```

packages/core/src/pagination/pagination.element.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @cssprop --color
+ * @cssprop --font-size
+ */
+
+export class CdsPagination extends LitElement {
+  @querySlot('cds-input[cds-pagination-number]') private numberInput: CdsInput;
+
+  render() {
+    const userLayout = html`<slot></slot>`;
+    const defaultLayout = html`<div cds-layout="${PAGINATION_LAYOUT}"><slot></slot></div>`;
```
coryrylan: Dynamic layouts can break the treeshaking from PurgeCSS since it wont be able to statically check the html for what attrs/classes are used. We should update this to keep the values inlined to the template.

```html
<div cds-layout="horizontal gap:md align:center"><slot></slot></div>
```

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -0,0 +1,106 @@
...
+    `;
+  }
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  updated(props: Map<string, any>) {
+    super.updated(props);
+    this.validateAriaLabel();
```
coryrylan, line 103: This probably can run in firstUpdated that way its not revalidating for each re-render.

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private validateAriaLabel() {
+    if ((this.cdsIcon || this.cdsIconDefault) && !this.getAttribute('aria-label')?.length) {
```
coryrylan: Similar prevent dev validation during prod

```typescript
if (notProductionEnvironment() && (this.cdsIcon || this.cdsIconDefault) && !this.getAttribute('aria-label')?.length) {
```

coryrylan (changes requested):

packages/core/src/internal/decorators/query-slot.ts
Shijir: Sounds good. Will do.

packages/core/src/pagination/register.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import '@cds/core/button/register.js';
+import '@cds/core/icon/register.js';
+import '@cds/core/input/register.js';
+import { registerElementSafely } from '@cds/core/internal';
+import { ClarityIcons } from '../icon/icon.service.js';
```
coryrylan: Import path should be `import { ClarityIcons } from '@cds/core/icon/icon.service.js';`

coryrylan (approved):
One import to update but other than that lgtm

PR #5739, opened 2021-03-16
fix(form): core control group to disable all children

### LINKED ISSUE #5734: Disabled control group doesn't disable the controls
## Describe the bug

`CdsInternalControlGroup` disabled property doesn't disable the controls inside the group container.

`disabled` attribute is not passed to the controls inside the group container.

## How to reproduce

https://codesandbox.io/s/control-group-disabled-e973f

Click on any of the checkboxes.

## Expected behavior

Any of the controls inside a child of `CdsInternalControlGroup` should be disabled when the group has a `disabled` property.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [x] v4.x
- [x] v5.x

bbogdanov:
Closes #5734

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #5734 

## What is the new behavior?

The controls inside a group are all disabled when a `disabled` attribute is set on the group.

https://user-images.githubusercontent.com/15037947/111314944-90449080-866a-11eb-8279-cdb3a3cdfbcf.mov

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/forms/control/control.element.ts
+++ b/packages/core/src/forms/control/control.element.ts
@@ -280,6 +280,8 @@ export class CdsControl extends LitElement {
   updated(props: Map<string, any>) {
     super.updated(props);
     this.messages.forEach(message => syncProps(message, this, { disabled: props.has('disabled') }));
+
+    syncProps(this.inputControl, this, { disabled: props.has('disabled') });
   }
 
   disconnectedCallback() {
--- a/packages/core/src/forms/control-group/control-group.element.spec.ts
+++ b/packages/core/src/forms/control-group/control-group.element.spec.ts
@@ -91,6 +91,22 @@ describe('cds-internal-control-group', () => {
     expect(controls[1].status).toBe('error');
   });
 
+  it('should set the disabled style on all cds-controls and native html inputs if group is disabled', async () => {
+    controlGroup.disabled = true;
+
+    await componentIsStable(controlGroup);
+    await componentIsStable(controls[0]);
+    await componentIsStable(controls[1]);
+
+    // cds-control has style hook attr '_disabled'
+    expect(controls[0].hasAttribute('_disabled')).toBe(true);
+    expect(controls[1].hasAttribute('_disabled')).toBe(true);
+
+    // native html input has attr 'disabled'
+    expect(controls[0].inputControl.hasAttribute('disabled')).toBe(true);
+    expect(controls[1].inputControl.hasAttribute('disabled')).toBe(true);
+  });
+
   it('should adjust for compact layouts and messages', async () => {
     controlGroup.layout = 'compact';
     await componentIsStable(controlGroup);


coryrylan (approved):

packages/core/src/forms/control/control.element.ts
```diff
@@ -280,6 +280,8 @@ export class CdsControl extends LitElement {
   updated(props: Map<string, any>) {
     super.updated(props);
     this.messages.forEach(message => syncProps(message, this, { disabled: props.has('disabled') }));
+
+    syncProps(this.inputControl, this, { disabled: props.has('disabled') });
```
coryrylan, line 284: We should add a test to check this, in the control tests you can test that the child element is set when the component disabled is set similar to this test https://github.com/vmware/clarity/blob/next/packages/core/src/forms/control/control.element.spec.ts#L63

coryrylan (changes requested):

packages/core/src/forms/control/control.element.ts
bbogdanov: Okay.

I am also trying to add a test in `control-group.element` but I can't find a way the `input` to receive a `disabled` attribute. 

`controlGroup.disabled = true` didn't work
`controlGroup.update(new Map().set('disabled' , true))` didn't work as well

bbogdanov: image

coryrylan: Its likely because lit renders async, while you have an await on the controlGroup it would also need to await on the child cds-control inputs that would asynchronously be setting the disabled attr on the input

coryrylan: This wont test the group use case but this will test that the control does set the native input correctly which I think is good for this fix. 

```javascript
it('should disable native input if cds-control is set disabled (set by cds-control-group)', async () => {
  await componentIsStable(control);
  expect(input.hasAttribute('disabled')).toBe(false);

  control['disabled'] = true; // control is a protected property only allowed to be set by the cds-control-group
  await componentIsStable(control);
  expect(input.hasAttribute('disabled')).toBe(true);
});
```

bbogdanov: OK but as I said in slack the `control['disabled']` is a possibility but do we really want to go into that direction? I had issues with typescript before when doing tests like this. I don't remember what was it but the issues appeared with upgrading to newer version of typescript before.

coryrylan: Core spec files have relaxed rules on dynamic property access mainly to set internal states for mocks. However you can avoid this and test this at the group level all the way down to the native input.


```typescript
// control-group.element.spec.ts
it('should set the disabled style on all cds-controls and native html inputs if group is disabled', async () => {
  controlGroup.disabled = true;

  await componentIsStable(controlGroup);
  await componentIsStable(controls[0]);
  await componentIsStable(controls[1]);

  // cds-control has style hook attr '_disabled'
  expect(controls[0].hasAttribute('_disabled')).toBe(true); 
  expect(controls[1].hasAttribute('_disabled')).toBe(true);

  // native html input has attr 'disabled'
  expect(controls[0].inputControl.hasAttribute('disabled')).toBe(true);
  expect(controls[1].inputControl.hasAttribute('disabled')).toBe(true);
});
```

mathisscott (approved):

coryrylan (approved):

PR #5755, opened 2021-03-18
feat(table): add Core table

gnomeontherun:
This adds the Core version of tables, which are just a CSS styling that fit with design tokens.

TODO

* [x] Fix the design tokens
* [x] Confirm design choices
* [x] Test more variations of tables
* [x] Add better documentation

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  --cds-alias-object-color-flat: #{$cds-global-color-gray-0};
+  --cds-alias-object-border-radius-50: 0.1rem; // 2px, existing radius on table
+}
+
+[cds-theme~='dark'] {
+  --cds-alias-object-color-flat: #{$cds-global-color-construction-1000};
+}
+
+$cds-alias-table-color-background: var(--cds-alias-object-color-flat);
+$cds-alias-object-border-radius-50: var(--cds-alias-object-border-radius-50);
```
coryrylan: The table should probably use `cds-alias-object-border-radius-100` for border radius so its consistent with the rest of core. If we need a 2pt we should sync up with Colin on if there should be a new border radius token and add it to the token file.

For the background using `cds-alias-object-container-background` will align with the other core container backgrounds and auto switch for dark theme. The `gray` palette is being deprecated internally with the components so all the components have moved over to use the `construction` palette.

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  td,
+  th {
+    line-height: #{$cds-global-typography-font-size-4};
+    padding: #{$cds-global-typography-font-size-1} #{$cds-global-typography-font-size-2};
+    text-align: center;
+    vertical-align: top;
+  }
+
+  th {
+    color: $cds-alias-status-neutral;
```
coryrylan: Using `$cds-global-typography-color-400` or 300 will get the color contrast and dark theme contrast automatically

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+$cds-alias-object-border-radius-50: var(--cds-alias-object-border-radius-50);
+
+// Need to get typography colors correct
+
+// Styles and default bordering
+[cds-table] {
+  background-color: $cds-alias-table-color-background;
+  border-collapse: separate;
+  border-spacing: 0;
+  color: #{$cds-global-color-gray-700};
```
coryrylan: Using `$cds-global-typography-color-400` or 300 will get the color contrast and dark theme contrast automatically

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  th {
+    line-height: #{$cds-global-typography-font-size-4};
+    padding: #{$cds-global-typography-font-size-1} #{$cds-global-typography-font-size-2};
+    text-align: center;
+    vertical-align: top;
+  }
+
+  th {
+    color: $cds-alias-status-neutral;
+    background-color: $cds-alias-status-neutral-tint;
```
coryrylan: ~~For the headings/highlight background you could use `$cds-alias-object-interaction-background-selected`. Its brighter than the original but aligns with what the new accordion has. It also gets you the dark theme switch automatically.~~

Looking at my previous suggestion it doesnt look right especially if there were selected rows like datagrid. Maybe instead of the `flat` we could add a tint option to the current container background to match the tint/shade convention with other tokens? Maybe something like

`cds-alias-object-container-background-tint: : var(--cds-global-color-construction-50)`
`cds-alias-object-container-background-tint: : var(--cds-global-color-construction-800)` // dark theme
https://clarity.design/storybook/core/?path=/story/foundation-object-styles--page 

Screen Shot 2021-03-23 at 10 10 16 AM


Screen Shot 2021-03-23 at 10 09 22 AM

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+    padding-bottom: #{$cds-global-layout-space-xs};
+  }
+}
+
+// Vertical options
+[cds-table~='vertical'] {
+  tr td:first-child,
+  tr th:first-child {
+    background-color: $cds-alias-status-neutral-tint;
+    // color: ??;
```
coryrylan, line 152: If you use `cds-alias-object-interaction-background-selected` and `cds-alias-object-interaction-color-selected` you will get the right contrast levels for light and dark theme.

packages/core/src/table/table.scss
gnomeontherun: The table radius at 4px is pretty stark and unlike current table. This is where I need to talk with design on that kind of a change being approved or not.

Can switch the background color.

gnomeontherun: Reworked tokens to use defaults or added them to tokens file.

packages/core/.storybook/public/demo.css
```diff
@@ -5,6 +5,7 @@
 @import './../../dist/core/styles/theme.dark.min.css';
 @import './../../dist/core/styles/theme.low-motion.min.css';
 @import './../../dist/core/list/list.min.css';
+@import './../../dist/core/table/table.min.css';
```
mathisscott, line 8: 🏆

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,205 @@
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@import './../styles/tokens/generated/index';
+
```
mathisscott, line 8: The list styles didn't make use of CSS custom properties. I believe that is largely because they were ported directly over from `clr/ui`. 

Would it make sense to have a few custom properites under `[cds-table]` to dry up all the repeated uses of global and alias design tokens? This would also make it easier for tables to be themed/customized independently of design tokens, reducing the cognitive load of having to understand how the aliases and global tokens work to change, say, the border radius of one table...

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,205 @@
...
+// Padding options
+[cds-table~='p:xxs'] {
+  th,
+  td {
+    padding: $cds-global-layout-space-xxs;
+  }
+}
+
+[cds-table~='p:xs'] {
```
mathisscott: Do we want to allow for different horizontal and vertical padding as well? For layouts, this would look like `p-x:md` and `p-y:md`. I don't think we would need to go any more granular than that.

Also, what about allowing for spacing to adjust responsively? Like the tokens can do. This looks like `p@md:lg` and `p-x@sm:xs`...

packages/core/src/table/table.stories.mdx
```diff
@@ -0,0 +1,74 @@
+import { Meta, Story, Preview, API } from '@storybook/addon-docs/blocks';
+import { html, LitElement } from 'lit-element';
+
+<Meta title="Components/Table" />
+
+# Tables
+
+Tables in Clarity Core come in three varieties – standard, compact, andv ertical.
```
mathisscott: `and vertical` not `andv ertical`

packages/core/src/table/table.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Installation
+
+To use the list styles import the global CSS.
```
mathisscott: `To use the table styles` not `To use the list styles`

packages/core/src/table/table.stories.ts
```diff
@@ -0,0 +1,491 @@
...
+/** @website */
+export function leftTable() {
+  return html`
+    <table cds-table="left">
+      <caption>
+        Demo with the cells left aligned in a Clarity table.
+      </caption>
+      <thead>
+        <tr>
+          <th>Wizard</th>
```
mathisscott: Lol. We're keeping the wizards? Love it. 🧙

packages/core/src/table/table.stories.ts
```diff
@@ -0,0 +1,491 @@
...
+          <th>Triwizard Champion?</th>
+          <th>Can Cast Fireball</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td cds-col="left">Use cds-col="left" to left align cell</td>
+          <td>Gryffindor</td>
+          <td>Yes</td>
+          <td cds-col="right">Use cds-col="right" to right align cell</td>
```
mathisscott: I was going to ask about this. 🏆

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,205 @@
...
+[cds-table] td[cds-col~='right'],
+[cds-table] th[cds-col~='right'] {
+  text-align: right;
+}
+
+// No Border options
+[cds-table~='noborder'],
+[cds-table~='noborder'] tr th,
+[cds-table~='noborder'] tr td {
+  border-width: 0 !important;
```
mathisscott: If we use custom properties, removing borders could become either a vanilla css override of the custom properties (and allow us to remove the `noborder` variant _or_ we could keep the variant and just override the custom properties (without needing the `!important`)

mathisscott (changes requested):

packages/core/src/table/table.scss
gnomeontherun: I started with making everything an alias for the tables before using it, then reverted to just using globals. It would make sense to do some, but need to consider how many are meaningful. Any guidance there?

packages/core/src/table/table.scss
gnomeontherun: I wasn't sure this was even desirable to support or not. I did it just to see, but there could be a lot of permutations here if we want to support the layout token styles.

packages/core/src/table/table.stories.mdx
gnomeontherun: Are you sure?

packages/core/src/table/table.stories.ts
gnomeontherun: Was not sure what other modifiers we might support though?

packages/core/src/table/table.scss
gnomeontherun: @mathisscott Do you mean to localize properties like this? 

```
:host {
  --background: #{$cds-alias-object-container-background};
  --background-tint: #{$cds-alias-object-container-background-tint};
  --background-shade: #{$cds-alias-object-container-background-shade};
  --color: #{$cds-global-typography-color-400};
  --border-radius: #{$cds-alias-object-border-radius-100};
  --border-width: #{$cds-alias-object-border-width-100};
  --border-color: #{$cds-alias-object-border-color};
  --border-tint: #{$cds-alias-object-border-color-tint};
}
```

packages/core/src/table/table.scss
gnomeontherun: If I do this, ought I make a mixin to generate these many permutations?

packages/core/src/table/table.scss
coryrylan: I think you could treat it like other web components localize just the vars to the table selectors. This could help align closer to the other component APIs and help migrate if table ever became a custom element.  

```scss
[cds-table] {
  --background: #{$cds-alias-object-container-background};
  --color: #{$cds-global-typography-color-400};
  --border-radius: #{$cds-alias-object-border-radius-100};
  --border: #{$cds-alias-object-border-width-100} solid #{$cds-alias-object-border-color};
}
```

packages/core/src/table/table.scss
coryrylan: I don't think the table should be doing that level of work. That should be controlled with layouts or the themes rather than a custom component specific feature. Tables cells can use layouts directly. Tables are usually dynamic so the repetitiveness is not as much of a concern.

```html
<td cds-layout="p:lg">

</td>
```

Or the padding can be controlled via themes/custom properties

```scss
[cds-table] {
  --cell-padding: $cds-global-typography-font-size-1 $cds-global-typography-font-size-2;
}

td, th {
  padding: var(--padding);
}
```

This would also better align using a compact theme rather than component specific APIs. See clubhouse story 602. This would help create compact options for datagrid and treeview as a single API solution for the entire system.

```html
<table cds-table cds-theme="compact-space">

</table>
```

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+[cds-table~='noborder'] tr td {
+  --border-width: 0;
+  background-color: inherit;
+}
+
+// Compact options
+[cds-table~='compact'] {
+  td,
+  th {
+    padding: calc(var(--cell-padding-vertical) / 2) calc(var(--cell-padding-horizontal) / 2);
```
bbogdanov: Is this the reason to go with separate values for cell padding?

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  --background-tint: #{$cds-alias-object-container-background-tint};
+  --background-shade: #{$cds-alias-object-container-background-shade};
+  --color: #{$cds-global-typography-color-400};
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --border-width: #{$cds-alias-object-border-width-100};
+  --header-separator-width: #{$cds-alias-object-border-width-100};
+  --border-color: #{$cds-alias-object-border-color};
+  --border-tint: #{$cds-alias-object-border-color-tint};
+  --cell-padding-vertical: #{$cds-global-space-5};
+  --cell-padding-horizontal: #{$cds-global-space-6};
```
coryrylan: To simplify the API could these be combined into a `--cell-padding` variable? Reasons described [here](https://clarity.design/storybook/core/?path=/story/internal-documentation-templates-and-styles--page#component-styles)

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  --font-size: #{$cds-global-typography-body-font-size};
+}
+
+// Styles and default bordering
+[cds-table] {
+  background-color: var(--background);
```
coryrylan: Can this be updated to `background` that way any shorthand values can be assigned?

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+  td,
+  th {
+    padding: var(--cell-padding-vertical) var(--cell-padding-horizontal);
+    text-align: center;
+    vertical-align: top;
+  }
+
+  th {
+    color: var(--color);
+    background-color: var(--background-shade);
```
coryrylan: Looking at the demo it looks like maybe this should should be the background tint?

packages/core/src/table/table.scss
gnomeontherun: Correct, this is a shortcut to just half the padding, and there might be differences with vertical and horizontal (there were in the previous table).

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,158 @@
...
+    border-bottom-width: var(--header-separator-width);
+    border-bottom-style: solid;
+    border-bottom-color: var(--border-color);
+    border-top-width: var(--border-width);
+    border-top-style: solid;
+    border-top-color: var(--border-color);
```
coryrylan: nit: CSS vars allow combined values so these can be border-top: var(--border-width) solid var(--border-color);

packages/core/src/table/table.scss
coryrylan: To preserve theming behavior, alternate style options through attrs should modify through the CSS props when possible.

```css
[cds-table~='compact'] {
  --cell-padding: calc(#{$cds-global-space-5} / 2) calc(#{$cds-global-space-6} / 2);
}
```

This allows users to define their own custom theme states by altering the custom properties based on the attr. An example of this, a user can adjust the look of cds-button in a particular state by hooking into the attr and custom prop.

```css
cds-button[status=success] {
  --background: purple;
}
```

packages/core/src/table/table.scss
bbogdanov: I thought about this as well but then I saw this https://github.com/vmware/clarity/pull/5755#discussion_r626271134

gnomeontherun:
@mathisscott @coryrylan I've updated the PR for review.

The expectation is that the default table will be `<table cds-table="b:row b:outside">` to achieve the same display as the Clarity Angular choices. All borders are opt-in, and default layout switches to left but could be changed to `cds-table="center"`. There are some specialized modifiers that are needed in quirky cases where you might use a colspan or rowspan on a cell. Since cells can span, the CSS selectors might not apply correctly and `cds-border` and `cds-radius` are just helpers to manually set the border and radius in such cases. The Quirks mode samples are just demonstrations of various types of tables that are functional HTML but might be missing things like a caption, tbody, etc.

packages/core/src/table/table.stories.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+- `<table cds-table>` - Style a table with Clarity styles, but no borders
+- `<table cds-table="b:all">` - Turn on all borders
+- `<table cds-table="b:outside">` - Turn on outside borders
+- `<table cds-table="b:row">` - Turn on row borders
+- `<table cds-table="b:col">` - Turn on col borders
+- `<table cds-table="b:col b:outside">` - Turn on both outsidie and col borders
+- `<table cds-table="vertical">` - Display table in vertical layout
+- `<table cds-table="compact">` - Display table in compact layout
+- `<table cds-table="zebra">` - Display table in with alternating background colors
```
mathisscott, line 36: This is awesome. 🏆

mathisscott (approved):

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,233 @@
...
+// Right Alignment options
+[cds-text~='right'] th,
+[cds-text~='right'] td {
+  text-align: right;
+}
+
+[cds-text~='center'] th,
+[cds-text~='center'] td {
+  text-align: center;
+}
```
coryrylan: These should be scoped to the table. Since `cds-text` is a global utility, it usually applies to only the element its placed on. This helps our styles not collide with projects that have a mix of clr / core or any other ui framework.

```css
[cds-table][cds-text=''] th { } 
[cds-table][cds-text=''] td { } 
```

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,233 @@
...
+  border-left: var(--border) !important;
+}
+
+[cds-border~='bottom'] {
+  border-bottom: var(--border) !important;
+}
+
+[cds-border~='right'] {
+  border-right: var(--border) !important;
+}
```
coryrylan: Both border and radius should be scoped to the table since these are specific to table but in a global style sheet. A couple of ideas, 

This aligns a bit closer to css only components like cds-list
```css
[cds-table~='border:top'] { }
```

or

```css
[cds-table-border='top'] { } 
```

Scoping with a CSS selector with the table attr is not ideal. It would mean cds-border/radius would be privileged APIs to only table. No other elements could use the attributes safely and get any static typing on the TS global [HTMLElement interface](https://github.com/vmware/clarity/pull/5755/files#diff-9a4ceebe7c6f86856371906c3f061d3b56b7457022b05179884a113e7ced67e8R16). 

https://code.visualstudio.com/blogs/2020/02/24/custom-data-format 

```css
[cds-table] [cds-border~='top'] { } /* cds-border would be a global attr but only ever available to table */
```

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,233 @@
...
+    border-right: var(--header-border);
+  }
+
+  tr td:not(:last-child) {
+    border-right: var(--cell-border);
+  }
+}
+// Outside borders
+table[cds-table~='b:all'],
+table[cds-table~='b:outside'] {
```
coryrylan: Here `b:` represents the border but there is also `cds-border` and `cds-radius`. Maybe the naming could be aligned? It would also help with the global scoping on some of the utils.

```html
<table cds-table="border:outside">
  <tr cds-table="radius:none">
    <td cds-table="border:right">
```

```html
<table cds-table="border:outside">
  <tr cds-table-radius="none">
    <td cds-table-border="right">
```

packages/core/src/table/table.scss
```diff
@@ -0,0 +1,233 @@
...
+
+  [cds-radius~='tr'] {
+    border-top-right-radius: var(--border-radius);
+  }
+
+  [cds-radius~='bl'] {
+    border-bottom-left-radius: var(--border-radius);
+  }
+
+  [cds-radius~='br'] {
```
coryrylan: For the border/radius utils I don't think there is a storybook example of either. Do we want to remove them or add a storybook demo?

packages/core/src/table/table.stories.mdx
```diff
@@ -0,0 +1,97 @@
...
+- `<td cds-radius="tl">` - Set the border radius override if needed for special cases with rowspan or colspan, can be top left `tl`, top right `tr`, bottom left `bl`, bottom right `br`, or disable `none`
+- `<td cds-border="top">` - Set the border on a cell if needed for special cases with rowspan or colspan, can be `top`, `left`, `right`, `bottom`, or `none`.
+
+You can also apply `cds-text` modifiers to the table, row, or specific cells for alignment, typography styles, etc.
+
+## Basic Table
+
+<Preview>
+  <Story id="stories-table--api" />
+</Preview>
```
coryrylan, line 89: The api demo is good, can we add a `basic` story use here instead of the API story? This is helpful since the code snippet that is generated doesnt have storybook specific code in the snippets and website docs. Example accordion https://github.com/vmware/clarity/blob/next/packages/core/src/accordion/accordion.stories.ts#L37

packages/core/tsconfig.json
```diff
@@ -37,7 +37,7 @@
         "rules": {
           "no-missing-import": "off",
           "no-unknown-attribute": "off",
-          "globalAttributes": ["cds-layout", "cds-list", "cds-text"]
+          "globalAttributes": ["cds-layout", "cds-list", "cds-table", "cds-text"]
```
coryrylan, line 40: Depending on the APIs above attrs may need to be added for border/radius.

coryrylan (changes requested):

gnomeontherun:
@coryrylan Let me summarize the API changes based on your input.

I removed the modifier styles, they are expect to be minor and edge case use only. Apps may be better suited to writing some simple CSS (demo in the quirks mode) to modify properties as needed.  We could always add something if needed, but that keeps everything scoped to `cds-table` as the only API we expose.

coryrylan (approved):

Issue #5563, opened 2021-02-04
Core component: Card


bbogdanov:
Implement Clarity Card in Clarity Core

## API Proposal

[Here](https://github.com/vmware/clarity/blob/22d80b6badbf3250d3a4db7d6ef8aba8de0a7428/packages/core/src/card/card.stories.mdx) and [Here](https://github.com/vmware/clarity/blob/22d80b6badbf3250d3a4db7d6ef8aba8de0a7428/packages/core/src/internal-components/panel/panel.stories.mdx)

bbogdanov:
#5634

PR #5785, opened 2021-03-29
feat(card): introducing card component

bbogdanov:

Closing #5563

### DIFF
--- a/packages/core/src/card/card.element.scss
+++ b/packages/core/src/card/card.element.scss
@@ -1,0 +1,20 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --width: fit-content;
+  --height: auto;
+  --padding: var(--cds-global-space-6) var(--cds-global-space-8);
+
+  --background: #{$cds-alias-object-container-background};
+  --border: #{$cds-alias-object-border-width-100} solid #{$cds-alias-object-container-border-color};
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --box-shadow: #{$cds-alias-object-shadow-100};
+
+  --card-remove-margin: var(--cds-global-space-8);
+}
+
+@supports (-moz-appearance: none) {
+  :host {
+    --width: -moz-fit-content;
+  }
+}
--- a/packages/core/src/card/card.element.spec.ts
+++ b/packages/core/src/card/card.element.spec.ts
@@ -1,0 +1,29 @@
+import '@cds/core/card/register.js';
+import { html } from 'lit-html';
+import { componentIsStable, createTestElement, removeTestElement } from '@cds/core/test';
+import { CdsCard } from '@cds/core/card';
+
+describe('cds-card element: ', () => {
+  let testElement: HTMLElement;
+  let component: CdsCard;
+  const placeholderText = 'Placeholder';
+
+  beforeEach(async () => {
+    testElement = await createTestElement(html`<cds-card>${placeholderText}</cds-card>`);
+    component = testElement.querySelector<CdsCard>('cds-card');
+  });
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  it('should create the component', async () => {
+    await componentIsStable(component);
+    expect(component).toBeTruthy();
+  });
+
+  it('should project the content', async () => {
+    await componentIsStable(component);
+    expect(component.innerText).toBe(placeholderText);
+  });
+});
--- a/packages/core/src/card/card.element.ts
+++ b/packages/core/src/card/card.element.ts
@@ -1,0 +1,65 @@
+import { property, globalStyle } from '@cds/core/internal';
+import { CdsInternalPanel } from '@cds/core/internal-components/panel';
+import { css } from 'lit-element';
+import { styles } from './card.element.css.js';
+
+export const CdsCardTagName = 'cds-card';
+
+/**
+ * Web component card.
+ *
+ * ```typescript
+ * import '@cds/core/card/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-card>
+    <div cds-layout="vertical gap:md">
+      <div cds-text="section" cds-layout="p-y:sm">
+        <!-- Card title -->
+      </div>
+
+      <cds-divider cds-card-remove-margin></cds-divider>
+
+      <div cds-text="body" cds-layout="p-y:md">
+        <!-- Card content -->
+      </div>
+
+      <cds-divider cds-card-remove-margin></cds-divider>
+
+      <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center">
+        <cds-button action="flat-inline">View</cds-button>
+      </div>
+    </div>
+  </cds-card>
+ * ```
+ *
+ * @element cds-card
+ * @slot - For projecting card content
+ * @cssprop --width
+ * @cssprop --height
+ * @cssprop --color
+ * @cssprop --overflow
+ * @cssprop --overflow-x
+ * @cssprop --overflow-y
+ * @cssprop --background
+ * @cssprop --border
+ * @cssprop --border-radius
+ * @cssprop --padding
+ * @cssprop --box-shadow
+ * @cssprop --cds-card-remove-margin
+ */
+export class CdsCard extends CdsInternalPanel {
+  @property({ type: String, reflect: true }) role = 'region';
+
+  @globalStyle() globalStyles = css`
+    [cds-card-remove-margin] {
+      margin-left: calc(-1 * var(--card-remove-margin));
+      width: calc(100% + calc(var(--card-remove-margin) * 2));
+    }
+  `;
+
+  static get styles() {
+    return [...super.styles, styles];
+  }
+}
--- a/packages/core/src/card/card.stories.mdx
+++ b/packages/core/src/card/card.stories.mdx
@@ -1,0 +1,43 @@
+import { Meta, Props, Story, Preview, API } from '@storybook/addon-docs/blocks';
+
+<Meta title="Components/Card" />
+
+# Card
+
+## Installation
+
+```typescript
+import '@cds/core/card/register.js';
+```
+
+## Usage
+
+```html
+<cds-card>
+  <div cds-layout="vertical gap:md">
+    <div cds-text="section" cds-layout="p-y:sm">
+      Card Title
+    </div>
+
+    <cds-divider cds-card-remove-margin></cds-divider>
+
+    <div cds-text="body" cds-layout="p-y:md">
+      Card Content
+    </div>
+
+    <cds-divider cds-card-remove-margin></cds-divider>
+
+    <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center">
+      <cds-button action="flat-inline">View</cds-button>
+    </div>
+  </div>
+</cds-card>
+```
+
+<Preview>
+  <Story id="stories-card--with-layout-two-dividers-and-button" />
+</Preview>
+
+## API
+
+<Props of={'cds-card'} />
--- a/packages/core/src/card/card.stories.ts
+++ b/packages/core/src/card/card.stories.ts
@@ -1,0 +1,287 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import '@cds/core/card/register.js';
+import '@cds/core/divider/register.js';
+import { getElementStorybookArgs, getElementStorybookArgTypes, spreadProps } from '@cds/core/internal';
+import { html } from 'lit-html';
+import customElements from '../../dist/core/custom-elements.json';
+import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
+import { form } from '../forms/forms.stories';
+
+const placeholder = 'Placeholder';
+const basicContent = `
+    Message: Can you imagine what we will be downloading in another twenty years?
+    Who would have ever thought that you could record sound with digital quality fifty years ago?
+    Now we routinely download whole albums worth of music in a couple of minutes.`;
+const longerContent = `
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+    Praesent volutpat tortor eget quam auctor, quis sagittis libero auctor.
+    Nulla augue ante, tincidunt sit amet semper vitae, tempus at ipsum.
+    Vestibulum elementum, turpis quis ullamcorper fermentum, elit turpis placerat ipsum, quis convallis ex nisi sit amet lacus.
+    Ut enim ipsum, tincidunt nec luctus id, pharetra id velit. Aliquam nec elit ut neque lacinia mattis id ac lorem.
+    Vivamus egestas massa nulla, ac elementum purus pretium eu. Duis ultrices nec tortor varius consectetur.
+    Maecenas convallis bibendum urna, et aliquam quam tincidunt in.
+    Nam viverra, leo vitae imperdiet ultrices, ante mi facilisis mi, ac vehicula erat nulla sed nibh.
+    Nam euismod gravida est, convallis placerat nunc pharetra sed. Nullam scelerisque dui augue, eu porta erat tempus in.
+
+    Integer egestas massa orci, id scelerisque libero lobortis vitae.
+    Vivamus aliquet sem massa, eget sodales sapien dictum eu.
+    Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos.
+    Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos.
+    Suspendisse non interdum sem, nec egestas nulla.
+    Duis vitae leo justo. Fusce ante enim, tincidunt nec faucibus sed, fringilla et ipsum.
+    Proin quis diam elit. Cras nec finibus libero. Sed bibendum lectus eget magna ultricies eleifend.
+    Duis ut molestie urna. Proin cursus dolor ac ex rutrum, eu condimentum elit tincidunt.
+    In a dolor faucibus, vehicula ipsum eu, viverra lorem. Aenean congue fermentum commodo.
+    Sed ex purus, consectetur mollis urna eget, vestibulum mollis dolor.
+    Donec sodales, felis sit amet gravida ornare, magna dolor ullamcorper turpis, vel facilisis odio ipsum semper sem.
+  `;
+
+export default {
+  title: 'Stories/Card',
+  component: 'cds-card',
+  argTypes: getElementStorybookArgTypes('cds-card', customElements),
+  parameters: {
+    options: { showPanel: true },
+  },
+};
+
+export function Basic(args: { [key: string]: unknown }) {
+  return html`<cds-card ...="${spreadProps(getElementStorybookArgs(args))}">
+    <p cds-text="body">${placeholder}</p>
+  </cds-card>`;
+}
+
+export function WithLayout() {
+  return html` <cds-card>
+    <div cds-layout="vertical gap:md">
+      <div cds-text="section" cds-layout="p-y:sm">
+        Card Title
+      </div>
+
+      <div cds-text="body" cds-layout="p-y:lg">
+        ${basicContent}
+      </div>
+
+      <cds-divider cds-card-remove-margin></cds-divider>
+
+      <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+    </div>
+  </cds-card>`;
+}
+
+export function WithContainerOfCards() {
+  return html`
+    <div cds-layout="grid cols@md:6 cols@lg:3 gap:md">
+      <cds-card>
+        <div cds-layout="vertical gap:md">
+          <div cds-text="section" cds-layout="horizontal align:vertical-center">
+            Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+          </div>
+
+          <div cds-text="body" cds-layout="p-y:lg">
+            ${basicContent}
+          </div>
+
+          <cds-divider cds-card-remove-margin></cds-divider>
+
+          <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+        </div>
+      </cds-card>
+      <cds-card>
+        <div cds-layout="vertical gap:md">
+          <div cds-text="section" cds-layout="horizontal align:vertical-center">
+            Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+          </div>
+
+          <div cds-text="body" cds-layout="p-y:lg">
+            ${basicContent}
+          </div>
+
+          <cds-divider cds-card-remove-margin></cds-divider>
+
+          <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+        </div>
+      </cds-card>
+      <cds-card>
+        <div cds-layout="vertical gap:md">
+          <div cds-text="section" cds-layout="horizontal align:vertical-center">
+            Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+          </div>
+
+          <div cds-text="body" cds-layout="p-y:lg">
+            ${basicContent}
+          </div>
+
+          <cds-divider cds-card-remove-margin></cds-divider>
+
+          <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+        </div>
+      </cds-card>
+      <cds-card>
+        <div cds-layout="vertical gap:md">
+          <div cds-text="section" cds-layout="horizontal align:vertical-center">
+            Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+          </div>
+
+          <div cds-text="body" cds-layout="p-y:lg">
+            ${basicContent}
+          </div>
+
+          <cds-divider cds-card-remove-margin></cds-divider>
+
+          <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+        </div>
+      </cds-card>
+    </div>
+  `;
+}
+
+export function WithAlert() {
+  return html`
+    <cds-card>
+      <div cds-layout="vertical gap:md">
+        <div cds-text="section" cds-layout="horizontal align:vertical-center">
+          Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+        </div>
+
+        <div cds-text="body" cds-layout="p-y:lg">
+          <cds-alert-group>
+            <cds-alert status="info">This is an alert with a status of "info"</cds-alert>
+          </cds-alert-group>
+        </div>
+
+        <cds-divider cds-card-remove-margin></cds-divider>
+
+        <div cds-layout="horizontal gap:sm p-y:sm align:vertical-center"></div>
+      </div>
+    </cds-card>
+  `;
+}
+
+export function WithForms() {
+  return html`
+    <cds-card>
+      <div cds-layout="vertical gap:md">
+        <div cds-text="section" cds-layout="horizontal align:vertical-center">
+          Card Title <cds-icon shape="times" cds-layout="align:right"></cds-icon>
+        </div>
+
+        <div cds-text="body" cds-layout="p-y:lg">
+          ${unsafeHTML(form().getHTML())}
+        </div>
+
+        <cds-divider cds-card-remove-margin></cds-divider>
+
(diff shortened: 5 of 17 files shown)


packages/core/src/card/card.element.scss
```diff
@@ -0,0 +1,18 @@
...
+  --border: #{$cds-alias-object-border-width-100} solid #{$cds-alias-object-container-border-color};
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --box-shadow: #{$cds-alias-object-shadow-100};
+}
+
+/**
+ * TODO: Have to figure out why is not working the `:not` selector to `cds-divider` component
+ */
+.private-host [cds-layout] :not(cds-divider) {
+  padding: 1rem;
```
mathisscott: padding should probably be configurable with a custom property as well...

packages/core/src/card/card.element.scss
```diff
@@ -0,0 +1,18 @@
...
+  --height: auto;
+
+  --background: #{$cds-alias-object-container-background};
+  --border: #{$cds-alias-object-border-width-100} solid #{$cds-alias-object-container-border-color};
+  --border-radius: #{$cds-alias-object-border-radius-100};
+  --box-shadow: #{$cds-alias-object-shadow-100};
+}
+
+/**
+ * TODO: Have to figure out why is not working the `:not` selector to `cds-divider` component
```
mathisscott: You may have an issue here due to those elements (the `cds-divider` and its containing `div`) being slotted...

mathisscott (changes requested):

bbogdanov:
I have been thinking about adding some additional testing to the component but the only differences between this one and the `panel` component are the CSS properties.

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -15,7 +15,6 @@ export const CdsInternalPanelTagName = 'cds-internal-panel';
  *  // content
  * </cds-internal-panel>
  * ```
```
hippee-lee, line 17: Should this change be reverted?

packages/core/src/styles/layout/layout.stories.ts
```diff
@@ -1774,9 +1775,9 @@ export function patternsContentSiteSingleRail() {
       </div>
       <div class="demo-content">
         <div cds-layout="grid cols@sm:4 gap:lg p:lg p@md:xxl container:xl container:center" cds-text="center">
-          <cds-card cds-layout="p:xl p@md:xxl">card</cds-card>
-          <cds-card cds-layout="p:xl p@md:xxl">card</cds-card>
-          <cds-card cds-layout="p:xl p@md:xxl">card</cds-card>
+          <cds-card> <div cds-layout="p:xl p@md:xxl">card</div> </cds-card>
```
hippee-lee, line 1778: It seems like a card should handle tokens on themselves, is there a reason that an extra div was added?

packages/core/src/card/card.element.scss
```diff
@@ -0,0 +1,11 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --width: auto;
+  --height: auto;
+
+  --background: #{$cds-alias-object-container-background};
+  --border: #{$cds-alias-object-border-width-100} solid #{$cds-alias-object-container-border-color};
```
hippee-lee, line 9: Do we need to offer more granularity here to make it easy for the case where consumerswant to do different things for width/color/lineType but keep the other defaults? E.g:

```html
--border-width: #{$cds-alias-object-border-width-100}
--border-color: #{$cds-alias-object-container-border-color}
--border-line: solid
--border: var(--border-width) var(--border-line) var(--border-color)
```

packages/core/src/card/card.element.ts
```diff
@@ -0,0 +1,53 @@
+import { CdsInternalPanel } from '@cds/core/internal-components/panel';
+import { styles } from './card.element.css.js';
+
+export const CdsCardTagName = 'cds-card';
```
hippee-lee, line 6: Was there a reason for doing this different from the other components? afaik this is the only external component (internal components seem to do it this way) to export the string as a const, the others just use the string directly in `register.ts`.

packages/core/src/card/card.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+    <div class="footer" cds-layout="p:lg">
+      <cds-button> View </cds-button>
+    </div>
+  </div>
+</cds-card>
+```
+
+<Preview>
+  <Story id="card-stories--basic" />
```
hippee-lee: When I run storybook, this demo isn't loading for me on: http://localhost:6006/?path=/story/components-card--page

packages/core/src/card/card.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+</Preview>
+
+## API
+
+<Props of={'cds-card'} />
```
hippee-lee, line 43: Do you know why the slot items show up even though there are no slots for cds-card?

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+    In a dolor faucibus, vehicula ipsum eu, viverra lorem. Aenean congue fermentum commodo.
+    Sed ex purus, consectetur mollis urna eget, vestibulum mollis dolor.
+    Donec sodales, felis sit amet gravida ornare, magna dolor ullamcorper turpis, vel facilisis odio ipsum semper sem.
+  `;
+
+export default {
+  title: 'Stories/Card',
+  component: 'cds-card',
+  argTypes: getElementStorybookArgTypes('cds-card', customElements),
+  parameters: {
```
hippee-lee, line 48: Something may be off with the storybook config. I see this alert at the top of the API Options pane: `This story is not configured to handle controls.` So when I try to adjust the  any of the `API Option` values its not reflected in the UI.

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+    </style>
+    <cds-card id="cds-card-with-fixed-height">
+      <div cds-layout="vertical">
+        <div cds-layout="p:lg">
+          <p cds-text="h4">Card Title</p>
+        </div>
+
+        <cds-divider></cds-divider>
+
+        <div cds-layout="p:lg" cds-text="p1" height="100px">
```
hippee-lee: It looks like this should have a `class="body"` on it to make the example work with overflow scrolling.

packages/core/src/card/card.element.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('Card element: ', () => {
+  let testElement: HTMLElement;
+  let component: CdsCard;
+  const placeholderText = 'Placeholder';
+
+  afterEach(() => {
+    removeTestElement(testElement);
+  });
+
+  describe('create the card component', () => {
```
hippee-lee: This describe isn't needed, move the `beforeEach` up to the top level describe and eliminate this nested one.

packages/core/src/card/card.element.spec.ts
```diff
@@ -0,0 +1,26 @@
+import '@cds/core/card/register.js';
+import { html } from 'lit-html';
+import { componentIsStable, createTestElement, removeTestElement } from '@cds/core/test';
+import { CdsCard } from '@cds/core/card';
+
+describe('Card element: ', () => {
```
hippee-lee: change this to `cds-card`

packages/core/src/card/card.element.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  describe('create the card component', () => {
+    beforeEach(async () => {
+      testElement = await createTestElement(html`<cds-card>${placeholderText}</cds-card>`);
+      component = testElement.querySelector<CdsCard>('cds-card');
+    });
+
+    it('should create the component', async () => {
+      await componentIsStable(component);
+      expect(component.innerText).toBe(placeholderText);
```
hippee-lee: This can be simplified to `expect(component).toBeTruthy();`

If you also want to test the projection a second `it('projects the content', () => {})` could be added.

hippee-lee (changes requested):
I've requested some changes and also have a few open questions about the storybook content. And, I had some confustion around the example html used in `card.element.ts` becasue it looked like the content classes were not used in storybook. It seems like they should be similar to teh html used in the basic story. 

In general it seems like this component leaves layout and typography up to the application side which is good.

packages/core/src/styles/layout/layout.stories.ts
bbogdanov: It should but you have a `private-host` element because the `card` extends the `panel`. The layout will be applied to the `host` instead to the `private-host`.

packages/core/src/card/card.element.scss
bbogdanov: We had a discussion before with Cory about this one and we end up with the conversation here https://github.com/vmware/clarity/pull/5538#discussion_r580341079 

The `--broder-line` was not part of the APi and we didn't thought about it before. I tried to see how `inherit` `unset`, etc will behave and nothing helped to inherit the initial value we set in the component. I believe switching to the API you are proposing will be beneficial. 

@coryrylan Any thoughts?

packages/core/src/card/card.element.ts
bbogdanov: The internal components way seems like a better idea to me cause there is single point of truth for the component selector so that's why I decided to go in that direction. We may do the same for the rest of components. @mathisscott @coryrylan @Shijir @jeeyun, what do you think?

packages/core/src/card/card.stories.mdx
bbogdanov: Yea, I have to switch the `stories` and `card` places in the selector. Thank you for catching that!

packages/core/src/styles/layout/layout.stories.ts
hippee-lee: Does this mean that this won't work for consumers who want to use the element to set up the layout?

```html
<cds-card cds-layout="vertical">
  <div cds-token"align:top"></div>
  <div cds-token"align:bottom"></div>
</cds-card>
```

coryrylan: Layouts have some issue with aligning elements when its projected content. The DOM treats the elements in the component template as elements between  the projected content. Sort of in a way of flattening out the content

```html
<div class="private-host">
  <slot></slot>
</div>
```

```html
<cds-card cds-layout="vertical">
  <div class="private-host"> <!-- the projected items are not direct children of the card so flex/grid items dont work -->
    <div cds-token"align:top"></div>
    <div cds-token"align:bottom"></div>
 </div>
</cds-card>
```

Once CSS Sub grid is widely supported or Safari fixes display:content in a more consistent way we don't have many ways around this.

packages/core/src/card/card.element.scss
coryrylan: We typically avoid splitting out shorthand properties like this primarily for maintenance. By exposing the highest level property possible they can override any number of properties without us maintaining a very large number of css properties.

For example we expose `--background` so developers have access to all shorthand values (size, clip, color, img) without maintaining 8 more custom properties for background or having to go back and add them later. Similar patterns can be found for `--padding` and `--margin`. We found early on it became hard to maintain and document so many property values. It may be more work to re-set defaults but we should encourage use of the theme aliases instead of component level properties when possible.

I do think we could add a new object border token to represent the style value at the system level.

```css
:root {
  --cds-alias-object-border-style: solid;
}

:host {
  --border: #{$cds-alias-object-border-width-100} #{$cds-alias-object-border-style} #{$cds-alias-object-container-border-color};
}
```

hippee-lee (approved):
Based on the conversations and fixed pushed this lgtm. 👍🏼
Thanks for adding this and helping me understand whats going on here.

packages/core/src/card/card.element.ts
mathisscott: This is something @coryrylan and I have discussed a while back. IIRC, the `const` approach you have here is what we decided on as the way forward in the future. I don't see any issue with switching existing components over to use it if you wanted to take that on in a separate PR.

packages/core/src/card/card.stories.mdx
mathisscott: Card extends the panel. The panel has a slot element, iirc.

mathisscott (approved):

packages/core/build/tokens.ts
```diff
@@ -12,6 +12,7 @@ const layout = {
     xs: token(4),
     sm: token(6),
     md: token(12),
+    xmd: token(18),
```
bbogdanov: In order to follow the design spces for the spaces in the card we need to introduce a new t-shirt size. 

We decided to go in this direction cause the other way is to provide CSS styles that you will need to add when your application is using the card component and this will complicate the things a lot.

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -32,13 +31,15 @@ export const CdsInternalPanelTagName = 'cds-internal-panel';
...
+  @id() private panelId: string;
+
   static get styles() {
     return [baseStyles, styles];
   }
 
   render() {
     return html`
-      <div class="private-host">
+      <div class="private-host" role="region" aria-labelledby="${this.panelId}">
```
bbogdanov: @mathisscott We chat about the `role` yesterday but I forgot to ask you about the `aria-labelledby`. I look around in the `core` package and I saw we used `@id()` for aria labels and aria descriptions. Let me know if I missed something here and I have to go in another direction.

mathisscott: The `aria-labelledby` will need to point to a header or other html block element provided by the end user. Likewise, the id will be provided on their end as well. You probably want a property that then projects its value (a string id) down into the private host wrapper.

packages/core/build/tokens.ts
mathisscott: @colinreedmiller 
Why are we already extending our spacing tokens? We did a lot of research on this and I believed we were confident in our design token sizes.

This feels like a slippery slope and we are having to make special cases here just to avoid a pretty major breaking change. 

It's a little disappointing to run across this less than a year into the design tokens.

colinreedmiller: The research into T-shirt sizes did not include design and the result omitted some necessary sizes. I put in the request to make some adjustments as necessary and was informed that though we had performance concerns, and that these adjustments were possible, especially in cases where we were lacking in standard sizes. There seems to be a communication disconnect between where spacing tokens are and where they need to be relative to what design is requesting.

mathisscott: @bbogdanov : I'm going to have to block this for the time being while we get aligned with the UX folks over these spacing sizes. We just need to make sure we are all on the same page here so that we can move forward in the spirit of consistency and long-term stability here. 

Thankfully, this is not a breaking change. You've done well. Maybe we could do with naming this `xm` instead of `xmd` because all the other sizes are just two letters.

But there may be more sizes like this coming our way and we need to understand the scope and reasoning. Right now, we don't know if this will go back to `24px` like the padding in the modal and overlays or remain `18px` and introduce new sizes in different areas.

mathisscott (changes requested):

packages/core/build/tokens.ts
coryrylan: Something to note on this there is a 18px spacing token already in the system `--cds-global-space-8` The system has two sets of tokens. [Space Tokens](https://clarity.design/storybook/core/?path=/story/foundation-spacing--page#component-spacing) and [Layout Tokens](https://clarity.design/storybook/core/?path=/story/foundation-spacing--page#application-layout-spacing). Space provides more values for more specific styles in a component. The layout values are specific to the whitespace between components. This distinction is important because for compact/large touch target themes to work properly they need to be tread as separate scales. Performance wise the layout values are limited due to the amount of utility CSS they generate. Look at [this example](https://clarity.design/storybook/core/?path=/story/themes-dynamic-themes--page) of why the two scales work together. Adjust the scale and density options.

A question on the API design why does the Card not retain its own padding but require wrapper elements around the content? Our other container components (modal, alerts, accordion) allow any layout but still provide the default padding. Without this it would make upgrades and compact/space themes difficult to implement. Example it would not be possible to define a compact theme for card or mobile dynamic sizes since the padding is defined outside the scope of the component.

The panel/card has a `--padding` property already defined.  So the card could set its own default host padding which is whats documented in the [API](https://github.com/vmware/clarity/pull/5785/files#diff-73551c481fbce025eadb1d64dd06c383dbc10713f2d9d1fba59aa69a71275130R46) currently. Rather than introduce a new layout token or require explicit padding each time we could add a attr to stretch items full width.

```scss
// card.element.scss
:host {
  --padding:  var(--cds-global-space-8);
}

::slotted([cds-card-stretch]) {
  margin-left: calc(-1 * var(--padding));
  width: calc(100% + var(--padding));
}
```

```html
<cds-card>
  <h3>heading</h3>
  
  <cds-divider cds-card-stretch></cds-divider>
  <cds-button>...</cds-button>
</cds-card>
```

packages/core/src/internal-components/panel/panel.element.ts
coryrylan: The aria labelledby and role need to exist on the host element. A aria-labelledby in the shadow dom cannot reference an id in the light DOM. This is because `id` values are unique to each. The role can be applied to the host similar to the alert component

```typescript
@property({ type: String }) role = 'region';
```

Then apps can assign the aria labelledby on the host

```html
<cds-card aria-labelledby="title"> <!-- role will be auto set to this element -->
  <h3 id="title"></h3>
</cds-card>
```

packages/core/build/tokens.ts
bbogdanov: @coryrylan We thought about the scenario you are talking about and I documented our thoughts [here](https://app.clubhouse.io/clarity-design-system/story/114/card-component)

The `slotted` selector can only select direct children of the element. In our case the direct child of `cds-card` is the `div.private-host`.

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+  title: 'Stories/Card',
+  component: 'cds-card',
+  argTypes: getElementStorybookArgTypes('cds-card', customElements),
+  parameters: {
+    options: { showPanel: true },
+  },
+};
+
+export function Basic(args: { [key: string]: unknown }) {
+  return html`<cds-card ...="${spreadProps(getElementStorybookArgs(args))}">
```
coryrylan, line 54: The args demo should only be on a single API demo function and not for all stories. This helps storybook performance but also the website not showing storybook code in the demos. [Example](https://github.com/vmware/clarity/blob/next/packages/core/src/button/button.stories.ts)

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export function WithLayout(args: { [key: string]: unknown }) {
+  return html` <cds-card ...="${spreadProps(getElementStorybookArgs(args))}">
+    <div cds-layout="vertical">
+      <div cds-layout="p-t:xmd p-l:xmd">
+        <p cds-text="section">Card Title</p>
+      </div>
+
+      <div cds-layout="p:xmd">
+        <p cds-text="p1">${basicContent}</p>
```
coryrylan: We should update this to use `body` rather than `p1`. The p and h utils are temporary to help migration and will eventually go away. We want to encourage teams to use the semantic utilities when possible.

packages/core/src/card/card.element.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @cssprop --padding
+ * @cssprop --box-shadow
+ */
+export class CdsCard extends CdsInternalPanel {
+  @property({ type: String }) role = 'region';
+
+  connectedCallback() {
+    super.connectedCallback();
+    setAttributes(this, ['role', this.role]);
+  }
```
coryrylan: You can remove this block by adding reflect true to the prop decorator, then it will add the attr for you automatically.

```typescript
@property({ type: String, reflect: true }) role = 'region';
```

packages/core/src/card/card.element.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @cssprop --height
+ * @cssprop --color
+ * @cssprop --overflow
+ * @cssprop --overflow-x
+ * @cssprop --overflow-y
+ * @cssprop --background
+ * @cssprop --border
+ * @cssprop --border-radius
+ * @cssprop --padding
+ * @cssprop --box-shadow
```
mathisscott, line 49: Probably need to note the `--card-remove-margin` here. We also need to verify all the other properties listed. It seems like we have too many overflow properties.

packages/core/src/internal-components/panel/panel.element.ts
mathisscott: I don't think this is what @coryrylan was suggesting...

packages/core/src/internal-components/panel/panel.element.ts
```diff
@@ -32,13 +31,15 @@ export const CdsInternalPanelTagName = 'cds-internal-panel';
  * @cssprop --box-shadow
  */
 export class CdsInternalPanel extends LitElement {
+  @id() private panelId: string;
```
mathisscott: I just fixed an issue with [tabbing into a scrollable panel in the modal](https://github.com/vmware/clarity/pull/5918).

We may want to double-check that isn't a problem here. Put an input or button inside a scrollable panel or card and then see if you can scroll the content with just the keyboard arrow keys. And then see if you can tab through to your button.

mathisscott (changes requested):

packages/core/src/internal-components/panel/panel.element.ts
bbogdanov: Yea I forgot to remove that one. Thank you for reminding me!

bbogdanov: @coryrylan I remember we actually talked about the `role="region"` should be part of the `div.private-host` so isn't it what we want as part of the `panel`?  If this case is the correct one then doesn't this means we have to pass down the `aria-labelledby` attribute that will be set on the `cds-card` to the `div.private-host`?

coryrylan: What you have currently on the PR looks correct. With the role on the host its remains in the light DOM and the a11y is preserved.

```html
<!-- App code -->
<cds-card aria-describedby="heading">
  <h2 id="heading">heading</h2>
</cds-card>

<!-- Rendered output code -->
<cds-card role="region" aria-describedby="heading">
  <h2 id="heading">heading</h2>
</cds-card>
```

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,147 @@
...
+        overflow: auto;
+      }
+    </style>
+    <cds-card id="cds-card-with-fixed-height">
+      <div cds-layout="vertical gap:md">
+        <div>
+          <p cds-text="section">Card Title</p>
+        </div>
+
+        <cds-divider card-remove-margin></cds-divider>
```
coryrylan: We should probably prefix this attribute with `cds-` to to ensure its within a scope of sorts and lowers attr collision risk on elements that are not Core.

```html
cds-card-remove-margin
```

coryrylan (approved):

mathisscott (approved):

packages/core/src/card/card.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+    <cds-divider cds-card-remove-margin></cds-divider>
+
+    <div cds-text="body" cds-layout="p-y:lg">
+      Content
+    </div>
+
+    <cds-divider cds-card-remove-margin></cds-divider>
+
+    <div cds-layout="horizontal gap:sm align:vertical-center">
+      <cds-button action="flat-inline">View</cds-button>
```
coryrylan, line 31: Looking at the demo I think we need to remove the line height on the flat-inline button as well with the height. Its inheriting the parent line height which pushed the text down.

```css
:host([action='flat-inline']) {
  --height: auto;
  --padding: 0;
  line-height: 0;
```

Screen Shot 2021-05-19 at 4 27 36 PM

gnomeontherun:
@bbogdanov I took a look today and I'm not sure the latest push looks quite right with the spacing. I'd also like to request that we make the demos more realistic and not have just _lorem ipsum_ content. Given the granularity and flexibility the card offers, we ought to bring them to life with realistic content to help give people a better impression of outcomes. We would also benefit from adding some alternative layouts or combining multiple cards like we see in applications. We don't need to create every possible example, but it would be helpful to have a handful of them. For example:

* What would a list or grid of cards potentially look like? We often see this with teams building a 'card layout' or datagrid toggle to display a list of items 
* How might a card behave that contains more complex titles, like icons?
* What about a card that has images or more complex content? We have samples of things like lists, alerts, progress indicators, and such.

I know we are focused on the low level card, and perhaps we also need to clarify that in our Storybook documentation. Right now it just shows a sample, without any context. Cards appear to be deceptively simple though we know they can be complicated, largely due to the complexity of content being projected in and if apps will correctly apply the right structure. Providing a few more sample types would help give developers and designers a broader understanding of how to construct them.

colinreedmiller:
I have created an analysis file comparing NG and Core cards here:
https://www.figma.com/file/dYY4HrbIOpDCgmVqoJxsn8/card-CORE?node-id=0%3A1

packages/core/src/card/card.element.scss
```diff
@@ -0,0 +1,14 @@
+@import './../styles/tokens/generated/index';
+
+:host {
+  --width: auto;
+  --height: auto;
```
coryrylan, line 5: After testing in storybook I think `--height` and `--width` need to be `100%` so they fill their container. In multi grid/column layouts they don't fill as expected. Example:

```html
<section cds-layout="grid gap:lg cols@md:6">
  <cds-card  style="--height: 100%">
    <div cds-layout="vertical gap:lg">
      <h3 cds-text="section">Card Title</h3>

      <p cds-text="body">
        Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.  
      </p>

      <p cds-text="body">
        Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.  
      </p>
    </div>
  </cds-card>

  <cds-card style="--height: 100%">
    <div cds-layout="vertical gap:lg">
      <h3 cds-text="section">Card Title</h3>

      <p cds-text="body">
        Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.  
      </p>
    </div>
  </cds-card>
</section>
```

Also it would be nice to see a column layout demo like this on the mdx page to show a bit more with card layouts.

bbogdanov: This is what we wanted in the first place. 

The panel to be able to fit the container, the card to be able to resize depending on its content.

coryrylan: Yeah you're right, I'll try something else see if the cards can fill without adjusting the --height 👍

packages/core/src/card/card.stories.mdx
bbogdanov: https://github.com/vmware/clarity/pull/5970

packages/core/src/card/card.element.scss
bbogdanov: There will be a CSS after all. If you expect every card to take the full width and height of a container, you will be able to simply do 

```
cds-card {
 --height: 100%;
 --width: 100%;
}
```

bbogdanov:
@coryrylan @mathisscott Added a slight change to the CSS. `--width: auto` to `--width: fit-content`. Seems like Firefox has some issues with `fit-content` so I added also one more line `--width: -moz-fit-content` to solve the issue on Firefox.

bbogdanov:
@coryrylan @mathisscott Added a slight change to the CSS. `--width: auto` to `--width: fit-content` cause the content on `auto` was shrinking to fit the container instead of taking the width necessary to fit the content.

On the other hand I see Firefox is having some issues with `fit-content` and I'll check that tomorrow.

Edit: Found that `-moz-fit-content` is a fix.

packages/core/src/card/card.stories.ts
```diff
@@ -0,0 +1,292 @@
...
+}
+
+export function WithContainerOfCards() {
+  return html`
+    <style>
+      cds-card.cds-card__fit {
+        flex: 1;
+      }
+    </style>
+    <div cds-layout="horizontal gap:md">
```
coryrylan: Non blocker: if you change this layout to use grid then the flex style is not needed and it will be responsive.

```html
<div cds-layout="grid cols@md:6 cols@lg:3 gap:md">
```

coryrylan (approved):

PR #5821, opened 2021-04-09
build: adoption guide v1 (Beta)

bdryanovski:
An initial version of the adoption guide v1

* This includes the adoption-guide as its own application as part of apps/
* Snippet and samples for demo.
* Guides on how to migrate some of the components into Core versions.

🧐 Known issues:

Part of the beta release was planned to include Stackblitz or Codesandbox integration for interactive demos - but for the moment both platforms have issues with running the latest release of the Clarity packages - so at the moment this functionality is hidden - more time is needed for resolving this so I don't want to block the Beta release at the moment.

More information on what is not working for interactive demos: 
 * Stackblitz - has issues compiling Clarity v5 and it's throwing NGCC error saying that It could not parse correctly the `*.umd.js` file. - Need investigation why?
 * Codesandbox - It's running fine pure Clarity Angular application - but have a hard issue when it comes to registering Clarity Core components - the error thrown there is that it could not find Stylesheet that must be part of `HTMLElement` - I suspect that some of the web components shims are not working well there. 
So both platforms could run only half of the Clarity packages.

I manage to run both Core and Angular components on separate projects but never together. I don't have any of the above issues when I run them locally.

Some of the examples require additional text and explanation.

Open for ideas of more examples into some of the sections like Alerts, Accordion, Icons, etc.

The next steps will be to add an adoption guide for Modal and Forms.

⚠️ **Important note**: this guide is not linked anywhere into the main documentation - this will be subject to next releases 

❗️ **UPDATE** - I'm enabling again the live embed and running on top Stackblitz and Codesandbox

* Stackblitz: To be able to run demos there I had to downgrade the `@clr/angular` to the latest v4 (4.0.14) but at the same time I'm running `@cds/core` & `@cds/angular` to the latest v5 (5.1.1). This seems to work ok with the latest Angular (v11). Need more investigation why v5 is not working ok with Stackblitz - This change is temporary until some solution is found - the best outcome will be to be able to run v5 for all packages.
* Codesandbox - is working ok with angular but don't want to compile the Core components it seems some local issue with how the shims is implemented.


apps/adoption/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "lint": "ng lint",
+    "e2e": "ng e2e"
+  },
+  "private": true,
+  "browserslist": [
+    "> 0.5%",
+    "last 2 versions",
+    "Firefox ESR",
+    "not dead",
+    "IE 11",
```
mathisscott: we don't support IE11?

apps/adoption/src/app/components/stackblitz/templates/core-latest.ts
```diff
@@ -0,0 +1,130 @@
...
+    '@angular/compiler': '10.2.4',
+    '@angular/core': '10.2.4',
+    '@angular/forms': '10.2.4',
+    '@angular/platform-browser': '10.2.4',
+    '@angular/platform-browser-dynamic': '10.2.4',
+    '@angular/router': '10.2.4',
+    '@cds/angular': '5.0.2',
+    '@cds/core': '5.0.2',
+    '@clr/angular': '4.0.11',
+    '@clr/core': '4.0.11',
```
mathisscott: Why clr-core AND cds-core...? That could result in some weird collisions...

apps/adoption/src/app/migrations/alert/alert.2.core.txt
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
+
+import '@cds/core/alert/register.js';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <cds-alert-group status="danger" type="banner">
+      <cds-alert closable>
+        <cds-icon shape="node-group" aria-hidden="true"></cds-icon>
```
mathisscott: These icons would not be aria-hidden. They need an aria-label, per a11y.

apps/adoption/src/app/pages/getting-started.page.html
```diff
@@ -0,0 +1,34 @@
...
+</p>
+
+<h3>Where to find more information?</h3>
+<p>
+  The guide won't always reflect the latest API changes inside the components so if you want to know more on specific
+  section you could go and check the <a href="https://clarity.design/" target="_blank">official documentation</a>
+</p>
+
+<!--
+<h3>Boilerplates</h3>
```
mathisscott, line 23: Are we keeping this around or should we remove it?

apps/adoption/src/app/templates/angular/styles.css.txt
```diff
@@ -0,0 +1,3 @@
+body {
+  padding: 1rem;
```
mathisscott, line 2: What's up with this text file?

mathisscott (changes requested):

apps/adoption/src/app/components/stackblitz/templates/core-latest.ts
bdryanovski: You are right this probably was left from some of my experiments to make Stackblitz work .. I was trying to make them work with older version - will still have hard time making Stackblitz v5 work without NGCC error

apps/adoption/src/app/pages/getting-started.page.html
bdryanovski: When Stackblitz or Codesandbox start working - here will be the place that you could find "Getting started kit" - A basic working example of Angular App or Core App 

I commented it just because I didn't want to drop the code - I have some deep inside big hopes that this will be a working solution for the moment.

apps/adoption/src/app/templates/angular/styles.css.txt
bdryanovski: This is for one reason only - presentation. 

When you use the embed option for Stackblitz or Codesandbox and even when you open them in their own interfaces - standalone components are displayed stuck to top-left corner so sometimes even their own border is hard to see. So this `padding: 1rem` is just giving some space so whatever we try to demo is more clearly visible - in demo mode.

apps/adoption/src/app/components/demo.component.ts
```diff
@@ -0,0 +1,182 @@
...
+  // Experimental
+  @Input('sideBySide') _sideBySideState: boolean | string = false;
+
+  constructor(private stackblitz: StackblitzService, private codesandbox: CodesandboxService) {}
+
+  ngOnInit(): void {
+    if (this.tabs) {
+      // auto attach uniq id for every tab - later use for embeding the preview
+      this.tabs = this.tabs.map(tab => {
+        // Use the first file name to generate hash - it must be always uniq name
```
sis0k0:
```suggestion
        // Use the first file name to generate hash - it must always be unique name
```

apps/adoption/src/app/components/demo.component.ts
```diff
@@ -0,0 +1,182 @@
...
+  @Input('tabs') tabs: DemoTabData[] = [];
+
+  // Experimental
+  @Input('sideBySide') _sideBySideState: boolean | string = false;
+
+  constructor(private stackblitz: StackblitzService, private codesandbox: CodesandboxService) {}
+
+  ngOnInit(): void {
+    if (this.tabs) {
+      // auto attach uniq id for every tab - later use for embeding the preview
```
sis0k0:
```suggestion
      // auto attach unique id for every tab - later used for embedding the preview
```

apps/adoption/src/app/components/demo.component.ts
```diff
@@ -0,0 +1,182 @@
...
+      this.tabs = this.tabs.map(tab => {
+        // Use the first file name to generate hash - it must be always uniq name
+        // or at least uniq for the page.
```
sis0k0:
```suggestion
        // or at least unique for the page.
```

apps/adoption/src/app/components/demo.component.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  @Input('tabs') tabs: DemoTabData[] = [];
+
+  // Experimental
+  @Input('sideBySide') _sideBySideState: boolean | string = false;
+
+  constructor(private stackblitz: StackblitzService, private codesandbox: CodesandboxService) {}
+
+  ngOnInit(): void {
+    if (this.tabs) {
```
sis0k0:
```suggestion
    if (this.tabs && this.tabs.length) {
```

apps/adoption/src/app/components/demo.component.ts
```diff
@@ -0,0 +1,182 @@
...
+  public async openCodesandbox(tab: DemoTabData): Promise<void> {
+    this.codesandbox.open(tab.template, tab.files);
+  }
+
+  public sideBySide(): void {
+    this._sideBySideState = !this._sideBySideState;
+  }
+
+  /**
+   * Side by side only work where are two and exactly two tabs defined.
```
sis0k0:
```suggestion
   * Side by side only works where there are exactly two tabs defined.
```

sis0k0:
I get the following error when trying to open the code samples in `Codesandbox`:

```
TypeError
Cannot read property 'styleSheet' of undefined
```

Jinnie:
> I get the following error when trying to open the code samples in `Codesandbox`:
> 
> ```
> TypeError
> Cannot read property 'styleSheet' of undefined
> ```

I've seen this too, trying to run Core in CodeSandbox. Manually. 
As, currently we're not making this public, I will consider this issue with CodeSandbox, not in our own code. And we'll need to work on it, if needed contact them, log a bug, etc. If we manage to overcome the problems with Stackblitz, we can drop CodeSandbox. But now we're in a state where Core works better in Stackblitz, while Angular is better in CodeSandbox.

apps/adoption/src/app/components/demo.component.ts
bdryanovski: Basically this is handle below but a better check may be will be 
```
if (Array.isArray(this.tabs)) { 
```

`this.tabs.length` won't prevent `String` from passing

Jinnie (approved):

sis0k0 (approved):

bdryanovski:
> > I get the following error when trying to open the code samples in `Codesandbox`:
> > ```
> > TypeError
> > Cannot read property 'styleSheet' of undefined
> > ```
> 
> I've seen this too, trying to run Core in CodeSandbox. Manually.
> As, currently we're not making this public, I will consider this issue with CodeSandbox, not in our own code. And we'll need to work on it, if needed contact them, log a bug, etc. If we manage to overcome the problems with Stackblitz, we can drop CodeSandbox. But now we're in a state where Core works better in Stackblitz, while Angular is better in CodeSandbox.

I left Codesandbox and added the `beta` badge for one reason only. Codesandbox works a lot better with building and creating demos. But have some issues with the core and angular package in one place. Not sure why? It seems for some strange reason that some Shims for styles are missing or overwritten by something. 

If you try to run pure core on a single HTML page I think I manage to make it work - but when you try to add something else to the mix like angular it stops working ?!

For Stackblitz ... it's running with older angular v4 (latest v4) and latest Core package a bit of a mix of both. But it works. So I'm suspecting that there is something that we miss or need to change so the v5 package will work again. Maybe we compiled something differently or maybe some attribute is missing from the `package.json`. I made a quick diff of some of the files and did not find any changes that could result in this.

Both of them are for experimental purpose and trying to get feedback so feel free to test them/ break them and so on.

apps/adoption/src/app/migrations/alert/alert.2.core.txt
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
+
+import '@cds/core/alert/register.js';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <cds-alert-group status="danger" type="banner">
+      <cds-alert closable>
+        <cds-icon shape="node-group" aria-label="Close alert"></cds-icon>
```
mathisscott: It seems like this aria label should be something like `Example alert`. Calling it `Close alert` would confuse people using screen readers. Because they would think they could click on it to close the alert.

apps/adoption/src/app/templates/angular/app.module.ts.txt
```diff
@@ -0,0 +1,22 @@
+import { NgModule } from "@angular/core";
+import { BrowserModule } from "@angular/platform-browser";
+import { BrowserAnimationsModule } from "@angular/platform-browser/animations";
+import { ClarityModule } from "@clr/angular";
+
+import { CdsModule } from "@cds/angular";
+
+import { AppComponent } from "./app.component";
+
+import '@clr/icons';
```
mathisscott, line 13: Do we want people importing the old clarity icons here?

apps/adoption/src/app/templates/core/app.module.ts.txt
```diff
@@ -0,0 +1,22 @@
+import { NgModule } from "@angular/core";
+import { BrowserModule } from "@angular/platform-browser";
+import { BrowserAnimationsModule } from "@angular/platform-browser/animations";
+import { ClarityModule } from "@clr/angular";
+
+import { CdsModule } from "@cds/angular";
+
+import { AppComponent } from "./app.component";
+
+import '@clr/icons';
```
mathisscott: Same question here

mathisscott (changes requested):

apps/adoption/src/app/templates/angular/app.module.ts.txt
bdryanovski: You are right here - we don't need to include the old icons. But because we are running v4 at the moment - Stackblitz is not working well with v5 at the moment - all the changes with the icons are not available at the moment. 

So when we update to v5 packages this line will be reworked to use `core icons`. 

Without this line for example this demo: 

```ts
import { Component } from "@angular/core";

@Component({
  selector: "app-root",
  template: `
    <clr-alert [clrAlertType]="'danger'" [clrAlertAppLevel]="true">
      <clr-alert-item>
        <span class="alert-text">
          Proin varius lacinia tincidunt. In elementum suscipit lectus eget
          hendrerit. Vestibulum rutrum risus ut turpis interdum, ac auctor nisi
          ultrices
        </span>
      </clr-alert-item>
    </clr-alert>
  `
})
export class AppComponent {}
```

Will render like that

Screenshot 2021-04-22 at 16 51 44

But this is a good point and must be documented when Stackbliz unblocks us.

apps/adoption/src/app/templates/core/app.module.ts.txt
bdryanovski: Here we could drop this - but in cases when we try to mix some old v4 components and core - we gonna miss some icons.

mathisscott (approved):

gnomeontherun:
@bdryanovski should this be closed?

bdryanovski:
Some build issues on the CI need to be resolved before merging but yes ;)

PR #5866, opened 2021-04-17
Adjust colors of disabled inputs

astorije:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the before/after behavior?

The bottom border of disabled inputs was actually darker than the bottom border of non-disabled inputs, while labels and control messages were ligther:

Before | After
--- | ---
disabled border before | disabled border after

The icon color of the `cds-search` element was not using a disabled color when the input was disabled:

Before | After
--- | ---
disabled search icon before | disabled search icon after

The text value of a disabled input was using the same color than when not disabled, giving the impression it was still editable:

Before | After
--- | ---
disabled input text color before | disabled input text color after

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/input/input.element.scss
+++ b/packages/core/src/input/input.element.scss
@@ -108,4 +108,5 @@
 
 :host([_disabled]) {
   --border-color: #{$cds-alias-object-border-color-tint};
+  --color: #{$cds-alias-status-disabled};
 }
--- a/packages/core/src/input/input.element.scss
+++ b/packages/core/src/input/input.element.scss
@@ -107,5 +107,5 @@
 }
 
 :host([_disabled]) {
-  --border-color: #{$cds-alias-status-disabled};
+  --border-color: #{$cds-alias-object-border-color-tint};
 }
--- a/packages/core/src/search/search.element.scss
+++ b/packages/core/src/search/search.element.scss
@@ -1,0 +1,9 @@
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+// This software is released under MIT license.
+// The full license information can be found in LICENSE in the root directory of this project.
+
+@import './../styles/tokens/generated/index';
+
+:host([_disabled]) .icon {
+  --color: #{$cds-alias-status-disabled};
+}
--- a/packages/core/src/search/search.element.ts
+++ b/packages/core/src/search/search.element.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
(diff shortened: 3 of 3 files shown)


packages/core/src/input/input.element.scss
```diff
@@ -107,5 +107,6 @@
 }
 
 :host([_disabled]) {
-  --border-color: #{$cds-alias-status-disabled};
+  --border-color: #{$cds-alias-object-border-color-tint};
```
mathisscott, line 110: Could we make the border color here the same as the text color on the next line? `$cds-alias-status-disabled`

Right now, the text is of a disabled input is darker than the border and it looks inconsistent.

mathisscott (changes requested):

packages/core/src/input/input.element.scss
astorije: So effectively reverting this change, right?

That would make the bottom border when disabled actually darker than when not disabled.

For comparison, this is what it looks with my PR (`--cds-alias-object-border-color` when enabled, `--cds-alias-object-border-color-tint` when disabled):

Screen Shot 2021-04-19 at 11 56 01 PM

This is what it looks reverted (`--cds-alias-object-border-color` when enabled, `--cds-alias-status-disabled` when disabled):

Screen Shot 2021-04-19 at 11 58 52 PM

And for reference, this is how it currently looks (i.e. without this PR) on a real-world use case, you can see the border of the disabled search input looks more prominent than the non-disabled select box, since `--cds-alias-status-disabled` is darker than `--cds-alias-object-border-color`:

Screen Shot 2021-04-20 at 12 04 39 AM

I think this looks wrong: the search border should definitely be darker than the select border.

What do you think?

mathisscott:
```
I think this looks wrong: the search border should definitely be darker than the select border.

What do you think?
```

I see what you are saying. I say keep it as you have it now. But I think this has uncovered a deeper visual issue in general that we won't solve right here and right now.

Pinging @colinreedmiller so that this is on his radar for future consideration.

mathisscott (approved):

hippee-lee (approved):

hippee-lee (changes requested):
Hey @astorije - this looks good and is ready to go. Can you squash and rebase if needed so that the commit history is clean?

astorije:
I had actually split this into 3 commits on purpose so that each commit solves a specific thing and makes for a more appropriate commit history, as demonstrated by the screenshots in the PR description. Do you still want me to squash them?

hippee-lee:
Ok, that makes sense. I'll unblock and merge. Thanks for fixing this!

hippee-lee (approved):

astorije:
Thanks!!

PR #5876, opened 2021-04-20
feat(breadcrumb): add breadcrumb component

derkoe:
- Add new core component cds-breadcrumb
- Add 'breadcrumb' to commit-lint scopes

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?

This adds a new core component for breadcrumbs. 

Here is the Figma design for the component: 
https://www.figma.com/file/lyq8iK32X3nlbHXTb0z9Bk/Breadcrumb?node-id=1%3A664

- [ ] Yes
- [x] No

## Other information

This implementation was done in coordination with the Clarity team.

### DIFF
--- a/packages/core/src/breadcrumb/breadcrumb.element.ts
+++ b/packages/core/src/breadcrumb/breadcrumb.element.ts
@@ -1,0 +1,80 @@
+import { baseStyles, getCssPropertyValue, internalProperty } from '@cds/core/internal';
+import { html, LitElement } from 'lit-element';
+import { styles } from './breadcrumb.element.css.js';
+
+export const CdsBreadcrumbTagName = 'cds-breadcrumb';
+
+interface ItemEntry {
+  element: Element;
+  separator: Element;
+}
+
+/**
+ * ```typescript
+ * import '@cds/core/breadcrumb/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-breadcrumb aria-label="breadcrumb">
+ *   <a href="/home" cds-text="link">Home</a>
+ *   <a href="/parent" cds-text="link">Parent page</a>
+ *   <span aria-current="page">Current page</span>
+ * </cds-breadcrumb>
+ * ```
+ *
+ * @beta
+ * @element cds-breadcrumb
+ * @cssprop --font-size
+ * @cssprop --font-weight
+ * @cssprop --letter-spacing
+ * @cssprop --gap - Gap between items. Valid values: xxs, xs, sm, md, lg, xl, xxl (default: sm)
+ * @cssprop --color
+ */
+export class CdsBreadcrumb extends LitElement {
+  @internalProperty({ type: String, reflect: true, attribute: 'role' })
+  protected role = 'navigation';
+
+  get gap() {
+    const gapValue = getCssPropertyValue('--gap', this);
+    return gapValue ? gapValue : 'sm';
+  }
+
+  render() {
+    return html`
+      <div class="private-host">
+        <ol cds-layout="horizontal gap:${this.gap} align:vertical-center">
+          ${this.getItems().map(
+            item =>
+              html`<li cds-layout="horizontal gap:${this.gap} align:vertical-center">
+                ${item.element} ${item.separator}
+              </li>`
+          )}
+        </ol>
+      </div>
+    `;
+  }
+
+  getItems(): ItemEntry[] {
+    const customSeparator = Array.from(this.children).find(c => c.getAttribute('slot') === 'cds-separator');
+    return Array.from(this.children)
+      .filter(c => !(c.getAttribute('slot') || '').startsWith('cds-separator'))
+      .map(element => {
+        let separator: Element;
+        if (customSeparator) {
+          separator = customSeparator.cloneNode(true) as Element;
+          separator.setAttribute('part', 'separator');
+          separator.removeAttribute('slot');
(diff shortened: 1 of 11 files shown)


packages/core/src/breadcrumb/breadcrumb.element.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @cssprop --gap - Gap between items. Valid values: xxs, xs, sm, md, lg, xl, xxl (default: sm)
+ * @cssprop --color
+ */
+export class CdsBreadcrumb extends LitElement {
+  @internalProperty({ type: String, reflect: true, attribute: 'role' })
+  protected role = 'navigation';
+
+  get gap() {
+    const gapValue = getCssPropertyValue('--gap', this);
+    return gapValue ? gapValue : 'sm';
```
mathisscott: Using `gap` like this here is clever. But how do we know we are going to catch dynamic changes to the custom property and run an update? It would be great to see a test on this. The test would look something like...

1. set up a breadcrumb
2. add a classname to it or add a style value to it that changes `--gap` 
3. verify that the `ol` and `li` down below have their `gap:*` values updated in `cds-layout`

This is a bonus, imo. I think people changing the gap after the component is rendered is an edge case. It would be painful (relatively speaking) to update it on a window resize. And I can't think of any other immediate edge.

Still, it's a good to know sort of thing. If you have time, a test validating it works would be awesome. If not, could you add a `// TODO: ` here so we can remember to be aware?

mathisscott (changes requested):

packages/core/src/breadcrumb/breadcrumb.element.ts
derkoe: The gap does not change when changing the custom CSS property - @jeeyun any idea how to change this?

jeeyun: @mathisscott brings up a valid point. I thought about this some more and think we should get rid of the `--gap`.  Since custom properties are commonly used for dynamic changes, I think using it in a way that violates user expectation is not ideal. 

An application could easily add more padding to a custom separator to achieve similar results:

```
<cds-breadcrumb aria-label="breadcrumb">
  <span cds-layout="p-x:lg" slot="cds-separator">🚘</span>
  <a href="#" cds-text="link">Home</a>
  <a href="#" cds-text="link">Parent page</a>
  <span aria-current="page">Current page</span>
</cds-breadcrumb>
 ```
 
Above code rendered: 
Screen Shot 2021-04-27 at 4 25 39 PM

Comment (derkoe) [reply packages/core/src/breadcrumb/breadcrumb.element.ts]: Removed the gap CSS property and adapted the example in the stories

File: packages/core/src/breadcrumb/breadcrumb.element.ts [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+                ${item.element} ${item.separator}
+              </li>`
+          )}
+        </ol>
+      </div>
+    `;
+  }
+
+  getItems(): ItemEntry[] {
+    const customSeparator = Array.from(this.children).find(c => c.getAttribute('slot') === 'cds-separator');
```
mathisscott, line 52: @jeeyun : Do you think this work here could be simplified by using `@querySlot` or `@querySlotAll`?

mathisscott (changes requested):

packages/core/src/breadcrumb/breadcrumb.element.ts
jeeyun: Yes, we should be using `@querySlot` here. Thank you for catching that.

I just realized that the code for rendering the items and separators has become more imperative. We try to make our templates to be as declarative as possible, to see more easily the markup that you'd see in the rendered version. Here are my proposed changes:

```
export class CdsBreadcrumb extends LitElement {
  ...
  // add the following (querySlot can be imported from '@cds/core/internal')
  @querySlot('[slot="cds-separator"]') private customSeparator: HTMLElement;

 render() {
    return html`
      <div class="private-host">
        <ol cds-layout="horizontal gap:sm align:vertical-center">
          ${this.getItems().map(
            item => html`
              <li cds-layout="horizontal gap:sm align:vertical-center">
                <slot name="${item.getAttribute("slot")}"></slot>
                <span part="separator" aria-hidden="true">
                  ${this.separator}
                </span>
              </li>
            `
          )}
        </ol>
      </div>
    `;
  }

  get separator() {
    let separator: Element;
    if (this.customSeparator) {
      separator = this.customSeparator.cloneNode(true) as Element;
      separator.removeAttribute('slot');
    } 
    return html`${separator? separator : html`/`}`;
  }

  getItems(): Element[] {
    return Array.from(this.children)
      .filter(c => c.getAttribute('slot') !== 'cds-separator')
      .map(element => {
        element.setAttribute("slot", createId());
        return element;
      });
  }
  ...
}
```

jeeyun:
[Porsche Informatic](https://github.com/porscheinformatik) has asked us to take over and wrap up the last 10%. Opening a new PR (https://github.com/vmware/clarity/pull/5940) and closing this one. Will add co-authoring information appropriately in the new PR. Thank you for this valuable contribution!

derkoe:
Thanks @jeeyun - sorry that I couldn't find time to fix this myself.

PR #5885, opened 2021-04-21
build: visual diff version 2

bdryanovski:
Fully reworking the VisualDiff tool that we had.

Trying to base the internal API as much as possible on generic spec-oriented testing.

**Current specs at the moment serve more as examples that a stable test for production** - located into `tools/visual-tester/specs/*`

For more information check `tools/visual-tester/README.md`

Known issues are ShadowDOM elements are hard to modify at runtime - so Icons, progress bar, spinners could result as failing tests.

`describe` and `it` could not be focused but could be ignored if needed.

```js
fdescribe('description', () => {})
// will work the same way as normal 
describe('description', () => {})
// but ignored blocks won't be executed at all
xdescribe('description', () => {}) 

// Same is valid for 
it('test', () => {})
fit('test', () => {})
xit('test', () => {})
```
This will be fixed in later PR.

Snapshots are not part of the tracked files from Git. (and they must stay this way)

An example of a writing test could be found in `tools/visual-tester/specs` or in the `README.md`

This is the initial PR (one of many)


bdryanovski:
Update: 

Provide a way to run Visual Diff code with Jest or Jasmine runners if needed. 

From my test, this is the result.

JavaScript version
Screenshot 2021-04-26 at 16 47 29
TypeScript version
Screenshot 2021-04-26 at 16 24 13


Compared to the basic simple runner (from scratch)
Screenshot 2021-04-26 at 16 25 32

It's around 2x/3x times slower when jest or other runner is used (suspect that jest is doing a lot more before start running) - Or I'm doing something wrong?

Benefit from running `jest` that I'm getting TypeScript support (not needing at the moment) and Watch mode that is useful when making changes on spec files. 

Also reporting with `jest` is a bit cryptic probably need to see how Jest Error object is structure and implement it. 

Screenshot 2021-04-26 at 16 20 17


📝 EDIT: for more information check `README.md`

gnomeontherun:
Sorry I'm a little lazy but not clear on how to run this? Can you add specific commands how to execute tests?

bdryanovski:
There are two options: running with the small and basic runner (missing functionality like watching files and auto rerunning tests for example or missing support for TypeScript) or using jest as the test runner. 

### Basic runner:
```bash
yarn install && yarn build && yarn link

vdiff --specs="./specs/**/*.vspec.js"
```

To overwrite base images in tests
```bash
UPDATE_SNAPSHOTS=true vdiff --specs="./specs/**/*.vspec.js"
```

### Jest runner:
```bash
yarn install

jest --testMatch="**/__tests__/**/*.vspec.ts"
```

All tests are located inside the `./specs` folder. Files inside `__tests__` are just importing from `./specs` - the reason for that was that I suspected that I may require some tweaking of some of the Jest setup methods that will not apply for the basic runner. At the moment I did not found any case for that - so this could be dropped at some point.

## Creating a new spec

Create a file and place it inside the `./specs` folder.

```js
// specs/<url>.vspecs.js

const VisualDiffContext = require('../dist/visual/visual-diff');

const TARGET_URL = process.env.TARGET_URL || 'https://localhost';
const UPDATE_SNAPSHOTS = process.env.UPDATE_SNAPSHOTS || false;

describe(`Testing ${TARGET_URL}`, () => {
  let browser, page;

  beforeAll(async () => {
    // Create new visual diff context
    VisualDiff = new VisualDiffContext('<url>.vspecs.js', { updateSnapshots: UPDATE_SNAPSHOTS });

    // prepare visual diff folders and setup the context
    await VisualDiff.prepare();

    browser = await VisualDiff.createBrowser();
  });

  afterAll(async () => {
    // Don't leave open browser after spec is finish
    await browser.close();
  });

  afterEach(async () => {
    // Make sure that after the test is finish we gonna close the page
    await page.close();
  });

  it('goto URL and create snapshot of the page', async () => {
    page = await VisualDiff.goto(`${TARGET_URL}/home-page`, browser);

    const screenshot = await VisualDiff.screenshot(
      {
        name: 'snapshot-picture-name',
        selector: 'main.content',
      },
      page    // page object
    );

    await VisualDiff.expectToMatchBase(screenshot);
  });
});
```

More details could be found here - [`./tools/visual-tester/README.md`](https://github.com/vmware/clarity/blob/51187cbac051100ecc587257ec6d18703d9a330c/tools/visual-tester/README.md) (link to commit hash)

Will try to keep it updated as much as possible over time.

bdryanovski:
Some questions that @Jinnie had so here is a summary of them:

### What to do to run it?

```bash
yarn install 
yarn build
yarn link 

vdiff

// or 

jest
```
Something strange that is happening is that `yarn link` did not work for him, so maybe there is some issue with that - need to investigate it.

Workaround this is to run it like that `node ./dist/runner/cli.js` instead of `vdiff` everything else is the same.

Remember that `tools` are not part of the monorepo so `yarn` won't install anything - you need to do this for yourself.

#### How to create base images ?

Base images are created on the first run - and the first run of new `it` (tests) if images with the same name are not found.

The visual diff flow is like that:

  * First run will create the base image for every test 
  * If an image with the same name is found it will try to compare it with the current snapshot
  * The second run will compare current images with the base images and return diff images
  
#### Some tests are failing - around 40th - why?
Right now some tests will fail when comes to the `dev.vspec` - the reason for that could be one of two 

  * Missing icon - on some places we did not import correctly the icons so we are getting these "flying dots" for the icon and this will result in a mismatch of images.
  * We could not prevent any animations at the moment inside core components - so things like progress-bar, icons, or any other continuous animations won't be prevented from running so we are not sure when exactly to make the screenshot so we won't get a mismatch. 

This could be resolved by fixing the missing icons inside the dev-app and preventing animations inside the core components - @mathisscott mention that there is a flag for that - didn't tested it yet.

#### Why we don't keep the base images as part of the repository

The base images are not part of the git tracked files for few reasons

  * Github actions - where the primary use of the diff tools will be is stateless so storing images there is impossible. 
  * We don't want to mess with the git repository and create more tracked objects inside that will only lead to more garbage collection cycles. Also, images are big and Git will store a full copy of every one of them - over time this will make the repo huge without giving us any benefit by the ability to go back in time and compare old images.
  * Committing and tracking binary files is hard and there is no guarantee that the images that you have are correct - we had a lot of cases over the year that wrong images were committed.
 
The idea here is to compare our current changes against our released versions - and ring a bell before is too late. 

This won't prevent us to run against multiple branches and for example, compare changes between v2 and v3 - but this will be something that the person that needs to know it will have to do for himself.

#### When this will be run.

Initial idea is to be run nightly as part of Github Schedule action and compare next-branch with the already released version of the code - if something did not match to notify us. Without preventing us from working or blocking PR. 

As part of these failed nightly builds we will get a report with all the commands and steps needed to reproduce this locally and check for ourselves. 

Again this is more of a Notification approach than blocking you from doing something. 
This of course could be a good indication of "what's new" between branches/versions.

#### If I want to run it do I need to make my own base images?

The quick answer is yes - you need to run it at least once to create base images against a reference provider - next branch or latest release for example `v5.0.12`


Some more documentation will be needed to provide a more clear view of how things work. Also, everything here is a `beta` version so it could change in the next PR or things above may not be applicable next week. So for that reason will try to keep everything at once place and this will be the `README.md` file into the `tools/visual-tester` folder.

Jinnie (approved):
To make this usable, we also need to fix all icon imports in the dev-app, because now we have many instances of animated ellipsis icons, which fail test, because of the animtion.
But this can be added as a separate task.

tools/visual-tester/src/runner/cli.ts
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+const { run, defaultOptions } = require('./index');
+const packageJson = require('../../package.json');
+
+const argv = require('minimist')(process.argv.slice(2), {
+  default: defaultOptions,
+});
+
+const banner = `
+██╗   ██╗██╗███████╗██╗   ██╗ █████╗ ██╗     ██████╗ ██╗███████╗███████╗
```
bbogdanov, line 10: I like that one!

bbogdanov (approved):

Issue #5916, opened 2021-04-28
@cds/angular not playing nice with element bindings


gnomeontherun:
## Describe the bug

The element bindings and attributes are not getting reflected correctly by the @cds/angular wrappers to the core elements.

## How to reproduce

https://stackblitz.com/edit/angular-ivy-d3mqge?file=src%2Fapp%2Fapp.component.html

1. Start focus on the first link
2. Hit tab
3. Expect focus to move to second link, because tag is supposed to be `readonly`.
4. Change from using pure Core in the AppModule to using CdsModule.
5. Start focus again on first link
6. Hit tab
7. Observe focus hits tag. 
8. Review the element and see that it does not reflect the readonly state properly into the web component.

## Expected behavior

That bindings will work 1:1.

## Versions

**Clarity project:**

- [X] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [X] v5.x

**Framework:**

- [X] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: [e.g. MacBook]

## Additional notes

_Add any other notes about the problem here._

Shijir:
Resolved by https://github.com/vmware/clarity/pull/5932

PR #5932, opened 2021-05-07
fix: empty value boolean properties in @cds/angular

Shijir:
## PR Checklist

## PR Type

Previously, the component with boolean properties were getting falsy due to the following:

```ts
@Input() set readonly(value) {
    this.element['readonly'] = value;
}
```

And empty attributes would be translated as `falsy`. So this PR fixes that with the following:

```ts
@Input() set readonly(value: any) {
    if (value === '') {
      this.element['readonly'] = true;
      return;
    }
    this.element['readonly'] = value;
}
```

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #5916 

- [ ] Yes
- [x] No

### DIFF
--- a/apps/core-angular-cli/src/app/app.component.ts
+++ b/apps/core-angular-cli/src/app/app.component.ts
@@ -73,7 +73,7 @@ export class AppComponent {
     return this.form.controls.password.touched && this.form.controls.password.hasError('minlength');
   }
 
-  expandedChange(event): void {
+  expandedChange(event: CustomEvent): void {
     this.panel1Expanded = event.detail;
   }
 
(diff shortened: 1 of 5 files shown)


apps/core-angular-cli/src/app/app.component.ts
```diff
@@ -73,7 +73,7 @@ export class AppComponent {
     return this.form.controls.password.touched && this.form.controls.password.hasError('minlength');
   }
 
-  expandedChange(event): void {
+  expandedChange(event: any): void {
```
mathisscott: I don't think this needs to be an `any` type. `CustomEvent` should work here as the type.

mathisscott (changes requested):

mathisscott (approved):

apps/core-angular-cli/src/app/app.component.ts
Shijir: Gotcha. I've made use of it in the event emitter types.

bbogdanov (approved):

PR #5951, opened 2021-05-14
fix(dropdownmenu): keyboard accessibility issue when there is a div t…

SaiSatwik:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Keyboard navigation with down arrow button is not working on clarity dropdown options when there is a wrapper div around dropdown items inside clarity dropdown menu. This issue is because @ContentChildren is not detecting the content with div tag in between with Angular9 and above.

Issue Number: N/A

## What is the new behavior?

Adding { descendants: true } on Content Children to get the nested items. With this change keyboard accessibility is working and focus is moving to first dropdown option on triggering dropdown menu even in the presence of nested div around dropdown items. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/popover/dropdown/dropdown-menu.ts
+++ b/packages/angular/projects/clr-angular/src/popover/dropdown/dropdown-menu.ts
@@ -104,7 +104,7 @@ export class ClrDropdownMenu extends AbstractPopover implements AfterContentInit
   }
 
   private focusHandler: DropdownFocusHandler;
-  @ContentChildren(FocusableItem) items: QueryList<FocusableItem>;
+  @ContentChildren(FocusableItem, { descendants: true }) items: QueryList<FocusableItem>;
 
   ngAfterContentInit() {
     this.focusHandler.container = this.el.nativeElement;


bdryanovski (approved):

Jinnie (changes requested):
Thanks for the contribution, first!
Using `{descendants: true}` seems to be the correct way to fix this behavior at first, but I have some memories that we've tried it and it may be leading to serious issues with the nested menus. Can you please verify the scenario?

jeeyun:
> Using `{descendants: true}` seems to be the correct way to fix this behavior at first, but I have some memories that we've tried it and it may be leading to serious issues with the nested menus. Can you please verify the scenario?

I just tested real quick and this solution does not work for the reason @Jinnie mentioned. When we have nested menu with submenus that are eagerly loaded (e.g. without using `clrIfOpen`), using `{descendants: true}` has effect of including all the descendents in the keyboard navigation when using up/down arrow.

SaiSatwik:
Hi @Jinnie/ @jeeyun, then is there any workaround for this. We have div tag inside clr-drodown-menu  in vmware nsx-t codebase. We have used it for some styling purpose for the dropdown. This used to work for us with clarity version 2.4.0 and angular 8.2.2. Keyboard accessibility broken when we upgraded to clarity version 4.0.3. This is an serious issue for us as we have used div tags at many places inside clr-dropdown-menu.

jeeyun:
Hi @SaiSatwik,
I would suggest getting rid of the wrapper div and modify your css customization in the application side. 

Here's a modified fork of your initial example to apply customization without using the wrapper div: https://stackblitz.com/edit/clarity-light-theme-playground-n2fixc?file=src%2Fapp%2Fapp.component.html

hippee-lee:
Closing this for now due to the work around.

PR #5958, opened 2021-05-18
fix: core dynamic rem calc

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

The current spacing tokens in Core rely on a base 20 font size. This was done as a calculation during the token build step. When converting a px value to rem the browser had to re-compute the base rem size which is very expensive in the pxToRem function. Also due to the rem calculation happening at build time the base font was not easily changed dynamically in the browser. This is a problem when dynamically switching themes.

Issue Number: N/A

## What is the new behavior?
This PR updates the tokens to use a CSS calc the create a rem based on a base font variable. This allows the base font to be any value not just 20 or 16. Our components will automatically scale appropriately and scale with the browser font size settings. 

Before:
```css
:root {
  --cds-global-layout-space-xxs: 0.10rem;
  --cds-global-layout-space-xs: 0.20rem;
  --cds-global-layout-space-sm: 0.30rem;
  --cds-global-layout-space-md: 0.60rem;
  --cds-global-layout-space-lg: 1.20rem;
  --cds-global-layout-space-xl: 2.40rem;
  --cds-global-layout-space-xxl: 4.80rem;
}
```

After:
```css
:root {
  --cds-global-layout-space-xxs: calc((2 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-xs: calc((4 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-sm: calc((6 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-md: calc((12 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-lg: calc((24 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-xl: calc((48 / var(--cds-global-base, 20)) * 1rem);
  --cds-global-layout-space-xxl: calc((96 / var(--cds-global-base, 20)) * 1rem);
}

/* base 16 font simply set the base var */
:root {
  --cds-global-base: 16;
}
```

This change also improves the performance of the `pxToRem` function. The function no longer needs to call
`window.getComputedStyle(document.body, null).getPropertyValue('--cds-global-typography-base-font-size')` which is a expensive CSS recalc and causes performance issues for components that iterate often (icons, grid rows, nodes).

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/build/tokens.ts
+++ b/packages/core/build/tokens.ts
@@ -17,14 +17,14 @@ const layout = {
     xxl: token(96),
   },
   grid: {
-    cols: token(12, { absolute: true }),
+    cols: token(12, { static: true }),
   },
   width: {
-    xs: token(576),
-    sm: token(768),
-    md: token(992),
-    lg: token(1200),
-    xl: token(1440),
+    xs: token('576px', { static: true }),
+    sm: token('768px', { static: true }),
+    md: token('992px', { static: true }),
+    lg: token('1200px', { static: true }),
+    xl: token('1440px', { static: true }),
   },
 };
 
@@ -340,8 +340,8 @@ const typography = {
     8: token(32),
     9: token(40),
   },
-  baseFontSize: token('125%'),
-  baseFontSizePx: token(20),
+  baseFontSize: token('125%'), // deprecated for removal in 6.0
+  baseFontSizePx: token(20), // deprecated for removal in 6.0
   fontFamily: token("'Clarity City', 'Avenir Next', sans-serif"),
   headerFontFamily: token("'Clarity City', 'Avenir Next', sans-serif"),
   topGapHeight: token('0.1475em'), // line-height eraser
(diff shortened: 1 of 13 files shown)


packages/core/build/tokens.ts
```diff
@@ -17,14 +17,14 @@ const layout = {
...
-    xs: token(576),
-    sm: token(768),
-    md: token(992),
-    lg: token(1200),
-    xl: token(1440),
+    xs: token('576px', { static: true }),
+    sm: token('768px', { static: true }),
+    md: token('992px', { static: true }),
+    lg: token('1200px', { static: true }),
+    xl: token('1440px', { static: true }),
```
coryrylan, line 27: Static ensures the value is not transformed in the token build process. This is important for values that are computed in sass mixins or values that can be css vars in media queries.

packages/core/src/internal/utils/css.ts
```diff
@@ -41,17 +41,7 @@ export function updateElementStyles(el: HTMLElement, ...styleTuples: [string, st
...
-
-  if (!base) {
-    const prop = window
-      .getComputedStyle(document.body, null)
-      .getPropertyValue('--cds-global-typography-base-font-size');
-    base = (16 * parseInt(prop !== '' ? prop : '100%')) / 100;
-    document.documentElement.setAttribute('cds-base-font', `${base}`);
-  }
-
-  return `${pxValue / base}rem`;
```
coryrylan, line 54: While pxToRem could do this math, by passing it through a CSS calc it can skip the expensive getComputedStyle lookup and keep the value dynamic. This is helpful when dynamically switching themes that adjust the size/space tokens.

packages/core/src/styles/layout/_optimize.scss
```diff
@@ -7,22 +7,22 @@
...
-$cds-global-layout-space-lg: var(--clg);
-$cds-global-layout-space-xl: var(--cxl);
-$cds-global-layout-space-xxl: var(--cxxl);
+$cds-global-layout-space-xxs: var(--δ1);
+$cds-global-layout-space-xs: var(--δ2);
+$cds-global-layout-space-sm: var(--δ3);
+$cds-global-layout-space-md: var(--δ4);
+$cds-global-layout-space-lg: var(--δ5);
+$cds-global-layout-space-xl: var(--δ6);
+$cds-global-layout-space-xxl: var(--δ7);
```
coryrylan, line 25: Static was no longer needed here and using a unicode greek char can give a namespace while reducing the large amount of repetitiveness in the css layout utils. This overall decreases the bundle sizes for layouts.

mathisscott (approved):

bbogdanov (approved):

packages/core/src/internal/utils/css.ts
bbogdanov: That's huge!

dima-runecast:
It actually does introduce a breaking change if the CSS variable is not added to the project.
```
:root {
  --cds-global-base: 16;
}
```

We are importing icons as collections so that might be the reason the CSS var it's not automatically included.
Maybe it would be good to add this info to docs.
Thanks for all your work!

coryrylan:
@dima-runecast I am working on opening a PR that will backport the selector to the vars css file and will go out with the next release. This should fix the case of only loading specific css modules as before.

PR #5962, opened 2021-05-19
feat(button): core flat-inline action

### LINKED ISSUE #5775: Add a variant to core flat buttons to align correctly when sitting left or right in a component
## Summary
As we are moving to clarity core, we’ve bumped up against a layout issue with the flat buttons. They’re not super useful for cases where they are positioned in layouts or components on the left or right as they won’t align correctly.  

- What is the change?
My proposal is that we add a variant to the flat button for these cases, just removing the left/right inside padding (or even 2px would do the trick) and changing the radius on the button container to zero.

- Why should it go in Clarity?
I think the biggest impact will be for cards…since most of the VMware apps have dashboards with cards and the clarity system leans on the flat buttons at the bottom of cards.  It’s really the only thing I have run across that you can’t make using core out of the box (without an alignment problem that is)..  attaching a few screens from the design system site.  You can’t actually do this with with core unless we all add customizations.

- If this change introduces a new behavior, is this behavior accessible?
The change only removes left/right padding, but leaves the container intact for focus and disabled state.  I don't see any impact on a11y

## Examples

Examples of the new variant in a card attached.  

Flat Button Variant- No Left-Right Padding

Other examples we have include: flat buttons over a datagrid and the back button positioned over the page title.  I can see this being an issue with flat buttons at the bottom of menus, forms and any time you need to add a button above or below another element and need to it visually align with that element.

bbogdanov:
Closes #5775

We also needed a way to have a button with no padding on the right and left this one is giving us just what we need.

## PR Checklist

- [] If applicable, have a visual design approval

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #5775 

## What is the new behavior?

Having a flat button with no padding on left and right

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/base-button.element.scss
+++ b/packages/core/src/button/base-button.element.scss
@@ -168,11 +168,17 @@
   --border-color: var(--color);
 }
 
-:host([action='flat']) {
+:host([action*='flat']) {
   --background: #{$cds-alias-object-opacity-0};
   --border-color: #{$cds-alias-object-opacity-0};
   --color: #{$cds-alias-status-info};
   --box-shadow-color: #{$cds-alias-object-opacity-0};
+}
+
+:host([action='flat-inline']) {
+  --height: auto;
+  --padding: 0;
+  --min-width: auto;
 }
 
 :host([size='sm']) {
@@ -212,7 +218,7 @@
   --color: #{$cds-alias-status-disabled};
 }
 
-:host([disabled][action='flat']) {
+:host([disabled][action*='flat']) {
   --border-color: #{$cds-alias-object-opacity-0};
   --background: #{$cds-alias-object-opacity-0};
   --color: #{$cds-alias-status-disabled};
--- a/packages/core/src/button/button.element.scss
+++ b/packages/core/src/button/button.element.scss
@@ -23,7 +23,7 @@
   --letter-spacing: 0.073em;
 }
 
-:host([action='flat']) ::slotted(cds-badge),
+:host([action*='flat']) ::slotted(cds-badge),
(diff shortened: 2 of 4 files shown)


coryrylan (approved):

mathisscott (approved):

coryrylan (approved):

packages/core/src/button/base-button.element.scss
```diff
@@ -168,13 +168,19 @@
...
+:host([action*='flat']) {
   --background: #{$cds-alias-object-opacity-0};
   --border-color: #{$cds-alias-object-opacity-0};
   --color: #{$cds-alias-status-info};
   --box-shadow-color: #{$cds-alias-object-opacity-0};
 }
 
+:host([action='flat-inline']) {
+  --height: auto;
+  --padding: 0;
```
mathisscott, line 180: I thought we only wanted to remove padding from the left and the right. Why would we remove all padding here?

mathisscott (changes requested):

packages/core/src/button/base-button.element.scss
bbogdanov: The buttons in the card footer are adding additional bottom space and it is not possible to remove that space without an additional token. Like we do with `  --card-remove-margin`. 

`--height: auto` will help us accomplish to remove that space and have an even padding that won't make the footer bigger than what's needed.

Also I and Cory discuss that and we thought having a `--height: auto` will bring more `inline` feeling.

mathisscott (approved):

PR #5966, opened 2021-05-20
fix(button): disabled button should remain disabled after loadingState changes to default

### LINKED ISSUE #5965: Disabled Button should remain disabled regardless of loadingState
## Describe the bug

A disabled Button becomes enabled after loadingState gets updated to default. However, the button should remain to be disabled when `disabled="true"` regardless of loadingState.

## How to reproduce

```jsx
import React from 'react';

export default function Component() {
  const [loading, setLoading] = useState('loading');

  setTimeout(() => {
    setLoading('default');
  }, 1000);

  return (
    <Button disabled={true} loadingState={loading}>
      Confirm
    </Button>
  );
}
```

When loadingState changed from 'loading' to 'default', the disabled attribute wouldn't affect any more.

I checked the source code at `packages/core/src/button/button.element.ts`

```js
  private updateLoadingState() {
    switch (this.loadingState) {
      case ClrLoadingState.loading:
        this.disableButton();
        return;
      case ClrLoadingState.success:
        this.disableButton();
        return;
      case ClrLoadingState.error:
        this.disableButton();
        return;
      default:
        // HERE IS WHY: the logic is missing disabled attribute checking. If `disabled`, call `this.disableButton()`, otherwise `this.enableButton()`
        this.enableButton();
    }
  }
```

## Expected behavior

For a disabled button, as long as disabled is true, the button should always be disabled, regardless of loadingState.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.x

**Framework:**

- [ ] Angular
- [x] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: React 17.0.0

**Device:**

- Type: MacBook
- OS: Catalina
- Browser: Any
- Version: Any

wghglory:
Close: #5965

Signed-off-by: Derek Wang <guanghui-wang@foxmail.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/vmware/clarity/issues/5965

Issue Number: #5965

## What is the new behavior?

Disabled button remains disabled regardless of loadingState

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/button.element.ts
+++ b/packages/core/src/button/button.element.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { baseStyles, CdsBaseButton, getElementWidth, property, spanWrapper } from '@cds/core/internal';
+import { baseStyles, CdsBaseButton, property, spanWrapper } from '@cds/core/internal';
 import { ClarityIcons } from '@cds/core/icon/icon.service.js';
 import { errorStandardIcon } from '@cds/core/icon/shapes/error-standard.js';
 import { checkIcon } from '@cds/core/icon/shapes/check.js';
@@ -155,12 +155,12 @@ export class CdsButton extends CdsBaseButton {
         this.disableButton();
         return;
       default:
-        this.enableButton();
+        this.disabled ? this.disableButton() : this.enableButton();
     }
   }
 
   private disableButton() {
-    this.style.width = getElementWidth(this);
+    this.style.removeProperty('width');
     this.disabled = true;
   }
 
--- a/apps/core-create-react-app/src/App.tsx
+++ b/apps/core-create-react-app/src/App.tsx
@@ -13,6 +13,7 @@ ClarityIcons.addIcons(userIcon);
 interface AppProps {}
 interface AppState {
   show: boolean;
+  loadingState: "default" | "success" | "loading" | "error" | undefined;
 }
 
 export default class App extends Component<AppProps, AppState> {
@@ -20,7 +21,15 @@ export default class App extends Component<AppProps, AppState> {
     super(props);
     this.state = {
       show: false,
+      loadingState: 'loading'
     };
+
+    setTimeout(() => {
+      this.setState({
+        ...this.state,
+        loadingState: 'default'
+      });
+     }, 1000);
   }
 
   render() {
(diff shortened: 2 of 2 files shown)


packages/core/src/button/button.element.ts
```diff
@@ -155,12 +155,12 @@ export class CdsButton extends CdsBaseButton {
         this.disableButton();
         return;
       default:
-        this.enableButton();
+        this.disabled ? this.disableButton() : this.enableButton();
```
hippee-lee, line 158: Personally I would rather see this handled with a guard. If a button is disabled it shouldn't be changing states, right? 

So above line #_147 with a guard for disabled might look like this

```
if (this.disabled) {
    return;
}
```

and then in the firstUpdated fn it would add a `!this.disabled` condition before it checks the loading state. 

@mathisscott or @coryrylan - thoughts?

hippee-lee (changes requested):

packages/core/src/button/button.element.ts
```diff
@@ -155,12 +155,12 @@ export class CdsButton extends CdsBaseButton {
...
     }
   }
 
   private disableButton() {
-    this.style.width = getElementWidth(this);
+    this.style.removeProperty('width');
```
mathisscott, line 163: What is this change here intended for?

packages/core/src/button/button.element.ts
mathisscott: A guard would work fine from my understanding of what the issue is. If we put the guard above line 147, no other checks would be needed in `updated` or `firstUpdated`.

We need a unit test to verify this behavior though...

mathisscott (changes requested):

hippee-lee:
Closing this because the initial idea was adopted and refined here: https://github.com/vmware/clarity/pull/6048

packages/core/src/button/button.element.ts
wghglory: When switching between button with loading icon and button with text, the button width will be changed. This line ensures the button width looks normal, but with the guard, I think this is not needed anymore. Great suggestions @mathisscott !

hippee-lee: Thanks again for the contribution @wghglory. FYI I moved your commit over to the new PR: https://github.com/vmware/clarity/pull/6048/commits

wghglory: @hippee-lee Great! Thank you for letting me know and making this change!

PR #6013, opened 2021-05-31
fix: accordion and stepper issue with clrAccordionPanelOpenChange event

### LINKED ISSUE #5957: Regression: clrAccordionPanelOpenChange emits only when the panel is expanded
## Describe the bug

There is a regression in Accordion/Stepper component. `clrAccordionPanelOpenChange` event emits only when a panel is expanded.

## How to reproduce

The bug is 100% reproducible on ClrStepper and ClrAccordion. `ClrAccordionPanel.panelOpenChange` does not emit when the panel is collapsed.

## Expected behavior

`clrAccordionPanelOpenChange` should emit `true` when a panel is expanded and `false` when a panel is closed.

## Versions

**Clarity project:**

- [ ] Clarity Core
- [X] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [X] v5.x

**Framework:**

- [X] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: MacBook
- OS: iOS
- Browser:  Chrome
- Version: 90

## Additional notes

The regression is introduced by this commit:
https://github.com/vmware/clarity/commit/4e338f3bde2ea10c8b0b7a704c8624058cf2e2f7

### LINKED ISSUE #5981: Regression: ClrStepperPanel and ClrAccordionPanel do not change the caret icon when expanded
## Describe the bug

In previous releases (before v5) the accordion and stepper panel changed their caret icon depending on the expanded state of the panel. This can be seen in the official Clarity Design page:

https://clarity.design/angular-components/stepper/
https://clarity.design/images/angular-components/stepper/stepper-style.png

## How to reproduce

Run the sample code from https://clarity.design/angular-components/stepper/ in stackblitz.
Observe the behaviour when panels are expanded/collapsed.

Stepper caret not rotated

## Expected behavior

When the panel is expanded, the caret should rotate down (as shown in the official documentation).

## Versions

**Clarity project:**

- [ ] Clarity Core
- [X] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [X] v5.x

**Framework:**

- [X] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: MacBook
- OS: iOS
- Browser: Chrome
- Version: 90

## Additional notes

Looking at the source code:
https://github.com/vmware/clarity/blob/c77dee037dda21556dba30aa4eb742ea1832677b/packages/angular/projects/clr-angular/src/accordion/accordion-panel.html

it seems that there is no logic for changing the icon at all:
`<clr-icon shape="angle" dir="right" class="clr-accordion-angle"></clr-icon>`

@bdryanovski

bdryanovski:
Fixing issue #5957 with not firing both states of the clrAccordionPanelOpenChange (true and false)

Also closing #5981 - could not reproduce it - probably fixed in previous versions (tested with latest v5)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #5981 & #5957 

- [ ] Yes
- [ ] No

## Other information

closes #5957 
closes #5981

### DIFF
--- a/packages/angular/projects/clr-angular/src/accordion/accordion-panel.spec.ts
+++ b/packages/angular/projects/clr-angular/src/accordion/accordion-panel.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2019 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -20,7 +20,11 @@ import { IfExpandService } from '../utils/conditional/if-expanded.service';
 @Component({
   template: `
     <clr-accordion>
-      <clr-accordion-panel [(clrAccordionPanelOpen)]="open" [clrAccordionPanelDisabled]="disabled">
+      <clr-accordion-panel
+        [(clrAccordionPanelOpen)]="open"
+        [clrAccordionPanelDisabled]="disabled"
+        (clrAccordionPanelOpenChange)="change($event)"
+      >
         <clr-accordion-title>title</clr-accordion-title>
         <clr-accordion-description *ngIf="showDescription">description</clr-accordion-description>
         <clr-accordion-content>panel</clr-accordion-content>
@@ -32,6 +36,26 @@ class TestComponent {
   open = false;
   disabled = false;
   showDescription = false;
+  change = state => {
+    return state;
+  };
+}
+
+@Component({
+  template: `
+    <clr-accordion>
+      <clr-accordion-panel (clrAccordionPanelOpenChange)="change($event)">
+        <clr-accordion-title>title</clr-accordion-title>
+        <clr-accordion-description>description</clr-accordion-description>
+        <clr-accordion-content>panel</clr-accordion-content>
+      </clr-accordion-panel>
+    </clr-accordion>
+  `,
+})
+class TestNoBindingComponent {
+  change = state => {
+    return state;
+  };
 }
 
 describe('ClrAccordionPanel', () => {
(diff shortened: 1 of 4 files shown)


Jinnie (approved):

hippee-lee (changes requested):
This lgtm. Before I give it a :shipit: I'd like to see a test that covers the change so we know if 
```this.panelOpen``` 
stops staying in sync with 
```panel.open```

bdryanovski:
@hippee-lee done.

In addition to that will try to explain the case here for historical reasons.

Because we use the `BehaviorSubject` we have a bit strange result that the `(clrAccordionPanelOpenChange)` is called when the Accordion is rendering for the first time with its initial value. And that is creating an unneeded event - also for some reason that I could not found out is that I could not use RxJs mechanics for preventing emitting an event only when changed - Previous and Next values for some strange reasons are already the same.

To NOT create additional private property to track internally the state of the Accordion and decide to emit or not an event - I'm using the `panelOpen` as a reference point to the previous value. 

But when there is no Angular binding to it - it goes off sync (we no longer go into `ngOnChanges`) so this is the reason for manually assigning it into the emitter method.

In addition to that, the `panelOpen` is used only into the `ngOnInit` to set the value and to be outside property to trigger change - when it comes to the template we use only `panel.open` - so no side effects into the template could be expected.

Trying to go back and fix the root of the problem `BehaviorSubject` - I'm getting a lot of lifecycle errors when it comes to rendering it. Really really hard to debug it. Also any changes made to the Accordion have effect on the Stepper and until someone decide to rework both of them - I think that this patch version is good enough for the moment

hippee-lee (approved):

Issue #6002, opened 2021-05-26
core: cds-button submit error


coryrylan:
## Describe the bug

The cds-button will incorrectly trigger a submit event in a form when set to type button.

## How to reproduce

Focus the html cancel button and use space or enter. Neither submit form due to type button. However cds-button will submit.

https://stackblitz.com/edit/typescript-gtxjpo

https://stackblitz.com/edit/react-ts-sudzgj

## Expected behavior

When a cds-button uses type button it should not trigger a form submit.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.x

**Framework:**

- [ ] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: [e.g. MacBook]

## Additional notes

_Add any other notes about the problem here._

PR #6017, opened 2021-06-01
fix: core button submit behavior

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Base button would fire submit events incorrectly when set to type "button". The base button also incorrectly marked the active state for keydown and mousedown events. 

Issue Number: #6002 

## What is the new behavior?

Now the button will prevent submit events if type is set to button. Active state will only apply during mousedown and keydown events.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/button.stories.ts
+++ b/packages/core/src/button/button.stories.ts
@@ -44,13 +44,14 @@ export function form() {
       cds-layout="vertical gap:md"
       @submit="${(e: Event) => {
         e.preventDefault();
-        action('submit')(e);
+        console.log('submit');
       }}"
     >
       <div cds-layout="vertical gap:sm">
         <label for="name" cds-text="caption">Name</label>
         <input id="name" />
       </div>
+      <cds-button type="button" action="outline" @click=${() => console.log('click')}>cancel</cds-button>
       <cds-button type="submit">submit</cds-button>
     </form>
   `;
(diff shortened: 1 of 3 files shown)


packages/core/src/internal/base/button.base.ts
```diff
@@ -104,23 +118,27 @@ export class CdsBaseButton extends LitElement {
...
         ></button>`;
 
-        this.showClick();
-        this.appendChild(buttonTemplate.getTemplateElement().content.cloneNode(true) as HTMLElement);
-        const button = this.querySelector('.cds-hidden-button') as HTMLButtonElement;
-        button.dispatchEvent(new MouseEvent('click', { relatedTarget: this, composed: true }));
-        button.remove();
+        const marker = document.createElement('div');
+        this.appendChild(marker);
+        render(buttonTemplate, marker);
```
coryrylan, line 123: This fixes the issue of lit not rendering the type attr from the initial render.

jeeyun (approved):

bdryanovski (approved):

packages/core/src/button/button.stories.ts
```diff
@@ -44,13 +44,14 @@ export function form() {
       cds-layout="vertical gap:md"
```
mathisscott, line 44: The build is asking about the import on line 12. Do we still need that in this file?

mathisscott (approved):

packages/core/src/button/button.stories.ts
coryrylan: I don't think so, I'll remove it 👍

PR #6019, opened 2021-06-02
fix: decouple adoption guide from dist folder

bdryanovski:
Make Adoption guide independent from already build packages inside `dist` folder and use only released packages from NPM


yarn.lock
```diff
@@ -3841,7 +3841,21 @@
...
+    lit-element "^2.3.1"
+    lit-html "^1.2.1"
+    ramda "^0.27.0"
+    tslib "^2.0.0"
+  optionalDependencies:
+    "@cds/city" "^1.1.0"
+    normalize.css "^8.0.1"
+
+"@cds/core@latest":
+  version "5.3.0"
```
mathisscott, line 3856: Something is weird in this yarn.lock file. 

The `@cds/core` here has the latest version (5.3.0) but there are other packages here pointing at `./dist` and 5.2.1.

That doesn't seem right.

mathisscott (changes requested):

gnomeontherun:
This should be fixed in the `angular` branch anyways.

bdryanovski:
@gnomeontherun the Adoption Guide for the moment is part of the website Build & Deploy (as subfolder of the website https://clarity.design/adoption/) so until it has its own deploy domain or Netlify build (something that we need to talk more about) it's better to stay on the `next` branch. 

The change here is just to make more easily moving the app if needed and we don't really need to use the unreleased code so depending on only published packages will be a better and simple way to go.

yarn.lock
bdryanovski: There are still packages that use `dist/<package>` to link - or at least on the `next` branch right now. 

The change here is to make the adoption guide not depend on the pre-build packages but use only published ones. This is making moving the app super painless.

At the moment `apps/adoption` (where the app is located) is into the `nohoist` paths for Yarn, but it still affects the `yarn.lock` file when some of the packages change.

So this is the reason for the change into the `yarn.lock`

mathisscott:
> This should be fixed in the `angular` branch anyways.

Does this mean

mathisscott (approved):

PR #6020, opened 2021-06-02
feat(tree-view): core tree view

jeeyun:
Core web component implementation of the tree view.

## Simple
Screen Shot 2021-06-02 at 1 05 03 PM

## Multi-select
Screen Shot 2021-06-02 at 1 04 53 PM

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/core/src/internal-components/visual-checkbox/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export * from './visual-checkbox.element.js';
```
mathisscott, line 7: I don't think this approach will work after the Lit 2 upgrade PR. IIRC, rollup has a problem with barrel imports referencing barrel imports. Barrel Inception?

Maybe we should remove this file? What say you, @coryrylan ?

packages/core/src/internal/interfaces/index.ts
```diff
@@ -17,6 +17,8 @@ export type StatusTypes = 'neutral' | 'info' | 'success' | 'warning' | 'danger';
 // MISC
 export type AriaLivePolitenessSettings = 'off' | 'polite' | 'assertive';
 
+export type AriaBooleanAttributeValues = 'true' | 'false';
```
mathisscott, line 20: Why are we making a distinction here with the string boolean values for aria?

packages/core/src/internal/motion/animations/cds-tree-item-expand.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { TargetedAnimation } from '../interfaces.js';
+
+export const AnimationTreeItemExpandName = 'cds-tree-item-expand';
+export const AnimationTreeItemExpandConfig: TargetedAnimation[] = [
+  {
```
mathisscott, line 11: What did you think about the animations? Hard? Easy?

packages/core/src/internal/motion/index.ts
```diff
@@ -14,6 +14,7 @@ export {
   AnimationAccordionPanelOpenConfig,
   AnimationAccordionPanelOpenName,
 } from './animations/cds-accordion-panel-open.js';
+export { AnimationTreeItemExpandConfig, AnimationTreeItemExpandName } from './animations/cds-tree-item-expand.js';
```
mathisscott: This is fine here right now. But I wanted to give a heads up to @coryrylan because I think this file goes away in the lit 2 upgrade.

packages/core/src/internal/utils/array.ts
```diff
@@ -14,3 +14,17 @@ export function arrayToObject(arr: any[], key: string) {
...
+
+export function previousInArray(current: any, arr: any[]) {
+  const idx = arr.indexOf(current);
+  return idx === -1 ? void 0 : arr[Math.max(idx - 1, 0)];
+}
+
+export function nextInArray(current: any, arr: any[]) {
+  const idx = arr.indexOf(current);
+  return idx === -1 ? void 0 : arr[Math.min(idx + 1, arr.length - 1)];
+}
```
mathisscott, line 30: 🏆

packages/core/src/tree-view/tree-item.element.scss
```diff
@@ -0,0 +1,135 @@
...
+  [part='expand-collapse-icon'] {
+    --icon-transform: rotate(180deg);
+  }
+}
+
+[part='expand-collapse-icon'] {
+  --icon-transform: rotate(90deg);
+  color: var(--cds-alias-object-interaction-color);
+  cursor: inherit;
+  transform: var(--icon-transform);
```
mathisscott: I don't know what we're getting by defining a CSS custom property and then immediately using it in the same selector.

I also don't think this is how part is used.

packages/core/src/tree-view/tree-item.element.ts
```diff
@@ -0,0 +1,232 @@
...
+  @property({ type: Boolean, reflect: true })
+  indeterminate = false;
+
+  @property({ type: Boolean, reflect: true })
+  loading = false;
+
+  @property({ type: Boolean, reflect: true })
+  selected = false;
+
+  // The :scope selector allows us to query the cds-tree-item nodes that are direct children
```
mathisscott, line 105: 👍

mathisscott (changes requested):

packages/core/src/internal/interfaces/index.ts
jeeyun: If you use a boolean type for the property, it will conditionally add the attribute if the value is true (much like `disabled`, `checked`, etc).

But for `aria-` attributes we need an explicit string value of `true` or `false` assigned to the attribute (e.g. `aria-selected='false'`).

Are there other non `aria-` attributes that work the same way? Happy to name the type differently if so.

mathisscott: No changes needed here. I just wanted to understand the why.

packages/core/src/internal/motion/animations/cds-tree-item-expand.ts
jeeyun: I liked using the decorator. It was helpful to see examples of animations already in place in some of the components.

packages/core/src/tree-view/tree-item.element.scss
jeeyun: Yeah you are right. I was trying to make the component super flexible (allow any custom content projection for the expand/collapse icon) and in the process got confused about how to use the part correctly. 😅 

I've made changes to expose the expand/collapse through part so it can be customized, but got rid of the slot projection ability. App side can do the following to customize the icon and the checkbox (example code available in storybook):

```
.app-custom cds-tree-item::part(checkbox) {
  --color: var(--cds-global-color-jade-700);
  --check-color: var(--cds-global-color-jade-50);
}
.app-custom cds-tree-item::part(expand-collapse-icon) {
  --color: var(--cds-global-color-jade-700);
}
```

Does this seem reasonable? 

Also, I've moved up the `--icon-transform` back to host level. Is this something we should consider exposing at the icon level (e.g. `--transform` property that's consumed by cds-icon)?

packages/core/src/internal-components/visual-checkbox/index.ts
coryrylan: This should be fine because its the index of the entrypoint `visual-checkbox`

packages/core/src/internal/motion/index.ts
coryrylan: Yeah this is a nested barrel which rollup does not allow due to treeshaking issues. This would be exported in the `/internal/index.ts` file instead

packages/core/src/tree-view/tree-item.element.ts
```diff
@@ -0,0 +1,232 @@
...
+  protected ariaExpanded: AriaBooleanAttributeValues | undefined = 'false';
+
+  @internalProperty({ type: String, reflect: true, attribute: 'aria-selected' })
+  protected ariaSelected: AriaBooleanAttributeValues | undefined = 'false';
+
+  @property({ type: Boolean, reflect: true, attribute: 'multi-select' })
+  multiSelect = false;
+
+  @property({ type: String, reflect: true })
+  id: string;
```
coryrylan: `id` is a native property/attribute to all HTML elements so you don't have to declare it here with `@property`

packages/core/src/tree-view/tree.element.ts
```diff
@@ -0,0 +1,214 @@
...
+ *  </cds-tree>
+ * ```
+ *
+ * @beta
+ * @element cds-tree
+ * @slot - Content slot for inside the tree
+ */
+export class CdsTree extends LitElement {
+  @property({ type: String, reflect: true })
+  role = 'tree';
```
coryrylan: Could `role` use `inernalProperty` or `@state` after rebase? This seems like it wouldnt be a documented public API to change/customize.

packages/core/src/tree-view/tree.element.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  @property({ type: Boolean, attribute: 'multi-select' })
+  multiSelect = false;
+
+  @property({ type: String, reflect: true, attribute: 'aria-activedescendant' })
+  ariaActiveDescendant: string;
```
coryrylan, line 54: Could `ariaActiveDescendant` use `inernalProperty` or `@state` after rebase? This seems like it wouldn't be a documented public API to change/customize.

packages/core/src/tree-view/tree.element.ts
```diff
@@ -0,0 +1,214 @@
...
+    this.addEventListener('focus', this.initAriaActiveDescendant);
+    this.addEventListener('click', this.clickHandler);
+    this.addEventListener('keydown', this.keyboardNavigationHandler);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('focus', this.initAriaActiveDescendant);
+    this.removeEventListener('click', this.clickHandler);
+    this.removeEventListener('keydown', this.keyboardNavigationHandler);
```
coryrylan: For custom elements when adding an event listener to the host element itself (`this`) you don't have to remove the event listeners in `disconnectedCallback`. Since the listener is attached to the element it will automatically get garbage collected when removed from the DOM. It's also not an issue to remove them in the disconnectedCallback.

packages/react/src/tree-view/index.test.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import { mount, shallow } from 'enzyme';
+import { CdsTree, CdsTreeItem } from './index';
+
+describe('CdsTag', () => {
```
coryrylan: nit: `CdsTree`

packages/core/src/tree-view/tree-item.element.scss
```diff
@@ -0,0 +1,132 @@
...
+  .lhs-container {
+    width: #{$cds-global-space-7};
+  }
+
+  .expand-collapse-icon {
+    --color: var(--cds-alias-object-interaction-color);
+    transform: var(--icon-transform);
+    transition: transform var(--animation-duration);
+  }
+}
```
coryrylan: nit: Since the CSS is scoped could we remove this first group of nesting since it will generate excess CSS?

packages/core/src/tree-view/tree.stories.mdx
```diff
@@ -0,0 +1,81 @@
+import { Meta, Props, Story, Preview, API } from '@storybook/addon-docs/blocks';
+
+<Meta title="Components/Tree View" />
+
+# Tree View
```
coryrylan, line 5: nit: add beta warning message like [this one](https://github.com/vmware/clarity/blame/next/packages/core/src/progress-circle/progress-circle.stories.mdx#L7)

packages/core/src/tree-view/tree-item.element.scss
coryrylan: I like the use of the CSS part API 👍  I think you could get away with not needing `--icon-transform` since the icon is exposed as a part they can apply the transform directly themselves. 

```css
.app-custom cds-tree-item::part(expand-collapse-icon) {
  transform: rotate(180deg);
}
```

Is there any reason to not apply the `expand-collapse-icon` part directly to the `cds-icon` rather than the wrapper div? I guess custom slotted icons would no longer rotate automatically

packages/core/src/internal/motion/index.ts
jeeyun: I won't move the existing exports but will add a comment that we should be moving them to `/internal/index.ts`.

packages/core/src/tree-view/tree-item.element.scss
jeeyun: I'm wondering if you are looking at old code, because the part is directly on the `cds-icon`! (It used to not be in my earlier code). I also got rid of the ability to add custom slotted icons for expand / collapse icon.

As for the name of the part, I wanted to keep it specific so as to distinguish from any other projected icons that maybe part of the content of tree node.

packages/react/src/tree-view/index.test.tsx
jeeyun: Caught copy-pasting 😂  Updated!

packages/core/src/tree-view/tree.element.ts
jeeyun: I have a bunch of [specs](https://github.com/vmware/clarity/blob/3aee6317fc65901d1de464c7e59f994a3dfe4cd0/packages/core/src/tree-view/tree.element.spec.ts) that ascertain that the ariaActiveDescandant is updated to the right value.

I am kind of torn because using `internalProperty` or `state` decorator forces the field to be `protected`, which makes me lose the ability to access them in my tests. I know that making a field public for tests' sake isn't always the best but I feel that this is a more robust test than just asserting that the update function was called.

What if I just annotated this field as `/** @private */` so it doesn't show up in docs? Or if you have a different suggestion I'm pretty open to it!

coryrylan: You should be able to use the `@state` decorator as a public class property. It uses the regular property decorator under the hood with some extras. We use a public property with a `@state` decorator [here](https://github.com/vmware/clarity/blob/next/packages/core/src/alert/alert.element.ts#L150) Mainly for the same reasons of inner component communication but without exposing it as a public API.

The analyzer treats any property marked with `@internalProperty`, `@state` or `/** @private */` as internal APIs regardless if it's a public TS property and won't make them available in the editor autocomplete or API docs. If the build/lint is preventing it let me know there might be something I misconfigured with the lit 2 update.

jeeyun: Ah, I haven't rebased with lit 2 update yet since it's mid-review. I'll verify if it'll let me do `@state` + public after the update.

packages/core/src/tree-view/tree-item.element.ts
```diff
@@ -0,0 +1,229 @@
...
+  indeterminate = false;
+
+  @property({ type: Boolean, reflect: true })
+  loading = false;
+
+  @property({ type: Boolean, reflect: true })
+  selected = false;
+
+  // The :scope selector allows us to query the cds-tree-item nodes that are direct children
+  @querySlotAll(':scope > cds-tree-item') private treeItemChildren: NodeListOf<CdsTreeItem>;
```
Jinnie, line 106: I still want to document my concern that we won't be able to support the case where our users will want to nest cds-tree-item nodes in their framework-specific component wrappers. This may be often attempted when building recursive dynamic trees.
As we talked with Jeeyun, we do have alternative approaches, but I am afraid that not supporting wrapper elements of cds-tree-item will lead to frequent support requests/issues. We need to, at least, document the scenario well.

coryrylan: This is a result of the API limitations of custom elements and Angular combined. Custom Elements can only safely access direct children of slots for encapsulation reasons. The Core forms and alerts have similar API limitations.

Almost all major frameworks default or provide container-less components with the exception of Angular. Most outside of Angular will have no issue with wrapping with their own framework components. Looks like there might be [possible workarounds](https://github.com/angular/angular/issues/18877#issuecomment-844354605).

Jinnie: Thanks, Cory - that's an awesome workaround, though a bit peculiar. I'll add it to our Support DB scenarios

packages/core/src/internal/motion/index.ts
mathisscott: This is probably something you'll need to do with the Lit 2 rebase...

mathisscott (approved):

packages/core/src/internal/motion/index.ts
jeeyun: addressed with the rebase 👍

packages/core/src/tree-view/tree.element.ts
jeeyun: Addressed with the rebase 👍

coryrylan (approved):

bbogdanov (approved):

PR #6024, opened 2021-06-03
feat: mvp cds-navigation

hippee-lee:
- adds default layout (vertical)
- custom expand/collapse icons for root navigation and navigation groups
- navigation groups
- arrow and home and end key navigation + a11y enhancements

## PR Checklist

## PR Type

MVP For CdsNavigation compoent. 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Core does not have navigation.

Issue Number: N/A

## What is the new behavior?
Core has a navigation with a vertical layout.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/internal/index.ts
+++ b/packages/core/src/internal/index.ts
@@ -56,3 +56,8 @@ export {
 } from './motion/animations/cds-accordion-panel-open.js';
 export { AnimationHingeConfig, AnimationHingeName } from './motion/animations/cds-overlay-hinge-example.js';
 export { AnimationShakeConfig, AnimationShakeName } from './motion/animations/cds-component-shake.js';
+export {
+  AnimationNavigationGroupOpenConfig,
+  AnimationNavigationGroupOpenName,
+} from './motion/animations/cds-navigation-group-open.js';
+export { AnimationNavigationOpenConfig, AnimationNavigationOpenName } from './motion/animations/cds-navigation-open.js';
--- a/packages/core/src/internal/motion/animations/cds-navigation-open.ts
+++ b/packages/core/src/internal/motion/animations/cds-navigation-open.ts
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { TargetedAnimation } from '../interfaces.js';
+
+export const AnimationNavigationOpenName = 'cds-navigation-open';
+export const AnimationNavigationOpenConfig: TargetedAnimation[] = [
+  {
+    target: 'cds-navigation',
+    animation: [{ width: 'var(--collapsed-width)' }, { width: 'var(--expanded-width)' }],
+    options: {
+      duration: '--animation-duration',
+      easing: '--animation-easing',
+      fill: 'forwards',
+    },
+  },
+];
--- a/packages/core/src/internal/services/i18n.service.ts
+++ b/packages/core/src/internal/services/i18n.service.ts
@@ -58,6 +58,12 @@ export const componentStringsDefault = {
     contentBox: 'Scrollable Modal Body',
     contentEnd: 'End of Modal Content',
   },
+  navigation: {
+    navigationElement: 'navigation',
+    navigationLabel: 'navigation menu',
+    navigationAbridgedText: 'View abridged menu',
+    navigationUnabridgedText: 'View unabridged menu',
+  },
   password: {
     showButtonAriaLabel: 'Show password',
     hideButtonAriaLabel: 'Hide password',
--- a/packages/core/src/navigation/navigation-group.element.scss
+++ b/packages/core/src/navigation/navigation-group.element.scss
@@ -1,0 +1,48 @@
+/*!
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+@import './../styles/tokens/generated/index';
+
+:host {
+  --animation-duration: #{$cds-global-animation-duration-secondary};
+  --animation-easing: #{$cds-global-animation-easing-primary};
+  --background: inherit;
+}
+
+.navigation-group-items {
+  height: 0;
+  overflow-y: hidden;
+}
+
+:host([cds-motion='off']) {
+  .navigation-group-items {
+    height: 0;
+  }
+}
+
+:host([cds-motion='off'][expanded]) {
+  .navigation-group-items {
+    height: auto;
+  }
+}
+
+:host([cds-motion][_cds-animation-status='ready']:not([cds-motion='off'])) {
+  .navigation-group-items {
+    height: 0;
+  }
+}
+
+:host([cds-motion][expanded][_cds-animation-status='ready']:not([cds-motion='off'])) {
+  .navigation-group-items {
+    height: auto !important;
+    transform: none;
+  }
+}
+
+.group-items-container,
+.group-items-wrapper {
+  width: 100%;
+}
--- a/packages/core/src/navigation/navigation-group.element.ts
+++ b/packages/core/src/navigation/navigation-group.element.ts
@@ -1,0 +1,191 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { html, LitElement, PropertyValues } from 'lit';
+
+import {
+  Animatable,
+  animate,
+  AnimationNavigationGroupOpenName,
+  baseStyles,
+  event,
+  EventEmitter,
+  i18n,
+  I18nService,
+  id,
+  property,
+  querySlot,
+  querySlotAll,
+  reverseAnimation,
+  state,
+  syncProps,
+  syncPropsForAllItems,
+} from '@cds/core/internal';
+import styles from './navigation-group.element.scss';
+import { CdsNavigationItem } from './navigation-item.element';
+import { CdsNavigationStart } from './navigation-start.element';
+
+export const CdsNavigationGroupTagName = 'cds-navigation-group';
+
+/**
+ *
+ * ```typescript
+ * import '@cds/core/navigation/register.js';
+ * ```
+ *
+ * ```html
+ * <cds-navigation-group>
+ *   <cds-navigation-start></cds-navigation-start>
+ *    <cds-navigation-item><a href="/home">Home</cds-navigation-item>
+ *    <cds-navigation-item><a href="/account">Account</cds-navigation-item>
+ * </cds-navigation-group>
+ * ```
+ *
+ * @beta
+ * @element cds-navigation-group
+ * @event expandedChange - notify when the user has clicked the navigation expand/collapse button
+ * @cssprop --animation-duration
+ * @cssprop --animation-easing
+ * @cssprop --background
+ * @slot
+ */
+@animate({
+  expanded: {
+    true: AnimationNavigationGroupOpenName,
+    false: reverseAnimation(AnimationNavigationGroupOpenName),
+  },
+})
+export class CdsNavigationGroup extends LitElement implements Animatable {
+  @property({ type: String })
+  cdsMotion = 'on';
+
+  // Keep this from polluting the wca output and causing @cds/angular issues when the directives are generated.
+  /** @private **/
+  @event()
+  expandedChange: EventEmitter<boolean>;
+
+  @event()
+  cdsMotionChange: EventEmitter<string>;
+
+  /**
+   * @desc
+   * Associate the (projected) cds-navigation-button with group-items-wrapper (aria-labelledby)
+   *
+   * @private
+   */
+  @id() navigationGroupId: string;
+
+  @i18n() i18n = I18nService.keys.navigation;
+
+  /**
+   * @description
+   * Getter method for a reference to the selector cds-navigation-group > cds-navigation-start
+   * This lets each group flag its cds-navigation-start element and sync that info down. This is
+   * needed because cds-navigation-start elements can be used at the root level and inside
+   * cds-navigation-group elements.
+   *
+   * @private
+   */
+  @state()
+  private get isGroupStart(): boolean {
+    return !!this.groupStart;
+  }
+
+  @property({ type: Boolean, reflect: true })
+  expanded = false;
+
+  @property({ type: Boolean })
+  active: boolean;
+
+  /**
+   * @desc
+   * The value of this property is passed down to start and item children. It is used to query for visible items when
+   * managing focus with key events in the root cds-navigation element.
+   *
+   * Note: eslint-disable  @typescript-eslint/no-unused-vars isn't ignoring the line
+   // eslint error happens because the value is set but never read.
+
+   * @private
+   */
+
+  @state()
+  expandedGroup = false;
+
+  /**
+   * @desc
+   *
+   * Used to coordinate css things and the keyboard navigation focus changes.
+   */
+  hasFocus = false;
+
+  @querySlot(':scope > cds-navigation-start', { assign: 'group-start' })
+  protected groupStart: CdsNavigationStart;
+
+  @querySlotAll(':scope > cds-navigation-item')
+  protected groupItems: NodeListOf<CdsNavigationItem>;
+
+  @querySlotAll(':scope > cds-navigation-group')
+  protected nestedGroups: NodeListOf<CdsNavigationItem>;
+
+  render() {
+    return html`
+      <div role="listitem" class="private-host" cds-layout="vertical wrap:none align:horizontal-stretch">
+        <slot name="group-start"></slot>
+        <div role="region" class="group-items-wrapper" aria-labelledby="${this.navigationGroupId}">
+          <div role="list" class="group-items-container">
+            <div
+              class="navigation-group-items"
+              aria-expanded="${this.expanded}"
+              cds-layout="vertical wrap:none align:horizontal-stretch"
+            >
+              <slot></slot>
+            </div>
+          </div>
+        </div>
+      </div>
+    `;
+  }
+
+  static get styles() {
+    return [baseStyles, styles];
+  }
+
+  private toggle() {
+    this.expandedChange.emit(!this.expanded);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.groupStart.removeEventListener('click', this.toggle.bind(this));
+  }
+
+  protected firstUpdated(props: PropertyValues) {
+    super.firstUpdated(props);
+    if (this.groupStart) {
+      this.groupStart.addEventListener('click', this.toggle.bind(this));
+    }
+  }
+
+  updated(props: PropertyValues<this>) {
+    super.updated(props);
+    if (props.has('expanded')) {
+      this.expandedGroup = this.expanded;
+    }
+
+    if (this.groupStart) {
+      syncProps(this.groupStart, this, {
+        active: true,
+        expanded: true,
+        isGroupStart: this.isGroupStart,
+        navigationGroupId: true,
+      });
+    }
+
+    syncPropsForAllItems(Array.from(this.groupItems), this, {
+      expandedGroup: true,
+    });
+  }
+}
--- a/packages/core/src/navigation/navigation-item.element.scss
+++ b/packages/core/src/navigation/navigation-item.element.scss
@@ -1,0 +1,67 @@
+/*!
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+@import './../styles/tokens/generated/index';
+
+:host {
+  --color: inherit;
+  --line-height: inherit;
+  --font-size: inherit;
+  --font-weight: inherit;
+  --letter-spacing: inherit;
+  --padding: inherit;
+}
+
+:host(:hover:not([disabled])),
+:host([_has-focus]) {
+  background: #{$cds-alias-object-interaction-background-hover};
+}
+
+:host(:active),
+:host([active]) {
+  background: #{$cds-alias-object-interaction-background-active};
+}
+
+:host([selected]) {
+  background: #{$cds-alias-object-interaction-background-selected};
+}
+
+:host([disabled]) {
+  --color: #{$cds-alias-object-interaction-color-disabled};
+
+  // disabled items should not respond to hover
+  &:hover {
+    --background: inherit;
+  }
+
+  ::slotted(a) {
+    cursor: not-allowed;
+  }
+}
+
+:host([_group-item]) {
+  padding: var(--padding);
+}
+
+::slotted(cds-icon) {
+  color: inherit;
+}
+
+// These are not our elements and the consumer can easily override us here.
+::slotted(a) {
+  color: var(--color) !important;
+  padding: var(--padding) !important;
+  text-decoration: none !important;
+  outline: none !important;
+}
+
+.private-host {
+  font-size: var(--font-size);
+  font-weight: var(--font-weight);
+  letter-spacing: var(--letter-spacing);
+  line-height: var(--line-height);
+  width: 100%;
+}
--- a/packages/core/src/navigation/navigation-start.element.scss
+++ b/packages/core/src/navigation/navigation-start.element.scss
@@ -1,0 +1,48 @@
+/*!
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+@import './../styles/tokens/generated/index';
+@import './../styles/mixins/mixins';
+
+:host {
+  --color: inherit;
+  --line-height: inherit;
+  --font-size: inherit;
+  --font-weight: inherit;
+  --text-transform: capitalize;
+}
+
+:host(:hover),
+:host([_has-focus]) {
+  background: #{$cds-alias-object-interaction-background-hover};
+}
+
+.private-host {
+  color: var(--color);
+  background: inherit;
+  border: 0;
+  font-size: var(--font-size);
+  font-weight: var(--font-weight);
+  height: var(--line-height);
+  outline: none;
+  text-transform: var(--text-transform);
+
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.icon-slot {
+  color: var(--color);
+}
+
+::slotted(span) {
+  color: var(--color);
+}
+
+:host([is-group-start][_active]:not([_expanded])) {
+  background: #{$cds-alias-object-interaction-background-active};
+}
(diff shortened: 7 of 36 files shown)


packages/core/package.json
```diff
@@ -11,7 +11,8 @@
     "test": "npm-run-all -s test:unit test:performance test:package",
     "test:watch": "karma start ./karma.conf.js --auto-watch=true --single-run=false --coverage=false",
     "test:package": "yarn run package-check --cwd ./dist/core",
-    "test:unit": "tsc --b ./tsconfig.spec.json --force && karma start ./karma.conf.js && yarn run build:package",
+    "test:unit": "tsc --b ./tsconfig.spec.json --force && karma start && karma start ./karma.conf.js && yarn run build:package",
+    "test:dev": "stmux -w always -e ERROR -m beep,system [ \"tsc --b ./tsconfig.spec.json -w --force\" : \"sleep 20 && yarn run build:watch\" : \"sleep 40 && karma start ./karma.conf.js --auto-watch=true --single-run=false --coverage=false\" ]",
```
hippee-lee: I added a thing so I could run everything in one window and see when it failed in one place rather than two. I wanted to watch src and spec and be able to debug things in the browser.

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -58,6 +58,12 @@ export const componentStringsDefault = {
     contentBox: 'Scrollable Modal Body',
     contentEnd: 'End of Modal Content',
   },
+  navigation: {
+    navigationElement: 'navigation',
+    navigationLabel: 'navigation menu',
+    navigationAbridgedText: 'View abridged menu',
```
hippee-lee, line 64: A11y decided on the word abridged rather than expanded. I didn't change the attribute tho, it is still expanded so abridged (consumer/user) =  expanded (dev / attr)

packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+@import './../styles/tokens/generated/index';
+
+:host {
+  font-family: var(--font-family);
+  font-size: var(--font-size);
+  font-weight: var(--font-weight);
+  letter-spacing: var(--letter-spacing);
+  //background: var(--background);
```
bbogdanov: Not needed anymore?

packages/core/src/navigation/navigation-group.element.ts
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * ```html
+ * <cds-navigation-group>
+ *
+ * </cds-navigation-group>
+ * ```
+ *
+ * @element cds-navigation-group
+ * @slot
+ */
```
bbogdanov, line 54: We should add the props and events here so we can expose them in Storybook

packages/core/src/internal/utils/dom.ts
```diff
@@ -149,11 +149,12 @@ export function isVisible(element: HTMLElement) {
   return !!element && element?.offsetHeight > 0 && element?.hasAttribute('hidden') === false;
 }
 
-export function spanWrapper(nodeList: NodeListOf<ChildNode>): void {
+export function spanWrapper(nodeList: NodeListOf<ChildNode>, spanClass?: string): void {
   Array.from(nodeList)
     .filter(node => node.textContent && node.textContent.trim().length > 0 && node.nodeType === 3 && node.parentElement)
     .forEach(node => {
       const spanWrapper = document.createElement('span');
+      spanClass ? spanWrapper.classList.add('navigation-item-text') : null;
```
mathisscott: This seems very specific to one use case. Is there any way we could generalize this functionality? Also, do we have any tests for this?

packages/core/src/navigation/navigation-group.element.scss
```diff
@@ -0,0 +1,59 @@
...
+.group-items-wrapper {
+  width: 100%;
+}
+
+.navigation-group-start {
+  width: 100%;
+}
+
+:host([active]) {
+  ::slotted(cds-navigation-start:not([expanded])) {
```
mathisscott: We may be able to use a CSS part here. `::slotted()` isn't the easiest thing to work with...

packages/core/src/navigation/navigation-group.element.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * @desc
+   * The element id is used to update cds-navigation aria-activedescendent when a group is the current
+   * active item.
+   */
+  @property({ type: String, reflect: true, attribute: 'id' })
+  id: string = createId();
+
+  @internalProperty()
```
mathisscott: Usually we define a `type` for these...

packages/core/src/navigation/navigation-group.element.ts
```diff
@@ -0,0 +1,198 @@
...
+    return !!this.groupStart;
+  }
+
+  /**
+   * @desc
+   * The element id is used to update cds-navigation aria-activedescendent when a group is the current
+   * active item.
+   */
+  @property({ type: String, reflect: true, attribute: 'id' })
+  id: string = createId();
```
mathisscott: I don't know if we need `: string` here...

packages/core/src/navigation/navigation-group.element.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @desc
+   * The value of this property is passed down to start and item children. It is used to query for visible items when
+   * managing focus with key events in the root cds-navigation element.
+   *
+   * @private
+   */
+  // Note: eslint-disable  @typescript-eslint/no-unused-vars isn't ignoring the line
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  private expandedGroup = false;
```
mathisscott: Why was typescript/eslint giving you trouble here? This property is being used on line 181.

packages/core/src/navigation/navigation-group.element.ts
```diff
@@ -0,0 +1,198 @@
...
+      syncProps(this.groupStart, this, {
+        navigationGroupId: true,
+        expanded: true,
+        isGroupStart: this.isGroupStart,
+      });
+    }
+
+    this.groupItems.forEach(item => {
+      syncProps(item, this, {
+        expandedGroup: true,
```
mathisscott: Are we removing `expandedGroup` from the child items when the group element is collapsed?

packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -0,0 +1,67 @@
...
+  &:hover {
+    --background: inherit; // conflicts with the container element bg
+  }
+
+  ::slotted(a) {
+    cursor: not-allowed;
+  }
+}
+
+::slotted([cds-layout~='align:bottom']) {
```
mathisscott: CSS parts may be better than using `::slotted()`

packages/core/src/navigation/navigation-start.element.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.group-start-action {
+  color: inherit;
+  background: inherit;
+  border: 0;
+  font-size: var(--font-size);
+  height: var(--line-height);
+  padding: var(--navigation-padding);
+  text-transform: capitalize;
```
mathisscott: We may want to make these text-transforms a custom property. It's the kind of thing people will want to be able to customize.

packages/core/src/navigation/navigation.element.ts
```diff
@@ -0,0 +1,479 @@
...
+
+  @property({ type: String })
+  cdsMotion = 'on';
+
+  @event()
+  cdsMotionChange: EventEmitter<string>;
+
+  @event() protected expandedChange: EventEmitter<boolean>;
+
+  // TODO: why do I need this? What happens when I comment it out?
```
mathisscott: Did you mean to leave this here?

packages/core/src/navigation/navigation.element.ts
```diff
@@ -0,0 +1,479 @@
...
+   *    the cds-navigation element is not expanded
+   * 9. home key will move focus to the first focusable item
+   * 10. end key will move focus to the last focusable item
+   *
+   * We may need a way to let consumers mark elements and include them in the focusable elements, not sure how yet
+   *
+   * @param e
+   * @private
+   */
+  private keyboardNavigationHandler = (e: KeyboardEvent) => {
```
mathisscott: This message has a lot of consecutive conditional statement inside the `onKey` handlers. Have you considered doing an early return or a switch statement if only one conditional is intended to pass?

packages/core/src/navigation/navigation.element.ts
```diff
@@ -0,0 +1,479 @@
...
+    return returnHTML;
+  }
+
+  protected get visibleChildren(): FocusableElement[] {
+    return Array.from(this.allNavigationElements).filter(n => isVisible(n));
+  }
+
+  firstUpdated() {
+    if (this.rootNavigationStart) {
+      this.rootNavigationStart.addEventListener('click', () => {
```
mathisscott: Can people replace/ remove the start element here with something like an `ngIf`? If so, you might lose your click  handler...

packages/core/src/navigation/navigation.element.ts
```diff
@@ -0,0 +1,479 @@
...
+    // set all visible navigation elements to tabindex = -1
+    this.allNavigationElements.forEach(item => {
+      setAttributes(item, ['tabindex', '-1']);
+    });
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('focus', this.initAriaActiveDescendant);
+    this.removeEventListener('keydown', this.keyboardNavigationHandler);
```
mathisscott: It seems we are removing two out of four event listeners. Is that intentional?

packages/core/src/navigation/navigation.element.ts
```diff
@@ -0,0 +1,479 @@
...
+
+    if (props.has('expanded')) {
+      this.expandedRoot = this.expanded;
+    }
+
+    this.updateChildrenProps();
+  }
+
+  updateChildrenProps() {
+    if (this.navigationGroupItems) {
```
mathisscott, line 403: There's a lot of repeated work here. It might be nice to have a utility that does the mechanics of these code blocks in the `lit-element.ts` helpers. Something like...

`syncPropsForAllItems<T>(items: T[], source: Element, conditions: { [prop: string]: boolean }`

So then these calls would all look like...

`syncPropsForAllItems(this.navigationGroupItems, this, { groupItem: true });`

...and you can move the logic for testing for length or truthyness into the helper fn.

It would also decrease the conditional branches we need to cover in this component method.

packages/core/src/navigation/navigation.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+import '@cds/core/navigation/register.js';
+
+<Meta title="Components/Navigation" />
+
+# Navigation
+
+A sound navigation layout offers a high degree of discoverability and feedback, letting users know where they are at all
+times and ensuring they can easily get to where they want to go.
+
+Navigation enables discoverability and feedback forms users. It lets them know where they are at all times and ensures
```
mathisscott: `for users`? instead of `forms users`

packages/core/src/navigation/navigation.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+they can easily get where they want to go.
+
+Core navigation includes features that enables developers to create multiple navigation layouts from one component that
```
mathisscott: `Core navigation provides a foundation that will enable developers...`

packages/core/src/navigation/utils/index.ts
```diff
@@ -0,0 +1,113 @@
...
+export const DEFAULT_NAVIGATION_LAYOUT = 'vertical';
+
+export type FocusableElement = CdsNavigationItem | CdsNavigationStart;
+
+export function expandedGroupEvent(event: KeyboardEvent): boolean {
+  const element = event.target as HTMLElement;
+  return element?.parentElement?.tagName === 'CDS-NAVIGATION-GROUP' && element?.parentElement?.hasAttribute('expanded');
+}
+
+// Something wrong here
```
mathisscott: What is wrong here?

packages/core/src/navigation/utils/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import { Directions, setAttributes, setOrRemoveAttribute } from '@cds/core/internal';
+import { CdsNavigationGroup, CdsNavigationItem, CdsNavigationStart } from '../index.js';
+
+export const NAVIGATION_ITEM_TEXT = 'navigation-item-text';
+export const DEFAULT_NAVIGATION_LAYOUT = 'vertical';
+
+export type FocusableElement = CdsNavigationItem | CdsNavigationStart;
+
+export function expandedGroupEvent(event: KeyboardEvent): boolean {
```
mathisscott: I was confused  Reading these helpers in the component code. The name seems like it was calling an event. But all of us really doing is figuring out if an event target is part of an expand event. would it be better to name it something like `isExpandedGroupEvent`? Just so we know to expect a boolean and no other work coming out of it...

packages/core/src/navigation/utils/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  const element = event.target as HTMLElement;
+  return element?.parentElement?.tagName === 'CDS-NAVIGATION-GROUP' && element?.parentElement?.hasAttribute('expanded');
+}
+
+// Something wrong here
+export function expandedNavigation(event: KeyboardEvent): boolean {
+  const element = event.target as HTMLElement;
+  return element?.parentElement?.tagName === 'CDS-NAVIGATION' && element?.parentElement?.hasAttribute('expanded');
```
mathisscott: Same comment here as above... I wasn't expecting this to return to Boolean like this.

mathisscott (changes requested):

packages/core/src/internal/motion/animations/cds-navigation-open.ts
```diff
@@ -0,0 +1,20 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { TargetedAnimation } from '../interfaces.js';
+
+export const AnimationNavigationOpenName = 'cds-navigation-open';
+export const AnimationNavigationOpenConfig: TargetedAnimation[] = [
+  {
+    target: 'cds-navigation',
+    animation: [{ width: 'var(--collapsed-width)' }, { width: 'var(--expanded-width)' }],
```
bbogdanov, line 13: This one bothers me for a situation if we decide to change the name of that variable or remove it. Do you have something in mind that could guard us in case this variable is changed or removed?

hippee-lee: This is not working like I initially thought it was. I'll likely need to make changes to the cds-motion in order to support an animation that animates something like this ( a width that can be set by the consumer).

That said, the custom property name is part of the public API and that is the guard against changing or removing it. In this case, the two custom props are something consumers will likely want to override. The only way to allow that is with the name defined in the public API.

I may pull animations from the MVP depending on more discussion about this.

mathisscott: Why can't you pull the width/height from the element itself? The custom property is applying a width/height to something. Seems like it cuts out the middle man to go to the element the custom property applies to. The problem with custom properties being used in this way is that the lookup is expensive. More expensive than getting a boundingRect...

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: nope. Thanks!

packages/core/src/navigation/navigation-group.element.ts
hippee-lee: I updated all of the element annotations with the cssprop changes I made and added missing events.

packages/core/src/internal/utils/dom.ts
hippee-lee: I broke this out into a utility that can use a `childNodes` 😉 and add the class to it there. I still need to add a test for it but will add that once I finish making PR changes and update all specs at the same time.

packages/core/src/navigation/navigation-group.element.ts
hippee-lee: I defined a type for this and hasFocus because it is similar. How for do you think I should go with this? There are a bunch of boolean properties that could follow the same pattern. Is there something in the dev guidlines I missed that talks about this?

packages/core/src/navigation/navigation-group.element.ts
hippee-lee: Removed.

packages/core/src/navigation/navigation-group.element.ts
hippee-lee: The issue stems from the fact that I need this to sync the property down but I never actually use it anywhere. If I use it in the template the eslint error goes away because its getter is called. That rule seems to want the property to do both get and set or it complains. 

I updated the note to be more descriptive.

packages/core/src/navigation/navigation.stories.mdx
hippee-lee: updated

packages/core/src/navigation/navigation.stories.mdx
hippee-lee: updated

packages/core/src/navigation/utils/index.ts
hippee-lee: ¯\_(ツ)_/¯

Removed until / unless I can figure out what is wrong here.

packages/core/src/navigation/utils/index.ts
hippee-lee: I updated all of the util fn names to follow the `isSomethingSomethingEvent` pattern.

packages/core/src/navigation/utils/index.ts
hippee-lee: Updated names to be explicit and use the form `isSomethingSomethingExpanded`.

packages/core/src/navigation/navigation-group.element.scss
hippee-lee: I changed the code to sync enough _private_ info down to the start element so that it can use simpler css to identify when collapsed groups have an active item and visually apply the active bg color to the start element.

packages/core/src/internal/motion/animations/cds-navigation-open.ts
hippee-lee: Where would I pull the clientRect from? I can't run code on the element in the config object and the motion is added as an annotation. I'm not sure where to get the bounding rect and how to apply it to the config object.

packages/core/src/navigation/navigation.element.ts
hippee-lee: Nope. I updated this and we are now removing five event listeners in disconnectedCallback.

packages/core/src/navigation/navigation.element.ts
hippee-lee: I changed the template to handle this with an @slotchange event listener that calls a fn to add the event listener to the start element.

packages/core/src/navigation/navigation.element.ts
hippee-lee: I'm adding returns to the onKey listeners with multiple branches here for now. I'd like to make it simpler but am not 100% sure how yet. Switch (imo) makes it more verbose and harder to follow. 

My thinking is at the event level

packages/core/src/navigation/navigation.element.ts
hippee-lee: ty for this idea I added what is essentially a wrapper for syncProps but can iterate on the array of items.

packages/core/src/navigation/navigation-group.element.ts
hippee-lee: yes, it gets removed when updated.

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: I refactored this to not have to use slots and accomplish the same behavior.

packages/core/src/navigation/navigation-start.element.scss
hippee-lee: done

bbogdanov:
@hippee-lee  Can we change the commit message to `feat(core:navigation): mvp cds-navigation`?

packages/core/src/internal/index.ts
```diff
@@ -56,3 +56,8 @@ export {
 } from './motion/animations/cds-accordion-panel-open.js';
 export { AnimationHingeConfig, AnimationHingeName } from './motion/animations/cds-overlay-hinge-example.js';
 export { AnimationShakeConfig, AnimationShakeName } from './motion/animations/cds-component-shake.js';
+export {
+  AnimationNavigationGroupOpenConfig,
+  AnimationNavigationGroupOpenName,
+} from './motion/animations/cds-navigation-group-open.js';
```
mathisscott, line 62: Am I seeing this right? It looks like line 62 and 63 are duplicates.

packages/core/src/navigation/navigation-start.element.scss
```diff
@@ -0,0 +1,50 @@
...
+  --color: inherit;
+  --line-height: inherit;
+  --font-size: inherit;
+  --font-weight: inherit;
+  --text-transform: capitalize;
+}
+
+:host(:hover),
+:host([_has-focus]) {
+  // Needs to be #{$cds-scss-var}
```
mathisscott: Is this comment still valid?

packages/core/src/navigation/navigation-start.element.scss
```diff
@@ -0,0 +1,50 @@
...
+.icon-slot {
+  color: var(--color);
+}
+
+::slotted(span) {
+  color: var(--color);
+}
+
+:host([is-group-start][_active]:not([_expanded])) {
+  // Needs to be #{$cds-scss-var}
```
mathisscott: Same here…

packages/core/src/navigation/navigation.element.spec.ts
```diff
@@ -0,0 +1,387 @@
...
+      element.focus();
+    }
+
+    function arrowDownEvent(element: any) {
+      const ade = new KeyboardEvent('keydown', {
+        code: 'ArrowDown',
+        key: 'ArrowDown',
+        bubbles: true,
+      });
+      console.log(ade);
```
mathisscott: Console.log needed here?

packages/core/src/navigation/register.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '@cds/core/internal';
+import { CdsNavigation, CdsNavigationTagName } from './navigation.element.js';
+import { CdsNavigationGroup, CdsNavigationGroupTagName } from './navigation-group.element.js';
+import { CdsNavigationStart, CdsNavigationStartTagName } from './navigation-start.element.js';
+import { CdsNavigationItem, CdsNavigationItemTagName } from './navigation-item.element.js';
+
+import { ClarityIcons } from '@cds/core/icon/icon.service.js';
+import { angleIcon } from '@cds/core/icon/shapes/angle.js';
+import { angleDoubleIcon } from '@cds/core/icon/shapes/angle-double.js';
+import '@cds/core/icon/register.js';
```
mathisscott: Duplicate registration of the icon?

mathisscott: line 8 vs. line 26...

packages/core/web-test-runner.config.mjs
```diff
@@ -10,6 +10,13 @@ import { fromRollup } from '@web/dev-server-rollup';
 import execute from 'rollup-plugin-shell';
 import baseConfig from './web-dev-server.config.mjs';
 
+// import the browser launcher you want to use, chromeLauncher is the default
+import { chromeLauncher } from '@web/test-runner';
+// import { playwrightLauncher } from '@web/test-runner-playwright';
```
mathisscott: Do we still need these commented lines?

packages/core/src/internal/utils/dom.ts
```diff
@@ -155,10 +156,20 @@ export function spanWrapper(nodeList: NodeListOf<ChildNode>): void {
     .forEach(node => {
       const spanWrapper = document.createElement('span');
       node.after(spanWrapper);
+      spanWrapper.setAttribute(NAVIGATION_TEXT_WRAPPER, '');
```
mathisscott: This seems very specific to the navigation element. Can you think of a way to make it more generalized?

mathisscott (changes requested):

packages/core/src/internal/index.ts
hippee-lee: line 62 is the animation for cds-navigation-group, line 63 is the animation cds-navigation

packages/core/src/navigation/navigation-start.element.scss
hippee-lee: Yes, I missed it will convert these (there are a couple others too).

packages/core/src/navigation/navigation-start.element.scss
hippee-lee: Will update.

packages/core/src/navigation/navigation.element.spec.ts
hippee-lee: Nope.

packages/core/src/navigation/register.ts
hippee-lee: Will remove.

packages/core/web-test-runner.config.mjs
hippee-lee: Nope, I'll revert back to whats in next. I was learning how the new test harness worked.

packages/core/src/internal/utils/dom.ts
hippee-lee: Drat, i forgot about this one. I should be adding it in the elements but they are nodes, not elements. This is specific only to the item/start elements when they wrap button  / a text nodes. I'll move it where it is supposed to go and handle adding the attribute there like I meant to.

packages/core/src/internal/motion/animations/cds-navigation-open.ts
hippee-lee: As discussed I'll address this optimization in a future PR.

hippee-lee:
> @hippee-lee Can we change the commit message to `feat(core:navigation): mvp cds-navigation`?

@bbogdanov - I will amend the commit when I squash.

packages/core/web-test-runner.config.mjs
```diff
@@ -45,5 +52,5 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
       <script type="module" src="./web-test-runner.jasmine.js"></script>
     </head>
     <body></body>
-  </html>`
+  </html>`,
```
mathisscott, line 55: Is this file alright with this trailing comma?

mathisscott (approved):

packages/core/web-test-runner.config.mjs
hippee-lee: I'll see if I can revert it. Prettier might have formatted it after my other changes ...

hippee-lee: Prettier reformats it on any change, even if I revert the changes. I could manipulate the system to comit it and turn prettier off, commit and then turn prettier back on and commit that change but imo I think it is probably ok and easier to keep it.

mathisscott: That's fine. Just seemed weird.

coryrylan (approved):

PR #6041, opened 2021-06-09
feat(adoption/card): no-clr-card eslint rule

bbogdanov:


hippee-lee (approved):

Shijir (approved):

hippee-lee:
@bbogdanov - it looks like this is ready to merge. Is there a reason to keep the pr as wip?

bbogdanov:
> @bbogdanov - it looks like this is ready to merge. Is there a reason to keep the pr as wip?

We have to explore if we can do fixers. I'll work on the next week.

PR #6043, opened 2021-06-09
chore(commitlint): updating commitlint configuration

bbogdanov:
Adjusting the commitlint configuration for the new commit strategy.

Every commit has to follow the structure of `$type($project/$projectScope): $subject`

e.g. We added a feature to Button component in Clarity Core.
The commit message will look like `feat(core/button): added a feature x.y.z`

Video using `yarn commit`

https://user-images.githubusercontent.com/15037947/121713118-29a60600-cae5-11eb-9237-59a59f0b6a77.mp4


commitlint.config.js
```diff
@@ -1,60 +1,140 @@
...
+      lowerLimitWarning: 'below limit',
+    },
+    questions: {
+      type: {
+        description: "Select the type of change that you're committing:",
+        enum: {
+          a11y: {
+            description: 'Commit that provides Accessibility change',
+            title: 'Accessibility',
+            emoji: '🛠',
```
mathisscott: Emoji suggestion... ♿️

commitlint.config.js
```diff
@@ -1,60 +1,140 @@
...
+            description:
+              'Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)',
+            title: 'Builds',
+            emoji: '🛠',
+          },
+          ci: {
+            description:
+              'Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)',
+            title: 'Continuous Integrations',
+            emoji: '⚙️',
```
mathisscott: Emoji suggestion... 🤖

commitlint.config.js
```diff
@@ -1,60 +1,140 @@
...
+          },
+          chore: {
+            description: "Other changes that don't modify src or test files",
+            title: 'Chores',
+            emoji: '♻️',
```
mathisscott: Emoji suggestion... 🧹

commitlint.config.js
```diff
@@ -1,60 +1,140 @@
...
+          },
+          feat: {
+            description: 'A new feature',
+            title: 'Features',
+            emoji: '✨',
+          },
+          fix: {
+            description: 'A bug fix',
+            title: 'Bug Fixes',
+            emoji: '🐛',
```
mathisscott: Emoji suggestion... 🐞

commitlint.config.js
```diff
@@ -1,60 +1,140 @@
...
-  'stack-view',
-  'stepper',
-  'table',
-  'tabs',
-  'textarea',
-  'timeline',
-  'toggle',
-  'tooltip',
-  'tree-view',
-  'vertical-nav',
```
mathisscott, line 46: Is our plan to get rid of the scopes here?

scripts/commitlint/angularScopes.js
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+
+const components = [
+  'accordion',
+  'alert',
+  'badge',
+  'breadcrumb',
+  'button',
```
mathisscott, line 8: nvrmind. i found the scopes. 😊

mathisscott (approved):

mathisscott (approved):

hippee-lee (commented):
This looks great. I really like the details that it puts into commits, I like the separation of scopes and granularity and I ♥️ the emoji's! 

I do have a few questions. 

1. It seems like this requires devs to commit with `yarn commit`. What if they use their editor or already have a script on command line that adds and signs and codesigns each commit - will this still work? 
2. What happens if the user doesn't know about the commit script and commits something anyway? 
3. Since there is a dialog it seems like I would have to do this every time I make a commit - but what I'f I am just making lots of small commits during development wont this take up more time than needed for something that will eventually get rebased into the official commit?

bbogdanov:
> This looks great. I really like the details that it puts into commits, I like the separation of scopes and granularity and I ♥️ the emoji's!
> 
> I do have a few questions.
> 
> 1. It seems like this requires devs to commit with `yarn commit`. What if they use their editor or already have a script on command line that adds and signs and codesigns each commit - will this still work?
> 2. What happens if the user doesn't know about the commit script and commits something anyway?
> 3. Since there is a dialog it seems like I would have to do this every time I make a commit - but what I'f I am just making lots of small commits during development wont this take up more time than needed for something that will eventually get rebased into the official commit?

The new `yarn commit` is just a feature that we can use for simplicity for bigger commits but you are able to continue with what you are used to.  The commitlint will just guard the commit message to meet the requirements we set. If the message you add with `git commit -m` doesn't meet the requirements you will see a message in the console what is wrong and what you need to adjust.

bdryanovski (approved):

Jinnie (approved):

PR #6055, opened 2021-06-11
fix(button): unsubscribe from `ButtonInGroupService.changes`

arturovt:
This PR adds unsubscription from the `ButtonInGroupService.changes` inside the `crl-button-group`
component, since its `next` callback captures `this` and prevents the component from being GC'd.

## PR Type

- [x] Bugfix

You may see there's no distance from the GC root but component is still not getting GC'd:

image

- [x] No


hippee-lee (approved):
lgtm - thanks for adding this!!

mathisscott (approved):

Shijir (approved):

PR #6058, opened 2021-06-11
feat(combobox): reduce change detection cycles

arturovt:
This PR reduces change detection cycles for the combobox by replacing
`HostListener` with `Renderer.listen` outside of the zone. We don't need to
run change detections when any keyboard button is pressed or when the multiselect
mode is off. Note that the `HostListener` wraps the actual listener under the hood
into the internal Angular function which runs `markDirty()` before running the actual
listener (the decorated class method).

## PR Checklist

## PR Type

- [x] Feature

- [x] No

### DIFF
--- a/packages/angular/golden/clr-angular.d.ts
+++ b/packages/angular/golden/clr-angular.d.ts
@@ -326,7 +326,7 @@ export declare class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContai
     smartPosition: ClrPopoverPosition;
     textbox: ElementRef;
     trigger: ElementRef;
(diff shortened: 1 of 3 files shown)


hippee-lee:
There is a similar PR for datagrid here: https://github.com/vmware/clarity/pull/6059. @Jinnie  I left a comment there with my thoughts and while I am not opposed to this change I wonder if its a good idea for the angular components at this time. What do you think wrt the combo box since there could be more of them on a page at one time?

Jinnie:
The code looks great I think, it's well explained and documented. I like it. In the same time, I share Matt's concern that the expected value should surpass the extra maintenance and readability costs. Especially since Clarity Angular is in predominantly maintenance phase now, we don't have active issues about the keyboard navigation performance and we would like to minimize the risk of regressions. Would be nice to hear some more thoughts from the team before we decide to merge it or not.

Shijir:
Hi @arturovt, we have moved our Angular source to `angular` branch. Could you re-submit this PR against `angular` branch and close this one?

bbogdanov:
@arturovt Any chance you could get back to this soon?

hippee-lee:
Hi @arturovt - I adopted and manually migrated the changes into a PR for the angular branch. Thank you for another contribution!

This pr is being closed in favor of https://github.com/vmware/clarity/pull/6193

PR #6059, opened 2021-06-11
feat(datagrid): reduce change detection cycles for `clr-dg-detail`

arturovt:
This PR reduces change detection cycles for the `clr-dg-detail` by replacing
`HostListener` with `Renderer.listen` outside of the zone. We don't need to run
change detections when any keyboard button is pressed. Note that the `HostListener`
wraps the actual listener under the hood into the internal Angular function which runs

## PR Type

- [x] Feature

- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-detail.ts
+++ b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-detail.ts
@@ -1,12 +1,13 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, ContentChild, HostListener } from '@angular/core';
+import { Component, ContentChild, NgZone, OnDestroy, Renderer2 } from '@angular/core';
 import { DetailService } from './providers/detail.service';
 import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 import { ClrDatagridDetailHeader } from './datagrid-detail-header';
+import { ESC } from '../../utils/key-codes/key-codes';
 
 @Component({
   selector: 'clr-dg-detail',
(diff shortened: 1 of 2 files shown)


packages/angular/projects/clr-angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -31,13 +32,34 @@ import { ClrDatagridDetailHeader } from './datagrid-detail-header';
...
+    renderer: Renderer2
+  ) {
+    ngZone.runOutsideAngular(() => {
+      this.unlisten = renderer.listen('document', 'keyup', (event: KeyboardEvent) => {
+        // We could've actually used `render.listener('document', 'keyup.esc')`, and that would definitely work.
+        // Unfortunately, Angular delegates the event handling to the `KeyEventsPlugin` when it meets the `.` (dot)
+        // in the event name. The `KeyEventsPlugin` calls `ngZone.runGuarded`, which triggers change detection anyway
+        // (no matter if we added event listeners in the root zone).
+        // In that case, the change detection will be run only when the user clicks the `esc` button.
+        if (event.keyCode === ESC) {
```
hippee-lee, line 53: `event.keyCode` is deprecated for `event.code`.

hippee-lee (changes requested):
I'm having a hard time understanding how this is better for the datagrid. There is usually only one datagrid on a page at a time so there are no likely to be additional change detection cycles triggered for other datagrid instances. Am I missing something re: this optimization technique for Angular apps? If I understand it correctly is usually applied when there are many of the same elements on each page, each with their own listener. 

From a code maintenance and readability standpoint  It replaces one line of code (An angular tested decorator that encapsulates all of the listening behavior) with the renderer and some custom logic for unregistering the Renderer fn.

Is there a way to see how many cycles this is reducing and what the performance increase is?
I would rather not make substantiative changes to datagrid at this point in its lifecycle (being more in maintenance mode as we await the `@cds/core` datagrid under active development). 

I am not opposed to making the change but imo it should substantially improve performance in order to justify the extra code. Also it would need to include a test to cover the change.

arturovt:
Hey @hippee-lee . Not sure what you mean by:

> there are no likely to be additional change detection cycles triggered for other datagrid instances

The change detection is triggered for **all** active components on the page. If there're multiple datagrid instances, then the change detection will be triggered for multiple datagrid instances caused by a single `keyup` event (no matter what keyboard button is pressed, the change detection will be always run).

hippee-lee:
Hi @arturovt - thank you for the ideas to optimize the angular code. I want to make sure you understand why my response is asking _how much improvement_ as a predicate for these types of changes. When we first started the system back in 2016 Angular was the focus for implementing the components for a variety of reasons - some technical and some non-technical. As the design system has matured and web components has become mainstream we have shifted efforts into upgrading the Clarity components in a framework independent way so they can be used anywhere modern web-standards and API's are available. This is why the Angualar components are in maintenance mode. 

That said, **I appreciate the effort and contribution** so my full line of thinking is related to this thought:
> There is **usually** only one datagrid on a page at a time so there are not likely to be additional change detection cycles triggered for other datagrid instances.

It is not often that you have multiple datagrids on a page, and even more less likely that they all implement detail pane. 

As I mentioned before, given the maintenance state of the datagrid codebase, its size and its complexity  I am trying to weigh:
> **the amount of improvement** for apps that implement 1,2,3, etc datagrids with detail pane 

vs

> code changes to a stable and tested but not completely optimized codebase

Can you provide any insight into **how much performance improvement** this optimization brings? eg - is it possible to calculate the performance impact of 1 round of change detection for a datagrid with detail pane?

bbogdanov:
@arturovt Any chance you could get back to this soon?

hippee-lee:
Thanks for the contribution @arturovt. I am going to close this for the reason mentioned above, 1) ClrDatagrid is in maintenance mode and it is not clear what the performance optimization will be for an app with these changes.

PR #6061, opened 2021-06-14
build: github actions for angular branch

bdryanovski:
- Add Github Actions to build the angular package and dev app
 - Change git ignore paths
 - Format generate news script
 - Change build order for the website
 - Added back few packages back like netlify and dotenv


bdryanovski:
The best way to test this is to run this before any command

```bash
git clone -fxd
```

Anf after that to run `yarn run build:ci` and so on...

package.json
```diff
@@ -110,7 +113,7 @@
     "karma-notify-reporter": "^1.3.0",
     "karma-parallel": "^0.3.1",
     "karma-typescript": "^5.5.1",
-    "netlify-lambda": "^2.0.6",
+    "netlify-lambda": "1.4.5",
```
hippee-lee, line 116: Why was this downgraded to the older version?

projects/website/generate-news.js
```diff
@@ -105,11 +105,7 @@ function generateRoutes() {
     }
   });
   var final_routes = ROUTES_TEMPLATE.replace('${routes}', routes);
-  fs.writeFileSync(
-    FINAL_ROUTES_PATH,
-    `// tslint:disable
-  ${final_routes}`
-  );
+  fs.writeFileSync(FINAL_ROUTES_PATH, `// tslint:disable\n${final_routes}`.trim());
```
hippee-lee: This looks like just a formatting change. Did something change between branches that made prettier re-format this line?

Shijir: I don't think you need `// tslint:disable` now as we don't use it any longer. We should resort to eslint-disable if necessary.

Shijir (approved):
Thanks for working on this!

package.json
bdryanovski: It's not a downgrade it the same version we have on the `next` branch. I think this package was just updated by mistake on the `angular` branch. It must not be there without the `netlify` package anyway.

bbogdanov (approved):

PR #6073, opened 2021-06-15
build: updating adoption guide

bdryanovski:
Replacing the way we create demos inside the adoption guide - Now the templates for Stackblitz are based on top of `node`. 

Providing this type of templates (at the moment)

* Core - Angular app with Core components only
* Angular - Angular app with Angular components only (@clr/angular)
* Hybrid -  Angular app with Angular component and Core components into a single app (@clr/angular & @cds/angular & @cds/core)

Fully reworking how Stackblitz service is working (need some more polishing but with the next PR)

Adding basic demo apps building links on the "Getting Started" page for quick testing.

DEMO: https://60c8a88362fef5fff180a92f--vmware-clarity.netlify.app/adoption/getting-started


mathisscott (approved):

gnomeontherun (changes requested):
I want us to move the adoption guide out of `next` and into its own branch so we can deploy it as a standalone subdomain.

bdryanovski:
We need to discuss this @gnomeontherun  and I think that this is not part of the scope of the changes above or at least for now. 

We could talk more about that next week when we discuss what and how the adoption guide will be promoted and publicly available

gnomeontherun (approved):

PR #6079, opened 2021-06-16
feat: add description list element styling

ashleyryan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?

Adds support for `cds-list` on a dl element, as well as layout styling with dl/dd/dt

- [ ] Yes
- [x] No

## Other information

This is still in draft, but I wanted to get some early feedback. 

A couple of open questions:
- Should this be part of the existing `list` files/css?
- If it should remain separate, should it use a different attribute than `cds-list`?

Screenshot:

image

### DIFF
--- a/packages/core/src/list/list.scss
+++ b/packages/core/src/list/list.scss
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
@@ -88,5 +88,17 @@ ul[cds-layout~='horizontal'],
 ol[cds-layout~='horizontal'] {
   li {
     list-style-position: inside;
+  }
+}
+
+dl[cds-list] {
+  margin: 0;
+
+  dt {
+    font-weight: bold;
+  }
+
+  dd {
+    margin-left: 0;
   }
 }
--- a/packages/core/src/list/list.stories.mdx
+++ b/packages/core/src/list/list.stories.mdx
@@ -4,9 +4,9 @@ import { Meta, Story, Preview } from '@web/storybook-prebuilt/addon-docs/blocks.
 
 # Lists
 
-Lists in Clarity Core come in three varieties – ordered, unordered, and
-unstyled. Any "compact" style variations for list display is now handled through
-Clarity Core layouts.
+Lists in Clarity Core come in four varieties – ordered, unordered,
+unstyled, and description lists. Any "compact" style variations for
+list display is now handled through Clarity Core layouts.
 
 ## Installation
 
@@ -54,4 +54,16 @@ To use the list styles import the global CSS.
 
 <Preview>
   <Story id="stories-list--dark-theme" />
+</Preview>
+
+## Description List - dl/dd/dt
+
+<Preview>
+  <Story id="stories-list--description-list" />
+</Preview>
+
+## Description List Custom Spacing
+
+<Preview>
+  <Story id="stories-list--custom-space-description-list" />
 </Preview>
--- a/packages/core/src/list/list.stories.ts
+++ b/packages/core/src/list/list.stories.ts
@@ -489,3 +489,27 @@ export function darkTheme() {
     </div>
   `;
 }
+
+/** @website */
+export function descriptionList() {
+  return html`<dl cds-list>
+    <dt>Name</dt>
+    <dd>Description</dd>
+    <dt>Name</dt>
+    <dd>Description</dd>
+  </dl>`;
+}
+
+/** @website */
+export function customSpaceDescriptionList() {
+  return html`<dl cds-list cds-layout="vertical gap:md">
+    <div>
+      <dt>Name</dt>
+      <dd>Description</dd>
+    </div>
+    <div>
+      <dt>Name</dt>
+      <dd>Description</dd>
+    </div>
+  </dl>`;
+}


packages/core/src/description-list/description-list.scss
```diff
@@ -0,0 +1,35 @@
+@import './../styles/tokens/generated/index';
+
+@mixin renderGapStyles($sizes: null) {
```
ashleyryan, line 3: oops, these aren't both supposed to be in there

packages/core/src/description-list/description-list.scss
```diff
@@ -0,0 +1,24 @@
...
+
+  dd {
+    margin-left: 0;
+  }
+}
+
+@mixin renderGapStyles($sizes: null) {
+  @each $size in $sizes {
+    dl[cds-list][cds-layout~='vertical'][cds-layout*='gap:#{$size}'] {
+      gap: 0;
```
ashleyryan: if you try and wrap your `dd` and `dt` in a div like the horizontal example and use a `vertical` layout with spacing, this will break unfortunately and you won't get any spacing. 

So it's weird to support the `div` for horizontal, but not vertical.

ashleyryan: I suppose I could remove all of the complexity and say "if you want a gap between items, wrap in a div" and then `cds-layout` would work out of the  box

astorije:
Clarity, what about a default gap (md? sm? which do you prefer?) between dt+dd pairs? It seems like a decent thing to have by default, and then `<dl cds-list="unstyled">` would remove both the gap _and_ the `font-weight` on the `dt`

astorije:
> * Should this be part of the existing `list` files/css?

I think they should be combined, `ul`/`ol`/`dl` kinda belong with each other IMO and I think using `cds-list` makes sense here.

packages/core/.storybook/public/demo.css
```diff
@@ -3,6 +3,7 @@
 @import './../../dist/core/global.min.css';
 @import './../../dist/core/styles/theme.dark.min.css';
 @import './../../dist/core/styles/theme.low-motion.min.css';
+@import './../../dist/core/description-list/description-list.min.css';
```
coryrylan: I agree with @astorije this should be combined as part of the existing `cds-list` API.

packages/core/src/description-list/description-list.scss
```diff
@@ -0,0 +1,24 @@
+@import './../styles/tokens/generated/index';
+
+dl[cds-list] {
```
coryrylan: Set margins to 0 on the `dl` so there is no wrapping white space.

packages/core/src/description-list/description-list.scss
coryrylan: Let's recommend wrapping the `dd` and `dt` in a div for vertical layouts as well. This will keep the layout behavior consistent when using `cds-layout`. Also ensures the with theming/responsive the gap behavior is the same. The vertical layouts seem to work as long as they are wrapped. https://stackblitz.com/edit/js-1qackg?file=index.html

packages/core/src/description-list/description-list.stories.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+import { html } from 'lit-html';
```
coryrylan: This import should be `lit` instead of `lit-html`

coryrylan (changes requested):

packages/core/.storybook/public/demo.css
mathisscott: +1

packages/core/src/description-list/description-list.stories.ts
```diff
@@ -0,0 +1,52 @@
...
+    <dd>Description 2</dd>
+    <dt>Name</dt>
+    <dt>Name 2</dt>
+    <dd>Description</dd>
+  </dl>`;
+}
+
+/** @website */
+export function horizontalList() {
+  return html`<dl cds-list cds-layout="horizontal gap:xl">
```
mathisscott: The thing about a horizontal layout is that I would expect it to be like the horizontal layout of forms. Such that the `dt` and the `dd` would be left and right, instead of stacked on top of one another.

The trick is (as we've seen with forms) setting it such that they all share a common width for the left (key). 

Maybe @coryrylan would have some ideas on how this could be accomplished with a custom property or two. Maybe a shared `--horizontal-dd-width` that can be set? I'm not sure.

I'm also happy to ditch the horizontal definition list for now too. It's possible that we could add it later if needed.

mathisscott (changes requested):

ashleyryan:
Updated screenshot: 

image

packages/core/src/description-list/description-list.stories.ts
ashleyryan: When I spoke with Scott about it the other day on slack, the horizontal was basically extra credit. Not my call, but I don't think it's necessary to support the horizontal layout you describe - also because it would probably break down if someone had more than one `dd` or `dt` next to each other.

mathisscott:
@colinreedmiller 
Does this meet expectations UX-wise? The `dl` definition list largely follows the specification for forms. We are punting on the horizontal layout of having the `dt` and `dd` appear on the same line. Maybe the team can revisit that in the near future to lock down responsive behaviors, etc.?

mathisscott (approved):

colinreedmiller:
> @colinreedmiller
> Does this meet expectations UX-wise? The `dl` definition list largely follows the specification for forms. We are punting on the horizontal layout of having the `dt` and `dd` appear on the same line. Maybe the team can revisit that in the near future to lock down responsive behaviors, etc.?

I agree with waiting on horizontal, if it is needed we will hear about it and can address adding it based on requirements - seems like an odd case offhand. Same line for dt and dd is not the sort of thing I'd recommend anyway simply because of usability / harder to scan and read. What I see in vertical with space and layouts is good.

colinreedmiller (approved):

coryrylan (approved):

PR #6092, opened 2021-06-21
fix: remove no-longer valid url redirects

Shijir:

- Removed invalid redirects in `netlify.tom`
- Ensured that the angular app (website) has control over routing instead of netlify. Without these change, users won't be able to access the component pages directly. For example: https://angular.clarity.design/documentation/alerts.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,82 +5,8 @@
   publish = ""
   environment = { CI = "true" }
 
-# Legacy Website Redirects
-
-# getting started
+# With this, the app has full control over the routing logic
 [[redirects]]
-  from = "/documentation/get-started"
-  to = "/get-started/"
-  status = 301
-
-# news -> releases
-[[redirects]]
-  from = "/news/*"
-  to = "/releases/"
-  status = 301
-
-# community -> (re-direct to homepage)
-[[redirects]]
-  from = "/community"
-  to = "/"
-  status = 301
-
-# Icons
-[[redirects]]
-  from = "/icons"
-  to = "/foundation/icons/"
-  status = 301
-
-# foundation pages
-[[redirects]]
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -5,82 +5,8 @@
...
-  status = 301
-
-[[redirects]]
-  from = "/documentation/typography"
-  to = "/foundation/typography/"
-  status = 301
-
-# Redirects for all images to CDN
-[[redirects]]
-  from = "/images/*"
```
hippee-lee, line 72: angular application still has an images folder right? I want to make sure we don't still need a redirect rule for images here like we do for the vuepress site.

hippee-lee (approved):
This looks correct to me for an isolated Angular app. These rules were for people who bookmarked the original angular app routes and were now coming to the vuepress routes. 

Do you think its worth looking at the git history for the state of the netlify.toml file before the vuepress app was used as the website, just to make sure we aren't forgetting anything? 

There should probably be a complementary PR for the next branch that updates these redirects there to redirect people who bookmarked `/documentation/get-started` to the angular sub-bomain, `https://angular.clarity.design/documentation/get-started`. If that is even possible. I'm not sure off-hand if we can make something like that work or not tho.

Shijir:
@hippee-lee  Yeah, I will follow up this PR with another PR in the `next` branch that addresses the old URL redirects.

As for the images, as we discussed, the images are stored within the website source directly so we don't need redirects there.

Jinnie (approved):

PR #6097, opened 2021-06-23
fix: core update service to global state service

### LINKED ISSUE #5318: core: singleton services
We need a way to store data globally for Clarity core services as singleton instances. Candidates for this include: 

- Icon Registry  
- Internationalization strings
- Keycodes

We'll leverage the `window` object for this.

### LINKED ISSUE #6095: core: i18n sync
## Describe the bug

When switching the i18n language throught the i18n service the components do not update with the latest string values.

## How to reproduce

Switch the i18n here and inspect internal aria-labels not updating https://clarity.design/storybook/core/?path=/story/foundation-internationalization--page 

## Expected behavior

When updating the i18n options the components should reflect the updates.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.x

**Framework:**

- [ ] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: [e.g. MacBook]

## Additional notes

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There are cases where we relied on DOM behavior to check for global state such as focus trap indicators. This can be an issues with VDOM libraries like react.

## What is the new behavior?
This PR updates our internal services to use a single shared global state object to keep things in sync.

- [ ] Yes
- [x] No

closes https://github.com/vmware/clarity/issues/6095
closes https://github.com/vmware/clarity/issues/5318


packages/core/src/internal/utils/global.ts
```diff
@@ -50,26 +41,41 @@ export function setupCDSGlobal() {
...
+  if (!window.CDS._stateProxy) {
+    window.CDS._stateProxy = true;
+    window.CDS._state = new Proxy(window.CDS._state, {
+      set: (target: any, key: string, value) => {
+        const detail = { key, prev: (window.CDS._state as any)[key], current: value };
+        target[key] = value;
+        document.dispatchEvent(new CustomEvent('CDS_STATE_UPDATE', { detail }));
+        return true;
+      },
+    });
```
coryrylan, line 58: todo: add test here and explainer in comment

The proxy wraps the state object to allow us to trigger events when any of the state object properties are updated. This allows services to know when the state was updated even if there are duplicates or multiple versions. The i18n and icon services leverage this to update/rerender components when new values are added.

packages/core/src/internal/services/focus-trap-tracker.service.spec.ts
```diff
@@ -6,220 +6,160 @@
...
   describe('getDocroot(): ', () => {
     it('should return something', () => {
-      const myDocroot: HTMLElement = service.getDocroot();
-      expect(myDocroot).not.toBeNull('Docroot node should exist');
-      expect(myDocroot.tagName.toLowerCase()).toBe('html', 'should return the documentElement(html) node');
+      expect(document.documentElement).not.toBeNull('Docroot node should exist');
+      expect(document.documentElement.tagName.toLowerCase()).toBe(
+        'html',
+        'should return the documentElement(html) node'
+      );
```
coryrylan: todo: remove test, I think this can be removed since we no longer track on the doc root but in the state object

packages/core/src/internal-components/overlay/overlay.element.spec.ts
```diff
@@ -67,11 +67,11 @@ describe('Overlay helper functions: ', () => {
     });
 
     it('identifies top-most overlay', async () => {
-      const fttIds = FocusTrapTracker.getTrapIds();
+      const elements = FocusTrapTrackerService.getTrapElements();
       await componentIsStable(overlay);
 
-      expect(overlayIsActive(fttIds.pop())).toBe(true, 'the top overly is not active');
-      fttIds.forEach(item => expect(overlayIsActive(item)).toBe(false, 'a hidden overlay is active'));
+      expect(overlayIsActive(elements.pop().focusTrapId)).toBe(true, 'the top overly is not active');
```
mathisscott: I think i had the text here wrong? `the top overly is not active`

Our test is expecting the top overlay to be active... 

Same mix up with the line below it...

packages/core/src/internal/services/focus-trap-tracker.service.spec.ts
mathisscott: It can

packages/core/src/internal/services/focus-trap-tracker.service.ts
```diff
@@ -1,84 +1,43 @@
...
-      // the function in the conditional handles all nil references. zero chance of null making it through here.
-      const myAttribute = docroot.getAttribute(CDS_FOCUS_TRAP_DOCUMENT_ATTR) || '';
-      // TS forcing us to write an unreachable codepath. this is where monads would be useful.
-      return myAttribute === '' ? [] : myAttribute.split(' ');
-    } else {
-      return [];
-    }
+  static setTrapElements(elements: { focusTrapId: string }[]): void {
+    elements.length
+      ? document.documentElement.setAttribute(CDS_FOCUS_TRAP_DOCUMENT_ATTR, '')
```
mathisscott: I don't think this will work in React. When I was investigating the modal backdrop-click and esc keypress not firing, this is where it would die. `documentElement` was not defined. Maybe you've done something here differently such that it works now. But I was moving towards `document.querySelector("html")` because that was coming back defined in React.

packages/core/src/internal/utils/global.ts
```diff
@@ -11,35 +11,26 @@ import { LogService } from '../services/log.service.js';
...
-  reactVersion?: string | undefined;
-  vueVersion?: string | undefined;
-  environment: {
-    production: boolean;
-  };
 }
 
 export interface CDSState {
-  focusTraps?: string[];
+  focusTrapRegistry: Readonly<{ focusTrapId: string }[]>;
```
mathisscott: Does it make sense to have all the items here be `*Registry`? Maybe this interface isn't `CDSState` but `CDSRegistries`? The keys in the interface seem to be compensating for the name or architecture of `CDSState` not being descriptive here.

packages/core/src/internal/utils/global.ts
```diff
@@ -50,26 +41,41 @@ export function setupCDSGlobal() {
...
-      angularVersion: getAngularVersion(false),
-      angularJSVersion: getAngularJSVersion(false),
-      reactVersion: getReactVersion(false),
-      vueVersion: getVueVersion(false),
-      loadedElements: window.CDS._loadedElements,
-    };
-    return log;
+function intializeCDSStateProxy() {
+  if (!window.CDS._stateProxy) {
+    window.CDS._stateProxy = true;
```
mathisscott: It seems like `_stateProxy` isn't a proxy but instead notes whether the `_state` object has been proxied?

I was expecting this property to be the proxy at first. Maybe we can update that to be `_stateProxied` or `_isStateProxied`?

packages/core/src/internal/utils/global.ts
```diff
@@ -50,26 +41,41 @@ export function setupCDSGlobal() {
...
+        target[key] = value;
+        document.dispatchEvent(new CustomEvent('CDS_STATE_UPDATE', { detail }));
+        return true;
+      },
+    });
   }
+}
 
-  return void 0;
+function getVersion() {
```
mathisscott: If we are focused here on returning the version info, do we need to include the state too?

mathisscott (changes requested):

packages/core/src/internal/utils/global.ts
coryrylan: I was thinking for debugging issues getting a snapshot of the state might help triage. Maybe the name here could be different? Maybe `getDetails()` and `logDetails()`?

mathisscott: I think either would be better. thanks!

packages/core/src/internal/utils/global.ts
coryrylan: Yeah you are right.  I don't think they all need to be the same. The focusTrap isnt really a registry. Its not registering anything like icons/i18n where they are registered for later lookups. Its really just in memory state.

packages/core/src/internal/services/focus-trap-tracker.service.ts
coryrylan: I wish I could find a way to test this consistently in react.

packages/core/src/internal/utils/global.ts
coryrylan: updated

packages/core/src/internal/utils/global.ts
coryrylan: added

mathisscott (approved):

bbogdanov:
@coryrylan Could we change the commit message to something like `fix(core): global state service sync`? I believe also adding the content of `What is the current behavior?` and `What is the new behavior?` as a description will be helpful.

coryrylan:
Fixed merge conflict and fixed commit message

hippee-lee (approved):

PR #6101, opened 2021-06-24
fix: core control label/message alignments

### LINKED ISSUE #5891: Ability to nest a form field inside the label of a checkbox or radio button
Results of a conversation on this issue follow below:

Following up a conversation on Slack, I would like to request the ability to design nested controls as such:

At CloudHealth, we have a lot of form controls following this format, often having to select an option and provide 1 or 2 settings for that option. We are currently converting our platform to Clarity and encountering a lot of those patterns. Some definitely need rethinking the page/form, but in many cases it makes sense UX wise.

I was hoping that the current components would let us do something like this, but each attempt had one small thing preventing it. I’ve been trying all sorts of combinations using helper components such as `CdsInputGroup` and `CdsControl` but so far haven’t been able to make the input even just show up at all.
In a nutshell, to produce the mock above, that third row cannot be a `CdsControl` because that will only let me use 1 control (vs. 2: checkbox + input), and it cannot be a `CdsInputGroup` (which lets me have more than 1 control) because the `CdsCheckboxGroup` only fills slots with `cds-control` elements while the `CdsInputGroup` is a `cds-control-group`).
Would you be open to consider this in Clarity Core?

_Originally posted by @astorije in https://github.com/vmware/clarity/discussions/5879_

===

Of interest here is that the primary reason this is not viable right now is that the checkbox group orientation is being forced on the input inside the label. It's hard to gauge the difficulty of preventing the group from doing that. But that is the reason this won't work today.

### LINKED ISSUE #6053: cds: long message in cds-control-message switches form layout to vertical
## Describe the bug

A long message in a `cds-control-message` switches form layout from horizonal to vertical. If the `cds-control-message` message is longer after a failed input validation the input will switch between horizontal and vertical layout.

image

## How to reproduce

https://stackblitz.com/edit/clarity-light-theme-v5-upxtde?file=src%2Fapp%2Fapp.component.html

See the stackbliz demo

## Expected behavior

A horizontal form layout shouldn't become vertical if `responsive = false` and the `cds-control-message` contains a long message.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.3.0

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
Angular 11

**Device:**

- Type: PC
- OS: Windows 10
- Browser: Chrome
- Version: 91

## Additional notes

Changing `flex-flow: row wrap` to `flex-flow: row` seems like a good workaround but I haven't been able to apply it outside Chrome's dev console. (Can't figure out the correct `::ng-deep` selectors)

image

### LINKED ISSUE #6098: cds-file poor long file names handling
## Describe the bug

When selecting a file with a long filename browse button shows the whole string breaking layouts.

## How to reproduce

1.  Go to https://clarity.design/core-components/file/
2.  Click on any browse button
3.  Select a file with about 100 or more chars long filename
4.  See how browse button overflows other content

screenshot

## Expected behavior

Mimic native html file input behaviour.

screenshot2

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The cds-control component was missing a default width for both label and control messages. This was causing several related bugs when wrapping content in responsive form mode.

closes: https://github.com/vmware/clarity/issues/6098
closes: https://github.com/vmware/clarity/issues/6053
closes: https://github.com/vmware/clarity/issues/5891

## What is the new behavior?
This aligns all control layouts to the same default label/control/message width. Inline controls now correctly show missing cds-messages when placed within parent inline control groups. This adds a new `--control-width` property to better control custom widths of the inputs.

- fix label max-width
- fix control message max width
- fix missing inline group message
- fix long label wrap issue with responsive
- fix missing control width property
- fix file button control width overflow

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/file/file.element.scss
+++ b/packages/core/src/file/file.element.scss
@@ -1,9 +1,26 @@
-// Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
 // This software is released under MIT license.
 // The full license information can be found in LICENSE in the root directory of this project.
 
 @import './../styles/tokens/generated/index';
 
+$fallback-width: calc(#{$cds-global-layout-space-xxl} * 2);
+
+:host {
+  --control-width: #{$fallback-width};
+}
+
 ::slotted(input) {
   display: none !important;
+}
+
+.input-container {
+  min-width: initial;
+}
+
+cds-button span {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  padding: $cds-global-space-4 0;
+  max-width: var(--control-width);
 }
--- a/packages/core/src/file/file.element.ts
+++ b/packages/core/src/file/file.element.ts
@@ -48,7 +48,7 @@ export class CdsFile extends CdsControl {
       <div>
         <cds-button size="sm" action="outline" @click="${() => this.label.click()}" ?disabled=${this.disabled}>
           <cds-icon shape="folder" aria-hidden="true"></cds-icon>
-          ${this.buttonLabel}
+          <span>${this.buttonLabel}</span>
         </cds-button>
         ${this.clearFilesControlTemplate}
       </div>
--- a/packages/core/src/forms/control/control.element.scss
+++ b/packages/core/src/forms/control/control.element.scss
@@ -4,9 +4,24 @@
 
 @import './../../styles/tokens/generated/index';
 
+$fallback-width: calc(#{$cds-global-layout-space-xxl} * 2);
+
 :host {
   display: block;
   width: 100%;
+  --control-width: #{$fallback-width};
+}
+
+:host([layout='compact']:not([control-width='shrink'])) .input-container {
+  min-width: var(--control-width);
+}
+
+:host(:not([layout*='vertical']):not([control-width='shrink'])) .input-message-container {
+  flex-basis: $fallback-width;
+}
+
+:host([layout='compact']:not([control-width='shrink'])) ::slotted(cds-control-message) {
+  --max-width: #{$fallback-width};
 }
 
 .input-container {
--- a/packages/core/src/internal/services/global.service.ts
+++ b/packages/core/src/internal/services/global.service.ts
@@ -10,7 +10,7 @@ import { LogService } from './log.service.js';
 export class GlobalState {
   static get state(): Readonly<CDSState> {
     setupCDSGlobal();
-    return window?.CDS?.state;
+    return window?.CDS?.state as CDSState;
   }
 
   static getValue(key: string): any {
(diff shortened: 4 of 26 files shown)


packages/core/src/internal/services/global.service.ts
```diff
@@ -10,7 +10,7 @@ import { LogService } from './log.service.js';
 export class GlobalState {
   static get state(): Readonly<CDSState> {
     setupCDSGlobal();
-    return window?.CDS?.state;
+    return window?.CDS?.state as CDSState;
```
coryrylan, line 13: This was throwing a error on next, this fixes the error or the global service PR will

packages/core/src/file/file.element.scss
```diff
@@ -7,3 +7,14 @@
 ::slotted(input) {
   display: none !important;
 }
+
+.input-container {
+  min-width: initial;
+}
+
+cds-button span {
```
mathisscott, line 21: This may be okay for now. But the ask internally was for the name to be ellipsized in the middle such that `reallyLongFileName.png` would become `reallyLon...me.png` or something like that.

packages/core/src/file/file.element.ts
```diff
@@ -48,7 +48,7 @@ export class CdsFile extends CdsControl {
       <div>
         <cds-button size="sm" action="outline" @click="${() => this.label.click()}" ?disabled=${this.disabled}>
           <cds-icon shape="folder" aria-hidden="true"></cds-icon>
-          ${this.buttonLabel}
+          <span>${this.buttonLabel}</span>
```
mathisscott, line 51: should we add an aria label here that includes the full text of the buttonLabel? In the event it gets truncated, what will happen with screen readers?

mathisscott (approved):

bbogdanov:
@coryrylan Could we change the commit message to something like - `fix(core:control): label message alignment`?

packages/core/src/forms/control/control.element.scss
```diff
@@ -4,11 +4,31 @@
 
 @import './../../styles/tokens/generated/index';
 
+$fallback-width: calc(#{$cds-global-layout-space-xxl} * 2);
+
 :host {
+  --control-width: #{$fallback-width};
```
coryrylan: todo: when control width shrink this likely needs to be set to auto

mathisscott (approved):

packages/core/src/file/file.element.ts
coryrylan: Tested this and verified, CSS truncated text is still readable to SR. However I think we can still improve with a followup PR to split and show the file extension on the end like you pointed out earlier.

bbogdanov (approved):
Don't forget to change the commit message.

PR #6104, opened 2021-06-28
fix(angular:timeline): apply css sizing only to template cds-icon

### LINKED ISSUE #5366: Icons within timeline content are incorrectly sized
When placing an icon within the title or description of a timeline step, the icon inherits the same size as the step's main icon. This is unintuitive and makes styling icons within the content difficult if not impossible. I can reproduce this behavior in a [Stackblitz](https://stackblitz.com/edit/clarity-v4-timeline-icons?file=src/app/app.component.html). All four icons in this sample are supposed to be different sizes, yet they all use the size inherited from the timeline.

Angular 10.2.3
Clarity 4.0.7
Chrome 87

image

mikelizard5:
…ments

closes #5366

Signed-off-by: Miguel Amaya <mamaya@vmware.com>

## PR Checklist

## PR Type

This adds more specificity for the timeline icon css.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When users project icons into the timeline component our css sets the size.

Issue Number: N/A

## What is the new behavior?
Users can project icons into the timeline component and can set any size that is valid.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/projects/clr-angular/src/timeline/_timeline.clarity.scss
+++ b/packages/angular/projects/clr-angular/src/timeline/_timeline.clarity.scss
@@ -17,8 +17,7 @@
     min-width: $clr-timeline-step-min-width;
     margin-left: $clr-timeline-gutter-width;
 
-    cds-icon,
-    clr-icon {
+    cds-icon:not([size]) {
       @include min-equilateral($clr-timeline-icon-size);
     }
 


hippee-lee (approved):
lgtm - @shijir This fix will have to be cherry-picked over to the angular branch, right?

packages/angular/projects/clr-angular/src/timeline/_timeline.clarity.scss
```diff
@@ -17,8 +17,7 @@
     min-width: $clr-timeline-step-min-width;
     margin-left: $clr-timeline-gutter-width;
 
-    cds-icon,
-    clr-icon {
+    cds-icon:not([size]) {
```
Shijir, line 20: I don't think this is the right approach. With this solution, users have to set the size manually on each icon used in the timeline content. The culprit of the problem is, I believe, is that the style defined for the timeline description icon size is leaking into the timeline content. That's why the icons in the content also getting as big as timeline description icons. So the correct approach I think is to improve the target specificity of that timeline description icon size and prevent it from affecting other areas.

Shijir (changes requested):

packages/angular/projects/clr-angular/src/timeline/_timeline.clarity.scss
bbogdanov: The selector could be `> cds-icon, > clr-icon` and we need both!

bbogdanov (changes requested):

packages/angular/projects/clr-angular/src/timeline/_timeline.clarity.scss
hippee-lee: I don't think we need both cds-icon and clr-icon any more. The template was upgraded to use cds-icon, internally we don't use clr-icon anymore. Since our template isn't using clr-icon anymore we shouldn't need to target it. If consumers are using clr-icon then we don't risk affecting it. 

@shijir i think this fix is ok b/c it establishes a 'default' icon size, the same as what we use in the template. If the consumer wants to use cds/clr-icon they can set the size thats right for their content and our style won't affect it. 

alternatively the fix could get more specific to apply to only the icons used in our template as I originally proposed.  The benefit of that is that the styles don't cascade down into the description component.

bbogdanov: @hippee-lee We do need both. https://github.com/vmware/clarity/pull/5579

hippee-lee: As I understand it, this sizing style should only apply to icons in the template and we only use cds-icon elements in the templates. Am I missing something? 

Keep in mind this fix is for latest angular only. One of the core team members will need to handle any fixes that get applied to previous versions.

mikelizard5:
Closing this pull request in favor of https://github.com/vmware/clarity/pull/6139.

PR #6114, opened 2021-07-01
fix(core:controls): enable aria-labelledby

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Form controls support aria-label but fail when using aria-labelledby. This is needed for certain complex UI patterns like selectable grid rows

Issue Number: N/A

## What is the new behavior?
- enable controls to be labeled with aria-labelledby
- enable radio style to associate when not within a parent control-group

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/forms/control/control.element.ts
+++ b/packages/core/src/forms/control/control.element.ts
@@ -22,6 +22,7 @@ import {
   syncProps,
   pxToRem,
   getElementUpdates,
+  hasAriaLabelTypeAttr,
 } from '@cds/core/internal';
 import { ClarityIcons } from '@cds/core/icon/icon.service.js';
 import { exclamationCircleIcon } from '@cds/core/icon/shapes/exclamation-circle.js';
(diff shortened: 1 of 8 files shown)


bbogdanov (approved):

packages/core/src/forms/control/control.element.ts
```diff
@@ -170,6 +170,10 @@ export class CdsControl extends LitElement {
     return [baseStyles, styles];
   }
 
+  private get hasAriaTypeLabel() {
+    return this.inputControl?.hasAttribute('aria-label') || this.inputControl?.hasAttribute('aria-labelledby');
```
mathisscott: Seems like line 174 would be a good candidate for a utility in `a11y.ts`. It would make it easier to unit test too. The private getter could still remain. Out of scope for this PR, IMO, but could you add a todo comment?

mathisscott (approved):

packages/core/src/forms/control/control.element.ts
coryrylan: Fixed, refactored it out to the a11y.ts file

PR #6126, opened 2021-07-06
fix(core:accordion): a11y fixes and improvements

jeeyun:
- We have a wrapper container that groups an accordion header and its content. We call this a 'panel' but this leads to a lot of confusion as the [aria- documentation refers to just the content part as 'panel'](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion). We'll rename the container to avoid this confusion.
- The accordion content has `role='region'` but moving this up to the host element for better Screen Reader support
- Adding `aria-labelledby` and `aria-controls` attributes and ID values auto-assigned (can be overriden if they use their own ID values)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

We are renaming `cds-accordion-panel` to `cds-accordion-section`. Since accordion is still in beta, this is a good time to introduce this breaking change. The migration path would be to rename the tags in the applications appropriately.


bbogdanov:
We should use `a11y` type in the commit message `a11y(core:accordion): fixes and improvements`.

apps/adoption/src/app/migrations/accordion/accordion.1.core.txt
```diff
@@ -4,20 +4,20 @@ import { Component } from '@angular/core';
   selector: 'app-root',
   template: `
     <cds-accordion>
-      <cds-accordion-panel expanded>
+      <cds-accordion-section expanded>
```
bdryanovski, line 7: Thanks for addressing this ;)

bdryanovski (approved):
LGTM

apps/core-angular-cli/src/app/app.component.html
```diff
@@ -18,27 +18,27 @@
...
-      <cds-accordion-panel [expanded]="panel1Expanded" (expandedChange)="expandedChange($event)">
+      <cds-accordion-section [expanded]="panel1Expanded" (expandedChange)="expandedChange($event)">
         <cds-accordion-header>Item 1</cds-accordion-header>
         <cds-accordion-content *ngIf="panel1Expanded">
           <section>Content 1</section>
         </cds-accordion-content>
-      </cds-accordion-panel>
-      <cds-accordion-panel>
+      </cds-accordion-section>
+      <cds-accordion-section>
```
mathisscott, line 27: Is this not a breaking change?

packages/core/src/accordion/accordion-content.element.ts
```diff
@@ -39,9 +39,16 @@ import styles from './accordion-content.element.scss';
...
  */
 export class CdsAccordionContent extends LitElement {
+  @state({ type: String, reflect: true, attribute: 'role' })
+  protected role = 'region';
+
   connectedCallback() {
     super.connectedCallback();
     this.setAttribute('slot', 'accordion-content');
+
+    if (!this.id) {
```
mathisscott, line 49: Does id need to be a property?

packages/core/src/accordion/accordion-header.element.ts
```diff
@@ -44,6 +44,10 @@ export class CdsAccordionHeader extends LitElement {
   connectedCallback() {
     super.connectedCallback();
     this.setAttribute('slot', 'accordion-header');
+
+    if (!this.id) {
```
mathisscott, line 48: Same here. Should id be a lit property? Disregard if it already is. I'm reviewing on my phone and what I can see of the file is limited

mathisscott (changes requested):
This seems like a breaking change to me. As in, it would break anyone who is currently using the accordion and updates. 

As such, I think we should wait until v6 to merge this in.

packages/core/src/accordion/accordion-content.element.ts
hippee-lee: I think id is inherited from LitElement but now I can't find confirmation of that from the docs.

coryrylan: `id` is an inherited property because `LitElement` extends `HTMLElement`. So all our components automatically get `id` without having to have `@property` . The nice part is if you set the `id` property it automatically reflects to the `id` attribute.

apps/core-angular-cli/src/app/app.component.html
coryrylan: I think this is a breaking change. Accordion with a few others were moved out of beta in march. This would be good for 6.0

packages/core/src/accordion/accordion-content.element.ts
mathisscott: I agree that `id` is part of the DOM spec for HTMLElement. My question was more concerned about whether we want this to be Lit property because of the tie-in to the update lifecycle and the React wrapper. React already has problems with us updating the DOM without it knowing. So I tend to only do this (updating the DOM outside of the Lit lifecycle) for stealth attributes that drive CSS selectors or internal logic (like with Clarity Motion). Things React wouldn't need to keep track of in the Virtual DOM.

jeeyun: I also updated the tree-view PR to not be a Lit property per @coryrylan 's suggestion. Either way we pick, we should keep it consistent across components which way we go.

apps/core-angular-cli/src/app/app.component.html
jeeyun: Yes, this is a breaking change. Makes sense to go into 6.0. 👍

packages/core/src/accordion/accordion-content.element.ts
coryrylan: Yeah all good points @mathisscott  So many DOM props/attr are not consistent in staying in sync together.

packages/core/src/accordion/accordion-section.element.ts
```diff
@@ -80,12 +99,10 @@ export class CdsAccordionPanel extends LitElement implements Animatable {
         type="button"
         @click="${() => this.toggle()}"
         ?disabled="${this.disabled}"
-        aria-disabled="${this.disabled}"
-        aria-expanded="${this.expanded}"
```
scroniser, line 84: @gnomeontherun - aria-expanded should be associated with the button element, as it was previously.

hippee-lee:
I am closing this in favor of two PR's that will replace it. 
[This](https://github.com/vmware/clarity/pull/6200) PR contains the accessibility changes that can be merged as soon as they are approved. [This](https://github.com/vmware/clarity/pull/6195) PR contains the renamed element changes that are a breaking change slated for v6.

Thank you @jeeyun for this and all of your contributions!!! ♥️

PR #6145, opened 2021-07-12
fix: realistic group demos

hippee-lee:

## PR Checklist

## PR Type

When reviewing the navigation demos a11y team asked if I could provide more realistic demos for them to test with. In doin ghtat I found a small css bug and am also including the fix for that here. 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Navigation story demos are only for dev purposes. 

Issue Number: N/A

## What is the new behavior?
More realistic storybook demos. 

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: '14.x'
+          node-version: '14'
+          check-latest: true
       - uses: microsoft/playwright-github-action@v1
       - uses: actions/cache@v2
         with:
--- a/packages/core/src/navigation/navigation.element.scss
+++ b/packages/core/src/navigation/navigation.element.scss
@@ -65,3 +65,7 @@
   height: 100%;
   overflow: auto;
 }
+
+::slotted([slot='cds-navigation-substart']) {
+  width: 100%;
+}
--- a/packages/core/src/navigation/navigation.stories.ts
+++ b/packages/core/src/navigation/navigation.stories.ts
@@ -10,7 +10,38 @@ import customElements from '../../dist/core/custom-elements.json';
 import { html } from 'lit';
 import { CdsNavigation } from '@cds/core/navigation';
 import { CdsNavigationGroup } from '@cds/core/navigation';
+
+import { ClarityIcons } from '@cds/core/icon/icon.service.js';
 import { CdsIcon } from '@cds/core/icon/icon.element.js';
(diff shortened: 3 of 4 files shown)


packages/core/src/navigation/navigation.stories.ts
```diff
@@ -10,7 +10,36 @@ import customElements from '../../dist/core/custom-elements.json';
 import { html } from 'lit';
 import { CdsNavigation } from '@cds/core/navigation';
 import { CdsNavigationGroup } from '@cds/core/navigation';
-import { CdsIcon } from '@cds/core/icon/icon.element.js';
+
+import '@cds/core/icon/register.js';
+import { ClarityIcons } from '@cds/core/icon/icon.service.js';
+
+import {
```
Shijir: I think we can include `ClarityIcons` import here as well.

Shijir (approved):

packages/core/src/navigation/navigation.stories.ts
hippee-lee: 👍. Good catch. Removed.

hippee-lee: Build fails with the icons used the correct way. Storybook seems to have a hard to trackdown issue with importing _only some_ of the icons. 

I've separated them out for now.  The build is still failing b/c it can't find chromium so I'll try to run it again tomorrow.

hippee-lee: And now i've come full circle. The isse was not having the filepaths in the import map for the `web-dev-server` config.

.github/workflows/build.yml
```diff
@@ -11,7 +11,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: '14.x'
+          node-version: '14'
+          check-latest: true
```
hippee-lee, line 15: This is slightly less performant,but it seems to fix the playwright not being installed issue. It forces the job to download the latest node version each time. 
The links to other related issues for playwright and node version are in the commit message.

packages/core/src/navigation/navigation.element.scss
```diff
@@ -65,3 +65,7 @@
   height: 100%;
   overflow: auto;
 }
+
+::slotted([slot='cds-navigation-substart']) {
```
hippee-lee, line 69: Small fix for something I must have broke whenI was cleaning up the shadowDOM and css.

mathisscott (approved):

PR #6159, opened 2021-07-19
fix(angular:schematics): ng add for 12.0.0-next.0

bbogdanov:
The issue was introduced when we decided to sync Clarity Angular version naming with Angular.
The schematic was executing logic of `{Angular version} - 6` that is no longer correct for the new change in version naming.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ng add @clr/angular@12.0.0-next.0` installs version of `@clr/angular` that doesn't exist.

## What is the new behavior?

`ng add @clr/angular@12.0.0-next.0` installs the correct version of `@clr/angular`

- [x] Yes
- [ ] No

### DIFF
--- a/projects/schematics/src/add/index.spec.ts
+++ b/projects/schematics/src/add/index.spec.ts
@@ -7,6 +7,7 @@
 import { Tree, HostTree, SchematicsException } from '@angular-devkit/schematics';
 import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
 import { join } from 'path';
+import { getVersion } from '.';
 import { getFileContent } from '../utility/get-file-content';
 import { setupProject } from '../utility/setup-project';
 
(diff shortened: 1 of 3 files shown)


projects/schematics/src/add/index.ts
```diff
@@ -48,13 +48,12 @@ function setProjectSettings(options: NgAddOptions) {
 }
 
 // Checks if a version of Angular is compatible with current or next
-function getVersion(ngVersion: string, clrVersion: string): string {
-  const diff = 6; // Number disparity between Angular and Clarity, this works as long as we stay in sync with versioning
+export function getVersion(ngVersion: string, clrVersion: string): string {
+  const diff = 0; // Number disparity between Angular and Clarity, this works as long as we stay in sync with versioning
```
gnomeontherun: So we need to do more here, if the Angular version is less than 12, we should use the original logic. That way it will still work on Angular 11 or earlier projects. The CLI will always download the 'latest' version of a package to run `ng-add` commands, so we have to retain backwards compatability.

gnomeontherun (changes requested):

projects/schematics/src/add/index.spec.ts
```diff
@@ -59,6 +60,18 @@ describe('ng add @clr/angular', () => {
...
+
+      expect(correct).toEqual('12.0.0-next.0');
+    });
+
+    it('should return Latest Clarity version', async () => {
+      const correct = getVersion('^12.0.0', '13.0.0');
+
+      expect(correct).toEqual('^12.0.0');
+    });
+
```
gnomeontherun, line 76: Add a test that checks for Angular 11, 10 that you get Clarity 5, 4, respectively.

hippee-lee (approved):

projects/schematics/src/add/index.ts
```diff
@@ -48,13 +48,36 @@ function setProjectSettings(options: NgAddOptions) {
...
+
+  if (majorNgVersion >= 12) {
+    return getVersionsAfterTwelve(ngVersion, clrVersion);
+  } else {
+    // Otherwise, just link to installed version for latest or next releases
+    return getVersionsBeforeTwelve(ngVersion, clrVersion);
+  }
+}
+
+function getVersionsAfterTwelve(ngVersion: string, clrVersion: string): string {
```
gnomeontherun, line 62: This really just needs to return the Angular major version number and return it, doesn't need to calculate a diff.

gnomeontherun (approved):

Issue #6129, opened 2021-07-07
cds-button loading doesn't reverse side effects


loucyx:
## Describe the bug

Hi folks! I don't know if this was reported already, but the behavior of `cds-button` with the `loading-state` attribute is kinda bugged. If you set that property to the value `"loading"` and then you either set it to `"default"` or you just remove it, the button keeps being disabled.

## How to reproduce

You can reproduce the bug and play with it here: https://stackblitz.com/edit/clarity-button-bug?file=index.html 

1. Set the `loading-state` attribute of a `cds-button` to `"loading"`  (you can use `document.querySelector("cds-button").setAttribute("loading-state", "loading")`).
2. Change that attribute to `"default"` or remove it (you can use `document.querySelector("cds-button").setAttribute("loading-state", "default")` to set it or `document.querySelector("cds-button").removeAttribute("loading-state")`).

This is what happens:

| loading-state | Output                                   | Expected?          |
|---------------|---------------------------------------------------|--------------------|
| "loading"     | Button disabled with spinner instead of content   | :heavy_check_mark: |
| "success"     | Button disabled with checkmark instead of content | :heavy_check_mark: |
| "error"       | Button disabled with "x" instead of content       | :heavy_check_mark: |
| "default"     | Button keeps being disabled, showing content      | :x:                |
| unset/removed | Button keeps being disabled, content is hidden    | :x:                |

## Expected behavior

Loading has the side effect of adding the `disabled` property, so it should go back to the original state when stops loading (when `loading-state` has the values `"loading"`, `"success"` or `"error"` it should show the button as disabled, otherwise it should take the `disabled` state from whatever the developer set):

| loading-state | Output                                            |
|---------------|------------------------------------------------------------|
| "loading"     | Button disabled with spinner instead of content            |
| "success"     | Button disabled with checkmark instead of content          |
| "error"       | Button disabled with "x" instead of content                |
| "default"     | Button with disabled state set by dev, and showing content |
| unset/removed | Button with disabled state set by dev, and showing content |

## Versions

**Clarity project:**

- [x] Clarity Core

**Clarity version:**

- [x] v5.x

**Framework:**

- [x] Angular
- [x] React
- [x] Vue
- [x] Other: Vanilla

coryrylan:
Thanks for filing this. I just saw this issue as well on https://cds.wtf/

coryrylan:
closed by fix https://github.com/vmware/clarity/pull/6183

PR #6183, opened 2021-07-26
fix(core:button): disabled/loading state

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If a button was disabled prior to the loading state change it would attempt to revert to inital disabled state once loading was back to default. However the logic was not quite right and would break the loading state in subsequent updates.

Issue Number: https://github.com/vmware/clarity/issues/6129

## What is the new behavior?
Now the loading state will disable the button when not default and when switch back to default set the disabled property the original value properly.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/base-button.element.scss
+++ b/packages/core/src/button/base-button.element.scss
@@ -33,7 +33,6 @@
   cursor: pointer;
   font-size: var(--font-size);
   height: 100%;
-  justify-content: center;
   line-height: 1em;
   min-width: var(--min-width);
   overflow: visible;
@@ -52,21 +51,10 @@
     @include center-content('block');
     height: 100%;
   }
+}
 
-  .button-status-icon {
-    cds-icon {
-      @include min-equilateral(#{$cds-global-space-8});
-      fill: var(--color);
-    }
-
-    cds-icon[shape='error-standard'] {
-      --color: #{$cds-alias-status-danger};
-    }
-
-    .button-spinner {
-      --ring-color: #{$cds-global-color-construction-400};
-    }
-  }
+cds-progress-circle {
+  --ring-color: #{$cds-global-color-construction-400};
 }
 
 :host(:active) .private-host,
--- a/packages/core/src/button/button.element.spec.ts
+++ b/packages/core/src/button/button.element.spec.ts
@@ -5,7 +5,7 @@
  */
 
 import { html } from 'lit';
-import { CdsButton, ClrLoadingState, iconSpinner } from '@cds/core/button';
+import { CdsButton, ClrLoadingState } from '@cds/core/button';
 import '@cds/core/badge/register.js';
 import '@cds/core/button/register.js';
 import { componentIsStable, createTestElement, getComponentSlotContent, removeTestElement } from '@cds/core/test';
@@ -221,15 +221,6 @@ describe('button element', () => {
   });
 
   describe('LoadingStateChange', () => {
-    it('should fallback to default state as expected', async () => {
-      await componentIsStable(component);
-      component.loadingState = null;
-      await componentIsStable(component);
-      expect(component.loadingState).toEqual(ClrLoadingState.default);
-      expect(component.hasAttribute('disabled')).toEqual(false);
-      expect(component.style.getPropertyValue('width')).toBe('');
-    });
-
     it('should set default state as expected', async () => {
       await componentIsStable(component);
       component.loadingState = ClrLoadingState.default;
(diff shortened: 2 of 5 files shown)


packages/core/src/button/base-button.element.scss
```diff
@@ -52,21 +51,10 @@
...
-      fill: var(--color);
-    }
-
-    cds-icon[shape='error-standard'] {
-      --color: #{$cds-alias-status-danger};
-    }
-
-    .button-spinner {
-      --ring-color: #{$cds-global-color-construction-400};
-    }
```
coryrylan, line 68: Was able to use the component properties and simplify the CSS

packages/core/src/button/button.element.spec.ts
```diff
@@ -280,51 +271,74 @@ describe('button element', () => {
...
+      expect(component.shadowRoot.querySelector<any>('cds-progress-circle').size).toBe('12');
     });
+  });
+});
 
-    it('should set button to be readonly', async () => {
-      await componentIsStable(anchorButton);
-      expect(anchor.querySelector('cds-button').readonly).toBe(true);
-    });
+describe('Button link', () => {
```
coryrylan, line 303: Moved this describe since it was nested in the describe above which made it a bit confusing what component it was referencing.

packages/core/src/button/button.element.ts
```diff
@@ -100,75 +81,55 @@ export class CdsButton extends CdsBaseButton {
...
+      ${this.loadingState === ClrLoadingState.success
+        ? html`<cds-icon shape="check" status="success" size="18"></cds-icon>`
+        : ''}
+      ${this.loadingState === ClrLoadingState.error
+        ? html`<cds-icon shape="error-standard" status="danger" size="18"></cds-icon>`
+        : ''}
+      ${this.loadingState === ClrLoadingState.loading
+        ? html`<cds-progress-circle .size=${this.size === 'sm' ? '12' : '18'} status="info"></cds-progress-circle>`
+        : ''}
+      ${this.loadingState === ClrLoadingState.default ? html`<slot></slot>` : ''}
```
coryrylan, line 119: Layout was simplified and didn't need the wrapper elements anymore

packages/core/src/button/icon-button.element.ts
```diff
@@ -44,15 +43,5 @@ export class CdsIconButton extends CdsButton {
...
-  render() {
-    return html`
-      <div class="private-host">
-        ${this.loadingState === ClrLoadingState.loading ? iconSpinner(this.size) : ''}
-        ${this.loadingState === ClrLoadingState.success ? iconCheck : ''}
-        ${this.loadingState === ClrLoadingState.default ? html`<slot></slot>` : ''}
-      </div>
-    `;
-  }
-
```
coryrylan, line 56: Icon button was the same template/logic duplicated as the inherited template

packages/core/src/button/register.ts
```diff
@@ -10,6 +10,11 @@ import { registerElementSafely } from '@cds/core/internal';
 import { CdsButton } from './button.element.js';
 import { CdsIconButton } from './icon-button.element.js';
 import { CdsInlineButton } from './inline-button.element.js';
+import { ClarityIcons } from '@cds/core/icon/icon.service.js';
+import { errorStandardIcon } from '@cds/core/icon/shapes/error-standard.js';
+import { checkIcon } from '@cds/core/icon/shapes/check.js';
+
+ClarityIcons.addIcons(errorStandardIcon, checkIcon);
```
coryrylan, line 17: Moving the side effects to the register to follow the rest of the component register files

packages/core/src/button/button.element.spec.ts
```diff
@@ -346,23 +360,6 @@ describe('buttonSlots: ', () => {
...
-
-  it('should set spinner size to "18" if not passed size "sm"', () => {
-    const templateResult = iconSpinner('anything-at-all');
-    expect(templateResult.values.indexOf('18') > -1).toBe(true);
-  });
-
-  it('should set spinner size to "12" if size "sm"', () => {
-    const templateResult = iconSpinner('sm');
-    expect(templateResult.values.indexOf('12') > -1).toBe(true);
-  });
```
coryrylan, line 363: Moved to the rest of the loading state specs

packages/core/src/button/button.element.ts
```diff
@@ -100,75 +81,55 @@ export class CdsButton extends CdsBaseButton {
...
-
-    // Find and wrap any text nodes into span elements
-    spanWrapper(this.childNodes);
-
-    if (this.loadingState !== ClrLoadingState.default) {
-      this.updateLoadingState();
+  set loadingState(value: keyof typeof ClrLoadingState) {
+    // track prior disabled state to set prior value after button is re-enabled from a loading state
+    if (this._loadingState === ClrLoadingState.default) {
+      this._disabled = this.disabled;
```
coryrylan, line 91: The `_disabled` property was the key fix. To preserve the disabled state while not breaking or causing side effects we need to grab what the current disabled state is when updating from `default` to something else. This way when we go back to `default` we can use what the prior `disabled` state was.

mathisscott (approved):

bbogdanov (approved):

bdryanovski (approved):

PR #6193, opened 2021-07-28
fix(angular:combobox): reduce change detection cycles

hippee-lee:
This is an adoption of: https://github.com/vmware/clarity/pull/6058

This PR reduces change detection cycles for the combobox by replacing
`HostListener` with `Renderer.listen` outside of the zone. We don't need to
run change detections when any keyboard button is pressed or when the multiselect
mode is off. Note that the `HostListener` wraps the actual listener under the hood
into the internal Angular function which runs `markDirty()` before running the actual
listener (the decorated class method).

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Unoptimized change detectino cycles in combobox. 

Issue Number: #6058 (og pr)

## What is the new behavior?
Fewer change detection cycles in combobox. 

- [ ] Yes
- [x] No

### DIFF
--- a/golden/clr-angular.d.ts
+++ b/golden/clr-angular.d.ts
@@ -326,7 +326,7 @@ export declare class ClrCombobox<T> extends WrappedFormControl<ClrComboboxContai
     smartPosition: ClrPopoverPosition;
     textbox: ElementRef;
     trigger: ElementRef;
-    constructor(vcr: ViewContainerRef, injector: Injector, control: NgControl, renderer: Renderer2, el: ElementRef, optionSelectionService: OptionSelectionService<T>, commonStrings: ClrCommonStringsService, toggleService: ClrPopoverToggleService, positionService: ClrPopoverPositionService, controlStateService: IfControlStateService, containerService: ComboboxContainerService, platformId: any, ariaService: AriaService, focusHandler: ComboboxFocusHandler<T>, cdr: ChangeDetectorRef);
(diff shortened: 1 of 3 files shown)


bbogdanov (approved):

Jinnie (commented):
I understand the optimization, but I'm still not sure why is it necessary.
Using host listener for the keyboard input is an Angular standard and leads to simple and easily maintainable code. 
Is there any actual issue (performance, or any side effect) that will make us go after these several milliseconds per click?

hippee-lee:
@Jinnie - I'm not sure. I just adopted the original or and moved it to the angular branch. It seemed like on the original or we were ok with this in. 

If you don't think it's worth adding lmk and I'll just close it too, like the datagrid one.

Jinnie (commented):
I'm hesitating. I like it, technically, it does what it aims for.
But I don't like that we're crossing the Angular boundaries without a good real-world reason for this.
We've been doing the same very sparingly in several places where it's about focus management, but our point there is that the Angular renderer does not have focus handlers built-in.
On the other hand, here we also have keyboard interaction, but we're not going through "setTimeout" cycles as we do with the focus, which was the main problem in that case.
I'd like to leave it some more time, so we can have the chance to hear more opinions from the team, or the contributor.

hippee-lee:
It's been about six weeks since I've seen anything posted by the contributor. I'm going to close it for now for a couple of reasons:

1. Your hesitancy with combo box is similar to my thoughts with datagrid
2. The change is optimizing for an issue we haven't seen or heard about yet
3. Both `@clr` combobox and datagrid components are in maintenance phase

Thanks for adding your thoughts. We have the PR and can always re-open it.

PR #6202, opened 2021-08-01
build: refactoring visual-diff tool and covering core stories with it

bdryanovski:
Updating the Visual Diff tool

* Update on the documentation
* Bugfix into the basic runner for the tool
* Tests to cover Core storybook
* Skipping Angular Dev app tests until a better way of testing is found
* Updated examples 
* other small changes

Runtime around 3min 

Todo: 
* Provide a better reporter with steps on how to reproduce locally
* Try running two tests in parallel
* Provide more tests for Core


tools/visual-tester/README.md
```diff
@@ -144,7 +154,7 @@ This issue is not there when using `npm`.
 If you don't want to do this you could use `node`
 
 ```bash
-node ./dist/runner/cli.js --specs="./specs/**/*.vspec.js"
+./node_modules/.bin/tsc && node ./dist/runner/cli.js --specs="./specs/**/*.vspec.js"
```
Jinnie, line 157: nit: Can we use `npx` here, or we just don't want to take chances about the global npm version?

tools/visual-tester/specs/angular.dev.vspec.js
```diff
@@ -14,6 +14,13 @@ const VisualDiffContext = require('../dist/index').VisualDiffContext;
 const CLARITY_ANGULAR_DEV = process.env.CLARITY_ANGULAR_DEV || 'http://localhost:4200';
 const UPDATE_SNAPSHOTS = process.env.UPDATE_SNAPSHOTS || false;
 
+/**
+ * ! Angular Dev Specs are not suitable to run inside the `next` branch cause
+ * the how application is missing, so for the moment this spec is not runnable
```
Jinnie: nit: I don't clearly understand the comment. Which application is missing?

Jinnie (approved):
With few minor questions/details, I'm ok with this.

tools/visual-tester/specs/core/stories.vspec.js
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * vdiff --specs="./specs/core.foundations.vspec.js"
+ *
+ */
+const VisualDiffContext = require('../../dist/index').VisualDiffContext;
+
+const CLARITY_STORYBOOK_URL = process.env.CLARITY_STORYBOOK_URL || 'https://clarity.design/storybook/core';
+const UPDATE_SNAPSHOTS = process.env.UPDATE_SNAPSHOTS || false;
+
+const stories = [
```
bbogdanov, line 12: Can we make some kind of iterator that goes through each story and takes the name of the story instead of listing them manually in here? 

I guess what we could do is add manually the stories we **don't** want the diff to handle cause x,y,z.

bdryanovski: We could but we don't want all of them .. for example, some of them include loaders or just are redundant stories that are already covered.. 

This is the minimum basic list of stories. We could extend them but have to write additional code for some of them - in the vase of loading animation we need to ignore or skip them overall - grouped alerts are a good example of that. Other like some of the accordion stories are duplicates and there is no point of making 5 screenshot of the same accordion

bbogdanov: That's what I meant we could specify which we don't want the diff to handle. I assume it would be much smaller list than the current one, wouldn't it?

bdryanovski: Yes, it could be done - feel free to update it - It will be a good pre-flight check to see how easy it will be to use the tool and collect some information around the documentation and what could be improved.

But keep in mind the only way to extract the list of stories will be to parse code inside the storybook of the core project (or try to parse a page somewhere) - and this is with the idea that the diff-tool will always be in the same branch - that is something I could not guarantee at the moment

Another issue that I see with that is that the tool is designed to go and visit a random page somewhere without any knowledge about version/branches/codebase - with the change that you suggest, this is not true anymore.

The flow of the test is something like that ...
 * Go visit some website that I think is the latest and correct version and create base images for me.
 * Go now to my local instance or Netlify build of some branch and compare it against the base images I created before that.

In the approach above I'm just telling him - Go and snapshot for me the list of these pages  I don't care about the rest. And maybe the ultimate best way to do it is to have a different test for every page with their own set of code and so on. But this is slow, hard to maintain and we try to have something quick and almost buletproof - that we could say between Branch A and Branch B there is no change at all and we did not break anything or we have change into the padding here and there but this is expected.

bbogdanov (approved):

tools/visual-tester/specs/core/stories.vspec.js
bbogdanov: I understand the point with branching and it makes perfect sense. Thank you for sharing your thoughts! I believe the last comment is a good doc of why we wanted to go in this direction.

PR #6208, opened 2021-08-03
Adoption Guide - Range, Select and Textarea

bdryanovski:
Adding Range, Select and Textarea documentation pages for Adoption guide.

### DIFF
--- a/src/app/migrations/select/select.1.angular.txt
+++ b/src/app/migrations/select/select.1.angular.txt
@@ -1,0 +1,19 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <clr-select-container>
+        <label>Select</label>
+        <select clrSelect>
+          <option value="one">One</option>
+          <option value="two">Two</option>
+          <option value="three">Three</option>
+        </select>
+        <clr-control-helper>Helper text</clr-control-helper>
+      </clr-select-container>
+    </form>
+  `
+})
+export class AppComponent {}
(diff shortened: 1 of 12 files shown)


src/app/migrations/select/select.1.angular.txt
```diff
@@ -0,0 +1,19 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <clr-select-container>
+        <label>Select</label>
+        <select clrSelect>
```
bbogdanov, line 9: Do you think it will be a good idea to add `ngModel` usage in the templates? 

Assuming as an Angular developer coming into Clarity Core I would appreciate an example how would a usage of `ngModel` will look like in Clarity Angular and how would be in Clarity Core and how error handling messages would happen.

mathisscott: +1

bdryanovski: I agree, but also think that it will be best to cover it inside the `Form` page,  there we could create an example for all form elements that we provide and integrate them with their state `Help`, `Error` and `Success`. And every other form control to provide just a link to the `Form` page. 

The goal of the PR was to provide some initial content for the pages and try to minimize the "We are working on it" pages.

bbogdanov: @bdryanovski Let's have a conversation about this. My expectations were different from what you are describing. Let's talk about it when you get back.

mathisscott (approved):

bbogdanov (approved):

PR #6210, opened 2021-08-03
a11y(core:navigation): a change set that improves accessibility for navigation

hippee-lee:
1. refactor for keyboard handlers by separating the wide/narrow toggle
   button and items into two tab-able areas
2. use aria-pressed on wide/narrow toggle button
3. update keypress listeners for structural changes
4. Fix enter / space event listeners that activate projected anchors and btns
5. Adds aria-label to root start button
6. Updated styles for (visual) focus ring

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -9,20 +9,32 @@
 :host {
   --color: inherit;
   --line-height: inherit;
+  --focus-outline: Highlight solid var(--focus-width);
```
hippee-lee: Per conversation with accessibility and design - this was the most practical way to go with this change. Using the browser focus-ring means this can only use roving tabindex instead of aria-activedescendent. aria-activedescendent is used b/c there roving tabindex doen't play nice with the keyboard handlers.

packages/core/src/navigation/navigation-start.element.ts
```diff
@@ -120,36 +123,70 @@ export class CdsNavigationStart extends LitElement implements FocusableItem {
 
   private handleStartButtonText() {
     spanWrapper(this.childNodes);
+
     // get the projected text now wrapped in a span and add the sr attribute.
     this.querySelector('span')?.setAttribute(NAVIGATION_TEXT_WRAPPER, '');
   }
 
+  private renderRootStart() {
```
hippee-lee, line 131: For a more accessible navigation, if there is a root start element, it is tab-able and has its own onKey handlers for expand/collapse behavior.

packages/core/src/navigation/navigation-start.element.ts
```diff
@@ -120,36 +123,70 @@ export class CdsNavigationStart extends LitElement implements FocusableItem {
...
+    return html`
+      <slot @slotchange=${this.handleStartButtonText}></slot>
```
hippee-lee, line 133: For accessibility reasons when a navigation start is used ion the root, the text is projected outside of the button so that the button can act like a toggle and the text can remain independent from it.

packages/core/src/navigation/navigation.element.spec.ts
```diff
@@ -48,77 +48,17 @@ describe('cds-navigation', () => {
     expect(count).toBe(1);
   });
 
-  describe('should handle events', () => {
+  // TODO: figure out how to test the eventListeners added, dispatching events on the shadowDom elements behaves
```
hippee-lee, line 51: It is really hard to mimic events generated in the browser by user actions. Previously I was able to dispatch events on the component and they propagated down as expected. That doesn't happen and dispatching `tab` or `focus` events doesn't trigger the added eventListeners for the elements in the shadowDom. 

I spent 5-6 hours on this, again. (getting testing to work for the event listeners with the MVP release was tricky too). Coverage is still good with the changes but I've been unable to find a way to test the dual zone onKey changes for the root start and items container.

packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -9,20 +9,32 @@
...
   --font-weight: inherit;
   --letter-spacing: inherit;
   --padding: inherit;
+  padding-left: var(--focus-width);
 }
 
 :host(:hover:not([disabled])),
 :host([_has-focus]) {
   background: #{$cds-alias-object-interaction-background-hover};
+  border: 1px solid var(--cds-alias-object-interaction-background-highlight);
```
mathisscott: Instead of `1px` can we use `--cds-alias-object-border-width-100`?

packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -9,20 +9,32 @@
...
 :host([active]) {
   background: #{$cds-alias-object-interaction-background-active};
+  padding-left: 0;
+  border-left: var(--focus-width) solid var(--cds-alias-object-interaction-background-highlight);
+}
+
+:host(:active:focus),
+:host([_has-focus][active]) {
+  background: #{$cds-alias-object-interaction-background-hover};
+  border: 1px solid var(--cds-alias-object-interaction-background-highlight);
```
mathisscott: Same here with `1px`

packages/core/src/navigation/navigation-item.element.ts
```diff
@@ -109,7 +112,6 @@ export class CdsNavigationItem extends LitElement implements FocusableItem {
 
   protected updated(props: PropertyValues) {
     super.updated(props);
-    this.disabled ? this.setAttribute('tabindex', '-1') : this.removeAttribute('tabindex');
```
mathisscott, line 112: Are we keeping `tabindex` on disabled items?

packages/core/src/navigation/navigation-start.element.scss
```diff
@@ -10,14 +10,17 @@
...
   --font-size: inherit;
   --font-weight: inherit;
   --text-transform: capitalize;
+  padding-left: var(--focus-width);
 }
 
 :host(:hover),
 :host([_has-focus]) {
   background: #{$cds-alias-object-interaction-background-hover};
+  border: 1px solid var(--cds-alias-object-interaction-background-highlight);
```
mathisscott: `1px` here too

packages/core/src/navigation/navigation-start.element.scss
```diff
@@ -45,4 +48,17 @@
...
 :host([is-group-start][_active]:not([_expanded])) {
   background: #{$cds-alias-object-interaction-background-active};
+  padding-left: 0;
+  border-left: var(--focus-width) solid var(--cds-alias-object-interaction-background-highlight);
+}
+
+:host([is-group-start][_active]:not([_expanded]):hover),
+:host([_has-focus][is-group-start][_active]:not([_expanded])) {
+  background: #{$cds-alias-object-interaction-background-hover} !important;
+  border: 1px solid var(--cds-alias-object-interaction-background-highlight);
```
mathisscott: Another `1px`

packages/core/src/navigation/navigation-start.element.ts
```diff
@@ -120,36 +123,70 @@ export class CdsNavigationStart extends LitElement implements FocusableItem {
 
   private handleStartButtonText() {
     spanWrapper(this.childNodes);
+
     // get the projected text now wrapped in a span and add the sr attribute.
     this.querySelector('span')?.setAttribute(NAVIGATION_TEXT_WRAPPER, '');
```
mathisscott, line 128: What happens if the button has a span tag inside of it instead of just text? Should we extend `spanWrapper` so we can add a classname or other identifier to it?

packages/core/src/navigation/navigation-start.element.ts
```diff
@@ -120,36 +123,70 @@ export class CdsNavigationStart extends LitElement implements FocusableItem {
...
+        type="button"
+      >
+        <div cds-layout="horizontal align:vertical-center">
+          <span class="icon-slot" cds-layout="${this.expandedRoot ? 'align:right' : 'align:left'}">
+            <slot name="cds-icon-slot">
+              ${this.startIcon
+                ? ''
+                : html` <cds-icon
+                    size="${!this.expandedRoot && this.isGroupStart ? 'xs' : 'sm'}"
+                    shape="${this.isGroupStart ? 'angle' : 'angle-double'}"
```
mathisscott, line 148: Not a suggestion, just a note. To customize the icon here, end users would need to override either `angle` or `angle-double` in `ClarityIcons`...

packages/core/src/navigation/navigation.element.spec.ts
mathisscott: To my understanding, events don't propagate down so much as bubble up. Did you try triggering the event on DOM children?

mathisscott (changes requested):

packages/core/src/navigation/navigation-item.element.ts
hippee-lee: Yes, it prevents them from being tabbable.

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: Yes for all of them. I meant to variable-ize it and then forgot.

packages/core/src/navigation/navigation-start.element.ts
hippee-lee: The text will come from here:
```
<cds-navigation-start>Root or Group Text</cds-navigation-start>
```

Could they put other elements in there? yeah. I had this conversation early on when I implemented the span wrapping for sr text with one or two of us and came to the conclusion that while they shouldn't, for now, if it needs it I should write that up in the documentation. If folks start adding or requestion to add lots of other elements in there, then I can see if it should be extended to handle other things if they make sense. Or, just be firm and explicit that folks should only be projecting text or icons or the override icon for the start button.  
There is a specific slot that allows them to put anything they want after the (optional) root start if they need to. 

wrt  adding a classname or identifier to `spanWrapper`, thats what  the attribute  `NAVIGATION_TEXT_WRAPPER` (cds-navigation-sr-text) is for it acts as a marker for the querySlot so that starts and items can find and update the `display:screen-reader-only` utility token as needed for the projected text. I ended up with that solution after initial PR feedback. I was initially doing just that with a spanWrapper fn that could take an optional class argument and it would get applied to all the spans that get wrapped. Iir the feedback request to separate the classname out was to keep the spanWrapper fn focused on only its wrapping job. Thats when I went in the specific attribute for marking spans that need to be managed for the sr-only token instead of a class.

packages/core/src/navigation/navigation-start.element.ts
hippee-lee: There is an `@querySlot` that checks for a projected icon with the [[cds-navigation-start-icon]](https://github.com/vmware/clarity/pull/6210/files/432432d88001480949acd08993505f6f692b8461#diff-e8fd8c90d14cc0e03558a42b2c491caa15b9028a3ca76685b64a96009faa3b59R102) attribute. If they put that on anything (hopefully an icon) we project that instead of the default icons. They catch is that if they customize it, then they need to manage its state if they want it to change or rotate etc. I worked up an early demo that used eye and eye-hide for the root navigation that controls the wide/narrow state of a navigation. I think it got removed when I re-worked the demos.

packages/core/src/navigation/navigation.element.spec.ts
hippee-lee: I was using `element.dispatchEvent(...)` on the elements that the host attaches the event listeners to for both of the now dual tab-able zones in a cds-navigation component with a root start element. The event listeners used to be on the host and it was worked (but was finicky then too) with `component.dispatchEvent(...)`

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: Updated to `var(--cds-alias-object-border-width-100)`

packages/core/src/navigation/navigation-start.element.scss
hippee-lee: Updated to `var(--cds-alias-object-border-width-100)`

packages/core/src/navigation/navigation-start.element.scss
hippee-lee: Updated to `var(--cds-alias-object-border-width-100)`

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: Updated to var(--cds-alias-object-border-width-100)

mathisscott (approved):

packages/core/src/navigation/navigation-item.element.scss
coryrylan: All components have a normalized focused selector available that applies the browser focus ring without requiring tabindex. You can add the `_has-focus` selector here and then you will get the native focus style without tabindex.

https://github.com/vmware/clarity/blob/next/packages/core/src/internal/base/base.element.scss#L60  

```scss
:host([tabindex='0']:focus),
:host([_focused]) .input,
:host([_has-focus]) {
  outline: Highlight solid $cds-global-space-2;
  outline-offset: $cds-global-space-1;
}

@media (-webkit-min-device-pixel-ratio: 0) {
  :host([tabindex='0']:focus),
  :host([_focused]) .input,
  :host([_has-focus]) {
    outline-color: -webkit-focus-ring-color;
  }
}
```

I recently had a similar issue with a similar fix with the [datagrid](https://github.com/vmware/clarity/blob/topic/datagrid/packages/core/src/internal/base/base.element.scss#L64)

hippee-lee: I removed this line, it is dead code. I looked back into why this css is the way it is. Here is whats driving it:

1. There is an issue with using outline; namely because part of the root layout has overflow on it means that its really hard to get the starts and items outline consistent. Items inside the area with overflow cut off the outline on the x-axis. This results with a start or item having a top/bottom outline but no left/right outline. 
2. The `Highlight` color with light theme and the background color fails accessibility with the `--cds-alias-object-interaction-background-hover` color

After discussions with Colin and Santina I changed to 

- use the border so that nothing gets cut off due to overflow
- use the `--cds-alias-object-interaction-background-highlight`
- keep the `--cds-alias-object-interaction-background-hover`

`--cds-alias-object-interaction-background-highlight` **+**  `--cds-alias-object-interaction-background-hover` gives an accessible color contrast
Screen Shot 2021-08-17 at 3 32 01 PM

vs `Highlight` **+** `--cds-alias-object-interaction-background-hover` does not give a good color contrast
Screen Shot 2021-08-17 at 3 31 02 PM

I'll push the removal change now.

hippee-lee: One other thing that was weird is that I do apply `_has-focus` to the elements via `@state` property with reflection. But, I don't get the outline applied to the elements.

coryrylan: Not blocking the PR if a11y and design has given the thumbs up on this approach but wanted to add some things to note,

- You might be able to fix the outline being cut off using `outline-offset: -2px`. This pulls the outline into the element rather than outside. This is what a handful of our components do. 

- The `Highlight` outline color is not a particular color. It changes from browser to browser and will also change based on the device preferences such as prefers dark theme, high contrast mode, mobile touch and custom color styles. Highlight also will automatically adjust its contrast in many browsers based on the background its placed on. This is not as much of an issue for navigation but it is for smaller components its helpful since we don't know what the background might be. I wrote up a detailed blog post here about the differences/trade-offs https://coryrylan.com/blog/dont-override-css-outline-focus-styles Overall if the native outline can work I'd recommend adding it with these changes as well.

packages/core/src/navigation/navigation-item.element.scss
```diff
@@ -9,20 +9,32 @@
...
   --letter-spacing: inherit;
   --padding: inherit;
+  padding-left: var(--focus-width);
+  border: var(--cds-alias-object-border-width-100) solid transparent;
 }
 
 :host(:hover:not([disabled])),
 :host([_has-focus]) {
   background: #{$cds-alias-object-interaction-background-hover};
+  border-color: var(--cds-alias-object-interaction-background-highlight);
```
coryrylan: All the uses of `var(--cds-*)` should be updated to use the sass var `#{$cds-*}` so we get compile time checks

coryrylan (changes requested):

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: Done

packages/core/src/navigation/navigation-item.element.scss
hippee-lee: I addressed this per our discussion by moving the style for focus back to outline with the pseudo element and an inset to prevent outline clipping.

.github/workflows/build.yml
```diff
@@ -14,10 +14,6 @@ jobs:
           node-version: '14'
           check-latest: true
       - uses: microsoft/playwright-github-action@v1
-      - uses: actions/cache@v2
```
hippee-lee, line 17: I tried several different things to deal with the missing playwright deps

1. setting PLAYWRIGHT_BROWSERS_PATH=0 (both on the command line and as a env variable in the workflow file)
2. adding a  `run: npx playwright install-deps` after yarn install
3. trying to find the folder that the browsers are installed in and caching that (this might actually work but I didn't find the path to the browser executables)

Since there is no `npm ci` for yarn this seems like the best way to make sure we always get the playwright browsers installed for testing.

coryrylan (approved):

PR #6218, opened 2021-08-05
feat(adoption:password): add password examples

bbogdanov:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/src/app/migrations/password/password.1.angular.txt
+++ b/src/app/migrations/password/password.1.angular.txt
@@ -1,0 +1,15 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <clr-password-container>
+        <label>Password</label>
+        <input clrPassword placeholder="Type a password" name="password" />
+      </clr-password-container>
+    </form>
+  `,
+})
+export class AppComponent {
+}
--- a/src/app/migrations/password/password.1.core.txt
+++ b/src/app/migrations/password/password.1.core.txt
@@ -1,0 +1,15 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <cds-password control-width="shrink">
+        <label>Password</label>
+        <input type="password" placeholder="Type a password" name="password" />
+      </cds-password>
+    </form>
+  `,
+})
+export class AppComponent {
+}
--- a/src/app/migrations/password/password.2.angular.txt
+++ b/src/app/migrations/password/password.2.angular.txt
@@ -1,0 +1,21 @@
+import { Component } from '@angular/core';
+import { FormGroup, FormControl } from '@angular/forms';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form [formGroup]="model" clrForm>
+      <clr-password-container>
+        <label>Password with hint</label>
+        <input clrPassword placeholder="Type a password" formControlName="input" required />
+        <clr-control-helper>message text</clr-control-helper>
+        <clr-control-error>error text</clr-control-error>
+      </clr-password-container>
+    </form>
+  `,
+})
+export class AppComponent {
+  model = new FormGroup({
+    input: new FormControl(''),
+  });
+}
--- a/src/app/pages/password.page.ts
+++ b/src/app/pages/password.page.ts
@@ -5,12 +5,57 @@
  */
 
 import { Component } from '@angular/core';
+import { DemoTabData } from '../components/demo.component';
 
 @Component({
   selector: 'app-password',
   template: `
     <h1>Password</h1>
-    <not-ready name="Password"></not-ready>
+    <demo [tabs]="demo1">
+      <h3>Basic</h3>
+    </demo>
+
+    <demo [tabs]="demo2">
+      <h3>Password with Reactive form</h3>
+    </demo>
   `,
 })
-export class PasswordPage {}
+export class PasswordPage {
+  demo1: DemoTabData[] = [
+    {
+      name: 'Angular',
+      files: {
+        'src/app/app.component.ts': 'password/password.1.angular.txt',
+      },
+      language: 'ts',
+      template: 'angular',
+    },
+    {
+      name: 'Core',
+      files: {
+        'src/app/app.component.ts': 'password/password.1.core.txt',
+      },
+      language: 'ts',
+      template: 'core',
+    },
+  ];
+
+  demo2: DemoTabData[] = [
+    {
+      name: 'Angular',
+      files: {
+        'src/app/app.component.ts': 'password/password.2.angular.txt',
+      },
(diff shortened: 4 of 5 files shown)


src/app/pages/password.page.ts
```diff
@@ -5,12 +5,103 @@
...
+    <demo [tabs]="demo1">
+      <h3>Basic</h3>
+    </demo>
+
+    <demo [tabs]="demo2">
+      <h3>password with label</h3>
+    </demo>
+
+    <demo [tabs]="demo3">
+      <h3>password with label</h3>
```
bdryanovski: Duplicate Header `password with label`

Also capitalize the `Password` 

`Password with label`

src/app/migrations/password/password.1.angular.txt
```diff
@@ -0,0 +1,15 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <clr-password-container>
+        <input clrPassword placeholder="input" name="input" [(ngModel)]="input" />
```
bdryanovski: The core example includes `label` here it is missing - is there a reason for that ?

src/app/migrations/password/password.1.core.txt
```diff
@@ -0,0 +1,16 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <cds-password labelLayout="hidden-label" control-width="shrink">
+        <label>input</label>
```
bdryanovski: Why the label say `input` and not `Password`

src/app/migrations/password/password.1.core.txt
```diff
@@ -0,0 +1,16 @@
...
+        <input type="password" placeholder="input" name="input" [(ngModel)]="input" />
```
bdryanovski: `placeholder="Type a password"` - same for all `placeholders`

src/app/migrations/password/password.4.core.txt
```diff
@@ -0,0 +1,25 @@
+import { Component } from '@angular/core';
+
+@Component({
```
bdryanovski: Angular demo both fields are linked together by binding `input` but here there are not - so this is creating a bit strange demo to show ?! 

What do we want to show here?

src/app/migrations/password/password.2.angular.txt
```diff
@@ -0,0 +1,16 @@
+import { Component } from '@angular/core';
+
+@Component({
```
bdryanovski, line 4: What is the difference between `password.1`, `password.2` and `password.3` examples here ?

src/app/pages/password.page.ts
```diff
@@ -5,12 +5,103 @@
...
+    </demo>
+
+    <demo [tabs]="demo4">
+      <h3>All type of paswords</h3>
```
bdryanovski: What do you mean by `All type of paswords` (there is a typo in password) 

Do you mean `Containers for password` ?

bdryanovski (changes requested):
Changes that apply for all examples:

* Renaming `input` to `password` so we could have some visual-q on what are we looking
* If one of the sides includes a label the other one have to include it as well - we try to create mirror versions of the examples

src/app/migrations/password/password.1.angular.txt
bbogdanov: Yes, in the `core` example there is attribute `labelLayout="hidden-label"`

src/app/migrations/password/password.1.core.txt
bbogdanov: I don't see what is bothering you here but if you want to change it I will.

src/app/migrations/password/password.1.core.txt
bbogdanov: I don't see what is bothering you here but if you want to change it I will.

src/app/migrations/password/password.2.angular.txt
bbogdanov: password 1 - has no visible label
password 2 - has a label
password 3 - shows how you can use `[ngModel]` -> I have to remove the `ngModel` from 1 and 2.

src/app/migrations/password/password.4.core.txt
bbogdanov: 1st one - shows text at the bottom
2nd one - shows an error state.

src/app/migrations/password/password.1.angular.txt
```diff
@@ -0,0 +1,14 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  template: `
+    <form clrForm>
+      <clr-password-container>
```
bdryanovski, line 7: I have a few questions with `password.1` example when I try to understand it

* Angular demo doesn't provide label but Core demo have - why?
* Core demo doesn't snow the label anyway ? - Bug or bath implementation 

I know that the reason for that is `labelLayout="hidden-label"` But why to make it like that? 

The basic example must show how to create the most basic form of password field and go to properties that could create confusion 

So if the Angular component doesn't have a label there is no need to implement the label and then hide it on the Core example. Because this could be a bad pattern if someone just copy-paste the code. 

I think the basic example here must include, password with a label without `<form>` wrapper and any additional attributes.

src/app/migrations/password/password.2.angular.txt
```diff
@@ -0,0 +1,15 @@
...
+  selector: 'app-root',
+  template: `
+    <form clrForm class="clr-form-horizontal">
```
bdryanovski: There is no need for the `class` the default layout is this one anyway 

```ts
layout: ClrFormLayout | string = ClrFormLayout.HORIZONTAL;
```

src/app/migrations/password/password.4.core.txt
bdryanovski: Both side of the example don't do the same thing and show different things so its hard for someone to understand why and how things work:

* Angular demo: Both inputs are linked together with a single model object - typing to one update the other one - This is not the same on the core side.
* Core demo error message is always visible no matter what (no validation) but Angular demo handle and show/hide the error message
* Core mention layout but that is not the same one for Angular example. 
* When using `[(ngModel)]` there is almost no need of using `name` in attributes (simplify the examples)
* Angular provides validation rules but Core example does not

bdryanovski (changes requested):

src/app/migrations/password/password.2.angular.txt
bbogdanov: It was strange for me at the beginning as well. I remember I added that cause on stackblitz example for some reason the form sets the container to be vertical. I'll take a another look at it

bdryanovski (approved):

PR #6241, opened 2021-08-09
feat(adoption): add pages for all components clarity has

bbogdanov:

## PR Checklist

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -28,34 +28,61 @@ import { SelectPage } from './pages/select.page';
 import { TextareaPage } from './pages/textarea.page';
 import { TogglePage } from './pages/toggle.page';
 import { CardPage } from './pages/card.page';
+import { WizardPage } from './pages/wizard.page';
+import { TimelinePage } from './pages/timeline.page';
+import { DropdownPage } from './pages/dropdown.page';
+import { ComboboxPage } from './pages/combobox.page';
+import { ProgressbarPage } from './pages/progressbar.page';
+import { TreeviewPage } from './pages/treeview.page';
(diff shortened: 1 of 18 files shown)


bdryanovski (changes requested):
In a way, I'm ok with merging it because it will let us have a lot of stubbed pages to work with later on but at the same time I'm with mixed feelings do we need it right now.

Providing pages for all components that we have into the Angular side is not a bad approach but we could not go only with "not ready". And in a way, it will just create more empty pages that could make the Adoption guide look as not "ready" or too empty.

I really like the change for the building the navigation.

But for now, I think that these pages must be not visible for anyone else except us - or at least we decide on a better way of building this "Not Ready" page with more text and documentation. 

I think that we need to add flag for "draft" pages and not display them.

bbogdanov:
> In a way, I'm ok with merging it because it will let us have a lot of stubbed pages to work with later on but at the same time I'm with mixed feelings do we need it right now.
> 
> Providing pages for all components that we have into the Angular side is not a bad approach but we could not go only with "not ready". And in a way, it will just create more empty pages that could make the Adoption guide look as not "ready" or too empty.
> 
> I really like the change for the building the navigation.
> 
> But for now, I think that these pages must be not visible for anyone else except us - or at least we decide on a better way of building this "Not Ready" page with more text and documentation.
> 
> I think that we need to add flag for "draft" pages and not display them.

This PR is a reflection of a quick discussion I and @gnomeontherun had on the planning last week. We can discuss it tomorrow

bdryanovski (approved):

Issue #6233, opened 2021-08-06
CdsButton with `action="outline"` using `@cds/react` and `@cds/core` is sometimes losing one of its borders


dturcotte:
## Describe the bug

Using @cds/react and @cds/core, I'm seeing an issue in Chrome where CdsButton with action="outline" is sometimes losing one of its borders. Most often this is when there are many of them on the screen and there is a layout change. See repro case below.

![image](https://user-images.githubusercontent.com/1950438/128548913-20559acc-625b-482f-8eaf-35a0c92928eb.png)

## How to reproduce

https://stackblitz.com/edit/react-ts-wmbpcm?file=package.json

1. Load the above stackblitz 
2. Resize the preview pane until there is one more or one fewer column of buttons.
3. Observe missing borders on the buttons like in the screenshot above

## Expected behavior

Expected buttons to continue to have their borders

## Versions

**Clarity project:**

- [X] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [X] v5.x

The above repro is on the latest, 5.5.1 (but I've been seeing it since 5.0)

**Framework:**

- [ ] Angular
- [X] React
- [ ] Vue
- [ ] Other:

**Framework version:**
16.14 (haven't tried 17 yet)

**Device:**

- Type: MacBook
- OS: Big Sur
- Browser: Chrome
- Version: 92.0.4515.131

PR #6244, opened 2021-08-09
fix(core:button): button border disappears on window resize

mathisscott:
I have verified that this is not an issue in vanilla clarity core. This appears to be a browser issue (see PR notes) and it is unclear whether it is exacerbated by react or some other culprit. I tried several different approaches to keep the button borders from disappearing. This includes the unworkable thin keyword as well as other transform settings. This is the only solution that worked consistently.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In very specific circumstances, `@cds/react` button borders can disappear when a window is resized.

Issue Number: #6233 

## What is the new behavior?

The borders no longer disappear.

### Before

Screen Shot 2021-08-09 at 4 37 14 PM

### After

Screen Shot 2021-08-09 at 4 37 46 PM

- [ ] Yes
- [x] No

I've gone through all of the button demos and verified that this change has no impact on them.

## Other information

https://support.google.com/chrome/forum/AAAAP1KN0B0r1neUxqo5Gc/?hl=en&gpf=%23!topic%2Fchrome%2Fr1neUxqo5Gc

### DIFF
--- a/packages/core/src/button/base-button.element.scss
+++ b/packages/core/src/button/base-button.element.scss
@@ -41,7 +41,7 @@
   text-align: center;
   text-decoration: var(--text-decoration);
   text-overflow: ellipsis;
-  transform: translateZ(0px); // used for ::after hover
+  transform: scale(1);
   user-select: none;
   vertical-align: middle;
   white-space: nowrap;


packages/core/src/button/base-button.element.scss
```diff
@@ -41,7 +41,7 @@
   text-align: center;
   text-decoration: var(--text-decoration);
   text-overflow: ellipsis;
-  transform: translateZ(0px); // used for ::after hover
+  transform: scale(1);
```
mathisscott, line 44: I think the `translateZ()` that was previously here was a legacy style that is no longer used. In clarity UI, we used to use translate to push the button down one pixel when it was clicked. We are no longer doing that.

I tried toggling this translate on and off to see if it made any difference. I didn't see any change.

hippee-lee (approved):

coryrylan (approved):

PR #6291, opened 2021-08-25
fix(core:button): not submitting forms on Enter key

### LINKED ISSUE #6263: cds-button prevents form submission when there is more than one input field
## Describe the bug

A form that uses `cds-button` and has more than a single field, can't be submitted with the Enter key.

## How to reproduce

https://stackblitz.com/edit/clarity-keyboard-submit-dlrdzm?file=src/app/app.component.html

* The first example with two input fields, can't be submitted with the Enter key. 
* The second example with just a single text input field can be submitted ok. 
* Adding a regular `<button>` in the third example, enables the form to be submitted via the keyboard again.

## Expected behavior

The form should be submitted by the user pressing the Enter key when in a text field, regardless of the number of input fields.

## Versions

**Clarity project:**

- [X] Clarity Core
- [X] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [X] v5.5.1 / 12.0.1

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
Angular 12.0.2

**Device:**
MacBook
Chrome 92.0.4515.131

## Additional notes

Jinnie:
Close: #6263

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Forms with button with type="submit" having more than 1 input field are not submitted on Enter.
A temporary native button is created dynamically only on Click.

Issue Number: #6263 

## What is the new behavior?

Forms are submitted on Enter if button type="submit" or not defined (default is "submit"). Not submitted on type="button".
If the form has submit type button, it will pre-create a native button with role="presentation".
Changing type dynamically changes button behavior.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/button.element.spec.ts
+++ b/packages/core/src/button/button.element.spec.ts
@@ -8,7 +8,13 @@ import { html } from 'lit';
 import { CdsButton, ClrLoadingState } from '@cds/core/button';
 import '@cds/core/badge/register.js';
 import '@cds/core/button/register.js';
-import { componentIsStable, createTestElement, getComponentSlotContent, removeTestElement } from '@cds/core/test';
+import {
+  componentIsStable,
+  createTestElement,
+  getComponentSlotContent,
+  removeTestElement,
+  emulatedClick,
+} from '@cds/core/test';
 
 describe('button element', () => {
   let testElement: HTMLElement;
@@ -59,7 +65,7 @@ describe('button element', () => {
       expect(component.getAttribute('aria-disabled')).toBe('true');
     });
 
-    it('should work with form elements when clicked', async done => {
+    it('should work with form elements when clicked; defaults to type="submit"', async done => {
       await componentIsStable(component);
       testElement.querySelector('form').addEventListener('submit', e => {
         e.preventDefault();
@@ -67,10 +73,10 @@ describe('button element', () => {
         done();
       });
 
-      component.click();
+      emulatedClick(component);
     });
 
-    it('should work with form elements when clicked via keyboard', async done => {
+    it('should work with form elements when clicked via keyboard; defaults to type="submit"', async done => {
       await componentIsStable(component);
       testElement.querySelector('form').addEventListener('submit', e => {
         e.preventDefault();
@@ -100,20 +106,42 @@ describe('button element', () => {
       expect(o.f).not.toHaveBeenCalled();
     });
 
-    it('should not interact with form elements if type is button', async () => {
-      component.type = 'button';
-      await componentIsStable(component);
+    it('should handle dynamic changes in button type', async done => {
       const o = {
         f: () => {
           // Do nothing
         },
       };
       spyOn(o, 'f');
+
+      // submit case needs this setup to prevent actual form submission and change of locaiton
+      // which lead to errors in the tests
+      const p = (e: Event) => {
+        e.preventDefault();
+        expect(true).toBe(true);
+        done();
+      };
+
+      // change from default (implicit "submit") to type="button"
+      component.type = 'button';
+      await componentIsStable(component);
       testElement.querySelector('form').addEventListener('submit', o.f);
-      component.click();
-      const event = new KeyboardEvent('keyup', { key: 'Enter' });
-      component.focus();
-      component.dispatchEvent(event);
+      emulatedClick(component);
+      expect(o.f).not.toHaveBeenCalled();
+
+      // change from type="button" to type="submit"
+      component.type = 'submit';
+      await componentIsStable(component);
+      testElement.querySelector('form').removeEventListener('submit', o.f);
+      testElement.querySelector('form').addEventListener('submit', p);
+      emulatedClick(component);
+
+      // change from type="submit" to type="button"
+      component.type = 'button';
+      await componentIsStable(component);
+      testElement.querySelector('form').removeEventListener('submit', p);
+      testElement.querySelector('form').addEventListener('submit', o.f);
+      emulatedClick(component);
       expect(o.f).not.toHaveBeenCalled();
     });
 
--- a/packages/core/src/internal/base/button.base.ts
+++ b/packages/core/src/internal/base/button.base.ts
@@ -24,8 +24,7 @@ export class CdsBaseButton extends LitElement {
 
   @property({ type: Boolean }) disabled = false;
 
-  @state({ type: String, attribute: 'aria-disabled', reflect: true }) protected ariaDisabled: 'true' | 'false' | null =
-    'false';
+  @state({ type: String, attribute: 'aria-disabled', reflect: true }) ariaDisabled: 'true' | 'false' | null = 'false';
 
   @state({ type: Number, attribute: 'tabindex', reflect: true }) protected tabIndexAttr: number | null; // don't override native prop as it stops native focus behavior
 
@@ -41,6 +40,15 @@ export class CdsBaseButton extends LitElement {
 
   @querySlot('cds-badge') protected badge: HTMLElement;
 
+  private marker: HTMLElement;
+
+  // When submitting forms with Enter key, the default submit button receives a click event from the form.
+  // We need to be able to differentiate between the real/explicit clicks and those form produced clicks.
+  // Otherwise we will submit the form twice.
+  private explicitClick = false;
+
+  private markerSignificantProperties = ['type', 'name', 'value', 'disabled'];
+
   protected render() {
     return html` <slot></slot> `;
   }
@@ -51,8 +59,16 @@ export class CdsBaseButton extends LitElement {
   }
 
   protected updated(props: Map<string, any>) {
+    if (this.hasMarkerSignificantProperty(props)) {
+      if (this.markerAttached) {
+        this.marker.remove();
+      }
+      this.createMarker();
+    }
+
     super.updated(props);
     this.updateButtonAttributes();
+    this.addDefaultSubmitBehavior();
 
     if (props.has('readonly')) {
       this.setupNativeButtonBehavior();
@@ -76,6 +92,7 @@ export class CdsBaseButton extends LitElement {
 
   private emulateActiveMouseUp() {
     this.active = false;
+    this.explicitClick = true; // mouse down/up are only associated with real clicks
   }
 
   private setupNativeButtonBehavior() {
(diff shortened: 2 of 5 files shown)


packages/core/src/button/button.element.spec.ts
```diff
@@ -100,20 +111,42 @@ describe('button element', () => {
       expect(o.f).not.toHaveBeenCalled();
     });
 
-    it('should not interact with form elements if type is button', async () => {
```
Jinnie, line 103: This test was duplicate of the one before it. That's why it looks overwritten here.

packages/core/src/internal/base/button.base.ts
```diff
@@ -104,26 +119,49 @@ export class CdsBaseButton extends LitElement {
...
+    marker.style.display = 'none';
+    render(buttonTemplate, marker);
+    this.marker = marker;
+  }
+
+  private insideForm(element: HTMLElement): boolean {
+    if (!element.parentElement) {
+      return false;
+    }
+    if (element.parentElement.tagName.toLowerCase() === 'form') {
```
bbogdanov: `if (element.parentElement instanceof HTMLFormElement)`

packages/core/src/internal/base/button.base.ts
```diff
@@ -51,6 +58,13 @@ export class CdsBaseButton extends LitElement {
   }
 
   protected updated(props: Map<string, any>) {
+    if (props.has('type') || props.has('name') || props.has('value')) {
+      if (this.marker && this.marker.parentElement) {
```
bbogdanov: I do not understand what these lines do.

If the `marker` and his parent element exist it will remove the marker and on the next like it sets it again. Why?

bbogdanov (changes requested):

packages/core/src/internal/base/button.base.ts
Jinnie: Recreates the marker if any of the bindable properties get changed.
We may already have a marker, attached to the DOM (in case it is a Submit button). So, we detach the old one (remove only does that) before it gets recreated (on the next line) and reattached further down in "updateButtonAttributes" if needed (the type may have already changed to button, in which case we don't need to attach it, as it will only get instantiated on click and disposed immediately after this).

packages/core/src/button/button.element.spec.ts
```diff
@@ -31,6 +31,17 @@ describe('button element', () => {
     removeTestElement(testElement);
   });
 
+  // We need the full set of mouse events to differentiate between real clicks
```
mathisscott: This would be a good utility function to export out of `test/utils.ts`

packages/core/src/button/button.element.spec.ts
```diff
@@ -100,20 +111,42 @@ describe('button element', () => {
...
+
+      // submit case needs this setup to prevent actual form submission and change of locaiton
+      // which lead to errors in the tests
+      const p = (e: Event) => {
+        e.preventDefault();
+        expect(true).toBe(true);
+        done();
+      };
+
+      // default => 'button'
```
mathisscott: The comment here says the default type is button. But up above it says the default is submit. Which is it? And why?

packages/core/src/internal/base/button.base.ts
mathisscott: this conditional looks like it would be difficult to test effectively. maybe we should break it out so we can unit test the conditional separately from the marker removal?

packages/core/src/internal/base/button.base.ts
```diff
@@ -104,26 +119,49 @@ export class CdsBaseButton extends LitElement {
...
+  }
+
+  private insideForm(element: HTMLElement): boolean {
+    if (!element.parentElement) {
+      return false;
+    }
+    if (element.parentElement.tagName.toLowerCase() === 'form') {
+      return true;
+    }
+    return this.insideForm(element.parentElement);
```
mathisscott: So this recursively walks up the tree until it hits the top of the document or a form tag?

This is another bit of code that looks difficult to test. Maybe this would be better as a utility function in `internal/utils/dom.ts`?

mathisscott (changes requested):

packages/core/src/button/button.element.spec.ts
Jinnie: I'll add better comment. It says we're testing the transition from default (no type attribute) to type="button"
Same with the comments below - testing type="button" changed to type="submit" and vice versa.

packages/core/src/internal/base/button.base.ts
Jinnie: About comment on lines 61-61: They are two separate conditions. First line translates to "we have change in a bindable property", while the second means "marker attached to DOM". I'll extract them in named methods, so the code can be read better.

Comment on lines 161-164: Yes, exactly it tests if we have a "form" parent. I will extract as utility function and do some more extensive testing on it specifically. I may generalize it to test against any parent, based on a parameter.

packages/core/src/internal/base/button.base.ts
Jinnie: Thanks for the suggestion.
As per Scott's comment, we'll extract this as utility function and will probably keep the tagName comparison, so that we can test for other types of parent elements too, not only forms.

bbogdanov: We can still provide a `Type` instead of a `string`. Providing type that is a child of `HtmlElement` is way more stable than providing a `string` that could have anything

packages/core/src/internal/base/button.base.ts
Jinnie: Conditions extracted to more human readable functions.

packages/core/src/button/button.element.spec.ts
Jinnie: Moved to `test/utils.ts`

packages/core/src/button/button.element.spec.ts
Jinnie: done

packages/core/src/internal/base/button.base.ts
Jinnie: Moved to `internal/utils/dom.ts`. Tests added.

packages/core/src/internal/base/button.base.ts
Jinnie: I think we'll have problems identifying some less generic elements, like `section` or `nav`, which seem to directly instantiate the `HTMLElement` and have no specific `HTMLElement` child type associated.
See the `DOM Interface` definition for `section`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section
vs 'div'
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div

packages/core/src/internal/utils/dom.ts
```diff
@@ -166,3 +166,13 @@ export function queryChildFromLightOrShadowDom(hostEl: Element, selector?: strin
...
+
+export function elementHasParent(element: HTMLElement, parentTagName: string): boolean {
+  if (!element || !element.parentElement) {
+    return false;
+  }
+  if (element.parentElement.tagName.toLowerCase() === parentTagName?.toLowerCase()) {
+    return true;
+  }
+  return elementHasParent(element.parentElement, parentTagName);
+}
```
coryrylan: You can use a native method `closest` to find a parent element of a node 
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest  

```js
element.closest('parent-selector')
```

Jinnie: Awesome! Love removing code. 
Done!

packages/core/src/internal/base/button.base.ts
Jinnie: Changed to `element.closest()` per Cory's suggestion.

packages/core/src/internal/base/button.base.ts
Jinnie: Changed to `element.closest()` per Cory's suggestion.

bbogdanov (approved):

packages/core/src/internal/base/button.base.ts
```diff
@@ -104,26 +119,48 @@ export class CdsBaseButton extends LitElement {
...
-          value="${ifDefined(this.value)}"
-          name="${ifDefined(this.name)}"
-          type="${ifDefined(this.type)}"
-        ></button>`;
-
-        const marker = document.createElement('div');
-        this.appendChild(marker);
-        render(buttonTemplate, marker);
+      } else if (!event.defaultPrevented && this.marker && this.explicitClick) {
+        if (this.markerAttached()) {
```
mathisscott: So we are only appending a marker if a marker is already attached? Maybe the language here is confusing me. But it seems to me like we would only append a marker if a marker wasn't attached...

packages/core/src/internal/base/button.base.ts
```diff
@@ -104,26 +119,48 @@ export class CdsBaseButton extends LitElement {
...
+    ></button>`;
+
+    const marker = document.createElement('div');
+    // avoid an empty space inside button
+    marker.style.display = 'none';
+    render(buttonTemplate, marker);
+    this.marker = marker;
+  }
+
+  private markerAttached(): boolean {
```
mathisscott: any reason why this is a method and not a getter?

packages/core/src/internal/base/button.base.ts
```diff
@@ -104,26 +119,48 @@ export class CdsBaseButton extends LitElement {
...
+    return !!(this.marker && this.marker.parentElement);
+  }
+
+  // A marker significant property is one that should be dynamically bound to a property of the native button
+  private hasMarkerSignificantProperty(props: Map<string, any>): boolean {
+    return props.has('type') || props.has('name') || props.has('value') || props.has('disabled');
```
mathisscott: seems like you could have the string values here as an array of strings in a constant (or a private property if preferred) and then loop through them. that way the `or` conditional doesn't get out of hand and you have the strings handy if you need to use them somewhere else in the code.

packages/core/src/internal/base/button.base.ts
```diff
@@ -154,5 +191,14 @@ export class CdsBaseButton extends LitElement {
       this.tabIndexAttr = this.disabled ? -1 : 0;
       this.ariaDisabled = this.disabled ? 'true' : 'false';
     }
+
+    // Forms handle default submit buttons without type="submit" automatically.
+    // Because of the dynamic createion of the native element, we prefer to set the attribute explicitly.
+    if (!this.type && this.closest('form')) {
```
mathisscott, line 206: if we make `type` a getter/setter and store the actual value in a private property (a pattern we've used a lot in the angular components), then could this logic could live in the setter?

packages/core/src/test/utils.ts
```diff
@@ -85,3 +85,13 @@ export function componentIsStable(component: any) {
       })
   );
 }
+
+// Full set of mouse events, generated on click
+export function emulatedClick(component: HTMLElement) {
```
mathisscott, line 90: 🏆

mathisscott (changes requested):

packages/core/src/internal/base/button.base.ts
Jinnie: I guess it won't get called if it's instantiated without type="something" attribute. But I'll test it.

packages/core/src/internal/base/button.base.ts
Jinnie: My understanding is that a property would most often reveal a state, attaching certain behaviors to setters/getters (especially setters), while functions bring some more complex logic along.
In this case, you are right - it could very well be a getter. Will be changed.

packages/core/src/internal/base/button.base.ts
Jinnie: Very good question. Looking at it now, it seems pointless the way it is. But I first need to check why it is passing tests and make test changes if necessary.

Jinnie: You are right here. Logically it has to be `if(!markerAttached())` 
And it works without having a native clickable button just because click can be attached to anything. That's why no test had any objections.
So, apart from the form-submit case, I am not really sure why else we need the native button.
Nevertheless, I'm changing it to reflect the correct behavior.

packages/core/src/internal/base/button.base.ts
Jinnie: If we don't have a type attribute set on the button, it will not get called, and we'll miss the "default submit in forms" case. So, we still need it to be in the update cycle, which is both called on initialization and on change.
But it does not really belong to the `updateButtonAttributes()`. I'll extract it to a separate, more readable, `addDefaultSubmitBehavior()` private method.

mathisscott: The code in base button here is not a great example of how we try to create utility functions that are easy to test without complicated mocks (which complex private functions will require to test branches in the logic). It's not on you to clean all this up but if you would like to be nudged in the right direction. I'm happy to hop on a call.

Another private method is not something I would block on a PR. But we do need to clean this file up to make it easier to unit test.

packages/core/src/internal/base/button.base.ts
Jinnie: done

packages/core/src/internal/base/button.base.ts
Jinnie: done

mathisscott (approved):
Please give @coryrylan a chance to review this PR before merging. Thanks!

coryrylan (approved):

PR #6297, opened 2021-08-26
fix(angular): rework the cds-angular build to use the new custom-elements json format

### LINKED ISSUE #6080: Disabled buttons are broken in core ver 5.4.0
## Describe the bug

Before 5.4.0 it was possible to use the `disabled` property on `<cds-button>`. With 5.4.0 this binding is no longer allowed and the alternative `attr.disabled` doesn't work correctly.

## How to reproduce

**Behaviour in 5.3.0**

Clicking the first button toggles the `disabled` property of the second button. The second button can be clicked when it's not disabled.

https://stackblitz.com/edit/clarit-angular-node-disabled-buttons

**Behaviour in 5.4.0**

Clicking the first button makes the second button disabled but it can never be re-enabled. (This example uses `[attr.disabled]` because the `[disabled]` binding is not valid.)

https://stackblitz.com/edit/clarit-angular-node-disabled-buttons-54?

## Expected behavior

Either the behaviour from 5.3.0 should continue or `[attr.disabled]` should work.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.x

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

bdryanovski:
* Fix some of the jsdoc attributes inside some of the Core
components
* Refactor the Angular Cds Wrapper
* Updating the specs for Angular Cds Wrapper
* Start to use the `custom-elements.json`

### How to test it 

Make sure that everything is updated 

```bash
$ yarn 
$ yarn build:ci
```

After this finish, the new `cds/angular` must be generated as expected. After that, the autogenerated code could be compared against `@cds/angular` package 

```bash
$ mkdir tmp && cd tmp
$ npm init -y && npm install @cds/angular
$ cd node_modules/@cds/angular
```

### How it works

```bash
$ node ./scripts/core-ng-module-generator.js
```

It will go pick the `packages/core/dist/core/custom-elements.json` and use it to generate Angular Wrappers.

This new `custom-elements.json` is different in structure and content, also includes `private` and `protected` fields, additional files like `*.scss.js` files, and much more.

close #6080

### DIFF
--- a/scripts/core-ng-module-generator.js
+++ b/scripts/core-ng-module-generator.js
@@ -10,12 +10,10 @@ const fs = require('fs');
 const path = require('path');
 const Mustache = require('mustache');
 
-const SOURCE_PATH = path.join(__dirname, './../packages/core/dist/core/custom-elements.legacy.json');
+const SOURCE_PATH = path.join(__dirname, './../packages/core/dist/core/custom-elements.json');
 const TARGET_PATH = path.join(__dirname, './../packages/angular/projects/cds-angular/src/cds/components');
 const MUSTACHE_TEMPLTATE_PATH = path.join(__dirname, './../packages/angular/projects/cds-angular/src/cds/_stubs');
 
-// TODO: check if a directory is already created for the same tag
-const IGNORE_DIRECTORY_NAMES = ['index.d.ts', 'internal-components', 'icon-shapes'];
 const MODULE_NAME = 'cds-angular.module.ts'; // hard-coded in the root index mustache template
 
 const directiveTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/directive.ts.mustache', 'utf8');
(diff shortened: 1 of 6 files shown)


scripts/core-ng-module-generator.js
```diff
@@ -24,142 +22,179 @@ const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.musta
...
+
+  // Write root index
+  writeToDisk(
+    path.join(TARGET_PATH, `index.ts`),
+    Mustache.render(rootIndexTemplate, {
+      modules: Object.keys(modulesToCreate)
+        .sort()
+        .map(moduleName => {
+          return {
+            moduleDirectoryName: moduleName,
```
bdryanovski: I could optimize the `modules` object and re-use it both `root-index` and `root-module`

scripts/core-ng-module-generator.js
```diff
@@ -24,142 +22,179 @@ const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.musta
...
-        props: tag.properties && tag.properties.map(p => ({ ...p, isBoolean: p.type === 'boolean' })),
-        events: tag.events,
-        hasProps: !!tag.properties,
-        hasEvents: !!tag.events,
+        module: directory,
+        tagName: directive.tagName,
+        directiveClassName: `${directive.name}Directive`,
+        props: (directive.members && reformatDirectiveProps(directive.members)) || [],
+        events: (directive.members && reformatDirectiveEvents(directive.members)) || [],
+        hasProps: false,
```
bdryanovski: `TypeScript` mental state, made me do this. But in reality, I could rework this like that and not duplicate `hasProps` and `hasEvents`

```js
const templateDate = {
        module: directory,
        tagName: directive.tagName,
        directiveClassName: `${directive.name}Directive`,
        props: (directive.members && reformatDirectiveProps(directive.members)) || [],
        events: (directive.members && reformatDirectiveEvents(directive.members)) || [],
}
templateData.hasEvents = !!templateData.events.length;
templateData.hasProps = !!templateData.props.length;
```

scripts/core-ng-module-generator.js
```diff
@@ -24,142 +22,179 @@ const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.musta
...
+    .map(event => {
+      const { name } = event;
+      return { name };
     });
+}
 
-    createRootIndex(moduleData);
-  }
+function writeToDisk(filename, content) {
+  fs.mkdirSync(path.dirname(filename), { recursive: true });
```
bdryanovski, line 83: This makes sure that directories will be created only if there will be a file inside them.

scripts/core-ng-module-generator.js
```diff
@@ -24,142 +22,179 @@ const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.musta
...
-  if (moduleData) {
-    const templateData = {
-      modules: Object.keys(moduleData).map(md => ({ moduleDirectoryName: md })),
-    };
-
-    createFileOnTemplate(rootIndexTemplate, templateData, 'index.ts', TARGET_PATH);
-  }
+function reformatDirectiveProps(props) {
+  return props
+    .filter(prop => prop.privacy === undefined) // public
```
bdryanovski, line 63: Can I extract this and re-use it!?

scripts/core-ng-module-generator.js
```diff
@@ -24,142 +22,179 @@ const indexTemplate = fs.readFileSync(MUSTACHE_TEMPLTATE_PATH + '/index.ts.musta
...
+    .map(function (word) {
+      return word.charAt(0).toUpperCase() + word.slice(1);
+    })
+    .join('');
 }
 
-function getDirectiveName(tagName) {
-  return tagName + '.directive.ts';
+function extractDirectoryFromPath({ path }) {
+  return path.split('/')[0];
```
bdryanovski, line 35: Can I use `path.dirname` ?

bdryanovski: The answer is No!

```js
path.dirname('forms/control-message/control-message.element.d.ts')
// => forms/control-message
```

This will result into generating Module or files like that
```js
import { CdsForms&#x2F;controlModule } from './forms&#x2F;control/index';
```

bbogdanov (approved):

coryrylan (approved):

PR #6314, opened 2021-09-03
fix(adoption): aligning example headers and spacing between examples

bbogdanov:


bdryanovski (approved):
LGTM

PR #6350, opened 2021-09-20
Cross Website components (Experimental - Beta - Draft)

bdryanovski:
This is EXPERIMENTAL work so everything could change without notice. Keep in mind that all naming and code location is not permanent so don't rely on it. 

### Test use inside the website

https://6148a3962bf54ea5bf92663b--vmware-clarity.netlify.app/components/accordion/

### index.html

Below is an example of requesting components from the outside page. ~It depends on the `component-information.json` file that will provide the data needed for the component~. Update (Fri 24 Sep): the component no longer requires an external file - all the data that it needs is part of the internal implementation.

Both the CDC component and the JSON file are generated on build time and served from the main website - `clarity.design`
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width" />
        <script type="module" src="https://6148a3962bf54ea5bf92663b--vmware-clarity.netlify.app/cdc/index.js"></script>
    <title>
      Clarity Text Cross Component
    </title>

  </head>
  <body>

    <cdc-component-status name="accordion">
      An accordion allows generic content to be collapsed and allows users to expand to show more
      detail.
    </cdc-component-status>

  </body>
</html>
```

### CDC (Clarity Design Components)

Are located here `apps/cross-components` and are build with `https://github.com/coryrylan/clarity-addon-kit` - but modify to handle not only `lib` but Web-ready bundle.

By Web-ready bundle we talk about code that is bundled into a single import that will include LitElement, TypeScript, and all components that will auto-register itself into the global scope. This is needed because we could not import modules or ESM bundles without having some pre-compile stage - something that CDN out there does for us. (maybe there is a better way - but for now, this is an optional improvement)

### Netlify changes

There is a small change into the Netlify configuration to enable CORS so we could request JavaScript to any source file from the outside page.

### Not all files are needed.

The Clarity Addon Kit provides a lot of code that is not needed and could be dropped. 

The package is outside the Yarn workspace - no need to mix package version and we want to be sure that this won't affect other packages.

### Component dependencies

The components don't import Fonts so the website that gonna use them has to include Clarity City font and handle it. The reason is that having a component bundle inside the component could get a bit tricky and will make the bundle a bit too big.

The component is built into two bundles. 
  * Lib - that is used inside the website as part of this build steps so the component will be part of the website without the need of script-tag 
  * Web - this is a full bundle that includes LitElement, Clarity Core, and everything needed to run this component without any other script tags.
 
Inside the website there is a wrapper around the `component-status` this is needed because we want to automate it a little bit and hide some of the data handlers, also if some changes are required, later on, we could fix it in one place.

### Component Information JSON

Update (Fri 24 Sep): This section is no longer valid and it's not needed. There is a private service internally to handle this type of data. 

This file is generated and must contain information about every angular and core component that be used inside the `component-status` component.

This data could be changed and it's not full yet.

In the case of the website, the data is loaded into `window` object and this is just to simplify the access of this data - but again this could be done only inside the website wrapper and done every time when the component is loaded.

```bash
https://clarity.design/component-information.json
```
[Example](https://6148a3962bf54ea5bf92663b--vmware-clarity.netlify.app/component-information.json)

Format of the file
```json
{
  "angular":[
       {
          "name": "component-name",
          "version": "1.0.0",
          "link": "https://angular.clarity.design/documentation/component-name",
          "eslint": "true | false",
          "figma": "link",
        },
  ],
  "core": [
      {
          "name": "component-name",
          "version": "4.0.0",
          "link": "https://clarity.design/core-components/component-name",
          "storybook": "",
          "figma": "link",
        }
  ]
   
}
```

### Angular usage

To use inside an Angular application, a wrapper module must be created to import the web component, and two files must be included inside the app, the JSON file, and the web bundle. These changes must be part of other PR - this branch doesn't include angular code. 

Update (Fri 24 Sep): Angular application will get `CdcModule` to handle all components and merge them as part of the Angular Modules. (This will be part for another PR)

### Known Issues

The Vue component could use the `<slot></slot>` pattern to project the content but because they are used inside Markdown files it's easy to inject additional tags and content that you don't want  - this is the reason for the use of `prop` to pass data.

```html
<CdcComponentStatusWrapper
name="accordion"
description="An accordion allows generic content to be collapsed and allows users to expand to show more detail."
/>
```

This work but it's tricky and will add additional `<p>` tag inside

```html
<CdcComponentStatusWrapper name="accordion">
  An accordion allows generic content to be collapsed and allows users to expand to show more detail.
</CdcComponentStatusWrapper>
```

The correct way of doing it
```html
<CdcComponentStatusWrapper name="accordion">An accordion allows generic content to be collapsed and allows users to expand to show more detail.</CdcComponentStatusWrapper>
```


apps/cross-components/.editorconfig
```diff
@@ -0,0 +1,29 @@
+# EditorConfig helps developers define and maintain consistent
```
bdryanovski: This file could be removed

apps/cross-components/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
```
bdryanovski: This file could be removed

apps/cross-components/README.md
```diff
@@ -0,0 +1,50 @@
+# Clarity Design Addon Kit
```
bdryanovski: This file could be removed

apps/cross-components/package.json
```diff
@@ -0,0 +1,116 @@
...
+    "rollup-plugin-sourcemaps": "0.6.3",
+    "rollup-plugin-terser": "7.0.2",
+    "sass": "1.34.0",
+    "source-map-explorer": "2.5.2",
+    "ts-lit-plugin": "1.2.1",
+    "tslib": "^2.1.0",
+    "typescript": "^4.2.4",
+    "web-component-analyzer": "1.2.0-next.0"
+  },
+  "eslintConfig": {
```
bdryanovski: I think that this is in conflict with the other eslint file !?

apps/cross-components/rollup.config.web.js
```diff
@@ -0,0 +1,115 @@
+/**
```
bdryanovski: some of the code here is not needed when bundle it into single file

apps/cross-components/rollup.performance.js
```diff
@@ -0,0 +1,53 @@
+/**
```
bdryanovski: At the moment this maybe not needed?!

apps/cross-components/sandbox/index.css
```diff
@@ -0,0 +1,3 @@
+body:not([cds-version]) {
```
bdryanovski: Not needed

apps/cross-components/src/global.scss
```diff
@@ -0,0 +1,3 @@
+// empty
```
bdryanovski: Script fail if file is empty, but may be better will be to just remove the how file.

apps/cross-components/src/index.ts
```diff
@@ -0,0 +1,15 @@
+export { test } from 'lib/internal';
```
bdryanovski: This duplicates the registered task but is a lot easier to bundle it for web like that, need to refactor this.

apps/cross-components/web-test-runner.config.mjs
```diff
@@ -0,0 +1,49 @@
+/**
```
bdryanovski: this maybe is not needed

apps/cross-components/rollup.config.lib.js
```diff
@@ -0,0 +1,124 @@
...
+const version = fs.readJsonSync('./npm.json').version;
+
+export default [
+  ...globalStyles({
+    baseDir: config.baseDir,
+    outDir: config.outDir,
+    styles: config.entryPoints.styles,
+  }),
+  {
+    external: [
```
bbogdanov: @bdryanovski  After having this, does the dist bundle still includes `lit` and `@cds/core`?

bdryanovski: No, it does not include them ... there are not needed when you bundle it for `lib` - lib is designed to depend on external builder like webpack, rollup, angular and so on. So dependencies like lit and @cds/core will have to be resolve by the parent. 

In our case for example the website is the parent, so @cds/core and lit element are coming from it.

bbogdanov (approved):

CHANGELOG.md
```diff
@@ -1,3 +0,0 @@
-# Changelog
-
-You can view the full release notes at https://clarity.design/releases.
```
gnomeontherun: Please revert.

apps/cross-components/.storybook/index.css
```diff
@@ -0,0 +1,9 @@
+@import 'normalize.css';
+@import '@cds/core/global.min.css';
+@import '@cds/core/styles/theme.dark.min.css';
+@import '@cds/city/css/bundles/default.min.css';
+@import '../dist/lib/global.min.css';
+
+body {
+  padding: 24px;
```
gnomeontherun: Use a token `var(--cds-global-space-9)`

apps/cross-components/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016-2021 VMware, Inc.  All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
```
gnomeontherun: Don't need another license file, remove.

apps/cross-components/npm.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "clarity-design-components",
+  "version": "0.0.1",
+  "license": "MIT",
+  "private": true,
+  "main": "./index.js",
+  "module": "./index.js",
+  "typings": "./index.d.ts",
+  "type": "module",
+  "files": ["*"],
```
gnomeontherun: Remove, we aren't going to publish to NPM.

apps/cross-components/package.json
```diff
@@ -0,0 +1,112 @@
...
+    "test:unit:coverage": "wtr --coverage",
+    "test:performance": "rollup --config ./rollup.performance.js",
+    "test:performance:report": "npm-run-all 'test:performance --sourcemap'",
+    "build:watch": "npm-run-all 'build:lib -- --watch'",
+    "build:web": "rollup --config ./rollup.config.web.js",
+    "build:lib": "rollup --config ./rollup.config.lib.js",
+    "postinstall": "patch-package"
+  },
+  "dependencies": {
+    "@cds/city": "^1.1.0",
```
gnomeontherun: Are we really depending on these two packages?

apps/cross-components/patches/@web+storybook-prebuilt+0.1.20.patch
```diff
@@ -0,0 +1,8 @@
+diff --git a/node_modules/@web/storybook-prebuilt/web-components.js b/node_modules/@web/storybook-prebuilt/web-components.js
+index 1a28bdd..14e2f7e 100644
+--- a/node_modules/@web/storybook-prebuilt/web-components.js
++++ b/node_modules/@web/storybook-prebuilt/web-components.js
```
gnomeontherun: Is this file supposed to be here?

apps/cross-components/sandbox/index.css
```diff
@@ -0,0 +1,3 @@
...
+  display: none;
```
gnomeontherun: What is this?

apps/cross-components/src/component-status/component-status.element.scss
```diff
@@ -0,0 +1 @@
+@import '~@cds/core/list/list.min';
```
gnomeontherun: Should this be imported here or assumed to be from the app? Otherwise we are bundling this twice in apps.

apps/cross-components/src/services/component-information.service.ts
```diff
@@ -0,0 +1,197 @@
...
+        link: `https://angular.clarity.design/documentation/${componentName}`,
+        eslint: eslintRules.includes(componentName),
+        figma: 'https://www.figma.com/file/ZvaQGGktjGoW6gz9DqwvrLtz/?node-id=1007%3A0',
+      },
+      core: {
+        name: componentName,
+        version: versionMap[`core-${componentName}`] || '4.0.0',
+        link: `https://clarity.design/core-components/${componentName}`,
+        storybook: '', // not sure how to auto-generate it for the moment?!
+        figma: 'https://www.figma.com/file/cjen2ts5Vz0W37NacV1eBE/Clarity-Core-Library-BETA?node-id=1169%3A3267',
```
gnomeontherun: I think we can link to specific Figma locations for each component, but its fine to start with just the overview page.

apps/cross-components/src/internal/base.element.scss
```diff
@@ -0,0 +1 @@
+@import '~@cds/core/global.min';
```
gnomeontherun: Again I don't think we want to load these again.

apps/cross-components/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
```
gnomeontherun: Why export test?

apps/adoption/src/app/migrations/card/card.1.core.txt
```diff
@@ -1,31 +0,0 @@
-import { Component } from '@angular/core';
-
-import '@cds/core/card/register.js';
-
-@Component({
```
gnomeontherun, line 5: Why were these files changed?

apps/website/.vuepress/theme/global-components/CdcComponentStatusWrapper.vue
```diff
@@ -0,0 +1,12 @@
+<template>
+  <cdc-component-status :name="name">
+    <slot>{{ description }}</slot>
+  </cdc-component-status>
+</template>
+
+<script>
+export default {
+  name: 'CdcComponentStatusWrapper',
```
gnomeontherun: Let's remove this wrapper stuff we don't need it.

apps/cross-components/.storybook/main.js
```diff
@@ -0,0 +1,8 @@
+const rollupStorybook = require('../rollup.storybook');
```
gnomeontherun: I think we should move this to `packages/`. Its not an app. I'm not sure `cross-components` is the right name either.

gnomeontherun (changes requested):

apps/cross-components/npm.json
bdryanovski: You're right here, but need as well to modify the rollup config to not make a sanity check of the package.

apps/cross-components/package.json
bdryanovski: IN this case right now yes - but in feature build maybe not

bdryanovski: The storybook is using it

apps/cross-components/src/component-status/component-status.element.scss
bdryanovski: The cross-component is out of the workspace so we just consume it from `node_modules` that is already builded

apps/cross-components/src/index.ts
bdryanovski: It's a legacy left from the `addon-kit` the code that is testing it will be removed

apps/cross-components/src/internal/base.element.scss
bdryanovski: We do .. this is the only way to share the styles inside the component - they could not come from above - this is the issue with having styles locked inside ShadowDOM

apps/cross-components/src/services/component-information.service.ts
bdryanovski: Yep, that is correct, I just left this as "todo" part of the information 

This also includes `version` numbers .. some components have different "first version"

apps/website/.vuepress/theme/global-components/CdcComponentStatusWrapper.vue
bdryanovski: I think that we don't need it as well so will remove it

apps/website/components/accordion/README.md
```diff
@@ -6,6 +6,11 @@ beta: true
 
 <DocPreviewWarning/>
 
+<CdcComponentStatusWrapper
```
bdryanovski: This must be reworked to `<cdc-component-status name="accordion" />` only.

Description and everything else will come from the internal service

apps/cross-components/.storybook/main.js
bdryanovski: I agree the name is not the best one. 

But also have some reservations for the `packages/` folder - and also about the node_modules as well - I prefer to use clarity core as an already released package and not use the `dist` version of it. at the same time some of the packages could be reused from other packages but others could have a conflict (yarn workspaces are tricky)

bdryanovski (commented):
Thanks @gnomeontherun for spending time to review this

apps/cross-components/sandbox/index.css
coryrylan: This was part of the core sandbox demo, it prevents a flash of unstyled content until core has booted up

apps/cross-components/src/internal/base.element.scss
coryrylan: This is correct. This I think is a big pain point for teams wanting to build higher level UI web components easily. Shadow DOM is great for encapsulation but theres no widely supported way to enable inheritance of global styles on the page. We eliminate the duplication or reliance on global styles in Core by pulling in the CSS and then treeshaking it.

Constructible Style Sheets will help solve this once it has proper browser support. Heres a demo I made showing two web components on a page with our global styles. One uses Constructable Style Sheets to share the global CSS and the other does not. 
https://stackblitz.com/edit/http-server-maqpui?file=src%2Fglobal.d.ts 
https://developers.google.com/web/updates/2019/02/constructable-stylesheets

apps/cross-components/src/component-status/component-status.element.scss
coryrylan: I believe the import is correct based on the discussion below on about the base.element.scss file

apps/cross-components/.storybook/preview.js
```diff
@@ -0,0 +1,17 @@
...
+  viewMode: 'docs',
+  docs: {
+    transformSource: (_src, storyContext) => {
+      return storyContext.storyFn().strings;
+    },
+    source: {
+      type: 'dynamic',
+    },
+  },
+};
```
coryrylan: I wonder if it's worth removing storybook from here and use the sandbox to demo the cross website components. Since these are internal to us idk if storybook is worth the complexity/dependencies. No strong opinion here, if the number of components grow then it may be worth leaving.

apps/cross-components/.storybook/main.js
coryrylan: (edit) For the location of this project, a possible idea is a new top level directory. We have `apps` for demo apps and `packages` for publicly consumed npm packages. We could add a `projects` directory for projects that are neither of those two categories. The website even could fit in that category.

So the repo folder structure would look something like this

```
/packages
  /core
  /angular
  /react
/apps
  /demo
  /demo
/projects
  /website
  /website-components (docs, or whatever name we land on)
```

That would give the repo some flexibility to add additional projects like common internal tooling that is fine to be open source but does not need to be part of our public API packages like `@cds` and `@clr`.

apps/cross-components/patches/@web+storybook-prebuilt+0.1.20.patch
coryrylan: This was a workaround that was in the starter kit and core for storybook but I don't know if its needed anymore on the latest storybook. Also see https://github.com/vmware/clarity/pull/6350#discussion_r725234065

apps/cross-components/.storybook/preview.js
bdryanovski: In a way I agree with you at first, I even explore how much effort will be to remove the Storybook - but after working a little bit on another component that I want to add to the list here I found that Storybook helped a lot with documenting components, testing them and providing easy to copy/paste examples. 

Right now the Storybook dependencies that we use into the now know as docs-components is the same that we use for the core package so .. our yarn dependency tree will stay the same. (No additional packages) - It will be nice to keep in mind so when updating to a later version we could use the same version. 

I will vote to keep it for the moment even that we have a single component at the moment with the idea that if we don't need it always could drop it. 

When it comes to the sandbox/demo I think that is the one that will go, there is no real reason to have two demo projects and it's hard to create readable and simple demos without writing a lot of code and keep it in sync later on.

PR #6358, opened 2021-09-24
fix(a11y): fix color contrasts and styles for buttons and form elements

Jinnie:
Box-shadow changed to outline.
Color changed to default browser color for outlines.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

We have multiple A11Y issues about color contrast of the outlines of buttons and form elements.

Issue Number: N/A

## What is the new behavior?

Outlines made same as in Clarity Core.

**Button groups:**
before 
before-buttongroup
after
after-buttongroup

**Buttons**
before
before-buttons-flat
before-buttons-outline
before-buttons-solid
after
after-buttons-flat
after-buttons-outline
after-buttons-solid

**Checkboxes**
before
before-checkbox-selected
before-checkbos
after
after-checkbox-selected
after-checkbox

**Close button**
before
before-close-button
after
after-close-button

**Radios**
before
before-radios-selected
before-radios
after
after-radios-selected
after-radios

**Select multiple**
before
before-multiselect-error
before-multiselect
after
after-multiselect-error
after-multiselect

**Toggles**
before
before-toggles
after
after-toggles

**Textareas**
before
before-textarea-error
before-textarea
after
after-textarea-error
after-textarea

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/utils/_mixins.scss
+++ b/projects/angular/src/utils/_mixins.scss
@@ -24,13 +24,15 @@ $clr-outline-color-inverse: hsla(0, 0%, 100%, 0.09) !default;
 $clr-outline-blur: $clr_baselineRem_2px;
 $clr-outline-spread: $clr_baselineRem_2px;
 
-@mixin include-outline-style-form-fields(
-  $outline-color: $clr-outline-color,
-  $outline-blur: $clr-outline-blur,
-  $outline-spread: $clr-outline-spread
-) {
-  outline: 0;
-  box-shadow: 0 0 $outline-blur $outline-spread $outline-color;
+@mixin include-outline-style-form-fields() {
+  outline: Highlight solid $clr_baselineRem_2px;
+  outline: -webkit-focus-ring-color solid $clr_baselineRem_2px;
+  outline-offset: $clr_baselineRem_1px;
+}
+
+@mixin include-slim-outline-style-form-fields() {
+  outline: Highlight auto $clr_baselineRem_2px;
+  outline: -webkit-focus-ring-color auto $clr_baselineRem_2px;
 }
 
 @mixin include-outline-style-links() {
--- a/projects/angular/src/utils/_reboot.clarity.scss
+++ b/projects/angular/src/utils/_reboot.clarity.scss
@@ -317,8 +317,16 @@
   //
   // Credit: https://github.com/suitcss/base/
   button:focus {
-    outline: 1px dotted;
-    outline: 5px auto -webkit-focus-ring-color;
+    outline: 2px solid -webkit-focus-ring-color;
+    outline-offset: 1px;
+  }
+
+  button:-moz-focusring,
+  [type='button']:-moz-focusring,
+  [type='reset']:-moz-focusring,
+  [type='submit']:-moz-focusring {
+    outline: 2px solid Highlight;
+    outline-offset: 1px;
   }
 
   input,
--- a/projects/angular/src/data/datagrid/_datagrid.clarity.scss
+++ b/projects/angular/src/data/datagrid/_datagrid.clarity.scss
@@ -697,7 +697,7 @@
 
         cds-icon,
         clr-icon {
-          @include css-var(color, clr-color-neutral-500, $clr-color-neutral-500, $clr-use-custom-properties);
+          @include css-var(color, clr-color-neutral-600, $clr-color-neutral-600, $clr-use-custom-properties);
         }
 
         &:hover {
@@ -1190,7 +1190,7 @@
         padding-right: 0;
         cds-icon,
         clr-icon {
-          @include css-var(color, clr-color-neutral-500, $clr-color-neutral-500, $clr-use-custom-properties);
+          @include css-var(color, clr-color-neutral-600, $clr-color-neutral-600, $clr-use-custom-properties);
         }
 
         &:hover {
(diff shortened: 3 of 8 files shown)


bbogdanov (approved):

projects/angular/src/utils/_mixins.scss
```diff
@@ -24,13 +24,13 @@ $clr-outline-color-inverse: hsla(0, 0%, 100%, 0.09) !default;
...
 
-@mixin include-outline-style-form-fields(
-  $outline-color: $clr-outline-color,
-  $outline-blur: $clr-outline-blur,
-  $outline-spread: $clr-outline-spread
-) {
-  outline: 0;
-  box-shadow: 0 0 $outline-blur $outline-spread $outline-color;
+@mixin include-outline-style-form-fields() {
+  outline: -webkit-focus-ring-color solid $clr_baselineRem_2px;
```
hippee-lee, line 29: I assume this will work in Chrome and Edge, was it tested in Firefox and Safari?

projects/angular/src/utils/_reboot.clarity.scss
```diff
@@ -318,7 +318,8 @@
   // Credit: https://github.com/suitcss/base/
   button:focus {
     outline: 1px dotted;
-    outline: 5px auto -webkit-focus-ring-color;
+    outline: 2px solid -webkit-focus-ring-color;
```
hippee-lee, line 320: Same concern here with `-webkit`

hippee-lee (commented):
It looks good to me except for the `-webkit` specific css as commented on below. If it has been tested in All browsers : Chrome, Edge, FF & Safari everything else looks 👌🏼 to me.

projects/angular/src/utils/_mixins.scss
Jinnie: You are right. Though it works fine in Safari, Firefox needs "Highlight" color instead. And some more overrides, as we have some way too specific code in the normalize.css that needs to be overridden.
Changed! Thanks

projects/angular/src/utils/_reboot.clarity.scss
Jinnie: Done!

Jinnie:
Also added in the last commit are a couple of color contrast updates for the datagrid and tabs.

mathisscott (changes requested):
If/when we move this project to the design tokens from `@cds/core` for color, how will we ensure that these changes are carried forward?

Another way to ask, have we validated that all these issues are either fixed in `@cds/core` or do we have a path towards making those fixes?

Jinnie:
> If/when we move this project to the design tokens from `@cds/core` for color, how will we ensure that these changes are carried forward?
> 
> Another way to ask, have we validated that all these issues are either fixed in `@cds/core` or do we have a path towards making those fixes?

Thanks for the question! It's a long story :) 
I discussed these changes with @colinreedmiller and he offered some tokenized colors from the CDS palette. Sadly, it appeared that 1) we don't have similar tokenized colors in Clarity Angular (yet) and even the clr-color-neutral-xyz colors don't have entries matching our requirements. So we had to either chose the nearest already available color from the neutral space, or introduce colors that are alien to the current CLR palette. Here we go for the first approach, though Colin was not entirely happy with it.
About the question if they are ok in Core - the last time I checked the WIP sources - at least some of them were okay.
What I can do is record the colors (tokenized) suggested by Colin, either here and/or in the sources, so we can change them after the `@cds/tokens` are imported. If we have a recent date for this event, I may revert this part of the code and wait after it's merged?

mathisscott:
> Thanks for the question! It's a long story :) I discussed these changes with @colinreedmiller and he offered some tokenized colors from the CDS palette. Sadly, it appeared that 1) we don't have similar tokenized colors in Clarity Angular (yet) and even the clr-color-neutral-xyz colors don't have entries matching our requirements. So we had to either chose the nearest already available color from the neutral space, or introduce colors that are alien to the current CLR palette. Here we go for the first approach, though Colin was not entirely happy with it. About the question if they are ok in Core - the last time I checked the WIP sources - at least some of them were okay. What I can do is record the colors (tokenized) suggested by Colin, either here and/or in the sources, so we can change them after the `@cds/tokens` are imported. If we have a recent date for this event, I may revert this part of the code and wait after it's merged?

### re: timeline for move to design tokens
The move to the `@cds/core` would constitute a breaking change. Even if we offer a smooth/easy path forward for teams that doesn't actually break their environments, we should treat it _as if_ we expected breakages. I don't think we should switch out the internals of the project in a minor version. The potential for regression is too high.

So the earliest we could deliver the design tokens would be v13.

Knowing that, I think this change should go into v12 and potentially v5. I don't think the timeline of the design tokens work impacts the need here.

### re: recording colors for future reference

I think that's a great idea. Maybe add comments to this PR that would be easy to track down when the design tokens work gets underway?

If any of these are not accessible in Core, they should be added or fixed if you wouldn't mind doing that or opening another issue for it.

Jinnie:
> If any of these are not accessible in Core, they should be added or fixed if you wouldn't mind doing that or opening another issue for it.

Tabs are a not-planned-yet stretch goal in Core, while the rest is in datagrid. I just double-checked the latest published datagrid demo and color contrasts there are even higher (using `--cds-global-typography-color-300`).

mathisscott (approved):

Jinnie:
### Color contrast future guidelines:
(after merging the design tokens from Core in Clarity Angular)

**re: Tabs**

We don't have Tabs in Core yet, so i asked Colin for guidance:
> Colin Miller  6:42 PM
> For now - even --cds-global-typography-color-200 meets contrast and we can start there - I may wish to change it later but that only requires changing the -200- to a different number

**re: Datagrid**

For parity with Clarity Core Datagrid `--cds-global-typography-color-300` should be used.

PR #6359, opened 2021-09-27
feat: use cross-component to display component information (experimental)

bdryanovski:
Use cross-components to display the same content to both the website and adoption guide. 

NOTE right now the URL for loading the component is only for demo - later on it will be changed


bdryanovski:
Not needed anymore

PR #6362, opened 2021-09-28
refactor(react): upgrade to @lit-labs/react package for generating react components

hippee-lee:

## Note
The majority of the changes are related to updating snapshots and tests to adjust for the package from `@lit-labs/react`. 

## PR Checklist

## PR Type

This change swaps the custom managed react wrapper for the `@lit-labs/react` package. 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

React components are currently generated with an older fork of the `@lit-labs/react` POC.

Issue Number: N/A

## What is the new behavior?
React components are now generated with the `@lit-labs/react` package. 

- [ ] Yes
- [x] No

## Other information
1. This adds a package dependency in the generated react package.json for `@lit-labs/react`
2. It moves the React version checker and writing to the component files instead of happening in the  `createComponent` function
3. It also has to use the exported class prototype to add in the componet displayName for react dev tools until these get addressed:
```javascript
/**
 * Setting displayName on the prototype is a workaround for now.
 * See: https://github.com/lit/lit/issues/2154
 * Follow: https://github.com/lit/lit/pull/2155
 */
```


packages/react/package.json
```diff
@@ -23,6 +23,7 @@
     "url": "https://github.com/vmware/clarity.git"
   },
   "dependencies": {
+    "@lit-labs/react": "^1.0.0",
```
hippee-lee, line 26: New dependency for react consumers of the package.

packages/react/src/accordion/index.test.tsx
```diff
@@ -14,7 +14,7 @@ describe('CdsAccordion', () => {
         </CdsAccordion>
       </div>
     );
-    const renderedComponent = wrapper.find('CdsAccordion');
+    const renderedComponent = wrapper.find(CdsAccordion);
```
hippee-lee, line 17: Passes the prototype now, not a string.

packages/react/src/accordion/index.tsx
```diff
@@ -3,12 +3,28 @@ import { CdsAccordionContent as AccordionContent } from '@cds/core/accordion';
...
-export const CdsAccordionContent = createComponent('cds-accordion-content', AccordionContent);
+export const CdsAccordionHeader = createComponent(React, 'cds-accordion-header', AccordionHeader);
+export const CdsAccordionContent = createComponent(React, 'cds-accordion-content', AccordionContent);
+
+/**
+ * Setting displayName on the prototype is a workaround for now.
+ * See: https://github.com/lit/lit/issues/2154
+ * Follow: https://github.com/lit/lit/pull/2155
+ */
+CdsAccordion.displayName = 'CdsAccordion';
```
hippee-lee, line 26: Until the wrapper tools are modified to add this, I'm doing it this way to still give React dev tools more information about these components.

packages/react/src/accordion/index.tsx
```diff
@@ -3,12 +3,28 @@ import { CdsAccordionContent as AccordionContent } from '@cds/core/accordion';
...
+
+// Setting react version has to be done at component level since we now consume the @lit-labs/react package
+if ((window as any).CDS && !(window as any).CDS._react.version) {
```
hippee-lee: Checking and adding the react version to the global CDS object for DX. This is done in each component for now.

packages/react/src/navigation/index.tsx
```diff
@@ -11,15 +11,31 @@ import {
...
+ * See: https://github.com/lit/lit/issues/2154
+ * Follow: https://github.com/lit/lit/pull/2155
+ */
+CdsNavigation.displayName = 'CdsNavigation';
+CdsNavigationGroup.displayName = 'CdsNavigationGroup';
+CdsNavigationStart.displayName = 'CdsNavigationStart';
+CdsNavigationItem.displayName = 'CdsNavigationItem';
+
+// Setting react version has to be done at component level since we now consume the @lit-labs/react package
+if ((window as any).CDS && !(window as any).CDS._react.version) {
```
bbogdanov: The `window` object in some cases will not exist. Do you think it is a good idea to assure the `window` object exist before executing next lines?

packages/react/src/select/index.tsx
```diff
@@ -1,5 +1,18 @@
...
+
+/**
+ * Setting displayName on the prototype is a workaround for now.
+ * See: https://github.com/lit/lit/issues/2154
+ * Follow: https://github.com/lit/lit/pull/2155
+ */
+CdsSelect.displayName = 'CdsSelect';
+
+// Setting react version has to be done at component level since we now consume the @lit-labs/react package
+if ((window as any).CDS && !(window as any).CDS._react.version) {
```
bbogdanov: Seems like we reuse the same code across all components. Let's do a util function and reuse it for flexibility

packages/react/src/accordion/index.tsx
mathisscott: Dumb question alert: How does this give React dev tools more info? Or maybe a slightly less dumb question, what info does this help provide?

This is not a challenge to the implementation. I'm filling a knowledge gap on my part.

packages/react/src/accordion/index.tsx
mathisscott: You may want to use optional chaining here.

So `if ((window as any).CDS && !(window as any).CDS._react.version)`

...can become...

`if (window && (window as any)?.CDS?._react.version)`

if `_react` exists, then we can bank on `.version` existing...

packages/react/src/card/index.tsx
```diff
@@ -1,5 +1,18 @@
...
+
+/**
+ * Setting displayName on the prototype is a workaround for now.
+ * See: https://github.com/lit/lit/issues/2154
+ * Follow: https://github.com/lit/lit/pull/2155
+ */
+CdsCard.displayName = 'CdsCard';
+
+// Setting react version has to be done at component level since we now consume the @lit-labs/react package
+if ((window as any).CDS && !(window as any).CDS._react.version) {
```
mathisscott: see previous comment...

packages/react/src/visual-checkbox/index.tsx
```diff
@@ -1,4 +1,18 @@
...
+ * Setting displayName on the prototype is a workaround for now.
+ * See: https://github.com/lit/lit/issues/2154
+ * Follow: https://github.com/lit/lit/pull/2155
+ */
+CdsInternalVisualCheckbox.displayName = 'CdsInternalVisualCheckbox';
+
+// Setting react version has to be done at component level since we now consume the @lit-labs/react package
+if ((window as any).CDS && !(window as any).CDS._react.version) {
+  (window as any).CDS._react.version = React.version;
+}
```
mathisscott: @bbogdanov 's suggestion to make a utility function is a good one. if we have even a minor change to this setup, it will result in a huge number of files needing manual changes.

I wonder if we could do something like this....

```
export const CdsInternalVisualCheckbox = createReactWrapper('cds-internal-visual-checkbox', VisualCheckbox);
```

...and then inside that utility function we have...

```
import { kebabCaseToPascalCase } from '@cds/core/internal';
import { createComponent } from '@lit-labs/react';
import * as React from 'react';

export function createReactWrapper(tagName: string, cdsComonent: any) {
    const reactComponent = createComponent(React, tagName, cdsComponent);
    reactComponent.displayName = kebabCaseToPascalCase(tagName);

    if (window && (window as any)?.CDS?._react && !(window as any).CDS._react.version) {
        (window as any).CDS._react.version = React.version;
    }
} 
```

The end result doesn't dramatically shift what we had with the original converter but I think this or something in this ballpark will make the whole thing more maintainable in the long term. One change can impact 50+ files. As opposed to changing 50+ files by hand.

```

**Review by mathisscott — CHANGES_REQUESTED** [review]
If we are looking to get this into V5, I would like to get it merged before I submit the dropdowns PR.

File: packages/react/src/accordion/index.tsx (hunk 3 above)
Comment (hippee-lee) [reply packages/react/src/accordion/index.tsx:26]: We used to handle this in our fork of the createComponent fn. Basically, with this line, they see <CdsButton> in the react dev tools instead of <ForwardRef> (I think, i forget exactly what surfaces without it, it might be undefined). It's not a blocker if we want to remove it, there is already an issue / PR for it in the lit repo.

File: packages/react/src/visual-checkbox/index.tsx (hunk 8 above)
Comment (hippee-lee) [reply packages/react/src/visual-checkbox/index.tsx]: I spent some time trying to make the `createReactWrapper` (in a funny coincidence, I used the _exact_ same name :)) function. There are typing issue with Reacts types that complicate it a lot and I couldn't get it to work. I can add a util for the React.version check.

This change is for the v6 work so I'm ok not putting it in until after we make the v5 branch. It's not a breaking change.

File: packages/react/src/accordion/index.tsx (hunk 3 above)
Comment (coryrylan) [reply packages/react/src/accordion/index.tsx:26]: Without this the react dev tools report the component as the base class name in the tool rather than the name of the component

File: packages/react/src/accordion/index.tsx (hunk 4 above)
Comment (hippee-lee) [reply packages/react/src/accordion/index.tsx]: I modified this slightly and removed the (window as any) stuff. Had to make it the way it is for tests to pass.

File: packages/react/src/navigation/index.tsx (hunk 5 above)
Comment (hippee-lee) [reply packages/react/src/navigation/index.tsx]: The original code had it this way. I modified it from that to current form.

File: packages/react/src/visual-checkbox/index.tsx (hunk 8 above)
Comment (hippee-lee) [reply packages/react/src/visual-checkbox/index.tsx]: Will stay with the prototype approach as discussed. The correct fix here is in the createComponent function but this is a workaround for React dev tools until then.

File: packages/react/src/utils/index.ts [hunk 9]
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export function logReactVersion(react: { version: string }) {
+  if (window && (window as any).CDS && !(window as any).CDS._react.version) {
```
hippee-lee, line 8: I ended up going back to what we had in the original forked code and added a window guard.

packages/react/src/card/index.tsx
hippee-lee: I had trouble getting it to build with a modified form of this. I neded up going back to what we had in the wrapper fork and adding a guard for `window`.

bbogdanov (approved):

mathisscott (approved):

packages/react/package.json
```diff
@@ -33,6 +34,7 @@
     "@cds/core": "./dist/core",
     "@babel/core": "7.10.2",
     "@babel/preset-env": "7.10.2",
+    "@lit-labs/react": "^1.0.0",
```
coryrylan: Do we need to have this listed as a devDependency if its in dependencies?

coryrylan (approved):

packages/react/package.json
hippee-lee: Nope. I'll remove it when I squash for merge.

PR #6396, opened 2021-10-11
fix(core:styles): cleanup focus css prop

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
This is some work I am pulling out of the datagrid branch. This adds the native focus outline as a token value to help make the outline usage consistent within components. This also adds a focus outline state demo for the interaction examples in storybook. 

- [ ] Yes
- [x] No


packages/core/src/button/button.element.ts
```diff
@@ -120,7 +119,7 @@ export class CdsButton extends CdsBaseButton {
     </div>`;
   }
 
-  static styles = [baseStyles, baseButtonStyles, styles];
+  static styles = [baseStyles, styles];
```
coryrylan, line 122: I combined base button styles and the buttons styles int a single style sheet. Icon button inherits button styles from the button element.

packages/core/src/button/icon-button.element.ts
```diff
@@ -43,5 +42,7 @@ export class CdsIconButton extends CdsButton {
   @property({ type: String, required: 'warning' })
   ariaLabel: string;
 
-  static styles = [baseStyles, baseButtonStyles, styles];
+  static get styles() {
+    return [super.styles, styles];
+  }
```
coryrylan, line 47: Rather than generate the base button css twice just inherit from the button

packages/core/src/forms/control-inline/control-inline.element.scss
```diff
@@ -19,3 +23,17 @@ cds-internal-control-label {
...
+
+[focusable] {
+  content: '';
+  position: absolute;
+  top: calc(calc(calc(var(--cds-alias-object-interaction-touch-target) - var(--height)) * -1) / 2);
+  left: calc(calc(calc(var(--cds-alias-object-interaction-touch-target) - var(--width)) * -1) / 2);
+  width: var(--cds-alias-object-interaction-touch-target);
+  height: var(--cds-alias-object-interaction-touch-target);
+  cursor: var(--cursor);
+}
```
coryrylan, line 43: This ensures the inline controls (checkbox, radio, switch) get the appropriate standard focus outline and minimum touch target size

packages/core/src/forms/control-inline/control-inline.element.ts
```diff
@@ -59,7 +59,8 @@ export class CdsInternalControlInline extends CdsControl {
             ? 'order:reverse'
             : ''}"
         >
-          <div class="input" @click=${this.selectInput}></div>
+          <div role="presentation" class="input" @click=${this.selectInput}></div>
+          <div role="presentation" focusable @click=${this.selectInput}></div>
```
coryrylan, line 63: This fixes an a11y bug where the SR can get confused on treating the click handler as an interactive group rather than focusing to the native slotted checkbox or radio input.

packages/core/src/forms/control-inline/control-inline.element.scss
```diff
@@ -1,11 +1,15 @@
...
 // The full license information can be found in LICENSE in the root directory of this project.
 
 @import './../../styles/tokens/generated/index';
 
 :host {
   width: initial;
+  position: relative;
+  --width: #{$cds-global-space-7};
+  --height: #{$cds-global-space-7};
+  --cds-alias-object-interaction-touch-target: #{$cds-global-space-8};
```
coryrylan, line 12: Sets a minimum touch target for inline elements like radio/checkbox. Also useful for datagrid to align target/focus to cell easily.

packages/core/src/input/input.element.scss
```diff
@@ -12,7 +12,10 @@
   --border: 0;
   --border-bottom: #{$cds-alias-object-border-width-100} solid var(--border-color);
   --border-radius: 0;
-  --outline: none;
+  --outline: var(--cds-alias-object-interaction-outline);
+  --outline-offset: var(--cds-alias-object-interaction-outline-offset);
+  --cds-alias-object-interaction-outline: none;
+  --cds-alias-object-interaction-outline-offset: 0;
```
coryrylan, line 18: Input overrides the default outline since it uses the underline focus effect. However I still use the variable so its consistent with theming and can be adjusted like any other component.

packages/core/src/button/base-button.element.scss
```diff
@@ -1,244 +0,0 @@
-// Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
```
mathisscott, line 1: Why are we deleting base button?

mathisscott (changes requested):

packages/core/src/button/base-button.element.scss
coryrylan: See:
https://github.com/vmware/clarity/pull/6396#discussion_r726506590
https://github.com/vmware/clarity/pull/6396#discussion_r726506969

packages/core/src/accordion/accordion-panel.element.ts
```diff
@@ -96,6 +96,7 @@ export class CdsAccordionPanel extends LitElement implements Animatable {
         ?disabled="${this.disabled}"
         aria-disabled="${this.disabled}"
         aria-expanded="${this.expanded}"
+        focusable
```
Shijir, line 99: Would this `focusable` attribute work in React when users need to assign boolean values?

packages/core/src/forms/control-inline/control-inline.element.scss
```diff
@@ -19,3 +23,21 @@ cds-internal-control-label {
 cds-control-action.status {
   padding-top: 0 !important;
 }
+
+.input {
+  outline: 0 !important;
+}
+
+[focusable] {
```
Shijir, line 31: Related question. I remember for boolean attributes, I needed to specify `[focusable]=true/false` in React. I wonder if that applies here too.

packages/core/src/accordion/accordion-panel.element.ts
coryrylan: This in particular is internal in the shadow dom so no consumer would access/use it. It is for making it easy to apply the focus element styles internal to our templates.

packages/core/src/forms/control-inline/control-inline.element.scss
coryrylan: Same as the other example, this is internal to the template. It doesnt use`::slotted` so its not applying as part of the public API. If its a property of the component React can set it like any other property however if its a plain html attribute (`@property` was not used in the web component) then users will run into this problem https://github.com/facebook/react/issues/9230

Shijir (approved):

packages/core/src/forms/control-inline/control-inline.element.scss
Shijir: Makes sense. Thanks for the explanation.

packages/core/src/button/base-button.element.scss
mathisscott: Got it. I was confused. I thought we were deleting `base-button.element.ts`...

mathisscott (approved):

PR #6421, opened 2021-10-19
chore(core): performance tests

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

## What is the new behavior?
This is more work on the datagrid branch I am splitting out into smaller PRs so its easier to review and we can start using new utilities. I have a experimental side project I made a while back [web-test-runner-performance](https://github.com/coryrylan/web-test-runner-performance) that makes it easier to use the [web-test-runner plugin API](https://modern-web.dev/docs/test-runner/overview/) to test render time and bundle sizes in unit tests. It works with any test setup that uses web-test-runner. 

I started using this plugin to help set a baseline for the datagrid rendering during development. The datagrid branch retains the grid specific tests for now. This PR is an initial setup with a few basic baseline examples (icons, badges, alerts). We can add additional tests to components to measure render/bundle sizes. We can also potentially in the future use the test reporter record performance over time. This replaces the need for our [custom rollup script](https://github.com/vmware/clarity/blob/next/packages/core/rollup.performance.js) and the [bundle-size package](https://github.com/siddharthkp/bundlesize) which we use for bundle size checking.

- [ ] Yes
- [x] No


packages/core/src/alert/alert.performance.ts
```diff
@@ -0,0 +1,38 @@
...
+describe('cds-badge performance', () => {
+  const alert = html`<cds-alert>Alert</cds-alert>`;
+  const alertGroup = (count: number) => html`
+    <cds-alert-group>
+      ${Array.from(Array(count).keys()).map(() => html`<cds-alert>Alert</cds-alert>`)}
+    </cds-alert-group>
+  `;
+
+  it(`should bundle and treeshake alert`, async () => {
+    expect((await testBundleSize('@cds/core/alert/register.js')).kb).toBeLessThan(23);
```
hippee-lee, line 19: Q: How do we determine values (kb, ms) when putting a component through performance testing? Is it arbitrary and usage of the first test results constitutes an initial baseline or something else?

hippee-lee (approved):

packages/core/src/alert/alert.performance.ts
coryrylan: Yeah thats correct, the first test is our baseline to help catch regressions. Overall lower the better for both test types. We can also use the initial test baseline as a way to help determine which existing components could be improved on. The render tests will by default render three times and average the results together. There is an option to increase the average count for particular tests if needed.

mathisscott: Not a blocker but maybe we can export an object or map that has all the expected KB sizes in one place? It would allow us to have a default as well as one place to increase/decrease if we need to. So instead of...

`expect((await testBundleSize('@cds/core/alert/register.js')).kb).toBeLessThan(23);`

...we would have something like...

`expect((await testBundleSize('@cds/core/alert/register.js')).kb).toBeLessThan(BundlesInKilobytes.alert);`

mathisscott: We could do the same with renderTime if we want. Just have a master object with both?

yarn.lock
```diff
@@ -3241,7 +3241,7 @@
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
 "@cds/angular@./dist/angular":
-  version "5.5.4"
+  version "5.5.7"
```
mathisscott: I assume we want `yarn.lock` updated?

mathisscott (approved):

yarn.lock
coryrylan: Not sure, I have to rebase so it will probably get updated. I'll double check

packages/core/src/alert/alert.performance.ts
coryrylan: Possibly, I had something kind of like that but it got complicated trying to name things since the components have multiple render/bundle tests. I'll try to take a second shot at this on the datagrid branch since it has a lot of complex tests to try out

PR #6434, opened 2021-10-25
fix(angular): support sub modules

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The cds-angular generator did not support nested folders for component entrypoints and was not properly building the sub modules of the package. This caused us to have workarounds like this https://github.com/vmware/clarity/blob/next/packages/core/src/selection-panels/index.ts#L8 as well as prevent proper tree shaking.

## What is the new behavior?
This add support for nested sub modules in `@cds/angular` as well as enabling full treeshaking at the component level by importing only the required component module.

- [ ] Yes
- [x] No


packages/angular/projects/cds-angular/_stubs/module.ts.mustache
```diff
@@ -1,18 +1,17 @@
...
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
+import '@cds/core/{{{moduleDirectory}}}/register.js';
 
 {{#directives}}
-import { {{directiveClassName}} } from './{{directiveFileName}}.directive';
+import { {{directiveClassName}} } from './{{{directiveFileName}}}.directive';
```
mathisscott, line 12: Are the triple brackets around  `./{{{directiveFileName}}}.` intentional?

mathisscott: Looking into it, I'm guessing that they are...

mathisscott (approved):

packages/angular/projects/cds-angular/_stubs/module.ts.mustache
coryrylan: Yeah, mustache defaults for rendering html so it was encoding `/` to `%2F`, the triple brackets prevent it from doing a html encode

mathisscott: Yeah. I found the html rendering when I dug into it. Didn't think about it encoding `/`

scripts/core-ng-module-generator.js
```diff
@@ -11,52 +11,103 @@ const path = require('path');
...
-    .split('-')
-    .map(function (word) {
-      return word.charAt(0).toUpperCase() + word.slice(1);
-    })
-    .join('');
-}
+const metadata = JSON.parse(fs.readFileSync(SOURCE_PATH));
+const components = metadata.modules.filter(module => module.declarations.find(d => d.customElement));
+
+const modules = metadata.modules
```
bdryanovski: It will be really nice to have some `/* comments */` here on how things work. Even that I love the functional approach to the problem, it's getting pretty quickly into Haskell/Lisp rabbit hole - and it's really hard to understand what is going on quickly. 

And because this is the single point of failure when it comes to the `cds/angular` package it will be good to be more readable or at least be document why things are like that. 

It took me some time to follow the path and I'm not sure that I understand it that well...

scripts/core-ng-module-generator.js
```diff
@@ -11,52 +11,103 @@ const path = require('path');
...
+    const indexFilePath = path.resolve(TARGET_PATH, path.join(path.dirname(module.path), 'index.ts'));
+    const publicAPIFilePath = path.resolve(TARGET_PATH, path.join(path.dirname(module.path), 'public-api.ts'));
+    const packageFilePath = path.resolve(TARGET_PATH, path.join(path.dirname(module.path), 'package.json'));
+    let moduleClassName = path
+      .dirname(module.path)
+      .replace(/\//g, '-')
+      .split('-')
+      .map((item, index) => (index ? item.charAt(0).toUpperCase() + item.slice(1).toLowerCase() : item))
+      .join('');
+    moduleClassName = `Cds${moduleClassName.charAt(0).toUpperCase() + moduleClassName.slice(1)}Module`;
```
bdryanovski: Why we doing `moduleClassName.charAt(0).toUpperCase() + moduleClassName.slice(1)` twice, I think that we already done it on line 57 with the last `map` ?

I guess the `map` above tries to solve issues like `progress-bar` but still why we **don't want to do it** on the `0` indexed item? Is there are case when this could be an issue? 

from the above example we will get:
```js
let moduleClassName = 'progressBar'       // path/dirname/replace/split/map/join
moduleClassName = 'CdsProgressBarModule'  // string interpolation 
```
Right?

My question here is more is there something I'm not seeing for the need of adding `if` and the addition operations?

scripts/core-ng-module-generator.js
```diff
@@ -65,7 +116,31 @@ function reformatDirectiveProps(props) {
     .map(prop => {
       const { name, type } = prop;
       const isBoolean = type && type.text === 'boolean';
-      return { name, isBoolean };
+
+      let formattedType = type.text;
```
bdryanovski: Some help here: I can't figure out what are we trying to get from `formattedType`.


For example below on line 129, if the type is `boolean` we are setting it to be `boolean | ''` is this because the Angular template strict checks? If this is the case why we don't quit, why need to continue making additional checks ? The `===` make sure that there are no additional types.

On line 121, we check for multiple types, and in the case that some of them include `string<name>` as type and it's null we set `any` as type - we don't quit here?

The check on line 133, is working with some already transformed or not data it's a bit hard to read what we expect from it - from what I'm thinking that I understand is that everything except `boolean | '' | number | string | HTMLElement` will be set to `any`, right? But it's hard to understand it like that. 

I think that the transformation of the `boolean` type is making the if on line 133 a bit complex.

```js
if (
 // If type is not found from the list below it must be `any`
 !['boolean', 'number', 'string', 'HTMLElement'].find(t => t === formattedType) 
 // If type is complex one and include string keys as types or is null - it must be `any`
 || formattedType.split('|').find(t => !t.includes("'") || !t.includes('null')
) {
 formattedType = 'any'
}

// Adding `string` type to boolean types to make Angular stringTemplate check pass.
if (formattedType === 'boolean') {
  formattedType = `${formattedType} | string`
}
```

packages/core/src/badge/badge.element.ts
```diff
@@ -41,7 +41,7 @@ export class CdsBadge extends LitElement {
 
   /**
    * Sets the color of the badge
-   * @type {neutral | info | success | warning | danger}
+   * @type {'neutral' | 'info' | 'success' | 'warning' | 'danger'}
```
bdryanovski, line 44: I don't think that this is the only place where we have string keys as `@type` - this is only for documentation reason I know and must not have any effect on `cds/angular` build script. 

But for having consistent documentation generation it will be nice to fix it on all places or open an Issue with "Help Wanted" to provide some easy tasks for contribution.

```
packages/core/src/alert/alert-group.element.ts:
  60     * Sets the overall height and width of the alerts inside the alert group
  61:    * @type {default | sm}
  62     */

  67     * Passed down into the alerts inside the alert-group
  68:    * @type {default | banner | light}
  69     */

  80     * Sets the status of the alerts inside the alert group
  81:    * @type {neutral | info | success | warning | danger | alt | loading}
  82     */
```

.gitignore
```diff
@@ -12,7 +12,40 @@ packages/react/coverage
 .rollup.cache
 
 # auto-generated ng-core components
-packages/angular/projects/cds-angular/src/cds/components
+packages/angular/projects/cds-angular/accordion
```
bdryanovski: I initially try to optimise and minimise the long paths listing. And there is a way for sure to do that - but open a potential hole that someone could forget to update the `gitignore` path and lead to missing files. 

```gitignore
packages/angular/projects/cds-angular/*
!packages/angular/projects/cds-angular/*.json
!packages/angular/projects/cds-angular/*.md
!packages/angular/projects/cds-angular/karma.conf.js
!packages/angular/projects/cds-angular/yarn.lock
!packages/angular/projects/cds-angular/src
!packages/angular/projects/cds-angular/src/cds-angular.module.ts
!packages/angular/projects/cds-angular/src/public-api.ts
```

The above ignore everything inside the `packages/angular/projects/cds-angular` but excludes `src` and a few other files.

Again if you create a file for example `eslint.conf` and not mention it inside the `.gitignore` file this file won't be tracked and won't be visible for `git`. 

⚠️ But that makes me think why we need to ignore the generated files ? We don't change them and every time when we build them we expect to have identical copies unless we change something in `core`. They are not minified or compressed, have no timestamps, no dynamic data.

So why don't we keep them as markers for change, if you get changes into the `cds-angular` folder that means something changed into the `core` package and needs to be reviewed. 

This also will let us check if the types are correctly translated from `core` to `cds/angular` package or even the props are there (this is a common issue that we have lately).

The impact on the `git` cache and repository will be really small over time.

🐞 The only thing that I see as an issue is that if we refer for some reason in this repository the `cds/angular` we could run against old not an updated version of the package - and there will be no error if we don't search for something `new`. But I don't know that we do such a thing in this branch so this is only a "Keep in mind - thing" - in this current state this is still a valid point anyway - I don't see any task to delete the generated content.

bdryanovski (commented):
My comments here are more of documentation and support questions nature - no need to block the PR, but it will be nice to address them in some form or another.

scripts/core-ng-module-generator.js
coryrylan: Yeah good point. I originally was hoping to get the fix with little change as possible but it became more involved than I expected. I'll add some comments and clean up some functions

packages/core/src/badge/badge.element.ts
coryrylan: I changed this mainly for testing that it worked but didn't update all, I'll open a Help Wanted issue.

.gitignore
coryrylan: I had the negated ignore prior and was having trouble with it marking what was already in source control as a rename rather than a delete. Thinking about it now, I probably need to make it two commits so it marks the file as a rename then commit the gitignore. 

Agree it wouldnt hurt to source control the file either or seems like low impact/risk. I'll try the negated gitignore for this PR see if that works and then can do a follow up if it make sense to source control all.

hippee-lee (approved):

scripts/core-ng-module-generator.js
coryrylan: Cleaned up classname logic

scripts/core-ng-module-generator.js
coryrylan: Split out into better function names and cleaned up the logic so it wasnt so nested with loops/maps

scripts/core-ng-module-generator.js
coryrylan: After going through this I realized the Angular compiler will infer the TS types so by simply applying the Web Component class as the directive element type the directives improved substantially without needing any custom logic from the json metadata. Auto-complete works much better in VS code now for Angular templates. Thanks for walking through this!

```typescript
// example output
import { Directive, ElementRef, Input } from '@angular/core';
import { CdsBadge } from '@cds/core/badge';

@Directive({ selector: 'cds-badge' })
export class CdsBadgeDirective {
  protected element: CdsBadge;
 
  get status() {
    return this.element.status;
  }
  @Input() set status(value) {
    this.element.status = value;
  }

  constructor(elementRef: ElementRef) {
    this.element = elementRef.nativeElement;
  }
```

Screen Shot 2021-10-27 at 11 09 13 PM

.gitignore
coryrylan: For now I added the simpler gitignore you posted

bdryanovski (approved):

scripts/core-ng-module-generator.js
bdryanovski: That's even better

.gitignore
bdryanovski: Agree

Issue #6295, opened 2021-08-26
Core icons don't support non-square icons


rbirkgit:
## Describe the bug

When using rectangular icons e.g. 96x124, clarity core icons are rendered in a 124x124 space (which is correct), but the svg inside is rendered as 124x160. Part of the bottom also gets cut off.

## How to reproduce

Assign a 96x124 icon to cds-icon.

If needed I can probably create a stackblitz example.

## Expected behavior

Icons render like they did with Clarity Angular icons, i.e. the svg as 124x124 as well.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.5

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
Angular 12

## Additional notes

The style used in angular that is "missing" in core is this, which causes the rectangular icon to render correctly.
projects/icons/src/clr-icons.scss:92
```
& > * {
  height: 100%;
  width: 100%;
}

```

rbirkgit:
Adding one of our icons in case one is needed to test with:
[...]

PR #6436, opened 2021-10-26
feat(core:icon): support non-equilateral icon sizing

mathisscott:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

This was a lot trickier than I thought it would be. The CSS identified in the github issue was not the culprit. The challenge was the `contain` CSS style.

Issue Number: #6295

## What is the new behavior?

The icon now accepts `fit` as part of its value. Putting this in the size attribute what force the icon to take the size the icon inside of it. Note that the icon will still scale up and down with sizing. It will just allow for the SVG inside of it to push out the host's boundary.

```
<cds-icon shape="my-custom-icon" size="xl fit"></cds-icon>
```

This had to be added as a new feature because there was no way we could extend the current icon with its performance-focused CSS contain style to make it re-size to the internal image. This was the easiest way to maintain our current icon features and support the edge case of a non-equilateral icon added to the registry.

An example is provided under the storybook icon stories.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/icon/icon.element.scss
+++ b/packages/core/src/icon/icon.element.scss
@@ -15,6 +15,7 @@
   --color: currentColor;
   display: inline-block;
   @include equilateral(#{$cds-global-space-7});
+  @include min-equilateral(#{$cds-global-space-7});
   margin: 0;
   vertical-align: middle;
   fill: var(--color);
@@ -28,33 +29,49 @@ svg {
 }
 
 // sizing
-:host([size='xs']) {
+:host([size*='xs']) {
   @include equilateral(#{$cds-global-space-5});
+  @include min-equilateral(#{$cds-global-space-5});
 }
 
-:host([size='sm']) {
+:host([size*='sm']) {
   // weirdly, the default... 16px
   @include equilateral(#{$cds-global-space-7});
+  @include min-equilateral(#{$cds-global-space-7});
 }
 
-:host([size='md']) {
+:host([size*='md']) {
   // 24px
   @include equilateral(#{$cds-global-space-9});
+  @include min-equilateral(#{$cds-global-space-9});
 }
 
-:host([size='lg']) {
+:host([size*='lg']) {
   // 36px
   @include equilateral(#{$cds-global-space-11});
+  @include min-equilateral(#{$cds-global-space-11});
 }
 
-:host([size='xl']) {
+// note: the selectors for `xl` and `xxl` work as expected due to the order in
+// which they are listed here. this is fine and expected for CSS. just know
+// that if their order is switched around, they will break.
+// likewise if we add an `xxs` it will have to go AFTER `xs`!
+:host([size*='xl']) {
   // 48px
   @include equilateral(#{$cds-global-space-12});
+  @include min-equilateral(#{$cds-global-space-12});
 }
 
-:host([size='xxl']) {
+:host([size*='xxl']) {
   // 64px
   @include equilateral(calc(#{$cds-global-space-13} - #{$cds-global-space-5}));
+  @include min-equilateral(calc(#{$cds-global-space-13} - #{$cds-global-space-5}));
+}
+
+:host([size*='fit']) {
+  height: auto;
+  width: auto;
+  contain: layout;
 }
 
 // colors
--- a/packages/core/src/icon/icon.stories.ts
+++ b/packages/core/src/icon/icon.stories.ts
@@ -5,6 +5,7 @@
  */
 
 import '@cds/core/icon/register.js';
+import '@cds/core/button/register.js';
 import { ClarityIcons } from '@cds/core/icon/icon.service.js';
 import { imageIcon } from '@cds/core/icon/shapes/image.js';
 import { userIcon } from '@cds/core/icon/shapes/user.js';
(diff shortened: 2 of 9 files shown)


packages/core/src/icon/icon.stories.ts
```diff
@@ -5,32 +5,15 @@
...
 import '@cds/core/icon/register.js';
+import '@cds/core/button/register.js';
 import { ClarityIcons } from '@cds/core/icon/icon.service.js';
 import { imageIcon } from '@cds/core/icon/shapes/image.js';
 import { userIcon } from '@cds/core/icon/shapes/user.js';
 import { html, LitElement } from 'lit';
 import { customElement } from 'lit/decorators/custom-element.js';
 import { state } from 'lit/decorators/state.js';
 import { homeIcon } from '@cds/core/icon/shapes/home.js';
-// import { boolean, select, text } from '@storybook/addon-knobs';
```
mathisscott, line 15: I removed all of this commented out stuff. It didn't seem like something we wanted to keep around.

packages/core/src/icon/icon.stories.ts
```diff
@@ -656,3 +495,69 @@ export function lazyLoading() {
     }
   }
 }
+
+const testNonequilateralIcon =
```
mathisscott, line 499: Here is the story that demonstrates the icon with different height and width. It's not on the main storybook page because this is an edge case requiring a custom icon being added to the icon registry and that icon having different height/width _and_ that being meaningful. I don't want to confuse people with different sizing options that 80% of them will not need.

packages/core/src/icon/utils/icon.classnames.ts
```diff
@@ -4,8 +4,8 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { isTshirtSize, pxToRem, updateElementStyles } from '@cds/core/internal';
-import isNil from 'ramda/es/isNil.js';
+import { isTshirtSize, pxToRem, replaceWord, updateElementStyles, unsetElementStyles } from '@cds/core/internal';
+import isNil from 'ramda/es/isNil.js'; // TODO: REPLACE WITH INTERNAL FN AFTER MERGING DROPDOWN BRANCH
```
mathisscott, line 8: Added a note here to use the internal `isNil` function to avoid/prevent circular dependencies in the future.

packages/core/src/internal/utils/css.ts
```diff
@@ -40,6 +40,13 @@ export function updateElementStyles(el: HTMLElement, ...styleTuples: [string, st
   return el;
 }
 
+export function unsetElementStyles(el: HTMLElement, ...styleProperties: string[]): HTMLElement {
```
mathisscott, line 43: This is a convenience utility that zeros out a list of style properties. Easier DX than using `updateElementStyles` set to empty strings.

packages/core/src/internal/utils/events.spec.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import { html } from 'lit';
-import { removeTestElement, createTestElement, onceEvent } from '@cds/core/test';
```
mathisscott, line 8: `onceEvent` was causing build warnings because it wasn't being used in this file.

packages/core/src/internal/utils/string.ts
```diff
@@ -121,3 +121,12 @@ export function removePrefix(str: string, prefix: string): string {
 export function removeSuffix(str: string, suffix: string): string {
   return removePrefixOrSuffix(str, suffix, 'suffix');
 }
+
+export function replaceWord(str: string, wordToReplace: string, replaceWith = '') {
```
mathisscott, line 125: This is less of a performance drag than bespoke regex replacement with the `g` flag. By default this utility will remove words from a string.

mathisscott:
CC: @rbirkgit

bdryanovski (approved):

packages/core/src/icon/icon.element.scss
```diff
@@ -28,33 +29,45 @@ svg {
...
 
-:host([size='lg']) {
+:host([size*='lg']) {
   // 36px
   @include equilateral(#{$cds-global-space-11});
+  @include min-equilateral(#{$cds-global-space-11});
 }
 
-:host([size='xl']) {
+:host([size*='xl']) {
```
Shijir, line 59: `size*='xl'` also covers `size*='xxl'` too, right? I think right now it works as `size*='xxl'` is defined after `size*='xl'`. But wondering if we need to tighten these selectors.

Shijir (approved):

packages/core/src/icon/icon.element.scss
mathisscott: You are correct that the order here is why the selectors work as expected. I cycled on possible ways to make that not so. But I don't want to make the size then dependent on the order that sizes appear in the attribute. 

So I will instead add a comment noting that the order here is important. That way, we have something in the code that explicitly tells us to keep the order in a certain way.

bbogdanov: @mathisscott Might be good to add this in our documentation for future reference

mathisscott: I don't intend to add this to our public documentation. There is an example of this feature in storybook and it is very much an edge case. Only a small percentage of our user base will need to know about it and i don't want to flood the zone with bells and whistles that most folks won't need.

Issue #6334, opened 2021-09-10
core(progress-circle): default aria-label value


astorije:
## Summary

- What is the change?

Add a default `aria-label="Loading…"` on the `cds-progress-circle` component. When `value` is used, this could instead be `Loading: ${value}%…`.

- Why should it go in Clarity?

Some Core components currently contain a default `aria-label`, such as the modal close button. This would be in line to provide a similar experience.

- Does this change impact existing behaviors? If so how?

It would impact existing use cases but not if the `aria-label` is already used: explicitly providing one would override the default value.

- If this change introduces a new behavior, is this behavior accessible?

I believe this would increase the accessibility of the existing component. There are internationalization considerations to keep in mind, but this would be no different than the modal close button label or similar other elements.

## Use case

At CloudHealth, we are replacing our existing circular progress component with Clarity's. Our component currently contains a default `aria-label` and we would like to avoid requiring all our developers to remember to add `aria-label` as it would be forgotten most of the time.

## API

There is no change to the existing API. However, if it is desirable to maintain the possibility to have no `aria-label`, one could explicitly give `aria-label=""`. It is probably not a good idea to not include an `aria-label` in general, so it seems like a good idea to make exceptions visually identifiable.

mathisscott:
this sounds reasonable to me. it may take us a few weeks to get to it though...

astorije:
Thanks @mathisscott! Let me know if I can help, I'd be happy to :)

PR #6444, opened 2021-10-29
feat(core:progress-circle): intelligently default aria label to i18n string

mathisscott:
This PR adds the following functionality:

- progress circles now have a default aria label if end-users do not define one
- the default aria label leans on new i18n keys in the i18n service
- overriding the i18n keys is one path for customization for the progress circle's aria labels
- if a custom `aria-label` is given, the progress-circle will ignore the default

#### Built for the future

Most of the work here translates easily to the progress bar. I've made notes in a few places where I think code can become shared between the progress circle and progress bar.

#### Advanced override

As an advanced customization users can override the default aria label's loading message composition. The default label is set up like `$i18n $val%` where the `$i8n` token represents the string stored in the i18n `progress.loading` key and the `$val` token represents the value of the progress circle. By default, this is transformed to `Loading 48%` where the progress circle's value is `48`. 

Through the following customization `<cds-progress-circle aria-label-template="Now $i18n... $val of 100" ...>` the default loading aria label's message would read `Now Loading... 48 of 100` and stay in sync with the changing value.

This was introduced as a way for users to customize/localize the loading message.

## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features)*

* - These are largely advanced customization features. A number of power users will use them. Most will not.

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The complaint was that creating an aria label that stayed in sync with the changing value of the progress circle was too onerous. The request was for the component to just handle it on its own.

Issue Number: #6334 

## What is the new behavior?

An auto syncing default are you label has been introduced. But done in a way that custom audio labels still override the default. There are a couple of tricky edge cases in here. But if you stay away from mixing the customization victors, you should be fine.

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/internal/decorators/i18n.spec.ts
+++ b/packages/core/src/internal/decorators/i18n.spec.ts
@@ -48,5 +48,8 @@ describe('i18n decorator', () => {
     component.setAttribute('cds-i18n', `{ "close": "${closeText}" }`);
     await componentIsStable(component);
     expect(component.i18n.close).toEqual(closeText);
+    component.setAttribute('cds-i18n', `{ "close": "ohai" }`);
+    await componentIsStable(component);
+    expect(component.i18n.close).toEqual('ohai', 'double set i18n');
   });
 });
--- a/packages/core/src/internal/services/i18n.service.ts
+++ b/packages/core/src/internal/services/i18n.service.ts
@@ -35,6 +35,10 @@ export interface I18nStrings {
     showButtonAriaLabel: string;
     hideButtonAriaLabel: string;
   };
+  progress: {
+    loading: string;
+    looping: string;
+  };
   treeview: {
     loading: string;
   };
@@ -72,6 +76,10 @@ export const componentStringsDefault = {
   password: {
     showButtonAriaLabel: 'Show password',
     hideButtonAriaLabel: 'Hide password',
+  },
+  progress: {
+    loading: 'Loading',
+    looping: 'Loading',
   },
   treeview: {
     loading: 'Loading',
--- a/packages/core/src/internal/utils/identity.spec.ts
+++ b/packages/core/src/internal/utils/identity.spec.ts
@@ -47,6 +47,8 @@ describe('Functional Helper: ', () => {
       expect(isNilOrEmpty({ item: 1 })).toEqual(false);
       expect(isNilOrEmpty([1, 2])).toEqual(false);
       expect(isNilOrEmpty('ohai')).toEqual(false);
+      expect(isNilOrEmpty(0)).toEqual(false);
+      expect(isNilOrEmpty(-1)).toEqual(false);
       expect(isNilOrEmpty(true)).toEqual(false);
       expect(isNilOrEmpty(false)).toEqual(false);
     });
--- a/packages/core/src/progress-circle/progress-circle.element.spec.ts
+++ b/packages/core/src/progress-circle/progress-circle.element.spec.ts
@@ -155,4 +155,147 @@ describe('progress circle element – ', () => {
     });
   });
   // circle classname arcstroke/backstroke
+
+  describe('ariaLabel', () => {
+    it('should show indeterminate label if no label is set and there is no value', async () => {
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('aria-label')).toBe('Loading');
+    });
+    it('should show loading label and value if no label is set', async () => {
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label')).toBe('Loading 49%');
+      component.value = 80;
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label')).toBe('Loading 80%');
+    });
+    it('should not show default label if a custom label is set', async () => {
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('aria-label')).toBe('Loading');
+      componentUnset.setAttribute('aria-label', 'ohai');
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('aria-label')).toBe('ohai');
+
+      await componentIsStable(component);
+      component.setAttribute('aria-label', 'ohai');
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label')).toBe('ohai');
+      component.value = 88;
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label')).toBe('ohai');
+    });
+  });
+
+  describe('i18n', () => {
+    const newLoading = 'ohai';
+    const newLooping = 'kthxbye';
+
+    it('uses i18n strings by default', async () => {
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label').indexOf(component.i18n.loading) > -1).toBe(
+        true,
+        'has loading i18n string'
+      );
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('aria-label').indexOf(componentUnset.i18n.looping) > -1).toBe(
+        true,
+        'has looping i18n string'
+      );
+    });
+    it('updates looping i18n strings as expected', async () => {
+      await componentIsStable(componentUnset);
+      componentUnset.setAttribute('cds-i18n', `{ "loading": "${newLoading}", "looping": "${newLooping}" }`);
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('aria-label').indexOf(newLooping) > -1).toBe(true);
+    });
+    it('updates loading i18n strings as expected', async () => {
+      await componentIsStable(component);
+      component.setAttribute('cds-i18n', `{ "loading": "${newLoading}", "looping": "${newLooping}" }`);
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label').indexOf(newLoading) > -1).toBe(true);
+    });
+  });
+
+  describe('setAriaAttributes', () => {
+    it('should return expected for indeterminate progress', async () => {
+      await componentIsStable(componentUnset);
+      componentUnset.setAriaAttributes();
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('role')).toBe('img');
+      expect(componentUnset.hasAttribute('aria-valuemin')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuemax')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuenow')).toBe(false);
+      expect(componentUnset.getAttribute('aria-label')).toBe('Loading');
+
+      componentUnset.ariaLabel = 'howdy';
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('role')).toBe('img');
+      expect(componentUnset.hasAttribute('aria-valuemin')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuemax')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuenow')).toBe(false);
+      expect(componentUnset.getAttribute('aria-label')).toBe('howdy');
+    });
+    it('should return as expected for progress with value', async () => {
+      await componentIsStable(component);
+      component.setAriaAttributes();
+      await componentIsStable(component);
+      expect(component.getAttribute('role')).toBe('progressbar');
+      expect(component.getAttribute('aria-valuemin')).toBe('0');
+      expect(component.getAttribute('aria-valuemax')).toBe('100');
+      expect(component.getAttribute('aria-valuenow')).toBe('49');
+      expect(component.getAttribute('aria-label')).toBe('Loading 49%');
+
+      component.value = 0;
+      await componentIsStable(component);
+      expect(component.getAttribute('role')).toBe('progressbar');
+      expect(component.getAttribute('aria-valuemin')).toBe('0');
+      expect(component.getAttribute('aria-valuemax')).toBe('100');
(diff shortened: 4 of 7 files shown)


packages/core/src/internal/decorators/i18n.spec.ts
```diff
@@ -48,5 +48,8 @@ describe('i18n decorator', () => {
     component.setAttribute('cds-i18n', `{ "close": "${closeText}" }`);
     await componentIsStable(component);
     expect(component.i18n.close).toEqual(closeText);
+    component.setAttribute('cds-i18n', `{ "close": "ohai" }`);
+    await componentIsStable(component);
+    expect(component.i18n.close).toEqual('ohai', 'double set i18n');
```
mathisscott, line 53: I added this test to force a reset of the i18n string. The test passes but, as noted below, this functionally does not work in all cases.

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -73,6 +77,10 @@ export const componentStringsDefault = {
     showButtonAriaLabel: 'Show password',
     hideButtonAriaLabel: 'Hide password',
   },
+  progress: {
```
mathisscott, line 80: New i18n keys/strings.

packages/core/src/internal/utils/identity.spec.ts
```diff
@@ -47,6 +47,8 @@ describe('Functional Helper: ', () => {
       expect(isNilOrEmpty({ item: 1 })).toEqual(false);
       expect(isNilOrEmpty([1, 2])).toEqual(false);
       expect(isNilOrEmpty('ohai')).toEqual(false);
+      expect(isNilOrEmpty(0)).toEqual(false);
```
mathisscott, line 50: Needed to prove out that these cases passed/failed as expected.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -70,15 +79,8 @@ export class CdsProgressCircle extends LitElement {
...
-    // we need to keep this so that circular-progress remains aligned with our icons
-    // we shouldn't change it unless we change the default viewbox of the icons!
-    const halfOfViewbox = 18;
-
-    // line offset takes the width/thickness of the progress circle into account
-    const lineOffset = Math.ceil(this.line / 2);
-
-    return halfOfViewbox - lineOffset;
+    // 36 is the default viewbox dimensions
+    return getProgressCircleRadius(this.line);
```
mathisscott, line 83: Moved to a utility so I could test/prove the expected output in isolation.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
     }
   }
 
-  private setAriaAttributes() {
+  private _ariaLabel: string | undefined | null;
+
+  /**
```
mathisscott, line 122: See comment.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
...
+   * the assumption is that knowledge of this feature is deep enough to manage
+   * the side effects.
+   *
+   * @memberof CdsProgressCircle
+   */
+  @property({ type: String })
+  ariaLabelTemplate = '$i18n $val%';
+
+  /**
+   * The aria-label attribute is added here as a convenience. It is not set
```
mathisscott, line 123: See comment.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
...
+    if (hasStringPropertyChanged(newAriaLabel, this._ariaLabel as string)) {
+      const oldVal = this._ariaLabel;
+      this._ariaLabel = newAriaLabel;
+      this.requestUpdate('ariaLabel', oldVal);
+    }
+  }
+
+  @i18n() i18n = I18nService.keys.progress;
+
+  // note: this aria attr logic could be reused when we introduce progress bars
```
mathisscott, line 149: See comment.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
...
+  // consider promoting this to a shared utility in internal in the future
+  /** @private */
+  setAriaAttributes(oldValueForAriaLabelCheck?: number) {
     const currentValue = this.value;
-    const ariaAttrsIfValue: HTMLAttributeTuple[] = [
```
mathisscott, line 120: I split this code out into a utility. So it could be tested in isolation. It could also be used as-is for progress bars in the future.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -139,10 +198,39 @@ export class CdsProgressCircle extends LitElement {
...
+    this.observers.push(
+      listenForAttributeChange(this, 'cds-i18n', () => {
+        const oldAriaLabel = this.ariaLabel;
+        this.ariaLabel = '';
+        this.requestUpdate('ariaLabel', oldAriaLabel);
+      })
+    );
+  }
+
+  // note: this update logic could be reused when we introduce progress bars
```
mathisscott, line 204: See comment.

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
...
+  /**
+   * Updating the ariaLabelTemplate is an advanced edge case. Note that it will
+   * force a reset of the aria-label. If manipulating the aria label template,
+   * the assumption is that knowledge of this feature is deep enough to manage
+   * the side effects.
+   *
+   * @memberof CdsProgressCircle
+   */
+  @property({ type: String })
+  ariaLabelTemplate = '$i18n $val%';
```
coryrylan: I'm not sure if a new syntax is needed to support this. Since the component is already stateless and is the source of truth for `value` then the app can set the label like a regular attr/property.

https://stackblitz.com/edit/http-server-fbpgdj?file=src%2FApp.tsx
```jsx
<CdsProgressCircle size="lg" value={this.state.value} aria-label={`Searching ${this.state.value} of ${this.state.total}`}>
  Hello There
</CdsProgressCircle>
```

https://stackblitz.com/edit/node-4u87qx?file=src%2Fapp%2Fapp.component.html
```html
<cds-progress-circle size="lg" [value]="value" [attr.aria-label]="'Loading ' + value + ' of ' + total"></cds-progress-circle>
```

If thats the case then the default should become simpler to create. This also potentially prevents excess updates from the framework since there is no risk of the setter/getter being different.

```typescript
  @i18n() i18n = I18nService.keys.progress;

  @property({ type: String }) ariaLabel: string;

  updated(props: Map<string, any>) {
    super.updated(props);
    this.setAriaAttributes();

    if (props.has('value') && (this.value !== props.get('value'))) {
      this.ariaLabel = this.ariaLabel ?? this.i18n.loading;
    } else {
      this.ariaLabel = this.ariaLabel ?? this.i18n.looping;
    }
  }
```

packages/core/src/progress-circle/progress-circle.element.spec.ts
```diff
@@ -155,4 +155,160 @@ describe('progress circle element – ', () => {
...
   // circle classname arcstroke/backstroke
+
+  describe('ariaLabelTemplate', () => {
+    it('should return expected default aria label template', async () => {
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label')).toBe('Loading 49%');
+    });
+    it('should allow for custom aria label template', async () => {
+      await componentIsStable(component);
+      component.setAttribute('aria-label-template', 'Now $i18n... $val out of 100 items');
```
Jinnie: I don't clearly understand the idea about the template. Isn't it enough that they can override the label with their own custom string? Why do we want to handle these dynamic updates internally?

packages/core/src/progress-circle/progress-circle.element.spec.ts
```diff
@@ -155,4 +155,160 @@ describe('progress circle element – ', () => {
...
+      component.setAttribute('cds-i18n', `{ "loading": "${newLoading}", "looping": "${newLooping}" }`);
+      await componentIsStable(component);
+      expect(component.getAttribute('aria-label').indexOf(newLoading) > -1).toBe(true);
+    });
+  });
+
+  describe('setAriaAttributes', () => {
+    it('should return expected for indeterminate progress', async () => {
+      await componentIsStable(componentUnset);
+      componentUnset.setAriaAttributes();
```
Jinnie, line 221: Why do we need this call? Initialize aria attributes manually? Why would we want it uninitialized.

note from myself after reading more code:
Obviously it resets/triggers the defaults for the min/max/now attributes. I still don't have a good understanding about the balance between automation and manual control with this approach.

packages/core/src/progress-circle/progress-circle.element.spec.ts
```diff
@@ -155,4 +155,160 @@ describe('progress circle element – ', () => {
...
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('role')).toBe('img');
+      expect(componentUnset.hasAttribute('aria-valuemin')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuemax')).toBe(false);
+      expect(componentUnset.hasAttribute('aria-valuenow')).toBe(false);
+      expect(componentUnset.getAttribute('aria-label')).toBe('Loading');
+
+      componentUnset.ariaLabel = 'howdy';
+      await componentIsStable(componentUnset);
+      expect(componentUnset.getAttribute('role')).toBe('img');
```
Jinnie, line 231: Why do we change between img and progressbar? Isn't it still a progressbar when uninitialized?

packages/core/src/progress-circle/progress-circle.element.ts
```diff
@@ -115,21 +117,78 @@ export class CdsProgressCircle extends LitElement {
...
-      ['aria-valuemin', false],
-      ['aria-valuemax', false],
-      ['aria-valuenow', false],
-    ];
-    const attrsToSet = isNil(currentValue) ? ariaAttrsNoValue : ariaAttrsIfValue;
+    const oldValIsNil = isNilOrEmpty(oldValueForAriaLabelCheck);
+    let ariaLabel: string;
+
+    if (oldValIsNil && !isNilOrEmpty(currentValue) && this.ariaLabel === this.i18n.looping) {
+      if (!isNilOrEmpty(currentValue) && this.ariaLabel === this.i18n.looping) {
```
Jinnie, line 158: Isn't this "if" already included in the above? Is it possible to get in the else case at all?
Maybe we can extract this same line to a named method, helping for readability. At least that's the advise I got from you for a similar block of code in a previous PR :)

Jinnie (commented):
I have several questions below, mostly helping me understand the desired API and behavior.

packages/core/src/internal/decorators/i18n.spec.ts
mathisscott: updated. works in all cases for now.

packages/core/src/progress-circle/progress-circle.element.ts
mathisscott: Per conversations over the last couple of days...

This feature (the aria label template) has been removed. We would be better served by extending the i18n service than doing a one-off just for progress circles.

packages/core/src/progress-circle/progress-circle.element.spec.ts
mathisscott: this feature has been removed in favor of a more general implementation

packages/core/src/progress-circle/progress-circle.element.spec.ts
mathisscott: the automation here is a convenience. setting them manually at this point is a way to dodge issues with how react supports aria attributes (specifically how it struggles with removing aria attributes). once that support improves, we may be able to set these as properties on the host component and derive their values through getters or as a set default.

packages/core/src/progress-circle/progress-circle.element.spec.ts
mathisscott: when the progress does not have a value, it is a "looping"/"spinning"/"indeterminate" progress. per a11y, that is considered an image. it is not `role="progress"` because it has no value, value-min, or value-max.

packages/core/src/progress-circle/progress-circle.element.ts
mathisscott: good catch. i simplified this logic. ❤️

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -73,6 +77,10 @@ export const componentStringsDefault = {
...
+    loading: 'Loading',
+    looping: 'Loading',
```
Jinnie, line 82: looping is 'Loading' too?

Jinnie (changes requested):
Something small in the i18n, the rest looks good to me.

packages/core/src/internal/services/i18n.service.ts
mathisscott: Presently, yes, as a default. If UX wants it to be something different, it's a minor change. If end users want it to be different, they can change it in the i18n here or with a custom aria-label.

bdryanovski (approved):

Jinnie (approved):

PR #6455, opened 2021-11-08
chore: update to Angular 13

bdryanovski:
# Overview

 - Update TypeScript to `4.4.4`
 - Update cds/angular to use Angular 13.0.0
 - Update `golden/clr-angular.d.ts`
 - Update `@cds/core` & `@cds/angular` (`"^5.6.0"`) used into sub packages to latest version - to resolve some warnings when building.
 - Migrate files to use Angular 13
 - Fixing Dev app
 - Fixing Clarity Schemtics to work Angular 13
 - Fixing `build:icons:zip`
 - Removing `build:icons:zip` from build steps - zip files no longer part of @clr/icons
 - Updating development packages to resolve warnings (`@typescript-eslint/eslint-plugin` & `@typescript-eslint/parser`)
 - Updating code to handle ClientRect deprecation
 - Drop support of Internet Explorer polyfils

This PR includes work based on collaboration with @bbogdanov and @Jinnie 

### How to test it

```bash

# checkout this branch 

$ git clean -fdx;  # make sure that everything is clean (node_modules, dist, golden files)

$ npm install  

$ npm run build:ci
```

To test the dev app:

```bash
$ npm run start
```

To test website

```bash
$ npm run start:website
```

### Testing packages

Setup testing tools

```bash
$ npm install --global verdaccio

$ verdaccio
```

Release packages `locally`

```bash
$ npm run publish:local

```

Check package information and dependencies
```bash
$ npm info @clr/angular --registry http://localhost:4873

```

Use the new package to create new application
```bash
$ npx @angular/cli@13 new demo-app

$ cd demo-app

$ npx @angular/cli@13 add @clr/angular@13.0.0-next --registry http://localhost:4873

$ npm run start
```

### Performance

```bash
$ npm install;

added 4467 packages from 2332 contributors and audited 4015 packages in 105.532s

$ npm run build:ci

265.35s user 25.50s system 139% cpu 3:29.22 total
```

### Known issues

When building, there is a warning that if I'm not wrong, it is already fixed, but it seems not backported. Need investigation. 
```bash
Recommendation: math.div($clr-baseline-denominator, $browser-default-size)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
18 │ $clr-basefont-size: ($clr-baseline-denominator/$browser-default-size) * 100% !default;
   │                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
[...]

PR #6456, opened 2021-11-08
fix: support native aria property reflection

coryrylan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The newer aria reflection AP enables elements to set an aria property and have it auto reflect to the attribute.

```javascript
const element = document.querySelector('div');
element.ariaDisabled = 'true'
```

```html
<div aria-disabled="true"></div>
```

This means we do not need the property decorators to handle the reflection on `aria*` properties and need to adjust the types to match TypeScript in the newer version 4.3.4+. To support the newer versions of TS we need to adjust and remove overrides on any native aria property type. We also need a shim for firefox as it does not fully support aria property reflection yet.
https://wicg.github.io/aom/caniuse.html
https://wicg.github.io/aom/aria-reflection-explainer.html

## What is the new behavior?
Components can now use the native aria properties and fallback to a shim for Firefox. This aligns and fixes the mismatched types with TypeScript. See https://github.com/microsoft/TypeScript/issues/46456 for more details

Unblocks part of the Angular 13 update https://github.com/vmware/clarity/pull/6455 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/alert/alert-group.element.ts
+++ b/packages/core/src/alert/alert-group.element.ts
@@ -71,12 +71,6 @@ export class CdsAlertGroup extends LitElement {
   type: AlertGroupTypes = 'default';
 
   /**
-   * Autosets the alert groups aria role to 'region'
-   */
-  @property({ type: String })
-  role = 'region';
-
-  /**
    * Sets the status of the alerts inside the alert group
    * @type {neutral | info | success | warning | danger | alt | loading}
    */
@@ -109,6 +103,11 @@ export class CdsAlertGroup extends LitElement {
         </div>
       </div>
     `;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.role = 'region';
   }
 
   firstUpdated(props: Map<string, any>) {
--- a/packages/core/src/internal-components/overlay/overlay.element.ts
+++ b/packages/core/src/internal-components/overlay/overlay.element.ts
@@ -101,7 +101,8 @@ export class CdsInternalOverlay extends CdsBaseFocusTrap implements Animatable {
   // renderRoot needs delegatesFocus so that focus can cross the shadowDOM
   // inside an element with aria-modal set to true
   /** @private */
-  static get shadowRootOptions() {
+  static get shadowRootOptions(): any {
+    // any is used until TS 4.4.x adopted through other @cds/* libraries. Can be removed in 6.0
     return { ...super.shadowRootOptions, delegatesFocus: true };
   }
 
--- a/packages/core/src/internal/index.ts
+++ b/packages/core/src/internal/index.ts
@@ -4,10 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-/** @internal private module to Clarity Core */
-
+import '@cds/core/polyfills';
 import styles from './base/base.element.scss';
 
+/** @internal private module to Clarity Core */
 export const baseStyles = styles;
 export * from './base/button.base.js';
 export * from './base/focus-trap.base.js';
--- a/packages/core/src/polyfills/aria-reflect.ts
+++ b/packages/core/src/polyfills/aria-reflect.ts
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+/*
+ * Shim for aria atribute reflection missing in Firefox
+ * https://wicg.github.io/aom/aria-reflection-explainer.html
+ * https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabel
+ * Patching missing types as needed
+ */
+declare global {
+  interface Element {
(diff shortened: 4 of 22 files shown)


packages/core/src/polyfills/aria-reflect.ts
```diff
@@ -0,0 +1,104 @@
...
+    role: string | null;
+    // ARIAMixin interface in TS lib.dom.d.ts https://github.com/microsoft/TypeScript/issues/46456
+    ariaActiveDescendant: string | null;
+    ariaControls: string | null;
+    ariaLabelledBy: string | null;
+    ariaDisabled: string | null;
+    ariaExpanded: string | null;
+    ariaSelected: string | null;
+  }
+}
```
coryrylan, line 57: This augments some of the aria properties as nullable. TypeScript marks them as string only but this is incorrect and causing issues when trying to set them as null. Once this is fixed in TS we can remove the overloaded `Element` interface.

 This is impacting other Design systems as well
 https://github.com/microsoft/TypeScript/issues/46456  
 https://github.com/microsoft/TypeScript/issues/45047

packages/core/src/polyfills/aria-reflect.ts
```diff
@@ -0,0 +1,104 @@
...
+    },
+    set: function (value) {
+      if (value !== null) {
+        this.setAttribute(attributeName, value);
+      } else {
+        this.removeAttribute(attributeName);
+      }
+    },
+  });
+}
```
coryrylan, line 141: This is a patch to support the aria property reflection for firefox
https://wicg.github.io/aom/caniuse.html
https://wicg.github.io/aom/aria-reflection-explainer.html

bdryanovski (approved):

packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -101,7 +101,8 @@ export class CdsInternalOverlay extends CdsBaseFocusTrap implements Animatable {
   // renderRoot needs delegatesFocus so that focus can cross the shadowDOM
   // inside an element with aria-modal set to true
   /** @private */
-  static get shadowRootOptions() {
+  static get shadowRootOptions(): any {
+    // any is used until TS 4.4.x adopted through other @cds/* libraries. Can be removed in 6.0
```
coryrylan, line 105: I set these to any so the 5.x @cds/* libraries continue to work in the older TS verions. Once we branch for 6.x we can remove the `any` and use the updated types that TS provides.

hippee-lee (approved):

packages/core/src/internal-components/overlay/overlay.element.ts
mathisscott: Zero concerns about leaving `any` here as long as we want.

packages/core/src/internal/index.ts
```diff
@@ -4,10 +4,10 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-/** @internal private module to Clarity Core */
-
+import '@cds/core/polyfills';
```
mathisscott, line 7: Most of our other imports point to a specific `.js` file. Should this one do that too?

`import '@cds/core/polyfills/index.js` vs. `import '@cds/core/polyfills'`?

mathisscott (approved):
Question about a barrel import. Beyond that, 👍

packages/core/src/internal/index.ts
coryrylan: `.js` is required for relative imports or imports that by pass an entrypoint module (we typically avoid but `register.js` would be an example exception). This is mainly to prevent users from reaching into modules that are [intended to be private](https://unpkg.com/@cds/core@5.6.0/package.json). I made polyfills an entrypoint to ensure its side effect of running the polyfill code and not getting tree-shaken away by bundlers. So now in the generate package.json it will have these entries added,

```json
"sideEffects": [
    "./polyfills/index.js" 
],
"exports": {
    "./polyfills": "./polyfills/index.js",
}
```

packages/core/src/polyfills/aria-reflect.ts
```diff
@@ -0,0 +1,108 @@
...
+    ariaExpanded: string | null;
+    ariaSelected: string | null;
+  }
+}
+
+let roleRegistered = false;
+let ariaRegistered = false;
+
+function isNode() {
+  return (globalThis as any)?.process?.env?.JEST_WORKER_ID !== undefined; // Jest and JSDom breaks on property reflection
```
coryrylan, line 63: hack: Jest throws since it doesn't seem to support aria properties correctly, once the shim is no longer required for firefox we can delete it

packages/core/src/alert/alert-group.element.ts
```diff
@@ -111,6 +105,11 @@ export class CdsAlertGroup extends LitElement {
     `;
   }
 
+  connectedCallback() {
+    super.connectedCallback();
+    this.role = 'region';
```
mathisscott, line 110: Other files we are just setting `role = 'region'` as a property on the class. Why are we doing it in `connectedCallback` here?

coryrylan: I had it as a property as you describe but then hit issues with Jest, I can double check to see if the isNode check prevents the issue here

PR #6466, opened 2021-11-11
fix(react): move react to peer dependency to allow React 17

### LINKED ISSUE #6464: @cds/react doesn't technically support React 17
## Describe the bug

The package.json requires a 16.x version of react. When using Clarity in a React 17 app, this results in two React versions in the yarn.lock. Two versions of react, especially two different major versions can cause problems in an app.

https://github.com/vmware/clarity/blob/next/packages/react/package.json#L27

I believe that you may have to switch React to a peer dependency (and a devDependency for development) and then you can support both versions: `"react": "^16.13.1 || ^17.0.2"`

## How to reproduce

Create a create-react-app app, add @cds/core and @cds/react and then check the yarn.lock file and notice a 16.X an 17.X version of react/react-dom. 

## Expected behavior

Clarity requires either v16 or v17.

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.x

**Framework:**

- [ ] Angular
- [x] React
- [ ] Vue
- [ ] Other:

**Framework version:**
React 17

**Device:**

- Type: [e.g. MacBook]

## Additional notes

ashleyryan:
Fix #6464

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue #6464 

@cds/react has react 16 as a dependency which can cause two version of react if apps are using v17.

## What is the new behavior?

react is now a peer dependency 

- [x] Yes
- [ ] No

I think moving a dependency to a peerDependency is technically a breaking change, but there's no way an app would use @cds/react without already having react/react-dom.

## Other information

There's probably a lot more to test with this, I was hoping this could be a jumping off point for discussion

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -23,12 +23,12 @@
     "url": "https://github.com/vmware/clarity.git"
   },
   "dependencies": {
-    "@lit-labs/react": "^1.0.1",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "@lit-labs/react": "^1.0.1"
   },
   "peerDependencies": {
-    "@cds/core": "./dist/core"
+    "@cds/core": "./dist/core",
+    "react":">= 16.13.1",
+    "react-dom": ">= 16.13.1"
   },
   "devDependencies": {
     "@cds/core": "./dist/core",
@@ -57,6 +57,8 @@
     "parcel": "^2.0.0-beta.1",
     "prettier": "2.0.5",
     "pretty-quick": "1.11.0",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
     "replace": "1.1.0",
     "ts-jest": "^26.1.0",
     "typescript": "^4.4.2"


packages/react/package.json
```diff
@@ -23,12 +23,12 @@
...
-    "@lit-labs/react": "^1.0.1",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "@lit-labs/react": "^1.0.1"
   },
   "peerDependencies": {
-    "@cds/core": "./dist/core"
+    "@cds/core": "./dist/core",
+    "react":"^16.13.1 || ^17.0.2",
+    "react-dom": "^16.13.1 || ^17.0.2"
```
astorije-vmware: It's probably fine to be a little looser for v17:

```suggestion
    "react":"^16.13.1 || 17",
    "react-dom": "^16.13.1 || 17"
```

(Maybe even for v16 but hard to say for sure)

packages/react/package.json
```diff
@@ -23,12 +23,12 @@
...
   "dependencies": {
-    "@lit-labs/react": "^1.0.1",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "@lit-labs/react": "^1.0.1"
   },
   "peerDependencies": {
-    "@cds/core": "./dist/core"
+    "@cds/core": "./dist/core",
+    "react":"^16.13.1 || ^17.0.2",
```
gnomeontherun: Do we want to relax this further like `>16.13.1`?

ashleyryan: Probably - that would include the React 18 beta, which you may or may not want to support yet

packages/react/package.json
bbogdanov: @ashleyryan If we don't expect React to change the behaviour of the code we use what do you think about doing what we did for the [@cds/angular](https://github.com/vmware/clarity/blob/4006bea6dfe7b31a4708cc656dc628aad47562fa/packages/angular/projects/cds-angular/package.json#L17)?

We had issues before when a newer version of the Angular was released. The compiler was raising an error that `@cds/angular` expects to work with 3 specific versions of Angular which was not true and we had to relax the version specification. 

I believe that's what @gnomeontherun had in mind with [this comment](https://github.com/vmware/clarity/pull/6466/files#r751557277)

ashleyryan: Right, I assumed that's what he was getting at - changing the peer dependency to be >16. They've been making a lot of changes in React 18 and I don't know how that will behave with Clarity, which is why I'm hesitant to add blanket support for it. 

But it's only in Beta so I think it would be safe to say `> 16` with a disclaimer that we don't guarantee React 18 support yet

gnomeontherun: I prefer to keep it less strict unless we know it doesn't work, otherwise we might not get feedback because people won't be able to install it.

ashleyryan: Makes sense - updated!

mathisscott (approved):

coryrylan (approved):

PR #6488, opened 2021-11-23
fix(core): lint/logging

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
This fixes the last linting/warning errors in the Core codebase for both the build and tests. There should be no warnings/logs in the console now.

- remove last of the deprecated jasmine done use/warnings 
- update private property to use `@state` to remove private api warning
- fix statement warning caught by TS/ESBuild compiler (boolean check guaranteed will always be true)
- fix the unused public export warning from rollup

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/core/src/internal-components/popup/popup.element.ts
+++ b/packages/core/src/internal-components/popup/popup.element.ts
@@ -23,6 +23,7 @@ import {
   reverseAnimation,
   setAttributes,
   setPopupPosition,
+  state,
 } from '@cds/core/internal';
 import { html } from 'lit';
 import { query } from 'lit/decorators/query.js';
@@ -102,7 +103,7 @@ export class CdsInternalPopup extends CdsInternalStaticOverlay implements Animat
    * positioning logic may override your changes.
    */
   /** @private */
-  @property({ type: Boolean, reflect: true })
+  @state({ type: Boolean, reflect: true, attribute: 'responsive' })
   responsive = false;
 
   @event()
--- a/packages/core/src/internal/motion/utils.spec.ts
+++ b/packages/core/src/internal/motion/utils.spec.ts
@@ -109,7 +109,7 @@ describe('Animation Helpers: ', () => {
   ];
 
   describe('runPropertyAnimations()', () => {
-    it('logs a warning if the element it is passed is not animatable', async done => {
+    it('logs a warning if the element it is passed is not animatable', async () => {
       const testElement = await createTestElement(html`<div>ohai</div>`);
       spyOn(LogService, 'warn');
       expect(testElement).toBeDefined();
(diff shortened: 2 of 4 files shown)


packages/core/src/internal/utils/focus-trap.ts
```diff
@@ -51,7 +51,7 @@ export function refocusIfOutsideFocusTrapElement(
     const focusableChildren = queryAllFocusable(focusTrapElement);
     const orderedFocusableChildrenAsArray = reorderCloseButtonSlot(Array.from(focusableChildren));
 
-    if (isReboundEl !== null && orderedFocusableChildrenAsArray !== []) {
+    if (isReboundEl !== null) {
```
coryrylan, line 54: Because the functions were typed checked TS/ESBuild catches this as a statement that will always be true

packages/core/src/internal/motion/utils.spec.ts
```diff
@@ -170,10 +166,9 @@ describe('Animation Helpers: ', () => {
       const didRun = await runPropertyAnimations(propMap, (component as unknown) as AnimatableElement);
       expect(didRun).toBe(true);
       removeTestElement(testElement);
-      done();
     });
 
-    it('does NOT run if there is an animation label associated with the property value but no animation in ClarityMotion', async done => {
+    it('does NOT run if there is an animation label associated with the property value but no animation in ClarityMotion', async () => {
```
coryrylan, line 171: Jasmine deprecated the `done` function when used in an async function. This was the last of them and the warning will no longer be in the build console

packages/core/src/internal-components/popup/popup.element.ts
```diff
@@ -101,8 +102,7 @@ export class CdsInternalPopup extends CdsInternalStaticOverlay implements Animat
    * do not manipulate the responsive property because the
    * positioning logic may override your changes.
    */
-  /** @private */
-  @property({ type: Boolean, reflect: true })
+  @state({ type: Boolean, reflect: true, attribute: 'responsive' })
```
coryrylan, line 106: Fix for lit code linter to use `@state` on private api properties rather than `@property`

mathisscott: 🤔

I made this a "private" property on purpose because code outside of the popup (like the animations) needed to reference it. If we are reflecting it, it may _still_ work because I added smarts to the animations to look up values in attributes (and not just object properties).

We just want to be sure that the responsive modals/dropdowns and animations work as expected here.

mathisscott (changes requested):

packages/core/src/internal-components/popup/popup.element.ts
coryrylan: It still works, `@state` uses `@property` under the hood defaulting to not reflecting. It's just a way to mark that it's a private/internal property rather than public to consumers outside of using TS. We do this in a lot of places where we need to [communicate between components](https://github.com/vmware/clarity/blob/next/packages/core/src/navigation/navigation-item.element.ts#L54) but mark it as a non-public API. The web component metadata generator will treat `@state` as private https://lit.dev/docs/api/decorators/#state I went ahead and added the jsdoc private back, that makes sense to keep.

Side note, thinking about it we probably should be using `/** @private */` + `@state()` when not using TS `private` to get the maximum DX tooling support. 🤔

mathisscott (approved):

hippee-lee (approved):

PR #6493, opened 2021-11-30
feat: test if content is scrollable before set tabindex for modal body

KingMario:
Signed-off-by: KingMario <gcyyq@hotmail.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

tabindex is set for modal body for no condition.

Issue Number: N/A

## What is the new behavior?

test if content is scrollable before set tabindex for modal body

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/modal/modal-body.ts
+++ b/projects/angular/src/modal/modal-body.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Directive, ElementRef, NgZone, OnDestroy, Renderer2 } from '@angular/core';
+import { AfterViewChecked, Directive, ElementRef, NgZone, OnDestroy, Renderer2 } from '@angular/core';
 
 /**
  * Allows modal overflow area to be scrollable via keyboard.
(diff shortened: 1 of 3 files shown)


bbogdanov:
Could you please provide us more context about this feature? 

Why we would need to add a new class and how is this class going to stop the keyboard scrolling functionality?

KingMario:
Hi @bbogdanov,

We have a VPAT jira ticket complains that while using tab key to navigate inside a modal dialog, the focus is moved to a `hidden` element before the first input field. It turns out that, the `hidden` element is the modal body element.

From the WCAG document, the tab stop should be set on an **operable** element.

The ClrModalBody directive helps add the tabindex attribute to the modal body, so that the element can be focused. And when focused, it can be scrolled by up and down arrow keys.

But if the modal content is not higher than the modal height, the modal body is **not** operable and should not be focusable.

The updated selector won't affect the original behavior of an element with `modal-body` class as usual. When the additional `no-tab-stop` class is added to a `.modal-body` element, the element doesn't match the updated selector `.modal-body:not(.no-tab-stop)`, hence, the directive is not hit and the focusable behavior is not applied.

KingMario:
Hi @bbogdanov,

I've just updated the directive to test if content is scrollable before set tabindex for modal body. Please review again.

Thanks

projects/angular/src/modal/modal-body.ts
```diff
@@ -35,6 +38,17 @@ export class ClrModalBody implements OnDestroy {
     });
   }
 
+  ngAfterViewChecked(): void {
```
gnomeontherun, line 41: Why do you think we should run this after every view checked? This would likely be something you set once unless the browser is resized.

gnomeontherun:
This is interesting but I'm concerned about how it is actually executing, and what side effects might be caused particularly in wizards. My concern is mostly on performance, if we are running this check at the right times.

projects/angular/src/modal/modal-body.ts
KingMario: Sometimes a dynamic grid or a dynamic form is put inside a dialog, then the size is not fixed on user's actions. And yes, the performance impact is an issue we should take care of.

steve-haar:
I agree that it's not ideal to run this check after every AfterViewChecked. One solution I could think of would be to observe height changes on the modal's content and only run this check when the height changes. To do this we could use the `ResizeObserver` api.

However, this approach has a few drawbacks:

1) `ResizeObserver` is not supported in all browsers and we may have to add a polyfill depending on our browser support (https://caniuse.com/resizeobserver)

2) The modal-body currently uses `overflow: hidden` and so we'd have to add another wrapper element inside the modal body in order to detect height changes (`ResizeObserver` doesn't detect scrollHeight). This would probably be a breaking change.

For these reasons, I don't recommend doing this approach, just trying to be thorough in what are options are.

mathisscott:
> 1. `ResizeObserver` is not supported in all browsers and we may have to add a polyfill depending on our browser support (https://caniuse.com/resizeobserver)

We do not support IE11 and `ResizeObserver` is in all the versions of all the browsers we support. So no. 1 here is a non-concern.

As for adding a new element inside of `.modal-body`. I think it's possible to do that without a breaking change to the end users. But it's been a while since I've been in the `@clr/angular` modal code.

steve-haar:
closing this as it should be fixed with https://github.com/vmware/clarity/pull/6546

PR #6497, opened 2021-12-03
build(angular): update dependencies for angular 13

hippee-lee:
This change updates some deps to address some github security warnings.
This ended up being a little more than I expected. It turns out angular
is roughly coupled to a few dependencies: typescript and eslint come to
mind in this case. I updated eslint because there was an issue with type
script. I bumped typescript (but not to latest yet) because angular 13
needs 4.4.2. Here is a list of things affected from the change:

- Some of our code was failing the eslint type checks so I updated the
  code or ignored in one case.
- Added a .gitignore rule because angular 13 generates an .angular/cache
  folder. Options were to turn it off or ignore it so I chose to ignore
it.
- remember: another reason for bumping the root level typescript and the
  package/schematic typescript is because the @scehamtics package pulls
in typescript as dep but they do it from a third-party folder inside
their package. This means that the @clr/schematics typescript version
and needs to match that typescript version or the schematics will not
build.
- updated lite-server dep to address gh sec warnings
- updated @commitlint deps to address gh sec warnings
- updated conventional-changelog-cli dep to address gh sec warnings
- updated copy-cli dep to address gh sec warnings
- 

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes

@cds/angular package does not build with angular 13 and some of our devDeps cause gh security warnings.

Issue Number: N/A

## What is the new behavior?
@cds/angular builds with angular 13, still ned to confirm the gh security warnings. 

- [ x] Yes
- [ ] No

## Other information
Angular applications will need to upgrade to angular v13 to use the @cds/angular package.


packages/angular/projects/cds-angular/src/all.spec.ts
```diff
@@ -670,25 +670,25 @@ describe('CDS Modal Header', () => {
...
-    fixture = TestBed.createComponent(ModalTestComponent);
-    fixture.detectChanges();
-    instance = fixture.componentInstance;
-  });
-
-  it('cds-modal should be defined', () => {
-    expect(instance.vcCdsModalDirective).toBeDefined();
-  });
-});
+// describe('CDS Modal', () => {
```
hippee-lee, line 694: I removed this only to submit the PR to get some eyes on it. The test was failing but the component works fine in the core-angular-cli application. There is something odd with the focus-first.controller that seems to be causing the issue for the test. Need to confirm that the private observer in that class won't be an issue for angular.

packages/core/.eslintrc.js
```diff
@@ -17,7 +17,7 @@ const rules = {
   // cause slow analysis on TS files with Storybook, see
   // https://github.com/typescript-eslint/typescript-eslint/issues/1856
   '@typescript-eslint/no-use-before-define': OFF,
-  '@typescript-eslint/ban-types': [ERROR, { types: bannedTSTypes }],
+  // '@typescript-eslint/ban-types': [ERROR, { types: bannedTSTypes }],
```
hippee-lee: I think I meant to put this back but forgot. I'll double check or remove

packages/core/src/internal/controllers/first-focus.controller.ts
```diff
@@ -30,7 +30,7 @@ export class FirstFocusController {
       // usually handled in closable controller but here to force the issue if there is a problem there
       FocusTrapTrackerService.removeTrapElement({ focusTrapId: hostFocusTrapId });
     }
-    this.observer.disconnect();
+    this.observer?.disconnect();
```
hippee-lee: Added this to try and fix the failing @cds/angular test.

packages/core/src/internal/decorators/animate.ts
```diff
@@ -14,6 +14,7 @@ import { runPropertyAnimations } from '../motion/utils.js';
 
 // https://www.typescriptlang.org/docs/handbook/decorators.html#class-decorators
 export function animate(config: PropertyDrivenAnimation) {
+  // eslint-disable-next-line @typescript-eslint/ban-types
   return function _DecoratorName<T extends { new (...args: any[]): {} }>(constr: T) {
```
hippee-lee, line 18: There doesn't seem to be a way to type this properly and the updated eslint does not like the {} object as a type.

packages/core/src/internal/decorators/event.ts
```diff
@@ -24,7 +24,7 @@ export class EventEmitter<T> {
 }
 
 // Legacy TS Decorator
-function legacyEvent(descriptor: PropertyDescriptor, protoOrDescriptor: {}, name: PropertyKey) {
+function legacyEvent(descriptor: PropertyDescriptor, protoOrDescriptor: Record<string, unknown>, name: PropertyKey) {
```
hippee-lee, line 27: There are a bunch of these. `Record<string, unknown>` is a stand in type for `{}`

packages/core/src/internal/motion/utils.ts
```diff
@@ -355,12 +355,12 @@ export function sizeDimensionKeyframes(
   }
 
   return animationKeyframes.map((kf: Keyframe) => {
-    if (kf?.hasOwnProperty('height') && isPrefixedBy(kf?.height?.toString() || '', 'from:')) {
+    if (Object.prototype.hasOwnProperty.call(kf, 'height') && isPrefixedBy(kf?.height?.toString() || '', 'from:')) {
```
hippee-lee, line 358: Eslint did not like us calling hasOwnProperty, I forget the exact rule. This is the proper way to call methods on objects.

packages/eslint-plugin-clarity-adoption/src/rules/no-clr-alert/index.ts
```diff
@@ -12,7 +12,6 @@ export default createESLintRule({
     type: 'problem',
     docs: {
       description: 'Disallow use of clr-alert',
-      category: 'Best Practices',
```
hippee-lee, line 15: `meta.docs.category` has been removed from eslint the typescript-eslint package. See more [here](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v5.0.0).

packages/react/package.json
```diff
@@ -11,7 +11,7 @@
     "build:ts": "tsc --b ./tsconfig.project.json --force",
     "build:ts:watch": "tsc --b ./tsconfig.project.json -w",
     "core:build": "yarn --cwd ../core run build",
-    "lint": "eslint \"*/**/*.{ts,tsx}\" --ignore-pattern \"dist\"",
+    "lint": "eslint \"*/**/*.{ts,tsx}\" --ignore-pattern \"dist\" --resolve-plugins-relative-to ../../node_modules",
```
hippee-lee, line 14: I needed to add this to tell eslint where to find the plugins that are installed in the root node_modules.

packages/schematics/src/add/index.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('ng add @clr/angular', () => {
     }
   });
 
-  describe('in a new Angular workspace', async () => {
+  describe('in a new Angular workspace', () => {
```
hippee-lee, line 28: Asyncs are not allowed in describes here.

packages/schematics/src/utility/get-typescript-source-file.ts
```diff
@@ -20,7 +20,5 @@ export function getTypeScriptSourceFile(host: Tree, path: string): ts.SourceFile
   }
 
   const content = buffer.toString();
-  const sourceFile = ts.createSourceFile(path, content, ts.ScriptTarget.Latest, true);
-
-  return sourceFile;
+  return ts.createSourceFile(path, content, ts.ScriptTarget.Latest, true);
```
hippee-lee, line 23: Simplify the code.

packages/angular/projects/cds-angular/src/all.spec.ts
```diff
@@ -410,6 +411,26 @@ describe('CDS Control Action', () => {
   });
 });
 
+// describe('CDS Dropdown', () => {
```
hippee-lee, line 414: This adds a missing sanity check for @cds/angular wrappers.

hippee-lee:
I added a sanity check test for Dropdown in the cds-angular package. It demonstrates the same issue with testing. There is a build/tooling/webpack? issue that is compiling regular async/awaits into a weird yield generator and that is breaking the core code. 

I'm going to keep the two affected tests (one new:dropdown and one existing: modal) commented out because the fix is to identify why the angular compiler works for a standard angular app, but not the @cds/angular testing build.

mathisscott (approved):

packages/core/src/internal/controllers/first-focus.controller.ts
hippee-lee: removed.

apps/core-angular-cli/tsconfig.json
```diff
@@ -6,16 +6,18 @@
     "outDir": "./dist/out-tsc",
     "forceConsistentCasingInFileNames": true,
     "strict": true,
+    "noImplicitAny": false,
```
hippee-lee: The changes here mirror the tsconfig file for packages/angular tsconfig. I made them similar while trying to figure out why two tests were failing.

packages/angular/package.json
```diff
@@ -57,7 +57,7 @@
     "jasmine-spec-reporter": "~7.0.0",
     "karma": "~6.3.9",
     "karma-chrome-launcher": "~3.1.0",
-    "karma-coverage-istanbul-reporter": "~3.0.3",
+    "karma-coverage": "~2.1.0",
```
hippee-lee: @angular/cli@13 drops the `karma-coverage-istanbul-reporter` in favor of this one.

packages/angular/projects/cds-angular/karma.conf.js
```diff
@@ -9,12 +9,21 @@ module.exports = function (config) {
...
+      require('karma-coverage'),
       require('@angular-devkit/build-angular/plugins/karma'),
     ],
     client: {
+      jasmine: {
+        // you can add configuration options for Jasmine here
+        // the possible options are listed at https://jasmine.github.io/api/edge/Configuration.html
+        // for example, you can disable the random execution with `random: false`
+        // or set a specific seed with `seed: 4321`
+      },
```
hippee-lee: This was in the new karma.conf generated by @angular/cli@13 so I kept it.

packages/angular/projects/cds-angular/karma.conf.js
```diff
@@ -9,12 +9,21 @@ module.exports = function (config) {
...
       clearContext: false, // leave Jasmine Spec Runner output visible in browser
     },
+    jasmineHtmlReporter: {
+      suppressAll: true, // removes the duplicated traces
+    },
```
hippee-lee: This was in the new karma.conf generated by @angular/cli@13 so I kept it.

packages/angular/tsconfig.json
```diff
@@ -3,25 +3,27 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/out-tsc",
-    "noImplicitAny": false,
-    "noImplicitReturns": false,
-    "noImplicitThis": false,
+    "forceConsistentCasingInFileNames": true,
```
hippee-lee: Changes here are to update `tsconfig.json` to similar as generated by @angular/cli@13

bdryanovski (changes requested):
@hippee-lee Thanks for updating the packages, this pretty much covers 99% of my updated packages so we could go with these changes. Thanks again.

I have one thing that I want to bring as a conversation here before committing this into the `next` branch. 

We had a discussion a month ago with @coryrylan when we updated the `cds/angular` package that could be a good idea to put the outputted files as part of the repository - so we could track them and check them for changes overtime - they must not change that often after all.

And you made just that. But now when I'm looking at it - I think that I did not think about every possible case.

The issues that I see here is that this may create a list of strange changes over time - that could be misleading. The `cds/angular` package is built only if you run one of the tasks related to it `build`, `build:ci`, and so on. But making changes into `cds/core` that could trigger different output of the files won't trigger the rebuild. So these changes could be hidden somewhere else for example inside the BuildCI task or when releasing the packages - even when making unrelated changes and running build to verify something you may get some changes that you don't know from where are coming.

As a small but obvious side effect of that will be committed messages that no longer include changes for specific topics, but now files from unrelated changes.

I'm thinking that we need some way to make sure that this type of "at first random changes" are more controlled and won't depend on some "common team knowledge" on how to run tasks in a specific order.

There few things that I think we could do with their own pros and cons;
  * Having this task running before push could be an option. (only at git push)
  * Having a quicker way to determine are we gonna have a different output - not sure yet how - maybe if there is a change inside the `cds/core/*.element.ts` file we must automatically run `cds/angular` (husky could do that)
  * Not tracking the files (I don't like that)
  * Something else?
 
I know that this won't have an effect until you don't release the package to NPM, but still, this could be a really good indicator on is the `cds/angular` build as it should be. Right now it depends on jsdoc, so it is easily affected when even the doc attribute is on the wrong line. 


I'm not blocking this PR for any other reason than just discussing this before we continue.

packages/angular/projects/cds-angular/accordion/cds-accordion-panel.directive.ts
```diff
@@ -0,0 +1,38 @@
...
+  @Input() set expanded(value: boolean | '') {
+    this.element.expanded = !!value;
+  }
+
+  @Output() cdsMotionChange: EventEmitter<CustomEvent> = new EventEmitter();
+
+  constructor(elementRef: ElementRef) {
+    this.element = elementRef.nativeElement;
+  }
+}
```
coryrylan: Looking at the git ignore there were no changes to how the generated files are being ignored, so were these file unintentionally committed somehow?

hippee-lee: Yeah, these shouldn't be there. The !operator in the `.gitignore` doesn't seem to be working the way I'd expect it to.

I will go back and clean this up, this should not be committing the generated files to the repo. But in order to commit changes to the specs I have to comment out the `.gitignore` rule, make my changes and then re-enable the .gitignore rule. IMO - i do not think we should be committing generated files to the repo b/c the source and build configuration files (tsconfig.json) controls the generated output. And the generated output will be the same for each commit.

cc/ @bdryanovski - I think this covers your feedback to, correct?

bdryanovski: @hippee-lee - in that case - yes :) 

In some form or another, I think that having the generated files is a good idea - just because we could monitor the changes over time .. and have some indication that we are creating any change at all. In some cases, this could be even breaking change for example the case when we change a type or extend type in `cds/core` could have effect into `cds/angular` - right now we depend only on people to have this notion of linked changes at the back of there todo list.


But we could not implement this without a good way of committing and tracking changes without creating unwanted side-effects

Thanks, @coryrylan for noticing the `gitignore` file.

bdryanovski (approved):

packages/angular/projects/cds-angular/accordion/cds-accordion-panel.directive.ts
hippee-lee: All of the generated files and folders have been removed.

coryrylan (approved):

PR #6502, opened 2021-12-07
feat(core:grid): dynamic i18n functionality

mathisscott:

## PR Checklist

## PR Type

This PR introduces a beta for "dynamic" i18n strings and fixes a couple of issues with the legacy i18n code that had crept in over the last couple of versions.

### Dynamic i18n

The way these work is you put in a template token in an i18n string. For example, creating an i18n value that says something like... `{ "pageOf": "Page ${currentPage} of ${totalPages}" }` would tell the i18n service to look at the component to pull values for `component.currentPage` and `component.totalPages`. This means the final output would render as `Page 2 of 49` if those values were `2` and `49` in the component instance when the component is rendered.

#### Do these update with the component?

Yes.

### Fixes

A couple of fixes were handled in this PR as well.

#### 1. Updating from the `cds-i18n` HTML attribute

Somehow this got broken. It now works.

#### 2. Properly gathering overrides with keys

This also used to work, afaik. But when I took this on, I discovered that trying to override, say, `{ "loading": "Busy" }` would hard set the i18n on a component to just the override and blow away all the rest of the keys. If your component's i18n keys were `{ success: "Success", error: "Error", loading: "Loading" }` then customizing the loading i18n string would kill success and error.

Previously there were no tests in the i18n decorator to guard against this breaking. I've added those as well.

### Points for Improvement/Future Work

The following areas are all areas that fell outside the scope of this effort (which was driven more by an a11y need).

1. Implementing dynamic aria labels in the progress circle. We held off a bit on dynamic updates on the progress circle aria-label until this was in. We can do that now.
2. Forcing i18n updates from the parent down. Presently, if I override an i18n string at the topmost level (say `cds-alert-group`), it only propagates if it is forced to (similar to how we've used `syncProps` in the past). This is a present issue with i18n that is resolved in some components but not all. More work to be done here.
3. Documentation. This is a beta easter egg right now. When we feel comfortable with it, it would be good to make it a first class citizen in our i18n docs.

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See above.

Issue Number: N/A

## What is the new behavior?

See above.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/grid/docs/async-data.stories.ts
+++ b/packages/core/src/grid/docs/async-data.stories.ts
@@ -1,6 +1,8 @@
 import { html, LitElement, PropertyValues } from 'lit';
+import { query } from 'lit/decorators/query.js';
 import { state, customElement } from '@cds/core/internal';
 import { DemoGrid, DemoService } from '@cds/core/demo';
+import { CdsGrid } from '@cds/core/grid';
 
 export default {
   title: 'Stories/Grid',
@@ -11,6 +13,8 @@ export function asyncData() {
   @customElement('demo-grid-async-data')
   class DemoGridAsyncData extends LitElement {
     @state() private grid: DemoGrid = { label: '', rowActions: [], columns: [], rows: [] } as unknown as DemoGrid;
+
+    @query('cds-grid') private gridElement: CdsGrid;
 
     render() {
       return html`
@@ -39,6 +43,7 @@ export function asyncData() {
     private async load() {
       this.grid = { label: '', rowActions: [], columns: [], rows: [] } as unknown as DemoGrid;
       this.grid = (await DemoService.asyncData).grid;
+      this.gridElement.requestUpdate(); // <- we need this so the i18n hydrates itself after load
     }
   }
 
--- a/packages/core/src/grid/docs/cell-editable.stories.ts
+++ b/packages/core/src/grid/docs/cell-editable.stories.ts
@@ -1,4 +1,4 @@
-import { css, html, LitElement } from 'lit';
+import { html, LitElement } from 'lit';
 import { queryAll } from 'lit/decorators.js';
 import { customElement, getInputValueType, state } from '@cds/core/internal';
 import { DemoGridCell, DemoService, exportElementsToCSV, parseCSV } from '@cds/core/demo';
--- a/packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
+++ b/packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
@@ -1,5 +1,5 @@
 import { html, LitElement } from 'lit';
-import { customElement, firstFocus, focusTrap, InlineFocusTrapController } from '@cds/core/internal';
+import { customElement, firstFocus, focusTrap } from '@cds/core/internal';
 import { componentIsStable, createTestElement, removeTestElement } from '@cds/core/test';
 
 @focusTrap<InlineFocusTrapControllerTestElement>()
@@ -20,7 +20,7 @@ class InlineFocusTrapControllerTestElement extends LitElement {
 describe('inline-focus-trap.controller', () => {
   let component: HTMLElement;
   let element: HTMLElement;
-  let root: any;
+  // let root: HTMLDocument | ShadowRoot;
   let shadowRoot: ShadowRoot;
 
   beforeEach(async () => {
@@ -36,7 +36,7 @@ describe('inline-focus-trap.controller', () => {
     component = element.querySelector<InlineFocusTrapControllerTestElement>(
       'inline-focus-trap-controller-test-element'
     );
-    root = component.getRootNode() as any;
+    // root = component.getRootNode() as any;
     shadowRoot = component.shadowRoot as any;
   });
 
--- a/packages/core/src/internal/decorators/i18n.spec.ts
+++ b/packages/core/src/internal/decorators/i18n.spec.ts
@@ -4,10 +4,17 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+import {
+  customElement,
+  i18n,
+  getI18nValues,
+  getI18nUpdateStrategy,
+  I18nService,
+  I18nElement,
+  LogService,
+} from '@cds/core/internal';
 import { html, LitElement } from 'lit';
-import { customElement } from 'lit/decorators/custom-element.js';
 import { componentIsStable, createTestElement, removeTestElement } from '@cds/core/test';
-import { i18n } from './i18n.js';
 
 const i18nValues = {
   open: 'Open my element',
@@ -21,6 +28,18 @@ class TestI18nElement extends LitElement {
 
   render() {
     return html`<slot></slot>`;
+  }
+}
+
+/** @element test-alert-18n-element */
+@customElement('test-alert-18n-element')
+class TestAlertI18nElement extends LitElement {
+  @i18n() i18n: Record<string, any> = I18nService.keys.alert;
+
+  greeting = 'hello';
+
+  render() {
+    return html`<p>ohai</p>`;
   }
 }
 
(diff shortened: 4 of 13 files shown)


packages/core/src/grid/docs/async-data.stories.ts
```diff
@@ -39,6 +43,7 @@ export function asyncData() {
     private async load() {
       this.grid = { label: '', rowActions: [], columns: [], rows: [] } as unknown as DemoGrid;
       this.grid = (await DemoService.asyncData).grid;
+      this.gridElement.requestUpdate(); // <- we need this so the i18n hydrates itself after load
```
mathisscott, line 46: The async functionality was rendering the i18n strings before it had values to put in them. Forcing an update here helped solve that. Leaving in as a good practice/example.

packages/core/src/grid/docs/cell-editable.stories.ts
```diff
@@ -1,4 +1,4 @@
-import { css, html, LitElement } from 'lit';
```
mathisscott, line 1: There are a lot of little build fixes like this that TS/ESLint was complaining about.

packages/core/src/internal/decorators/i18n.spec.ts
```diff
@@ -24,6 +30,18 @@ class TestI18nElement extends LitElement {
   }
 }
 
+/** @element test-alert-18n-element */
+@customElement('test-alert-18n-element')
+class TestAlertI18nElement extends LitElement {
```
mathisscott, line 36: I recreated the situation of @coryrylan 's stackblitz of the i18n not properly merging overrides with the i18n keys.

packages/core/src/internal/decorators/i18n.spec.ts
```diff
@@ -52,3 +70,122 @@ describe('i18n decorator', () => {
...
+    expect(testMe.loading).toBe('Loading');
+    expect(testMe.success).toBe('Success');
+  });
+
+  it('overrides from cds-i18n attr as expected, even partially', async () => {
+    component.setAttribute('cds-i18n', '{ "closeButtonAriaLabel": "${greeting} world" }');
+    await componentIsStable(component);
+    const testMe = component.i18n;
+    expect(testMe.closeButtonAriaLabel).toBe('hello world');
+    expect(testMe.loading).toBe('Loading');
```
mathisscott, line 100: Here are those tests. Note that if the override blows away the underlying keys, `testMe.loading` will come back as `undefined`.

packages/core/src/internal/decorators/i18n.spec.ts
```diff
@@ -52,3 +70,122 @@ describe('i18n decorator', () => {
...
+    expect(testMe.success).toBe('Success');
+  });
+
+  it('overrides from i18n as expected, even partially', async () => {
+    component.i18n = { closeButtonAriaLabel: '${greeting} from the other side...' };
```
mathisscott, line 105: I tested both updated the attr and property. This may seem like overkill but the decorator was behaving a little erratically on this matter.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -78,13 +81,24 @@ export function i18n() {
 
     const descriptor = {
       get(this: I18nElement) {
-        return { ...(I18nService.keys as any)[this.__i18nKey], ...this.__i18n };
+        const i18nObj = mergeObjects((I18nService.keys as any)[this.__i18nKey], this.__i18n || {});
+        return I18nService.hydrate(i18nObj, this);
```
mathisscott, line 86: This is where the i18n strings get values from the component instance.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -78,13 +81,24 @@ export function i18n() {
 
     const descriptor = {
       get(this: I18nElement) {
-        return { ...(I18nService.keys as any)[this.__i18nKey], ...this.__i18n };
+        const i18nObj = mergeObjects((I18nService.keys as any)[this.__i18nKey], this.__i18n || {});
```
mathisscott, line 85: The spread operators only do a shallow merge. We have i18n objects now that go more than one level deep. `mergeObjects` does that.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -78,13 +81,24 @@ export function i18n() {
...
+        return I18nService.hydrate(i18nObj, this);
       },
-      set(this: I18nElement, value: {}) {
-        (this.__i18nKey as any) = Object.keys(I18nService.keys).find(key => (I18nService.keys as any)[key] === value);
+      set(this: I18nElement, value: Record<string, unknown>) {
+        const newValues = getI18nValues(value, this);
+        const testKey = I18nService.findKey(newValues);
+        const strat = getI18nUpdateStrategy(testKey || '', this.__i18nKey, newValues, this.__i18n);
+
+        if (typeof strat.key !== 'undefined') {
```
mathisscott, line 93: I used a strategy pattern here to make this whole block easier to test.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -96,3 +110,45 @@ export function i18n() {
       : standardI18n(descriptor, protoOrDescriptor);
   };
 }
+
+export function getI18nValues(values: Record<string, unknown>, component: I18nElement): Record<string, unknown> {
+  if (isNilOrEmpty(values)) {
+    let returnVal = {};
+
+    if (component.hasAttribute('cds-i18n')) {
```
mathisscott, line 119: This block here is a sledgehammer to force a lookup on the HTML attr in the event the i18n property gets in a bad state. It does no harm but handles a weird edge case I was hitting early on.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -96,3 +110,45 @@ export function i18n() {
...
+      const stringVal = component.getAttribute('cds-i18n') + '';
+      if (isNilOrEmpty(stringVal)) {
+        returnVal = {};
+      } else {
+        try {
+          returnVal = JSON.parse(stringVal);
+        } catch {
+          console.warn('Clarity i18n: Invalid JSON passed to cds-i18n');
```
mathisscott: The i18n needs to be valid JSON. That's probably the hardest thing to understand about it.

packages/core/src/internal/decorators/i18n.ts
```diff
@@ -96,3 +110,45 @@ export function i18n() {
...
+  }
+}
+
+export function getI18nUpdateStrategy(
+  newKey: string,
+  oldKey: string,
+  newValues: Record<string, unknown>,
+  oldValues: Record<string, unknown>
+): { values?: object; key?: string; update: boolean } {
+  if (!isNilOrEmpty(newKey)) {
```
mathisscott, line 144: Here's that strategy pattern. You send it your data. It returns with an object that tells you how the i18n should update/change.

packages/core/src/internal/services/i18n.service.spec.ts
```diff
@@ -24,10 +28,109 @@ describe('I18nService', () => {
...
+  });
+
+  describe('findKey()', () => {
+    it('finds key if object matches a keyed value', () => {
+      const findMe = I18nService.get('alert');
+      const testMe = I18nService.findKey(findMe);
+      expect(testMe).toBe('alert');
+    });
+
+    it('does what?', () => {
```
mathisscott: The title here is a leftover. Funny though. I will update it.

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -183,17 +184,42 @@ export class I18nService {
   static get keys(): Readonly<I18nStrings> {
     // intialize registry if not yet set
     if (Object.keys(GlobalStateService.state.i18nRegistry).length === 0) {
-      GlobalStateService.state.i18nRegistry = { ...componentStringsDefault, ...GlobalStateService.state.i18nRegistry };
+      GlobalStateService.state.i18nRegistry = mergeObjects(
```
mathisscott, line 187: Same problem with deep vs. shallow merges here.

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -183,17 +184,42 @@ export class I18nService {
...
+        GlobalStateService.state.i18nRegistry
+      );
+    }
+
+    return mergeObjects(componentStringsDefault, GlobalStateService.state.i18nRegistry) as I18nStrings;
+  }
+
+  static findKey(keyValueObj: any) {
+    const keyVals = I18nService.keys;
+    return Object.keys(keyVals).find(key => objectNaiveDeepEquals((keyVals as any)[key], keyValueObj));
```
mathisscott, line 198: Previously we had a `===` here. That was failing. This was the root cause of the overrides overwriting the i18n keys. Fixed now.

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -183,17 +184,42 @@ export class I18nService {
...
+  static get(key: string) {
+    if (!key) {
+      return {};
     }
 
-    return { ...componentStringsDefault, ...GlobalStateService.state.i18nRegistry };
+    return (I18nService.keys as any)[key] || {};
+  }
+
+  static reset() {
```
mathisscott, line 210: This is used primarily in the unit tests. Other folks could use it if they wanted. 🤷‍♂️

packages/core/src/internal/services/i18n.service.ts
```diff
@@ -183,17 +184,42 @@ export class I18nService {
...
+    // largely useful for unit testing
+    GlobalStateService.state.i18nRegistry = mergeObjects({}, componentStringsDefault);
+  }
+
+  static hydrate(i18nObject: object, componentInstance: any) {
+    return JSON.parse(interpolateNaively(JSON.stringify(i18nObject), componentInstance));
```
mathisscott, line 216: Surprisingly, getting component values into the i18n strings was not the hardest part of this...

packages/core/src/internal/utils/identity.ts
```diff
@@ -65,6 +65,31 @@ export function createId(prefix = '_') {
   return `${prefix}${Math.random().toString(36).substr(2, 9)}`;
 }
 
+// simplistic way to test objects for equality
+// note that it ignores/removes methods from objects
+export function objectNaiveDeepEquals(obj1: {}, obj2: {}) {
```
mathisscott, line 70: Note comment above. It's naive because it ignores function/method equality.

packages/core/src/internal/utils/identity.ts
```diff
@@ -65,6 +65,31 @@ export function createId(prefix = '_') {
...
+  return JSON.stringify(obj1) === JSON.stringify(obj2);
+}
+
+export function getFromObjectPath(path: string, dataObj: any, fallback = `$\{${path}}`) {
```
mathisscott, line 74: `getFromObjectPath('a.a1', { a: { a1: 'ohai', a2: 'howdy }, b: 'zdrasti' })` will return `"ohai"`. If you tried to find `getFromObjectPath('a.a3', { a: { a1: 'ohai', a2: 'howdy }, b: 'zdrasti' })` it would return `"${a.a3}"` as a string because there is no `a3`.

packages/core/src/internal/utils/identity.ts
```diff
@@ -250,3 +282,27 @@ export function allAreDefined<U>(...items: U[]): boolean {
...
+  });
+
+  clones.forEach(srcObj => {
+    Object.keys(srcObj).forEach(prop => {
+      const propVal = (srcObj as any)[prop];
+
+      if (Array.isArray(propVal)) {
+        (returnObj as any)[prop] = Array.from(propVal);
+      } else if (isObject(propVal)) {
+        (returnObj as any)[prop] = mergeObjects((returnObj as any)[prop] || {}, propVal);
```
mathisscott, line 300: recursion fun!

packages/core/src/internal/utils/string.ts
```diff
@@ -161,3 +161,12 @@ export function pluckValueFromStringUnit(val: string, unit: string) {
 export function pluckPixelValue(val: string): number {
   return !val ? 0 : pluckValueFromStringUnit(val.trim(), 'px');
 }
+
+export function interpolateNaively(template: string, dataObj: any, fallback?: string) {
```
mathisscott, line 165: "naive" because it will just say `Function ()` if you try to get it to interpolate a function.

packages/core/src/grid/docs/async-data.stories.ts
```diff
@@ -1,6 +1,8 @@
 import { html, LitElement, PropertyValues } from 'lit';
+import { query } from 'lit/decorators/query.js';
 import { state, customElement } from '@cds/core/internal';
 import { DemoGrid, DemoService } from '@cds/core/demo';
+import { CdsGrid } from '../grid/grid.element.js';
```
coryrylan: Let's pull this import in through `@cds/core/grid`

packages/core/src/internal/services/i18n.service.ts
coryrylan: We may wan't to mark this one with a `@private`. I'd be worried since this is global state. In microfrontend situation, apps could unintentionally reset the i18n values and wipe out values set by other parts of the application they don't control.

coryrylan (approved):

packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('inline-focus-trap.controller', () => {
     component = element.querySelector<InlineFocusTrapControllerTestElement>(
       'inline-focus-trap-controller-test-element'
     );
-    root = component.getRootNode() as any;
+    // root = component.getRootNode() as any;
```
hippee-lee, line 39: Extra comment?

packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
```diff
@@ -20,7 +20,7 @@ class InlineFocusTrapControllerTestElement extends LitElement {
 describe('inline-focus-trap.controller', () => {
   let component: HTMLElement;
   let element: HTMLElement;
-  let root: any;
+  // let root: any;
```
hippee-lee: Extra comment? I know it is just a test but I wonder if this could be typed?

```javascript
let root: HTMLDocument | ShadowRoot; 
```

https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode#return_value

hippee-lee (approved):

packages/core/src/internal/decorators/i18n.ts
mathisscott: I should use the LogService here and test against that...

packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
mathisscott: Nope. This was commented out as part of the tests @coryrylan commented out below. TS was complaining because nothing was using this variable...

packages/core/src/internal/controllers/inline-focus-trap.controller.spec.ts
mathisscott: Same here. Commented out to make TS happy after the tests were commented out further down. I will update the typing for when it gets uncommented in the future though...

PR #6505, opened 2021-12-07
feat(core): action button

coryrylan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?

https://61ba7ce549e858a779684d11--vmware-clarity.netlify.app/storybook/core/index.html?path=/story/components-button-action--page

- Action buttons are used internally and externally as part of our public API. They are used heavily in the datagrid api. This PR adds action button types as part of the upcoming datagrid work as well as refactors similar custom isolated icon buttons in the system. (form control actions, internal close buttons, accordion, treeview) 

- Action button ensures consistency visually and for a11y with all icon actions in the system. This includes touch targets and interaction states

- This also refactors the base button class into isolated lit controller behaviors. This allows us to isolate component behaviors in a way that is composable and reusable in other components similar to mixins. Some of these controllers also are used in datagrid work that will be in followup PRs.

- The controllers do have internal docs/demos in storybook but I plan to add those as a stand alone PR to keep this one manageable. 

- [ ] Yes
- [x] No

## Other information

Screen Shot 2021-12-07 at 4 31 16 PM


packages/core/src/alert/alert.element.ts
```diff
@@ -251,14 +251,13 @@ export class CdsAlert extends LitElement {
...
-                  aria-label="${this.i18n.closeButtonAriaLabel}"
-                ></cds-internal-close-button></slot
-            ></span>`
+          ? html`<slot name="close-button">
+              <cds-internal-close-button
+                icon-size="${this.type === 'banner' ? '20' : '16'}"
+                @click="${() => this.closeAlert()}"
+                aria-label="${this.i18n.closeButtonAriaLabel}"
+              ></cds-internal-close-button
+            ></slot>`
```
coryrylan, line 260: The span was no longer needed with the slight space adjustment in the action button

packages/core/src/button/button.element.spec.ts
```diff
@@ -174,23 +174,6 @@ describe('button element', () => {
...
-      expect(component.removeEventListener).toHaveBeenCalledWith('click', jasmine.any(Function));
-      expect(component.removeEventListener).toHaveBeenCalledWith('keydown', jasmine.any(Function));
-
-      spyOn(component, 'addEventListener').and.callThrough();
-      component.readonly = false;
-      await componentIsStable(component);
-      expect(component.addEventListener).toHaveBeenCalledWith('click', jasmine.any(Function));
-      expect(component.addEventListener).toHaveBeenCalledWith('keydown', jasmine.any(Function));
-    });
-
```
coryrylan, line 193: All managed and tested by the controllers

packages/core/src/button/button.element.spec.ts
```diff
@@ -372,41 +355,3 @@ describe('buttonSlots: ', () => {
...
-    const component = element.querySelector('cds-button');
-    await componentIsStable(component);
-    expect(component.hasAttribute('_active')).toBe(false);
-
-    component.dispatchEvent(new MouseEvent('mousedown'));
-    await componentIsStable(component);
-    expect(component.hasAttribute('_active')).toBe(false);
-    removeTestElement(element);
-  });
-});
```
coryrylan, line 412: Moved and tested in isolated controllers

packages/core/src/button/inline-button.element.ts
```diff
@@ -29,6 +29,8 @@ import styles from './inline-button.element.scss';
  * @cssprop --letter-spacing
  */
 export class CdsInlineButton extends CdsBaseButton {
+  @querySlot('cds-icon') protected icon: HTMLElement;
```
coryrylan, line 32: Icon is no longer part of the base button class as only a few actually used it and created false api docs

packages/core/src/forms/control-action/control-action.element.spec.ts
```diff
@@ -31,21 +30,11 @@ describe('cds-control-action', () => {
...
-    await componentIsStable(controlAction);
-    expect(LogService.warn).toHaveBeenCalled();
-  });
-
   describe('syncAria: ', () => {
     it('should set aria-hidden on read-only', async () => {
       const testElement = await createTestElement(html` <cds-control-action readonly>test</cds-control-action>`);
       const testControlAction = testElement.querySelector<CdsControlAction>('cds-control-action');
-      expect(testControlAction.getAttribute('aria-hidden')).toBe('true');
+      expect(testControlAction.ariaHidden).toBe('true');
```
coryrylan, line 37: aria changes managed/tested at the controller level

packages/core/src/forms/control-action/control-action.element.ts
```diff
@@ -4,89 +4,31 @@
...
-    super.connectedCallback();
-    this.syncAria();
-  }
-
-  private syncAria() {
-    const iAmReadonly = this.readonly;
-    const iHaveAriaLabel = hasAttributeAndIsNotEmpty(this, 'aria-label');
-    setOrRemoveAttribute(this, ['aria-hidden', 'true'], () => {
-      return iAmReadonly && !iHaveAriaLabel;
-    });
```
coryrylan, line 68: all logic is shared within action button and base button

packages/core/src/icon/collections/core.ts
```diff
@@ -158,5 +162,7 @@ declare module '@cds/core/internal' {
     [vmBugIconName]: string;
     [vmBugInverseIconName]: string;
     [warningStandardIconName]: string;
+    [detailExpandIconName]: string;
+    [detailCollapseIconName]: string;
```
coryrylan, line 166: new icons for the detail expand states that are used in the datagrid and action expand button

packages/core/src/input/input.element.spec.ts
```diff
@@ -98,7 +98,7 @@ describe('cds-control with prefix/suffix icon', () => {
   it('should apply padding inline styles when suffix/prefix controls are applied in input', async () => {
     await componentIsStable(control);
     expect(input.getAttribute('style')).toBe(
-      'padding-left: calc((30 / var(--cds-global-base)) * 1rem) !important; padding-right: calc((30 / var(--cds-global-base)) * 1rem) !important;'
+      'padding-left: calc((22 / var(--cds-global-base)) * 1rem) !important; padding-right: calc((22 / var(--cds-global-base)) * 1rem) !important;'
```
coryrylan, line 101: slight visual adjustment due to action button

packages/core/src/internal-components/close-button/close-button.element.spec.ts
```diff
@@ -31,34 +30,10 @@ describe('internal close button element', () => {
...
-    const myIcon = component.shadowRoot.querySelector<HTMLElement>('cds-icon');
-    expect(component.iconShape).toBe('times', 'default set');
-    expect(myIcon.getAttribute('shape')).toBe('times', 'shape set');
-  });
-
-  it('should allow me to change icon shape', async () => {
-    const myIcon = component.shadowRoot.querySelector<HTMLElement>('cds-icon');
-    component.iconShape = 'jabberwocky';
-    await componentIsStable(component);
-    expect(myIcon.getAttribute('shape')).toBe('jabberwocky', 'shape set');
```
coryrylan, line 61: all managed at action button level

packages/core/src/internal/base/button.base.ts
```diff
@@ -27,189 +37,5 @@ export class CdsBaseButton extends LitElement {
...
-  // Forms handle default submit buttons without type="submit" automatically.
-  // Because of the dynamic createion of the native element, we prefer to set the attribute explicitly.
-  private addDefaultSubmitBehavior() {
-    if (!this.type && this.closest('form')) {
-      this.type = 'submit';
-    }
-    if (this.type === 'submit' && !this.markerAttached) {
-      this.appendChild(this.marker);
-    }
-  }
```
coryrylan, line 214: all logic is split into the controllers defined above 

```typescript
@active<CdsBaseButton>()
@ariaButton<CdsBaseButton>()
@ariaPressed<CdsBaseButton>()
@ariaDisabled<CdsBaseButton>()
@ariaPopupTrigger<CdsBaseButton>()
@buttonSubmit<CdsBaseButton>()
@buttonAnchor<CdsBaseButton>()
```

This makes the logic generic and reusable for other components that may need similar a11y behaviors but are not a traditional button. This also helps ensure our APIs are consistent with how components state is described (disabled, expanded, pressed, readonly...)

packages/core/src/internal/controllers/aria-disabled.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  hostUpdated() {
+    if (this.host.disabled !== null) {
+      this.host.ariaDisabled = this.host.disabled;
+    }
+
+    if (this.host.readonly) {
+      this.host.ariaDisabled = null;
+    }
```
coryrylan, line 24: When a host component is `readonly` typically this means the component is content and not an interactive element. For example readonly tags display information but are not clickable. In these cases we do not want to describe any aria states and treat it as if it was static content.

packages/core/src/internal/controllers/aria-popup-trigger.controller.spec.ts
```diff
@@ -33,16 +31,16 @@ describe('aria-popup-trigger.controller', () => {
...
 
   it('should NOT initialize aria attributes for triggers without a popup id', async () => {
     await componentIsStable(noponent);
     expect(noponent.hasAttribute('popup')).toBe(false);
-    expect(noponent.hasAttribute('aria-controls')).toBe(false);
-    expect(noponent.hasAttribute('aria-haspopup')).toBe(false);
-    expect(noponent.hasAttribute('aria-expanded')).toBe(false);
+    expect(noponent.ariaControls).toBe(undefined);
+    expect(noponent.ariaHasPopup).toBe(null);
+    expect(noponent.ariaExpanded).toBe(null);
```
coryrylan, line 44: We can now use aria properties directly.  In some cases like datagrid/treeview this seems to help performance as it enables the property to be set in AOM and let the browser manage updating the DOM

packages/core/src/internal/controllers/aria-pressed.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  hostUpdated() {
+    if (this.host.pressed !== null && this.host.pressed !== undefined) {
+      this.host.ariaPressed = this.host.pressed ? 'true' : 'false';
+    }
+
+    if (this.host.readonly) {
+      this.host.ariaPressed = null;
+    }
+  }
```
coryrylan, line 26: similar to the disable controller, in readonly states we want to treat the component as neutral content

packages/core/src/actions/action-expand.element.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should create the component', async () => {
+    await componentIsStable(component);
+    expect(component).toBeTruthy();
+  });
+
+  it('should set default angle icon', async () => {
+    await componentIsStable(component);
+    expect(component.shadowRoot.querySelector('cds-icon').shape).toBe('angle');
```
mathisscott: Do we want to test if the icon shape can be changed?

packages/core/src/actions/action-expand.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    } else if (this.action === 'horizontal') {
+      return this.pressed ? 'left' : 'right';
+    } else {
+      return null;
+    }
+  }
+
+  private get iconShape() {
+    if (this.action === 'detail') {
+      return this.pressed ? 'detail-collapse' : 'detail-expand';
```
mathisscott: It looks like we are hard-coding the icon shapes in actions. We don't do that in other areas that use icons. Why did we decide to do that here?

packages/core/src/actions/action-sort.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.addEventListener('click', () => this.sortClick());
+
+    if (this.ariaLabel) {
+      this.ariaLabel = this.i18n.sort;
+    }
+  }
+
+  private sortClick() {
+    let sort = this.sort;
+    switch (sort) {
```
mathisscott: Because this is private we have to test it by setting up the component and forcing events and whatnot. Would it be easier to break this switch statement into a helper so we could test it in isolation and then just test whether the event fires or not (versus granularly testing that the event fires results in a specific order?). The event-based tests seem brittle.

packages/core/src/card/card.element.scss
```diff
@@ -1,8 +1,6 @@
 @import './../styles/tokens/generated/index';
 
 :host {
-  --width: fit-content;
```
mathisscott: This seems out of scope with actions and buttons. What's happening here?

packages/core/src/internal/controllers/button-anchor.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export class ButtonAnchorController<T extends ButtonAnchor> implements ReactiveController {
+  constructor(private host: T) {
+    this.host.addController(this);
+  }
+
+  hostUpdated() {
+    const hasAnchor = this.host.parentElement?.tagName === 'A';
+    if (hasAnchor && this.host.parentElement) {
```
mathisscott, line 31: Wouldn't the `?.` part of `this.host.parentElement?.tagName` on the previous line handle the `this.host.parentElement` check for us?

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -45,58 +45,46 @@ import styles from './pagination-button.element.scss';
...
-
-  @querySlot('cds-icon', { assign: 'cds-icon-slot' }) cdsIcon: HTMLElement;
+  @property({ type: String }) action: 'first' | 'prev' | 'next' | 'last';
 
-  @query('cds-icon') private cdsIconDefault: HTMLElement;
+  @state() private direction: 'up' | 'right' | 'down' | 'left';
 
-  private get customContent() {
-    return !this.action && !this.cdsIconDefault;
+  get cdsIcon() {
```
mathisscott: Why wouldn't a `@query()` property work here?

packages/core/src/pagination/pagination-button.element.ts
```diff
@@ -45,58 +45,46 @@ import styles from './pagination-button.element.scss';
...
-    return [baseStyles, styles];
-  }
+  updated(props: PropertyValues) {
+    super.updated(props);
 
-  firstUpdated(props: Map<string, any>) {
-    super.firstUpdated(props);
-    this.validateAriaLabel();
-  }
+    switch (this.action) {
```
mathisscott: This seems like another good candidate for a helper function...

mathisscott (changes requested):

packages/core/src/actions/register.ts
```diff
@@ -0,0 +1,36 @@
...
+import { CdsAction } from './action.element.js';
+import { CdsActionSort } from './action-sort.element.js';
+import { CdsActionExpand } from './action-expand.element.js';
+import { CdsActionHandle } from './action-handle.element.js';
+import '@cds/core/icon/register.js';
+
+ClarityIcons.addIcons(angleIcon, dragHandleIcon, ellipsisVerticalIcon, detailExpandIcon, detailCollapseIcon, timesIcon);
+ClarityIcons.addAliases(['times', ['close']]);
+
+registerElementSafely('cds-action', CdsAction);
```
hippee-lee: Question about optimization: Since we only have one register file does this mean that if a consumer only uses one, say cds-action, won't they also get the others? For tree shaking shouldn't each element (in this case) have its own registration code?

packages/core/src/actions/action-expand.element.ts
```diff
@@ -0,0 +1,71 @@
...
+    } else {
+      return 'angle';
+    }
+  }
+
+  render() {
+    return html`
+      <div class="private-host">
+        <cds-icon .shape=${this.iconShape} .direction=${this.iconDirection}></cds-icon>
```
hippee-lee: Note, this is not customizable by consumers.

packages/core/src/actions/action-handle.element.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export class CdsActionHandle extends CdsAction {
+  static get styles() {
+    return [super.styles, styles];
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.pressed = false;
+    this.shape = 'drag-handle';
```
hippee-lee: Does this need a render fn to make the icons customizable?

packages/core/src/actions/action-sort.element.ts
```diff
@@ -0,0 +1,71 @@
...
+  @event() sortChange: EventEmitter<ActionSort>;
+
+  static get styles() {
+    return [super.styles, styles];
+  }
+
+  render() {
+    return html`
+      <div class="private-host">
+        <div cds-layout="vertical align:center">
```
hippee-lee: Is this, should this be customizable by users. I dont think so but wanted to ask the question.

hippee-lee (commented):
Just the register questions and icon customization questions for this pass. Other than that, it lgtm.

hippee-lee (approved):
Pre-approval so I don't block this.

packages/core/src/actions/action-expand.element.ts
coryrylan: Will fix with default slot

packages/core/src/actions/action-handle.element.ts
coryrylan: Will fix with default slot

packages/core/src/actions/action-sort.element.ts
coryrylan: Will fix with default slots

packages/core/src/actions/register.ts
coryrylan: Yeah thats right, I can split this up into multiple registers. I was kind of on the fence with it since they are all small/similar. However some could get a bit bigger like the resize/draggable actions.

```typescript
import '@cds/core/action/register.js';
import '@cds/core/action-sort/register.js';
import '@cds/core/action-expand/register.js';
import '@cds/core/action-handle/register.js';
import '@cds/core/action-resize/register.js';
```

packages/core/src/internal/controllers/button-anchor.controller.ts
coryrylan: Yeah I thought so as well, but the `parentElement` property on `HTMLElement` technically can be null on elements that were created but not yet appended to the DOM. So this means a required null check by TS.

ashleyryan (commented):
How is an Action Button different then a CdsIconButton using the `flat` action? I know that combination isn't documented in storybook, but I've used it a few times at CloudHealth

coryrylan:
@ashleyryan good callout! There are a couple of differences between the action and button components but potential opportunity to consolidate as well. The action button consolidates all the custom icon style buttons internally in the system like close, expand, and various control/form actions.

The `cds-button` has a lot of option/types that we would want to keep separate (for now). This includes visual options like statuses, sizes and loading. But primarily they have specific aria/a11y states like sorting, pressed/drag and expanded. If possible we try to avoid ["impossible states"](https://kentcdodds.com/blog/make-impossible-states-impossible) in the component APIs specifically with this case the a11y differences/complexities. I'm also hoping to avoid any "mega" components even if that means multiple derivative types.

However you got me thinking about a possible way to rename on the actions. I think I can rename the actions to better align with existing cds-button-* conventions and consolidate the API a bit with no breaking changes. I'll post back here and slack with what I find. Thanks!

packages/core/src/actions/action-sort.element.ts
coryrylan: Fixed, split out into util function

packages/core/src/actions/action-expand.element.ts
coryrylan: Added default slots for all action buttons so they are always overridable.

packages/core/src/actions/action-expand.element.spec.ts
coryrylan: Added

packages/core/src/pagination/pagination-button.element.ts
coryrylan: Updated to use query

packages/core/src/pagination/pagination-button.element.ts
coryrylan: Split out into helper function

packages/core/src/actions/register.ts
coryrylan: Updated

packages/core/src/actions/action-expand.element.ts
coryrylan: Added default slots for all action buttons so they are always overridable.

packages/core/src/actions/action-handle.element.ts
coryrylan: Added default slots for all action buttons so they are always overridable.

packages/core/src/actions/action-sort.element.ts
coryrylan: Added default slots for all action buttons so they are always overridable.

packages/core/src/forms/control-action/control-action.element.ts
```diff
@@ -4,89 +4,31 @@
...
+import { assignSlotNames, property } from '@cds/core/internal';
+import { CdsButtonAction } from '@cds/core/button-action';
 
 /**
- * Control Action
+ * Control Action Button
  *
  * ```typescript
- * import '@cds/core/forms/register.js';
+ * import '@cds/core/button/register.js';
```
mathisscott: Looking at the code, I'm not sure where control-action needs to be imported from but I'm 100% sure it's not being exported from `button/register.ts`...

packages/core/src/forms/control/control.element.scss
```diff
@@ -71,7 +71,7 @@ $fallback-width: calc(#{$cds-global-layout-space-xxl} * 2);
   cursor: default;
 }
 
-cds-control-action.status {
+[cds-button-action].status {
```
mathisscott, line 74: `cds-control-action` vs. `cds-button-action` is confusing to me now. I don't see us registering a `cds-control-action` element. And all references to that element are being changed to `cds-button-action`.

Yet we still have `control-action.element.ts`. Was our intent to get rid of `control-action.element.ts`? Is this a deprecation? Or a removal that didn't come all the way through?

packages/core/src/forms/register.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 import '@cds/core/icon/register.js';
+import '@cds/core/button-action/register.js';
```
mathisscott, line 8: Control action is being imported 6 lines south of here. So it seems like the import comment in `control-action.element.ts` should be switched back to this file? Or maybe that comment should be removed altogether?

packages/core/src/select/select.element.ts
```diff
@@ -44,7 +44,7 @@ import styles from './select.element.scss';
  */
 export class CdsSelect extends CdsControl {
   protected get suffixDefaultTemplate() {
-    return html`<cds-control-action readonly><cds-icon shape="angle" direction="down"></cds-icon></cds-control-action>`;
+    return html`<cds-button-expand pressed readonly></cds-button-expand>`;
```
mathisscott, line 47: Are we forcing pressed state here for styles?

mathisscott (changes requested):

packages/core/src/forms/control-action/control-action.element.ts
coryrylan: Yeah will fix, that looks like I copy/pasted to much

packages/core/src/forms/control/control.element.scss
coryrylan: All the action buttons have a `[cds-button-action]` attribute that way you can query/slot any action button or alias (`cds-button-action` or `cds-control-action`). The datagrid has some similar queries since you can project a regular action button or one of the other types. I can add a comment to this line here as well https://github.com/coryrylan/clarity/blob/topic/core-action-button/packages/core/src/button-action/button-action.element.ts#L43

packages/core/src/forms/register.ts
coryrylan: Yeah you're right, not sure why I have that import comment there

packages/core/src/select/select.element.ts
coryrylan: Yeah styles only, but I could add a CSS rotate if needed. The `ariaPressed` controller will only set the `aria-pressed` if the button is not set to readonly that way you can visually change the button however needed (pressed, disabled) but continue to treat it as content only.

mathisscott: That's fine. I was asking for my own understanding. I don't think it needs to be changed.

packages/core/src/forms/control/control.element.scss
coryrylan: I'm not a fan of the name however, if I think of something I'll update it. Could technically could do something like `[cds-button]` on the base then components could query for any `[cds-button]` attr which basically all clarity buttons would then have.

packages/core/src/forms/register.ts
coryrylan: fixed

packages/core/src/forms/control-action/control-action.element.ts
coryrylan: fixed

mathisscott (approved):

PR #6513, opened 2021-12-13
feat(core): clean up clarity motion decorator

mathisscott:

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Initially, the scope of work was to convert the `@animate` class decorator into a Reactive Controller. As development got underway, it became clear that a reactive controller would not meet our needs. 

The decorator needed access to `props`. Reactive controllers don't get those. The decorator also needed to interject itself in the `updated` lifecycle loop. Specifically, at a point in time prior to the component running its own `updated` callbacks. The reactive controller can't do that either.

That said, we learned somethings from the datagrid work and I was able to clean up the `animate` decorator and eliminate the need for the `implements Animatable` step – removing the `Animatable` interface entirely.

I also added some guards and moved a couple of things around (moving the `_animations` config to the instance instead of (errantly on my part) the constructor. Anecdotally, this seemed to improve performance (slightly).

- [x] Yes – `Animatable` interface removed
- [ ] No

### DIFF
--- a/packages/core/src/internal-components/overlay/overlay.element.ts
+++ b/packages/core/src/internal-components/overlay/overlay.element.ts
@@ -5,7 +5,6 @@
  */
 
 import {
-  Animatable,
   AnimationModalEnterName,
   animate,
   baseStyles,
@@ -223,7 +222,7 @@ export class CdsInternalStaticOverlay extends CdsBaseFocusTrap {
     false: AnimationModalEnterName,
   },
 })
-export class CdsInternalOverlay extends CdsInternalStaticOverlay implements Animatable {
+export class CdsInternalOverlay extends CdsInternalStaticOverlay {
   @property({ type: String })
   cdsMotion = 'on';
 
--- a/packages/core/src/internal/decorators/animate.spec.ts
+++ b/packages/core/src/internal/decorators/animate.spec.ts
@@ -5,14 +5,7 @@
  */
 
 import { animate } from './animate.js';
-import {
-  Animatable,
-  AnimatableElement,
-  event,
-  EventEmitter,
-  property,
-  registerElementSafely,
-} from '@cds/core/internal';
+import { AnimatableElement, event, EventEmitter, property, registerElementSafely } from '@cds/core/internal';
 import { html, LitElement } from 'lit';
 import { componentIsStable, createTestElement, removeTestElement } from '@cds/core/test';
 
@@ -29,7 +22,7 @@ declare global {
     no: 'nothing',
   },
 })
-class TestAnimateElement extends LitElement implements Animatable {
+class TestAnimateElement extends LitElement {
   @property({ type: String })
   cdsMotion = 'on';
 
@@ -43,7 +36,7 @@ class TestAnimateElement extends LitElement implements Animatable {
 
 // sanity check if bad data comes through
 @animate(null)
-class TestEmptyAnimateElement extends LitElement implements Animatable {
+class TestEmptyAnimateElement extends LitElement {
   @property({ type: String })
   cdsMotion = 'on';
 
@@ -78,13 +71,13 @@ describe('animate decorator', () => {
 
   it('should set _animations hidden property', async () => {
     await componentIsStable(component);
-    expect(component._animations).toBeDefined();
+    expect(component._animations).not.toBeNull();
     expect(component._animations.isValid.yes).toBe('something');
     expect(component._animations.isValid.no).toBe('nothing');
   });
 
   it('should handle bad data', async () => {
     await componentIsStable(emptyComponent);
-    expect(emptyComponent._animations).toBeUndefined();
+    expect(emptyComponent._animations).toBeNull();
   });
 });
--- a/packages/core/src/internal/decorators/animate.ts
+++ b/packages/core/src/internal/decorators/animate.ts
@@ -12,26 +12,36 @@ import {
 } from '../motion/interfaces.js';
 import { runPropertyAnimations } from '../motion/utils.js';
 
-// https://www.typescriptlang.org/docs/handbook/decorators.html#class-decorators
+// decorator factory that extends the component constructor to inject animations code into it
 export function animate(config: PropertyDrivenAnimation) {
   // eslint-disable-next-line @typescript-eslint/ban-types
-  return function _DecoratorName<T extends { new (...args: any[]): {} }>(constr: T) {
-    const _constr = constr as any;
-    return (class extends _constr {
-      _animations: any;
+  return function _animationDecorator<T extends { new (...args: any[]): AnimatableElement }>(constructor: T) {
+    return (class extends constructor {
+      private _animationReady = false;
+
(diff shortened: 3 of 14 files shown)


packages/core/src/internal-components/overlay/overlay.element.ts
```diff
@@ -223,7 +222,7 @@ export class CdsInternalStaticOverlay extends CdsBaseFocusTrap {
     false: AnimationModalEnterName,
   },
 })
-export class CdsInternalOverlay extends CdsInternalStaticOverlay implements Animatable {
```
mathisscott, line 226: No longer need this `implements`. The decorator will complain if `cdsMotion` or `cdsMotionChange` are not present.

packages/core/src/internal/decorators/animate.spec.ts
```diff
@@ -78,13 +71,13 @@ describe('animate decorator', () => {
...
-    expect(component._animations).toBeDefined();
+    expect(component._animations).not.toBeNull();
     expect(component._animations.isValid.yes).toBe('something');
     expect(component._animations.isValid.no).toBe('nothing');
   });
 
   it('should handle bad data', async () => {
     await componentIsStable(emptyComponent);
-    expect(emptyComponent._animations).toBeUndefined();
+    expect(emptyComponent._animations).toBeNull();
```
mathisscott, line 81: Changes wound up converting this lookup to be a config object or `null`.

packages/core/src/internal/decorators/animate.ts
```diff
@@ -12,25 +12,39 @@ import {
...
+        // const self = (this as unknown) as AnimatableElement;
+        // self._animations = config || void 0;
 
-        if (!self.hasAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME)) {
-          self.setAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME, AnimationStatus.ready);
-        } else if (!self.hasAttribute('_demo-mode')) {
-          // ignore if element has the _demo-mode attribute set; _demo-mode is used for docs, static examples, and stories
-          runPropertyAnimations(props, self);
+        // init-style booleans here to skip need for expensive attr lookup
+        if (!this._animationReady && !this.hasAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME)) {
```
mathisscott, line 29: Conditional complexity is due to guards that simplify/reduce the number of lookups the decorator needs to do.

packages/core/src/internal/motion/interfaces.ts
```diff
@@ -64,9 +65,9 @@ export interface MotionRegistry {
 
 /* ANIMATED COMPONENTS */
 
-export interface Animatable {
+export type AnimatableElement = ReactiveElement & {
```
mathisscott, line 68: `AnimatableElement` type improvements helped a lot. Thanks @hippee-lee ! ❤️

packages/core/src/internal/motion/motion.stories.mdx
```diff
@@ -34,13 +34,12 @@ Clarity Motion is customizable at several levels:
 
 ## Installation
 
-To animate a component with ClarityMotion, import the `animate` decorator and `Animatable` interface in your JavaScript. The `Animatable` interface will require the animated component to have a `cdsMotion` property and `motionChange` event emitter.
+To animate a component with ClarityMotion, import the `animate` decorator. The `@animate` decorator will require the component to have a `cdsMotion` property and `cdsMotionChange` event emitter.
```
mathisscott, line 37: Docs updated to remove references to `Animatable`

packages/core/src/internal/motion/utils.ts
```diff
@@ -61,13 +61,13 @@ export async function runPropertyAnimations(props: Map<string, any>, hostEl: Ani
     propertyAnimations.map((propNameAnimationTuple: TargetedAnimationAsPropertyTuple) => {
       const [propname, propertyAnimationsByValue] = propNameAnimationTuple;
 
-      if (props.get(propname) === hostEl[propname]) {
+      if (props.get(propname) === (hostEl as any)[propname]) {
```
mathisscott, line 64: TS was complaining...

packages/core/src/internal/decorators/animate.ts
```diff
@@ -12,25 +12,39 @@ import {
...
+          // this prevents animations from loading onpageload
+          this.setAttribute(PRIVATE_ANIMATION_STATUS_ATTR_NAME, AnimationStatus.ready);
+          this._animationReady = true;
+          return;
+        } else if (this._animationDemoMode) {
```
bdryanovski: I have a question more or less ?

Why we don't combine the two ifs into one the same way as we do for the above case.

```ts
if (this._animationDemoMode || this.hasAttribute('_demo-mode') { 
  this._animationDemoMode = true;
  return;
}
```
Here the v8 will not do the assignment twice because the internal state is not changed so it won't have a performance penalty, instead, gonna save `if-else-if` code branching?

We could even drop the last `else` block in that case

```ts
if (animation) { 
  return
}

if (preventAnimations) {
  return 
}

runPropertyAnimations(props, this);
```
Maybe I'm missing something it's only reading-and-compiling in my head. 

I think it's more clear what we do with the `else` block and the lot of `if-else` sections, and will require less unit-test for branched sections.

And something else, I like the idea of `_demo_mode` attribute to prevent animations, this will be super helpful for visual testing in the case where this was not an option (dropdown/select/loading and more) - why we don't rename it to something that could be part of the API like `disabledMotionAnimation` something pretty obvious and hard to miss-type ;)  

I think this could be a really good addition to the API, to be honest, I think it could have some added factor to unit-testing as well.

bdryanovski (approved):

packages/core/src/internal/decorators/animate.ts
mathisscott: I can clean up the conditional a bit. I thought it was easy to follow when I wrote it but I was down in it so... 🤷‍♂️

Thanks for the suggestions.

`_demo-mode` already exists. I didn't add anything new there. I just saved us lookups on the storybook site. The only caveat is that folks (meaning _us_) shouldn't be adding/removing demo-mode willy nilly.

bdryanovski: `_demo-mode` is new for me ... but I plan to abuse it :) 

I hoped that this will solve one of my issues with `accordion` for example I needed to wait additional `200ms` before snapshotting it and wait for the content inside to be rendered. Now when I think more about it - I see that I was totally wrong about it - and this won't do any difference for me. 

But will help with testing the `loading` animations - I hope so! 

Thanks. I don't have any more questions

end

coryrylan (approved):

packages/core/src/internal/decorators/animate.ts
mathisscott: 🤣

I'm here for the Perd Hapley memes...

PR #6514, opened 2021-12-13
fix(core:card): alignment in grid layout

### LINKED ISSUE #6457: core(cds-card): cards have incorrect sizes when using in a grid layout
## Describe the bug

When located next to other cards inside a grid, the `CdsCard` component shrinks when its content does not take the full size, or expands below other components when its content overflows.

Screen Shot 2021-11-08 at 11 45 03 AM

## How to reproduce

https://codesandbox.io/s/blissful-moon-ei8yr?file=/src/App.tsx

The top 6 cards are built with `CdsCard`, the bottom 6 cards are simple `div`s styled like cards. Both should look somewhat the same, as the bottom 6 cards are what we would expect to end up with when using `CdsCard`s which are just simple, generic containers.

## Expected behavior

Screen Shot 2021-11-08 at 11 42 51 AM

Note that the overflowing `div` is expected, this is on the consumer to handle the overflow correctly, it is just to demonstrate that the behavior differs.
In our case, we render charts in those cards but that would make for a much less contrived sandbox: when `CdsCards` they incorrectly overflow the container, but when using regular `div`s, they size themselves correctly:

`CdsCard` | Regular `div`
--- | ---
Screen Shot 2021-11-08 at 12 20 35 PM | Screen Shot 2021-11-08 at 12 20 54 PM
Screen Shot 2021-11-08 at 12 23 03 PM | Screen Shot 2021-11-08 at 12 22 44 PM

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.6.0

**Framework:**

- [ ] Angular
- [x] React 17.0.2
- [ ] Vue
- [ ] Other:

**Device:**

- OS: macOS
- Browser: Chrome
- Version 94

### LINKED ISSUE #6464: @cds/react doesn't technically support React 17

ashleyryan:
Fix #6464

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue #6464 

@cds/react has react 16 as a dependency which can cause two version of react if apps are using v17.

## What is the new behavior?

react is now a peer dependency 

- [x] Yes
- [ ] No

I think moving a dependency to a peerDependency is technically a breaking change, but there's no way an app would use @cds/react without already having react/react-dom.

## Other information

There's probably a lot more to test with this, I was hoping this could be a jumping off point for discussion

### DIFF
--- a/packages/core/src/card/card.element.scss
+++ b/packages/core/src/card/card.element.scss
@@ -1,8 +1,6 @@
 @import './../styles/tokens/generated/index';
 
 :host {
-  --width: fit-content;
-  --height: auto;
   --padding: var(--cds-global-space-6) var(--cds-global-space-8);
 
   --background: #{$cds-alias-object-container-background};
@@ -11,10 +9,4 @@
   --box-shadow: #{$cds-alias-object-shadow-100};
 
   --card-remove-margin: var(--cds-global-space-8);
-}
-
-@supports (-moz-appearance: none) {
-  :host {
-    --width: -moz-fit-content;
-  }
 }
--- a/packages/core/src/card/card.stories.ts
+++ b/packages/core/src/card/card.stories.ts
@@ -385,3 +385,40 @@ export function WithLayoutAndOverflow() {
     </div>
   </cds-card>`;
 }
+
+export function WithinGrid() {
+  return html`<div cds-layout="grid gap:md cols:6 align:stretch">
+    <cds-card aria-labelledby="gridCard1">
+      <h3 id="gridCard1" cds-text="section" cds-layout="p-y:sm">Card Title</h3>
+      <div cds-text="body light" cds-layout="p-y:md">
+        Card Content. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Praesent volutpat tortor eget quam
+        auctor, quis sagittis libero auctor. Nulla augue ante, tincidunt sit amet semper vitae, tempus at ipsum.
+        Vestibulum elementum, turpis quis ullamcorper fermentum, elit turpis placerat ipsum, quis convallis ex nisi sit
+        amet lacus. Ut enim ipsum, tincidunt nec luctus id, pharetra id velit.
+      </div>
+    </cds-card>
+    <cds-card aria-labelledby="gridCard2">
+      <h3 id="gridCard2" cds-text="section" cds-layout="p-y:sm">Card Title</h3>
+      <div cds-text="body light" cds-layout="p-y:md">
+        Card Content.
+      </div>
+    </cds-card>
+    <cds-card aria-labelledby="gridCard3">
+      <h3 id="gridCard3" cds-text="section" cds-layout="p-y:sm">Card Title</h3>
+      <div cds-text="body light" cds-layout="p-y:md">
+        Card Content. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Praesent volutpat tortor eget quam
+        auctor, quis sagittis libero auctor.
+      </div>
+    </cds-card>
+    <cds-card aria-labelledby="gridCard4">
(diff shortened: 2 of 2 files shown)


packages/core/src/card/card.element.scss
```diff
@@ -18,3 +18,18 @@
     --width: -moz-fit-content;
   }
 }
+
+:host-context([cds-layout~='grid'][cds-layout*='align:horizontal-stretch']) {
```
ashleyryan: This is the only place in the code base that uses host-context, so I'm not positive this is the correct approach?

coryrylan: This is because host-context does not have wide enough [browser support](https://caniuse.com/?search=%3Ahost-context()) unfortunately. We also probably would want take a different approach since this ties the implementation to specifically cds-layout. The card bug still happens if the card is placed in a standard CSS grid layout.

I think the fix might be to allow the card to fill its container by setting the width and height custom properties to default to 100% rather than fit-content and auto. This would align card to the rest of our block component behavior like alerts/forms/tables. This is slight breaking change from the default behavior (which is fine for this PR on 6.0/next).

coryrylan (changes requested):
See comments

packages/core/src/card/card.element.scss
ashleyryan: Ok I've updated it to remove the fit-content and auto. It now inherits the width/height 100% from the Panel component

coryrylan (approved):

packages/core/src/card/card.element.scss
```diff
@@ -1,8 +1,6 @@
 @import './../styles/tokens/generated/index';
 
 :host {
-  --width: fit-content;
-  --height: auto;
   --padding: var(--cds-global-space-6) var(--cds-global-space-8);
 
   --background: #{$cds-alias-object-container-background};
```
coryrylan, line 6: We may want to remove the moz prefixing on lines 14-18 otherwise I think firefox will override the width from inheriting from the panel  https://github.com/vmware/clarity/blob/7b55985612d27af7a170093a6b859bd32e58ddf0/packages/core/src/card/card.element.scss#L14

ashleyryan: Oh right, duh.

I noticed you have an open PR that removes some of the y-padding in the Card stories. Should I remove that here as well or in your PR?

coryrylan: I don't think you need to add to this PR. The other PR is for the spacing token updates and the adjustments for them. Edit: sorry I see your demo added here. I can get this in the space token PR since this probably will be merged first.

mathisscott (approved):

PR #6523, opened 2021-12-17
AdoptionGuide: updating documentation pages

bdryanovski:
Updating Adoption guide documentation like Get Started, Intro, Landing page, Additional Learning and more.


src/styles.css
```diff
@@ -7,10 +7,22 @@ pre {
 
 .branding img {
   height: 2rem;
+  vertical-align: middle;
+  width: 36px;
+  height: 36px;
+}
+
+.branding .logo-text {
+  color: var(--clr-header-title-color,#fafafa);
```
bbogdanov: could we use the `core` tokens in here?

bdryanovski: I could do even better .. "#fafafa" only :)

bdryanovski: I got the style from clarity.design - but that will change soon so it ok to simplify it for now

src/app/app-routing.module.ts
```diff
@@ -45,8 +45,9 @@ import { TabPage } from './pages/tab.page';
 import { LandingPage } from './pages/landing.page';
 import { OverviewPage } from './pages/overview.page';
 import { ApproachesPage } from './pages/approaches.page';
-import { DifferencesPage } from './pages/differences.page';
-import { TemplateBuilderPage } from './pages/template-builder.page';
+// import { TemplateBuilderPage } from './pages/template-builder.page';
```
mathisscott, line 48: comment still needed?

src/app/app-routing.module.ts
```diff
@@ -86,9 +87,9 @@ export const componentRoutes = [
 export const getStartedRoutes = [
   { path: 'introduction', component: IntroPage },
   { path: 'get-started', component: GettingStartedPage },
-  { path: 'differences', component: DifferencesPage },
-  { path: 'approaches', component: ApproachesPage },
-  { path: 'builder', component: TemplateBuilderPage },
+  { path: 'additional-learning', component: AdditionalLearningPage },
+  { path: 'additional-strategies', component: ApproachesPage },
+  // { path: 'builder', component: TemplateBuilderPage },
```
mathisscott, line 92: comment still needed?

src/styles.css
```diff
@@ -7,10 +7,22 @@ pre {
 
 .branding img {
   height: 2rem;
+  vertical-align: middle;
+  width: 36px;
```
mathisscott: why not rem?

src/styles.css
```diff
@@ -7,10 +7,22 @@ pre {
...
+  height: 36px;
+}
+
+.branding .logo-text {
+  color: #fafafa;
```
mathisscott, line 9: seems like you'd want to use a color token here. especially if you need to support a dark theme.

mathisscott (changes requested):

src/app/app-routing.module.ts
bdryanovski: yes, i'm not quite sure when we gonna make this public - and don't want to delete the code for it at the moment - maybe later on we could make it public

src/styles.css
bdryanovski: Got it from the clarity.design - but I guess rem could work better

src/styles.css
bdryanovski: I really want to use the headers from core but at the moment this is not part of the changes so I'm keeping it simple. The end result in my head will be to use only core components and remove the dependency for `clr/angular`

PR #6532, opened 2021-12-22
fix: breaking changes for core in v6

bdryanovski:
Deprecating code that no longer will be shipped with v6 when released. This
include things like:

  - focus-trap
  - button-inline


bbogdanov (approved):
Awesome!

mathisscott (approved):

PR #6537, opened 2021-12-23
fix(angular): size button properly when initialized to loading

### LINKED ISSUE #5967: Bug: Loading Button Sizing Problems
## Describe the bug

When using a clarity button with `[clrLoading]` the initial width is not wide enough.

image

After it stops loading it adjusts to the correct size.
But if it starts loading again the spinner icon contains an unexpected ellipsis.

image

## How to reproduce

https://stackblitz.com/edit/clarity-v4-loading-button

1.  Go to https://stackblitz.com/edit/clarity-v4-loading-button
2.  See the cut spinner icon
3.  Click 'Stop Loading' & 'Start Loading'
4.  See '...' behind loading spinner

## Expected behavior

* button should have initial min width of spinner icon
* in loading state the spinner icon should be shown without ellipsis

## Additional Notes

Maybe it would help to set a min-width for the initial state:
https://github.com/vmware/clarity/blob/next/packages/angular/projects/clr-angular/src/button/button-loading/loading-button.ts#L105

Maybe the ellipsis text-overflow can be disabled for `.btn-icon`.

steve-haar:
fixes #5967

Signed-off-by: Steve Haar <info@stevehaar.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Initializing a button to loading will cause it to have the width of the button as if it only contained the loading spinner. Subsequently (or just not initially) setting the button to loading, the button will have the width of the button in its default state.

Issue Number: https://github.com/vmware/clarity/issues/5967

## What is the new behavior?
Buttons initialized in the loading state will have the width of the button in its default state.

- [ ] Yes
- [x] No

## Other information
Before this commit a button initialized to loading would never have its default state rendered. Without rendering, it's impossible (I think) to determine what the width of the content would be. Therefore, in order for the button initialized to loading to have the width of the button in its default state, we must first render the default state, determine its width, then set the button to the loading state. I could think of two ways to solve this:

1) don't set the button to loading until after the view has initialized and the button has been initially rendered in its default state (we can use `ngAfterViewInit` in either the `clrLoading` directive or the loading button component). To do this we would wait for the view to initialize and then dispatch the state change. This will require a `setTimeout` otherwise we will get an `ExpressionChangedAfterItHasBeenCheckedError` in dev mode.

2) Always render the button's content, just set it to `visibility:hidden` and/or set it to have an absolute position. This way we could get the size of the content without having to show it.

Option 2 seemed more risky since changing the positioning of the content could potentially affect its width, also there could be performance implications with rendering more content. I thought option 1 was best, but please weigh in if you're reading this.

I also had to modify `/golden/clr-angular.d.ts`, so please let me know if there's anything further I need to do.

## Workaround
As a workaround, the developer could take care to not set `clrLoading` to `true` until `ngAfterViewInit`.

### DIFF
--- a/projects/angular/src/utils/loading/loading.ts
+++ b/projects/angular/src/utils/loading/loading.ts
@@ -3,7 +3,7 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Directive, Input, OnDestroy, Optional } from '@angular/core';
+import { AfterViewInit, Directive, Input, OnDestroy, Optional } from '@angular/core';
 
 import { LoadingListener } from './loading-listener';
 
@@ -15,13 +15,14 @@ export enum ClrLoadingState {
 }
 
 @Directive({ selector: '[clrLoading]' })
-export class ClrLoading implements OnDestroy {
+export class ClrLoading implements AfterViewInit, OnDestroy {
   public static ngAcceptInputType_loadingState: boolean | ClrLoadingState | null | string;
 
   // We find the first parent that handles something loading
   constructor(@Optional() private listener: LoadingListener) {}
 
   private _loadingState: ClrLoadingState | string = ClrLoadingState.DEFAULT;
+  private _viewInitialized = false;
 
   public get loadingState() {
     return this._loadingState;
@@ -40,9 +41,16 @@ export class ClrLoading implements OnDestroy {
     }
 
     this._loadingState = value;
-    if (this.listener) {
+    if (this.listener && this._viewInitialized) {
       this.listener.loadingStateChange(value);
     }
+  }
+
+  ngAfterViewInit() {
+    this._viewInitialized = true;
+    setTimeout(() => {
+      this.listener.loadingStateChange(this._loadingState);
+    });
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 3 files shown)


projects/angular/src/utils/loading/loading.ts
```diff
@@ -40,11 +41,18 @@ export class ClrLoading implements OnDestroy {
...
     this._loadingState = value;
-    if (this.listener) {
+    if (this.listener && this._viewInitialized) {
       this.listener.loadingStateChange(value);
     }
   }
 
+  ngAfterViewInit() {
+    this._viewInitialized = true;
+    setTimeout(() => {
```
Shijir, line 51: This will trigger change detections. Apps can have many ClrLoading instances within itself. So we have to avoid solutions that rely on timeouts.

Shijir (changes requested):

projects/angular/src/utils/loading/loading.ts
Shijir: We avoid the `setTimeout` workaround in almost all cases. We only use it when we exhaust all other options and its use cases would be limited within the app.

steve-haar: I have another solution that I proposed up in the comments (using visibility:hidden), but I thought this was the lesser of the two evils. I could avoid the `setTimeout` by forcing change detection with `ChangeDetectorRef.detectChanges`. That will still trigger another change detection, but at least it will be more explicit. Perhaps there's a better way, but my strategy was to render the button to get the proper size and then immediately hide it.

steve-haar:
closing in favor of https://github.com/vmware/clarity/pull/6573

PR #6546, opened 2022-01-04
fix: only focus on modal body if scrollable

steve-haar:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A modal body will always have a tab index, even if the body is not scrollable.

Issue Number: N/A
closes https://github.com/vmware/clarity/pull/6493

## What is the new behavior?
A modal body should only have a tab index if the content is tall enough to make the body scrollable. If the content is not tall enough, then the modal body is not operable and should not have a tab index. 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/modal/modal-body.spec.ts
+++ b/projects/angular/src/modal/modal-body.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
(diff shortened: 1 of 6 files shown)


projects/angular/src/modal/modal-body.spec.ts
```diff
@@ -21,11 +43,20 @@ describe('ClrModalBody Directive', () => {
     modalBodyEl = fixture.componentInstance.testElement.nativeElement;
   });
 
-  it('adds tabindex="0" to modal body to make content focusable and scrollable with keyboards', () => {
+  it('toggles tabindex="0" when modal body is scrollable', function () {
+    resizeObserverCallback();
```
mathisscott, line 48: What are the repeated `resizeObserverCallback()` executions doing?

mathisscott (changes requested):

projects/angular/src/modal/modal-body.spec.ts
steve-haar: Sorry, I just added a comment to the code right as you were reviewing this. There is an issue with ResizeObservable not working with angular tests: https://github.com/angular/angular/issues/31695#issuecomment-507832750. This was the only workaround I could find. Using fakeAsync/tick does not seem to work.

mathisscott: Got it. Thanks for the clarification!

mathisscott (approved):

coryrylan (approved):

Issue #6432, opened 2021-10-22
Angular combobox takes string but treat all values as true


rbirkgit:
## Describe the bug

The combobox clrMulti property takes a string, but treats any value (except empty string) as true. 

```
@Input('clrMulti')
set multiSelect(value: boolean | string) {
  if (value) {
    this.optionSelectionService.selectionModel = new MultiSelectComboboxModel<T>();
  } else {
    this.optionSelectionService.selectionModel = new SingleSelectComboboxModel<T>();
  }
```

## How to reproduce

We had a typo in our code:
`[clrMulti]="'false'"`

I expected the strictTemplate setting to catch it, but the source of combobox as seen above takes a string input but doesn't check the content of the string, just if a string has value.

## Expected behavior

Either not supporting string or parse the string to actual tue/false.

## Versions

**Clarity project:**

- [ ] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [x] v12.x

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 12

bdryanovski:
The reason for having `string` as a type is that the strict Angular (`fullTemplateTypeCheck` & `strictTemplates`) check requires every field that is `boolean` to have an additional type of `string`.

The reason is that internally Angular and DOM don't have an idea of what is a `boolean` everything is a string. Angular works with strings that look like boolean and try to handle this - sort of. The solution you had is right. And you could also notice that non-empty strings will be `true`. 

Here maybe will be a good place to have additional code to check strings (like `'true'` & `'false'`) and convert them to `boolean`.

I don't think that this is a bug - but we are open to accepting PR with changes to handle the above case.

rbirkgit:
To me this is a clear bug and caused issues in our code. Either it should only accept an empty string as type or you have to convert string to proper vavlue. That how it is done in other components.

rbirkgit:
It's seems inconsistent throughout the Clarity code base with many bugs. Some places do it this way:

```
  set clrInline(value: boolean | string) {
    if (typeof value === 'string') {
      this.inline = value === 'false' ? false : true;
    } else {
      this.inline = !!value;
    }
  }
```

Other places like this:
```
  @Input('disabled')
  set disabled(value: boolean | string) {
    if (this.dateFormControlService) {
      this.dateFormControlService.setDisabled(isBooleanAttributeSet(value));
    }
  }
```

[...]

PR #6547, opened 2022-01-05
fix(angular): Combobox clrMulti input handle string value of 'false'

IrisTien:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #6432

## What is the new behavior?
Combobox Component clrMulti input handles both string and boolean type values.
If clrMulti input is boolean, it will pass the boolean value directly;
If clrMulti input is string: it would treat "false" and empty string to boolean false, it would set to true for all the other strings.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/forms/combobox/combobox.ts
+++ b/projects/angular/src/forms/combobox/combobox.ts
@@ -194,13 +194,23 @@ export class ClrCombobox<T>
 
   @Input('clrMulti')
   set multiSelect(value: boolean | string) {
-    if (value) {
-      this.optionSelectionService.selectionModel = new MultiSelectComboboxModel<T>();
-    } else {
-      // in theory, setting this again should not cause errors even though we already set it in constructor,
-      // since the initial call to writeValue (caused by [ngModel] input) should happen after this
-      this.optionSelectionService.selectionModel = new SingleSelectComboboxModel<T>();
+    let isMultiSelectModel: boolean;
+
+    switch (true) {
+      case value === 'false' || value === false:
+        isMultiSelectModel = false;
+        break;
+      case value === 'true' || value === true:
+        isMultiSelectModel = true;
+        break;
+      default:
+        isMultiSelectModel = false;
     }
+    // in theory, setting this again should not cause errors even though we already set it in constructor,
+    // since the initial call to writeValue (caused by [ngModel] input) should happen after this
+    this.optionSelectionService.selectionModel = isMultiSelectModel
+      ? new MultiSelectComboboxModel<T>()
+      : new SingleSelectComboboxModel<T>();
     this.updateControlValue();
   }
 
--- a/projects/angular/src/forms/combobox/combobox.spec.ts
+++ b/projects/angular/src/forms/combobox/combobox.spec.ts
@@ -37,7 +37,7 @@ import { ClrComboboxModule } from './combobox.module';
   `,
 })
 class TestComponent {
-  multi: boolean;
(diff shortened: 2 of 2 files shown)


projects/angular/src/forms/combobox/combobox.ts
```diff
@@ -194,7 +194,14 @@ export class ClrCombobox<T>
 
   @Input('clrMulti')
   set multiSelect(value: boolean | string) {
-    if (value) {
+    let isMultiSelectModal = false;
```
gnomeontherun: I'm not sure in what situation you would be passing a string here. This is meant to be a boolean, and generally its good to restrict typing in these cases. It can lead to confusing outcomes for consumers if they think a string can be passed like `"0"`, is sending a boolean really a problem?

Regardless, this is a lot of logic. We can just coerce the value into a boolean and not try to handle individual type cases.

```js
value = !!value;
```

Jinnie: On our website the usage is documented as:
> The multi selection mode is triggered with a clrMulti="true" 

This implies that `clrMulti="false"` should also work.
And I don't thing that the coercion will not work with the case of `clrMulti="false"`

mathisscott: This would be better checking against more explicit values. There's a lot of unnecessary `typeof` checks here that are only necessary because we are duck typing the value.

So something more like...

```
let isMultiSelectModal: boolean;

switch (true) {
  case (value === 'false' || value === false):
     isMultiSelectModal = false;
     break;
  case (value === 'true' || value === true):
     isMultiSelectModal = true;
     break;
   default:
     isMultiSelectModal = false;
}
```

... then you could finish off with something like...

```
this.optionSelectionService.selectionModel = isMultiSelectModal ? new MultiSelectComboboxModel<T>() : new SingleSelectComboboxModel<T>();
```

...narrowing down a lot of conditional code here.

mathisscott (changes requested):

projects/angular/src/forms/combobox/combobox.ts
IrisTien: Please check the comments from Bozhidar Dryanovski in issue #6432 :
`The reason for having string as a type is that the strict Angular (fullTemplateTypeCheck & strictTemplates) check requires every field that is boolean to have an additional type of string.`

So here we have the fix to support string value 'true' and 'false'.

@mathisscott , addressed the comments.

mathisscott (approved):

bbogdanov (approved):

Jinnie:
Hello, @IrisTien,

There is a small formatting error that needs to be fixed. And also we need those 4 commits squashed, and then we can merge this fix in.

You can easily accomplish both tasks by soft resetting to the latest "good" commit, then running format:fix on the staged commits and force pushing all back. These are the commands you need:
```
git reset --soft 5911f81caae3e02ac9b9008c0d82bf07ad040529
npm run format:fix
```
Then commit (don't forget the "signed-off-by" line) and force push back to your fork. This PR will be automatically updated and build tasks rerun.

Thanks!

ashleyryan:
Hello, this repository has been archived. Please visit one of the new repositories:

- [Clarity Core](https://github.com/vmware-clarity/core)
- [Clarity Angular](https://github.com/vmware-clarity/ng-clarity)

PR #6563, opened 2022-01-10
feat(angular): do not preserve width of loading buttons

steve-haar:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, loading buttons preserve their width when they switched to the loading state.

fixes https://github.com/vmware/clarity/issues/5967

## What is the new behavior?
Loading buttons will no longer preserve their width. This also fixes styling issues when a button was initialized in the loading state.

- [ ] Yes
- [x] No

## Other Information
The previous fix for same size icons is no longer needed if the button is not going to preserve its width.

### DIFF
--- a/projects/angular/src/button/button-loading/loading-button.ts
+++ b/projects/angular/src/button/button-loading/loading-button.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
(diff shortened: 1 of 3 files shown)


bbogdanov (approved):

projects/angular/src/button/button-loading/loading-button.ts
```diff
@@ -98,11 +93,4 @@ export class ClrLoadingButton implements LoadingListener {
     }
     this.clrLoadingChange.emit(state);
   }
-
```
mathisscott, line 101: We want this behavior in there when a button switches from non-loading to loading so that the row of buttons does not shift when the loading spinner is narrower than the text of the button.

It seems the real issue here is that...

a) the minimum width for a button with a spinner may be too narrow to contain the spinner
b) a button that is initially rendered in the loading state isn't being set to the correct width

It may be a better strategy to cache a button's width when it is not in a loading state and update that cached width when/if the contents of the button change.

mathisscott (changes requested):

projects/angular/src/button/button-loading/loading-button.ts
```diff
@@ -76,20 +76,15 @@ export class ClrLoadingButton implements LoadingListener {
 
     switch (state) {
       case ClrLoadingState.DEFAULT:
-        this.renderer.removeStyle(this.el.nativeElement, 'width');
```
Shijir, line 79: Not sure how these style removals are solving the issue. The solution we talked about is that when a button is in the loading, keep its width at least at the spinner's width (it shouldn't be narrower than that). When it switches to the non-loading state, its width should expand to fit asynchronously loaded content.

steve-haar:
closing in favor of https://github.com/vmware/clarity/pull/6573

PR #6572, opened 2022-01-13
Address several tree-view element issues

hippee-lee:
This change handles several isses that were encountered when triaging
usage of the cds-tree-item with angular's *ngIf structural directive.

1. This change adds registration of for the new cds-button-expand
   element used in the temlate.
2. This change adds a code comment to further describe both the
   automatic and manual use cases for `expandable`
3. THis change adds a dynamic demo that demonstrates how to use the
   expandable with complex trees
4. This change adds a test in the tree-item spec to make sure the button
   is in the DOM when a cds-tree-item element is `expandable`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

As described above this adds a demo for conditional DOM on cds-tree-item elements. (think *ngIf for Angular apps). It also adds some context for the expandable property as it has an automatic and a manual use case. Finally, It adds a test into the spec to make sure the button is in the DOM when an element is 'expandable'. 

Missing register for cds-button-expand, no dev notes for the manual use case of the expandable property and there is no test to make sure the expand button is in the DOM when a cds-tree-item is expandable. 
Issue Number: N/A

## What is the new behavior?
Added register code, added dev context to the expandable property, added a spec for the expandable property. Finally, it adds a demo for conditional cds-tree-items in a complex tree. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/tree-view/tree-item.element.ts
+++ b/packages/core/src/tree-view/tree-item.element.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -75,6 +75,14 @@ export class CdsTreeItem extends LitElement {
   @property({ type: Boolean, reflect: true })
   expanded = false;
 
+  /**
+   * @description
+   *
+   * Expandable is used (see line 124) to automatically show the cds-button-expand element when a tree-item detects that
+   * there are children. However, when doing investigation into why styles were not applied for tree-items controlled
+   * via *ngIf (or any conditional DOM and code that will put tree-item elements there at some unknown point in the
+   * future) it was re-discovered as a public property without a demo for manual usage.
+   */
   @property({ type: Boolean, reflect: true })
   expandable = false;
 
--- a/packages/core/src/tree-view/tree.stories.ts
+++ b/packages/core/src/tree-view/tree.stories.ts
@@ -1,14 +1,18 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import '@cds/core/alert/register.js';
+import '@cds/core/button/register.js';
 import '@cds/core/icon/register.js';
 import '@cds/core/tree-view/register.js';
 import { CdsTreeItem } from '@cds/core/tree-view';
 
-import { html } from 'lit';
+import { css, html, LitElement } from 'lit';
+import { createItems, TreeItem } from './story-utils.js';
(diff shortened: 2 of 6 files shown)


packages/core/src/tree-view/tree-item.element.ts
```diff
@@ -75,6 +75,14 @@ export class CdsTreeItem extends LitElement {
   @property({ type: Boolean, reflect: true })
   expanded = false;
 
+  /**
+   * @description
+   *
+   * Expandable is used (see line 124) to automatically show the cds-button-expand element when a tree-item detects that
```
hippee-lee, line 81: These are the notes for future dev selves to describe the manual use case if a bug gets reported when a consumer puts something like the `*ngIf` structural directive on one of these elements with children.

packages/core/src/tree-view/tree.stories.ts
```diff
@@ -1,14 +1,18 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 import '@cds/core/alert/register.js';
+import '@cds/core/button/register.js';
```
hippee-lee, line 7: Registers the cds-button-expand element so consumers do not have to.

packages/core/src/tree-view/tree.stories.ts
```diff
@@ -379,6 +383,78 @@ export const interactive = () => {
   `;
 };
 
+export function conditionalItems() {
+  @customElement('demo-conditional-tree')
```
hippee-lee, line 387: Complex tree demo for conditional DOM.

packages/core/src/tree-view/utils.ts
```diff
@@ -0,0 +1,50 @@
+export type CheckboxSelection = true | false | 'indeterminate';
```
hippee-lee: I didn't think it was worth adding a utils directory since there are only two of them.

bbogdanov (approved):
Awesome!

mathisscott (approved):

packages/core/src/tree-view/utils.ts
coryrylan: May want to add a comment block at the top of this file that these are for storybook demo utils so it does not accidentally get pulled into published library code. Maybe the file could be `story-utils.ts` or `demo.ts`?

packages/core/src/tree-view/tree.stories.ts
```diff
@@ -1,14 +1,18 @@
...
  */
 import '@cds/core/alert/register.js';
+import '@cds/core/button/register.js';
 import '@cds/core/icon/register.js';
 import '@cds/core/tree-view/register.js';
 import { CdsTreeItem } from '@cds/core/tree-view';
 
-import { html } from 'lit';
+import { css, html, LitElement } from 'lit';
+import { createItems, TreeItem } from './utils';
```
coryrylan: Path should be have `.js` extension`./utils.js`. It can get a bit buggy if storybook is using import maps without it

Issue #6559, opened 2022-01-06
React(CdsButton): loadingStatus prop not working


ashleyryan:
## Describe the bug

The `loadingState` prop of `CdsButton` no longer does anything. `loading-state` must be used instead.

## How to reproduce

loadingState doesn't work: https://stackblitz.com/edit/clairty-core-react-dark-v5-zacfgr?file=src%2FApp.tsx
loading-state works: https://stackblitz.com/edit/clairty-core-react-dark-v5-56yx2w?file=src%2FApp.tsx

## Expected behavior

It's expected that the property setting for loading-state/loadingState should work when using the prop `loadingState`. It broke in 5.5.1

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [ x] v5.x

**Framework:**

- [ ] Angular
- [x] React
- [ ] Vue
- [ ] Other:

## Additional notes

Here's an example with the prop working in 5.5.0: https://stackblitz.com/edit/http-server-ugdktg?file=src%2FApp.tsx

The issue may be related to this PR: https://github.com/vmware/clarity/pull/6183/files

PR #6576, opened 2022-01-14
fix(react:button): move loadingStatus logic to `update` event

ashleyryan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In React, the `loadingStatus` prop no longer works, devs must use `loading-status`. It last worked in 5.5.0/

Issue Number: #6559

## What is the new behavior?

`loadingStatus` works in React now

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/button/button.element.ts
+++ b/packages/core/src/button/button.element.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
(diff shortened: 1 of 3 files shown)


hippee-lee (approved):

bbogdanov (approved):

packages/core/src/button/button.element.ts
```diff
@@ -80,28 +80,36 @@ export class CdsButton extends CdsBaseButton {
...
+        // track prior disabled state to set prior value after button is re-enabled from a loading state
+        this._disabled = this.disabled;
+      }
 
-    if (value === ClrLoadingState.default) {
-      this.enableButton();
-    } else {
-      this.disableButton();
+      if (props.get('loadingState') !== undefined) {
+        if (this.loadingState === ClrLoadingState.default) {
```
mathisscott: Can we move these duplicated checks into a getter? Something like `isLoading`?

mathisscott (changes requested):

packages/core/src/button/button.element.ts
mathisscott: I ask so that we can keep the code DRY. In the event that a further step is needed in determining if the button is loading, we would need to make changes in three places instead of one...

packages/core/src/button/button.element.ts
```diff
@@ -122,7 +130,6 @@ export class CdsButton extends CdsBaseButton {
   static styles = [baseStyles, styles];
 
   private async disableButton() {
```
mathisscott: If we removed the `await` we probably don't need the `async` declaration either...

mathisscott (changes requested):

packages/core/src/button/button.element.ts
ashleyryan: Updated, the other states aren't loading, even though the button is disabled for them, so I went with the very literal `isDefaultLoadingState`

mathisscott (approved):

PR #6612, opened 2022-01-31
fix(react): fix jsdom react testing issues

### LINKED ISSUE #5985: cds-select + React Testing Library: `TypeError: 'set' on proxy: trap returned falsish for property 'disabled'`
## Describe the bug

When Clarity is used in conjunction with the React Testing Library, selecting a `CdsSelect` by its corresponding label throws:

```
    TypeError: 'set' on proxy: trap returned falsish for property 'disabled'

      at node_modules/src/internal/utils/lit-element.ts:42:20
          at Array.forEach (<anonymous>)
      at syncProps (node_modules/src/internal/utils/lit-element.ts:42:6)
      at HTMLElement.updated (node_modules/src/forms/control/control.element.ts:291:5)
      at HTMLElement.performUpdate (node_modules/lit-element/src/lib/updating-element.ts:792:12)
      at node_modules/lit-element/src/lib/updating-element.ts:725:25
      at asyncGeneratorStep (node_modules/lit-element/lib/updating-element.js:8:103)
      at _next (node_modules/lit-element/lib/updating-element.js:10:194)
```

## How to reproduce

I could not reproduce this using a StackBlitz as the test runner runs outside of a browser context (and StackBlitz's super new Node beta support had some issues), but I set up a repro repo at [astorije/repro-clarity-proxy-set-disabled](https://github.com/astorije/repro-clarity-proxy-set-disabled).

1.  Clone the repo
2.  Install dependencies with `yarn install`
3.  Run `yarn test`
4.  The error posted above is thrown. This is triggered by [astorije/repro-clarity-proxy-set-disabled `index.test.tsx#L35-L54`](https://github.com/astorije/repro-clarity-proxy-set-disabled/blob/main/index.test.tsx#L35-L54)

## Expected behavior

When I comment out [this line](https://github.com/vmware/clarity/blob/a222f9972d5e0e384c4270ee036a3192236711f4/packages/core/src/forms/control/control.element.ts#L298), the test passes as expected.

Using the repo above, open `node_modules/@cds/core/forms/control/control.element.js` and comment out L191. Run the tests again and they will be green. By adding `screen.debug()` after the `expect` statement, we can see the DOM as it's supposed to be rendered:

```html
<body cds-supports="js no-flex-gap" cds-version="5.2.1">
  <div>
    <cds-select
      _nfg=""
      cds-control=""
      control-width="stretch"
      responsive=""
      status="neutral"
    >
      <style />
      <label for="_b3i52678e" slot="label">
        Foo
      </label>
      <select id="_b3i52678e" slot="input">
        <option>Bar</option>
      </select>
    </cds-select>
  </div>
</body>
```

I doubt the fix is to simply remove that line as it was recently added as a bugfix itself in https://github.com/vmware/clarity/pull/5739 or that upgrading `lit` to a non-fork will fix the issue.

I don't follow this code too well yet. `CdsSelect` extends `CdsControl`, could that mean `CdsSelect` is not set up properly to accept the `disabled` attribute? I'm happy to help with a fix but will need a little guidance 😅

## Versions

**Clarity project:** `@cds/core` + `@cds/react` v5.2.1, v5.3.0
**Framework:** Tested this is an issue with React 16 and 17.
**Node:** v14.16.1

### LINKED ISSUE #6525: Missing role attributes in Core 5.6.1
## Describe the bug

ARIA roles and attributes are missing in Clarity Core v5.6.1, failing our [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) tests that use `await screen.findByRole('button')`.

## How to reproduce

1.  I could not make this work with Stackblitz or Code Sandbox, so had to go the repro repo route: https://github.com/astorije/clarity-missing-role
2.  Clone, install dependencies (`yarn install`)
3.  Run `yarn test` on the `main` branch and see the error in the console output
4.  Switch to the `core-5.6.0` branch, install dependencies again (`yarn install`), and run `yarn test` to see the same tests pass

Here is the error output on v5.6.1:

```
 FAIL  src/App.test.tsx
  App
    ✕ foo? (1042 ms)

  ● App › foo?

    Unable to find role="button"

    Ignored nodes: comments, <script />, <style />
    <body
      cds-supports="js no-flex-gap"
      cds-version="5.6.1"
    >
      <div>
        <div
          cds-layout="m:md"
        >
          <cds-button
            _nfg=""
            action="solid"
            size="md"
            status="primary"
            tabindex="0"
          >
            Foo
          </cds-button>
        </div>
      </div>
    </body>

      10 |
      11 |     expect(
    > 12 |       await screen.findByRole("button", { name: "Foo" })
         |                    ^
      13 |     ).toBeInTheDocument();
      14 |   });
      15 | });
```

As you can tell, the `cds-button` element is missing the `role` attribute (and the `aria-disabled` one).
I suspect the issue originated at https://github.com/vmware/clarity/compare/v5.6.0...v5.6.1#diff-40c8e7c2cb3e2e864efc9529dbba5def804aa9131b0ab70c82fb87cfc86aba00
(warning: GitHub's renderer currently removes the # between `v5.6.1` and `diff` in this URL 🤷‍♂️)

## Expected behavior

The `core-5.6.0` contains the expected behavior. When running tests, we get:

```
 PASS  src/App.test.tsx
  App
    ✓ foo? (144 ms)

  console.log
    <body
      cds-supports="js no-flex-gap"
      cds-version="5.6.0"
    >
      <div>
        <div
          cds-layout="m:md"
        >
          <cds-button
            _nfg=""
            action="solid"
            aria-disabled="false"
            role="button"
            size="md"
            status="primary"
            tabindex="0"
          >
            Foo
          </cds-button>
        </div>
      </div>
    </body>

    /.../src/App.test.tsx:15:12
      13 |     ).toBeInTheDocument();
      14 |
    > 15 |     screen.debug();
         |            ^
```

## Versions

**Clarity project:**

- [x] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [x] v5.6.1

**Framework:**

- [ ] Angular
- [x] React
- [ ] Vue
- [ ] Other:

## Additional notes

To be noted that the role does appear fine in the browser but not in the test DOM, not sure what kind of magic is happening there:

Screen Shot 2021-12-20 at 12 53 09 PM

ashleyryan:
Fixes #5985
- Add isNode check to property syncing in cdscontrol to prevent error

Fixes #6525:
- Remove isNode check for aria-reflect, now that all roles are set in connectedCallback (#6593)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes

Issue #5985 
When rendering `CdsSelect` in jsdom (via react-testing-library), an exception will be thrown on render and all tests will fail.

Issue #6525
The aria-reflect polyfill doesn't run in node because of jsdom exceptions, meaning react unit tests can't target elements by `role`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/docs/react.stories.mdx
+++ b/packages/core/docs/react.stories.mdx
@@ -98,20 +98,9 @@ const Component = () => {
 ## Testing
 
 React Testing Library is [recommended by the React team](https://reactjs.org/docs/testing.html)
-for unit testing React components. Because of the way web components are rendered, the synchronous
-`get` methods of RTL might not work, use the asynchronous `find` ones instead.
-
-```javascript
-test('my test', async () => {
-  render(<CdsButton>My Button</CdsButton>);
-
-  // might not work
-  expect(screen.getByRole('button', { name: 'My Button' })).toBeInTheDocument();
-
-  // will work
-  expect(await screen.findByRole('button', { name: 'My Button' })).toBeInTheDocument();
-});
-```
+for unit testing React components. By default, RTL runs in jest using jsdom, a node-based implementation
+of browser standards. It does not have implementations of all the latest browser features,
+so there may issues rendering and interacting with Clarity components.
 
 ### IntersectionObserver
 
@@ -130,3 +119,20 @@ window.IntersectionObserver = jest.fn().mockReturnValue({
 <a href="https://github.com/vmware/clarity/tree/next/apps" target="_blank" rel="noopener">
   <cds-button>Example Apps</cds-button>
 </a>
+
+### Finding a Clarity element
+
+Because of the way web components are rendered, the synchronous
+`get` methods of RTL might not work, use the asynchronous `find` ones instead.
+
+```javascript
+test('my test', async () => {
+  render(<CdsButton>My Button</CdsButton>);
+
+  // might not work
+  expect(screen.getByRole('button', { name: 'My Button' })).toBeInTheDocument();
+
+  // will work
+  expect(await screen.findByRole('button', { name: 'My Button' })).toBeInTheDocument();
+});
+```
--- a/packages/core/src/internal/utils/environment.ts
+++ b/packages/core/src/internal/utils/environment.ts
@@ -1,0 +1,9 @@
+import { isNil } from '../utils/identity.js';
+
+export function isBrowser(win = window) {
+  return !isNil(win);
+}
+
+export function isJestTest() {
+  return (globalThis as any)?.process?.env?.JEST_WORKER_ID !== undefined;
+}
--- a/packages/core/src/internal/utils/events.ts
+++ b/packages/core/src/internal/utils/events.ts
@@ -1,8 +1,10 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
+
+import { isJestTest } from './environment.js';
 
 export function stopEvent(event: Event) {
   event.preventDefault();
@@ -29,7 +31,8 @@ export const getElementUpdates = (
 
   const updatedProp = Object.getOwnPropertyDescriptor(Object.getPrototypeOf(element), propertyKey) as any;
 
-  if (updatedProp) {
+  //  Jest and JSDom breaks defining a new property, so skip
+  if (updatedProp && !isJestTest()) {
     Object.defineProperty(element, propertyKey, {
       get: updatedProp.get,
       set: val => {
--- a/packages/core/src/internal/utils/environment.spec.ts
+++ b/packages/core/src/internal/utils/environment.spec.ts
@@ -1,0 +1,9 @@
+import { isBrowser } from './environment';
+
+describe('Environment Helper: ', () => {
+  describe('isBrowser():', () => {
+    it('returns true when expected', () => {
+      expect(isBrowser()).toBe(true);
+    });
+  });
+});
--- a/packages/core/src/internal/utils/exists.spec.ts
+++ b/packages/core/src/internal/utils/exists.spec.ts
@@ -1,10 +1,10 @@
 /*
- * Copyright (c) 2016-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { existsIn, existsInWindow, isBrowser } from './exists.js';
+import { existsIn, existsInWindow } from './exists.js';
 
 describe('Functional Helper: ', () => {
   describe('existsInWindow(): ', () => {
@@ -89,12 +89,6 @@ describe('Functional Helper: ', () => {
 
     it('returns false shallowly', () => {
       expect(existsIn(['notDefined'], myTestObject)).toEqual(false);
-    });
-  });
-
-  describe('isBrowser():', () => {
-    it('returns true when expected', () => {
-      expect(isBrowser()).toBe(true);
     });
   });
 });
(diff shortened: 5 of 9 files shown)


ashleyryan:
Putting this as a draft for now, I want to write some more RTL tests around CdsControls to test that the fix for #6525 doesn't break anything else

ashleyryan:
I repurposed the repo-clarity-proxy-set-disabled repo to test the fixes using RTL and they work now with yalc: https://github.com/ashleyryan/repro-clarity-proxy-set-disabled/blob/main/test.test.tsx

packages/core/src/internal/utils/events.ts
```diff
@@ -9,6 +9,10 @@ export function stopEvent(event: Event) {
   event.stopPropagation();
 }
 
+function isNode() {
```
mathisscott: Good job making this a utility. But I don't know if `events.ts` is a good home for it. We have a similar check for SSR called `isBrowser` that lives in `exists.ts`. 

It may make sense to take `isBrowser` and `isNode` and pull them into their own utility file. Maybe something like `environment.ts` or `platform.ts`? Whichever name makes sense.

These are internal files/utilities so moving them around like that is not considered a breaking change. Only power users are going to be aware of them and use them extensively.

mathisscott (changes requested):

packages/core/src/internal/utils/events.ts
ashleyryan: It actually already was a utility, I just moved it from one file to another 😆

Is it overkill to move it to a utility file is this is the only place it's used? I can see the benefits of having it live next to `isBrowser` though.

mathisscott: I don't think it's overkill. I was surprised that `isBrowser` is in `exists.ts`. That's weird. It's a lot easier to track these utilities down if they live somewhere sensible. The only caveat with that approach is avoiding a circular module import.

packages/core/docs/react.stories.mdx
```diff
@@ -130,3 +119,20 @@ window.IntersectionObserver = jest.fn().mockReturnValue({
 <a href="https://github.com/vmware/clarity/tree/next/apps" target="_blank" rel="noopener">
   <cds-button>Example Apps</cds-button>
 </a>
+
+### Finding a Clarity element
+
+Because of the way web components are rendered, the synchronous
+`get` methods of RTL might not work, use the asynchronous `find` ones instead.
```
mathisscott, line 126: 👍

packages/core/src/internal/utils/environment.ts
```diff
@@ -0,0 +1,9 @@
+import isNil from 'ramda/es/isNil.js';
```
mathisscott: We have our own `isNil` that can be imported from `identity.js`...

We added it because the ramda imports were going circular on us...

packages/core/src/internal/utils/events.ts
```diff
@@ -1,9 +1,11 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+import { isJestTest } from './environment';
```
mathisscott: I think this import needs to come from `./environment.js`...

mathisscott (changes requested):

packages/core/docs/react.stories.mdx
ashleyryan: I meant to update this and figured now was as good a time as any. I was going to mention puppeteer-testing-library or cypress-testing-library as alternative but don't know enough about using those with react to recommend it.

packages/core/src/internal/utils/events.ts
ashleyryan: vscode auto import, oops. Fixed

mathisscott (approved):

coryrylan (approved):

Issue #6622, opened 2022-02-01
Creat React App / Webpack 5


coryrylan:
## Describe the bug

_A clear and concise description of what the bug is, and the conditions when it occurs._

## How to reproduce

```
yarn create react-app test
cd test
yarn add @cds/core @cds/city @cds/react
yarn install --check-files
```

## Expected behavior

With Webpack 5 import paths must require a `.js` extention on the end of file imports. 
https://github.com/vmware/clarity/pull/5714 but then a regression re-introduced later

All imports in the `@cds/react` library that are a path to a direct file need to have the `.js` extention. Most are the following path

```javascript
import { logReactVersion } from '../utils/index';
```

https://github.com/vmware/clarity/blob/next/packages/react/src/badge/index.tsx#L5 

## Versions

**Clarity project:**

- [ ] Clarity Core
- [ ] Clarity Angular/UI

**Clarity version:**

- [ ] v3.x
- [ ] v4.x
- [ ] v5.x

**Framework:**

- [ ] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: [e.g. MacBook]

## Additional notes

_Add any other notes about the problem here._

mathisscott:
This should be fixed in #6627

coryrylan:
This is a different issue, https://github.com/vmware/clarity/pull/6627 did fix its intended issue

coryrylan:
Opened separate issue here https://github.com/vmware/clarity/issues/6695, thought this was a duplicate

PR #6627, opened 2022-02-02
fix(react): update react import paths to include extension

ashleyryan:
fixes: Creat React App / Webpack 5 #6622

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

#6622 Creact-react-app/webpack requires imports to have extensions

## What is the new behavior?

Update the import paths and add a new lint rule

- [ ] Yes
- [x] No


packages/react/.eslintrc.js
```diff
@@ -23,6 +23,7 @@ const rules = {
       varsIgnorePattern: 'React|Cds',
     },
   ],
+  'import/extensions': ['error', 'ignorePackages'],
```
ashleyryan: I only added this to the react package - should it be across the whole repo? @mathisscott  left a comment on my PR yesterday about file extensions in core so I'm thinking yes?

coryrylan: I thought we had this turned on in Core but it doesn't look like that's the case. We should definitely add this rule to the Core eslint config.

coryrylan (approved):

packages/react/.eslintrc.js
ashleyryan: Angular too? If I add it at the root there's only 2 or 3 issues in core, but a lot more in angular.

coryrylan: Maybe just start with Core. For Core it can make or break the library with CDN or ESM tools. Angular not as much because its always using a build step somewhere.

bdryanovski (approved):

Issue #4211, opened 2020-01-16
Wizard Focus Does Not Switch To Content After Selecting a Tab


jeff-bacon:
## Describe the bug

When selecting a tab on the left side of the wizard, the focus does not jump to that tabs content.  Keyboard only users and screen reader users will find it difficult to access the pop-up as they have to navigate through all the controls in the left navigation to reach the main content.

## How to reproduce

[https://clarity-light-theme-v2-p315xm.stackblitz.io/](https://clarity-light-theme-v2-p315xm.stackblitz.io/)

1. Tab to the **Open Wizard** button.
2. Press **Enter**
3. Tab until the **Next** button.
4. Press **Enter**
5. Repeat until the last page of the wizard can be accessed.
6. Tab until any tab has focus.
7. Press **Enter**
8. Notice that the content for that tab does not gain focus.

## Expected behavior

When clicking on a wizard tab, focus should move to that tabs content to make the wizard more accessible.

## Versions

**App**

* Angular: 8.2.4
* Clarity: 2.1.2

**Device:**

* Type: Surface Book
* OS: Windows 10
* Browser: Chrome
* Version: 79.0.3945.117

gnomeontherun:
Clarity cares deeply about accessibility, and we work with our accessibility team to identify and triage issues. We have an internal tracking system to identify and verify accessibility issues, which has captured this already. Some issues with Clarity Angular are difficult to solve without causing side effects or breaking changes, and we have to balance concerns with fixing issues and breaking changes. As we build out Clarity Core, we are able to address these root cause accessibility issues and recommend adopting the Clarity Core implementations as they become available. We are closing this issue as we have the issues tracked internally. Please follow our development and releases to see when we release relevant Clarity Core components to make this possible.

hippee-lee:
Looking into this as part of VPAT-749 (

It sounds like the a11y requirement is to move focus to the WizardPage after a WizardStepNav item is activated.

PR #6670, opened 2022-02-09
fix(wizard): poc for focus issue after nav item click

hippee-lee:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the ClrWizard component is opened focus is moved directly to the page title. When a user interacts with the wizard and clicks on a nvigation item or a page button that causes the page to change, focus stays on the clicked element. 

Issue Number: #4211 

## What is the new behavior?
After user interaction that changes the wizard page (content) focus will be moved to the page title element to make it more accessible. 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/wizard/providers/wizard-navigation.service.ts
+++ b/projects/angular/src/wizard/providers/wizard-navigation.service.ts
@@ -5,8 +5,7 @@
  */
 
 import { Injectable, OnDestroy, TemplateRef } from '@angular/core';
-import { Observable } from 'rxjs';
-import { Subject } from 'rxjs';
+import { Subject, Observable } from 'rxjs';
 import { Subscription } from 'rxjs';
 
 import { ClrWizardPage } from '../wizard-page';
--- a/projects/angular/src/wizard/wizard.ts
+++ b/projects/angular/src/wizard/wizard.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -434,7 +434,6 @@ export class ClrWizard implements OnDestroy, AfterContentInit, DoCheck {
     if (!pageId) {
       return;
     }
-
     this.navService.goTo(pageId);
   }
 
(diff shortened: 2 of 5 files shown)


mathisscott (approved):

projects/angular/src/wizard/wizard.ts
```diff
@@ -471,7 +471,15 @@ export class ClrWizard implements OnDestroy, AfterContentInit, DoCheck {
   }
 
   private listenForPageChanges(): Subscription {
-    return this.navService.currentPageChanged.subscribe(() => this.currentPageChanged.emit());
+    return this.navService.currentPageChanged.subscribe(() => {
```
Shijir, line 473: When we return these subscriptions, where do we unsubscribe them?

hippee-lee: Subscriptions are unsubscribed in the [ngOnDestroy on line 269](https://github.com/vmware/clarity/pull/6670/files#diff-76309545a2b4ee57076610f8514507e223a4ccf73b9fe8be31ff8b6cd81c4ebfR269)

hippee-lee: This component follows the familiar `hold observables in an array and then unsubscribe from all of them at once` in the ngOnDestroy lifecycle.

projects/angular/src/wizard/providers/wizard-navigation.service.ts
```diff
@@ -178,6 +177,15 @@ export class WizardNavigationService implements OnDestroy {
     return this._currentChanged.asObservable();
   }
 
+  // Added to address VPAT-749:
+  //   When clicking on a wizard tab, focus should move to that
+  //   tabs content to make the wizard more accessible.
+  private navItemChanged = new Subject<ClrWizardPage>();
```
Shijir: nit: I think we should stick to the convention of declaring a subject with _ and then having its observable without the underscore like:
```
private _currentChanged = new Subject<ClrWizardPage>();
public get currentPageChanged(): Observable<ClrWizardPage> {};
```

projects/angular/src/wizard/wizard.ts
```diff
@@ -471,7 +471,15 @@ export class ClrWizard implements OnDestroy, AfterContentInit, DoCheck {
...
+      this.currentPageChanged.emit();
+    });
+  }
+
+  private listenForNavItemChanges(): Subscription {
+    return this.navService.navItemChanged$.subscribe(() => {
+      this.wizardTitle?.nativeElement.focus();
```
Shijir, line 477: Instead of adding another Observable for it, can we make use of the existing `this.navService.currentPageChanged`? Basically, appending this `this.wizardTitle?.nativeElement.focus()` to  the line 475?

Shijir (approved):
I posted some minor questions. Otherwise LGTM

projects/angular/src/wizard/providers/wizard-navigation.service.ts
hippee-lee: good point. This went away when I used the other observable for the same purpose.

PR #6705, opened 2022-02-16
a11y(angular:datagrid): vpat-763

hippee-lee:
This change implements fixes for issues reported in vpat-763 report

1. Adds sibling span with button text and .clr-sr-only class
2. Adds aria-expanded attribute to the toggle button in datagrid footer

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Screen readers do not read out button text for the hid/show feature button controls. There is a missing aria-expanded attribute on the main toggle button in the datagrid footer. 

Issue Number: N/A

## What is the new behavior?
Screen readers read out the button text and for the main toggle button in datagrid footer they read out the aria-expanded attribute. 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-column-toggle.ts
+++ b/projects/angular/src/data/datagrid/datagrid-column-toggle.ts
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-import { Component, Inject, ContentChild, ElementRef, NgZone, PLATFORM_ID, ViewChild } from '@angular/core';
+import { Component, Inject, ContentChild, ElementRef, NgZone, PLATFORM_ID, ViewChild, OnDestroy } from '@angular/core';
 
 import { UNIQUE_ID, UNIQUE_ID_PROVIDER } from '../../utils/id-generator/id-generator.service';
 import { ClrPopoverPosition } from '../../utils/popover/interfaces/popover-position.interface';
@@ -19,6 +19,7 @@ import { ColumnState } from './interfaces/column-state.interface';
 import { DatagridColumnChanges } from './enums/column-changes.enum';
 import { ClrDatagridColumnToggleTitle } from './datagrid-column-toggle-title';
 import { ClrDatagridColumnToggleButton } from './datagrid-column-toggle-button';
+import { Subscription } from 'rxjs';
 
 @Component({
   selector: 'clr-dg-column-toggle',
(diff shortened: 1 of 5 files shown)


projects/angular/src/data/datagrid/datagrid-column-toggle.ts
```diff
@@ -31,9 +32,11 @@ import { ClrDatagridColumnToggleButton } from './datagrid-column-toggle-button';
       clrPopoverOpenCloseButton
       [attr.aria-controls]="popoverId"
       [attr.aria-owns]="popoverId"
+      [attr.aria-expanded]="openState"
       [attr.title]="commonStrings.keys.pickColumns"
     >
       <cds-icon shape="view-columns" aria-hidden="true" [attr.title]="commonStrings.keys.pickColumns"></cds-icon>
+      <span class="clr-sr-only">{{ commonStrings.keys.pickColumns }}</span>
```
Jinnie, line 38: Can you provide some more information why the "title" is not enough and we need to provide the `clr-sr-only` text?

hippee-lee: Not sure, the issue looks to have been raised by usage of the https://www.deque.com/axe/, testing tool. 

Note, this issue pre-dates the move to replace clr-icon with cds-icon which is inert and having the sr-only text as a sibling allows the screen reader to read correct. What we have today is an inert icon but no descriptive text for the button usage.

hippee-lee:
@Jinnie - I updated the title. When reviewing the code changes there was a patch that didn't get applied for the pagination buttons which are also part of this vpat issue.

projects/angular/src/data/datagrid/datagrid-column-toggle.ts
mathisscott: 🤔  Maybe the issue is that we are using `title` instead of `aria-label`? I know that we use `aria-label` on the icons in `@cds/core`. It may be possible that we could replace `title` with `aria-label` and remove the `clr-sr-only` span altogether...

CC:  @scroniser in case I'm confused on anything.

mathisscott (changes requested):

projects/angular/src/data/datagrid/datagrid-column-toggle.ts
scroniser: Yes, that's correct: the `title` isn't read consistently across screen readers. It's also a violation since one can't trigger it using keyboard-only navigation.

`clr-sr-only` text is a better option :)

mathisscott: > Yes, that's correct: the `title` isn't read consistently across screen readers. It's also a violation since one can't trigger it using keyboard-only navigation.
> 
> `clr-sr-only` text is a better option :)

@scroniser

If we changed the `title` to an `aria-label` could we get rid of the `clr-sr-only` span?

hippee-lee: @mathisscott - there was guidance in JIRA to remove the [attr.title]'s. I have updated the PR to remove them and adjust specs accordingly.

mathisscott (approved):

ashleyryan (approved):

PR #6739, opened 2022-03-03
fix(date-picker): better denote selected date

ashleyryan:
- include selected date in button label
- add aria-selected to date in calendar

fixes VPAT-662

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The selected date in the calendar in the modal does not have aria-selected on it after a date has been selected

## What is the new behavior?

Added aria-selected to the button, also changed the label on the button to switch between "Choose date" and "Change date, {date}" when selected.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/forms/datepicker/date-container.spec.ts
+++ b/projects/angular/src/forms/datepicker/date-container.spec.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
@@ -30,6 +30,7 @@ import { ClrPopoverEventsService } from '../../utils/popover/providers/popover-e
 import { ClrPopoverPositionService } from '../../utils/popover/providers/popover-position.service';
 import { ViewManagerService } from './providers/view-manager.service';
 import { IfControlStateService, CONTROL_STATE } from '../common/if-control-state/if-control-state.service';
+import { DayModel } from './model/day.model';
 
 const DATEPICKER_PROVIDERS: any[] = [
   ClrPopoverEventsService,
@@ -54,6 +55,7 @@ export default function () {
     let context: TestContext<ClrDateContainer, TestComponent>;
     let enabledService: MockDatepickerEnabledService;
     let dateFormControlService: DateFormControlService;
+    let dateNavigationService: DateNavigationService;
     let toggleService: ClrPopoverToggleService;
     let container: any;
 
@@ -72,6 +74,7 @@ export default function () {
       enabledService = context.getClarityProvider(DatepickerEnabledService) as MockDatepickerEnabledService;
       dateFormControlService = context.getClarityProvider(DateFormControlService);
       toggleService = context.getClarityProvider(ClrPopoverToggleService);
+      dateNavigationService = context.getClarityProvider(DateNavigationService);
       container = context.clarityDirective;
     });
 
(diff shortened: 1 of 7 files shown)


projects/angular/src/forms/datepicker/date-container.spec.ts
```diff
@@ -224,7 +230,7 @@ export default function () {
 @Component({
   template: `
     <clr-date-container [clrPosition]="position">
-      <input type="date" clrDate [(ngModel)]="model" [disabled]="disabled" />
+      <input type="date" clrDate [(ngModel)]="model" [disabled]="disabled" value="2020-02-01" />
```
ashleyryan: I think I added this to check something, I'm going to remove it.

projects/angular/src/forms/datepicker/date-container.ts
```diff
@@ -175,6 +178,36 @@ export class ClrDateContainer extends ClrAbstractContainer implements AfterViewI
...
+      return (
+        this.commonStrings.parse(this.commonStrings.keys.datepickerToggleChangeDateLabel, {
+          SELECTED_DATE: formattedDate,
+        }) || this.commonStrings.keys.datepickerToggle
+      );
+    }
+    return this.commonStrings.keys.datepickerToggleChooseDateLabel || this.commonStrings.keys.datepickerToggle;
+  }
+
+  private listenForDateChanges() {
```
ashleyryan, line 199: [This line](https://github.com/vmware/clarity/blob/angular/projects/angular/src/forms/datepicker/date-input.ts#L119) is the reason why we get the `ExpressionChangedAfterItHasBeenCheckedError` error.

projects/angular/src/forms/datepicker/date-container.ts
```diff
@@ -175,6 +178,36 @@ export class ClrDateContainer extends ClrAbstractContainer implements AfterViewI
...
 
+  /**
+   * Return the label for the toggle button.
+   * If there's a selected date, the date is included in the label.
+   */
+  private getToggleButtonLabel(day: DayModel): string {
+    if (day) {
+      const formattedDate = this.dateIOService.toLocaleDisplayFormatString(day.toDate());
+
+      return (
```
ashleyryan, line 190: both cases fallback to the old toggle label so it's not breaking if anyone supplied i18n strings

projects/angular/src/forms/datepicker/date-container.ts
```diff
@@ -175,6 +178,36 @@ export class ClrDateContainer extends ClrAbstractContainer implements AfterViewI
...
+    // using a databound attribute to change the button labels results in ExpressionChangedAfterItHasBeenCheckedError.
+    // so instead, update the attribute directly on the element
+    return this.dateNavigationService.selectedDayChange
+      .pipe(startWith(this.dateNavigationService.selectedDay))
+      .subscribe(day => {
+        const label = this.getToggleButtonLabel(day);
+        this.toggleButton.nativeElement.setAttribute('title', label);
+        this.toggleButton.nativeElement.setAttribute('aria-label', label);
+      });
+  }
```
Shijir: I am a bit concerned here as we are accessing DOM API without platform/environment checking. Usually, these kinds of dom changes should be done through [Renderer2](https://angular.io/api/core/Renderer2) API with an additional checking of document being present. You can see an example of how Angular checks whether document is present or not in `datepicker-enabled.service.ts`

ashleyryan: Ah, good catch. I was initially copying what I saw here in `date-input`, but then totally left out the renderer.setProperty part: https://github.com/vmware/clarity/blob/794ec3e19f3f70a91d6eb15944af6b3f98d28a30/projects/angular/src/forms/datepicker/date-input.ts#L278

ashleyryan: I'm looking at the file you mentioned, and I'm not sure how it's relevant? It makes sense it's checking for mobile in that case, since it changes how the popup behaves, but I'm not doing that here, just setting an aria label. 

Is switching to using the renderer sufficient? 

```
 const toggleEl = this.toggleButton.nativeElement;
 this.renderer.setAttribute(toggleEl, 'aria-label', label);
```

Shijir: We check dom `document` object is present or not in order to ensure these dom APIs work, especially in non-browser rendering environments such as SSR. You can see that in `datepicker-enabled.service.ts` we have this check:

```ts
constructor(@Inject(DOCUMENT) private _document: any) {
    if (this._document) {
       ...
    }
}
```

But I think using Renderer APIs might be sufficient without needing that additional document checking.

ashleyryan: Ah ok, I was thinking SSR might be the reason. I was googling to see if `ngAfterViewInit` gets called during SSR - in react lifecycle events after render don't get called then, so wasn't sure it would be an issue.

Thanks for the explanation

Shijir (approved):

bdryanovski (approved):

